PR #1919, opened 2015-01-05
Eng 7384 redundant predicate

alexeevm:
The reason for the redundant predicate was the duplicate call to the m_parsedStmt.m_joinTree.analyzeJoinExpressions
- the first one from the PlanAssembler.setupForNewPlans (l 335)
- the second one from the WriterSubPlanAssembler.nextPlan()

Each call splits the original WHERE (or JOIN) predicate and adds individual expressions into appropriate buckets resulting in dups.

The simple fix is to clone the original m_parsedStmt.m_joinTree inside the WriterSubPlanAssembler.nextPlan() and call the JoinNode.analyzeJoinExpressions on a clone instead of the original.

The better solution is outlined in the comment inside PlanAssembler.setupForNewPlans (ll 326-332) - similar to the SelectSubPlanAssembler.

The commit is
https://github.com/alexeevm/voltdb/commit/c9c01d4300db843714fad22078f024f96d356530

Mike


src/frontend/org/voltdb/planner/WriterSubPlanAssembler.java
```diff
@@ -66,13 +66,13 @@
     AbstractPlanNode nextPlan() {
         if (!m_generatedPlans) {
             // Analyze join conditions
-            m_parsedStmt.m_joinTree.analyzeJoinExpressions(m_parsedStmt.m_noTableSelectionList);
+            assert (m_parsedStmt.m_joinTree != null);
+            JoinNode tableNode = (JoinNode) m_parsedStmt.m_joinTree.clone();
+            tableNode.analyzeJoinExpressions(m_parsedStmt.m_noTableSelectionList);
```
wolffcm: This change looks good.  I've made a copy of your branch in the VoltDB repo and pushed it up to GitHub so it runs a bunch of tests in our continuous integration framework.  

Would you mind adding a regression test for this fix?  `TestPlansDML.java` would be a good place for it.  Just a short test locking down that we no longer create the redundant predicate would be good.  I could easily imagine some well-meaning person removing the call to `clone` thinking the only effect would be to save time and space.

alexeevm: Thanks Chris.
I have added the test case TestPlansDML.testDMLPredicate().
https://github.com/alexeevm/voltdb/blob/de60fc54d5c92cd8db4a461e71a509761180d324/tests/frontend/org/voltdb/planner/TestPlansDML.java
MIke

wolffcm:
This work was merged to master in pull request #1925.

PR #1923, opened 2015-01-06
Fix row count returned by INSERT when DELETE trigger fires

wolffcm:
Apparently for "sendless" DML the VoltDBEngine assumes responsibility for counting the number of rows affected by a plan fragment to return to the client.  I was surprised because I thought that the executors do that.

I changed m_tuplesModified in VoltDBEngine to be a stack (where the top contains the modified count of the most nested plan fragment currently executing), so that we can discard the modified count of the delete trigger.

I also modified a unit test case to test a DELETE trigger on a partitioned table.  Had I done that in the first place I would have found this a lot sooner.


tests/frontend/org/voltdb/regressionsuites/TestSQLFeaturesNewSuite.java
```diff
@@ -502,25 +503,65 @@ public void testLimitPartitionRowsDeleteWithLimit() throws Exception {
...
         VoltTable vt;
         for (int i = 0; i < 50; ++i) {
+            String uuid = UUID.randomUUID().toString();
             vt = client.callProcedure("@AdHoc",
-                    "INSERT INTO events_capped VALUES (NOW, " + i + ")")
+                    "INSERT INTO events_capped VALUES ('" + uuid + "', NOW, " + i + ")")
                     .getResults()[0];
+
+            // Note: this should be *one*, even if insert triggered a delete
             validateTableOfScalarLongs(vt, new long[] {1});
```
jxchinaren: Have you verified this assert used to fail before?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -674,7 +680,11 @@ int VoltDBEngine::executePlanFragment(int64_t planfragmentId,
     m_currExecutorVec = execsForFrag;
 
     int rc = execsForFrag->execute(this);
+
+    int64_t tuplesModified = m_tuplesModifiedStack.top();
+    m_tuplesModifiedStack.pop();
     resetExecutionMetadata();
+
     if (rc != 0) {
```
jxchinaren: Better to change this "0" to "ENGINE_ERRORCODE_SUCCESS".

jxchinaren:
Looks good to me, Chris.

tests/frontend/org/voltdb/regressionsuites/TestSQLFeaturesNewSuite.java
wolffcm: Yes, it was returning "2" when I changed the test to use a partitioned table instead of a replicated one.

PR #1927, opened 2015-01-07
Reduce diffing in DDL processing for performance

jhugg:
Add about 1/3 of a DDL parser and just a bit too much logic to make the XML diffing efficient.


mikeismert:
Haven't waded through the patch yet, but my pre-comment is:

Once you decide to pre-parse the action and the affected table, is there any reason to really keep doing any diffing at all?  You could go back to the previous implementation and just modify the export/partition/DR state for a table directly in the VoltDDLElementTracker.  Is generating the table diff actually winning any potential future benefits anymore?

mikeismert:
To clarify a little more: when I set out earlier, I chose between what it appears you've done now *or* the diffing approach, and opted for the diffing because it seemed like it might be less special-case, and it turned out to mostly be a disaster.  I'm wondering if you've now taken the first option, but kept the disaster, but we still have all the special-case per-action code the diffing was supposed to avoid, so we have the worst of all worlds?

jhugg:
To answer your question, I kept the diffing for the same reason you did, because it's safer. I know the diff path works, and this is mostly a reasonably safe restriction of that path for performance reasons. Had I had more time, I would probably just rip out all of the diffing, but I think that's more than a 2 day fix, which is about what this was. I'm not proud.

jhugg:
I just pushed a bunch of changes to merge better with Izzy's pull and to fix two tests.

src/frontend/org/voltdb/utils/SQLLexer.java
```diff
@@ -87,6 +90,78 @@ public static String extractDDLToken(String sql)
...
+            }
+
+            // cascade/if exists are interesting on alters and drops
+            boolean cascade = false;
+            boolean ifexists = false;
+            if (verb != HSQLDDLInfo.Verb.CREATE) {
+                matcher = DDL_IFEXISTS_OR_CASCADE_CHECK.matcher(ddl);
+                if (matcher.matches()) {
+                    for (int i = 0; i < 3; i++) {
+                        System.out.println(matcher.group(i));
```
mikeismert: Get rid of this debugging output?

src/frontend/org/voltdb/utils/SQLLexer.java
```diff
@@ -87,6 +90,78 @@ public static String extractDDLToken(String sql)
...
+                    }
+                    System.out.println();
```
mikeismert: Here too.

src/frontend/org/voltdb/utils/SQLLexer.java
```diff
@@ -87,6 +90,78 @@ public static String extractDDLToken(String sql)
...
+            ".*$", // all the rest
+            Pattern.CASE_INSENSITIVE | Pattern.MULTILINE | Pattern.DOTALL
+            );
+
+    // does ddl the statement end with cascade or have if exists in the right place
+    private static final Pattern DDL_IFEXISTS_OR_CASCADE_CHECK = Pattern.compile(
+            "^.*?" + // start of line, then anything (greedy)
+            "(\\s+if\\s+exists)?" + // may contain if exists preceded by whitespace
+            "(\\s+cascade)?" + // may contain cascade preceded by whitespace
+            "\\s*;?\\s*" + // then optional whitespace, a single optional semi, then ws
```
mikeismert: Is this going to choke on comments after the statement?:

DROP TABLE FOO CASCADE;  -- nuke it from orbit

src/hsqldb19b3/org/hsqldb_voltpatches/HSQLDDLInfo.java
```diff
@@ -0,0 +1,79 @@
...
+            else {
+                return null;
+            }
+        }
+    }
+
+    public final HSQLDDLInfo.Verb verb;
+    public final HSQLDDLInfo.Noun noun;
+    public final String name;
+    public final String secondName;
```
mikeismert: Perhaps a comment on what slides in here (It's the table an index is on, right?)

mikeismert:
That's all I've got.

PR #1929, opened 2015-01-07
ENG-7324 load/remove classes sqlcmd aliases

wijjo:


examples/voter/ddl.sql
```diff
@@ -61,7 +61,7 @@ AS
 ;
 
 -- Update classes from jar to that server will know about classes but not procedures yet.
-exec @UpdateClasses voter-procs.jar '';
+load classes voter-procs.jar;
```
jhugg: Capital letters maybe?
And can you update the other examples in this folder? (you can skip VoltKV as it has no java).

wijjo: sure

On Wed, Jan 7, 2015 at 10:49 AM, John Hugg <notifications@github.com> wrote:

> In examples/voter/ddl.sql
> <https://github.com/VoltDB/voltdb/pull/1929#discussion-diff-22593311>:
>
> > @@ -61,7 +61,7 @@ AS
> >  ;
> >
> >  -- Update classes from jar to that server will know about classes but not procedures yet.
> > -exec @UpdateClasses voter-procs.jar '';
> > +load classes voter-procs.jar;
>
> Capital letters maybe?
> And can you update the other examples in this folder? (you can skip VoltKV
> as it has no java).
>
> —
> <https://github.com/VoltDB/voltdb/pull/1929/files#r22593311>.
>



-- 


*[image: VoltDBLogoEmail] <http://www.voltdb.com/>*Steve Cooper, Software
Engineer

*VoltDB, Inc.**E: *scooper@voltdb.com

*voltdb.com <http://www.voltdb.com/>*Follow VoltDB [image: twitter]
<https://twitter.com/VoltDB> [image: facebook]
<https://www.facebook.com/VoltDB> [image: linkedin]

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -543,6 +653,46 @@ public static void interactWithTheUser() throws Exception
         return parsedQueries;
     }
 
+    /// Returns the original command, a replacement command, or null (on error).
+    private static String handleTranslatedCommands(String lineIn)
+    {
+        String lineOut = null;
+        try {
+            // LOAD CLASS <jar>?
+            if (lineOut == null) {
```
jhugg: Why not just return when you've found a match? The if guards made me do a double take, especially since the first one can't ever be false.

wijjo: That structure allowed the single debug print statement to cover everything. It's why I switched from inline returns to falling through to a single return point.

jhugg:
So there already has to be code that parses a line and tries to understand what to do with it. For example, we ALREADY trap "exec @UpdateClasses" and modify the args before we call the proc. Why couldn't this code be more integrated with that code?

Also you should try to get Paul/Xin to look at it, as they're on SQLcmd duty this iteration.

wijjo:
I saw what these new commands doing as being more analogous to FILE. because it needs to be parsed and it substitutes one command for other(s). And it needs pattern recognition and reasonable error checking. The @UpdateClasses/@UAC argument manipulation does not do this kind of parsing and error checking. Does that make sense?

I'll try to get other eyes.

jxchinaren:
Looks good to me.

jxchinaren:
As I worked on some missing parsing features on VMC SQL QUERY, I realize that your new LOAD/REMOVE will not work on this new UI.

PR #1930, opened 2015-01-07
Trigger disappears after truncate table

wolffcm:
I noticed that if I do "delete from t" where t has a row limit trigger, then the trigger will disappear afterwards.  It turns out that the persistent table object gets created anew for truncate table, and it needs to get the executor vector for the row limit trigger as well.


wolffcm:
I've a fix for another issue that depends on the fix for this one (ENG-7460).  I'm closing this pull request and will create a new one that fixes both issues.

PR #1931, opened 2015-01-07
ENG-7252: Fix NPE in ReplaySequencer.ReplayEntry.toString().

nshi:


tests/frontend/org/voltdb/iv2/TestReplaySequencer.java
```diff
@@ -111,6 +111,7 @@ public void testOfferSentinelThenFragments()
         TransactionInfoBaseMessage frag2 = makeFragment(1L);
 
         result = dut.offer(1L, sntl);
+        dut.dump(1); // sanity check the toString methods
```
afrogers: Maybe either wrap this and fail on throw, or add a comment that makes it slightly clearer why I shouldn't delete this line? I don't know, probably not a big deal

nshi: I'll add more comment. Without the fix, dump() throws NPE which fails the test.

PR #1935, opened 2015-01-07
Fix ENG-7466, extraneous quotes, by removing toString() call.

johnTcrawford:
This is a trivial, one-line bug, in the json-sessions example, which I noticed while writing the new, updated JSON blog(s): the special user sessions named 'voltdb' and 'voltdb2' were being created with extraneous double-quotes around the last-login value, unlike all the other, randomly generated user sessions, so I removed the .toString() call (which was used for those two user sessions, but not elsewhere), which fixes it.


wolffcm:
Looks good to me.

PR #1938, opened 2015-01-08
Disallow views on subqueries. Improve the error message

alexeevm:
Disallow views on subqueries. Only a single source table is allowed in a view

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -2333,9 +2333,14 @@ public class DDLCompiler {
         int displayColCount = stmt.m_displayColumns.size();
         String msg = "Materialized view \"" + viewName + "\" ";
 
+        if (!stmt.m_fromSubqueryList.isEmpty()) {
+            msg += "with subquery sources is not supported.";
+            throw m_compiler.new VoltCompilerException(msg);
+        }
+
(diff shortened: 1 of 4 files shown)


wolffcm:
Ok, these changes look good to me.  I'll merge to the VoltDB branch for this issue, which will kick off testing.

alexeevm:
Chris,
if this is not too late.

I found a better solution. The AbstractparsedStmt.hasSubquery should be
implemented as

    /**

     * Return true if a SQL statement contains a subquery of any kind

     * @return TRUE is this statement contains a subquery

     */

    public boolean hasSubquery() {

        // This method should be called only after the statement is parsed
and join tree is built

        return !getSubqueries().isEmpty();

    }

I missed that  the AbstractparsedStmt class already has getSubqueries()
method which traverses the JoinTree and collects StmtSubqueryScan.
There is no need for the JoinNode modification. Also I think we should
still keep the AbstractparsedStmt.hasSubquery method even it becomes a
one-liner for the reason it should deal with the subquery expressions in
the future. The AbstractparsedStmt.getSubqueries method only collects the
subqueries from the FROM clause.

Mike

wolffcm:
Hi Mike,

It's not too late.  Makes sense, I'm glad you caught that.  I see that you added another commit to your branch---I'll grab it and put it into the VoltDB branch, and merge when the testing passes.

Thanks!

Chris

PR #1940, opened 2015-01-08
New "Windowing with DDL" Example

wolffcm:
These commits add a new "windowing with DDL" example application to the examples directory.  Most of this app is cloned from the "windowing" example, but significantly simplified by using DDL to limit row count and purge old rows.

Is there a better name for this example than "windowing with DDL"?

Since some of this new example overlaps with the existing "windowing" example, it would be good to simplify the older example at some point, by removing the "inlined deletes" options, since it's mostly covered in this example now.


examples/windowing-with-ddl/deployment.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<deployment>
+    <cluster hostcount="1" sitesperhost="4" kfactor="0" />
```
jpiekos: I believe we default to 8 sites per host, why 4 here?  Also, do we need a deployment file (use default generated one instead)?

examples/windowing-with-ddl/README.md
```diff
@@ -0,0 +1,104 @@
...
+        EXECUTE (DELETE FROM timedata
+                 WHERE update_ts
+                       < TO_TIMESTAMP(SECOND, SINCE_EPOCH(SECOND, NOW) - 30)
+                 ORDER BY update_ts, uuid LIMIT 1500)
+    );
+    PARTITION TABLE timedata ON COLUMN uuid;
+
+The constraint caps each partition of the table to 165000 rows, and also says to execute a DELETE statement if an insert would cause the table to exceed the cap.  In this case, the DELETE statement says to get rid of the oldest rows that are older than 30 seconds, but not more than 1,500 rows at one go.  Being able to define a DELETE statement that helps to enforce a LIMIT PARTITON ROWS constraint is a new feature in version 5.0.
+
+If the insertion rate is 20k rows per second, then 600k rows are produced in 30 seconds.  A cluster with 4 nodes would need to store about 150k rows per partition.  We allow a little bit of leeway to ensure that we never fail to clear out space for new tuples and cap each partition of the table at 165,000 rows.  These settings (insertion rate of 20k per second and a 4-node cluster) are the default for this application.
```
jpiekos: Do you mean 4 sites per host?  This app is 1 node "cluster" with 4 partitions, so 150K * 4 = 600K (per node).  4 nodes would be 2.4M tuples, right?

examples/windowing-with-ddl/deployment.xml
wolffcm: I'm not familiar with generated deployment files.  Will it always choose 8 sites per host?  Or will it make a best guess on the optimal sites per host based on hardware?

I chose 4 sites so that the insertion rate and the table cap make the table store around 30 seconds worth of data.  Number of partitions, insertion rate, and table cap are all inter-related.  If the example is run on a machine with lower specs, I'd like to be reasonably sure that it can keep up with the insertion rate, as well.

examples/windowing-with-ddl/README.md
wolffcm: Yes, you're right, I confused my jargon.  I mean one host, 4 sites per host, so four partitions in all.

examples/windowing-with-ddl/deployment.xml
jpiekos: Makes sense. Check out this doc page for the default for sitesperhost:  http://docs.voltdb.com/UsingVoltDB/ChapAppRun.php

paulmartel:
This looks really good -- clearly and usefully documented. Please merge.

examples/windowing-with-ddl/deployment.xml
jhugg: We tend to de-emphasize using sitesperhost in deployments, except as an advanced option. I would say remove it and use 8. Sorry this review is 800 years late.

PR #1941, opened 2015-01-09
Fix issues with row limit trigger and truncate table

wolffcm:
When writing the example app I found a few issues that had to do with truncating a table:
* When truncating a table, the row limit trigger for a table would disappear.  It needs to be propagated to the new PersistentTable object that gets created.
* When the row limit trigger itself does a truncate table, then the insert executor's pointer to the PersistentTable being modified needs to be updated; otherwise we're just attempting to insert into the old, full table.

### DIFF
--- a/src/ee/executors/insertexecutor.cpp
+++ b/src/ee/executors/insertexecutor.cpp
@@ -125,7 +125,7 @@ bool InsertExecutor::p_init(AbstractPlanNode* abstractNode,
     return true;
 }
 
-bool InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable* table) {
+bool InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable** ptrToTable) {
     InsertPlanNode *insertPlanNode = static_cast<InsertPlanNode*>(getPlanNode());
     if (insertPlanNode->isMultiRowInsert()) {
         // Multi-row inserts triggering a purge is not supported yet.
@@ -134,6 +134,7 @@ bool InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable* table) {
         return true;
     }
 
+    PersistentTable* table = *ptrToTable;
     int tupleLimit = table->tupleLimit();
     int numTuples = table->visibleTupleCount();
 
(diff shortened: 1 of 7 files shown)


src/ee/executors/insertexecutor.cpp
```diff
@@ -270,7 +282,7 @@ bool InsertExecutor::p_execute(const NValueArray &params) {
                 // try to put the tuple into the target table
 
                 if (m_hasPurgeFragment) {
-                    if (!executePurgeFragmentIfNeeded(persistentTable))
+                    if (!executePurgeFragmentIfNeeded(&persistentTable))
```
jxchinaren: oh... Does this mean that UPSERT may delete a tuple when UPSERT is really a UPDATE? If so, this does not seem correctly.

wolffcm: No, this block is entered when we're inserting a row.  The test to see if we found a PK match is line 281.  You had me scared for a second!

src/ee/executors/insertexecutor.cpp
```diff
@@ -151,6 +152,13 @@ bool InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable* table) {
                        tupleLimit);
             return false;
         }
+
+        // If the purge fragment did a truncate table, then the old
+        // table is still around for undo purposes, but there is now a
+        // new empty table we can insert into.  Update the caller's table
+        // pointer to use it.
+        TableCatalogDelegate* tcd = m_engine->getTableDelegate(table->name());
```
jxchinaren: Instead of caching the variable PersistentTable* table and use its name to look up the new table, you can just call function "m_node->getTargetTable()" to get the new table. 

The node has the delegate already and can bypass the treemap look up.

src/ee/executors/insertexecutor.cpp
jxchinaren: I see now. That will make it correctly. 哈哈, thanks.

jxchinaren:
Finish this review round.

src/ee/executors/insertexecutor.cpp
jxchinaren: One more question: do we have UPSERT tests with capped table feature?

wolffcm: Yes, there is a simple test for it in TestSQLFeaturesNewSuite.  It just makes sure that we will delete for inserts, and that we don't delete any rows (even if table is full) for updates.  Maybe there should be more?

wolffcm:
I just pushed a new commit that addresses the map lookup issue.

src/ee/executors/insertexecutor.cpp
jxchinaren: It's better to have that test, but I am also ok to merge now. It's an edge case and separate from this pull request ticket.

jxchinaren:
Looks good to me.

src/ee/executors/insertexecutor.cpp
wolffcm: Ok, thanks Xin.  I can file a ticket for testing UPSERT with ROW LIMIT triggers.

PR #1945, opened 2015-01-12
ENG-6467: Add provenance statement to autogenerated DDL in catalog.

mikeismert:
Do some hacky things to avoid including the timestamp in tests that
care about diffs.


src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -188,8 +190,22 @@ public long getCRC() throws IOException {
             if (e.getKey().equals("buildinfo.txt") || e.getKey().equals("catalog-report.html")) {
                 continue;
             }
-            crc.update(e.getKey().getBytes("UTF-8"));
-            crc.update(e.getValue());
+            // Hacky way to skip the first line of the autogenerated ddl, which
+            // has a date which changes and causes test failures
+            if (e.getKey().equals(VoltCompiler.AUTOGEN_DDL_FILE_NAME)) {
```
wijjo: A regex that matches the timestamp line might be more future-proof. If it matches the regex skip the line otherwise handle it. Probably not a big deal.

wijjo:
Looks fine to me. I just have the one minor comment which you can address or not.

src/frontend/org/voltdb/utils/InMemoryJarfile.java
mikeismert: Yeah, I tried the regex thing to start and then had two problems, so I gave up.  I figure the tests will pick up any unintended consequences if someone mucks with the output and we can deal then.

PR #1946, opened 2015-01-12
Eng 7387 add clusterid and dr flag to deployment

wweiss-voltdb:


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -584,8 +586,31 @@ private String checkModifyWhitelist(final CatalogType suspect,
...
             return null;
-        if (suspect instanceof Cluster && field.equals("drMasterHost"))
+        if (suspect instanceof Cluster && field.equals("drProducerEnabled"))
             return null;
+
+        // Avoid over-generalization when describing limitations that are dependent on particular
+        // cases of BEFORE and AFTER values by listing the offending values.
+        String restrictionQualifier = "";
+        
+        if (suspect instanceof Cluster && field.equals("drClusterId") ||
```
nshi: I don't think we can allow changing cluster ID online if it's gonna be baked into each row for active/active.

wweiss-voltdb: If you look below, ClusterId can only be changed if DR is, was, or will be disabled. In each case a consumer snapshot is manditory after the change. We need this capability to resolve collisions in the id namespace, particularly since all databases after this change will have a default clusterId of 0 when the DR section in the deployment is not specified.

nshi: Do you rewrite all the rows in the EE if the cluster ID changes, even if DR is disabled?

I don't think ID collisions is going to be common. Since cluster ID is a required attribute, why would it default to 0?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -790,6 +790,9 @@ bool VoltDBEngine::loadCatalog(const int64_t timestamp, const string &catalogPay
         m_isELEnabled = true;
     }
 
+    // Set DR flag based on current catalog state
+    m_drStream.m_enabled = catalogCluster->drProducerEnabled();
```
nshi: Need to also do this in `VoltDBEngine::updateCatalog()`.

src/frontend/org/voltdb/NodeDRGateway.java
```diff
@@ -61,6 +59,5 @@
     public abstract void start();
     public abstract void shutdown() throws InterruptedException;
 
-    public abstract void calculateTableSignature(Database catalog);
-
+    public abstract void updateCatalog(boolean calledFromUAP, CatalogContext catalog);
```
nshi: What is the purpose of the parameter `calledFromUAP`?

src/catgen/in/javasrc/CatalogDiffEngine.java
wweiss-voltdb: Because this is a catalog field and the field is there irrespective of whether the deployment file has a DR section. As far as the catalog goes it is an is a change. I don't think each EE performs an independent Diff.

nshi: The cluster ID is gonna be stored with each row for A/A. If cluster ID can change online, every row has to be updated on catalog update.

afrogers: Just starting my review pass on this. Turning on A/A DR would be an operational event. It would, presumably, involve adding the cluster ID/sequence-number-or-timestamp field to every row. Unless we're saying enabling A/A is only supported at cluster start time, which is probably not going to fly. People aren't going to want to take their clusters down. Probably not anway. Or unless we're just going to force this field on everyone, regardless of whether they're using A/A. Which seems like it's also not going to fly.

So as long as DR is not enabled, they don't have this data written with every row anyway. Unless it's the cluster-start-or-nothing option. In which case yeah, cluster ID can't change online. But otherwise, we're going to have to write this to every row anyway, so why not let them change it until we do?

PR #1950, opened 2015-01-13
Fix the test cases

jxchinaren:
Forget to change the test case after changing the code. Do not understand how the tests used to succeed, but it succeeds now(manually tested).


tests/frontend/org/voltdb/utils/TestSqlCmdErrorHandling.java
```diff
@@ -443,26 +443,36 @@ private int callSQLcmdWithErrors(String optionArg, String errorMessage) throws E
...
+                assertTrue(message.contains(errorMessage));
+                return exitValue;
+            } catch (Exception e) {
+                System.err.println("External process (" + commandPath + ") has not yet exited after " + (i+1) * SLEEP + "ms");
+            }
+        }
 
-        return exitValue;
+        // sqlcmd process has not finished, time out this tests
+        return -1;
```
jhugg: Ha. I wrote almost this exact code this morning when looking into this fail.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1497,7 +1497,7 @@ private static void loadStoredProcedures(Map<String,Map<Integer, List<String>>>
 
     private static String extractArgInput(String arg) {
         String[] splitStrings = arg.split("=", 2);
-        if (splitStrings.length < 2) {
+        if (splitStrings[1].isEmpty()) {
```
jhugg: Wouldn't this crash out of bounds if the split returns an array of length 1? Don't you need both protections?

jxchinaren: It's only called when the input has a "=" character, maybe I will add a comment for it. Yes, CLI will be much better.

PR #1954, opened 2015-01-14
ENG-7474: account for bytes written so far for backpressure.

akhanzode:


src/frontend/org/voltcore/network/NetworkDBBPool.java
```diff
@@ -26,19 +26,29 @@
...
+    private static final int SIZE = Integer.getInteger("NETWORK_DBB_SIZE", (1024 * 32));
 
     private final int m_numBuffers;
+    private final int m_allocationSize;
     public NetworkDBBPool(int numBuffers) {
         m_numBuffers = numBuffers;
+        m_allocationSize = SIZE;
+    }
+
+    public NetworkDBBPool(int numBuffers, int allocSize) {
```
vtkstef: remove the public since it is used only by test in the same package

nshi:
Looks good to me.

akhanzode:
squash merged....

PR #1955, opened 2015-01-15
Execute row limit trigger for INSERT and UPSERT INTO ... SELECT

wolffcm:
This was pretty straightforward, except that it occurred to me that we get non-deterministic behavior when doing multi-row inserts into a table with a row limit trigger. So I added checks to guard against that.

### DIFF
--- a/src/frontend/org/voltdb/planner/PlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/PlanAssembler.java
@@ -354,10 +354,21 @@ public class PlanAssembler {
         }
 
         if (parsedStmt instanceof ParsedInsertStmt  && !plan.isOrderDeterministic()) {
+            ParsedInsertStmt parsedInsert = (ParsedInsertStmt)parsedStmt;
+            boolean targetHasLimitRowsTrigger = parsedInsert.targetTableHasLimitRowsTrigger();
+
             if (parsedStmt.m_isUpsert) {
                 throw new PlanningErrorException(
                         "UPSERT statement manipulates data in a non-deterministic way.  "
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -354,11 +354,22 @@ private static void failIfNonDeterministicDml(AbstractParsedStmt parsedStmt, Com
...
                         + "Adding an ORDER BY clause to UPSERT INTO ... SELECT may address this issue.");
             }
+            else if (targetHasLimitRowsTrigger) {
+                throw new PlanningErrorException(
+                        "Order of rows produced by SELECT statement in INSERT INTO ... SELECT is "
+                        + "non-deterministic.  Since the table being inserted into has a row limit "
+                        + "trigger, the SELECT output must be ordered.  Add an ORDER BY clause "
+                        + "to address this issue."
+                        );
+            }
```
paulmartel: Minor: It is possible in this case (as it was/is in the similar surrounding cases) that the user has provided an insufficient ORDER BY clause. suggesting that they "Add an ORDER BY clause" may be confusing in this case. We should either A) test for the presence of an ORDER BY and give more accurate advice "Extend the ORDER BY clause." or maybe "Strengthen the ORDER BY clause" or B) give conditional advice like "Add or strengthen an ORDER BY clause". or C) advise "Add ORDER BY columns." which sort of implies adding ORDER BY if there is none and adding columns if there is. Ask Andrew?

paulmartel:
Looks good with just the minor message wording issue. Your call -- fix now or log an "improve error message" ticket and move on.

wolffcm:
Thanks, Paul.  I'll log a ticket for the messages.

PR #1957, opened 2015-01-15
ENG-7426 less verbose sqlcmd feedback on successful ddl

paulmartel:


wolffcm:
This looks reasonable to me.  What does the output of `echo ddl.sql | sqlcmd` for voter look like now?

Also, it seems like it might be worth adding testing locking down the output?

wolffcm:
I mean, add testing at some point, not necessarily in this submission.

PR #1964, opened 2015-01-19
ENG-7372: Make an error message a little more precise/helpful.

mikeismert:


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -472,7 +472,7 @@ else if (isTableLimitDeleteStmt(suspect)) {
             String nullness = column.getNullable() ? "NULL" : "NOT NULL";
             retval[0] = parent.getTypeName();
             retval[1] = String.format(
-                    "Unable to add %s column %s because table %s is not empty.",
+                    "Unable to add %s column %s because table %s is not empty and no default value was specified.",
```
paulmartel: I think that I am just bringing up a missed opportunity to trim out misleading dead code brought to light by the new wording, but...

I'm pretty sure that this code path can never be executed when (column.getNullable() == true),
so the message could be hard-coded to ALWAYS start "Unable to add NOT NULL column ".

My understanding of sql -- VoltDB included, as I have just confirmed by manual test -- is that "NULL is the default for nullable column default values" -- IOW, for nullable columns "default null" is 100% redundant -- equivalent with specifying no default.

We shouldn't ever be refusing to add a nullable column just because the table has existing rows.

mikeismert: Yeah, I had the same thought when I read the code but just less motivation.  I'm happy to kill the "NULL" part as part of this change, though.

mikeismert: Aaaaand, gone.

PR #1975, opened 2015-01-20
Display Index usage information for limit partition row delete statement

jxchinaren:


src/frontend/org/voltdb/compilereport/IndexAnnotation.java
```diff
@@ -30,6 +30,7 @@
  *
  */
 public class IndexAnnotation {
+    // the statement is only used by LIMIT PARTITION ROW DELETE statement
```
wolffcm: Was the SortedSet<Statement> unused prior to now?  Maybe change the comment to "statements are only inserted into this set when the index is used in a LIMIT PARTITION ROWS trigger."

src/frontend/org/voltdb/compilereport/ReportMaker.java
```diff
@@ -138,6 +138,10 @@ static String genrateIndexRow(Table table, Index index) {
                 sb.append(StringUtils.join(procs, ", "));
                 sb.append("</p>");
             }
+            if (annotation.statementsThatUseThis.size() > 0) {
+                assert(annotation.statementsThatUseThis.size() == 1);
+                sb.append("<p>Used by the LIMIT PARTITION ROWS " + table.getTuplelimit() + " Statement</p>");
```
wolffcm: Maybe consult with Andrew on what the message should look like here?  Maybe something like, "Used by the table's LIMIT PARTITION ROWS DELETE statement."

paulmartel: Agreed. Having the number here doesn't really add anything -- the full
definition appears just above in the report.

*Paul Martel*, Software Engineer
*VoltDB, Inc.*
*E: *pmartel@voltdb.com
Follow VoltDB [image: twitter] <https://twitter.com/VoltDB> [image:
facebook] <https://www.facebook.com/VoltDB> [image: linkedin]

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1593,21 +1593,38 @@ private static void updateIndexUsageAnnotation(Index index, Statement stmt) {
...
+            stmt.setAnnotation(sa);
+        }
+        sa.indexesUsed.add(index);
+
         IndexAnnotation ia = (IndexAnnotation) index.getAnnotation();
         if (ia == null) {
             ia = new IndexAnnotation();
             index.setAnnotation(ia);
         }
-        ia.statementsThatUseThis.add(stmt);
```
paulmartel: It looks like this action was dropped -- I don't know why. Do we now have less information in the report?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1593,21 +1593,38 @@ private static void updateIndexUsageAnnotation(Index index, Statement stmt) {
...
+
+        assert(parent instanceof Procedure);
         Procedure proc = (Procedure) parent;
+
         // skip CRUD generated procs
         if (proc.getDefaultproc()) {
             return;
         }
 
+        StatementAnnotation sa = (StatementAnnotation) stmt.getAnnotation();
```
wolffcm: Should we also annotate the statement for the LIMIT PARTITION ROWS DELETE statement as well?

wolffcm:
I'm finished my review round.

src/frontend/org/voltdb/utils/CatalogUtil.java
paulmartel: Annotating the LIMIT PARTITION ROWS DELETE statement would only be useful if we also made a place to show the annotation in the report page. That same place could also provide an opportunity to show the explained plan for the delete statement. The latter should be made available in the report SOMEWHERE.
I can't decide if it would be overkill to redisplay this info as a kind of footnote on each and every insert/upsert statement in each stored procedure. I'm thinking that it could follow the normal information for that statement and be prefixed by something like 

    And optionally, to prevent violation of <table>'s LIMIT PARTITION ROWS constraint:

We should make new tickets for "explain the LIMIT PARTITION DELETE statement plan somewhere in the catalog report" and "display the index (if any) used by a LIMIT PARTITION DELETE statement in some obvious place in the catalog report -- not just as an annotation on the index".

src/frontend/org/voltdb/utils/CatalogUtil.java
jxchinaren: statementsThatUseThis in the IndexAnnotation is not used anywhere actually before, someone tries to make the code forward compatible.

src/frontend/org/voltdb/utils/CatalogUtil.java
wolffcm: Those two things could be the same ticket?  Doesn't the index usage for a statement appear with its explained plan?  I had thought that explaining the limit delete statement in VMC would be part of ENG-7024, but reading the description now it just mentions displaying the SQL text.

PR #1976, opened 2015-01-21
ENG-7567: Rework how the report gets cross-ref data now that statements ...

jhugg:
...are cached and not always compiled.

Move statement annotations into the catalog itself.
Generate the other references at report generation time.


src/frontend/org/voltdb/compilereport/ReportMaker.java
```diff
@@ -375,34 +379,69 @@ static String genrateStatementRow(Procedure procedure, Statement statement) {
...
-            }
-            for (Index i : annotation.indexesUsed) {
-                Table t = (Table) i.getParent();
-                String name = i.getTypeName().toUpperCase();
-                String link = "\"<a href='#s-" + t.getTypeName() + "-" + i.getTypeName() +"'>" + name + "</a>\"";
-                plan = plan.replace("\"" + name + "\"", link);
-            }
 
-            sb.append("<tt>").append(plan).append("</tt>");
+        String plan = escapeHtml4(Encoder.hexDecodeToString(statement.getExplainplan()));
```
wijjo: Not critical, but is this the best location for all of this code? A lot of it seems more general than report generation. It would be nice if somehow we could just see the HTML-building stuff here.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -2416,7 +2416,14 @@ Statement getCachedStatement(String keyPrefix, String sql) {
         }
 
         // check that no underlying tables have been modified since the proc had been compiled
-        String tablesTouched[] = candidate.getTablestouched().split(",");
+        String[] tablesTouched = candidate.getTablesread().split(",");
```
wijjo: Another style comment... Perhaps the split is better done in Statement? Yes, I know this is the way it was.

src/frontend/org/voltdb/compilereport/ReportMaker.java
```diff
@@ -529,56 +568,64 @@ static String generateProcedureRow(Procedure procedure) {
...
-                sb.append("<p>Read-only access to tables: ");
-                List<String> tables = new ArrayList<String>();
-                for (Table table : annotation.tablesRead) {
-                    tables.add("<a href='#s-" + table.getTypeName() + "'>" + table.getTypeName() + "</a>");
-                }
-                sb.append(StringUtils.join(tables, ", "));
-                sb.append("</p>");
+        if (annotation == null) {
+            annotation = new ProcedureAnnotation();
+            procedure.setAnnotation(annotation);
```
wijjo: Again, a little odd to do the annotation update from a report generator. Would it be better in a Procedure method that gets and sets as needed?

src/frontend/org/voltdb/compilereport/ReportMaker.java
jhugg: It's not ideal. I think there's some cleanup work that could be done in a second pass while making the HTML generation faster too. I have some thoughts.

Still, all of the annotation stuff being computed here is used solely for generating this HTML, so it's not too far off base?

src/frontend/org/voltdb/compiler/VoltCompiler.java
jhugg: Statement is a CatalogType subclass and since it's auto generated, can't do special stuff like this. It *could* contain a list of references, but that kills performance and is pretty complex.

src/frontend/org/voltdb/compilereport/ReportMaker.java
wijjo: It's a for-what-it's-worth type comment. Opportunistic cleanup later would be fine. Would it be reasonable to add a comment with the suggestion?

src/frontend/org/voltdb/compiler/VoltCompiler.java
wijjo: I was wondering if that was the explanation. No need for non-trivial refactoring just for this.

PR #1978, opened 2015-01-22
Remove erroneous assertion from EE

wolffcm:
This was causing a failed assertion when running sqlcoverage.  When searching for a value in an index that is larger than the index's component types (e.g., I have a bigint 1000 that I'm searching for in a tinyint index), then an assertion in the EE requires more than one component in the index to be specified.  Now, that I understand the code a bit better, I don't think there's any reason this needs to be the case, so I just removed the assertion.

### DIFF
--- a/src/ee/executors/nestloopindexexecutor.cpp
+++ b/src/ee/executors/nestloopindexexecutor.cpp
@@ -312,10 +312,6 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
                     if ((localLookupType != INDEX_LOOKUP_TYPE_EQ) &&
                         (ctr == (activeNumOfSearchKeys - 1))) {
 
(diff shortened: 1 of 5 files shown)


paulmartel:
The change looks good and very low risk. Yet, AFAIK, master is still officially frozen at the moment.
Tomorrow, I will look into getting a waiver, or getting the release branch split off master, or getting a CORE integration branch up to rev with master, so we can merge and close this.

wolffcm:
No need to make an exception for this.  This was only an issue in a debug build, because it was a failed assertion.  Since the fix was to remove the assertion, this change has zero effect for the release build.  I've attempted to reproduce with -Dbuild=release and I cannot.

wolffcm:
After discussion with Paul and Phil, we've decided to merge this since it has no effect on release build code, and has the benefit of adding more testing.

PR #1979, opened 2015-01-22
ENG-7515: Removing catalogs from tutorial examples.

nshi:


nshi:
I squashed the commits to address John's review comments together.

PR #1981, opened 2015-01-22
Eng 7574 sqlcmd semicolons

paulmartel:
Killed the annoying sqlcmd semi-colon insertion and tab completion mis-features which just get more out-of-date and ugly and downright dangerous as we push forward with DDL support.
Also addressed some error message uglies brought to light when this fix was tested -- mostly manually.
Removed or adjusted unit test cases that relied on implied semi-colons, depending on whether that seemed to be their focus.
More automated sqlcmd testing is coming soon with a new data-driven python framework.
There still seem to be lingering issues in how input tabs are handled (swallowed?) but at least this change eliminates one possible source of these issues.


jxchinaren:
Great to remove 'features'.

There are still test failures, FYI, 
1) junit.framework.AssertionFailedError
	at org.voltdb.TestAdHocQueries.testXopenSubSelectQueries(TestAdHocQueries.java:673)
2) junit.framework.AssertionFailedError: skipped a file-scripted post-error write
	at org.voltdb.utils.TestSqlCmdErrorHandling.test70BadNestedFileWithWriteThenWrite(TestSqlCmdErrorHandling.java:323)

src/hsqldb19b3/org/hsqldb_voltpatches/Statement.java
```diff
@@ -212,7 +212,7 @@ public ResultMetaData getParametersMetaData() {
     VoltXMLElement voltGetStatementXML(Session session)
     throws org.hsqldb_voltpatches.HSQLInterface.HSQLParseException
     {
-        throw new RuntimeException("Unsupported SQL verb in statement: \""+sql+"\"");
+        throw new org.hsqldb_voltpatches.HSQLInterface.HSQLParseException("a DDL statement is not allowed in this context");
```
paulmartel: One of the two jenkins failures is telling me that I have to be more generic in the wording here -- I was guessing that it was specific to DDL in a procedure context, but it comes up elsewhere.

wolffcm:
If the call to ParserStringKeywords pattern isn't needed, maybe it could be removed, but perhaps you *won't* want to remove it to minimize risk?

I don't see any red flags here.  I'm glad that this is getting done finally.  That's all my comments.

paulmartel:
Acknowledged: Removing ParserStringKeywords was an oversight but the effect is negligible, so in spite of the fact that I have a patch for that (and other small changes, mostly re: your feedback) I want to submit the branch AS IT STANDS to 5.0.  Ruth has reminded me that VMC has additional safeties built in, so I was wrong to want to push the comparable VMC change into 5.0. I HAVE added a commit to this branch/pull request to address the jenkins regressions. I'd appreciate a quick review of that. If I don't hear objections in the meantime, I'd like to merge this branch before I start my commute around 11:00.
It troubles me that sqlcmd with its (supposed?) default of --stop-on-error=true doesn't have the same level of safety that VMC apparently does which is why it urgently requires this patch. This unexpected behavior on error needs to be its own ticket for 5.1.

PR #1983, opened 2015-01-23
Eng7573

mikeismert:
There's a matching pro pull request, mostly test cleanups.

### DIFF
--- a/src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
+++ b/src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
@@ -27,6 +27,7 @@ import java.util.TimerTask;
 
 import org.voltdb.common.Constants;
 import org.voltdb.planner.BoundPlan;
+import org.voltdb.utils.Encoder;
 
 import com.google_voltpatches.common.cache.Cache;
 import com.google_voltpatches.common.cache.CacheBuilder;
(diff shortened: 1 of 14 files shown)


src/frontend/org/voltdb/iv2/MpRoSite.java
```diff
@@ -150,7 +150,7 @@ public long getCatalogCRC() {
 
         @Override
         public byte[] getCatalogHash() {
-            throw new RuntimeException("Not needed for RO MP Site, shouldn't be here.");
+            return m_context.getCatalogHash();
```
jxchinaren: Looking at the API functions got override before and after, they are all throwing runtime exception. Why this one is special? Add a comment or include the ticket number if it is too lengthy to be explained.

mikeismert: Not all, getCatalogVersion() actually returns a value, for just this purpose.

paulmartel:
Looks good.

src/frontend/org/voltdb/iv2/MpRoSite.java
mikeismert: And it had a comment...added one for getCatalogHash, too.

src/frontend/org/voltdb/planner/CorePlan.java
```diff
@@ -153,8 +154,8 @@ public int getSerializedSize() {
         else {
             size += 4;
         }
-        size += 3; // booleans
-        size += 4; // catalog version
+        size += 2; // booleans
+        size += 20;  // catalog hash SHA-1 is 20b
```
jxchinaren: This looks fine now. However, adding a new constant variable like CONSTANT SIZE_OF_HASH_IN_BYTES will help to read, especially the 20 constant has been used a couple of times above.

src/frontend/org/voltdb/planner/CorePlan.java
```diff
@@ -210,8 +211,9 @@ public static CorePlan fromBuffer(ByteBuffer buf) throws IOException {
         boolean isReplicatedTableDML = buf.get() == 1;
         boolean isReadOnly = buf.get() == 1;
 
-        // catalog version
-        int catalogVersion = buf.getInt();
+        // catalog hash
+        byte[] catalogHash = new byte[20];  // Catalog sha-1 hash is 20b
```
jxchinaren: This make my previous constant variable comment more true.

mikeismert: This is true in many more places that just this file, so the scope of a real fix is larger than just adding a constant to CorePlan.  Really, I think we need a hash class, because there's a lot of other code that we repeat when manipulating SHA-1 hashes.  Would a ticket to do this at some point in the future be sufficient for now?

jxchinaren: I agree. Looks good.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -50,23 +51,24 @@
     // STATIC CODE TO MANAGE CACHE LIFETIMES / GLOBALNESS
     //////////////////////////////////////////////////////////////////////////
 
-    // weak values should remove the object when the catalog version is no longer needed
-    private static Cache<Integer, AdHocCompilerCache> m_catalogVersionMatch =
+    // weak values should remove the object when the catalog hash is no longer needed
+    private static Cache<String, AdHocCompilerCache> m_catalogHashMatch =
```
vtkstef: Consider using BigInteger. 20 byte BigInteger lookups will be faster than a 40 character string lookup

mikeismert: This is called once per catalog update, I'm not too worried about it.

mikeismert:
Squashed and merged elsewhere

PR #1986, opened 2015-01-23
ENG-6842 REVIEW ONLY

afrogers:


src/ee/common/types.h
```diff
@@ -453,7 +453,9 @@ enum RecoveryMsgType {
 // ------------------------------------------------------------------
 enum TaskType {
     TASK_TYPE_VALIDATE_PARTITIONING = 0,
-    TASK_TYPE_APPLY_BINARY_LOG = 1
+    TASK_TYPE_APPLY_BINARY_LOG = 1,
+    TASK_TYPE_GET_DR_SEQUENCE_NUMBER = 2,
+    TASK_TYPE_SET_DR_SEQUENCE_NUMBER = 3
```
nshi: If you are landing my stuff before yours, you probably need another pair for the MP sequence number.

src/ee/storage/DRTupleStream.cpp
```diff
@@ -274,13 +264,13 @@ void DRTupleStream::endTransaction(int64_t spUniqueId) {
          extendBufferChain(END_RECORD_SIZE);
      }
 
-     m_currBlock->lastSpUniqueId(spUniqueId);
+     m_currBlock->recordCompletedTxnForDR(sequenceNumber, uniqueId);
```
nshi: Maybe consider having consistent ordering of these two things (sequenceNumber and uniqueId) in signatures.

afrogers: Yeah, that's part of the cleanup I'm working on. Sequence number should be first, since it's the primary identifier, but I wrote some of this before we decided to make it the primary identifier.

src/ee/common/types.h
afrogers: Yep. I think you landing first is still probably the best course.

PR #1987, opened 2015-01-24
Remove the newline and more stuffs; fix the test case failure and add tw...

jxchinaren:
...o more tests.

### DIFF
--- a/tests/frontend/org/voltdb/utils/TestSqlCommandParserInteractive.java
+++ b/tests/frontend/org/voltdb/utils/TestSqlCommandParserInteractive.java
@@ -201,8 +201,29 @@ public class TestSqlCommandParserInteractive extends TestCase {
         cmd.submitText("; --select * from goats;\n");
         cmd.waitOnResult();
         System.out.println("RESULT: " + result.get());
-        assertEquals(2, result.get().size());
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/utils/TestSqlCommandParserInteractive.java
```diff
@@ -201,9 +201,30 @@ public void testComments() throws Exception
...
+                );
+        cmd.waitOnResult();
+        System.out.println("RESULT: " + result.get());
+        assertEquals(1, result.get().size());
+
+        cmd.submitText("select * from T; -- select * from T;\n");
+        cmd.waitOnResult();
+        System.out.println("RESULT: " + result.get());
+        assertEquals(1, result.get().size());
+
```
paulmartel: Test cases like these, at least the cases with comments INSIDE a multi-line statement should be added to TestSqlCmdInterface to verify that all comments (and only comments) get stripped out. The statement count assert here is very weak for those cases, and long-term nobody is going to inspect the System.out when there is no assert failure to draw their attention -- which is why I always advise disabling most junit System.out calls except when debugging, or making it conditional to only print when an assert is going to fail.

PR #1988, opened 2015-01-24
Stripe out all the -- comments

jxchinaren:


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -152,6 +152,23 @@ public static Pattern getExecuteCall() {
             i++;
         }
 
+        int commentBegin = 0;
+        while ( (commentBegin = query.indexOf("--")) > 0) {
+            // stripe out inline dash dash comment
```
paulmartel: strip and stripe are not the same word.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -152,6 +152,23 @@ public static Pattern getExecuteCall() {
...
+
+            int commentEnd = query.indexOf("\n", commentBegin);
+            if (commentEnd > commentBegin) {
+                queriesAfterInlineComment = query.substring(commentEnd);
+            }
+
+            // concat the new query
+            query = queriesBeforeInlineComment + queriesAfterInlineComment;
+        }
+
```
paulmartel: Unless/until we need to get into really complex processing like quote-counting, I suspect that this change could be greatly simplified using regexp code -- possibly a one-liner -- by someone more skilled with regexp than me.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -152,6 +152,23 @@ public static Pattern getExecuteCall() {
             i++;
         }
 
+        int commentBegin = 0;
+        while ( (commentBegin = query.indexOf("--")) > 0) {
+            // stripe out inline dash dash comment
+            // this may have problem if the user has quoted data "--".
+            // Release note that SQLCMD does not support user quoted data dash dash.
```
paulmartel: Sorry about the long comment.
If we want to have a wider discussion, I can cut and paste it into an email.

One of the things I liked very much about your previous changeset was that it did not seem to have this problem with '--' in string literals. I am sad to see that go. It makes me uncomfortable to introduce this known failure with no known user workaround.

One possibility is that we make this code smarter by skipping any '--' that is preceded IN THE SAME LINE by an odd number of single quotes. The code would have to be smart about continuing to scan the line for another --. The only case that this would NOT catch is a valid (in sqlcmd anyway) multi-line string like this:

        col1 VARCHAR(1000) DEFAULT 'A string value started here that contains
    this non-comment -- on its own line with no preceding quote', -- and maybe a comment HERE!

And maybe that's too much logic to be adding to 5.0?

The version of the fix that I gave up on was also going to give up on '--' strings, BUT it included a workaorund.
An undocumented command line switch meant "for this sqlcmd session, the user promises to follow a few restrictions in where and how they use '--' comments, and what kind of punctuation they put into their '--' comments. In exchange, sqlcmd will just skip any risky steps like THIS one so the user can safely use '--' in a quoted string". That lets users (after they learn the secret) to combine in the same sql file '--' strings and MOST cases of '--' comments.

The OTHER possible workaround -- so maybe we would not even need a command line option -- is if hsql supports some way of disguising the ascii character '-'. This COULD be as simple as '\-' so users could be told to break up any '--' inside their strings by using something like '-\-'. Or maybe hsql supports some other escape sequence that uses numeric ascii or unicode values. I have never researched this question. It MIGHT be a "free" workaround that we can just release note. 
I don't know how hard it would be to ADD something like this to hsql's string parser if it was missing -- even just support for the special cases '\-' => '-' and '\\' => '\' (to escape the escape so that '\\-' => '\-').

Another alternative: we do support string EXPRESSIONS in the place of string constants so the user could break up

    'my string with a -- in the middle'

into

    'my string with a -' || '- in the middle'

though this could result in slower queries, unless/until we changed hsql or the planner to "fold the constant". We could fix that in 5.1. 
Also, this does not work for 2 cases:
  1) DDL default values must be constants or NOW
  2) "exec procedure" parameters must be constants
We could also fix these cases to specifically support ("fold") the || operator in string constants. Is this too far to go just for a workaround?

tests/frontend/org/voltdb/utils/TestSqlCmdInterface.java
```diff
@@ -525,4 +525,19 @@ private void notAssertThis2(String query, String cleanQryStr, int num, int testI
         err2 += "Actual queries: \n#" + parsedString + "#\n";
            assertFalse(msg+err2, cleanQryStr.equalsIgnoreCase(parsedString));
     }
+
+   // assert for inline comment query
+   private void assertThis3(String qryStr, String cleanQryStr, int numOfQry, int testID) {
+       List<String> parsed = SQLCommand.parseQuery(qryStr);
```
paulmartel: This new assert... helper function suffers even more than the others in this code from deflated 3- or 7- space indentation.

paulmartel: Actually, this helper seems to be identical in behavior to the assertThis overload with the same signature.
Remove it?

src/frontend/org/voltdb/utils/SQLCommand.java
jxchinaren: Haha, good to know that.

src/frontend/org/voltdb/utils/SQLCommand.java
jxchinaren: All these options or workaround do not solve the problem completely, maybe someone should make a decision about which one we really want for V5.0.

tests/frontend/org/voltdb/utils/TestSqlCmdInterface.java
jxchinaren: The main difference is :
1) assertThis:
assertEquals(msg+err1, numOfQry + blockCommentCount, parsed.size());
if (blockCommentCount == 0) {
            // If there is sql comments in block(s), then skip the assertion below
            assertTrue(msg+err2, cleanQryStr.equalsIgnoreCase(parsedString));
        }
        else {
            assertFalse(msg+err2, cleanQryStr.equalsIgnoreCase(parsedString));
        }
2) assertThis3:
assertEquals(msg+err1, numOfQry, parsed.size());
assertFalse(msg+err2, cleanQryStr.equalsIgnoreCase(parsedString));

The other way is to change the assertThis API.

paulmartel: You are comparing your assertThis3 to the assertThis overload with the 5 arguments.
And of course they are different -- your assertThis3 uses only 4 arguments.
I am comparing your assertThis3 to the assertThis overload with the SAME 4-argument signature. It does not have this API difference, and (I believe) it does not have a behavior difference with assertThis3 AND it is one line of code!

src/frontend/org/voltdb/utils/SQLCommand.java
paulmartel: I just looked at the code again. At the point where this change runs, ALL the quoted strings have been pulled out of the code because they may contain semicolons -- and they will not be restored until after the split. I think dash dash strings may be safe! I will test that.

paulmartel: I tested it. It works. Please remove the scary comment.

src/frontend/org/voltdb/utils/SQLCommand.java
paulmartel: I think I found the regexp "one-liner" to do this:

            query = MidlineComments.matcher(query).replaceAll("");

This works after defining

        private static final Pattern MidlineComments = Pattern.compile("(\\/\\/|--).*$", Pattern.MULTILINE);

tests/frontend/org/voltdb/utils/TestSqlCmdInterface.java
```diff
@@ -158,8 +158,8 @@ public void testParseQuery9() {
     public void testParseQuery10() {
         String raw = "SELECT * FROM table -- SELECT * FROM table2";
         ID = 10;
-        String expected = raw;
-        assertThis(raw, expected, 1, ID);
+        String expected = "SELECT * FROM table";
+        assertThis3(raw, expected, 1, ID);
```
paulmartel: There should be a test case forcing multiple mid-line comment substitutions like

    select -- comment no semicolon
    * -- comment no semicolon
    from -- comment no semicolon
    table -- comment with semicolon;

one should have a semi-colon inside the comment


    select -- comment no semicolon
    * -- comment with this ; a semicolon inside
    from -- comment with this ; a semicolon inside
    table -- comment with semicolon;

tests/frontend/org/voltdb/utils/TestSqlCmdInterface.java
jxchinaren: I must be so dumb at that night. Yes, you are right. There is one overloaded function available. I will get rid of my assertThis3 and use assertThis directly.

PR #1992, opened 2015-01-27
Add more unit tests for IN with params

wolffcm:
These are the unit tests I wrote to try and reproduce a user-reported issue that was closed as "not a bug."

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
@@ -1858,6 +1858,18 @@ public class TestFixedSQLSuite extends RegressionSuite {
                 new long[][] {});
     }
 
+    private void insertForInParamsTests(Client client) throws Exception {
+        for (int i = 0; i < 10; ++i) {
+            VoltTable vt = client.callProcedure("P1.insert",
+                    i, Integer.toString(i), i * 10, i * 100.0)
+                    .getResults()[0];
+            validateTableOfScalarLongs(vt, new long[] {1});
+        }
+    }
+
+    // Note: the following tests for IN with parameters should at some point
+    // be moved into their own suite along with existing tests for IN
+    // that now live in TestIndexesSuite.  This is ENG-7607.
     public void testInWithIntParams() throws Exception {
 
         // HSQL does not support WHERE f IN ?
@@ -1865,20 +1877,17 @@ public class TestFixedSQLSuite extends RegressionSuite {
             return;
 
         Client client = getClient();
-
-        for (int i = 0; i < 10; ++i) {
-            VoltTable vt = client.callProcedure("P1.insert",
-                    i, Integer.toString(i), i * 10, i * 100.0)
-                    .getResults()[0];
-            validateTableOfScalarLongs(vt, new long[] {1});
-        }
+        insertForInParamsTests(client);
 
         VoltTable vt = client.callProcedure("one_list_param", new int[] {1, 2})
                 .getResults()[0];
         validateTableOfScalarLongs(vt, new long[] {1, 2});
 
-        // param has to be an int array, not long array
-        // Also, this error message isn't that friendly
+        // The following error message characterizes what happens if the
+        // users passes long array to an IN parameter on an INTEGER column.
+        // VoltDB requires that the data types match exactly here.
+        //
+        // This error message isn't that friendly (ENG-7606).
         verifyProcFails(client,
                 "tryScalarMakeCompatible: "
                 + "Unable to match parameter array:int to provided long",
@@ -1899,45 +1908,25 @@ public class TestFixedSQLSuite extends RegressionSuite {
 
     }
 
-    static private final class SimpleCallback implements ProcedureCallback {
-
-        private ClientResponse m_clientResponse = null;
-
-        @Override
-        public void clientCallback(ClientResponse clientResponse) throws Exception {
-            m_clientResponse = clientResponse;
-        }
-
-        public ClientResponse getClientResponse() {
-            return m_clientResponse;
-        }
-    }
-
     public void testInWithStringParams() throws Exception {
         if (isHSQL())
             return;
 
         Client client = getClient();
-
-        for (int i = 0; i < 10; ++i) {
-            VoltTable vt = client.callProcedure("P1.insert",
-                    i, Integer.toString(i), i * 10, i * 100.0)
-                    .getResults()[0];
-            validateTableOfScalarLongs(vt, new long[] {1});
-        }
+        insertForInParamsTests(client);
 
         String[] stringArgs = {"7", "8"};
 
         // For vararg methods like callProcedure, when there is an array of objects
-        // (not an array of native types) passed as the last argument, the
-        // compile time type affects how the compiler presents the arguments to the
+        // (not an array of native types) passed as the only vararg argument, the
+        // compile-time type affects how the compiler presents the arguments to the
         // callee:
         //    cast to Object   - callProcedure sees just one param (which is an array)
         //    cast to Object[] - (or a subclass of Object[]) callee sees each array
-        //                       element as its own parameter
+        //                       element as its own parameter value
 
         // where desc in ?
-        // Cast it as an object and it's treated as a single paramter in the callee.
+        // Cast parameter value as an object and it's treated as a single parameter in the callee.
         VoltTable vt = client.callProcedure("one_string_list_param", (Object)stringArgs)
                 .getResults()[0];
         validateTableOfScalarLongs(vt, new long[] {7, 8});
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+                    i, Integer.toString(i), i * 10, i * 100.0)
+                    .getResults()[0];
+            validateTableOfScalarLongs(vt, new long[] {1});
+        }
+
+        VoltTable vt = client.callProcedure("one_list_param", new int[] {1, 2})
+                .getResults()[0];
+        validateTableOfScalarLongs(vt, new long[] {1, 2});
+
+        // param has to be an int array, not long array
```
paulmartel: I don't understand the point of this comment. What/who wants/expects a long array & why? What/who says it has to be an int array? Is this expressing some known limitation? Some minor bug that should have a ticket?

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+        // Also, this error message isn't that friendly
```
paulmartel: Please file a minor "unfriendly message" ticket -- possibly suggest a better alternative wording there.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+
+        Client client = getClient();
+
+        for (int i = 0; i < 10; ++i) {
+            VoltTable vt = client.callProcedure("P1.insert",
+                    i, Integer.toString(i), i * 10, i * 100.0)
+                    .getResults()[0];
+            validateTableOfScalarLongs(vt, new long[] {1});
+        }
+
```
paulmartel: Is this common setup code worth refactoring?

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+        assertTrue(b);
+        client.drain();
+        assertEquals(ClientResponse.GRACEFUL_FAILURE, callback.getClientResponse().getStatus());
+        assertTrue(callback.getClientResponse().getStatusString().contains(
+                "Array / Scalar parameter mismatch"));
+
+        b = client.callProcedure(callback,
+                "one_string_scalar_param", (Object[])stringArgs);
+        // This is queued, but execution fails as it should.
+        // (Presumably with wrong number of params, but we have no way of knowing.)
```
paulmartel: I don't understand the parenthetical remark.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+        // (not an array of native types) passed as the last argument, the
+        // compile time type affects how the compiler presents the arguments to the
+        // callee:
+        //    cast to Object   - callProcedure sees just one param (which is an array)
+        //    cast to Object[] - (or a subclass of Object[]) callee sees each array
+        //                       element as its own parameter
+
+        // where desc in ?
+        // Cast it as an object and it's treated as a single paramter in the callee.
+        VoltTable vt = client.callProcedure("one_string_list_param", (Object)stringArgs)
```
paulmartel: global replace /paramter/parameter/

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+        String[] stringArgs = {"7", "8"};
+
+        String adHocQueryWithListParam = "select id from P1 where desc in ?";
+        String adHocQueryWithScalarParam = "select id from P1 where desc in (?)";
+
+        VoltTable vt;
+
+        // where desc in ?
+        // Cast it as an object and it's treated as a single paramter in the callee.
+        // This fails, but it should not, this issue is ENG-7604
```
paulmartel: I think that ENG-7604 is not a fixable issue and is essentially "pilot error". and "This fails" is overly strong wording.
Java makes its well-defined documented determination on whether an array argument sent to an Object... varargs is intended as a single array-valued Object parameter or multiple Object-valued parameters -- not that I can ever recall off the top of my head what rule it uses -- it's covered in the Effective Java book. This determination can be overridden with a cast. There's really nothing the API designer can do about it other than to only use Object... where a more specific type would not work OR where array-valued arguments are not meaningful -- so they can be detected as an error or flattened by the function implementation. I think that neither callProcedure nor voltQueueSql have that luxury, so they are subject to java's ruling and foisting any overruling on the caller. "This fails" as you say, when the programmer casually makes an incorrect assumption about the behavior of java. That happens far too often for this particular language feature, but there's little to be done about that.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
                 new long[][] {});
     }
 
+    public void testInWithIntParams() throws Exception {
```
paulmartel: There really should be a dedicated InLists test suite to bring these together with the original IN tests that I grafted rather clumsily onto TestIndexesSuite. For all I know there may already be redundant tests there and here.
Actually, I consider "Fixed SQL" to be a sub-optimal classification for test cases like "misc. stuff that broke once -- and possibly wouldn't have if it had been supported with a proper test suite". When I want a quick report from jenkins of "what broke", the answer "something that broke before" does not have great value to me. This doesn't need to be addressed now. Let's just add a ticket for migrating existing tests into a TestInLists.java suite.
I wouldn't mind another ticket for going through the Fixed SQL suite and finding a loving home for each of its test cases.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+                "one_list_param", new long[] {1, 2});
+
+        // scalar param where list should be provided fails
+        verifyProcFails(client,
+                "Array / Scalar parameter mismatch",
+                "one_list_param", 1);
+
+        vt = client.callProcedure("one_scalar_param", 5)
+                .getResults()[0];
+        validateTableOfScalarLongs(vt, new long[] {5});
```
paulmartel: Tests might be easier to read if segregated between positive and negative -- maybe even have a separate test...Errors method.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1858,6 +1858,222 @@ public void testInnerJoinWithOverflow() throws Exception {
...
+        // This should succeed, but doesn't (ENG-7604 again)
+        b = client.callProcedure(callback, "@AdHoc",
+                adHocQueryWithListParam, stringArgs);
+        assertTrue(b);
+        client.drain();
+        assertEquals(ClientResponse.GRACEFUL_FAILURE, callback.getClientResponse().getStatus());
+        assertTrue(callback.getClientResponse().getStatusString().contains(
+                "Array / Scalar parameter mismatch"));
+    }
+
```
paulmartel: I'm a little conflicted about the asynch test variations here. I don't know if there is an existing suite that covers asynch error recovery -- access to error messages -- with possibly planner errors (using @AdHoc), parameter passing errors, ee runtime errors, disconnect errors, and timeout errors as special cases -- SHOULD be tested somewhere. The selection of IN LIST queries as the "query fodder" for this kind of testing seems kind of like a quirk of history. I'd LIKE to be able to claim that these asynch in list tests are redundant with the similar synch IN LIST tests on the one hand and simpler asynch error response tests on the other, and remove them, but I guess I'd want to KNOW that the simpler error response tests exist and either meet or exceed the coverage of the asynch tests here. Is there some inherent interaction of IN LISTS / array parameters with asynch that rates a specialized combo test like this?

paulmartel:
Finished review round -- there seems like there is new ground being covered by these new tests, so let's merge them. We can find and cull any dupes later when we get to the suggested unpackings of TestFixedSQLSuite and TestIndexesSuite and locate what we have for asynch callProcedure coverage elsewhere.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: I stumbled onto this by accident.  I guess my fingers have gotten used to typing `new long[]`.  

We seem to be able to pass a scalar long to a parameter that compares with an integer column, so I was surprised that passing a `long[]` to `intcol IN ?` fails.  My intent was just to characterize the behavior in this test case.

I think it's probably a reasonable limitation, but I can file a bug on it if you like.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: Done.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: Yes, I think so.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: I made that comment before I discovered that getStatusString existed.  I'll remove the comment

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: I wrote a little test program trying this out:

```Java
class MyTest {
    static private void testIt(Object... objs) {
        int i = 0;
        for (Object o : objs) {
            System.out.println(i + ": " + o);
            ++i;
        }

        System.out.println();
    }

    static public void main(String[] args) {
        String[] strings = {"aardvark", "bear", "chimpanzee"};

        testIt(strings);
        testIt((Object)strings);
        testIt((Object[])strings);

        System.out.println("**************************");

        testIt("zebra", strings);
        testIt("zebra", (Object)strings);
        testIt("zebra", (Object[])strings);
    }
}
```
It produces this output:
```Text
0: aardvark
1: bear
2: chimpanzee

0: [Ljava.lang.String;@103e584d

0: aardvark
1: bear
2: chimpanzee

**************************
0: zebra
1: [Ljava.lang.String;@103e584d

0: zebra
1: [Ljava.lang.String;@103e584d

0: zebra
1: [Ljava.lang.String;@103e584d
```
Sorry this comment is so long.  Anyway, for the stored procedure case, there is exactly the issue you're talking about: there's just one vararg parameter, and since it's a list there needs to be a cast to distinguish what the user means (and javac gives a helpful warning).  I cover both casting to `Object` and `Object[]` in the test for stored procedures, and I see expected behavior.

For the ad hoc case however, the SQL statement is always the first vararg parameter, so any successive parameters will always be seen as arrays in callProcedure.  So there's really no ambiguity here and that's why I think it should succeed.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: I'm definitely guilty of using TestFixedSQLSuite as a catch all for my unit tests.  It is getting unwieldy and takes a long time to run.  It even takes a long time for me to *comment out* the stuff that I don't want to run.  I'd like to see it pruned down too for this reason as well as the one you stated.  

I'll add tickets for both IN lists and the more general case of re-repatriating test cases in fixed SQL suite.

I should have looked harder for existing test coverage.  "IN" turns up a lot of hits when grepping and I got lazy.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: No, I haven't noted anything special about IN lists and array parameters with respect to async procedure calls.

I'm also not sure there's any tests that cover async error recovery, but I might not be finding them.  Should I file a ticket?

paulmartel: Let me do some quick looking around -- I'll file a ticket if I don't find a reasonable suite.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
paulmartel: Thanks for clarifying. I wasn't accounting for the effect of the query string to @AdHoc.
I need to take another look to understand this test case.

wolffcm:
Paul, I pushed a new commit that incorporates much of your comments.  I didn't break out the negative and positive test cases, but I did do some reorganization that makes for shorter test methods that hopefully makes them a bit easier to follow.

paulmartel:
OK to merge.

PR #1993, opened 2015-01-27
ENG-7472 fixed timestamps in @SnapshotScan

dweiss88:


src/frontend/org/voltdb/SnapshotScanAgent.java
```diff
@@ -243,8 +243,7 @@ private VoltTable getSnapshotScanResults(String path)
                                     f.getParent(),
                                     f.getName(),
                                     savefile.getTxnId(),
-                                    org.voltdb.TransactionIdManager.getTimestampFromTransactionId(
```
vtkstef: cant'use this becuase the savefile timestamp can change across cluster nodes. Do try to get the timestamp from the transaction

PR #1994, opened 2015-01-28
Do bounds-checking when getting SQL text for long running query

wolffcm:
This was causing an array out of bounds exception when the EE was
calling back into Java, and the EE was crashing.

This isn't a complete fix, as I still don't understand why m_sqlTexts did not have text for the fragment, but it least avoids taking down the EE.

### DIFF
--- a/src/frontend/org/voltdb/jni/ExecutionEngine.java
+++ b/src/frontend/org/voltdb/jni/ExecutionEngine.java
@@ -412,12 +412,37 @@ public abstract class ExecutionEngine implements FastDeserializer.Deserializatio
                         m_currMemoryInBytes,
                         m_peakMemoryInBytes);
 
-        // ENG-7610: array of SQL texts seems to get out of sync with the current
-        // batch index, so we need to do bounds checking here.
         if (m_sqlTexts != null
                 && m_currentBatchIndex >= 0
                 && m_currentBatchIndex < m_sqlTexts.length) {
             msg += "  Executing SQL statement is \"" + m_sqlTexts[m_currentBatchIndex] + "\".";
+        }
+        else if (m_sqlTexts == null) {
+            // Can this happen?
+            msg += "  SQL statement text is not available.";
+        }
+        else {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/jni/ExecutionEngine.java
```diff
@@ -411,7 +411,12 @@ private String getLongRunningQueriesMessage(long latency, String planNodeName, b
                         CoreUtils.hsIdToString(m_siteId),
                         m_currMemoryInBytes,
                         m_peakMemoryInBytes);
-        if (m_sqlTexts != null) {
+
+        // ENG-7610: array of SQL texts seems to get out of sync with the current
+        // batch index, so we need to do bounds checking here.
+        if (m_sqlTexts != null
+                && m_currentBatchIndex >= 0
+                && m_currentBatchIndex < m_sqlTexts.length) {
```
paulmartel: Since we don't understand how/when m_currentBatchIndex gets out of synch, maybe for our own good we should be getting MORE verbose in this case rather than LESS. 

I'm not even sure that we understand the use case that leads to the already handled (m_sqlTexts == null) condition.

At some small risk of annoying the user with an additional line of text that may mean nothing to them (but might help us), let's always append SOMETHING to the message in place of the executing SQL statement.
for m_sqlTexts == null, something like:
    "SQL statement text is not available for reporting" 
and for the m_currentBatchIndex out of range, something like:
    "Unable to report specific SQL statement text for batch index " + m_currentBatchIndex + ". It MAY be (one of) "
-- and then guardedly iterate and dump the length and contents of m_sqlTexts.

This might help us figure out when/where/how this is going south.

End of review round.

wolffcm:
I pushed another commit that produces a more verbose message, and added some testing.

paulmartel:
Looks good. Thanks. Please follow up w/ support re: does user need/want a backport of this fix.

vtkstef:
Thanks Chris

PR #1998, opened 2015-01-29
ENG-7556 Unified parsing/lexing

wijjo:
Reorganize lexer/parser

All tests pass.


src/frontend/org/voltdb/parser/JDBCParser.java
```diff
@@ -0,0 +1,83 @@
+/* This file is part of VoltDB.
```
jhugg: Any reason this file shouldn't be with the JDBC package?

wijjo: Nope, but I think there's some benefit to looking in one place for all parsing code so that the ripple of syntax changes is more manageable. Does that make any sense? If not I can move it.

mikeismert: I thought the idea was to try to get all the parsing/lexing in one place so that someone could do follow-on work to consolidate it all.  Pushing this stuff back into the JDBC package seems like it defeats that purpose.

jhugg: Yeah. I guess so.

jhugg:
So I haven't really thought this out line by line.
1. I want to check if there's any performance changes from this code on my catalog update benchmark.
2. I'm a little bummed that the consolidation of code added 1000 lines of code.

jhugg:
Ok it doesn't seem to be any different performance wise and the length is probably mostly licenses.

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -1394,7 +1395,7 @@ private void processExplainDefaultProc(AdHocPlannedStmtBatch planBatch) {
     ClientResponseImpl dispatchExplainProcedure(StoredProcedureInvocation task, ClientInputHandler handler, Connection ccxn, AuthUser user) {
         ParameterSet params = task.getParams();
         //String procs = (String) params.toArray()[0];
-        List<String> procNames = MiscUtils.splitSQLStatements( (String)params.toArray()[0]);
+        List<String> procNames = SQLLexer.splitStatements( (String)params.toArray()[0]);
```
paulmartel: AFAIK, we don't actually support multiple proc names in an ExplainProc call, so I THINK that the string is always a single procname symbol and all this splitting and iterating is a sad no-op, but this was broke when you found it, so no need to deal with it now.

src/frontend/org/voltdb/parser/JDBCParser.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public static ParsedCall parseJDBCCall(String jdbcCall) throws SQLParser.Exception
+    {
+        Matcher m = PAT_CALL_WITH_PARAMETERS.matcher(jdbcCall);
+        if (m.matches()) {
+            String sql = m.group(1);
+            int parameterCount = PAT_CLEAN_CALL_PARAMETERS.matcher(m.group(2)).replaceAll("").length();
+            return new ParsedCall(sql, parameterCount);
+        }
+        else {
```
paulmartel: Drop Redundant "else" after return for readability -- to get two serial if(...) { return ... }  constructs.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -0,0 +1,535 @@
...
+    // Simplest possible SQL DDL token lexer. (set in initializePatternsOnce())
+    private static Pattern PAT_ANY_DDL_FIRST_TOKEN = null;
+
+    // Generate supported renames (after we know what's allowed).
+    private static Pattern generateRenamePattern(String... renameables)
+    {
+        return SPF.statementLeader(
+                    SPF.token("alter"),
+                    SPF.capture(SPF.token(renameables)),
+                    SPF.optional(SPF.anything()), //TODO: Too forgiving?
```
paulmartel: I'd expand further on the "TODO: comment, something more like:

                    //TODO: too forgiving? Note that "anything" optonally matches nothing.
                    // Is the anything pattern really doing anything useful here?
                    // And this is currently its only caller.
                    // The create/drop pattern use SPF.symbol in a similar situation.
                    // On the other hand, only WHITE LIST patterns suffer from their generality
                    // (false positives). A BLACK LIST pattern, which this currently is,
                    // tends to become MORE effective as it becomes more general,
                    // as long as it doesn't cross the line and start catching false negatives.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -0,0 +1,535 @@
...
+            }
+            else {
+                unsupportedVerbCount++;
+            }
+        }
+        PAT_ANY_DDL_FIRST_TOKEN =
+            SPF.statementLeader(
+                SPF.capture(SPF.token(verbsAll)),
+                SPF.anyClause()
+            ).compile();
```
paulmartel: Many regexps that were replaced by SPF constructs like this were annotated in the original code with 
Pattern.CASE_INSENSITIVE | Pattern.MULTILINE | Pattern.DOTALL

Is this somehow happening by default now? I didn't notice how/where.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -0,0 +1,535 @@
...
+        // All handled (white-listed) patterns.
+        int renameableCount = 0;
+        String[] secondTokens = new String[OBJECT_TOKENS.length + MODIFIER_TOKENS.length];
+        for (int i = 0; i < OBJECT_TOKENS.length; ++i) {
+            secondTokens[i] = OBJECT_TOKENS[i].token;
+            if (OBJECT_TOKENS[i].renameable) {
+                renameableCount++;
+            }
+        }
+        for (int j = 0; j < MODIFIER_TOKENS.length; ++j) {
```
paulmartel: The use of modifiers as second tokens here is (continues to be) a little obscure. I'd add commentary something like:

        // Modifier tokens are supported in the place of object tokens following
        // a verb to allow the "verb modifier object" pattern like "CREATE UNIQUE INDEX".
        // For simplicity, "CREATE UNIQUE" et. al. are considered sufficient evidence that
        // the statement is a permitted white-listed DDL statement.
        // We seem to be more concerned about accidentally permitting
        // "CREATE <non-permitted-object> ..."
        // than "CREATE UNIQUE <non-permitted-object> ...".
        // Otherwise, we'd require the modifiers to be part of a nested
        // "modifier object" subpattern.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -0,0 +1,535 @@
...
+        String[] verbsSupported = new String[supportedVerbCount];
+        supportedVerbCount = 0;     // Reuse to build supported verb array.
+        for (int i = 0; i < VERB_TOKENS.length; ++i) {
+            if (VERB_TOKENS[i].supported) {
+                verbsSupported[supportedVerbCount++] = VERB_TOKENS[i].token;
+            }
+        }
+        Pattern patSupportedPreambles =
+            SPF.statementLeader(
+                SPF.capture(SPF.token(verbsSupported)),
```
paulmartel: The captures here are a misleading dead end -- not actually used.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -0,0 +1,535 @@
...
+                SPF.capture(SPF.token(secondTokens))
+            ).compile();
+        PAT_WHITELISTS = new Pattern[] {
```
paulmartel: Constructing and iterating over this single-element vector continues to be overkill.

src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -0,0 +1,1224 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Results from parseShowStatement
+     */
+    public static class ParseShowResults
+    {
+        public String command;
```
paulmartel: Preserving command is not actually useful.
The two possible values, list and show, are completely equivalent.
So what we really have here is a class to hold a single String value, subcommand.
We could just use a String directly.

src/frontend/org/voltdb/parser/SQLPatternFactory.java
```diff
@@ -0,0 +1,180 @@
...
+            return part;
+        }
+
+        public static SQLPatternPart optional(SQLPatternPart part)
+        {
+            part.m_flags |= SQLPatternFactory.OPTIONAL;
+            return part;
+        }
+
+        public static SQLPatternPart or(SQLPatternPart... parts)
```
paulmartel: The name "or" is kind of cryptic and hard to read. I'd suggest oneOf.
I don't think this overload has any callers. How do we know it works?

src/frontend/org/voltdb/parser/SQLPatternFactory.java
```diff
@@ -0,0 +1,180 @@
...
+        public static SQLPatternPart or(String... strs)
+        {
+            // Default to outer and inner non-capturing groups.
+            SQLPatternPartElement retElem = new SQLPatternPartElement(strs);
+            retElem.m_flags |= SQLPatternFactory.GROUP;
+            retElem.m_separator = "|";
+            return retElem;
+        }
+
+        public static SQLPatternPart token(String... strs)
```
paulmartel: While we're in the business of trivial wrapper aliases, I LOVE the name token when this is used with a single string argument to match a specific fixed token, but I find it much less descriptive when applied to a vector or parameter list of alternative token strings. In the latter case, I'd prefer something that more explicitly expresses choice like "tokenOneOf" or "tokenAlternative"
To that end, I'd change the signature of plain old "token" to take just a String, and call a differently named method wherever callers use the multi/vector argument signatures.
I don't know if there is or should be an optimizated implementation for the single-token case.
But even if they used exactly the same code, I think that the intent of single-token and multi-token use cases are diffferent enough that they deserve different method names.

src/frontend/org/voltdb/parser/SQLPatternFactory.java
```diff
@@ -0,0 +1,180 @@
...
+        {
+            return or(strs);
+        }
+
+        public static SQLPatternPart symbol()
```
paulmartel: Add TODO/warning comments:

        //TODO: symbol and ddlName could be unified or at least used more consistently
        // by different callers with the same intent.
        //TODO: neither symbol nor ddlName recognize quoted identifiers.

tests/frontend/org/voltdb/utils/TestSqlCmdInterface.java
```diff
@@ -365,16 +366,6 @@ public void testParseQuery23() {
     }
 
     // To test parseQueryProcedureCallParameters()
-    // To test a valid query: 'select * from dummy'
-    @Test
-    public void testParseQueryProcedureCallParameters22() {
```
paulmartel: Why were tests 22 and 25 dropped (vs. tweaked). 
Did they break?
Are they not testing something reasonable, worth having test coverage for?

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -0,0 +1,535 @@
...
+    private static final Pattern PAT_SINGLE_LINE_COMMENT = Pattern.compile(
+            "^\\s*" + // start of line, 0 or more whitespace
+            "--" + // start of comment
+            ".*$"); // everything to end of line
+
+    private static final Pattern PAT_STRIP_CSTYLE_COMMENTS = Pattern.compile(
+            "/\\*(.|\\n)*?\\*/"
+            );
+
+    //===== Derived parsing data (populated in initializePatternsOnce() on first demand)
```
paulmartel: I'm not seeing the justification for initializePatternsOnce() on first demand and all its callers.
Couldn't this function be replaced with an anonymous static code block to run automatically at class load time?

paulmartel:
I really like the new "more regular" abstract method of constructing regexps. I find it much easier to read than even commented/verbose regexp syntax.
You've done a great job of unifying the underpinnings of the different parsers, while -- IMPORTANTLY -- initially maintaining the original callers' behaviors, fluke for fluke. This puts us in a better position to identify and fix inconsistencies and maybe do some further unification where the callers' requirements overlap and we may still have unwittingly redundant or slightly incompatible patterns. These improvements can be trickled in and consciuously tested as we make them. I think that given this assignment I would have been tempted to try and "fix everything now" and could have caused a regression nightmare. 
I think I might have done more to merge the parser classes so they weren't so caller-oriented.
OTOH, figuring out where to find the "original" source for each method and pattern for this review was plenty hard enough without them being mashed up into fewer classes. Maybe that can come later with further unification (and a less churnful changeset).
Great work!

src/frontend/org/voltdb/ClientInterface.java
wijjo: I'll add a comment to this effect.

src/frontend/org/voltdb/parser/SQLLexer.java
wijjo: I'll study this and see if I can either explain it better, or even make it better.

src/frontend/org/voltdb/parser/SQLLexer.java
wijjo: The flags are added in SQLPatternPartElement.compile()

src/frontend/org/voltdb/parser/SQLLexer.java
wijjo: Added your comment. Had nothing more illuminating myself. Thanks.

src/frontend/org/voltdb/parser/SQLLexer.java
wijjo: Removed the captures. Easy enough to add back if we eventually need them.

src/frontend/org/voltdb/parser/SQLLexer.java
wijjo: I personally prefer that the white and black lists look and work the same. I think there's more mental overhead if one's an array and the other isn't.

src/frontend/org/voltdb/parser/SQLParser.java
wijjo: I'll make that change. Thanks

src/frontend/org/voltdb/parser/SQLPatternFactory.java
wijjo: Makes sense.

tests/frontend/org/voltdb/utils/TestSqlCmdInterface.java
wijjo: The tests seemed to be of marginal use, but I restored and tweaked them. A lot of that suite seems a little odd (to me) how it's doing the testing.

src/frontend/org/voltdb/parser/SQLLexer.java
wijjo: Could be me having a Java brain-belch. I'll look at making that change.

PR #2002, opened 2015-01-30
Use correct SQL text for logging long-running fragments

wolffcm:
Also, add more testing and clean up the fragment progress update test.

I put an explanation of what happened with this bug in the ticket:
https://issues.voltdb.com/browse/ENG-7610

### DIFF
--- a/src/frontend/org/voltdb/jni/ExecutionEngine.java
+++ b/src/frontend/org/voltdb/jni/ExecutionEngine.java
@@ -334,7 +334,7 @@ public abstract class ExecutionEngine implements FastDeserializer.Deserializatio
         TIME_OUT_MILLIS = newLatency;
     }
 
-    public long fragmentProgressUpdate(int batchIndex,
+    public long fragmentProgressUpdate(int indexFromFragmentTask,
             String planNodeName,
             String lastAccessedTable,
             long lastAccessedTableSize,
@@ -355,7 +355,7 @@ public abstract class ExecutionEngine implements FastDeserializer.Deserializatio
         long latency = currentTime - m_startTime;
 
         if (m_readOnly && TIME_OUT_MILLIS > 0 && latency > TIME_OUT_MILLIS) {
-            String msg = getLongRunningQueriesMessage(latency, planNodeName, true);
+            String msg = getLongRunningQueriesMessage(indexFromFragmentTask, latency, planNodeName, true);
             log.info(msg);
 
             // timing out the long running queries
@@ -380,7 +380,7 @@ public abstract class ExecutionEngine implements FastDeserializer.Deserializatio
             // future callbacks per log entry, ideally so that one callback arrives just in time to log.
             return LONG_OP_THRESHOLD;
         }
-        String msg = getLongRunningQueriesMessage(latency, planNodeName, false);
+        String msg = getLongRunningQueriesMessage(indexFromFragmentTask, latency, planNodeName, false);
         log.info(msg);
 
         m_logDuration = (m_logDuration < 30000) ? (2 * m_logDuration) : 30000;
@@ -391,7 +391,8 @@ public abstract class ExecutionEngine implements FastDeserializer.Deserializatio
         return LONG_OP_THRESHOLD;
     }
 
-    private String getLongRunningQueriesMessage(long latency, String planNodeName, boolean timeout) {
+    private String getLongRunningQueriesMessage(int indexFromFragmentTask,
+            long latency, String planNodeName, boolean timeout) {
         String status = timeout ? "timed out at" : "taking a long time to execute -- at least";
         String msg = String.format(
                 "Procedure %s is %s " +
@@ -413,21 +414,21 @@ public abstract class ExecutionEngine implements FastDeserializer.Deserializatio
                         m_peakMemoryInBytes);
 
         if (m_sqlTexts != null
-                && m_currentBatchIndex >= 0
-                && m_currentBatchIndex < m_sqlTexts.length) {
-            msg += "  Executing SQL statement is \"" + m_sqlTexts[m_currentBatchIndex] + "\".";
+                && indexFromFragmentTask >= 0
+                && indexFromFragmentTask < m_sqlTexts.length) {
+            msg += "  Executing SQL statement is \"" + m_sqlTexts[indexFromFragmentTask] + "\".";
         }
         else if (m_sqlTexts == null) {
             // Can this happen?
             msg += "  SQL statement text is not available.";
         }
         else {
-            // For some reason, the current index in the batch isn't a valid
+            // For some reason, the current index in the fragment task message isn't a valid
             // index into the m_sqlTexts array.  We don't expect this to happen,
             // but let's dump something useful if it does.  (See ENG-7610)
             StringBuffer sb = new StringBuffer();
-            sb.append("  Unable to report specific SQL statement text for batch index "
-                    + m_currentBatchIndex + ".  ");
+            sb.append("  Unable to report specific SQL statement text for "
+                    + "fragment task message index " + indexFromFragmentTask + ".  ");
             sb.append("It MAY be one of these " + m_sqlTexts.length + " items: ");
             for (int i = 0; i < m_sqlTexts.length; ++i) {
                 if (m_sqlTexts[i] != null) {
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/jni/ExecutionEngine.java
```diff
@@ -391,7 +391,8 @@ public long fragmentProgressUpdate(int batchIndex,
         return LONG_OP_THRESHOLD;
     }
 
-    private String getLongRunningQueriesMessage(long latency, String planNodeName, boolean timeout) {
+    private String getLongRunningQueriesMessage(int indexFromFragmentTask,
+            long latency, String planNodeName, boolean timeout) {
         String status = timeout ? "timed out at" : "taking a long time to execute -- at least";
         String msg = String.format(
```
jxchinaren: I can not comment on line: 411.

Do you still need to use "m_currentBatchIndex" instead of the new "indexFromFragmentTask"?

jxchinaren: Another question, can we just get rid of the "m_currentBatchIndex" member variable according to your change?

tests/frontend/org/voltdb/benchmark/tpcc/procedures/FragmentUpdateTestProcedure.java
```diff
@@ -71,6 +71,9 @@
     // Not meaningful, but should take long.
     public final SQLStmt item_big_del = new SQLStmt("DELETE FROM ITEM WHERE I_NAME <> 'NULL_NULL';");
 
+    // Just a quick query
+    public final SQLStmt quick_query = new SQLStmt("SELECT W_ID FROM WAREHOUSE LIMIT 0");
```
jxchinaren: "LIMIT 0" is tricky here depending on how smart the planner it is.  
"LIMIT 1" should be more reliable for testing or both if you want.

paulmartel:
Oops - I got turned around and commented directly on the commit vs. on the pull request. Sorry for any confusion.

tests/frontend/org/voltdb/benchmark/tpcc/procedures/FragmentUpdateTestProcedure.java
paulmartel: Agreeing with Xin. Limit 1 is more conservative.
IIRC, one of my first ever hsql patches was to convince hsql that LIMIT 0 was acceptable. More recently, this is one of the patches that has not fared well in the early going of my hsql upgrade, so there is some risk of near-term regression if I flubbed the merge. I would rather catch that regression in its own dedicated LIMIT test case -- I HOPE I wrote one -- rather than (also) trip over it in this tangential test case.

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
```diff
@@ -418,9 +522,6 @@ public void testTimingoutQueries() throws Exception {
     }
 
     private ExecutionEngine m_ee;
-    private static final int CLUSTER_ID = 2;
-    private static final long NODE_ID = 1;
-    private int m_tableSize;
```
paulmartel: Nice cleanup of "variable scope creep".

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
```diff
@@ -338,25 +422,25 @@ private void verifyLongRunningQueries(
...
                 break;
             case NO_STATEMENT:
                 sqlTexts = null;
                 break;
             default:
                 assert (sqlTextExpectation == SqlTextExpectation.STATEMENT_LIST);
-                sqlTexts = new String[] {};
+                // Leave off the last item, which is the one that needs to be
+                // reported.
+                sqlTexts = Arrays.copyOfRange(sqlTexts, 0, numFragsToExecute - 1);
```
paulmartel: This was "already broken", but I find this idiom

    default:
        assert(switchvariable == EnumName.LAST_REMAINING_CASE);
        // implementLastRemainingCase ...

not nearly as readable as 

    case LAST_REMAINING_CASE:
        // implementLastRemainingCase ...
    default:
        assert(false); // or fail("unexpected invalid value specified for switchvariable");

src/frontend/org/voltdb/jni/ExecutionEngine.java
wolffcm: We still put the "batch index" into the long-running log message---it refers to how many calls to voltExecuteSQL have preceded the one that's taking a long time.  

I filed ENG-7631 on the "batch index", since it will always be 0 for the MP/RO case.  In the ticket I also suggest removing it altogether, since it doesn't seem useful to the user.  What do you think, is it useful?

tests/frontend/org/voltdb/benchmark/tpcc/procedures/FragmentUpdateTestProcedure.java
wolffcm: Ok, I'll change it to 1.

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
wolffcm: I agree that it's not as readable.  I got into the habit of using the less-readable idiom because it ensures that code coverage tools will show every branch as being hit.  In the readable way we would never hit the default case.  If high code coverage numbers are a goal, or might be in the future, it's something to consider.

Then again, this is test code, so coverage is less important here.

paulmartel: Interesting point. So maybe for actual product code this slightly awkward hybrid covers both concerns?

    case LAST_REMAINING_CASE:
    default: // avoid "code coverage" gap by handling unexpected "default" case in a mainline code path.
        assert(switchvariable == EnumName.LAST_REMAINING_CASE); // fail the "default:" case.
        // implementLastRemainingCase ...

wolffcm: I don't think the code coverage tool that I used before (Bullseye) would have shown 100% coverage for that.  It would instrument the C++ code and insert calls to increment counters, and I think every case label would be counted separately.

I guess it really depends on the tool, and I don't know much about Emma.  It seems reasonable that failure to hit `default: assert(false);` shouldn't really be considered a gap in testing---hopefully a tool can be taught that.

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
```diff
@@ -278,7 +281,31 @@ public void testProgressUpdateLogSqlStmtList() throws Exception {
...
+
+        verifyLongRunningQueries(50, 0, "item_crazy_join", 5, false, SqlTextExpectation.NO_STATEMENT);
+    }
+
+    @SuppressWarnings("deprecation")
+    public void testProgressUpdateLogSqlStmtListRW() throws Exception {
+        // Set the log duration to be very short, to ensure that a message will be logged.
+        m_ee.setInitialLogDurationForTest(1);
+
+        verifyLongRunningQueries(50, 0, "item_crazy_join", 5, false, SqlTextExpectation.STATEMENT_LIST);
```
paulmartel: I am trying to follow the logic here. Correct me if I am wrong.

We had a "can not build a message if the fragment index is out of range" bug which was triggered by a "used the wrong fragment index" bug. We fixed the first bug by adding code. We fixed the second bug, thus making it next to impossible to verify the code we added to fix the first bug -- since it now MAY never be triggered again under normal usage. So, we have built this test with a purposely corrupted input to ensure that the first bug fix will be able to stand up to something like a regression in the second bug fix.

If I am correct, a comment would have helped much in helping me piece this together. I was beating my head over why this test was purposely truncating a perfectly good generated EE input -- was this simulating something that is really expected to happen in normal production usage? Answer: not really -- it's just stressing the long query message generator to slightly beyond its expected operational parameters. That's a fine thing to do, as long as you declare that it's what you're doing.

wolffcm:
I pushed another commit based on review feedback; will merge once Jenkins passes.

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
wolffcm: Just realized I never replied to your comment---yes, you have it right.  I added a comment that helps to clarify this.

PR #2008, opened 2015-02-04
Make HSQL amenable to OFFSET with no LIMIT clause

wolffcm:
Add unit tests for same, for both SELECT and DELETE.

### DIFF
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -902,7 +902,7 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
         m_limitNodeTop.setOffsetParameterIndex(offsetParamIndex);
 
         // check if limit can be pushed down
-        m_limitCanPushdown = !m_distinct;
+        m_limitCanPushdown = (hasLimit() && !m_distinct);
         if (m_limitCanPushdown) {
             for (ParsedColInfo col : m_displayColumns) {
                 AbstractExpression rootExpr = col.expression;
@@ -1320,10 +1320,17 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
         return Collections.unmodifiableList(m_orderColumns);
     }
 
+    private boolean hasLimit() {
+        return m_limit != -1 || m_limitParameterId != -1;
+    }
+
+    private boolean hasOffset() {
+        return m_offset > 0 || m_offsetParameterId != -1;
+    }
+
     @Override
     public boolean hasLimitOrOffset() {
-        if ((m_limit != -1) || (m_limitParameterId != -1) ||
-            (m_offset > 0) || (m_offsetParameterId != -1)) {
+        if (hasLimit() || hasOffset()) {
             return true;
         }
         return false;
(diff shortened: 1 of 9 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -1068,7 +1068,12 @@ private int getLimitCount(Session session, int rowCount) {
 
         int limitCount = Integer.MAX_VALUE;
 
+        // A VoltDB extension to support OFFSET without LIMIT
```
jxchinaren: I remember Paul has a fixed pattern to inject code into hsql for his automatically upgrade project. Something like:
    /************************* Volt DB Extensions *************************/
        voltdbCode();
    /*************************************************************************/

Paul should have more suggestions on his pattern.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -977,18 +977,21 @@ static VoltXMLElement voltGetXMLExpression(QueryExpression queryExpr, Expression
...
-                    Integer limitValue = (Integer)limitCondition.nodes[1].getValue(session);
-                    Expression expr = new ExpressionValue(limitValue,
-                            org.hsqldb_voltpatches.types.Type.SQL_BIGINT);
-                    limit.children.add(expr.voltGetXML(session));
-                    limit.attributes.put("limit", limitValue.toString());
-                } else {
-                    limit.attributes.put("limit_paramid", limitCondition.nodes[1].getUniqueId(session));
+                // Limit may be null (offset with no limit), or
+                // it may be a parameter
+                if (limitCondition.nodes[1] != null) {
```
jxchinaren: Could you use "limitCondition.getRightNode()" instead of hard coded "limitCondition.nodes[1]"? 

Maybe cache it as a local variable as it has been referenced a couple of times down below.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -902,7 +902,7 @@ private void prepareLimitPlanNode () {
         m_limitNodeTop.setOffsetParameterIndex(offsetParamIndex);
 
         // check if limit can be pushed down
-        m_limitCanPushdown = !m_distinct;
+        m_limitCanPushdown = (hasLimit() && !m_distinct);
```
jxchinaren: Could you add a comment saying something like "LimitPlanNode can have LIMIT clause only, OFFSET clause only, or both."? Then it's easier to understand the extra check.

jxchinaren:
Add another matview OFFSET only query to check?
Maybe around https://github.com/VoltDB/voltdb/blob/master/tests/frontend/org/voltdb/compiler/TestVoltCompiler.java#L2296

jxchinaren:
Nice work. Finish review round, maybe Paul has more suggestions.

wolffcm:
Thanks Xin, I'll push another commit with your feedback.

wolffcm:
I pushed another commit to address Xin's review feedback.

jxchinaren:
Looks good.

PR #2009, opened 2015-02-04
ENG-7649: Stop defaulting DR producer cluster ID to 0

afrogers:


src/frontend/org/voltdb/iv2/RejoinProducer.java
```diff
@@ -327,9 +327,16 @@ public void runForRejoin(SiteProcedureConnection siteConnection, TaskLog rejoinT
 
                     drSequenceNumbers = event.drSequenceNumbers;
 
-                    Pair<Long, Long> myDRIds = event.remoteDCLastIds.get(0).get(m_partitionId);
-                    if (myDRIds != null) {
-                        VoltDB.instance().getConsumerDRGateway().notifyOfLastAppliedBinaryLog(m_partitionId, myDRIds.getFirst(), myDRIds.getSecond());
+                    if (!event.remoteDCLastIds.isEmpty()) {
```
nshi: This if guard is probably redundant?

afrogers: True; pushing fix

src/frontend/org/voltdb/iv2/RejoinProducer.java
```diff
@@ -327,9 +327,16 @@ public void runForRejoin(SiteProcedureConnection siteConnection, TaskLog rejoinT
...
+                        if (event.remoteDCLastIds.size() != 1) {
+                            VoltDB.crashLocalVoltDB("Have DR data from more than one upstream cluster, which shouldn't be possible.");
```
nshi: This detects multiple upstreams at rejoin time. Is it detected at other places, like when DR comes up?

Seems like it's either unpossible to have multiple upstreams at the consumer or this check need to occur earlier, rather than during rejoin.

afrogers: So this is detected in a couple of places -- restore and rejoin -- just to sanity check the map that's present at other nodes. Although having extraneous data would actually be harmless, anyway. I have no strong feelings about this, to be honest -- the situation is not actually possible, and these sanity checks are at least a little silly.

nshi:
Review finished.

afrogers:
Rebasing and pushing again.

PR #2011, opened 2015-02-04
Fix TestFragmentProgressUpdate to avoid sporadic failures

wolffcm:
Some tests were tweaking the timeout latency setting and not unsetting it, and this was sometimes affecting subsequent tests, especially in memcheck mode.

I'm trying out using a helper class and Java's AutoCloseable and try-with-resources to tightly scope ExecutionEngine settings.  Does this idiom seem useful?

Also did some more general code cleanup in this test.


tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
```diff
@@ -251,61 +292,61 @@ public void testPeakLargerThanCurr() throws Exception {
...
     }
 
-    @SuppressWarnings("deprecation")
     public void testProgressUpdateLogSqlStmt() throws Exception {
         // Set the log duration to be very short, to ensure that a message will be logged.
-        m_ee.setInitialLogDurationForTest(1);
-
-        verifyLongRunningQueries(50, 0, "item_crazy_join", 5, true, SqlTextExpectation.SQL_STATEMENT);
+        try (AutoEngineSettings aes = new AutoEngineSettings()) {
+            aes.setInitialLogDuration(1);
```
jxchinaren: Suggestion:
Rename the function to "setInitialLogDurationForTest" to keep its original name.

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
```diff
@@ -251,61 +292,61 @@ public void testPeakLargerThanCurr() throws Exception {
...
+        }
     }
 
-    @SuppressWarnings("deprecation")
     public void testProgressUpdateLogNoSqlStmt() throws Exception {
         // Set the log duration to be very short, to ensure that a message will be logged.
-        m_ee.setInitialLogDurationForTest(1);
-
-        verifyLongRunningQueries(50, 0, "item_crazy_join", 5, true, SqlTextExpectation.NO_STATEMENT);
+        try (AutoEngineSettings aes = new AutoEngineSettings()) {
```
jxchinaren: Could you explain why should you try block here?

jxchinaren:
The others looks good.

tests/frontend/org/voltdb/jni/TestFragmentProgressUpdate.java
wolffcm: The idea is that the initial log duration will be set back to its original value when control flow exits the try block.  This happens because AutoEngineSettings inherits from AutoCloseable, so its close method is guaranteed to be called.

The reason I want the initial log duration set back to its original value is that I want each test to avoid "drooling" global settings it needs into other tests.  This ensures that the whole suite will pass no matter what order the tests are run in.

jxchinaren: I do see the test "testTimingoutQueries" has used AutoEngineSettings twice. For the other cases, AutoEngineSettings has only been used once per test. Will its close method be called after the test finishes?

All the use case of AutoEngineSettings share the same pattern of code:
        try (AutoEngineSettings aes = new AutoEngineSettings()) {
            aes.setInitialLogDuration(1);
            verifyLongRunningQueries(50, 0, "item_crazy_join", 5, true, SqlTextExpectation.STATEMENT_LIST);
        }

Will it be possible to hide the AutoEngineSettings inside of "verifyLongRunningQueries" function, reducing 4 lines to be 1 line?

wolffcm: This would make setting the initial log duration the same for all callers of verifyLongRunningQueries.  This seems harmless though, so I'll give it a shot, since it will simplify the test a lot.

Java guarantees that an instance of AutoCloseable created inside a `try (...)` will have its close method called when control flow exits the the try block (either by exiting naturally or if an exception is thrown).  Since all instances of AutoEngineSettings are created in such a context, the close method will always be called, no matter how many instances are created per test.

I'll push a new commit with your suggestions once you've finished your review round.

PR #2012, opened 2015-02-05
Eng 7612 ee crash message to log

jxchinaren:


src/ee/expressions/functionexpression.cpp
```diff
@@ -35,6 +35,15 @@ template<> inline NValue NValue::callUnary<FUNC_VOLT_SQL_ERROR>() const {
         const int32_t valueLength = getObjectLength_withoutNull();
         const char *valueChars = reinterpret_cast<char*>(getObjectValue_withoutNull());
         std::string valueStr(valueChars, valueLength);
+
+#if defined(DEBUG) || defined(_DEBUG) || defined(_DEBUG_)
+        if (valueStr == "crash_me") {
```
wolffcm: Can you do a comparison of an std::string to a const char* literal like this?  I have the feeling if this compiles that it always returns false?  

(Actually I just verified that the C++ library includes operator overloads to make this work right, so I just learned something.)

src/ee/expressions/functionexpression.cpp
```diff
@@ -35,6 +35,15 @@ template<> inline NValue NValue::callUnary<FUNC_VOLT_SQL_ERROR>() const {
         const int32_t valueLength = getObjectLength_withoutNull();
         const char *valueChars = reinterpret_cast<char*>(getObjectValue_withoutNull());
         std::string valueStr(valueChars, valueLength);
+
+#if defined(DEBUG) || defined(_DEBUG) || defined(_DEBUG_)
```
wolffcm: There is also the more standard `#ifndef NDEBUG` for this kind of thing.  (This is what the C library's `assert` uses.)  In a couple of places we use that too.  I'm not sure which one is better.

src/ee/expressions/functionexpression.cpp
```diff
@@ -35,6 +35,15 @@ template<> inline NValue NValue::callUnary<FUNC_VOLT_SQL_ERROR>() const {
...
+        if (valueStr == "crash_me") {
+            // generate a segmentation fault on purpose
+            int *ptr = NULL;
+            *ptr = 1; // EXC_BAD_ACCESS
```
wolffcm: I think writing to a null pointer is undefined behavior.  This likely does what you want, but may not be portable, or may not do what you want for -O3, or when the compiler is upgraded.  The most portable way is probably to call the C function `raise(SIGSEGV)`.

I'm probably being pedantic; feel free to ignore.

wolffcm:
Looks good to me.

paulmartel:
Rebasing branch with changes re:wolffcm's comments.

PR #2014, opened 2015-02-06
ENG-7389 Force white space after sqlcmd FILE and RECALL commands to redu...

paulmartel:
...ce false positives.


wolffcm:
Looks good.  I guess we had "recall" as well as a "file" problem.

Let's not forget to add tests for this kind of stuff when the testing framework is online.  It might be good to collect all the test cases we'll need to add into a ticket, if we haven't already.

PR #2015, opened 2015-02-06
Removed unused hqldb/dbinfo package to ease hsql upgrade.

paulmartel:
There are a number of hsqldb source code directories that we don't use (and more to come in 2.2). 
Most of them were just never added to our repo (or perhaps dropped by us long ago) -- the unused dbinfo package should have been one of them.


wolffcm:
Looks good to me.

PR #2016, opened 2015-02-06
Tweak code to conform to the voltdb change annotation regime.

paulmartel:
I decided that there was no easy way to break up these changes, so here they are all at once.
Most of them are fairly mechanical according to specific rules listed below.
None of the changes is intended to change system behavior.
It is all intended to streamline the upcoming hsql upgrade.

The change annotations have been verified as well-formed by an automated script (soon to be made public).

Normalize the syntax of the annotated comments around Volt changes and cherry picked changes from later hsql versions.

Back out extraneous whitespace changes from the original hsql code, including blank lines before and after the annotated changed blocks.

Back out most changes to "import blocks" as potentially "conflicty"
-- we may want to put some of these back in with proper annotations where they really help readability.

Drop "nested" annotated change blocks that were confusing my sed script.

Undo some well-intentioned IDEs' automatic changes:
Auto-reformat of:
    // fall through
to:
    // $FALL-THROUGH$
Auto-removal of redundant casts.
Auto-insertion of @Override annotations.

Make the replacement of the symbol:
    hsqldb
with:
    hsqldb_voltpatches
more universal (especially where it does not matter, in comments and unused help strings, etc.) to make it easier to automate that substitution.

In a couple of places, I reordered some changed statements to make it easier to cover in fewer annotated change blocks.

In a couple of places I tweaked brace placement etc.


src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -246,14 +246,12 @@ private void resolveColumnReferences() {
                                                rangeVariables.length, false);
         }
 
-    /************************* Volt DB Extensions *************************/
+        // A VoltDB extension to support ???
```
wolffcm: Something to do with aliases in the GROUP BY or ORDER BY clauses?  I seem to remember that Xin did this one.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -947,9 +942,9 @@ static VoltXMLElement voltGetXMLExpression(QueryExpression queryExpr, Expression
     }
 
     /** return a list of VoltXMLElements that need to be added to the statement XML for LIMIT and OFFSET */
-    protected static List<VoltXMLElement> voltGetLimitOffsetXMLFromSortAndSlice(Session session, SortAndSlice sortAndSlice)
-            throws HSQLParseException {
-        List<VoltXMLElement> result = new ArrayList<>();
+    protected static java.util.List<VoltXMLElement> voltGetLimitOffsetXMLFromSortAndSlice(Session session, SortAndSlice sortAndSlice)
+            throws org.hsqldb_voltpatches.HSQLInterface.HSQLParseException {
```
wolffcm: This seems to hurt readability a bit, but I'm willing to live with it if it makes upgrading easier.  Is it still okay to write new code using just "List" instead of "java.util.List" though?

wolffcm:
This looks good to me.  If everything still compiles I think you should merge it.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
jxchinaren: Yes, I believe it is for GROUP BY alias.

paulmartel: Thanks. I'll complete the comment.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
paulmartel: I may have gone a little overboard for some file(s), but the general idea is that the import section has been an area of high contention for code merges. I try to avoid changes there, but if changes are warranted, I try to make them in a single annotated block. It's a judgment call. My tipping point for when a change there is warranted is ROUGHLY when the annotated change to the import section contains more text than I am saving by not having to use prefixes. One use of a prefix does not justify an import -- as it would if I was coding native VoltDB code and letting eclipse do the work. Three or four probably make it worth it. And once the "ice is broken" by a file's first import change, I tend to be much more liberal about others.
As for what's still OK, practically speaking, under normal circumstances, if I saw ANY properly annotated change in the imports, I'd be tempted to "let it ride" even as the file's code evolved. If I ran into a non-annotated or improperly annotated import change, I'd probably stop to ask "what is this change really saving me?"

PR #2018, opened 2015-02-07
ENG-619 upload client jar to maven central

vtkstef:


upload.gradle
```diff
@@ -0,0 +1,190 @@
...
+
+               scm {
+                   url 'git@github.com:VoltDB/voltdb.git'
+                   connection 'scm:git@github.com:VoltDB/voltdb.git'
+                   developerConnection 'scm:git@github.com:VoltDB/voltdb.git'
+               }
+
+               licenses {
+                   license {
+                       name 'GNU Affero General Public License Version 3'
```
akhanzode: Is this correct lic?

upload.gradle
```diff
@@ -0,0 +1,190 @@
...
+    include 'org/voltdb/utils/SerializationHelper.java'
+    include 'org/voltdb/utils/RowWithMetaData.java'
+    include 'org/voltdb/utils/CSVDataLoader.java'
+    include 'org/voltdb/utils/CSVBulkDataLoader.java'
+    include 'org/voltdb/utils/CSVTupleDataLoader.java'
+    include 'org/voltdb/VoltOverflowException.java'
+    include 'org/voltdb/VoltTable.java'
+    include 'org/voltdb/VoltTableRow.java'
+    include 'org/voltdb/VoltType.java'
+    include 'org/voltdb/VoltTypeException.java'
```
akhanzode: Thought we were not packaging sources.

upload.gradle
vtkstef: Sonatype won't let us publish to maven central without sources. I talked it over with John H and we decided to publish the sources under the Affero license.

akhanzode:
this looks good. as long as folks know how to publish new versions (wiki) I am good with merging.

rmorgenstein:
Instead of writing a wiki, I would prefer you to write a jenkins job.  Make it start with 'kit-job' and add a call to it from kit-tools-publish-kit-masterjob.

akhanzode:
Should we package Export client classes as well so people can write custom connectors without having server.??

PR #2019, opened 2015-02-09
Fix issues with INSERT INTO ... SELECT and export tables

wolffcm:
These changes fix two issues found recently in this area:
- ENG-7633, planner crash on insert into select for export table with no partitioning column defined
- ENG-7665, dropped export rows for insert into select in a MP stored procedure (or MP ad hoc)

The fix for ENG-7633 is to guard against using insert into select with export tables with no partition column in the planner.

For ENG-7665 the fix is to skip the de-duping filter in the EE when the rows in the subquery are from a partitioned source.  This info is passed from the planner to the EE in the JSON for the insert plan node.

There is a complementary pro pull request with export tests that cover insert into select:
https://github.com/VoltDB/pro/pull/674


src/frontend/org/voltdb/planner/InsertSubPlanAssembler.java
```diff
@@ -59,6 +62,7 @@ AbstractPlanNode nextPlan() {
         Table targetTable = insertStmt.m_tableList.get(0);
         targetTable.getTypeName();
         StmtSubqueryScan subquery = insertStmt.getSubqueries().get(0);
+        boolean subqueryIsMultiFragment = subquery.getBestCostPlan().rootPlanGraph.hasAnyNodeOfType(PlanNodeType.SEND);
```
jxchinaren: PlanAssembler uses receive node to indicate distributed plan. Looking for PlanNodeType.RECEIVE should be more reliable to indicate multi fragment plan. SEND is a also needed for single fragment plan. Am I right?

jxchinaren:
Will ECO team work on tests or should I expect some test cases to validate:
1) Guards on export table cases: MP with partitioned export table without partition key (also AdHoc case)
2) Works for SP with partitioned export table without partition key
3) Partition to partition case without dropping rows
etc...

The other looks good to me.

src/frontend/org/voltdb/planner/InsertSubPlanAssembler.java
wolffcm: Hmmm... Maybe RECEIVE would be better, but I'm not sure. This code has been in there a while, and works okay for normal tables.  My guess is that the SEND node isn't added to the plan for subqueries until later?  I'll look into it.

wolffcm: I mean the top-level SEND that returns the results to the stored procedure isn't added until later, so we'd only have a SEND in a subquery if it was already multi-fragment.

wolffcm:
Thanks, Xin.  The tests for this are in the pro pull request.

jxchinaren:
Where are the tests for NPE guards on EXPORT table?

wolffcm:
Squash merged with master.

PR #2023, opened 2015-02-10
ENG-7673. Fix uniquedevices and voter demos so they work in docker

rmorgenstein:
The run.sh target that started the server in the background (nohup_server) was not working.  It was trying to load the ddl before the server was up.  
- Renamed function from nohup_server to background_server_andload - it no longer uses nohup, and does more than run the server
- Stopped using nohup - using voltdb to background it instead (-B)
- Added a wait loop before loading the ddl - it times out after 30 seconds.

While in there, I also removed rejoin. It is not part of any demo and is useless on a single server.

### DIFF
--- a/examples/uniquedevices/run.sh
+++ b/examples/uniquedevices/run.sh
@@ -110,7 +110,7 @@ function background_server_andload() {
     sleeptime=0
     until ../../bin/sqlcmd  --query=' exec @SystemInformation, OVERVIEW;' > /dev/null 2>&1
     do
-        sleeptime=$((sleeptime + 1))
+        sleeptime=$((sleeptime + 2))
         sleep 2
         echo " ... Waiting for VoltDB to start"
         if [[ $sleeptime -gt 30 ]]
--- a/examples/uniquedevices/run.sh
+++ b/examples/uniquedevices/run.sh
@@ -104,11 +104,22 @@ function init() {
     sqlcmd < ddl.sql
 }
 
-function nohup_server() {
-    jars-ifneeded
+function background_server_andload() {
     # run the server
-    nohup $VOLTDB create -d deployment.xml -l $LICENSE -H $HOST > nohup.log 2>&1 &
-    sqlcmd < ddl.sql
+    voltdb create -B -d deployment.xml -l $LICENSE -H $HOST > nohup.log 2>&1 &
+    sleeptime=0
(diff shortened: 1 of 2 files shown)


examples/uniquedevices/run.sh
```diff
@@ -104,11 +104,22 @@ function init() {
...
-    jars-ifneeded
+function background_server_andload() {
     # run the server
-    nohup $VOLTDB create -d deployment.xml -l $LICENSE -H $HOST > nohup.log 2>&1 &
-    sqlcmd < ddl.sql
+    voltdb create -B -d deployment.xml -l $LICENSE -H $HOST > nohup.log 2>&1 &
+    sleeptime=0
+    until ../../bin/sqlcmd  --query=' exec @SystemInformation, OVERVIEW;' > /dev/null 2>&1
+    do
+        sleeptime=$((sleeptime + 2))
```
prosegay: you can use $SECONDS to do this more nicely

examples/voter/run.sh
```diff
@@ -91,17 +91,22 @@ function init() {
...
-
-# run the voltdb server locally
-function rejoin() {
-    # run the server
-    voltdb rejoin -H $HOST -d deployment.xml -l $LICENSE
+function background_server_andload() {
+    # run the server in the background
+    voltdb create -B -d deployment.xml -l $LICENSE -H $HOST > nohup.log 2>&1 &
+    sleeptime=0
+    until ../../bin/sqlcmd  --query=' exec @SystemInformation, OVERVIEW;' > /dev/null 2>&1
```
prosegay: don't use the relative path, use the env. var for voltdb_home already in the script I believe.

examples/voter/run.sh
```diff
@@ -91,17 +91,22 @@ function init() {
...
-    jars-ifneeded
-    # run the server
-    nohup $VOLTDB create -d deployment.xml -l $LICENSE -H $HOST > nohup.log 2>&1 &
-    sqlcmd < ddl.sql
-}
-
-# run the voltdb server locally
-function rejoin() {
-    # run the server
-    voltdb rejoin -H $HOST -d deployment.xml -l $LICENSE
```
prosegay: just make a function for the test/wait and call it from the nohup_server

rmorgenstein:
Sqash-merged. Closing

PR #2024, opened 2015-02-10
[review only] Eng 7667

akhanzode:
will squash


vtkstef:
looks good. Let's remember to modify our app-fastdata demos too...

akhanzode:
squash merged.

PR #2025, opened 2015-02-10
ENG-7223 wildcard class drop

wijjo:
Improved the regex generator in ClassMatcher so that '*' allows an empty string and '.' is forced to be literal in the regex.

### DIFF
--- a/src/frontend/org/voltdb/compiler/ClassMatcher.java
+++ b/src/frontend/org/voltdb/compiler/ClassMatcher.java
@@ -67,8 +67,10 @@ public class ClassMatcher {
             classNamePattern = classNamePattern.substring(0, indexOfDollarSign);
         }
 
-        String regExPreppedName = preppedName.replace("**", "[\\w.\\$]+");
-        regExPreppedName = regExPreppedName.replace("*",  "[\\w\\$]+");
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/compiler/ClassMatcher.java
```diff
@@ -67,8 +67,15 @@ public ClassNameMatchStatus addPattern(String classNamePattern) {
...
         }
 
-        String regExPreppedName = preppedName.replace("**", "[\\w.\\$]+");
-        regExPreppedName = regExPreppedName.replace("*",  "[\\w\\$]+");
+        // Substitution order is critical.
+        // Keep track of whether or not this is a wildcard expression.
+        // '.' is specifically not a wildcard.
+        String regExPreppedName = preppedName.replace(".", "[.]");
+        boolean isWildcard = regExPreppedName.contains("*");
+        if (isWildcard) {
```
jhugg: Is this just a perf optimization or am I missing something?

jhugg: Oh I see. Never mind.

jhugg:
Looks good to this guy.

PR #2026, opened 2015-02-11
ENG-7676 patch for fixing RestRoundTripTest

dweiss88:


akhanzode:
Push merge if jenkins is good.

PR #2027, opened 2015-02-11
Eng 6923 inexists sqlcoverage

jxchinaren:


tests/scripts/examples/sql_coverage/basic-select.sql
```diff
@@ -89,3 +89,14 @@ SELECT * FROM @fromtables LHS INNER JOIN @fromtables B19RHS ON LHS.@idcol = B19R
...
+
+
+--- test IN/EXISTS predicate
+--- Use @columntype instead of @comparabletype because Hsql sometimes return wrong answers between integer and decimal comparison for IN.
+SELECT * FROM @fromtables A WHERE EXISTS ( SELECT * FROM @fromtables B WHERE B._variable[@columntype] _cmp A._variable[@columntype] )
+SELECT * FROM @fromtables A WHERE _variable[@columntype] IN ( SELECT _variable[@columntype] FROM @fromtables B )
+SELECT * FROM @fromtables A WHERE _variable[@comparabletype] _cmp @comparableconstant AND EXISTS ( SELECT * FROM @fromtables B WHERE B._variable[@columntype] _cmp A._variable[@columntype] )
+
+-- IN/EXISTS with other combination
+SELECT _variable[#grouped @columntype], COUNT(*) FROM @fromtables A WHERE EXISTS ( SELECT * FROM @fromtables B WHERE B._variable[@columntype] _cmp A._variable[@columntype] ) GROUP BY __[#grouped]
```
paulmartel: I don't think that the case of the GROUP BY in the parent is very interesting -- having a GROUP BY in the subquery is more interesting -- with the most interesting cases being something like:

     "_variable... IN (select MAX(_variable..."
     "EXISTS (SELECT _variable[#grouped]... HAVING MAX(_variable) _cmp  A._variable[@columntype] )"

Also, a simple join inside IN or EXISTS would also be an interesting case.

tests/scripts/examples/sql_coverage/advanced-ints.sql
```diff
@@ -1,2 +1,9 @@
 <configure-for-ints.sql>
 <advanced-template.sql>
+
+--- Test IN/EXISTS Advanced cases
+SELECT * FROM @fromtables A WHERE _variable[@columntype] _maybe IN ( SELECT MIN(_variable[@columntype]) FROM @fromtables B WHERE B._variable[@columntype] _cmp A._variable[@columntype] )
+SELECT * FROM @fromtables A WHERE EXISTS ( SELECT _variable[#GB]  FROM @fromtables B GROUP BY B.__[#GB] HAVING MAX(B._variable[@columntype]) _cmp  A._variable[@columntype] )
+
+{@jointype = "_jointype"}
+SELECT * FROM @fromtables LHS @jointype JOIN @fromtables B25RHS ON LHS.@idcol = B25RHS.@idcol where LHS._variable[@columntype] NOT IN (SELECT _variable[@columntype] FROM @fromtables IN_TABLE)
```
paulmartel: 2 comments:    

It's a little strange to set @jointype to _jointype this late -- just use _jointype in the 1 statement instead of @jointype.

Repeating a comment from last round: a join INSIDE an IN subselect would also be interesting.

PR #2032, opened 2015-02-13
ENG-7701 Fix explain command broken by ENG-7556

wijjo:
Was shifting past the verb in the string twice. Fix was to delete the original shifting code.


mikeismert:
Seems fine to me.

PR #2037, opened 2015-02-13
ENG-7647 Reject all quoted schema names

wijjo:
Still waiting for junit on Jenkins

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
@@ -64,8 +64,10 @@ public class ParserDQL extends ParserBase {
     HsqlException                  lastError;
     boolean                        strictSQLNames;
     boolean                        strictSQLIdentifierParts;
-    // VoltDB: reject quoted (delimited) names for now. TODO: Set from property?
-    boolean                        rejectQuotedSchemaObjectNames = true;
+    // A VoltDB extension to reject quoted (delimited) names.
+    // TODO: Set flag from property?
+    boolean rejectQuotedSchemaObjectNames = true;
+    // End of VoltDB extension
 
     //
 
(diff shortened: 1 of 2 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -64,6 +64,8 @@
     HsqlException                  lastError;
     boolean                        strictSQLNames;
     boolean                        strictSQLIdentifierParts;
+    // VoltDB: reject quoted (delimited) names for now. TODO: Set from property?
+    boolean                        rejectQuotedSchemaObjectNames = true;
```
paulmartel: 1) Have you considered the relative merits of a static vs non-static member for this  -- I think that a ParserDQL instance has a rather limited lifetime/accessibility. I'm not sure what makes more sense in a possible future where it is overridable -- maybe for "set from property" a static member set once is more efficient?

2) I have tools that expect all of our "manual" changes to hsqldb_voltpatches modules to be surrounded with marker comments that look like:

        // A VoltDB extension to <describe the intent of your changeset>
        <your new code here>
        // End of VoltDB extension

Indentation optional -- I usually indent these markers like normal comments.

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
```diff
@@ -474,4 +472,33 @@ public void testNullAnnotation() throws IOException {
...
+                catch (HSQLParseException e) {
+                    fail(String.format("Expected DDL to succeed: %s", ddl));
+                }
+            }
+            void testFailure(String ddl) {
+                try {
+                    hsql.runDDLCommand(ddl);
+                }
+                catch (HSQLParseException e) {
+                    return;
```
paulmartel: I usually try to be a little more rigorous and do some pattern matching on the test of the error message to assert that the error is the intended one.

Similarly, if the wrong exception type got thrown, do we want to catch and count that as a test Failure rather than as a test Error?

PR #2038, opened 2015-02-14
Eng 7389 sqlcmdtest framework

paulmartel:
Introducing the sqlcmdtest framework.
Includes tests for ENG-7389 (column names like 'fileXYZ' and similar cases).


tests/sqlcmd/sqlcmdtest.py
```diff
@@ -0,0 +1,216 @@
...
+
+from optparse import OptionParser
+
+
+# recursive directory creator like linux "mkdirs -p"
+# Credit: cribbed from stackoverflow.com/questions/600268
+def mkdir_p(path):
+    try:
+        os.makedirs(path)
+    except OSError as exc: # Python >2.5
```
wolffcm: Will this run with older versions of python?  I think there's a few machines in Jenkins that don't have python 2.6 yet.  Something to keep in mind if you want to automate these tests.

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -0,0 +1,216 @@
...
+        return os.path.join(baseline_dir, script_subdir)
+    # Otherwise, assume that its path up to the first separator is equivalent
+    # to script_dir so that everything beyond that is relative to script_dir.
+    subdir_path_part = script_subdir[script_subdir.find(os.pathsep)+1:]
+    return os.path.join(baseline_dir, subdir_path_part)
+
+
+def launch_and_wait_on_voltdb(reportout):
+    # Launch a single local voltdb server to serve all scripted sqlcmd runs.
+    # The scripts are expected to clean up after themselves  -- and/or defensively
```
wolffcm: It might be a good idea to make the scripts clean up after themselves, by asserting that "show tables" has the expected 0 tables.  It can be tricky to put the genie back into the bottle once the horse is out of the barn.

wolffcm:
I really like this.  Please merge it.

I have a few observations related to usability.  Most of this falls into the category of "nice to have at some point."

If the test fails, I don't see any indication on the command line other than a "See diffs in..." message.  I think if python exits with a non-zero exit code, then `ant` will show a foreboding BUILD FAILED message that my brain has been trained to look for.  You can also imagine showing red/green text next to each script that was run, etc, etc.

For the "understanding what failed" part, you can imagine outputting an html file someday.  In the short term, even just outputting links to files as URLs, like
> file:///home/cwolff/workspace/voltdb/tests/sqlcmd/sqlcmdtest.report
means that I can right-click and open the file from a terminal, which is really handy.

All of this can be done after you merge, of course.  I'm glad to see that we'll have a nice framework moving forward.

paulmartel:
Good ideas, Chris. Thanks. Ruth tells me that Python 2.6 is a reasonable prereq. I'd like a script failure or error to turn jenkins red, so I will implement the exit code now. The rest I will ticket.

PR #2044, opened 2015-02-17
Add method findInvalidStatementInDdlBatch to SQLCommand

wolffcm:
This pull request is pursuant to ongoing sqlcmd DDL batching work.

### DIFF
--- a/src/frontend/org/voltdb/utils/SQLCommand.java
+++ b/src/frontend/org/voltdb/utils/SQLCommand.java
@@ -538,9 +538,7 @@ public class SQLCommand
        List<String> stmts = SQLParser.parseQuery(batch);
        int i = 0;
        for (String stmt : stmts) {
-           boolean isDdl = SQLParser.queryIsDDL(stmt);
-           boolean isLoadOrRemoveClasses = SQLParser.isLoadOrRemoveClassesDirective(stmt);
-           if (! isDdl && ! isLoadOrRemoveClasses) {
+           if (! SQLParser.queryIsDDL(stmt)) {
                return Optional.of(Pair.of(i, stmt));
            }
 
--- a/src/frontend/org/voltdb/utils/SQLCommand.java
+++ b/src/frontend/org/voltdb/utils/SQLCommand.java
@@ -47,6 +47,7 @@ import jline.console.CursorBuffer;
 import jline.console.KeyMap;
 import jline.console.history.FileHistory;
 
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -515,6 +518,39 @@ public static String readScriptFile(File file)
...
+    */
+   static Optional<Pair<Integer, String>> findInvalidStatementInDdlBatch(String batch) {
+
+       batch = Strings.nullToEmpty(batch);
+
+       List<String> stmts = SQLParser.parseQuery(batch);
+       int i = 0;
+       for (String stmt : stmts) {
+           boolean isDdl = SQLParser.queryIsDDL(stmt);
+           boolean isLoadOrRemoveClasses = SQLParser.isLoadOrRemoveClassesDirective(stmt);
```
jxchinaren: Why LOAD/REMOVE is special? It's not DDL, does the SQLParser.queryIsDDL handle it differently?

wolffcm: I don't know if it's DDL or not.  Does it belong in a batch?  If not, I can take it out.

jxchinaren: LOAD/REMOVE are the sqlcmd directives and not DDL statements that the server can understand.

paulmartel: The SQLParser.queryIsDDL only recognizes alter, create, drop, export, and partition.
The load/remove check is redundant.

paulmartel: Sorry. I misread the code. I meant to say that the load/remove check is wrong.
These are not ddl in any useful sense.

wolffcm:
I pushed another commit.

src/frontend/org/voltdb/utils/SQLCommand.java
paulmartel: This code seems a bit bloated compared to the description in the ticket.
I thought that we had agreed that for the sake of expedience and consistency with direct calls to @AdHoc we were just going to check the first keyword in the batch and leave it to @AdHoc to enforce consistency and give a reasonable error message (for all callers, including non-ddl batch callers).

wolffcm:
I misunderstood what the ticket was asking for.  I'll push another branch today.

PR #2045, opened 2015-02-17
ENG-7752 - add retry for assert failure

tkcampbell:


tests/test_apps/schemachange/src/schemachange/SchemaChangeUtility.java
```diff
@@ -103,8 +104,15 @@ static ClientResponse callROProcedureWithRetry(Client client, String procName, i
...
                     log.error(((ClientResponseImpl)cr).toJSONString());
                     // for starters, I'm assuming these errors can't happen for reads in a sound system
-                    assert(false);
-                    System.exit(-1);
+                    if (retry < 3) {
+                        // Cluster might not have come back get, retry after a delay
+                        retry++;
+                        try{ Thread.sleep(5 * 1000); } catch (InterruptedException ignore) {}
+                        break;
+                    } else {
```
prosegay: Tim, make the retry for the specific conditions in the ticket, not just any error. Also, don't add a delay to the retry loop.

tkcampbell: I thought I did; the error refers to a block that's only called when ClientResponse.USER_ABORT shows up, so that's where I added the retry. I'm not sure how to make it more specific.

Removed the delay in the retry loop

PR #2047, opened 2015-02-17
Add initial support for file batch DDL sqlcmd feature

jxchinaren:
For CORE team review

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -881,8 +881,8 @@ public class SQLParser extends SQLPatternFactory
         private final File m_file;
         private final boolean m_batch;
 
-        FileInfo(File f, boolean b) {
-            m_file = f;
+        FileInfo(String fileName, boolean b) {
+            m_file = new File(fileName);
             m_batch = b;
         }
 
@@ -904,11 +904,11 @@ public class SQLParser extends SQLPatternFactory
     {
         Matcher batchMatcher = FileBatchToken.matcher(statement);
         if (batchMatcher.matches()) {
-            return new FileInfo(new File(batchMatcher.group(1)), true);
+            return new FileInfo(batchMatcher.group(1), true);
         }
         Matcher matcher = FileToken.matcher(statement);
         if (matcher.matches()) {
-            return new FileInfo(new File(matcher.group(1)), false);
+            return new FileInfo(matcher.group(1), false);
         }
 
         return null;
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -362,7 +362,8 @@ public class SQLParser extends SQLPatternFactory
     private static final Pattern SemicolonToken = Pattern.compile("^.*\\s*;+\\s*$", Pattern.CASE_INSENSITIVE);
     private static final Pattern RecallToken = Pattern.compile("^\\s*recall\\s+([^;]+)\\s*;*\\s*$", Pattern.CASE_INSENSITIVE);
     private static final Pattern FileToken = Pattern.compile("^\\s*file\\s+['\"]*([^;'\"]+)['\"]*\\s*;*\\s*", Pattern.CASE_INSENSITIVE);
-
+    private static final Pattern FileBatchToken = Pattern.compile("^\\s*file\\s+-batch\\s+['\"]*([^;'\"]+)['\"]*\\s*;*\\s*",
+            Pattern.CASE_INSENSITIVE);
     // Query Execution
     private static final Pattern ExecuteCall = Pattern.compile("^(exec|execute) ", Pattern.MULTILINE + Pattern.CASE_INSENSITIVE);
     // Match queries that start with "explain" (case insensitive).  We'll convert them to @Explain invocations.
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -876,17 +877,33 @@ public static ParseRecallResults parseRecallStatement(String statement, int line
...
-    public static File parseFileStatement(String statement)
+    public static FileInfo parseFileStatement(String statement)
     {
-        Matcher matcher = FileToken.matcher(statement);
-        if (matcher.matches()) {
-            return new File(matcher.group(1));
+        Matcher batchMatcher = FileBatchToken.matcher(statement);
+        if (batchMatcher.matches()) {
+            return new FileInfo(new File(batchMatcher.group(1)), true);
+        } else {
```
paulmartel: eliminate else after a return.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -236,6 +271,29 @@ private static String handleTranslatedCommands(String lineIn)
         return lineIn;
     }
 
+    public static class QueryInfo {
+        StringBuilder query;
+        boolean batch = false;
```
wolffcm: I could envision this class as having more members and used in more places in the future.  It would be good to make the members private and provide accessors.  Also, can this class be made immutable?  (These are both items in the Effective Java book.)

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -236,6 +271,29 @@ private static String handleTranslatedCommands(String lineIn)
         return lineIn;
     }
 
+    public static class QueryInfo {
+        StringBuilder query;
```
wolffcm: We seem to use the word "query" pretty freely in our code to refer to any SQL statement, but really it refers to statements (like SELECT) that return rows from the database.  Since I think you mean to include DDL statements too here, I think `StmtInfo` would be a better name for this class?

jxchinaren: Did you see that we use "query" variable as a StringBuilder all over to concat all the input DDL/Queries/Directives, etc. I guess I put a blind eye on it and follow it.

I would love to do a refactor work as a separate commit to fix it.

wolffcm: Sounds good, thanks.

src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -876,17 +877,40 @@ public static ParseRecallResults parseRecallStatement(String statement, int line
...
+    public static FileInfo parseFileStatement(String statement)
     {
+        Matcher batchMatcher = FileBatchToken.matcher(statement);
+        if (batchMatcher.matches()) {
+            return new FileInfo(new File(batchMatcher.group(1)), true);
+        }
         Matcher matcher = FileToken.matcher(statement);
         if (matcher.matches()) {
-            return new File(matcher.group(1));
+            return new FileInfo(new File(matcher.group(1)), false);
```
paulmartel: minor: you could factor the "new File(filepath)" step into the FileInfo constructor.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -287,7 +360,7 @@ public static void executeNoninteractive() throws Exception
     /// input loop can proceed to the next line.
     //TODO: There have been suggestions that some or all of these directives could be made
     // available in non-interactive contexts. This function is available to enable that.
-    private static boolean executesAsSimpleDirective(String line) throws Exception {
+    private static boolean executesShowCmdAsSimpleDirective(String line) throws Exception {
```
paulmartel: This renaming is wrong -- the function also implements the Help directive -- revert the name.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -434,7 +507,24 @@ private static void printTables(final String name, final Collection<String> tabl
...
     }
 
-    public static String readScriptFile(File file)
+    // function update the queryBatchList and query
+    private static StringBuilder updateQueryBatch(List<QueryInfo> contentInfo,
+            List<QueryInfo> queryBatchList, StringBuilder query) {
+        if (m_inFileBatchMode) {
+            assert(contentInfo.size() == 1);
+            // found contents for FILE cmd, flush the previous lines into a batch
+            queryBatchList.add(new QueryInfo(query, false));
```
paulmartel: Should we skip this QueryInfo entry if query.isEmpty()?

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -434,7 +507,24 @@ private static void printTables(final String name, final Collection<String> tabl
...
+            query = new StringBuilder();
+
+            queryBatchList.add(new QueryInfo(contentInfo.get(0).getQuery(), true));
+        } else {
+            queryBatchList.addAll(contentInfo);
+        }
+
+        return query;
```
paulmartel: It does not seem right in non-batch mode to "place on hold" the current contents of query while we queue batch contents before it.
Would it be better to always queue query in this function, have no return value, and let the callers always reset the StringBuilder?

tests/frontend/org/voltdb/utils/TestSqlCmdErrorHandling.java
```diff
@@ -189,12 +189,18 @@ private int callSQLcmdBase(boolean stopOnError, String inputText,
...
-            pb = new ProcessBuilder(commandPath, "--ddl-file=" + f.getPath());
+        if (batchModeDDL) {
+            pb = new ProcessBuilder(commandPath);
+            File batchScript = new File("script.sql");
+            batchScript.deleteOnExit();
+            fos = new FileOutputStream(batchScript);
+            fos.write(new String("FILE -batch " + f.getPath()).getBytes());
+            fos.close();
+
+            pb.redirectInput(batchScript);
```
paulmartel: I like having a few basic batch mode ddl unit tests here.

I don't remember any plans to deprecate all testing for "--ddl-file=".
Maybe these test cases should exist in parallel?

paulmartel:
End of review round.

PR #2048, opened 2015-02-17
Add method batchBeginsWithDDLKeyword to SQLParser

wolffcm:
Pursuant to ongoing FILE -batch.  I put this method into SQLParser since it seemed to belong there.

### DIFF
--- a/tests/frontend/org/voltdb/parser/TestSQLParser.java
+++ b/tests/frontend/org/voltdb/parser/TestSQLParser.java
@@ -106,5 +106,20 @@ public class TestSQLParser extends TestCase {
         assertFalse(SQLParser.batchBeginsWithDDLKeyword(
                 "/* create table was done earlier... */\n"
                 + "update t set i = 70 where i > 69;"));
+
+        // Near misses that might appear in a ddl.sql file
+        // but that cannot be batched
+
+        assertFalse(SQLParser.batchBeginsWithDDLKeyword(
+                "load classes foo-bar.jar"));
+
+        assertFalse(SQLParser.batchBeginsWithDDLKeyword(
+                "remove classes foo-bar.jar"));
+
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/parser/TestSQLParser.java
```diff
@@ -0,0 +1,110 @@
...
+                + "update t set i = 70 where i > 69;"));
+
+        assertFalse(SQLParser.batchBeginsWithDDLKeyword(
+                "// create table was done earlier...\n"
+                + "update t set i = 70 where i > 69;"));
+
+        assertFalse(SQLParser.batchBeginsWithDDLKeyword(
+                "/* create table was done earlier... */\n"
+                + "update t set i = 70 where i > 69;"));
+    }
```
paulmartel: Add negative test cases for near-miss non-ddl statements that could cause future confusion: load, remove, exec, import, file.

wolffcm: I can certainly do this.

By the way `queryIsDDL` will return true if the first keyword is `IMPORT`.  Should we actually return false for this case, or let the server take care of it?  It seems to be the first keyword in an IMPORT CLASS statement, but I'm not clear on how that statement is used.

wolffcm: IMPORT CLASS seems to be treated as DDL by the server (even if it's not supported by ad hoc DDL) so I'm tempted to make sqlcmd continue to consider IMPORT CLASS to be DDL---the error messages will be consistent.
```
1> import class org.voltdb.compiler.VoltCompiler
2> ;
AdHoc DDL contains an unsupported DDL statement: import class org.voltdb.compiler.VoltCompiler
```

jxchinaren:
The other looks good to me.

PR #2049, opened 2015-02-17
ENG-7496 Replace python process with server

wijjo:
Mainly had to deal with quoted arguments because exec doesn't filter them the way the shell does. We were improperly combining multiple arguments in quotes and also unnecessarily quoting paths. This solution won't handle all possible shell quoting, but is good enough for running volt.


jhugg:
This looks ok to me. Nice work.

PR #2052, opened 2015-02-18
ENG-7766: Fix dynamic DR DDL parsing failure.

nshi:
The DR DDL statement was not recognized as a DDL statement, added it to
the while list.


afrogers:
Looks good, merging

PR #2057, opened 2015-02-19
New sqlcmdtest features including stderr baselines.

paulmartel:


tests/sqlcmd/baselines/simple/sqlcmd1.errbaseline
```diff
@@ -0,0 +1 @@
+Strict java memory checking is enabled, don't do release builds or performance runs with this enabled. Invoke "ant clean" and "ant -Djmemcheck=NO_MEMCHECK" to disable.
```
wolffcm: I wonder if we shouldn't just filter this message after sqlcmd has finished but before comparing against the baseline.  That way error-free scripts wouldn't need stderr baselines.  Also, the tests would pass on both debug and release build.

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -95,40 +95,162 @@ def launch_and_wait_on_voltdb(reportout):
...
+                                 # survives as \g<1>
+            [0-9]+\.[0-9]+s      # also required, replaced with #.##s
+            """, re.VERBOSE)
+    for line in outbackin:
+        # Note len(cleanedline) here counts 1 EOL character.
+        # Preserve blank lines as is -- there's no need to try cleaning them.
+        if len(line) == 1:
+            cleanedout.write(line)
+            continue
+        cleanedline = memory_check_matcher.sub("", line)
```
wolffcm: Oh wait, I guess you are filtering it...

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -95,40 +95,162 @@ def launch_and_wait_on_voltdb(reportout):
...
     parser.add_option("-b", "--baselines", dest="baseline_dir", default="./baselines",
                       help="top level test output baseline directory")
-    parser.add_option("-o", "--reportfile", dest="reportfile",
+    parser.add_option("-o", "--report_file", dest="report_file",
                       default="./sqlcmdtest.report",
                       help="report output file")
     parser.add_option("-r", "--refresh", dest="refresh",
                       action="store_true", default=False,
                       help="enable baseline refresh")
+    parser.add_option("-p", "--purge_only", dest="purge_only",
```
wolffcm: Could also add the .out files, etc to `.gitignore`?

tests/sqlcmd/baselines/simple/sqlcmd1.errbaseline
wolffcm: Ok, so I see you are filtering this message in the .py script below.  So now I'm not sure how the comparison succeeds with this message here?

paulmartel: I see. The stderr version of the warning is not prefaced with "WARN: " DOH! 
I'll generalize the pattern.

BTW, don't get lulled into thinking there is anything so mundane as a "release build" vs. "debug build" distinction being made here. No. As the message explains, to disable this "development mode" memory checking you have to build with additional flags.
IIRC, this typically comes pretty far along in the test cycle (as in never on a branch test and not even for many master test runs). The flags get turned on for performance tests and I THINK for "one full test suite pass" before release.

wolffcm: Yes, I had forgotten about that.  It's a little strange that a true release build (i.e., the bits we ship with) needs flags in addition to `-Dbuild=release`.  I was going to say something about how maybe we should just disable the message in sqlcmd, but I've actually found it useful to remind me to rebuild with the NO_MEMCHECK flag.

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -95,40 +95,162 @@ def launch_and_wait_on_voltdb(reportout):
...
-            return
+    for waited in xrange(0, 19):
         empty_input.seek(0)
         waiting = subprocess.call(['../../bin/sqlcmd'], stdin=empty_input)
         if not waiting:
             break
+        if waited == 19:
+            reportout.write("voltdb server not responding -- so giving up\n")
+            kill_voltdb()
+            return
```
wolffcm: Another possible enhancement here: check and see if there's already a volt process running, and either error out or use the process if so.  (For example, I might start Volt from Eclipse for debugging.)

wolffcm:
Looks good to me.  Feel free to merge when you're ready.

wolffcm:
Looks good to me.  Feel free to merge when you're ready.

PR #2058, opened 2015-02-20
Fix the mismatch output columns for SendPlanNode, Not sure how it ever s...

jxchinaren:
...ucceeds before

### DIFF
--- a/src/frontend/org/voltdb/plannodes/SendPlanNode.java
+++ b/src/frontend/org/voltdb/plannodes/SendPlanNode.java
@@ -55,7 +55,7 @@ public class SendPlanNode extends AbstractPlanNode {
             int index = tve.resolveColumnIndexesUsingSchema(input_schema);
             tve.setColumnIndex(index);
         }
-        m_outputSchema.sortByTveIndex();
+        // output schema for SendPlanNode should not ever be changed
     }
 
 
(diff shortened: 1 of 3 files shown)


tests/testprocs/org/voltdb_testprocs/regressionsuites/fixedsql/fixed-sql-ddl.sql
```diff
@@ -285,3 +285,52 @@ create procedure one_string_scalar_param as
...
+  selling_price,
+  cost,
+  supplier,
+  safety_stock_days
+FROM product_changes
+WHERE location = ? 
+AND product_id = ? 
+ORDER by location, product_id, start_date;
+PARTITION PROCEDURE voltdbSelectProductChanges ON TABLE product_changes COLUMN location PARAMETER 0;
+-- ********************************** --
```
paulmartel: We should put a little more effort into simplifying/unifying regression repro cases -- did we really need a new custom test table with 20 columns to repro this issue. Reusing existing test schema for new tests keeps things simpler and more readable -- a table named P1 tells me "I'm (probably) partitioned" -- and reduces the risk if leaking proprietary info into our public code. It is worth the little bit of extra effort on the repro case -- which can also be educational (how general IS the bug?).

src/frontend/org/voltdb/plannodes/SendPlanNode.java
```diff
@@ -55,7 +55,7 @@ public void resolveColumnIndexes()
             int index = tve.resolveColumnIndexesUsingSchema(input_schema);
             tve.setColumnIndex(index);
         }
-        m_outputSchema.sortByTveIndex();
+        // output schema for SendPlanNode should not ever be changed
```
paulmartel: Now I am suspicious about the other 5 calls to sortByTveIndex.
Do they really have a purpose? -- I IMAGINE that they may be intended to make row-scans more cache-friendly on wide rows (?), would that even be a measurable effect? and maybe (?) some are needed for completeness/consistency with others (like maybe when inlining?). It's all very mysterious.

PR #2059, opened 2015-02-20
Eng 7761 annotate parser

paulmartel:
Accidentally picked up a master change from Ryan -- looks like a non-issue for CORE-team. Please ignore.
Github diff does not show this as well as meld, but this really is (I think) a whitespace/comment change (and breaking strings up with " + ".
I tried not to change the actual regexp content.

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -111,85 +111,73 @@ public class SQLParser extends SQLPatternFactory
         ).compile();
 
     //TODO: Convert to pattern factory usage below this point.
+    //TODO: Consider implementing FIXME/TODO comments -- and TESTING --
+    // prior to conversion to pattern factory.
 
     /**
      * CREATE PROCEDURE from Java class statement regex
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -111,85 +111,73 @@ public Exception(Throwable cause, String message, Object... args)
         ).compile();
 
     //TODO: Convert to pattern factory usage below this point.
+    //TODO: Consider implementing FIXME/TODO comments -- and TESTING --
```
wolffcm: Hmmm... a TODO item to implement other TODO items?  If it's risen to the level of a recursive TODO, maybe file a ticket?

wolffcm:
This change looks good to me.  It reduces the possibility of comments getting out-of-sync with what's actually there.  

Seems like you can't automatically merge because I merged a commit to CORE-team ahead of you.  Sorry about that.  GitHub's merge tool doesn't seem to be that smart---I don't think there's really any conflicts here.

PR #2060, opened 2015-02-20
ENG-7772 code refactor to reduce dependencies -- sqlcmd hang (likely) fi...

paulmartel:
...x.

In CoreUtils, use a throw-away "HOST" logger instance
when a utility thread runner unexpectedly catches an exception.
Avoids the class load dependencies of having a static VoltLogger member on stand-by.

In RealVoltDB, establish a "bring your own" VoltLogger policy to reduce dependencies
of RealVoltDB's "convenience statics callers" on its "HOST" VoltLogger.
They're more concerned with the deployment and build info than the RealVoltDB instance
or its logger. Allows the RealVoltDB instance to own its (non-static) VoltLoggers,
and avoids class load dependency chains through RealVoltDB.

In VoltLogger, eliminate circular CoreUtils dependencies by adding...
+    /// A thread factory implementation customized for asynchronous logging tasks.
See the remainder of that code comment for details.

Kill unused Volt*Logger.addSimpleWriterAppender(StringWriter writer) methods.
Kill unused Volt*Logger.setLevel(Level level) methods.

Also, simplify the code paths for synch log entry execution and asynch log entry submission
by branching them earlier in the VoltLogger method call stack and sharing only the truly
common low-level part (now factored out).


paulmartel:
Remember to synch this merge with the pro changeset: https://github.com/VoltDB/pro/pull/693

src/frontend/org/voltcore/logging/VoltLogger.java
```diff
@@ -65,77 +98,98 @@ public void run() {}
         public boolean isEnabledFor(Level level);
         public void log(Level level, Object message, Throwable t);
         public void l7dlog(Level level, String key, Object[] params, Throwable t);
```
wolffcm: What does "l7d" stand for?

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -547,9 +547,7 @@ boolean matches(String statement)
      */
     private static class WhitelistSupportedPreamblePattern extends CheckedPattern
     {
-        private static Pattern whitelistPattern = null;
-
-        static
+        private static Pattern initPattern()
```
wolffcm: I like Java's `static { ... }` a lot less after seeing the potential for weird race conditions here.

wolffcm: I mean, I'm just making an observation.  I like how you've addressed the problem by calling named static functions.

wolffcm:
This all looks fine to me.  Thanks for fixing this.

If we had coding standards I would say we should put among them, "never put anything in a static block that depends on non-system classes."

Will this work fix the sqlcmd unit test that's been sporadically failing?

paulmartel:
Your suggested coding standard seems a bit drastic.
As we discussed, static blocks are not much different from static attribute initializers.
Would you ban final static attributes of all non-system classes?
I like better the idea of being more attentive to findbugs reports.
These circularity issues are only listed as "Normal Priority". 
And there are other kinds of issues in the larger report that are listed as "High Priority".
Every now and then, someone writes a jira ticket that says "the findbugs report needs some attention"  I have no idea how the current report looks compared to just after it last got some attention.
In my experience, it's a rare combination of engineer and analysis tool that leads to a clean bill of health -- unless of course there's a "mute" option.
As for the particulars, I have a jenkins branch job defined somewhere that I can tweak to force it to retry the repro case on this branch. I'd like to see that run AND see this findbugs category closed out before I merge.

paulmartel:
I pushed one more commit -- should be reviewed -- when I realized that I had delayed the VoltLogger shutdown/flush to after the log4j shutdown. I used this order dependency as an excuse to more formally recognize the (optional) log4j shutdown as "THE FINAL" shutdown action (a new concept). I didn't refactor the ShutdownHooks class to radically simplify it, eliminating the (now) unused priority system and (now) unused capability to fire a normal hook on a crash -- but it's ripe for that (i.e. currently looking over-engineered).

wolffcm:
Hi Paul, I was wondering if you could provide a little context on the difference between log4j and VoltLogger services.  How do their responsibilities differ and why is it important for log4j to shutdown last?

PR #2061, opened 2015-02-20
Fix typo so build succeeds

wolffcm:
Build was failing on CORE-team branch.


paulmartel:
Sorry. Thx. I noticed that on my machine just before I left work today -- wasn't sure where/when/how it crept in. Still a little puzzled.

PR #2066, opened 2015-02-23
ENG-7761 re-enable sqlcmd list directive, parse other directives more co...

paulmartel:
...nsistently.

Generally improved the generality of whitespace matching.
The long-term goal is for directive patterns to be more forgiving so that better error messages can come from post-processing -- vs. fallback to @AdHoc on pattern mismatch.
The short-term goal is to not allow pattern matches like "list integer" or "recall integer" that could conceivably be mid-statement SQL and so SHOULD fall through to @AdHoc.
-- until we fix the SQLCommand directive processor to not look for mid-statement directives.

Also fixed ugly and crashy sqlcmd error message formatting.


.gitignore
```diff
@@ -11,6 +11,12 @@
 /.DS_Store
 .DS_Store
 .cache/
+*.err
```
wolffcm: It seems drastic to tell git not to track files with these endings for the whole tree.  Maybe I'm being paranoid.

wolffcm:
Will sqlcmdtest work with directives like show, list, recall, etc?  If so, it would be good to add the tests that lock down your fixes.

wolffcm:
End of review round.

.gitignore
paulmartel: Maybe -- I can back off.

paulmartel:
These changes were tested manually for now because not all commands are supported via stdin -- certainly not recall, and I've forgotten WRT list/show. I was going to look into this and add sqlcmdtest cases (if supported via stdin) as part of the testing ticket.
I THINK I recall some existing unit test framework that was supposed to be simulating interactive input, but I'm having trouble finding it now.

I am hopeful that the more advanced python subprocess API features will allow us to support "interactive mode" scripting as an option in a future version of sqlcmdtest.

paulmartel:
Pushed the reviewed part of this -- with a few small tests -- as ENG-7761-test-exec.

PR #2067, opened 2015-02-23
ENG-7552: Rip out spUniqueId

afrogers:


src/frontend/org/voltdb/SnapshotSaveAPI.java
```diff
@@ -140,13 +137,10 @@ public VoltTable startSnapshotting(
                 @Override
                 public void run() {
                     Map<Integer, Long> partitionTransactionIds = m_partitionLastSeenTransactionIds;
-                    Map<Integer, Long> partitionUniqueIds = m_partitionLastSeenUniqueIds;
 
-                    SNAP_LOG.debug("Last seen partition transaction ids " + partitionTransactionIds + " and unique ids " + partitionUniqueIds);
+                    SNAP_LOG.debug("Last seen partition transaction ids " + partitionTransactionIds);
```
wweiss-voltdb: After one full pass, my only concern is if we should remove the uniqueId from all debug output.

PR #2076, opened 2015-02-25
Fix EE issue with inverted parameters in IPC initialization causing ValG...

wweiss-voltdb:
...rind to fail. Also moved StackTrace from FatalException to debuglog. Finally added missing stored procedure back for DrEndToEnd.

### DIFF
--- a/src/ee/common/FatalException.cpp
+++ b/src/ee/common/FatalException.cpp
@@ -25,6 +25,7 @@
 #include <dlfcn.h>
 #include <stdio.h> // for fopen, fprintf, fclose
 #include <string.h>
+#include "common/debuglog.h"
 
 #ifdef MACOSX // mac os requires _XOPEN_SOURCE for ucontext for some reason
 #define _XOPEN_SOURCE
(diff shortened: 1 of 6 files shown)


src/ee/common/FatalException.cpp
```diff
@@ -41,67 +42,22 @@ FatalException::FatalException(std::string message,
...
-        }
-        ::free(function);
-    }
-
     if (bt) {
-        for (int ii=0; ii < m_traces.size(); ii++) {
-            const char* str = m_traces[ii].c_str();
+        StackTrace st;
+        // write header for backtrace file
+    	int numFrames = (int)st.m_traces.size();
```
paulmartel: Hint:  "./tools/licensescheck.py --fixws" run in the voltdb directory will fix the whitespace issues (replace the tabs) that are preventing ant licensecheck from passing on your branch.

src/ee/common/debuglog.cpp
```diff
@@ -0,0 +1,99 @@
...
+//	char **stack_strings;
+//
+//	stack_depth = backtrace(stack_addrs, max_depth);
+//	stack_strings = backtrace_symbols(stack_addrs, (int)stack_depth);
+//
+//	for (size_t i = 4; i < stack_depth; i++) {
+//	    printf("    %s\n", stack_strings[i]);
+//	}
+//	free(stack_strings); // malloc()ed by backtrace_symbols (1 array starting with a set of pointers to the data area of the same array)
+//}
```
paulmartel: Commented out code is not useful.
Commented out code with no commented justification for its semi-existence is worse than that.

src/ee/common/debuglog.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+StackTrace::~StackTrace() {
+    ::free(m_traceSymbols);
+}
+} // namespace voltdb
+
+using namespace voltdb;
+void printStackTrace_() {
+	StackTrace st;
```
paulmartel: Is this ugly little trailing '_' in the name supposed to mean something?

I think I would have named this print and made it a static method of StackTrace.

Also I would have taken the similar code out of FatalException and made it another static method of StackTrace -- give callers the option.

src/ee/common/debuglog.h
```diff
@@ -179,4 +179,24 @@ inline void outputLogHeader_(const char *file, int line, const char *func, int l
...
+
+	char** m_traceSymbols;
+	std::vector<std::string> m_traces;
+};
+} // namespace voltdb
+
+
+void printStackTrace_();
+#define PRINT_STACK_TRACE outputLogHeader_(__FILE__, __LINE__, __FUNCTION__, VOLT_LEVEL_TRACE);::printf("OUTPUTING STACK TRACE\n");printStackTrace_();::fflush(stdout)
+
```
paulmartel: Some people seem to have a problem \
with backslash newlines in macro definitions \
-- but not me.
Somepeoplefindlonglinesoftext(evenwithpunctuation,andespeciallywithnowhitespace)hardtoread.Thatwouldbeme.
So maybe the _ at the end of printStackTrace_ is there for breathing room?

src/ee/common/debuglog.h
```diff
@@ -179,4 +179,24 @@ inline void outputLogHeader_(const char *file, int line, const char *func, int l
     printf("[%s] [%s:%d:%s()] %s - ", type, file, line, func, time_str);
 }
 
+#include <vector>
```
paulmartel: mid-file #include is needlessly unconventional.
Also, no macros would be harmed by moving "namespace voltdb {" to its usual place near the top.

src/ee/voltdbipc.cpp
```diff
@@ -438,6 +438,7 @@ int8_t VoltDBIPC::initialize(struct ipc_command *cmd) {
     cs->hostId = ntohl(cs->hostId);
     cs->logLevels = ntohll(cs->logLevels);
     cs->tempTableMemory = ntohll(cs->tempTableMemory);
+    cs->createDrReplicatedStream = ntohl(cs->createDrReplicatedStream);
```
paulmartel: Because ntohl MIGHT turn a 0 into a non-0 or vice-versa?
Though I suppose it actually MIGHT if bool was actually a 64-bit value so you were reassigning it to a reordered 32-bit subset of itself. Seems risky. I'd skip it altogether, as the code was.

src/ee/voltdbipc.cpp
```diff
@@ -457,8 +458,8 @@ int8_t VoltDBIPC::initialize(struct ipc_command *cmd) {
                                  cs->partitionId,
                                  cs->hostId,
                                  hostname,
-                                 createDrReplicatedStream,
-                                 cs->tempTableMemory) == true) {
+                                 cs->tempTableMemory,
+                                 createDrReplicatedStream) == true) {
```
paulmartel: Interesting bit of trivia: VoltDBEngine::initialize always returns true -- if it returns at all -- so both the ipc and jni calls to it are testing its return value for nothing.
It's a fairly common form of silliness in and around VoltDBEngine.
Not terribly hard to clean up, taken a little at a time.

src/ee/common/debuglog.h
paulmartel: OUTPUTING?

src/ee/common/FatalException.cpp
```diff
@@ -41,67 +42,22 @@ FatalException::FatalException(std::string message,
...
+    	// Ignore the stack frames specific to StackTrace object
+    	fprintf(bt, "VoltDB Backtrace (%d)\n", numFrames-2);
```
paulmartel: As broke as you found it, but: the parenthesized number without units is pretty meaningless, as opposed to say "(%d stack frames)"

paulmartel:
End of review round -- the ntohl issue is the only real potential show-stopper I found.

src/ee/voltdbipc.cpp
wweiss-voltdb: This is a bad practice but in fact ExecutionEngineIPC represents the boolean as a 1 or 0 represented in an int and assigned into the stream as putInt. The conversion is not required, but it does add clarity for someone examining cs->createDrReplicationStream and seeing a 0x80000000. In any case it is only performed once during initialization.

src/ee/common/debuglog.cpp
wweiss-voltdb: Yea. I was following the pattern of those before me...
outputLogHeader_

PR #2077, opened 2015-02-26
ENG-7563 Fixed drop data when exporting a partitioned table bug

dweiss88:


tests/frontend/org/voltdb/catalog/TestCatalogDiffs.java
```diff
@@ -819,6 +819,44 @@ public void testRemoveNonUniqueIndex() throws IOException {
         verifyDiff(catOriginal, catUpdated);
     }
 
+    public void testExportRejectedIfNotEmpty() throws IOException {
```
akhanzode: Where in this test is export table that has data?

PR #2079, opened 2015-02-26
ENG-7564 create proc with dots

wijjo:
More extensive than needed, but "modernized" the complex CREATE PROCEDURE patterns.


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -356,6 +356,26 @@ private String checkIdentifierWithWildcard(
...
+     * @return the given identifier unmodified
+     * @throws VoltCompilerException
+     */
+    private String checkProcedureIdentifier(
+            final String identifier, final String statement
+            ) throws VoltCompilerException {
+        String retIdent = checkIdentifierStart(identifier, statement);
+        if (retIdent.contains(".")) {
+            String msg = String.format(
+                "Invalid procedure name \"%s\" in DDL: \"%s\"",
```
jhugg: Is this error message too generic? Do we know the problem is the "." here?

jhugg:
So I see anything specifically wrong here, but I'm not sure that's saying much. I like the direction of making everything more consistent, It's a long fix for a small issue, which adds risk; it should probably be two pull requests?

PR #2081, opened 2015-02-26
ENG-7570: removed -limitrows from JDBCLoader

tkcampbell:


src/frontend/org/voltdb/utils/JDBCLoader.java
```diff
@@ -461,10 +458,6 @@ private void produceFiles(long ackCount, long insertCount) {
 
             totalRowCnt = JDBCStatementReader.m_totalRowCount.get();
 
-            if (config.limitrows == -1) {
-                out_reportfile.write("Input stopped after "
```
vtkstef: you may remove the if block but please leave the report file write alone

tkcampbell: I thought I was; I only removed the if, and the  message that the writing stopped

vtkstef:
you have jenkins regression for org.voltdb.utils.TestJDBCLoader.testCommon please investigate and fix it

PR #2082, opened 2015-02-26
Stand alone interface to HSQL parser (Review only, do not merge)

wolffcm:
I did some work a week or two ago to see if using HSQL's tokenizer elsewhere in Volt would be feasible, in order to avoid some of the headaches caused by using regular expressions.  I hit the end of my timebox and this work still isn't ready for prime time, but I thought I'd share it in a pull request.

This changeset presents an interface to allow clients to create an iterable stream of tokens from a string of SQL text.  The big advantage here is:
- Comments and whitespace are automatically skipped
- Code that needs to reason about user input can be written more clearly

For an example of using these classes see the unit test in:
https://github.com/VoltDB/voltdb/blob/d0c83ba484e9ccc4edc69bb65113f4257e204bf7/tests/hsqldb/org/hsqldb_voltpatches/TestVoltTokenStream.java

Also see the example of a rewritten parseQuery() method in SQLParser.java below.

Issues with these changes:
- All identifiers are upper-cased, so for CREATE PROCEDURE this doesn't work, since Java identifiers are case-sensitive
- Not all input to sqlcmd are valid SQL tokens (e.g., the file names provided to the FILE command)

Neither of these are insurmountable---but even if they weren't, I think it would be possible to start replacing some of the functionality in SQLParser to reduce the amount of code that's there.


BillWhite:
Thanks.

The approach we're taking is to replace all of this lexical/parsing stuff with a generated lexer and parser, generated by Antlr4.  It's really convenient, which means it's probably somewhat slow.  But for the moment it seems to do the work for us.  I don't change the case on identifiers, but the lookup tables are all case-insensitive.  So error messages get the user's original case, but we get the right behavior.

Since this is mostly an exercise in learning SQL deeply, we will be happy if we can create a table, insert some data into it and select from it.

PR #2083, opened 2015-02-27
Support inline batching in sqlcmd

wolffcm:
This is a new pull request that adds a couple new commits to #2078 and addresses code review comments.  (I must have rebased my branch because I couldn't push the new commits.)

Only the last two commits contain new work that hasn't been reviewed.  Mostly this does better syntax error handling for the FILE command.  I also found that if we want to recover from exceptions thrown when locally parsing commands (like FILE) in sqlcmd, then we need to catch the execption sooner in order to continue.


wolffcm:
I pushed another commit that disallows file names that are only whitespace.  It's still not perfect, but I think we need to talk about what's valid here.

I also made more consistent that handling of empty batches.

PR #2086, opened 2015-02-27
ENG-7527 enable catalog updates with export properties changes

vtkstef:


akhanzode:
Can you add tests with LocalServer that modify this at runtime and test expected result of running export clients. Also test the cases when we will not allow changes.

vtkstef:
Considering the scope of the change (a couple of lines in CatalogDiffEngine) do you believe that a localServer test case would take other code paths which differ from the ones already tested in TestCatalogDiffs?

akhanzode:
Its not so much for diff path but test/code coverage so if accidental change breaks this we will catch.

vtkstef:
It's not just the diffing. It calls compileDeployment which will create an instance of the configured and modified export connector. I don't see yet what would make a local server test compelling.

PR #2088, opened 2015-02-27
ENG-7468: Exceptions caught and more useful errors given

tkcampbell:


src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -215,10 +215,14 @@ public static JSONObject parseJSONFile(String configInfoPath) {
             jsonObject = new JSONObject(builder.toString());
         } catch (FileNotFoundException e) {
             System.err.println("config log file '" + configInfoPath + "' could not be found.");
+            System.exit(1);
```
vtkstef: in line (126)[https://github.com/VoltDB/voltdb/blob/ENG-7468/src/frontend/org/voltdb/utils/Collector.java#L126] we exit with -1. Let's be consistent and use -1 here too

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -215,10 +215,14 @@ public static JSONObject parseJSONFile(String configInfoPath) {
...
         } catch (IOException e) {
             System.err.println(e.getMessage());
+            System.exit(1);
         } catch (JSONException e) {
-            System.err.println(e.getMessage());
+            System.err.println("Error with config file: " + configInfoPath);
+            System.err.println(e.getLocalizedMessage());
+            System.exit(1);
```
vtkstef: and here too

src/frontend/org/voltdb/utils/Collector.java
tkcampbell: I'll make a note that this is the standard return value for errors.

PR #2090, opened 2015-02-28
Fix IPC build compilation failure.

nshi:
Following the pattern of not catching SerializableEEException. The IPC
process will probably exit when it hits the exception.


afrogers:
Looks good, merging.

PR #2092, opened 2015-03-01
possible ee leak??

akhanzode:
Looks like fishy missing cleanup code.


paulmartel:
I'm pretty sure that you are patching dead code paths.
insertTuple possibly in ALL cases but definitely in these temp table insert cases ALWAYS returns true. These functions are virtually (SHOULD be literally, after a cast) calling "void TempTable::insertTempTuple". The if test SHOULD be removed.
I'm sorry that this code is so misleading.
IF the input tables are actually the leak, they must be leaking -- and will have to be cleaned up -- in a more generic code path. Xin was the last to deal with such issues. You might want to follow up with him.

src/ee/executors/distinctexecutor.cpp
```diff
@@ -108,6 +108,7 @@ bool DistinctExecutor::p_execute(const NValueArray &params) {
                            " output table '%s'",
```
jhugg: Is there any evidence we're hitting these paths? Do we see the "Failed to insert tuple from input table..." message in a log or in std out? We're supposed to see it in the log, right?

src/ee/execution/JNITopend.cpp
```diff
@@ -386,6 +386,7 @@ void JNITopend::pushExportBuffer(
     if (block != NULL) {
         jobject buffer = m_jniEnv->NewDirectByteBuffer( block->rawPtr(), block->rawLength());
         if (buffer == NULL) {
+            m_jniEnv->DeleteLocalRef(signatureString);
```
jhugg: This one, and the one in voltdbjni seem like sane things to do. We should make sure these changes get onto master, even if we're not sure if we're actually hitting these paths.

PR #2096, opened 2015-03-03
Eng 7792

paulmartel:
Indexes' use of  of CONCAT -- and || in recompiles where || gets translated into CONCAT! -- ran into a missing switch case.
It turns out the switch's default behavior (calling super to delegate to hsql) is not needed and not useful -- always leading to an unwanted exception.
MOST of the other switch cases were implementing identical behavior (in slightly different ways/code paths) -- made THAT the more reasonable default, applying also to CONCAT and likely MOST future cases.
Slightly reduced other dependencies on hsql super class -- attribute values that were hard-coded to be the same as VoltDB attributes that have VoltDB has tighter control over.

Had 1 minor fail in jenkins run -- seems unrelated (intermittent?).


wolffcm:
This looks good.  

The only danger I can see here is if a function has arguments the we *don't* want in output of getSQL, but this seems unlikely.  Is the getSQL codepath important for expression indexes mostly?  Do we have testing for indexes containing each VoltDB-specific function?  I guess I'm wondering if there's other latent bugs in this area.

If there's already good testing for expression indexes in this area, feel free to merge.  If not, consider adding it or filing a ticket.

PR #2099, opened 2015-03-03
ENG-7829: Fix lookupTuple() logic in the no PK path

afrogers:


src/ee/storage/persistenttable.cpp
```diff
@@ -869,14 +869,21 @@ TableTuple PersistentTable::lookupTuple(TableTuple tuple) {
...
             // to avoid matching duplicate tuples with different pointers to Object storage
             // -- which would cause erroneous releases of the wrong Object storage copy.
-            if (::memcmp(tableTuple.address(), tuple.address(), tuple_length) == 0) {
+            if ((forUndo || onlyInline)) {
+                char* tableTupleData = tableTuple.address() + TUPLE_HEADER_SIZE;
+                char* tupleData = tuple.address() + TUPLE_HEADER_SIZE;
+                if (::memcmp(tableTupleData, tupleData, tuple_length) == 0) {
+                    return tableTuple;
+                }
+            } else if (tableTuple.equalsNoSchemaCheck(tuple)) {
```
paulmartel: Pull the "if ((forUndo || onlyInline)) {" branching outside the loop and duplicate the hasNext/next calls inside each branch.
And then, since we're only calling it once, you could inline the call to m_schema->getUninlinedObjectColumnCount() == 0 so it got short-circuited out in the "forUndo" case.
And then you only have to initialize tuple_length before the loop that uses it.

BTW, nice catch on the off-by-one -- I thought I'd found the last of those -- I probably HAD but then added this as a new one.

tests/ee/storage/table_and_indexes_test.cpp
```diff
@@ -375,7 +375,7 @@ TEST_F(TableAndIndexTest, DrTest) {
     /*
      * Test that update propagates
      */
-    TableTuple toUpdate = districtTable->lookupTuple(temp_tuple);
+    TableTuple toUpdate = districtTable->lookupTuple(temp_tuple, false);
```
paulmartel: Seems like you'd want additional test cases to cover tables w/ no primary key, one all inline and one not.

src/ee/storage/persistenttable.cpp
```diff
@@ -860,7 +860,7 @@ void PersistentTable::deleteTupleForUndo(char* tupleData, bool skipLookup) {
     deleteTupleFinalize(target); // also frees object columns
 }
 
-TableTuple PersistentTable::lookupTuple(TableTuple tuple) {
+TableTuple PersistentTable::lookupTuple(TableTuple tuple, bool forUndo) {
```
paulmartel: To me, the forUndo false callers and the forUndo true callers clearly have different intent.
The fact that (largely) the same code satisfies them both is an implementation detail that I'd rather hide from the callers.
I'd suggest two new public inline member functions: lookupTupleByValues and lookupTupleForUndo each of which is a one-liner delegating to your new lookupTuple signature which can now be made private.

paulmartel:
End of review. In short, looks correct, could be cleaner, should be unit tested.

PR #2100, opened 2015-03-03
ENG-7484 Fix default to HTTP/JSON enabled

wijjo:
It's a little larger change set because I did some cleanup, e.g. un-hard-coded the port number in a couple of places. It should be easier to pick a different port number now by changing a constant.


tests/frontend/org/voltcore/utils/TestCOWNavigableSet.java
```diff
@@ -219,9 +219,9 @@ public void testAddDup() {
     public void testAddNonComparable() {
         try {
             COWNavigableSet q = new COWNavigableSet();
-            q.add(new Object());
```
wijjo: I'm not entirely sure why I needed to add the casts to make it build in Eclipse, but I don't think it affects the test to add them.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -61,6 +61,10 @@
     public static final int SITES_TO_HOST_DIVISOR = 100;
     public static final int MAX_SITES_PER_HOST = 128;
 
+    // Special port values to disable or trigger auto-scan.
+    public static final int PORT_DISABLED = -1;
+    public static final int PORT_AUTO = 0;
```
akhanzode: make these HTTP_ constants

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -1910,8 +1910,11 @@ else if (s.getIsorderdeterministic() == false) {
             }
             if (m_reportPath != null) {
                 outputStream.println("------------------------------------------\n");
-                outputStream.println("Full catalog report can be found at file://" + m_reportPath + "\n" +
-                            "\t or can be viewed at \"http://localhost:8080\" when the server is running.\n");
+                outputStream.println(String.format(
+                        "Full catalog report can be found at file://%s\n"
+                        + "\tor can be viewed at \"http://localhost:%d\" "
+                        + "when the server is running.\n",
+                        m_reportPath, VoltDB.DEFAULT_HTTP_PORT));
```
akhanzode: This is and will be a misleading message as 8080 could conflict and we should print at runtime what the URL will be.
VoltCompiler is useful in catalog mode only so maybe this message can be skipped?

tests/frontend/org/voltdb/TestJSONInterface.java
```diff
@@ -1235,8 +1242,8 @@ public void testGarbageProcs() throws Exception {
             server.start();
             server.waitForInitialization();
 
-            callProcOverJSONRaw("http://localhost:8080/api/1.0/Tim", 404);
-            callProcOverJSONRaw("http://localhost:8080/api/1.0/Tim?Procedure=foo&Parameters=[x4{]", 404);
+            callProcOverJSONRaw(getHTTPURL(null, "api/1.0/Tim"), 404);
```
akhanzode: Why did this need to be changed? this test server listens on 8095 and that should still hold valid right?

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -154,7 +158,7 @@ public synchronized static void setDefaultTimezone() {
         public String m_drInterface = "";
 
         /** HTTP port can't be set here, but eventually value will be reflected here */
-        public int m_httpPort = Integer.MAX_VALUE;
+        public int m_httpPort = PORT_DISABLED;
```
jhugg: Put new constants in org.voltdb.common.Constants

src/frontend/org/voltdb/compiler/VoltCompiler.java
jhugg: I think we can forgo it?

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -1010,13 +1020,18 @@ private String writeDeploymentFile(
...
+        if (m_httpdPortNo != null) {
+            deployment.setHttpd(httpd);
+            httpd.setEnabled(m_httpdPortNo != -1);
+            httpd.setPort(m_httpdPortNo);
+            if (m_jsonApiEnabled != null) {
+                Jsonapi json = factory.createHttpdTypeJsonapi();
+                httpd.setJsonapi(json);
+                json.setEnabled(m_jsonApiEnabled);
+            }
+        }
```
jhugg: I don't understand what was broken about this file? Is the new behavior any different? Why is NULL better than 0, especially when there's a constant. I might be missing something.

jhugg: I don't understand what was broken about this file? Is the new behavior any different? Why is NULL better than 0, especially when there's a constant? I might be missing something.

tests/frontend/org/voltcore/utils/TestCOWNavigableSet.java
jhugg: Weird. Errors? Not warnings?

But sure, looks harmless.

tests/frontend/org/voltdb/TestJSONInterface.java
```diff
@@ -1697,4 +1704,89 @@ public void testProfile() throws Exception {
             server = null;
         }
     }
+
+    public void testHTTPDefault() throws Exception {
```
jhugg: What do these next two tests test? Are they new code or moved from something else? Maybe a comment?

akhanzode:
Can you also manually test that --http option works there is no good test for it but CLI should override the defaults.

src/frontend/org/voltdb/VoltDB.java
wijjo: I actually had HTTP_ in front of the names originally and removed it because I thought these could/should apply to any port, whether or not the current code uses it that way. I don't feel strongly, if you do.

src/frontend/org/voltdb/VoltDB.java
wijjo: Sure.

src/frontend/org/voltdb/compiler/VoltCompiler.java
wijjo: Not sure I understand the issue. The message seen by the user is not changed by this, but if we ever change the default HTTP port it will remain correct.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
wijjo: Null is a distinct state that allows tests to fully omit either of the XML elements to test how it defaults.

tests/frontend/org/voltdb/TestJSONInterface.java
wijjo: It doesn't change the test since the url ends up being identical, but it will adapt if we change the default HTTP port.

tests/frontend/org/voltdb/TestJSONInterface.java
wijjo: Each test omits an XML element to test how the port and enabling defaults. The first omits the entire HTTP element and makes sure it can connect. The second omits the jsonapi element and tests the connection.

wijjo:
Sure, I'll make sure to manually test from the CLI. Thanks.

src/frontend/org/voltdb/VoltDB.java
jhugg: I agree with Anish. Make them HTTP and we can remove the thing if we find reuse later.

src/frontend/org/voltdb/compiler/VoltCompiler.java
jhugg: If the server is running HTTP on 8081, then this message is wrong. So is it worth keep a message that's often right, but sometimes wrong. I think it's probably best to remove at this point. Being wrong is bad.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
jhugg: Oh. I guess I would rather just test that manually, like via the deployment tests in TestVoltCompiler? It's confusing here because PORT_DISABLED, PORT_AUTO and PORT_DEFAULT already exist.

PR #2105, opened 2015-03-06
ENG-7850: Generate the binary log before the tuple is updated.

nshi:


afrogers:
Review complete. How did this survive so long? Were we just not updating things?

PR #2106, opened 2015-03-06
Eng 7480 volttype conversion

jxchinaren:
This change set removes guards of operation between DECIMAL and FLOAT either on Planner and EE engine, small change for scientific notation, and code clean up stuffs.

It does not make the SQL constants value DECIMAL as the default type, but keep the existing behavior.


src/ee/common/NValue.hpp
```diff
@@ -972,39 +976,16 @@ class NValue {
...
-        case VALUE_TYPE_TIMESTAMP:
-            return getTimestamp();
         default:
             throwCastSQLException(type, VALUE_TYPE_BIGINT);
             return 0; // NOT REACHED
         }
     }
 
+    /**
+     * Non-explicitly converting function to big integer type
```
wolffcm: Typo in comment here: you mean plain "integer" not "big integer".

Also, what does "non-explicitly" mean here?  Do you mean that this function is only called for implicit conversions (when there is no CastExpression)?

wolffcm: Also, since you are being good and documenting this functions behavior, what happens if the NValue doesn't fit into the range of a 32-bit integer?  Does it truncate, throw an exception or do something else?

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -73,6 +73,8 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Vector;
+import java.math.BigDecimal;
```
wolffcm: For the changes near the top of this file, we should probably follow the VoltDB annotation conventions for HSQL?

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1691,6 +1696,10 @@ VoltXMLElement voltGetXML(Session session, List<Expression> displayCols,
             }
 
             // Otherwise just string format the value.
+            if (dataType instanceof NumberType && ! dataType.isIntegralType()) {
+                exp.attributes.put("value", new BigDecimal(valueData.toString()).toPlainString());
```
wolffcm: Maybe add a comment here to indicate that you're doing this to remove the exponent notation from floating point toString()'s return value?  (That what I think you're doing here.)

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -44,6 +44,7 @@
 import org.hsqldb_voltpatches.result.Result;
 import org.hsqldb_voltpatches.result.ResultConstants;
 import org.hsqldb_voltpatches.result.ResultMetaData;
+import org.hsqldb_voltpatches.types.NumberType;
```
wolffcm: Paul's mentioned before that import blocks tend to have a lot of conflicts when upgrading HSQL.  If you're only using NumberType statically and in one place, consider just using the fully-qualified name instead of importing.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -1981,10 +1981,10 @@ public void testInWithStringParamsAdHoc() throws Exception {
         // VoltDB produces an error in this case, but this seems wrong:
         //
         verifyProcFails(client, "Array / Scalar parameter mismatch",
-                "@AdHoc", adHocQueryWithListParam, (Object)stringArgs);
+                "@AdHoc", adHocQueryWithListParam, stringArgs);
 
         verifyProcFails(client, "Array / Scalar parameter mismatch",
-                "@AdHoc", adHocQueryWithListParam, (Object[])stringArgs);
+                "@AdHoc", adHocQueryWithListParam, stringArgs);
```
wolffcm: Since these two calls (this one and the one before it) are now identical, you might remove one of them.  I had added the two calls since Java has some interesting rules (that I did not completely understand at the time) regarding vararg functions with array arguments.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -1306,7 +1307,11 @@ static protected void voltAppendParameters(Session session, VoltXMLElement paren
             parameter.attributes.put("index", String.valueOf(index));
             ++index;
             parameter.attributes.put("id", expr.getUniqueId(session));
-            parameter.attributes.put("valuetype", Types.getTypeName(paramType.typeCode));
+            if (paramType == NumberType.SQL_NUMERIC_DEFAULT_INT) {
+                parameter.attributes.put("valuetype", "BIGINT");
```
wolffcm: This seems like a reasonable change.  Was this an existing bug that your testing found?

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -2101,6 +2101,89 @@ public void testENG7724() throws Exception {
         assertEquals(13, vt.getColumnCount());
     }
 
+    private void runQueryGetDecimal(Client client, String sql, double value) throws Exception {
+        VoltTable vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertTrue(vt.advanceRow());
+        assertEquals(value, vt.getDecimalAsBigDecimal(0).doubleValue(), 0.0001);
```
wolffcm: I did not know that assertEquals for doubles provides an overload for the delta!  That's cool.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -2101,6 +2101,89 @@ public void testENG7724() throws Exception {
...
+
+        String sql;
+        sql = "insert into R1 Values(1, 'MA', 2, 2.2);";
+        client.callProcedure("@AdHoc", sql);
+        // query constants interpreted as DECIMAL
+
+        //
+        // operation between float and decimal
+        //
+        sql = "SELECT 0.1 + (1-0.1) + ratio FROM R1";
```
wolffcm: Help me out here a little bit.  What are the data types of operands here?  I'm not clear on why `0.1 + ratio` is not enough to reproduce the bug.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -2498,13 +2498,12 @@ public void testOverlay() throws NoConnectionsException, IOException, ProcCallEx
...
 
-        result = client.callProcedure("OVERLAY", "Jia", 4.2, 7, 1).getResults()[0];
-        assertTrue(result.advanceRow());
-        assertEquals("XinJia", result.getString(1));
-
-        result = client.callProcedure("OVERLAY", "Jia", 4.9, 7, 1).getResults()[0];
-        assertTrue(result.advanceRow());
-        assertEquals("XinJia", result.getString(1));
+        try {
+            result = client.callProcedure("OVERLAY", "Jia", 4.2, 7, 1).getResults()[0];
```
wolffcm: I guess we used to just truncate here?  Should we document this new behavior?

tests/frontend/org/voltdb/utils/TestVoltTypeUtil.java
```diff
@@ -113,25 +89,21 @@ public void testDetermineImplicitCastingNullWins()
...
-        assertEquals(VoltTypeUtil.determineImplicitCasting(VoltType.STRING,
-                                                           VoltType.STRING),
-                     VoltType.STRING);
+        assertEquals(VoltType.STRING,
+                VoltTypeUtil.determineImplicitCasting(VoltType.STRING,VoltType.STRING));
 
         // Check the easy non-coerced order
         VoltType[] winning_types = { VoltType.FLOAT,
-                                     VoltType.TIMESTAMP,
                                      VoltType.BIGINT };
```
wolffcm: I'm curious to know how timestamp ever worked in this loop.

tests/frontend/org/voltdb/utils/TestVoltTypeUtil.java
```diff
@@ -156,9 +127,16 @@ public void testDetermineImplicitCastingOrder()
...
             {
                 assertEquals(VoltType.BIGINT,
-                             VoltTypeUtil.determineImplicitCasting(promoted_types[i],
-                                                                   promoted_types[j]));
+                             VoltTypeUtil.determineImplicitCasting(promoted_types[i],promoted_types[j]));
             }
         }
+
+        assertEquals(VoltType.TIMESTAMP, VoltTypeUtil.determineImplicitCasting(VoltType.TIMESTAMP,VoltType.TIMESTAMP));
+
```
wolffcm: Wait, did you change the casting behavior of timestamp in this changeset?  Seems like operations between TIMESTAMP and other numeric types used to produce TIMESTAMP, and now we throw an exception?

wolffcm:
Okay, this looks *mostly* ready to go.  I've just got a question:
- What are the changes to changes to implicit casting for TIMESTAMP data type?  Or am I reading the test wrong?

src/ee/common/NValue.hpp
jxchinaren: Thanks, "Implicit" is the antonym of my word "non-explicitly". 

For out of range conversion, it will throw exceptions.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
jxchinaren: Yes, SQL Function OVERLAY hit this problem using SQL_NUMERIC_DEFAULT_INT but treated as FLOAT.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
jxchinaren: Hsql will pre-compute the operations (not sure whether it is always or not) and return DECIMAL type for certain cases. 

Anyway, I just do the same stuff to reverse the operation order as the other operations tests.

tests/frontend/org/voltdb/utils/TestVoltTypeUtil.java
jxchinaren: Not sure why we want to allow timestamp to operate with other Integer types here at first. I vaguely remembered that I have manually verified that we do return run time exception for operation between timestamp and integer on V5.0.

tests/frontend/org/voltdb/utils/TestVoltTypeUtil.java
jxchinaren: Yes. TIMESTAMP should not operate with INTEGER value directly without any time-related functions.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
jxchinaren: I will bring it up at the today's standup meeting.

wolffcm:
After our conversation at stand up these changes look good.  Feel free to merge when you are ready.

PR #2107, opened 2015-03-06
ENG-7851: internal unauth and unauthadmin clients needs to be whacked.

akhanzode:
ENG-7851: clear cache and dont bother with exceptions.


src/frontend/org/voltdb/client/AuthenticatedConnectionCache.java
```diff
@@ -241,10 +256,25 @@ public synchronized void releaseClient(Client client) {
...
+        } else {
+            conn.refCount--;
+        }
         attemptToShrinkPoolIfNeeded();
     }
 
+    private synchronized void closeClient(Client client)
+    {
+        try {
+            client.drain();
```
vtkstef: should wrap both statements in try blocks one per statement

PR #2110, opened 2015-03-09
ENG-7816: Fix DR replica command log recovery and chaining

nshi:


src/ee/storage/DRTupleStream.cpp
```diff
@@ -255,6 +255,22 @@ void DRTupleStream::endTransaction(int64_t sequenceNumber, int64_t uniqueId) {
          extendBufferChain(END_RECORD_SIZE);
      }
 
+     if (m_currBlock->startDRSequenceNumber() == std::numeric_limits<int64_t>::max()) {
+         throwFatalException(
+             "Appending end transaction message to a DR buffer with no matching begin transaction message."
+             " DR ID (%jd), unique ID (%jd)",
+             (intmax_t)sequenceNumber, (intmax_t)uniqueId);
```
afrogers: The internal numbers here are just sequence numbers, not DR IDs. Ambivalent about them showing up in exceptions, but it's a fatal that shouldn't occur, so I guess all debugging info anyway.

afrogers: Other than that terminology nitpick (DR ID has the cluster ID), I like the logging. Would corresponding logging be helpful on the begin as well?

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -225,14 +223,14 @@ public static synchronized void pushDRBuffer(
...
-                    System.out.println("Version " + version + " type TRUNCATE_TABLE table handle " + tableHandle + " table name " + tableName);
+                    log.trace("Version " + version + " type TRUNCATE_TABLE table handle " + tableHandle + " table name " + tableName);
                     break;
                 }
                 }
                 int calculatedChecksum = DBBPool.getBufferCRC32C(buf, startPosition, buf.position() - startPosition - 4);
                 if (calculatedChecksum != checksum) {
-                    System.out.println("Checksum " + calculatedChecksum + " didn't match " + checksum);
-                    System.exit(-1);
+                    log.error("Checksum " + calculatedChecksum + " didn't match " + checksum);
```
afrogers: This whole thing is a weird bit of code. It was private debugging info when we were implementing the buffers. I suppose turning it into trace is as good as anything. Good to see the System.exit(-1) calls go away, anyhow.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -158,65 +158,63 @@ public static synchronized void pushDRBuffer(
...
                     buf.position(buf.position() + lengthPrefix);
                     checksum = buf.getInt();
-                    System.out.println("Version " + version + " type DELETE table handle " + tableHandle + " length " + lengthPrefix + " checksum " + checksum);
+                    log.trace("Version " + version + " type DELETE table handle " + tableHandle + " length " + lengthPrefix + " checksum " + checksum);
                     break;
                 }
                 case UPDATE:
                     //Update
-                    //System.out.println("Version " + version + " type UPDATE " + checksum " + checksum);
+                    //log.trace("Version " + version + " type UPDATE " + checksum " + checksum);
```
afrogers: Probably no need to preserve commented tracing code for a record type we didn't implement. (Are we really ever going to implement it? Seems like we'd just stick with delete + insert)

src/ee/storage/DRTupleStream.cpp
nshi: I'll fix the terminology. You are right, these are just sequence numbers.

For the begins, what are you thinking that we can check for?

src/frontend/org/voltdb/PartitionDRGateway.java
nshi: I'll remove this.

src/ee/storage/DRTupleStream.cpp
afrogers: Hmm... yeah, guess there's not much we could do. Makes you want to track the last begin transaction we added for debugging purposes, but that's clunky. Would be nice to know if we're appending another begin transaction without ending the previous one, but wouldn't be nice to carry around extra, debug-only data.

src/frontend/org/voltdb/PartitionDRGateway.java
nshi: I just wanted a way to easily turn this debug code on and off without recompiling the whole jar, so I wrapped the whole block with a `isTraceEnabled()` check. Even though this message is at the ERROR level, it won't get printed if TRACE is not on. It's easier to grep for with ERROR.

afrogers:
Review complete

src/frontend/org/voltdb/PartitionDRGateway.java
afrogers: Oh, yes. Saw that. Pay no attention to where in this giant block of tracing code I inserted my comments -- the weirdness I speak of is the original tracing block itself, not your error level logging. More that we preserved it so long with a private boolean guard you had to programmatically set for yourself...

PR #2112, opened 2015-03-09
Eng 7858 real fix build client

rmorgenstein:
ENG-7858. Change build-client.xml to have strict list of included files.  
The build now makes a copy of the required sources before compiling them.   The 2nd commit shows which files were missing from the previous <include> list.


jhugg:
Thanks for doing this work. It looks correct, however depressing.

PR #2117, opened 2015-03-10
Eng 7629 daily logger bug fix

jxchinaren:


wolffcm:
Hi Xin, I'm having a hard time seeing how you've changed behavior here.  It looks like just a refactoring?  I'm sure I'm missing something.  Could you explain what the issue was and how this change solves it?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -965,56 +965,61 @@ public void initialize(VoltDB.Configuration config) {
...
+            } catch (NoSuchFieldException e) {
+                hostLog.error("Failed to set daily system info logging: " + e.getMessage());
+            }
+        }
+        final Field nextCheckField = field;
+        long nextCheck = System.currentTimeMillis();
+        // the next part may throw exception, current time is the default value
+        if (dailyRollingFileAppender != null && nextCheckField != null) {
+            try {
+                nextCheck = nextCheckField.getLong(dailyRollingFileAppender);
```
wolffcm: I guess the issue is that "dailyRollingFileAppender" was stale, because the underlying object will change when you call @UpdateLogging.  So a garbage value (0?) has being read for "nextCheck".  Your fix seems to be to loop over the appenders to get up-to-date DailyRollingFileAppender every time.

Could there ever be more than on DailyRollingFileAppender?

jxchinaren:
That means I did a very surgical fix and kept the original behavior as much as possible. The only change is in the RUN method, which it does not use the cached Appender class but update the reference first.

wolffcm:
Looks good.  Assuming you verified this by hand, feel free to merge.

PR #2119, opened 2015-03-10
ENG-7609 Cleaner DDL error messages

wijjo:
Chopped off the useless exception preamble from DDL error messages. Did not touch the other exception-wrapped messages in DDLCompiler. I did it by throwing VoltCompilerException instead of IOException for just the ordinary DDL errors and not wrapping the error when that is caught. I avoided risk by focusing tightly on the ticket requirements. Let me know if the scope should be expanded.


.gitignore
```diff
@@ -61,3 +61,4 @@ doc/tutorials/auction/csvloader_*.csv
 /src/ee/org_voltdb_utils_PosixAdvise.h
 /src/frontend/org/voltcore/utils/DBBPool.java
 /src/frontend/org/voltcore/utils/.flagfile
+/tests/sqlcmd/scripts/**/*.outdiffs
```
paulmartel: fyi, this will probably merge-conflict with a superset change about to be merged for 5.1.

tests/frontend/org/voltdb/TestAdhocCompilerErrorMessages.java
```diff
@@ -0,0 +1,63 @@
...
+            try {
+                m_client.callProcedure("@AdHoc", "drop table x y if exists;");
+            }
+            catch (ProcCallException pce) {
+                String message = pce.getLocalizedMessage();
+                Matcher m = pat.matcher(message);
+                assertTrue(String.format("'%s' mismatch: %s", pat.pattern(), message), m.matches());
+                threw = true;
+            }
+            assertTrue("Expected exception", threw);
```
paulmartel: Minor style suggestion: Use of "boolean threw" can be replaced by a one-liner:

    fail(""Expected exception");

at the bottom of the try block.

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -46,17 +46,24 @@ def mkdir_p(path):
 # kill the VoltDB server process, if any, running on the local machine.
 # The process is identified by its command line text via the ps command.
 def kill_voltdb():
+    # Works for Linux and Mac.
     # ps - find all running java commands.
-    # fgrep - filter out any headers or command lines that do not reference org.voltdb.VoltDB
-    # tr - compress multi-space delimiters to a single space.
-    # cut - extract the second space-delimited field, which is the PID.
+    # awk - finds commands ending in "/java" that refer to the org.voltdb.VoltDB class and outputs the PID(s)
```
paulmartel: Could use some mention of the rationale behind the awk script's use of magic numbers 8 and 9.

paulmartel:
This looks good. I see no need to widen scope before committing. There'll be other tickets.
AFAIK master is currently frozen except for 5.1-critical changes.
I could get behind merging this either in or after 5.1, but that's not my call.

PR #2122, opened 2015-03-11
ENG-7880: Fix unsafe use of temp tuple in the EE.

nshi:
Reset all the bits of temp tuple before use, or it may have unexpected
consequences for inline VARCHAR columns during byte comparison.

### DIFF
--- a/src/ee/storage/table.h
+++ b/src/ee/storage/table.h
@@ -145,6 +145,7 @@ class Table {
 
     TableTuple& tempTuple() {
         assert (m_tempTuple.m_data);
+        ::memset(m_tempTuple.address(), 0, m_tempTuple.tupleLength());
         return m_tempTuple;
     }
 
--- a/src/ee/storage/table.h
+++ b/src/ee/storage/table.h
@@ -145,6 +145,11 @@ class Table {
 
     TableTuple& tempTuple() {
         assert (m_tempTuple.m_data);
+        return m_tempTuple;
+    }
+
+    TableTuple& tempTupleWithZeroedBits() {
+        assert (m_tempTuple.m_data);
         ::memset(m_tempTuple.address(), 0, m_tempTuple.tupleLength());
         return m_tempTuple;
     }
--- a/src/ee/storage/BinaryLogSink.cpp
+++ b/src/ee/storage/BinaryLogSink.cpp
@@ -67,7 +67,7 @@ void BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_t, 
 
             PersistentTable *table = tableIter->second;
 
-            TableTuple tempTuple = table->tempTuple();
+            TableTuple tempTuple = table->tempTupleWithZeroedBits();
 
             ReferenceSerializeInputLE rowInput(rowData,  rowLength);
             tempTuple.deserializeFromDR(rowInput, pool);
--- a/tests/ee/storage/persistent_table_log_test.cpp
+++ b/tests/ee/storage/persistent_table_log_test.cpp
@@ -427,19 +427,19 @@ TEST_F(PersistentTableLogTest, LookupTupleUsingTempTupleTest) {
     m_table->insertTuple(nullTuple);
     delete[] nullTuple.address();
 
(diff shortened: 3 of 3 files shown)


src/ee/storage/table.h
```diff
@@ -145,6 +145,7 @@ class Table {
 
     TableTuple& tempTuple() {
         assert (m_tempTuple.m_data);
+        ::memset(m_tempTuple.address(), 0, m_tempTuple.tupleLength());
```
wolffcm: I've a couple concerns about this change.  

There are 19 callers of this method.  Are we sure that none of them intentionally make use of the existing behavior that leaves the bits as-is across calls to `tempTuple()`?  Provided that no callers are relying on the existing behavior, it might be good to rename the method to something that indicates the bits will be cleared.

My other concern is performance.  For wide tuples, the call to memset could be expensive.  For example, in projectionexecutor.cpp, we call `tempTuple()` for each input tuple.  I would suggest hoisting this call (and similar calls) out of loops wherever possible.

nshi: Thanks for the feedback, Chris.

I can make another flavor of this method that resets the bits and just use it in the DR code path, which will minimize the impact on existing code.

nshi: I pushed another commit that addresses your review comment. I'll rebase it after review. Don't merge it yet.

wolffcm:
This looks good to me with the latest commit.  Perhaps Paul should still have a look though.

I'm still concerned that this bug could manifest in other contexts, so I'll file a ticket to audit the other uses of tempTuple().

nshi:
Yes, That was my concern as well. Essentially any path that uses an uncleared temp tuple and calls `PersistentTable::lookupTuple()` could hit this.

wolffcm:
It occurs to me that zeroing out the tuple solves the problem for inlined varchar data, but for outlined varchar data, a `StringRef*` is stored in the tuple.  Depending on your context you might have two different StringRefs pointing to identical data, which make two logically identical tuples look different.  Could this be an issue here?

afrogers:
We solve this by checking if there are uninlined columns. If so, we do a
logical value compare instead of the raw ::memcmp.

On Wed, Mar 11, 2015 at 3:11 PM, Christopher M. Wolff <
notifications@github.com> wrote:

> It occurs to me that zeroing out the tuple solves the problem for inlined
> varchar data, but for outlined varchar data, a StringRef* is stored in
> the tuple. Depending on your context you might have two different
> StringRefs pointing to identical data, which make two logically identical
> tuples look different. Could this be an issue here?
>
> —
> <https://github.com/VoltDB/voltdb/pull/2122#issuecomment-78348512>.
>



-- 

*[image: VoltDBLogoEmail]* <http://www.voltdb.com/>
Alex Rogers, Software Engineer
*VoltDB, Inc.*
*E: *arogers@voltdb.com

Follow VoltDB [image: twitter] <https://twitter.com/VoltDB> [image:
facebook] <https://www.facebook.com/VoltDB> [image: linkedin]

<http://www.linkedin.com/company/voltdb>

nshi:
Paul pointed out to me that it's probably safer to have NValue always reset all the bits when setting a VARCHAR or VARBINARY column regardless of the length of the current value.

wolffcm:
I like Paul's idea, but it's unclear to me if it would impact performance.

nshi:
We talked about performance. The impact is probably negligible because it's only going to reset the bits if the column is inlined.

wolffcm:
That seems reasonable.

On Wed, Mar 11, 2015 at 12:39 PM, Ning Shi <notifications@github.com> wrote:

> We talked about performance. The impact is probably negligible because
> it's only going to reset the bits if the column is inlined.
>
> —
> <https://github.com/VoltDB/voltdb/pull/2122#issuecomment-78353765>.
>



-- 
Chris Wolff, Software Engineer
*VoltDB, Inc.*
*E:* cwolff@voltdb.com

*voltdb.com* <http://www.voltdb.com>
Follow VoltDB  <https://twitter.com/VoltDB>
<https://www.facebook.com/VoltDB>  <https://www.linkedin.com/company/voltdb>

PR #2123, opened 2015-03-11
ENG-7880: Fix unsafe use of temp tuple in the EE.

nshi:
Reset all the bits when setting an inline VARCHAR or VARBINARY, or it
may have unexpected consequences when used in an uncleared tuple during
byte comparison.

### DIFF
--- a/src/ee/common/NValue.hpp
+++ b/src/ee/common/NValue.hpp
@@ -1492,6 +1492,10 @@ class NValue {
      * inline object in the provided storage area
      */
     void inlineCopyObject(void *storage, int32_t maxLength, bool isInBytes) const {
+        // Always reset all the bits regardless of the actual length of the value
+        // 1 additional byte for the length prefix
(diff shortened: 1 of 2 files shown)


nshi:
Don't merge yet. I'll squash the two commits after review.

src/ee/common/NValue.hpp
```diff
@@ -1492,6 +1492,10 @@ class NValue {
      * inline object in the provided storage area
      */
     void inlineCopyObject(void *storage, int32_t maxLength, bool isInBytes) const {
+        // Always reset all the bits regardless of the actual length of the value
+        // 1 additional byte for the length prefix
+        ::memset(storage, 0, maxLength + 1);
+
```
paulmartel: In the non-null case, there's a check after this point and before the memcpy that can throw a sql exception.
I'd feel better if we didn't clear the storage until after we passed that check.

PR #2125, opened 2015-03-11
ENG-7886 Fix regression in sqlcmd LOAD CLASSES

paulmartel:
My recent fix to bypass the translation of LOAD/REMOVE commands 
into exec @UpdateClasses commands and rewrite them directly as 
m_client.callProcedure("@UpdateClasses" calls missed the fact that these 
particular exec calls never actually become m_client.callProcedure("@UpdateClasses"
calls at all -- they are special cased to become direct m_client.updateClasses calls.


wolffcm:
Ok, this looks like it does what it's supposed to.  The control flow for that method is more consistent and clear too, which is nice.  I've no objections to merging this.

I'm sure you've tested this by hand.  I'm guessing that automated tests will follow after we've branched 5.1?

PR #2128, opened 2015-03-12
one-line fix for sqlcmd save history

paulmartel:
Interactive sqlcmd was normally exiting via System.exit, bypassing the finally block that saved history between sessions.


wolffcm:
This looks good to me.  I was surprised to learn that calling exit bypasses finally blocks.

At some point it would be nice to get rid of all the calls to System.exit, in favor of throwing exceptions, which will ensure that finally blocks (those we have now and those we add in the future) are always executed.

Also, it would be good to have automated testing for this, at some point---seems like this kind of feature really wants "expect" style testing.

PR #2130, opened 2015-03-12
ENG-7569: CSVLoader ignoring --columnsizelimit

tkcampbell:
The option --columnsizelimit only operated on quoted strings, and it was only checking after encountering a newline.
The Tokenizer now checks column size within quoted columns all the time.


third_party/java/src/org/supercsv_voltpatches/tokenizer/Tokenizer.java
```diff
@@ -233,18 +232,18 @@ else if( c == delimeterChar ) {
...
-                        String
-                        .format(
-                                "oversized column while reading quoted column %d beginning on line %d and ending on line %d. " +
-                                "See --columnsizelimit.",
-                                columns.size() + 1, quoteScopeStartingLine, getLineNumber()));
-            }
+
+        if( currentColumn.length() > columnSizeLimit ) {
+            state = TokenizerState.NORMAL;
+            System.out.println("AWESOME");
```
akhanzode: remove AWESOME printing.

third_party/java/src/org/supercsv_voltpatches/tokenizer/Tokenizer.java
```diff
@@ -128,7 +128,6 @@ public boolean readColumns(final List<String> columns) throws IOException {
 	int charIndex = 0;
 	boolean espectQuote = this.strictQuotes;
 	boolean isEscape = false;
-	boolean sawNewLineInQuote = false;
```
akhanzode: Why did you need to remove this?

akhanzode:
Write a test in TestCSVLoader to test column size limit.

third_party/java/src/org/supercsv_voltpatches/tokenizer/Tokenizer.java
tkcampbell: Ah, oversight on my part. Was their for debugging. Removed.

third_party/java/src/org/supercsv_voltpatches/tokenizer/Tokenizer.java
tkcampbell: This variable was there only for checking if a newline had been encountered in a previous loop, so columnsize could be checked. I've changed it to check columnsize every time (not just after newlines), so this variable no longer serves any purpose.

third_party/java/src/org/supercsv_voltpatches/tokenizer/Tokenizer.java
```diff
@@ -299,6 +299,18 @@ else if( c == delimeterChar ) {
 			state = TokenizerState.NORMAL;
 			sawNewLineInQuote = false;
 			quoteScopeStartingLine = -1; // reset ready for next multi-line cell
+
+			// Check that we haven't gone over the column size limit
+			if( currentColumn.length() > columnSizeLimit ) {
```
akhanzode: Shouldn't this be a global check as opposed to within Quotes only? This is also not formatted correctly needs to indent in.

tkcampbell: According to the csvloader help text, columnsizelimit is the “max size of a quite column in bytes”, so I put the check in the quotes only section.

The formatting was like that originally. I kept it to me consistent. I can go back & change it.

PR #2133, opened 2015-03-13
Add table's delete statement to @SystemCatalog TABLES

wolffcm:
Also did some refactoring to pull out the DELETE statement for a table from the catalog.


paulmartel:
Looks good. Probably does not make the cut for 5.1, but probably also does not carry enough weight to force a branch, so I guess we should hold off just a bit on the merge until something else "breaks the ice" for 5.2.

PR #2135, opened 2015-03-13
ENG-6125 voltadmin stop command

wijjo:


lib/python/voltcli/voltadmin.d/stop.py
```diff
@@ -0,0 +1,125 @@
...
+    # Convert @SystemInformation results to objects.
+    hosts = Hosts(runner.abort)
+    for tuple in response.table(0).tuples():
+        hosts.update(tuple[0], tuple[1], tuple[2])
+
+    # Look up the hosts specified as command line arguments.
+    found, unused, missing = hosts.lookup(runner.opts.target_host)
+    if len(missing) > 0:
+        runner.abort('Host not found in cluster: %s' % ' '.join(missing))
+    if len(unused) == 0:
```
jhugg: I think there's already a server side error message for this. Do we want to have two error message or should we just try it anyway and pass the error message on from the server?

jhugg: Hmm... I guess in this case you have to pick a node to connect to... maybe you're right.

lib/python/voltcli/voltadmin.d/stop.py
```diff
@@ -0,0 +1,125 @@
...
+    hostname = unused[0].hostname
+    port = int(unused[0].adminport)
+    username = runner.opts.username
+    password = runner.opts.password
+    if username:
+        user_info = ', user: %s' % username
+    else:
+        user_info = ''
+    runner.info('Connecting to host: %s:%d%s' % (hostname, port, user_info))
+    runner.voltdb_connect(hostname, port, username, password)
```
jhugg: Are we disconnected at this point? Does this function smash and replace the current connection?

lib/python/voltcli/voltadmin.d/stop.py
wijjo: I was thinking of it more as argument validation.

lib/python/voltcli/voltadmin.d/stop.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Stop all the requested hosts using exec @StopNode HOST_ID
+    for host in found:
```
jhugg: I'm still not sure being able to kill more than one node at a time makes sense. Feels dangerous to me. I'm gonna ask Andrew G.

lib/python/voltcli/voltadmin.d/stop.py
wijjo: Yes, it disconnects and reconnects, possibly to the same node.

lib/python/voltcli/voltadmin.d/stop.py
wijjo: The argument validation only accepts one host, so found will only have one host for now. I didn't think it added much complexity to keep the support for many.

PR #2136, opened 2015-03-15
ENG-7902: Restartable @ApplyBinaryLogMP

nshi:


src/frontend/org/voltdb/VoltSystemProcedure.java
```diff
@@ -171,7 +172,7 @@ abstract public DependencyPair executePlanFragment(
         assert (txnState != null);
         txnState.setupProcedureResume(false, new int[] { aggregatorOutputDependencyId });
 
-        VoltTable[] results = new VoltTable[1];
+        final ArrayList<VoltTable> results = new ArrayList<>();
```
afrogers: I presume nothing relies on the behavior of always having one, potentially null, table?

src/frontend/org/voltdb/VoltSystemProcedure.java
```diff
@@ -181,14 +182,18 @@ abstract public DependencyPair executePlanFragment(
...
+                              "trusted.");
+                }
             } else {
-                results[0] = matchingTablesForId.get(0);
+                results.add(matchingTablesForId.get(0));
             }
         }
 
-        return results;
+        return results.toArray(new VoltTable[0]);
```
afrogers: In the only case where it would fit (empty ArrayList) and not just throw away that allocation, how is this different than .toArray() with no parameter? Both will allocate a new, size 0 array, but this call always allocates that array, usually unnecessarily.

src/frontend/org/voltdb/iv2/MpProcedureTask.java
```diff
@@ -121,7 +122,8 @@ public void run(SiteProcedureConnection siteConnection)
                 spName.startsWith("@") &&
                 !spName.startsWith("@AdHoc") &&
                 !spName.startsWith("@LoadMultipartitionTable") &&
-                !spName.equals("@UpdateApplicationCatalog"))
+                !spName.equals("@UpdateApplicationCatalog") &&
+                !spName.equals("@ApplyBinaryLogMP"))
```
afrogers: So your change just makes the ApplyBinaryLog itself fallible. Why not allow ApplyBinaryLogSP to roll back as well? Nothing about this seems specialized to the MP case.

src/frontend/org/voltdb/messaging/BorrowTaskMessage.java
```diff
@@ -65,7 +67,11 @@ public FragmentTaskMessage getFragmentTaskMessage()
 
     public void addInputDepMap(Map<Integer, List<VoltTable>> inputDeps)
     {
-        m_inputDeps = inputDeps;
+        final ImmutableMap.Builder<Integer, List<VoltTable>> builder = ImmutableMap.builder();
```
afrogers: Not that I in any sense object to this change, but how did it come about? Before this was just passing a mutable member map from the MpTransactionState?

afrogers: Never mind. Helpful when I read your commit messages.

afrogers:
Review complete.

src/frontend/org/voltdb/VoltSystemProcedure.java
nshi: The only case this would be null is described in the debug message I added below.

src/frontend/org/voltdb/VoltSystemProcedure.java
nshi: `toArray()` inconveniently gives you `Object[]`. The empty array is to tell `toArray()` that I want a `VoltTable[]` in return. If `results` actually has `VoltTable`s in it, it will allocate a new array of the given type.

src/frontend/org/voltdb/iv2/MpProcedureTask.java
nshi: You are right. Nothing prevents `@ApplyBinaryLogSP` from being able to rollback. In fact, it will if an exception is thrown inside the transaction.

This check here is specific to MP sysprocs. `@ApplyBinaryLogSP` never restarts, so no special handling is needed.

src/frontend/org/voltdb/messaging/BorrowTaskMessage.java
nshi: This is scary when I realized it was passing a mutable map to a totally different thread.

I guess the reason it never became a real issue is because for user MP procs, irrelevant fragment responses from before the restart are dropped on the floor, and the borrow task aggregation accepts empty dependency by accident. However, sysprocs don't tolerate empty dependencies.

PR #2138, opened 2015-03-16
ENG-7849: Fix a bogus check in BinaryLogSink.

nshi:
It is possible to get two begin txn entries with the same sequence
number in the binary log if there was a rollback. The check should make
sure that this is allowed unless there was real binary log data in the
previous transaction.


afrogers:
Review complete.

PR #2139, opened 2015-03-16
ENG-7907: Disallow add/drop/modify DR table columns.

nshi:


wijjo:
Is it okay that you're not doing anything about tables, just columns?

jhugg:
Sure.

PR #2140, opened 2015-03-16
Eng 5990 partial index

jxchinaren:
DO NOT MERGE!!! Code review ONLY!!!


tests/frontend/org/voltdb/regressionsuites/testpartialindexes-ddl.sql
```diff
@@ -0,0 +1,24 @@
...
+
+CREATE TABLE P1
+(
+  a bigint NOT NULL,
+  b bigint,
+  c bigint,
+  d bigint
+);
+PARTITION TABLE P1 ON COLUMN a;
+CREATE UNIQUE INDEX p1_pidx_1 ON P1 (a) where b is not null;
```
jxchinaren: It seems that there is missing tests:
1) Multiple columns index or expression index, other than ONLY single column index.
2) Partial index filter case on the same single column, like: 
CREATE UNIQUE INDEX p1_pidx_2 ON P1 (a) where a > 4;
CREATE UNIQUE INDEX p1_pidx_3 ON P1 (a) where a > b and c > 3;
...
3) The queries should cover more cases:
3.1) For index p1_pidx_2, it should cover query filter like "where a > 5", "where a > 4" and "where a > 3". Check out the plan also to verify the index usage of the query.
3.2) For index p1_pidx_3, the table data should have some rows satisfied with "a > b and c > 4" and some not. Try query filter variations on the where clause, check the results and the index usage information.

4)
There may be more necessary when it comes to parameterized queries. I do not think we have a dynamic query plan, not sure how should we handle this common case, like the procedure "where a > ?". For p1_pidx_2, different input parameters may result into different query plans.

src/ee/indexes/CompactingHashMultiMapIndex.h
```diff
@@ -72,13 +72,13 @@ class CompactingHashMultiMapIndex : public TableIndex
         return *reinterpret_cast<MapIterator*> (cursor.m_keyIter);
     }
 
-    bool addEntry(const TableTuple *tuple)
+    bool addEntryDo(const TableTuple *tuple)
```
jxchinaren: Why rename all the API to "xxxDo" ?

tests/frontend/org/voltdb/regressionsuites/testpartialindexes-ddl.sql
paulmartel: @jxchinaren I agree that these are good test case suggestions and the data ranges should span the values used in the index definition. WRT point 4 -- in the current limited version, we are focused on exact filter matches -- so an index where a > 4 applies to a query where a > 4 -- not even (for this release) a query where a > 5 (that can come a little later) -- and definitely not a query where a > ? (maybe MUCH later). We COULD/SHOULD be handling now the ad hoc query with a > 4 that got parameterized into a > ? -- for that to work, it just needs to use the existing "parameter binding" trick to put a "conditional" parameterized plan in the cache.
Then it becomes important to test ad hoc with a > 4 JUST BEFORE other ad hoc cases like a > 3 and a > 5 to make sure that we are purposely skipping the first cached plan.

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
```diff
@@ -337,6 +339,14 @@ public void testIndexedMinMaxViews() {
                "CREATE VIEW VT2 (V_D1_D2, V_D3, CNT, MIN_VAL1, SUM_VAL2, MAX_VAL3) " +
                "AS SELECT D1 + D2, ABS(D3), COUNT(*), MIN(VAL1), SUM(VAL2), MAX(VAL3) " +
                "FROM T " +
+               "GROUP BY D1 + D2, ABS(D3);" +
+               "CREATE VIEW VT3 (V_D1, V_D2, V_D3, CNT, MIN_VAL1_VAL2, MAX_ABS_VAL3) " +
```
jxchinaren: Not sure why adding two VIEWs here "VT3" and "VT4"? Have they been used somewhere else?

jxchinaren: Are the group by columns here a partial index? If yes, it DOES need more regression tests other than ONLY compiler tests.

If the source table has a partial index and matches the VIEW group by clause and where clause (Like the test case),  is there any way to tell the VIEW group by query really picks up the SOURCE table partial index?

tests/frontend/org/voltdb/regressionsuites/testpartialindexes-ddl.sql
jxchinaren: @paulmartel 
That means we should have negative tests to guard from the parameterized cases.

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +312,221 @@ public void testCaseWhenIndex()
         System.out.println(pn.toExplainPlanString());
         assertTrue(pn.toExplainPlanString().contains("using its primary key index (for deterministic order only)"));
     }
+
+    public void testPartialIndexNULLPredicate() {
+        {
+            AbstractPlanNode pn = compile("select * from c where a > 0;");
```
jxchinaren: Most of the check can be refactored into a function (String query, String expectedIndexName). Not sure how useful it is to check the INDEXSCAN plan node's comparison type.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -0,0 +1,389 @@
...
+                vt = client.callProcedure("@AdHoc", "select a, b, c from " + tb +" where c = 3 and b > 0 order by a,c;").getResults()[0];
+                validateTableOfLongs(vt, new long[][] { {6, 1, 3} });
+            }
+        }
+    }
+
+    public void testPartitionPartialUniqueIndex() throws Exception {
+        Client client = getClient();
+
+        if (!isHSQL()) { // HSQL doesn't support partial indexes
```
jxchinaren: It's better to do something like:

if (isHSQL()) return;

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -0,0 +1,389 @@
...
+                        " where a > 0 and b > 0 order by a").getResults()[0];
+                validateTableOfLongs(vt, new long[][] { {2,2} });
+            }
+        }
+    }
+
+    public void testPartialIndex() throws Exception {
+        Client client = getClient();
+
+        if (!isHSQL()) { // HSQL doesn't support partial indexes
```
jxchinaren: Same suggestion applies to all these cases.

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +312,221 @@ public void testCaseWhenIndex()
...
+            assertEquals(PlanNodeType.INDEXSCAN, pn.getPlanNodeType());
+            IndexScanPlanNode ipn = (IndexScanPlanNode) pn;
+            assertTrue(ipn.getPredicate() == null);
+            String json = pn.toJSONString();
+            assertTrue(json.contains("\"TARGET_INDEX_NAME\":\"A_PARTIAL_IDX_NOT_NULL_E\""));
+        }
+        {
+            // CREATE INDEX a_partial_idx_not_null_e ON c (a) where e is not null;
+            // range-scan covering from (A > 0) to end Z_FULL_IDX_A has higher cost
+            AbstractPlanNode pn = compile("select * from c where a > 0 and e = 0;");
```
jxchinaren: For this one test, predicate "e = 0" should be one of the post predicate for index scan. I think it's important to check that also.

jxchinaren:
Is that reasonable to have a demo benchmark showing that partial index is faster than the original index (either tree or hash)? That may be the best part of the work to show it really can work very well.

src/ee/indexes/CompactingHashMultiMapIndex.h
alexeevm: Initially addEntry/deleteEntry/etc. methods were declared pure virtual at the base class TableIndex and implemented by the concrete classes for each index type. I changed the pattern: All these methods are now non-virtual at the TableIndex class providing common functionality and the concrete work is delegated to the corresponding xxxxDo method which is pure virtual on the base class level and implemented at the child class level.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
alexeevm: Indeed. Corrected

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
alexeevm: corrected

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: The check is there: lines 352-353
            AbstractExpression p = ipn.getPredicate();
            assertEquals(ExpressionType.COMPARE_EQUAL, p.getExpressionType());
There is only one COMPARE_EQUAL expression (e=0) in the original SQL and I didn't check the TVEs to make sure one of them is "e"

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: Yes, it looks much better this way. Refactored. I think you meant the plan node predicate's comparison type, correct? The only reason I am doing it to make sure the expression is removed from the post-predicate in case it matches the index expression.

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
alexeevm: The purpose for the VT3 and Vt4 views is to verify  views based on a partial index.
The VT3 is supposed to use T_TREE_5 index and VT4 - T_TREE_6. The view predicate covers the partial index WHERE clause thus allowing the index to be selected: DDLCompiler.findBestMatchIndexForMatviewMinOrMax - lines 2205-2222

alexeevm: @jxchinaren I don't think I am following you with regard to the GROUP BY columns and the partial index. How does a partial index differ from a full one if the index columns cover all the group by ones?
There is actually a small problem in the view index selection logic - it picks the first index that satisfies the view definition. If a more efficient partial index happens to be not the first one it won't be even considered. Should we do a cost analysis there to pick the most efficient index?

alexeevm:
@jxchinaren Do you have any example of a benchmark?  I will use it as an example to build one for this project

tests/frontend/org/voltdb/regressionsuites/testpartialindexes-ddl.sql
alexeevm: @jxchinaren 1&2 The only reason the test cases here use the limited index types is that the index selection process (including multi-column and expression indexes and indexes with the same columns but different predicates) is tested in the planner suite (TestIndexSelection.java) where I can verify that the target index is actually selected. Here (the regression test suite) I concentrated on the EE side of the functionality. My thought was once the index is selected and access plan is built, the regression test should cover hash/tree/unique indexes. 
What is best way to verify which index is selected for an Ad-Hoc query? In my test runs I simply used debug statements to make sure the partial index is actually selected
3.1 As Paul already mentioned, this release only covers the exact match scenario.
3.2 i will add data to the test suite
@paulmartel @jxchinaren The planner itself does not apply a partial index to a parameterized query. I added a test case to the TestIndexSelection.java. The same, unfortunately, goes for the Ad-Hoc queries. Paul, what is this "parameter binding" trick you have mentioned?

src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -197,6 +197,17 @@ private boolean indexCovers(Index newIndex, Index existingIndex) {
...
+        if (existingIndex.getPredicatejson().length() > 0) {
+            if (existingIndex.getPredicatejson().equals(newIndex.getPredicatejson())) {
+                return true;
+            } else {
+                return false;
+            }
+        } else if (newIndex.getPredicatejson().length() > 0) {
+            return false;
+        }
+
```
jxchinaren: TODO: There should at list be a test case to cover INDEX duplicates and exercise this part of code on purpose. 

Place suggestion to add this test:
https://github.com/VoltDB/voltdb/blob/master/tests/frontend/org/voltdb/compiler/TestVoltCompiler.java#L1924

jxchinaren: Sorry, I see your test case now.

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
jxchinaren: @alexeevm I think the small problem is all right now. Even for the non-partial index case, currently VoltDB does not pick up the best index possible for MIN and MAX. I am fine right now.

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +316,211 @@ public void testCaseWhenIndex()
...
+            // CREATE INDEX partial_idx_5 ON c (b) where d > f;
+            AbstractPlanNode pn = compile("select * from c where b > 0 and f < d;");
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"PARTIAL_IDX_5\"");
+            String[] columns = {"F"};
+            checkIndexPredicateIsNull(pn);
+        }
+    }
+
+    public void testPartialIndexComparisonPredicateNonExactMatch()
+        {
```
jxchinaren: Indent problem?

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +316,211 @@ public void testCaseWhenIndex()
...
+            AbstractPlanNode pn = compile("select * from c where a > 0 and b > 0 and 0 < f;");
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"PARTIAL_IDX_4\"");
+            String[] columns = {"F"};
+            checkIndexPredicateDoesNotHave(pn, columns);
+        }
+        {
+            // CREATE INDEX partial_idx_5 ON c (b) where d > f;
+            AbstractPlanNode pn = compile("select * from c where b > 0 and d > f;");
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"PARTIAL_IDX_5\"");
+            String[] columns = {"F"};
```
jxchinaren: Never used "columns"?

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +316,211 @@ public void testCaseWhenIndex()
...
+        {
+        // At the moment expression exact match only
+        {
+            AbstractPlanNode pn = compile("select * from c where a > 0 and e > 0 or d < 5;");
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"Z_FULL_IDX_A\"");
+        }
+        {
+            AbstractPlanNode pn = compile("select * from c where b > 0 and e = 0;");
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"Z_FULL_IDX_A\"");
+//            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"PARTIAL_IDX_1\"");
```
jxchinaren: Not used, remove?

Also, could you add comments for this function testPartialIndexComparisonPredicateNonExactMatch() to explain why the index is picked? You sort of have them on the other tests.

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +316,211 @@ public void testCaseWhenIndex()
...
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"Z_FULL_IDX_A\"");
+//          checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"PARTIAL_IDX_3\"");
+        }
+    }
+
+    public void testParameterizedQueryPartialIndex() {
+        {
+            // CREATE INDEX partial_idx_4 ON c (a, b) where 0 < f; - not selected because of the parameter
+            AbstractPlanNode pn = compile("select * from c where a > 0 and b > 0 and ? < f;");
+            checkIndexName(pn, "\"TARGET_INDEX_NAME\":\"Z_FULL_IDX_A\"");
```
jxchinaren: Can you have a similar test on the expression index with parameter that can not use partial index?

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -310,4 +316,211 @@ public void testCaseWhenIndex()
         System.out.println(pn.toExplainPlanString());
         assertTrue(pn.toExplainPlanString().contains("using its primary key index (for deterministic order only)"));
     }
+
+    public void testPartialIndexNULLPredicate() {
+        {
```
jxchinaren: Indentation?

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
jxchinaren: Can you also verify that adding your VT3 or VT4 without adding partial indexes VoltDB will generate warnings?

I think it will work but a new test should not hurt. Thanks.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -212,6 +233,77 @@ public IndexableExpression extractEndFromPrefixLike() {
...
+            if (removeExactMatchCoveredExpressions(coveringExpr, exprsToCover)) {
+                exactMatchCoveringExprs.add(coveringExpr);
+            }
+            // Try the reversed expression for the exact match
+            if (reversedCoveringExpr != null && removeExactMatchCoveredExpressions(reversedCoveringExpr, exprsToCover)) {
+                // It is the original expression that we need to remember
+                exactMatchCoveringExprs.add(coveringExpr);
+            }
+            // @TODO commeneted out the non-exact match expressions
+//            // Now try to find non-exact match
```
jxchinaren: Are you planning to use the commented code? If not, could you clean it up?

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
alexeevm: @jxchinaren Added

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: Corrected

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: Yep

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: Removed commented out lines (invalid anyway) and added comments indicating the target index once the functionality is implemented

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: Yes, added

tests/frontend/org/voltdb/planner/TestIndexSelection.java
alexeevm: Corrected

src/frontend/org/voltdb/planner/SubPlanAssembler.java
alexeevm: Removed. I saved the code that deals with the non-exact match expressions on my local machine.

tests/frontend/org/voltdb/regressionsuites/testpartialindexes-ddl.sql
alexeevm: @jxchinaren I created few more indexes/test scenarios as you recommended.
How to check which index is used in the ad-hoc query? I currently use debug output for the verification but it has to be a better way of doing that.

src/ee/indexes/CompactingHashMultiMapIndex.h
jxchinaren: @alexeevm Thanks for the explanation.

tests/frontend/org/voltdb/regressionsuites/testpartialindexes-ddl.sql
jxchinaren: @alexeevm For the AdHoc query, there is a System procedure "Explain" to check the plan of the AdHoc query. I think I have an example here, please check: https://github.com/VoltDB/voltdb/blob/master/tests/frontend/org/voltdb/regressionsuites/TestGroupBySuite.java#L728

tests/frontend/org/voltdb/planner/TestIndexSelection.java
jxchinaren: @alexeevm Looks good.

PR #2142, opened 2015-03-16
ENG-7916 fix recall parser crash from failure to expect its argument

paulmartel:
Note to reviewer: expectFromAll test helper function just got moved unchanged -- real test change was addition of recall command parser tests.
Also, I manually  tested interactive sqlcmd.

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -364,6 +364,9 @@ public class SQLParser extends SQLPatternFactory
             "^\\s*" +         // optional indent at start of line
             "help" +          // required HELP command token
             "(\\W|$)" +       // require an end to the keyword OR EOL (group 1)
+            // Make everything that follows optional so that help
+            // command diagnostics can "own" any line starting with the word
+            // help.
             "\\s*" +          // optional whitespace before subcommand
             "([^;\\s]*)" +    // optional subcommand (group 2)
             InitiallyForgivingDirectiveTermination,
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -394,6 +397,8 @@ private Exception(Throwable cause, String message, Object... args)
             "(\\W|$)" +    // require an end to the keyword OR EOL (group 1)
             // Make everything that follows optional so that recall command
             // diagnostics can "own" any line starting with the word recall.
+            "\\s*" +          // extra spaces
+            "([^;\\s]*)" + // (first) non-space non-semicolon garbage word (group 2)
```
wolffcm: Some regex engines support named groups.  I am not sure if Java's does.  It seems like a nice feature to use here if it's available.

wolffcm:
Looks good.  Glad to see more testing here.

src/frontend/org/voltdb/parser/SQLParser.java
paulmartel: I'm pretty sure that Steve's structured regex builder leverages named groups.
I'd have to study the builder code a bit to know whether it can duplicate all the "forgiving" idioms I use here.
I suppose that switching to named groups in these regexs wouldn't completely be wasted effort even if we eventually migrated all of these to use the builders.

jxchinaren: Does the crash come from using the group 2 but sqlcmd RECALL pattern only used to have 1 group?

paulmartel: Almost -- the InitiallyForgivingDirectiveTermination string below adds two groups, so we were using group 4 when there were only 3 groups because group 2 was missing.
Chris is correct that if we used named groups in place of numbered groups, we might have caught or diagnosed this more easily. It is also true that this code error introduced by a refactoring that came late in the game could and should have been caught earlier with a trivial test in TestSQLParser.
Lesson for Paul: avoid refactoring anything that you can't immediately unit test.

PR #2143, opened 2015-03-16
ENG-7917 refresh sqlcmd metadata cache after a batch

paulmartel:
Tested by extending a sqlcmdtest script
 -- also added db schema cleanup steps to the test script.


wolffcm:
Looks good to me.

The outfiles are looking a lot easier to read with the echoing and "command succeeded" interleaved.

PR #2145, opened 2015-03-17
ENG-7918: Reverse the order of releasing the undo actions.

nshi:
Undo actions should be released in FIFO order, which is the reverse of
what undo does. Think about the case where you insert and delete a bunch
of tuples in a table, then does a truncate. You do not want to delete
that table before all the inserts and deletes are released.


jxchinaren:
Looks good to me. Probably we should log a new ticket to add more Junit tests for single transaction with multiple statements related to UPDATES/DELETES and TRUNCATE.

PR #2149, opened 2015-03-18
ENG-7922: Add callback to notify listener of ClientInterface partition

nshi:
master repairs.

### DIFF
--- a/src/frontend/org/voltdb/ConsumerDRGateway.java
+++ b/src/frontend/org/voltdb/ConsumerDRGateway.java
@@ -38,7 +38,7 @@ public interface ConsumerDRGateway extends Promotable {
 
     public abstract void shutdown(boolean blocking) throws InterruptedException;
 
-    public abstract void promotePartition(int partitionId, long maxDRId, long maxUniqueId);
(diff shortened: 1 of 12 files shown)


src/frontend/org/voltdb/ConsumerDRGateway.java
```diff
@@ -38,7 +38,7 @@
 
     public abstract void shutdown(boolean blocking) throws InterruptedException;
 
-    public abstract void promotePartition(int partitionId, long maxDRId, long maxUniqueId);
+    public abstract void recordLastSeenSegmentId(int partitionId, long maxDRId, long maxUniqueId);
```
afrogers: My only problem here is that "recordLastSeenSegmentId" vs. "notifyOfLastAppliedBinaryLog" makes this interface confusing, and that any comment explaining the difference would be rife with implementation details. Maybe better to specify purposes for these two methods. Maybe "recordLastSeenSegmentIdForPromote" vs "recordLastSeenSegmentIdFor[xxxxx]"; or maybe one method with a boolean flag

afrogers:
Review complete

PR #2151, opened 2015-03-19
Add a license expiring warning starting 30 days prior to expiration - co...

jpiekos:
...mmercial license only

### DIFF
--- a/src/frontend/org/voltdb/utils/MiscUtils.java
+++ b/src/frontend/org/voltdb/utils/MiscUtils.java
@@ -322,6 +322,23 @@ public class MiscUtils {
                                    expiresStr);
         consoleLog.info(msg);
 
+        // If this is a commercial license, and there is less than or equal to 30 days until expiration,
+        // issue a "days remaining" warning message.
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -322,6 +322,23 @@ else if (numberOfNodes > licenseApi.maxHostcount()) {
...
+        // There should thus be no issue with daylight savings time,
+        // but just in case, if the diff is a negative number, round up to zero.
+        if (diff < 0)
+        {
+            diff = 0;
+        }
+        long diffDays = diff / (24 * 60 * 60 * 1000);
+        if ((diff > 0) && (diff <= 30))
+        {
+            msg = "Warning, VoltDB commercial license expires in " + diffDays + " day(s).";
```
jhugg: It's more work, but I would love to have the message say the date it expires, then maybe the duration in parens after that?

Warning, VoltDB commercial license expires on March 31 (8 days).

jhugg: Also, this code won't get called for trial licenses. Is that a problem?

PR #2152, opened 2015-03-19
Regression fix for 5.1 ENG-7927 sqlcmd exec with quoted space parameters

paulmartel:


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -344,7 +344,7 @@ private Exception(Throwable cause, String message, Object... args)
             "'" +
             "[^',\\s]*" +  // arbitrary string content NOT matching param separators
             "[,\\s]" +     // the first match for a param separator
-            "[^']" +       // arbitrary string content
+            "[^']*" +      // arbitrary string content
```
wolffcm: This seems like it won't quite work for escaped single quotes (`''`) occurring after a parameter separator in a string.  I don't think you need to fix it here, but it would be good test.

wolffcm:
Looks good to me.

src/frontend/org/voltdb/parser/SQLParser.java
paulmartel: I'll add at least parser tests -- I purposely chose NOT to try to handle escaped single quotes here, but instead to match 'abc''def' as TWO strings. With no intervening space, they get split into the same parameter -- each parameter IS purposely allowed to have multiple serial sustitutions applied and the substitutions preserve their end quotes, so the quoted quote also gets preserved as is. I can't vouch for what happens post-parse, but I expect at least a parser test to succeed -- a sqlcmdtest may not. Easy to try.

paulmartel:
Added a more significant commit to fix multiple string params on a line.
And quoted quotes have a working test now.

wolffcm:
Ok, this looks good to me.  And the tests look good too.

One other question: do we handle embedded newlines in string parameters?  I'm just curious what we do for that.  It seems like it might work, looking at the code.  (Even if we don't do the right thing, I don't think it needs to be fixed in 5.1.)

paulmartel:
What I've seen in the past -- and as far as I know I haven't broken it -- is that a newline in a string gets replaced "somewhere down the line" with a space.
An edge case whose behavior probably changed recently is that a semicolon within a string that is the last non-whitespace character on its line will be treated as a premature end of statement.
IF that ever worked in the past, I'd argue that making it work cost us too much in other capabilities or edge case behaviors.

PR #2153, opened 2015-03-19
ENG-7929: Fix DR tuple stream serialization error.

nshi:
Wrong null array position, the header size already includes the null
array size.


afrogers:
Reviewed. Looks fine to me.

PR #2154, opened 2015-03-19
ENG-7931: Disallow exporting a DR table.

nshi:


afrogers:
We already handled other paths for this, right? If so, looks good

PR #2155, opened 2015-03-19
ENG-7929: Fix DR tuple stream serialization.

nshi:
Forgot to add the current IO position to the null array location.


afrogers:
As I said in email, reviewed. Although it seems like (at least the reserve) could use a comment

nshi:
Thanks, Alex. Comment added. Merging now.

PR #2156, opened 2015-03-20
Eng 7856: export connecter pushes performance stats to a socket

tkcampbell:


tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+            buffer.clear();
+            buffer.put((byte)message.length());
+            buffer.put(message.getBytes());
+            buffer.flip();
+
+            // Send message over socket
+            try {
+                channel.send(buffer, address);
+            } catch (IOException e) {
+                System.err.println("Couldn't send stats to socket");
```
akhanzode: Use export logger see some of the export connectors such as kafka.

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+                System.err.println(e.getLocalizedMessage());
+            }
+        }
+
+        @Override
+        public void sourceNoLongerAdvertised(AdvertisedDataSource source) {
+            timer.cancel();
```
akhanzode: You need to close socket resources too.

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+            address = new InetSocketAddress(InetAddress.getLocalHost(), port);
+        } else {
+            address = new InetSocketAddress(host, port);
+        }
+        channel = DatagramChannel.open();
+    }
+
+    class SocketExportDecoder extends ExportDecoderBase {
+        Timer timer;
+        ByteBuffer buffer = ByteBuffer.allocate(1024);
```
akhanzode: you can move this into print statistics. Also call printStatistics as sendStatistics as thats what you are doing.

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+        /**
+         * Periodically print statistics to a UDP socket
+         */
+        public void schedulePeriodicStats() {
+            timer = new Timer();
+            TimerTask statsPrinting = new TimerTask() {
+                @Override
+                public void run() { printStatistics(); }
+            };
+            timer.scheduleAtFixedRate(statsPrinting,
```
akhanzode: make this a configuration property of export connector so that we can tweak it if needed,

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+
+            // Time decodeRow
+            try {
+                long startTime = System.nanoTime();
+                decodeRow(rowData);
+                long endTime = System.nanoTime();
+
+                totalDecodeTime += (endTime - startTime);
+            } catch (IOException e) {
+                System.err.println(e.getLocalizedMessage());
```
akhanzode: user logger

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+                           + "\n";
+
+            buffer.clear();
+            buffer.put((byte)message.length());
+            buffer.put(message.getBytes());
+            buffer.flip();
+
+            // Send message over socket
+            try {
+                channel.send(buffer, address);
```
akhanzode: Check how many bytes sent. If less that capacity we need to send rest?

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
```diff
@@ -0,0 +1,177 @@
...
+            // Transaction count
+            transactions++;
+
+            // Time decodeRow
+            try {
+                long startTime = System.nanoTime();
+                decodeRow(rowData);
+                long endTime = System.nanoTime();
+
+                totalDecodeTime += (endTime - startTime);
```
akhanzode: You can calculate tps and avg decode time here and save it in AtomicLong values so print stats will not send wrong values. Currently the processRow and printStatstics will bump into each other and could send wrong stats.

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
akhanzode: You need to throw RestartBlockException which means this row will be retried by export. By catching and not rethrowing the export data is lost.

tkcampbell: Is this necessary though? This error relates to statistics printing, not to the processing of the data to be exported.

tests/test_apps/exportbenchmark/src/exportbenchmark/SocketExporter.java
tkcampbell: According to the docs, the send() method returns "either the number of bytes that were remaining in the source buffer when this method was invoked or, if this channel is non-blocking, may be zero if there was insufficient room for the datagram in the underlying output buffer"

As far as I can tell, it shouldn't ever be anything but the bytes sent, but I added an error message if it isn't for now.

PR #2158, opened 2015-03-20
Fix voltdb client parser for replica start action.

nshi:


lib/python/voltcli/verbs.py
```diff
@@ -453,15 +453,16 @@ def initialize(self, verb):
...
             if runner.opts.replica:
-                self.subcommand = 'replica'
+                final_args = [self.subcommand, 'replica']
         if self.supports_live:
             if runner.opts.block:
                 final_args = [self.subcommand]
             else:
                 final_args = ['live', self.subcommand]
-        else:
+        elif final_args == None:
```
wijjo: "is None" is preferable to "== None"

PR #2164, opened 2015-03-23
Tweaks to prevent unwarranted throws/crashes.

paulmartel:
In extractBuildInfo, use early return and fall-through instead of a RuntimeException throw+catch to implement normal flow control.

In "RestoreAgent.generatePlans", just fudge an error message in the unlikely (maybe impossible) event of a JSON exception during its formatting. This is preferable to leaking the exception to the caller. The caller is set up to call crashGlobalVoltDB on any exception.


jxchinaren:
Looks good to me.

PR #2166, opened 2015-03-24
ENG-7936 (Union on partitioned selects does not work)

atratus:
Problem: m_partitioning reference is replaced by new one so client of PlanAssember does not know about gathered  partitioning information.
Solution: set data in existent m_partitioning instance client logic refers to as well.


afrogers:
Thanks for your work on this. At this time, I’m going to close this pull request, since we’ve merged a fix for this issue from one of our engineers, which can be found here:
https://github.com/VoltDB/voltdb/pull/2168/files

For future reference, in general we won't merge a fix without unit tests. You can contact us for advice on where the relevant test suites are located.

PR #2167, opened 2015-03-24
Guice & Maven proof-of-concept

atratus:
This pull request introduces Maven build scripts as initial step in the way of ant-maven migration. (Ant build still works).
This pull request introduces Spring as IOC container. For now, Configuration class is moved out of VoltDB class and is served by Spring as a singleton.
Both works (Maven migration, Spring migration) imply deep structural modifications in future so probably it is better to continue work on them in main development branch. In other way seems it will be very hard to reintegrate them back.


atratus:
Will create able-to-merge 'guice' pull request instead

PR #2168, opened 2015-03-24
Eng 7936 union wrong answer fix

jxchinaren:


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -219,6 +219,9 @@ synchronized AdHocPlannedStatement planSql(String sqlIn, StatementPartitioning p
                 // If not caching or there was no cache hit, do the expensive full planning.
                 plan = planner.plan();
                 assert(plan != null);
+                if (plan != null && plan.getStatementPartitioning() != null) {
```
wolffcm: I guess you don't need to check if plan is null here since the previous line asserts it.

tests/frontend/org/voltdb/regressionsuites/TestUnionSuite.java
```diff
@@ -43,30 +43,53 @@ public TestUnionSuite(String name) {
...
+              " UNION " +
+              "select client_id, config_id, sum(cost) as cost from rpt_copy_p where client_id=140 group by client_id, config_id;";
+        result = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertEquals(3, result.getRowCount());
+
+        result = client.callProcedure("testunion_p", 140, 140).getResults()[0];
+        assertEquals(3, result.getRowCount());
+
+        result = client.callProcedure("testunion_p", 10, 10).getResults()[0];
+        assertEquals(0, result.getRowCount());
```
wolffcm: How much UNION verification do we do beyond just asserting the right number of rows?  It seems like we would also want to verify the contents of the result tables.

wolffcm:
I'm not clear on why we were getting a wrong answer here, so I'm not understanding why these changes fix the bug.

src/frontend/org/voltdb/compiler/PlannerTool.java
jxchinaren: "partitioning" got changed inside of the planning, but the code cached this variable which contained outdated information contributing wrong answer in later use.

tests/frontend/org/voltdb/regressionsuites/TestUnionSuite.java
jxchinaren: In order to make the results order deterministic, that's where I found the wrong answer problem on VoltDB with ORDER BY clause. And without ORDER BY, VoltDB has different order results with Hsqldb. 

This Union wrong answer bug used to return 0 rows for all cases because of losing the partitioning parameter information.

src/frontend/org/voltdb/compiler/PlannerTool.java
wolffcm: Ok, I think I'm understanding.  Normally it's the same instance of the Partitioning object.  But for UNION, we create multiple instances?  Looking at planner code, I see this is the case.  Okay, looks good.

tests/frontend/org/voltdb/regressionsuites/TestUnionSuite.java
wolffcm: Ok.  My question is, do we lack UNION testing?  Should we file a ticket to add more?

wolffcm:
Looks good to me.

tests/frontend/org/voltdb/regressionsuites/TestUnionSuite.java
jxchinaren: Add a new Union testing ticket ENG-7966, linked with the wrong answer ORDER BY ticket also.

paulmartel: IF we had variants (optional arguments?) to the RegressionSuite validate... helper functions that could handle non-deterministic ordering, that MIGHT be enough to motivate lazy test developers to "raise their expectations" from row counts to actual values.
These variants could use a common helper function to sort a "two dimensional table of longs" by its columns left-to-right before comparing to expected.

PR #2178, opened 2015-03-26
ENG-7753 fixed check for old topology

dweiss88:


src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2460,6 +2460,7 @@ public void run() {
                     if (ds != null) {
                         oldValue = ByteBuffer.allocate(ds.getSerializedSize());
                         ds.serialize(oldValue);
+                        oldValue.flip();
```
akhanzode: Add a comment here why so we remember.

akhanzode:
squash merged.

PR #2179, opened 2015-03-26
Add JSON string to Result column of Table statistics will be included wh...

wweiss-voltdb:
ENG-7969: Add JSON string to Result column of Table statistics will be included whenever DR is enabled for that table

### DIFF
--- a/tests/frontend/org/voltdb/TestAdhocCreateTable.java
+++ b/tests/frontend/org/voltdb/TestAdhocCreateTable.java
@@ -196,4 +196,52 @@ public class TestAdhocCreateTable extends AdhocDDLTestBase {
             teardownSystem();
         }
     }
+
+    public void testCreateDRTable() throws Exception {
+
+        String pathToCatalog = Configuration.getPathToCatalogForTest("adhocddl.jar");
+        String pathToDeployment = Configuration.getPathToCatalogForTest("adhocddl.xml");
+
+        VoltProjectBuilder builder = new VoltProjectBuilder();
+        builder.addLiteralSchema("--dont care");
+        builder.setUseDDLSchema(true);
+        boolean success = builder.compile(pathToCatalog, 2, 1, 0);
+        assertTrue("Schema compilation failed", success);
+        MiscUtils.copyFile(builder.getPathToDeployment(), pathToDeployment);
+
+        VoltDB.Configuration config = new VoltDB.Configuration();
+        config.m_pathToCatalog = pathToCatalog;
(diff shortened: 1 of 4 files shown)


tests/frontend/org/voltdb/TestAdhocCreateTable.java
```diff
@@ -196,4 +196,52 @@ public void testCreatePartitionedTable() throws Exception {
             teardownSystem();
         }
     }
+
+    public void testCreateDRTable() throws Exception {
```
nshi: This is a rather random place to put a test for `@SystemCatalog`. There is actually a `TestSystemCatalogSuite` which may suite this better.

nshi:
Mind adding the ticket number to the commit message so that Jira can link it to the ticket?

tests/frontend/org/voltdb/TestAdhocCreateTable.java
wweiss-voltdb: AdhocDDLTestBase is used in many tests to interpret SystemCatalog output. The function I added there is a convenience method for determinining if a Table is DRed that can be used by other tests. I chose Create Table because there is already a similar test in there that verifies the catalog values of an adhoc create table followed by a partition table.

nshi: I understand that it's convenient to add the test here because there is another test doing similar stuff. But it's unintuitive for what your test is trying to test. For example, if I changed the JSON string for DR tables in the @SystemCatalog result, I would probably go to TestSystemCatalogSuite and make the same changes there, but would never have thought of TestAdhocCreateTable. In the end, what you are trying to test is the result of @SystemCatalog, not creating a table.

wweiss-voltdb: If that were the case, then the other JSON strings under remarks (limitPartitionRowsDeleteStmt and sourceTable) would be tested in TestSystemCatalogSuite, but they are not.

nshi: > If that were the case, then the other JSON strings under remarks (limitPartitionRowsDeleteStmt and sourceTable) would be tested in TestSystemCatalogSuite, but they are not.

Then I guess this is the opportunity to set a good example.

wweiss-voltdb: I realized that the TestSystemCatalogSuite does touch the JSON so I added one case with the new string.

PR #2186, opened 2015-03-30
ENG-7910: add sha256

akhanzode:
Add sha256 hashing. make it default for java client. wire protocol change is backward compatible.


src/catgen/spec.txt
```diff
@@ -95,6 +95,7 @@ end
 begin User javaonly
   GroupRef* groups
   string shadowPassword "SHA-1 double hashed hex encoded version of the password"
+  string sha256ShadowPassword "SHA-256 double hashed hex encoded version of the password"
```
jhugg: Why would we just list all kinds of passwords? Wouldn't we list the bytes for the hash and then what kind of hash it is?
What if we need to add another kind of hash? There's lots of places in the code where this should be abstracted out instead of a giant OR. That's my main feedback on this path.

akhanzode: I can push auth hash scheme in catalog.

akhanzode:
squash merged.

PR #2187, opened 2015-03-30
Eng 7651 add dr to full ddl

wweiss-voltdb:


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -968,6 +975,9 @@ private void fillTrackerFromXML()
                 else {
                     m_tracker.removeExportedTable(tableName);
                 }
+                if (drTable != null) {
```
afrogers: All the others remove if it is null -- why not DR, too?

afrogers:
Review complete

src/frontend/org/voltdb/compiler/DDLCompiler.java
wweiss-voltdb: Let me double check the behaviour.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -778,16 +777,23 @@ else if (matchStatus == ClassNameMatchStatus.NO_WILDCARD_MATCH) {
...
-                "Invalid DR statement: table %s is an export table", tableName
-                ));
+            if (tableXML != null) {
+                if (tableXML.attributes.containsKey("export")) {
+                    throw m_compiler.new VoltCompilerException(String.format(
+                        "Invalid DR statement: table %s is an export table", tableName));
+                }
+                else {
+                    if ((statementMatcher.group(2) != null)) {
+                        tableXML.attributes.put("drTable", "DISABLE");
```
nshi: What about just dropping the `drTable` attribute if it's disabled?

nshi:
There is a check in `CatalogDiffEngine.java` that prevents DR table to be exported. I believe that can be removed now with the proper DDL compilation.

nshi:
Except for the first commit, the rest has nothing to do with ENG-7651. Can you edit the commit messages so that they describe what ticket they are fixing? The first commit should contain ENG-7651. Squashing the last three commits would also be appreciated.

src/frontend/org/voltdb/compiler/DDLCompiler.java
wweiss-voltdb: It is a style question. The disabled drTable flag translates to the same thing as a missing drTable flag. So if the behaviour is identical, efficiency breaks the tie for me. It is more efficient to do a put than a check followed by a delete.

PR #2194, opened 2015-04-01
ENG-7975: Allow DR to use default to External Interface configuration wh...

wweiss-voltdb:
...en DR interface is not specified in the command line. Ensure proper precedence for DR ports.

### DIFF
--- a/src/frontend/org/voltdb/VoltDB.java
+++ b/src/frontend/org/voltdb/VoltDB.java
@@ -151,7 +151,7 @@ public class VoltDB {
         public String m_internalInterface = DEFAULT_INTERNAL_INTERFACE;
 
         /** port number to use for DR channel (override in the deployment file) */
-        public int m_drAgentPortStart = -1;
+        public int m_drAgentPortStart = DEFAULT_DR_PORT;
         public String m_drInterface = "";
 
         /** HTTP port can't be set here, but eventually value will be reflected here */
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -348,7 +348,7 @@
     </xs:sequence>
     <xs:attribute name="id" type="clusterIdType" use="required"/>
     <xs:attribute name="listen" type="xs:boolean" default="false" />
-    <xs:attribute name="port" type="xs:int" default="5555"/>
+    <xs:attribute name="port" type="xs:int" default="0"/>
```
nshi: The default was specified as 5555 so that VMC can display the proper value on the Admin panel. Changing it to 0 will confuse users.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -151,7 +151,7 @@ public synchronized static void setDefaultTimezone() {
         public String m_internalInterface = DEFAULT_INTERNAL_INTERFACE;
 
         /** port number to use for DR channel (override in the deployment file) */
-        public int m_drAgentPortStart = -1;
+        public int m_drAgentPortStart = DEFAULT_DR_PORT;
```
nshi: I think this was initialized as -1 so that you can tell if user specified something on the command line.

PR #2196, opened 2015-04-03
ENG-7975: commmand line replication port always chosen over deployment f...

wweiss-voltdb:
...ile.


src/frontend/org/voltdb/Inits.java
```diff
@@ -572,6 +571,7 @@ public void run() {
             m_config.m_drAgentPortStart = replicationPort;
         }
     }
+    */
```
nshi: Is this commented out by accident or is it unused now? If not used, we should delete this block.

wweiss-voltdb: Sorry. I tested this to make sure it worked but forgot to remove it afterward.

PR #2197, opened 2015-04-06
Remove initFromTupleStorage bottleneck

wolffcm:


src/ee/executors/OptimizedProjector.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /** Perform the projection on a destination tuple. */
+    void exec(TableTuple& dstTuple, const TableTuple& srcTuple) const;
+
+    /** For testing, re-order the target fields so mem copies must be broken up. */
+    void permuteOnIndexBit(int numBits, int bitToFlip);
+
+    /** The number of steps needed to perform this projection */
+    size_t numSteps() const;
+
+    /** For testing, return an exppression for each step.  (This must be done before optimizing,
```
paulmartel: spelling: exppression

Also, I prefer something more explicit like a ..ForTest name suffix to indicate test-only functions.

tests/ee/harness.h
```diff
@@ -247,6 +247,14 @@ do { \
...
 } while (0)
 
+#define ASSERT_TRUE_MSG(value, msg)                     \
+do { \
+    if (!(value)) { \
+        STUPIDUNIT_ASSERT_BREAKPOINT_CODE \
+        fail(__FILE__, __LINE__, msg); \
+        return; \
+    } \
+} while (0)
```
paulmartel: Consider refactoring:

    #define ASSERT_TRUE(value) ASSERT_TRUE_MSG(value, "Expected true; " #value " is false")

    #define ASSERT_FALSE(value) ASSERT_TRUE_MSG(!(value), "Expected false; " #value " is true")

tests/ee/harness.cpp
```diff
@@ -158,13 +159,19 @@ int TestSuite::runAll() {
...
+                printf("PASSED.\n");
+            } else {
+                printf("FAILED.\n");
+                test->printErrors();
+                printf("\n");
+                failed_tests++;
+            }
+        }
+        catch (voltdb::SerializableEEException& eeExc) {
+            printf("Uncaught SerializableEEException: %s\n", eeExc.message().c_str());
```
paulmartel: Are we potentially squelching forensic details by not running

            test->printErrors();
            printf("\n");

in the caught exception code path?

tests/ee/executors/OptimizedProjectorTest.cpp
```diff
@@ -0,0 +1,480 @@
...
+int main(int argc, char* argv[]) {
+    int opt;
+    while ((opt = getopt(argc, argv, "r:c:")) != -1) {
+        switch (opt) {
+        case 'r':
+            NUM_ROWS = boost::lexical_cast<int64_t>(optarg);
+            break;
+        case 'c': {
+            unsigned long val = boost::lexical_cast<unsigned long>(optarg);
+            // number of column must be a power of two.
```
paulmartel: Might want to be a little more explicit about this power of 2 thing, like print to stderr if violated even if asserts are optimized out, and add to usage like '[-c <num_cols_must_be_power_of_2>]'

tests/ee/executors/OptimizedProjectorTest.cpp
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    assert (ExecutorContext::getExecutorContext() == NULL);
+
+    boost::scoped_ptr<Pool> testPool(new Pool());
+    UndoQuantum* wantNoQuantum = NULL;
+    Topend* topless = NULL;
+    boost::scoped_ptr<ExecutorContext> executorContext(new ExecutorContext(0, 0, wantNoQuantum, topless, testPool.get(),
+                                                                           NULL, "", 0, NULL, NULL));
+
```
paulmartel: Makes sense I guess to stash away this setup code here vs. in the test class where I've usually been putting it -- if the values don't need to be referenced in the test. I had not thought of that.

tests/ee/executors/OptimizedProjectorTest.cpp
```diff
@@ -0,0 +1,480 @@
...
+    std::vector<TypeAndSize> outlinedVarcharColumns;
+    for (int i = 0; i < NUM_COLS; ++i) {
+        outlinedVarcharColumns.push_back(VARCHAR_OUTLINE);
+    }
+
+    cout << "\n          " << "VARCHAR columns (outlined):\n";
+    runProjectionTest(outlinedVarcharColumns, outlinedVarcharColumns, projector);
+    BOOST_FOREACH(AbstractExpression* e, exprs) {
+        delete e;
+    }
```
paulmartel: I forget: is there not a boost substitute for vector<ptr> that would just take care of this cleanup?

tests/ee/executors/OptimizedProjectorTest.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+    template <typename T>
+    T log2(T n) {
+        return static_cast<T>(log(static_cast<double>(n)) / log(2.0));
+    }
+
+    void runProjectionTest(const std::vector<TypeAndSize>& srcTableTypes,
+                           const std::vector<TypeAndSize>& dstTableTypes,
+                           const OptimizedProjector& baselineProjector) {
```
paulmartel: srcTableTypes and dstTableTypes are redundant in the current use cases.

src/ee/executors/OptimizedProjector.cpp
```diff
@@ -0,0 +1,528 @@
...
+        } memcpyParams;
+
+        struct Eval {
+            AbstractExpression* expr;
+            int dstFieldIndex;
+        } evalParams;
+
+    };
+
+    Params m_params;
```
paulmartel: Readability would be helped by consistency, here -- Action and m_action are defined in one statement. 
Memcpy and memcpyParams in one statement, Eval and evalParams in one statement. Params and m_params get separate statements. I actually prefer separate statements.
I usually think of a "tagged union" like this as encapsulating its tag -- so that m_action would be in identical member of each Params fork -- or maybe defined by a common base class. But I guess that's not perfect since you'd have to arbitrarily pick on one of the two branches to use to address the identical/overlapping field.
Your placement of dstFieldIndex strangely goes the other way -- defining like-named fields that serve the same purpose/semantics at different offsets in the two branches. Why isn't dstFieldIndex a direct member of ProjectStep? 
Skipping the m_ convention for the union branch fields seems unhelpful -- search for dstFieldIndex and you find two different members and an accessor function and various initializers and other locals.

src/ee/executors/OptimizedProjector.cpp
```diff
@@ -0,0 +1,528 @@
...
+};
+
+// Implement less than.  We want to order by field index in the
+// destination tuple.  Source tuple field index is not appropriate
+// for ordering: fields in source tuple may be referenced more
+// than once, or projection expression may not be a TVE.
+struct StepComparator {
```
paulmartel: Feeling "twice-shy" about inlines? I'd probably have made most of these in-class-body inlines, but especially the trivial tests and accessors and comparator.

src/ee/executors/OptimizedProjector.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+    union Params {
+
+        Params(AbstractExpression* expr, int dstFieldIndex);
+
+        Params(int dstFieldIndex, int srcFieldIndex, size_t dstOffset, size_t srcOffset, size_t numBytes);
+
+        struct Memcpy {
+            int dstFieldIndex;
+            int srcFieldIndex;
```
paulmartel: srcFieldIndex is redundant -- contiguity could be checked by comparing the srcOffset+numBytes of one step to the srcOffset of the next -- maybe add an accessor "srcEndOffset() { return srcOffset+numBytes; }"

tests/ee/executors/OptimizedProjectorTest.cpp
wolffcm: The right thing to do would probably be to put smart pointers in a std::vector.  Boost's shared_ptr would do the trick, but I think it's weird to do that if there's really only one owner.  In C++11, containers have `emplace` methods to construct in-place, so one could use scoped_ptr (or alternatively use C++11's unique_ptr).

src/ee/executors/OptimizedProjector.cpp
wolffcm: I'm not sure about this one.  I wasn't ready to make the assumption that all the fields in a record will always be tightly packed.  There's a couple reasons we might want to put a few bytes of padding in between values in a record: for performance (since aligned loads can be faster), or if we ever port to a platform that requires loads to be aligned for some types.

src/ee/executors/OptimizedProjector.cpp
```diff
@@ -0,0 +1,528 @@
...
+    // Returns the field index in the destination tuple for this step.
+    // For memcpy steps that span multiple contiguous fields, returns the
+    // lowest field index.
+    int dstFieldIndex() const;
+
+    // Returns the source index for TVE expr steps, and memcpy steps
+    // that have not been coalesced, and -1 otherwise.
+    int srcFieldIndex() const;
+
+    // For memcpy steps, returns the offset in the dst tuple
```
paulmartel: The next 3 accessors only require private visibility.

src/ee/executors/OptimizedProjector.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+    assert (!steps.empty());
+
+    if (steps.size() == 1) {
+        return *(steps.begin());
+    }
+
+    size_t dstOffset = steps.begin()->dstOffset();
+    size_t srcOffset = steps.begin()->srcOffset();
+    size_t numBytes = (steps.rbegin()->dstOffset() - steps.begin()->dstOffset()) + steps.rbegin()->numBytes();
```
paulmartel: steps.begin()->dstOffset() is already cached in dstOffset.

src/ee/executors/OptimizedProjector.cpp
```diff
@@ -0,0 +1,528 @@
...
+                                          const TupleSchema* srcSchema,
+                                          const ProjectStepSet& steps) {
+    ProjectStepSet outSteps;
+    BOOST_FOREACH(const ProjectStep& step, steps) {
+        AbstractExpression *e = step.expr();
+        if (e->getExpressionType() == EXPRESSION_TYPE_VALUE_TUPLE) {
+            const TupleSchema::ColumnInfo* dstColInfo = dstSchema->getColumnInfo(step.dstFieldIndex());
+            const TupleSchema::ColumnInfo* srcColInfo = srcSchema->getColumnInfo(step.srcFieldIndex());
+
+            // XXX test this!
```
paulmartel: Comment is more scary than informative. Clarify?

src/ee/executors/OptimizedProjector.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include "expressions/abstractexpression.h"
+
+namespace voltdb {
+
+// Forward declarations
+class TableTuple;
+class TupleSchema;
+class ProjectStep;
+struct StepComparator;
+typedef std::set<ProjectStep, StepComparator> ProjectStepSet;
```
paulmartel: Taking a step back, I question the usefulness of the ProjectStep abstraction.
Wouldn't this implementation be simplified by having ProjectStepSet be a wrapper around 2 separate sets, one of ProjectStepEval and one of ProjectStepMemcpy?
This would avoid the need for the union and the frequent switching/filtering on Action -- especially in the critical exec runtime path.
The ProjectStepEvals could just get passed through to the final state of the optimized projector "undigested"
leaving just the ProjectStepMemcpys to be optimized.

src/ee/executors/OptimizedProjector.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /** The number of steps needed to perform this projection */
+    size_t numSteps() const;
+
+    /** For testing, return an exppression for each step.  (This must be done before optimizing,
+     * before expressions are replaced with mem copies.)*/
+    std::vector<AbstractExpression*> exprs() const;
+
+private:
+
+    boost::scoped_ptr<ProjectStepSet> m_steps;
```
paulmartel: Is it worth pursuing the further exec runtime benefit of a simpler final representation of the optimized projector -- like a vector (vs. tree) of each kind of ProjectStep? The set(s) could be transient artifacts of the optimization process.
Maybe what I'm suggesting is that a very simple (OptimizedProjection?) class with eval method gets its state set by a temporary OptimizedProjector (ProjectionOptimizer?) class.

tests/ee/executors/OptimizedProjectorTest.cpp
```diff
@@ -0,0 +1,480 @@
...
+                default:
+                    assert(false);
+                }
+
+                srcTuple.setNValue(j, nval);
+            }
+
+            tbl->insertTuple(srcTuple);
+        }
+    }
```
paulmartel: Much of the above is framework not all that specific to this test.
Some of it looks somewhat reinvented (I'm assuming it isn't preexisting/copy/pasted)
-- similar to other test setup code that builds and populates tables.
Standardization across tests would be nice long-term.
MAYBE this is as good a candidate to standardize on as any predecessor.
Maybe a reasonable small step in that direction would be to package it outside this particular test class definition
(even if kept inside this code module for now) either as a Test-based base class or as a helper class.

src/ee/executors/OptimizedProjector.cpp
wolffcm: I don't have a strong feeling either way---maybe for small methods, it's nice to have the definitions inline.  Sometimes I'll just scan a class definition to see if there's any methods that I'll find useful, and in that case, longer method bodies can make it harder to read.

src/ee/executors/OptimizedProjector.cpp
wolffcm: It's possible to have projections between different types, where values are implicitly casted.  The comment was a note to myself to exercise this code.  I'm not sure exactly how to do this.  I should figure this out before I merge (and then remove the comment).

tests/ee/executors/OptimizedProjectorTest.cpp
wolffcm: There's some code in existing tests but it's not very well encapsulated and assumes a fixed schema.  I wrote these methods thinking of re-use in the future.

I'd like to move these methods (for schema, table, and data creation) to a separate file, visible to the other tests, perhaps to a class or namespace.

src/ee/executors/OptimizedProjector.hpp
wolffcm: I agree with your thoughts on design, both in this comment and the preceding one.  Do you see this happening before I merge this branch to master, or as a follow-on ticket?

paulmartel:
End of review round.

wolffcm:
Paul, I've added some more commits that address most of your comments, except for the ones that call for some redesign, for which I've filed ENG-8039 like we talked about.

paulmartel:
The changes look good.
BTW, I don't think the "future world with alignment" argument holds up as a justification for using column index vs. column offset + column size as the criteria for contiguity. This code would kinda fall apart anyway in that world, since the source and destination tuples could have DIFFERENT alignment paddings -- frustrating the column-spanning memcpy. To me, that seems to favor an approach that relies MORE on offset+size and LESS on column index -- it's just that the concept of size would have to be extended to account for "padded size"  -- and optimization would require validation against destination's padded size. We'd probably want to track value size AND padded size and only include the padding bytes when spanning columns and not for the stand-alone column or for the last column in a contiguous span. So, I'm still thinking that source column index is expendable -- so, maybe let's lose it for the upcoming redesign.

wolffcm:
Squashing and creating a new pull request on a different branch.

PR #2201, opened 2015-04-07
Eng   7079

guotong1988:
there is no eng-7079 branch in the master

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -1245,6 +1245,10 @@ public class SQLParser extends SQLPatternFactory
             throw new SQLParser.Exception(msg);
         }
 
+		if (filename.startsWith("~")) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -1245,6 +1245,10 @@ public static FileInfo parseFileStatement(FileInfo parentContext, String stateme
             throw new SQLParser.Exception(msg);
         }
 
+        if (filename.startsWith("~")) {
+            filename = filename.replace("~", System.getProperty("user.home"));
```
jxchinaren: @guotong1988 
Will not work for case like:
File "~tong/ddl.sql" if "tong" is your username. 

However, this is very rare and low priority, we are going to merge it now. 

Thanks.

jxchinaren:
Very low risk, so merge it without running it on a branch.

PR #2202, opened 2015-04-07
Guice

atratus:
This pull request intended for acquaintance and analysis only.
It introduces Guice IoC and involves redesign of RealVoltDB class its dependencies in order to be handled by Guice container smoothly.
JUnit tests are passed on latest version of this branch except of TestMixedVersionClusters which is failed cause actual version is set after SocketJoiner consumes it. So some addtional redesign should be applied in order to make this test pass.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1178,6 +1259,7 @@ public void run() {
         }
     }
 
+    /*
```
afrogers: Please delete rather than commenting lines -- it should make the diff easier to read.

atratus:
Removed commented code from RealVoltDB in the commit 96e7c5ba1b1b82c490da5b69e0bd243224622671

vtkstef:
NIce work there! I just took a look at it. Very interesting approach.

PR #2207, opened 2015-04-07
Use replaceFirst to avoid too much replacing

jxchinaren:


jxchinaren:
@guotong1988 
Use found that this should be the expected function to replace the first character "~".

PR #2210, opened 2015-04-08
Add bitwise operation support including bitand, bitor and bitxor

jxchinaren:


src/ee/expressions/bitwisefunctions.h
```diff
@@ -19,4 +19,52 @@
...
+    return getBigIntValue(res);
+}
+
+
+template<> inline NValue NValue::call<FUNC_BITOR>(const std::vector<NValue>& arguments) {
+    assert(arguments.size() == 2);
+    const NValue& lval = arguments[0];
+    const NValue& rval = arguments[1];
+    if (lval.getValueType() != VALUE_TYPE_BIGINT || rval.getValueType() != VALUE_TYPE_BIGINT) {
+        throw SQLException(SQLException::dynamic_sql_error, "unsupported non-BigInt type for SQL BITAND function");
```
wolffcm: You mean BITOR here, and BITXOR below.

wolffcm:
The ticket calls for EE tests.  My BITNOT ticket calls for them too.  Not sure that they're necessary?  Maybe Paul has thoughts on this.

src/ee/expressions/bitwisefunctions.h
```diff
@@ -19,4 +19,52 @@
...
+    }
+
+    int64_t lv = lval.getBigInt();
+    int64_t rv = rval.getBigInt();
+
+    int64_t res = lv & rv;
```
wolffcm: All of these need to return the null value if either operand is NULL, or am I missing something?

src/ee/expressions/bitwisefunctions.h
jxchinaren: Good catch. I was thinking that the planner expects a BigInt, so it probably never will hit the this check. However, I feel safe to do the check and do not want to crash the EE.

src/ee/expressions/bitwisefunctions.h
jxchinaren: I think the default behavior is to treat the Long.Min correctly and not support real NULL value, assuming the caller does not need to handle NULL value.

We may need to come up with a new function to handle NULL correctly but does not support MIN if needed.

wolffcm: Ok.  Is this the long-term behavior, to be merged with master?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -518,14 +519,14 @@ private FunctionCustom(int id) {
...
             case FUNC_ROUND :
-            case FUNC_BITAND :
-            case FUNC_BITOR :
-            case FUNC_BITXOR :
             case FUNC_DIFFERENCE :
             // A VoltDB extension to customize the SQL function set support
             case FUNC_DATEDIFF :
                 voltDisabled = DISABLED_IN_FUNCTIONCUSTOM_CONSTRUCTOR;
                 // $FALL-THROUGH$
+            case FUNC_BITAND :
```
wolffcm: This should get the `// A VoltDB extension...` comment markers?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -1063,11 +1064,14 @@ Object getValue(Session session, Object[] data) {
                         return null;
                     }
                 }
-
+                /************************* Volt DB Extensions *************************/
```
wolffcm: Should probably use the standard comment pattern that Paul is using for the HSQL upgrade, like elsewhere in this file.

tests/frontend/org/voltdb/planner/testplans-functions-ddl.sql
```diff
@@ -2,3 +2,12 @@ CREATE TABLE ENG913 (
 	name varchar(32),
     city varchar(32)
 );
+
+
+CREATE TABLE bit (
```
wolffcm: I guess this table definition has tabs, which is why it's indented funny here.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
jxchinaren: I think line https://github.com/VoltDB/voltdb/pull/2210#discussion-diff-28018897L525 is the beginning of "// A VoltDB extension..." marker.  It ends at line 535. 

Does that make sense?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
jxchinaren: I was struggling about it and finally pick up the block comments example at end of this file: Line 1751. I thought it is a block of changes and should use the block comments style. 

I can do either way anyway.

tests/frontend/org/voltdb/planner/testplans-functions-ddl.sql
jxchinaren: screen shot 2015-04-08 at 8 48 18 pm

App using textWrangler.

screen shot 2015-04-08 at 8 48 56 pm

App using vim.

Not sure why it is indented not correctly here. Maybe it's my editor problem. I will investigate.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
wolffcm: Paul has said that the `// A VoltDB extension... ` style is preferred although this one might still be okay.  He sent me an email with the details a while back but I'm having trouble finding it.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -3297,6 +3297,63 @@ public void testAliasesOfSomeStringFunctions() throws IOException, ProcCallExcep
...
+        client.callProcedure("NUMBER_TYPES.insert", pk, 1, 1, null, 1.0, 1.0);
+        vt = client.callProcedure("BITWISE_AND_OR_XOR", in, in, in, pk).getResults()[0];
+        validateRowOfLongs(vt, new long[]{bignum & in, bignum | in, bignum ^ in});
+
+        pk = 101; bignum = Long.MAX_VALUE; in = Long.MIN_VALUE;
+        client.callProcedure("NUMBER_TYPES.insert", pk, 1, 1, bignum, 1.0, 1.0);
+        vt = client.callProcedure("BITWISE_AND_OR_XOR", null, null, null, pk).getResults()[0];
+        validateRowOfLongs(vt, new long[]{bignum & in, bignum | in, bignum ^ in});
+    }
+
```
paulmartel: I'd like to see an index test and a group by test that used BITAND or BITOR with a mask of some "middle bits" to define duplicate index keys and groups from column values in different ranges -- filtering on bitfields seems like an important requirement.

tests/frontend/org/voltdb/planner/testplans-functions-ddl.sql
paulmartel: I see tabs on the TINYINT and BIGINT lines when I select the text on the github screen.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
patriceberube: Are those functions available in sql

jxchinaren: @patriceberube All these bitwise functions will probably be merged by end of this week. We finished all the coding and are in review process.

jxchinaren: @patriceberube I take it back after reading John's email. We will follow whatever the plan it is.

wolffcm:
Looks good to merge to me.  I see you've got the index tests in with the bit shift branch.

PR #2211, opened 2015-04-08
ENG-7652: Use Cluster Create Time In DR Consumer and periodic logging of...

wweiss-voltdb:
... cluster create time if database was recovered.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -38,6 +38,7 @@ import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1712,6 +1713,10 @@ void logDebuggingInfo(int adminPort, int httpPort, String httpPortExtraLogMessag
         for (String line : lines) {
             hostLog.info(line.trim());
         }
+        if (m_messenger.getInstanceId().getTimestamp() != m_clusterCreateTime) {
+            hostLog.info("Cluster RECOVERed from CREATE performed on " +
```
nshi: How about "Database was originally created on #######"?

PR #2214, opened 2015-04-09
Possible fix for intermittent TypeError for datetime and NoneType

paulmartel:


tests/scripts/examples/sql_coverage/fuzzynormalizer.py
```diff
@@ -23,10 +23,42 @@
...
+    # cmp also supports comparison of lists with scalars, treating the scalar
+    # as a singleton list, but, so far, that is not required.
+    if isinstanceof(x, types.ListType) and isinstanceof(y, types.ListType):
+        for (xn, yn) in zip(x, y):
+            rn = safecmp(xn, yn)
+            if rn:
+                return rn // return first difference
+        # all elements the same, return 0
+        # unless one list is longer, but even that is not an expected use case
+        return cmp(len(x), len(y))
```
jxchinaren: Should we check the length of these two list first, then the contents? Does these function needs a tri-value compare or boolean compare is enough?

jxchinaren:
As it's test change, I have not reviewed too carefully, but it really "looks" good.   ^_^

tests/scripts/examples/sql_coverage/fuzzynormalizer.py
paulmartel: This IS a three-value compare, so the contents SHOULD be checked first, the length difference matters only when they are prefix-equal.

PR #2216, opened 2015-04-10
Fix FunctionCustom switch for bitwise functions

wolffcm:


src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -525,10 +525,16 @@ private FunctionCustom(int id) {
                 // $FALL-THROUGH$
             case FUNC_BITAND :
                 name = Tokens.T_BITAND;
+                parseList = doubleParamList;
```
paulmartel: This code looks a little messed up still.
Most notably there is a fall-through from ROUND/DIFFERENCE/DATEDIFF that is setting name to T_BITAND.
Granted, these are DISABLED, but I imagine that the name setting will lead to strange diagnostic messages.
Yet, I don't think name even needs to be set in these cases -- See line 346.
If, as I suspect, name doesn't need to be set here, AND/OR/XOR can all just share code with the REPEAT case that follows.

wolffcm: Ok, I did an experiment on my local machine, letting AND/OR/XOR all just share code with the REPEAT case.  This seems to work---but I noticed that in Xin's original pull request, he had exactly this code.  He then added another commit that sets the `name` field.  I'd like to understand why.

More generally, I'm not clear on why we need to set the name explicitly for some functions but not others.  Maybe for functions that are aliased to others?

paulmartel: Yes. The explicit name requirement is an alias and overload specialization thing.
Or possibly it's just superstitious nonsense.
Maybe Xin can explain why he thought more code would be better, here.

jxchinaren: The reason to have the name setting is that I want to use it in the getSQL() function in the same file, which is required by running function expression index tests.

Yes, line 346 really helps setting the name member for bitand/bitor/bitxor functions. I was sorry not to miss it. The Hsqldb code sets the function member "name" variable at the final step before using it, but I think it's better to set it in the constructor when creating it. Because all the members are public accessible, it's really easy to miss where the members are changed. 

On the other hand, Hsqldb 1.9.3 that we used is not really functionally ready with bitand/bitor/bitxor. The code is there partly but not returning correct answer or having null pointer exception, which involves some extra work here. 

As this file is Hsqldb code, I agree with removing the "name" assignment code, maybe leaving comments. My bad.

wolffcm:
Merged per Xin's request.  He'll follow on with some code cleanup.

PR #2217, opened 2015-04-10
Fix compilation error on Mac.

nshi:


jhugg:
Send to paul/xin?

wolffcm:
Looks fine to me too.  Just curious, what was the compilation error?  It doesn't happen on my machine.

nshi:
I upgraded clang on my Mac yesterday and it started failing because it thinks the pointer `this` can never be `NULL` in a proper C++ context.

paulmartel:
WOW! So I guess this is incorrect code that only works by taking advantage of g++ implementation details:

        AbstractExpression* ae = NULL;
        if (std::string("NULL") == ae->AbstractExpression::debug()) {
            cout << "g++ rocks!";
        }

By extension, what I should really be doing to avoid undefined behavior in my robust library code is to make all of my member functions private and wrap each one in a public static member function that tests for null before calling the member function. That should cover all the undefined cases.

That's one hard-core coding standard. Do the clang tools offer a utility for generating stubs for these wrappers and rewriting callers to call them? It's probably on their roadmap.

The workaround implemented here is MOSTLY harmless -- most calls to debug will have already crashed if this is null. But I don't know if I consider this progress. In the non-virtually-called edge case above, it trades one technically undefined behavior that mostly works for another practically undefined behavior that mostly crashes.   Thanks, clang, for saving us from an undeserved success.

PR #2218, opened 2015-04-10
ENG-8024: Allow --replica on recover in the startup script.

nshi:


lib/python/voltcli/voltdb.d/recover.py
```diff
@@ -34,6 +34,9 @@
                                 safemode_available=True,
                                 supports_daemon=True,
                                 supports_multiple_daemons=True),
+    options = (
+        VOLT.BooleanOption('-r', '--replica', 'replica', 'recover replica cluster', default = False),
```
wijjo: Could have probably pushed it down to ServerBundle with a "supports_replica" keyword argument, now that more ServerBundle commands have a replica option. Not a big deal though.

PR #2220, opened 2015-04-10
Eng 8033 bitshift

jxchinaren:


jxchinaren:
It contains too much un-useful commits history, close it and send one within one commit.

PR #2222, opened 2015-04-10
patch for running Valgrind with VoltDB

dweiss88:


lib/python/voltcli/verbs.py
```diff
@@ -374,7 +374,9 @@ def initialize(self, verb):
            cli.StringOption(None, '--http', 'httpport', 'specify the http port as [ipaddress:]port-number'),
            cli.StringOption(None, '--internalinterface', 'internalinterface', 'specify the network interface to use for internal communication, such as the internal and zookeeper ports'),
            cli.StringOption(None, '--externalinterface', 'externalinterface', 'specify the network interface to use for external ports, such as the admin and client ports'),
-           cli.StringOption(None, '--publicinterface', 'publicinterface', 'For hosted or cloud environments with non-public interfaces, this argument specifies a publicly-accessible alias for reaching the server. Particularly useful for remote access to the VoltDB Management Center.'))
+           cli.StringOption(None, '--publicinterface', 'publicinterface', 'For hosted or cloud environments with non-public interfaces, this argument specifies a publicly-accessible alias for reaching the server. Particularly useful for remote access to the VoltDB Management Center.'),
+           cli.BooleanOption(None, '--valgrind', 'valgrind', None),
+           cli.StringOption(None, '--ipcport', 'ipcport', None))
```
paulmartel: There's getting to be enough of these options that maybe it makes sense to process them in alphabetical order to make it easier to scan the code for them.

src/ee/storage/ExportTupleStream.h
```diff
@@ -47,6 +47,9 @@ class ExportTupleStream : public voltdb::TupleStreamBase {
     void setBytesUsed(size_t count) {
         assert(m_uso == 0);
         StreamBlock *sb = new StreamBlock(new char[1], 0, count);
+        // Valgrind needs the newly created block to be initialized
+        // Leaving the last 8 bytes for MAGIC_HEADER_SPACE_FOR_JAVA
+        ::memset(sb->rawPtr(), 0, (count-8));
```
paulmartel: Add if (count > 8) or at least assert(count >=8) so "count-8" will not catastrophically wrap.

paulmartel: I am suspicious of this memset. If valgrind is complaining, either it's a false alarm or the code is probably expecting valid data where it didn't write any valid data. Are we sure that reading zeros is OK in that case? Could the code easily be fixed to only read the bytes it wrote? Are we sure that there is not a bug where we INTENDED to write specific values to the memory before reading it?

akhanzode:
Is this something good to merge?

paulmartel:
@akhanzode I've personally got some reservations -- I am hoping to have some time next week to try to repro the valgrind issues Dan was addressing and to make a judgment whether this change represents a true fix vs. just a cover-up that satisfies valgrind but hides a performance or correctness issue.

PR #2223, opened 2015-04-11
Support hexadecimal literals in sqlcmd's EXEC syntax

wolffcm:
What's missing from here?
- No VMC support (yet)
- No sqlcmdtest tests (I tested manually)

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -1379,7 +1379,7 @@ public class SQLParser extends SQLPatternFactory
 
         /**
          * Given a parameter string, if it's of the form x'0123456789ABCDEF',
-         * return a string containing just the digits.
+         * return a string containing just the digits.  Otherwise, return null.
          */
         private static String getDigitsFromHexLiteral(String paramString) {
             Matcher matcher = SingleQuotedHexLiteral.matcher(paramString);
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -1374,6 +1377,44 @@ private static String friendlyTypeDescription(String paramType) {
...
+            if (hexDigits.length() > 16) {
+                throw new SQLParser.Exception("Too many hexadecimal digits for BIGINT value");
+            }
+
+            if (hexDigits.length() == 0) {
+                throw new SQLParser.Exception("Zero hexadecimal digits is invalid for BIGINT value");
+            }
+
+
+            long val = new BigInteger(hexDigits, 16).longValue();
```
paulmartel: Worth a comment here to explain that Long.parseLong(hexDigits, 16) is not an option since it doesn't support the negative range. Might mention for the future that in Java 1.8 or later this call could be replaced with Long.parseUnsignedLong(hexDigits, 16) .

src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -1426,10 +1476,17 @@ else if (paramType.equals("timestamp")) {
...
+                            String hexDigits = getDigitsFromHexLiteral(param);
+                            if (hexDigits == null) {
+                                hexDigits = Unquote.matcher(param).replaceAll("");
+                            }
+                            objParam = Encoder.hexDecode(hexDigits);
                             // Make sure we have an even number of characters, otherwise it is an invalid byte string
-                            if (param.length() % 2 == 1) {
+                            if (hexDigits.length() % 2 == 1) {
+                                // This seems to be dead code:
+                                // Encoder will throw an exception before we get here.
```
paulmartel: I'm OK with removing the dead code or replacing with an assert.

paulmartel:
Looks good. Thanks.

PR #2231, opened 2015-04-15
Add various tests for live schema change for sparse index

jxchinaren:
Please check pro also.


tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -584,4 +584,19 @@ static public void checkDeploymentPropertyValue(Client client, String key, Strin
         }
         assertTrue(found);
     }
+
+    static public void checkQueryPlan(Client client, String query, String[] patterns) throws Exception {
```
paulmartel: If you define this function's signature as String... patterns instead of String[] patterns, you can eliminate the (Client, String, String) overload below and make this function easier to call with multiple patterns (by just passing one or more separated by commas).

PR #2234, opened 2015-04-16
ENG-8062 Workaround for making rolling upgrade work

rmorgenstein:
DO NOT MERGE. I want to squash these.
Rolling upgrade should only require a regex on the hotfix patch to allow a match against the running version.  Unfortunately, there is a later check during rejoin against the entire buildstring (loaded from buildstring.txt in the voltdb .jar)  which contains git --describe output. This branch allows the build to take a -Dbuildstring="..." arg that will write buildstring.txt instead of creating it from the version and git sha.

### DIFF
--- a/build-client.xml
+++ b/build-client.xml
@@ -154,10 +154,19 @@ JAR BUILDING
 ***************************************
 -->
 
-<target name="buildinfo">
+<target name="getversion">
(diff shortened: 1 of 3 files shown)


rmorgenstein:
squash-merged - https://github.com/VoltDB/voltdb/commit/84296f8d831684615c133d9e7a92d6d2a1f361e2

PR #2235, opened 2015-04-17
Support x-quoted literals for VARBINARY, but not integers (yet)

wolffcm:
This brings us back to where we were before release 5.2EA1, but more formally supports x-quoted literals for VARBINARY types (which is standard SQL), and adds testing for same.

Supporting x-quoted literals in integer contexts is to follow (ticket ENG-8046).


paulmartel:
Looks good - thanks for the additional VARBINARY coverage.

wolffcm:
Closing this pull request, will merge into release-5.2 instead.

PR #2237, opened 2015-04-18
Add support for aggregate distinct push down optimization

jxchinaren:
FOR REVIEW ONLY. DO NOT MERGE.

SqlCoverage has patterns on agg distinct also.


paulmartel:
This looks good.

PR #2240, opened 2015-04-19
ENG-8108: Don't delete empty segment on PersistentBinaryDeque.close().

nshi:


afrogers:
Reviewed. Premise looks good, but several memcheck-related test errors in Jenkins.

nshi:
Need to be squashed before merging.

PR #2245, opened 2015-04-20
disable the partial index matching for certain seqscan aggregate queries...

jxchinaren:
... optimization


paulmartel:
Looks good.

PR #2247, opened 2015-04-20
Fix pretend synchronization discovered while debugging ENG-8086

paulmartel:


BillWhite:
This looks right to me.

PR #2257, opened 2015-04-22
Fix ENG-6761 sqlcoverage TypeError comparing possibly null datetimes

paulmartel:


jxchinaren:
Looks good, close the old pull request?

PR #2258, opened 2015-04-22
ENG-7978: Add primary keys to the windowing example apps.

nshi:


examples/windowing-with-ddl/ddl.sql
```diff
@@ -19,7 +19,7 @@ CREATE TABLE timedata
   uuid VARCHAR(36) NOT NULL,
   val BIGINT NOT NULL,
   update_ts TIMESTAMP NOT NULL,
-  CONSTRAINT update_ts_uuid UNIQUE (update_ts, uuid),
+  CONSTRAINT update_ts_uuid PRIMARY KEY (update_ts, uuid),
```
wolffcm: The primary key for the table seems like UUID, not (UPDATE_TS, UUID).  If the example were larger and you wanted to join to table TIMEDATA, you'd likely do it on the UUID field.  Changing the constraint from being UNIQUE to PRIMARY KEY doesn't effect execution at all but seems kind of hacky.

It seems like this is important to DR to have a primary key.  Maybe the answer is to create two indexes?  

(Same comments for the other windowing example.)

nshi: Using both `(UPDATE_TS, UUID)` as the primary key just in case there are collision in the `UUID`.

> It seems like this is important to DR to have a primary key. Maybe the answer is to create two indexes?

Yes. DR uses `PersistentTable::lookupTuple()` for tuple lookup, and that only leverages the primary key.

wolffcm: Well, it seems weird for the timestamp field to be first in the primary key.  You wouldn't be able to join to this table without also specifying a timestamp.

nshi: Good point. What about creating a primary key on `(UUID, UPDATE_TS)` and leaving this index alone?

nshi: In fact, one of the optimizations we are doing right now will solve this problem without having to add a primary key to this table, the unique index will suffice. So I'm gonna leave this schema alone, but add a primary key to the other windowing app.

wolffcm: Being able to use any unique index instead of only the primary key---that was my next question.

What you propose sounds good to me.

nshi:
Creating another PR addressing the review comments.

PR #2260, opened 2015-04-22
Support hexadecimal literals as integers

wolffcm:
This changeset supports using hexadecimal literals (`X'00ABCD'`) to express integers in SQL.

This required some changes to HSQL to convert VARBINARY literals into BIGINTs based on context.  This is done for arithmetic (+, -, *, /) and logical (<. >, =, etc) expressions, and the new bitwise functions.

I decided *not* to support hexadecimal literals for the other, non-bitwise functions, since it seems really unlikely that a user would ever type something like `SQRT(X'0100')`.  The way HSQL handles the non-VoltDB specific functions this would require changes for each function.  I can file a ticket to support these other functions, if we feel there's a need for it.

### DIFF
--- a/src/frontend/org/voltdb/ParameterConverter.java
+++ b/src/frontend/org/voltdb/ParameterConverter.java
@@ -108,8 +108,26 @@ public class ParameterConverter {
     }
 
     private static final Pattern thousandSeparator = Pattern.compile("\\,");
+
+    private static boolean isIntegralClass(Class<?> clz) {
+        if (clz == long.class
+            || clz == int.class
+            || clz == short.class
+            || clz == byte.class) {
+            return true;
+        }
+        return false;
+    }
+
     /**
      * Given a string, covert it to a primitive type or return null.
+     *
+     * If the string value is a VARBINARY constant of the form X'00ABCD', the
+     * expected class is one of byte, short, int or long, then we interpret the
+     * string as specifying bits of a 64-bit signed integer (padded with zeroes if
+     * there are fewer than 16 digits).
+     * Corresponding code for handling hex literals appears in HSQL's ExpressionValue class
+     * and in voltdb.expressions.ConstantValueExpression.
      */
     private static Object convertStringToPrimitive(String value, final Class<?> expectedClz)
     throws VoltTypeException
@@ -138,7 +156,24 @@ public class ParameterConverter {
             }
         }
         // ignore the exception and fail through below
-        catch (NumberFormatException nfe) {}
+        catch (NumberFormatException nfe) {
+        }
+
+        // If we failed to parse the string in decimal form it could still
+        // be a numeric value specified as X'....'
+        //
+        // Do this only after trying to parse a decimal literal, which is the
+        // most common case.
+        if (isIntegralClass(expectedClz)) {
+            String hexDigits = SQLParser.getDigitsFromHexLiteral(value);
+            if (hexDigits != null) {
+                try {
+                    return SQLParser.hexDigitsToLong(hexDigits);
+                }
+                catch (SQLParser.Exception spe) {
+                }
+            }
+        }
 
         throw new VoltTypeException(
                 "tryToMakeCompatible: Unable to convert string "
--- a/src/frontend/org/voltdb/ParameterConverter.java
+++ b/src/frontend/org/voltdb/ParameterConverter.java
@@ -109,20 +109,10 @@ public class ParameterConverter {
 
     private static final Pattern thousandSeparator = Pattern.compile("\\,");
 
-    private static boolean isIntegralClass(Class<?> clz) {
-        if (clz == long.class
-            || clz == int.class
-            || clz == short.class
-            || clz == byte.class) {
-            return true;
-        }
-        return false;
-    }
-
     /**
      * Given a string, covert it to a primitive type or return null.
      *
-     * If the string value is a VARBINARY constant of the form X'00ABCD', the
+     * If the string value is a VARBINARY constant of the form X'00ABCD', and the
      * expected class is one of byte, short, int or long, then we interpret the
      * string as specifying bits of a 64-bit signed integer (padded with zeroes if
      * there are fewer than 16 digits).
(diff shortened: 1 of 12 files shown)


src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -108,8 +108,26 @@ else if (value != null) {
...
+            return true;
+        }
+        return false;
+    }
+
     /**
      * Given a string, covert it to a primitive type or return null.
+     *
+     * If the string value is a VARBINARY constant of the form X'00ABCD', the
+     * expected class is one of byte, short, int or long, then we interpret the
```
paulmartel: Add "and" for  "... and the expected class is..."

src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -138,7 +156,24 @@ private static Object convertStringToPrimitive(String value, final Class<?> expe
...
-        catch (NumberFormatException nfe) {}
+        catch (NumberFormatException nfe) {
+        }
+
+        // If we failed to parse the string in decimal form it could still
+        // be a numeric value specified as X'....'
+        //
+        // Do this only after trying to parse a decimal literal, which is the
+        // most common case.
+        if (isIntegralClass(expectedClz)) {
```
paulmartel: Not sure how much we care, but this code comes after the CSV-friendly code:

         // remove commas and escape chars
         value = thousandSeparator.matcher(value).replaceAll("");

We COULD preserve and use the original raw value for the x-literal case.
Do we want to allow/ignore or disallow commas and/or back-slashes in and around our x literals?

paulmartel: Move this block into the catch block above to avoid for non-integer non-double.
In that context, isIntegralClass() could be simplified to a test for the only NumberFormatException that we don't retry as a possible x-literal: (expectedClz != double.class).

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLogical.java
```diff
@@ -1699,4 +1703,29 @@ boolean reorderComparison(Session session) {
 
         return true;
     }
+    // A VoltDB extension to support X'..' as numeric literals
+    /**
+     * If one child is an integer, and the other is a VARBINARY literal, try to convert the
+     * literal to an integer.
+     */
+    private boolean convertBinaryIntegerLiteral(Session session, Expression lhs, Expression rhs) {
```
paulmartel: With the hsql232 upgrade, I've been moving towards smurf-naming the attributes and methods we add(ed) with a volt prefix. It helps. So, voltConvertBinaryIntegerLiteral, voltMutateToBigintType, etc.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLogical.java
```diff
@@ -1699,4 +1703,29 @@ boolean reorderComparison(Session session) {
...
+            nonIntegralExpr = rhs;
+        }
+        else if (rhs.dataType.isIntegralType()) {
+            nonIntegralExpr = lhs;
+        } else {
+            return false;
+        }
+
+        if (nonIntegralExpr.opType == OpTypes.VALUE && nonIntegralExpr.dataType.isBinaryType()) {
+            int whichChild = (nonIntegralExpr == rhs) ? RIGHT : LEFT;
```
paulmartel: set whichChild in the if/else block above, rather than reset here.

Should the "opType == OpTypes.VALUE && dataType.isBinaryType() and cast to ExpressionValue" preamble common to all of mutateToBigintType's callers be pushed down into it? That WOULD mean that mutateToBigintType would not be an instance method or maybe make it private with a static wrapper method that executes the preamble.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionValue.java
```diff
@@ -97,4 +102,47 @@ protected String describe(Session session, int blanks) {
...
+     * We assume that the VARBINARY constant is representing a
+     * 64-bit two's complement integer:
+     * - a constant with no digits returns false (no conversion)
+     * - a constant with more than 16 digits returns false (too many digits)
+     * - a constant that is shorter than 16 digits is implicitly zero-extended
+     *   (i.e., constants with less than 16 digits are always positive)
+     *
+     * These are the VoltDB classes that handle hex literal constants:
+     *   voltdb.ParameterConverter
+     *   voltdb.expressions.ConstantValueExpression
```
paulmartel: It's not clear from your comment what the relationship is here -- that each of these code paths should consistently implement the same cases. sqlcmd and VMC's exec param parsing should also be mentioned (though AFAIK the latter's x-literal support is TBD).

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionValue.java
```diff
@@ -97,4 +102,47 @@ protected String describe(Session session, int blanks) {
...
+     * @return true for a successful conversion and false otherwise.
+     */
+    boolean mutateToBigintType(Expression parent, int childIndex) {
+        if (valueData == null) {
+            return false;
+        }
+
+        byte[] data = ((BinaryData)valueData).getBytes();
+        if (data == null || data.length <= 0 || data.length > 16) {
+            return false;
```
paulmartel: It seems a little sad and strange that all of these different failure cases are summed up with a false return value -- and even that return value is not checked by one of your callers. That suggests that the resulting error messages are likely to be pretty generic. If at any point we know that we are in an unrecoverable situation (vs. some kind of a trial interpretation) we'd be better off throwing more specific errors.

tests/frontend/org/voltdb/compiler/TestVoltCompilerErrorMsgs.java
```diff
@@ -209,29 +215,42 @@ public void testErrorOnLimitPartitionRows() throws Exception {
...
+                "create procedure insHex as insert into blah (tinyval) values (x'FfffFfffFfffFf7f');");
+
+        // Hex constants not allowed to initialize DECIMAL or FLOAT.
+
+        ddlErrorTest("invalid format for a constant float value",
+                "create procedure insHex as insert into blah (floatval) values (x'80');");
+
+        ddlErrorTest("invalid format for a constant decimal value",
+                "create procedure insHex as insert into blah (decval) values (x'80');");
+
```
paulmartel: Test that it is possible (albeit inconvenient) to set the high bit of a TINYINT by with 14 Fs + an '8' or higher + one more digit.

paulmartel: I see that you have something like this covered in TestHexLiteralsSuite.

tests/frontend/org/voltdb/compiler/TestVoltCompilerErrorMsgs.java
```diff
@@ -39,7 +39,10 @@ private void statementTest(String feature, boolean expectError, boolean testStmt
         String simpleSchema =
             "create table blah (" +
             "ival bigint default 0 not null, " +
-            "sval varchar(255) not null" +
+            "sval varchar(255) not null," +
+            "tinyval tinyint," +
+            "floatval float," +
+            "decval decimal" +
```
paulmartel: Add a table that demonstrates valid usage of x-literal default values (is this supported? -- does it work compatibly in hsql backend?).
Add a test demonstrating ddl error messages resulting from invalid hex literal default value cases.

paulmartel:
End of review round.

src/frontend/org/voltdb/ParameterConverter.java
wolffcm: I'm looking at the `thousandSeparator` code.  It doesn't seem to do anything with backslashes, it just removes commas.  

Removing commas arbitrarily can result in some strange behavior: for example, we'll parse `5.2,6e,-,32,,,` as a valid double value.  We don't seem to document this behavior in CSVLoader.  Might it be documented somewhere else?  I'm more inclined (as a separate ticket) to rip out the comma-replacement altogether so that we can just write in our docs, "CSVLoader parses numeric values using Java's `parseLong`, `parseDouble`, etc methods."

Of course, it's possible that someone is relying on this behavior now, so maybe the right thing would to do is to preserve the original value for when we try to parse a hex literal, so as to stanch this strange behavior.  So I guess that's what I'll do.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionValue.java
wolffcm: My thinking here was that if we can't convert the literal to an integer (too many or too few digits), then we fall back to treating it as a varbinary literal, so the error message would be something like "incompatible data types in combination."

I think you're right that this would be confusing, so I can instead return the error "malformed numeric constant" instead, which is a little better.

wolffcm:
@paulmartel I added another commit that addresses your comments, and supports default values.  It turns out that there is explicit code in HSQL for unary minus in default values, so I didn't need to do anything special to support `DEFAULT -X'7F'`.

PR #2265, opened 2015-04-23
Improve display of DELETE statement for row limit triggers in VMC

wolffcm:
Some minor changes to the catalog report for row limit delete statements.  I also change how we capture the statement in HSQL so that whitespace is condensed, similar to what we do with views.


jxchinaren:
Looks good.

jxchinaren:
@wolffcm does that mean we used to display DELETE statement but not pretty?

wolffcm:
@jxchinaren That's right.  We used to preserve all the whitespace from the original DDL, but newlines would be changed to spaces.  So there would be big gaps of whitespace if the DELETE statement spanned multiple lines and was indented on each line.

PR #2266, opened 2015-04-23
ENG-8111: Don't crash server if table not found when applying binary log.

nshi:


src/ee/common/SerializableEEException.h
```diff
@@ -20,6 +20,8 @@
 
 #include <string>
 
+#define throwSerializableEEException(...) { char msg[8192]; snprintf(msg, 8192, __VA_ARGS__); throw voltdb::SerializableEEException(VOLT_EE_EXCEPTION_TYPE_EEEXCEPTION, msg); }
```
afrogers: Would it make sense to create a specific exception type for this case? Strikes me as a bit strange that we're adding a macro to throw a generic serializable exception for this case.

afrogers:
Other comment aside (to which the answer is probably, "It's fine as it is"), looks good.

afrogers:
In fact, answering for myself that it's fine as it is.

PR #2268, opened 2015-04-23
HSQL Style tweaks from subquery code review.

paulmartel:
These are just stylistic tweaks of Mike's work.
There were a few cases of minor whitespace-only or dead-code only changes from the original HSQL code that I backed out.
I cleaned the structured comments to more clearly mark disabled lines, new VoltDB patches, and cherry-picks from future hsql -- Mike made wise use of future code to patch current bugs, so the amount of new VoltDB code in this part of his changes turned out to be pretty minimal. We'll be glad of that soon when we upgrade. It may even help with some of the "known problems" I've already detected in the upgrade merge.
In a few places, I streamlined out some redundancy, but I don't think that I found any need to change the code's effective behavior.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
@@ -239,8 +239,7 @@ public class ExpressionColumn extends Expression {
 
     void collectObjectNames(Set set) {
 
-        /************************* Volt DB Extensions *************************/
-        // Implementation is taken from the HSQL 2.3.0
+        // BEGIN Cherry-picked code change from hsqldb-2.3.2
         switch (opType) {
 
             case OpTypes.SEQUENCE :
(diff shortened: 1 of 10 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -270,7 +269,22 @@ void collectObjectNames(Set set) {
 
                 return;
         }
-        /**********************************************************************/
+        /* Disable 13 lines
+        if (opType == OpTypes.SEQUENCE) {
```
jxchinaren: That's nice.

But how do you know that Michael dropped these 13 lines? Using your script to compare the difference with hsqldb source file?

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
```diff
@@ -1153,10 +1153,6 @@ public void reset() {}
 
     /************************* Volt DB Extensions *************************/
 
-    private Expression subqueryExpression;
-
-    void setSubqueryExpression(Expression sqe) { subqueryExpression = sqe; }
```
jxchinaren: Nice cleaning for unused code.

jxchinaren:
Looks good to me.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
paulmartel: No, mostly, where I restored lines and marked them as disabled, I just picked them up from current master, but since they weren't marked originally, I believe that they match the original hsql19b code. As I clean up my hsql upgrade tools, I will try them on this branch to see if I did a good job on the tools and on these comment changes.
I don't actually remember now what my tools expect to see for cherry picked changes -- maybe they expect to see the old code disabled or maybe they just expect to see JUST the cherry picked future version with the comments around it or maybe they handle both cases. I will find out soon.

PR #2274, opened 2015-04-27
fix the null search key wrong answer for counting index

jxchinaren:


src/ee/executors/indexcountexecutor.cpp
```diff
@@ -203,8 +210,8 @@ bool IndexCountExecutor::p_execute(const NValueArray &params)
                     assert (localLookupType == INDEX_LOOKUP_TYPE_GT || localLookupType == INDEX_LOOKUP_TYPE_GTE);
 
                     if (e.getInternalFlags() & SQLException::TYPE_OVERFLOW) {
-                        m_outputTable->insertTuple(tmptup);
-                        return true;
+                        earlyReturnForSearchKeyOutOfRange = true;
+                        break;
```
paulmartel: Here and below, you are in a "catch (const SQLException &e) {" block that really has only two possible outcomes:
"throw" or "break;"

You can clean this code up by removing all the "break;"s except the one at the bottom of the catch block -- this includes the "break;"s you added and other "break;"s that were already there in the other if/else branches. They can all just fall through.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -2178,6 +2178,16 @@ private void nullIndexSearchKeyChecker(Client client, String sql) throws Excepti
...
+
+        String sql1 = sql.replace("SELECT ID", "SELECT COUNT(ID)");
+        assertTrue(sql1.contains("SELECT COUNT(ID) FROM"));
+        vt = client.callProcedure("@AdHoc", sql1, null).getResults()[0];
+        validateTableOfScalarLongs(vt, new long[]{0});
+
+        String sql2 = sql.replace("SELECT ID", "SELECT COUNT(*)");
+        assertTrue(sql2.contains("SELECT COUNT(*) FROM"));
+        vt = client.callProcedure("@AdHoc", sql2, null).getResults()[0];
+        validateTableOfScalarLongs(vt, new long[]{0});
```
paulmartel: Nice high-leverage refactoring to multiply the test coverage -- maybe someday the indexcountexecutor/indexexecutor/nestloopindexexecutor will be able to share code as well.

End of review round. No blockers found.

PR #2282, opened 2015-04-28
Call generateOutputSchema for expression subqueries

wolffcm:


tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -1273,6 +1273,27 @@ public void testWhereRowSubSelects() throws NoConnectionsException, IOException,
...
+                val = 0;
+            expected[i][0] = val;
+        }
+
+        validateTableOfLongs(client,
+                "select (select count(*) from R_ENG8145_1 where ID > parent.num) from R_ENG8145_2 parent order by id;",
+                expected);
+        validateTableOfLongs(client,
+                "select (select count(*) from R_ENG8145_1 where ID > parent.num) from R_ENG8145_2 parent group by id;",
+                expected);
```
jxchinaren: Can you have a combination or ORDER BY /  GROUP BY case ? Interesting that the above query does not have " order by 1 " syntax. Because the result may not be order deterministic.

jxchinaren:
Looks good to go.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
paulmartel: @jxchinaren This query looks deterministic to me -- I don't have 100% confidence that the planner logic is thorough enough to figure that out, but it's possible.

wolffcm: For the first query, I guess since we're inserting into a replicated table, and the rows are pretty small, we'll use just one block.  I think I've seen sequential scan be non-deterministic because blocks are not necessarily scanned in the order they're allocated (maybe ordered by address?).  Within a block, we tend to scan in insertion order (assuming we haven't deleted rows and filled holes).

I've added a combo GROUP BY/ORDER BY query. 

I noticed that if I add a "ORDER BY 1" that I get a different error (exception in planner).   I'll file another ticket for it.

jxchinaren: @paulmartel "select (select count(*) from R_ENG8145_1 where ID > parent.num) from R_ENG8145_2 parent group by id;"

There is no ORDER BY, why deterministic?

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -554,6 +554,11 @@ private boolean getBestCostPlanForExpressionSubQueries(List<AbstractExpression>
                 m_recentErrorMsg = IN_EXISTS_SCALAR_ERROR_MESSAGE;
                 return false;
             }
+
+            // Generate the output schema for this subquery, which also has the side effect of
+            // creating inlined projections where needed to define the schema for the EE.
+            // For top-level queries, this is normally done in QueryPlanner.compileFromXML.
+            bestPlan.rootPlanGraph.generateOutputSchema(m_catalogDb);
```
jxchinaren: @wolffcm Do you want to move this subquery best plan generateOutputSchema() function into function planForParsedSubquery() at line 751 before 
"
subqueryScan.setBestCostPlan(compiledPlan);
"

I guess that's the place where it get the best plan but forget to call this generateOutputSchema().

wolffcm: That would mean that we'll call generateOutputSchema() for both the FROM subquery and expression subquery codepaths.  I believe that right now FROM subqueries are already covered in the top-level call to generateOutputSchema() invoked from QueryPlanner.  Is it okay to have redundant calls to generateOutputSchema for FROM subqueries?

jxchinaren: @wolffcm I see what you mean. It's fine to call it multiple times but not good for performance. Sorry, I forgot it also shared with FROM subquery.

Looks good now.

PR #2283, opened 2015-04-28
Add a java unit test for rolling back truncate table.

BillWhite:
Also, fixed a bug in allTypesMultiOpsJavaError which caused nothing to actually be tested.

### DIFF
--- a/tests/testprocs/org/voltdb_testprocs/regressionsuites/rollbackprocs/AllTypesMultiOpsJavaError.java
+++ b/tests/testprocs/org/voltdb_testprocs/regressionsuites/rollbackprocs/AllTypesMultiOpsJavaError.java
@@ -34,15 +34,28 @@ import org.voltdb.types.TimestampType;
     singlePartition = false
 )
 public class AllTypesMultiOpsJavaError extends VoltProcedure {
-    public final static int INSERT = 0;
-    public final static int UPDATE = 1;
-    public final static int DELETE = 2;
-    public final SQLStmt[] statements =
-        {new SQLStmt("INSERT INTO ALL_TYPES VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);"),
-         new SQLStmt("UPDATE ALL_TYPES SET TINY = ?, SMALL = ?, BIG = ?,"
+    public final static int INSERT   = 0;
+    public final static int UPDATE   = 1;
+    public final static int DELETE   = 2;
+    public final static int TRUNCATE = 3;
+    // Note: Need to make these all:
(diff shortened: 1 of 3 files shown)


tests/testprocs/org/voltdb_testprocs/regressionsuites/rollbackprocs/AllTypesMultiOpsJavaError.java
```diff
@@ -34,15 +34,28 @@
...
-        {new SQLStmt("INSERT INTO ALL_TYPES VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);"),
-         new SQLStmt("UPDATE ALL_TYPES SET TINY = ?, SMALL = ?, BIG = ?,"
+    public final static int INSERT   = 0;
+    public final static int UPDATE   = 1;
+    public final static int DELETE   = 2;
+    public final static int TRUNCATE = 3;
+    // Note: Need to make these all:
+    //          public static final SQLStmt XXX_STMT = new SQLStmt("...");
+    //       The reflection mechanism can't find them if they are in
+    //       non-static member functions like this, so they don't get
```
paulmartel: the comment seems a little confused/confusing: replace: "in non-static member functions like this" w/ "anonymous elements in the initializer of an array attribute like 'statements' " -- the real issue, as far as I understand.

Good catch, BTW.

paulmartel:
End of review round.

tests/testprocs/org/voltdb_testprocs/regressionsuites/rollbackprocs/AllTypesMultiOpsJavaError.java
BillWhite: Sorry.  That comment was to remind me how to fix the problem while I
investigated other places
where we may ignore thrown ProcCallExceptions incorrectly.  I don't think
there needs to be any
comment here at all, and I'll just delete it.

It turns out that I can't really tell from the status values whether a
ProcCallException is caused
by a VoltDB system defect or by a stored procedure system defect, such as a
divide by zero.
As we discussed before, it's not clear to me how to tell whether an NPE
comes from a VoltDB library
procedure, as it does in this case, or whether it comes from an error in
the user's stored
procedure.  For example, we frequently divide by zero in test procedures to
cause transactions
to abort.  That's indistinuishable from a user DBZ error, I reckon.  We
treat them all as transaction
failures and abort the transaction equally.  So I didn't make any changes
to the handling of
ProcCallException.  (Just to beat the deceased equine while it's down.)

PR #2286, opened 2015-04-29
fix the hard coded BigInt type problem

jxchinaren:


src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -473,9 +472,8 @@ public static AbstractExpression addScalarValueExpression(AbstractSubqueryExpres
         scalarExpr.setLeft(expr);
         scalarExpr.setValueType(expr.getValueType());
         scalarExpr.setValueSize(expr.getValueSize());
-        // reset scalarSubqueryExpr type to BIGINT
-        expr.setValueType(VoltType.BIGINT);
-        expr.setValueSize(VoltType.BIGINT.getLengthInBytesForFixedTypes());
+        expr.setValueType(expr.getValueType());
+        expr.setValueSize(expr.getValueSize());
```
paulmartel: Using setters and getters to set a member to itself?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -667,6 +667,8 @@ else if (child.name.equals("tablesubquery")) {
             col.tableAlias = scalarSubqueryExpr.getSubqueryScan().getTableAlias();
             // Need to add a ScalarValueExpression on top of the subquery expression
             // to be able to extract the subquery expression result
+
+            scalarSubqueryExpr.changeToScalarExprType();
```
paulmartel: This call precedes 1 of the 2 callers to ExpressionUtil.addScalarValueExpression. 
The other caller is in SubPlanAssembler.getIndexAccessPlanForTable.
My question is whether changeToScalarExprType() should be called INSIDE ExpressionUtil.addScalarValueExpression so that is gets called in both code paths.
I'm not sure how to exercise the other code path on a non-bigint type -- maybe with an "IN (select DESC from R)"? -- but we should make sure we have junit test coverage for that code path.

paulmartel:
End of review round.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
jxchinaren: create a new ticket ENG-8175 to solve the index usage work. Currently, it seems like a dead code or code that is not working properly for performance reason.

PR #2289, opened 2015-04-29
Added tools/hsqlup a handy toolkit for managing upstream hsqldb version ...

paulmartel:
...merges.


jxchinaren:
Do these shell scrip file ends with .sh ?

PR #2293, opened 2015-04-30
Eng 7520 better sum star message

BillWhite:
I wrote a test to diagnose the problem and implemented a new java junit test.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
@@ -373,7 +373,7 @@ public class SetFunction implements Serializable {
         if (setType == OpTypes.COUNT) {
             return Type.SQL_INTEGER;
         }
-        
+
         // A VoltDB extension to handle aggfnc(*) syntax errors.
         // If the argument node does not have
         // a data type, it may be '*'.  If the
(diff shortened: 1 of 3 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
```diff
@@ -373,7 +373,17 @@ static Type getType(int setType, Type type) {
         if (setType == OpTypes.COUNT) {
             return Type.SQL_INTEGER;
         }
-
+
```
paulmartel: There's a tab added here outside the annotated block. Please remove it.

tests/hsqldb/org/hsqldb_voltpatches/TestHSQLDB.java
```diff
@@ -282,6 +284,28 @@ public void testInsertIntoSelectFrom() {
...
     }
 
+    public void testSumStarFails() {
+        HSQLInterface hsql = HSQLInterface.loadHsqldb();
+        assertNotNull(hsql);
+
+        // The elements of this arraylist tells us the statement to
+        // execute, and whether we expect an exception when we execute
+        // the statement.
+        ArrayList<Pair<Boolean, String>> allddl = new ArrayList<Pair<Boolean, String>>();
```
paulmartel: This test could be stronger in a few respects:

    The loop tend to obscure which ddl statement is failing which kind of assertion (negative or positive).
You could get the same benefit of code reuse by factoring the loop body into its own method and invoking directly rather than adding to the map.

There is a junit method "assertEqual(value1, value2)" that reads more clearly than assertTrue(value1 == value2).

There is an optional initial string argument to both assertEqual and assertTrue where you can put helpful context (like the failing ddl statement).

Rather than a Boolean, you could label each statement with a possibly null string and validate that the HSQLParseException's message .contains it as a substring (a null string meaning expec successt). This would flag any case that gave some kind of non-sequiter message for a false negative.

paulmartel:
End of review round.

PR #2295, opened 2015-05-01
ENG-8036: Initial pass at optimized DR delete record format

afrogers:


nshi:
Squash merged externally

PR #2298, opened 2015-05-01
Review EE code in subquery branch

wolffcm:
Not many changes here, I just added some comments to clarify my understanding of the code.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -185,12 +185,13 @@ public:
              it != m_fragment->executeListEnd(); ++it) {
             assert(it->second != NULL);
             const vector<AbstractPlanNode*>& planNodeList = *it->second;
-            std::vector<AbstractExecutor*> executorList;
+            auto_ptr<vector<AbstractExecutor*> > executorList(new vector<AbstractExecutor*>());
             BOOST_FOREACH (AbstractPlanNode* planNode, planNodeList) {
                 initPlanNode(engine, planNode);
-                executorList.push_back(planNode->getExecutor());
+                executorList->push_back(planNode->getExecutor());
             }
-            m_subplanExecListMap.insert(make_pair(it->first, &executorList));
+            m_subplanExecListMap.insert(make_pair(it->first, executorList.get()));
+            executorList.release();
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/ee/plannodes/plannodefragment.cpp
```diff
@@ -89,8 +89,15 @@ void PlanNodeFragment::constructTree(AbstractPlanNode* node)
 
 PlanNodeFragment::~PlanNodeFragment()
 {
-    for (PlanNodeMapIterator mapIt = m_stmtExecutionListMap.begin(); mapIt != m_stmtExecutionListMap.end();) {
+    // The need to delete this could be avoided just by storing
+    // an actual instance (instead of a pointer) in the map?
+    //
+    // Also, why is the call to erase needed below?
```
jxchinaren: @wolffcm Do you plan to create a new ticket to address this issue or just leave with it?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -300,9 +303,15 @@ class ExecutorVector {
...
 {
+    // We could avoid having call delete here if we stored an instance in the
+    // map, rather than a pointer to it?
+    //
+    // Also, why do we need to call erase here?
     map<int, vector<AbstractExecutor*>* >::iterator it = m_subplanExecListMap.begin();
     while (it != m_subplanExecListMap.end()) {
         vector<AbstractExecutor*>* executorList = it->second;
+        // Note: we need to increment the iterator here
+        // to avoid invalidating it.
```
jxchinaren: it++, means to use it first and increment later. So it's probably fine here, right?

jxchinaren:
Looks good to me.

src/ee/plannodes/plannodefragment.cpp
wolffcm: I'll file a (very low priority) ticket for the minor nits that I found during review.

src/ee/execution/VoltDBEngine.cpp
wolffcm: Maybe this comment is confusing.  If one just does `erase(it)`, then `it` is invalidated, and attempting to increment an invalid iterator leads to crashes, etc.  Doing `erase(it++)` means that the iterator is incremented before `erase` completes, so the iterator is still valid on the next iteration of the loop.

I had originally tried to restructure this loop to a more familiar idiom, but ran into this problem, so I just added the comment for clarity.

jxchinaren: I see what you mean. Now it seems to be a serious problem that can crash. Right?

wolffcm: Now the loop works as intended, as it did before.  I just added the comment to clarify the subtleties.

wolffcm:
Merging this now, will address comments on a subsequent commit, once valgrind is passing.

PR #2302, opened 2015-05-05
ENG-8147 fixed VMC null pointer exception

dweiss88:


src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -271,19 +272,19 @@ private AuthenticationResult getAuthenticationResult(Request request) {
         }
         String admin = request.getParameter(PARAM_ADMIN);
 
+        AuthenticatedConnectionCache connection = m_connections.get();
```
akhanzode: Use a correct var name as connection_cache and m_connection_cache connection sounds like 1 connection but this is a cache.

PR #2303, opened 2015-05-05
Guard the argument to REPEAT(str,count) to prevent memory explosion.

BillWhite:
1. I had to move the const int POOLED_MAX_VALUE_LENGTH into the ThreadLocalPool.  It's still static, but it's now class static and not file static.
2. I wrote a small check in the existing C++ code.
3. I wrote a C++ unit test for this.  This required making the binary test function more generic.
3. I wrote a small test in the existing Java code.

### DIFF
--- a/src/ee/expressions/stringfunctions.h
+++ b/src/ee/expressions/stringfunctions.h
@@ -139,7 +139,7 @@ template<> inline NValue NValue::call<FUNC_REPEAT>(const std::vector<NValue>& ar
     const int32_t valueUTF8Length = strValue.getObjectLength_withoutNull();
     if ((count * valueUTF8Length) > ThreadLocalPool::POOLED_MAX_VALUE_LENGTH) {
         char msg[1024];
-        snprintf(msg, sizeof(msg), "data exception: string too long (%d > %d)",
+        snprintf(msg, sizeof(msg), "REPEAT function call would create a string of size %d which is larger than the maximum size %d",
                  count * valueUTF8Length, ThreadLocalPool::POOLED_MAX_VALUE_LENGTH);
         throw SQLException(SQLException::data_exception_string_data_length_mismatch,
                            msg);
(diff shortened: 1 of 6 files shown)


tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -2435,6 +2435,17 @@ public void testRepeat() throws NoConnectionsException, IOException, ProcCallExc
...
         assertEquals("foofoofoo", result.getString(1));
+
+        boolean sawException = false;
+        try {
+            cr = client.callProcedure("REPEAT", 1000000, 1);
+            assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+            result = cr.getResults()[0];
+            assertTrue("Exception expected here.", false);
+        } catch (ProcCallException ex) {
+            sawException = true;
```
wolffcm: Parent class `RegressionSuites` provides a method called `verifyProcFails` to make it easier to verify that something fails.  I try to use it (and `verifyStmtFails` for ad hoc statements) since it makes tests easier to read.

src/ee/expressions/stringfunctions.h
```diff
@@ -137,6 +137,13 @@ template<> inline NValue NValue::call<FUNC_REPEAT>(const std::vector<NValue>& ar
     }
 
     const int32_t valueUTF8Length = strValue.getObjectLength_withoutNull();
+    if ((count * valueUTF8Length) > ThreadLocalPool::POOLED_MAX_VALUE_LENGTH) {
+        char msg[1024];
+        snprintf(msg, sizeof(msg), "data exception: string too long (%d > %d)",
```
wolffcm: How does this message appear to the user?  Is there some indication that this error is occurring because the output of REPEAT would be very long?  If not, please add to this message to make it more helpful.  (The "substring error" above is not a terribly good example of an error message...)

wolffcm:
Thanks for adding the comments to the C++ unit test.  I'm not familiar with that code, so it was helpful.

End of review round.  Mostly I'd like to be assured that the message will be helpful to the user.

wolffcm:
Looks good.  Thanks for moving the Junit test.  Merge when you're ready.

PR #2304, opened 2015-05-05
Address valgrind issue and minor changes

wolffcm:
The first commit addresses the valgrind issue.

I added back in `std::` prefixes in VoltDBEngine.cpp.  I made this its own commit, since it makes it difficult to see the other changes.

I convinced myself that calls to `erase` that were in a couple destructors are not needed, so I removed them.

I also updated TestSubQueriesSuite now that valgrind is passing there.


jxchinaren:
Good work, Looks good to me.

PR #2311, opened 2015-05-07
Fix the bad parameterized AdHoc plan cache problem

jxchinaren:
Fix the bug on release branch, merged first to trigger the Jenkins run. Still needs to do code review.


paulmartel:
Looks good.

PR #2313, opened 2015-05-07
Eng 8209 adhoc parameters caching

jxchinaren:
Copy the patch changes of 4.0.1.x to master


src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -218,7 +218,9 @@ public synchronized AdHocPlannedStatement getWithSQL(String sql) {
     public synchronized void put(String sql,
                                  String parsedToken,
                                  AdHocPlannedStatement planIn,
-                                 String[] extractedLiterals)
+                                 String[] extractedLiterals,
+                                 boolean hasUserQuestionMark,
```
wolffcm: Let's add entries in the header comment for these new parameters.

tests/frontend/org/voltdb/regressionsuites/TestAdHocPlannerCache.java
```diff
@@ -0,0 +1,581 @@
...
+public class TestAdHocPlannerCache extends RegressionSuite {
+    private static final int CACHE_MISS1 = 0;
+    private static final int CACHE_MISS2 = 1;
+    private static final int CACHE_MISS2_MISS1 = 2;
+    private static final int CACHE_HIT1 = 3;
+    private static final int CACHE_HIT2 = 4;
+    private static final int CACHE_HIT2_HIT1 = 5;
+
+    private static final int CACHE_PARAMS_EXCEPTION = 6;
+    private static final int CACHE_SKIPED = -1;
```
wolffcm: "Skipped" has two P's in it, should be `CACHE_SKIPPED`.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -229,50 +231,55 @@ public synchronized void put(String sql,
...
-            // another invocation did not.  These are not counted separately (which would have to happen below
-            // after each call to boundVariants.add) because they are not evicted separately.
-            // It seems saner to use consistent units when counting insertions vs. evictions.
-            ++m_planInsertions;
-        } else {
-            for (BoundPlan boundPlan : boundVariants) {
-                if (boundPlan.equals(unmatched)) {
-                    matched = boundPlan;
-                    break;
+        // deal with L2 cache
```
wolffcm: In general, I'm finding this code pretty baffling.  As a reader I would benefit from a lot more comments here.

It wasn't immediately obvious that you've given new terminology to the different kinds of caches here.  There's not any member variables with L1 or L2 in the name, so I was't sure what you were referring to.

Why don't we want to populate the coreCache if the query failed to plan with extracted params?  I guess the coreCache is only for queries that have had literals extracted.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -229,50 +231,55 @@ public synchronized void put(String sql,
...
-        // then deal with the
-        AdHocPlannedStatement cachedPlan = m_literalCache.get(sql);
-        if (cachedPlan == null) {
-            m_literalCache.put(sql, plan);
-            ++m_literalInsertions;
-        }
-        else {
-            assert(cachedPlan.equals(plan));
+        // then deal with the L1 cache
+        if (! hasUserQuestionMark) {
```
wolffcm: I guess we don't put ad hoc queries with user question marks in the literal cache, since... actually this is not clear to me.  Isn't it possible that you might have a query that was unplannable when it was "force-paramterized", but can be planned literally, even if there are user question marks?

Can you elucidate with a comment in the code?

wolffcm: Is it possible that hasUserQuestionMark and wasBadParameterized could both be true?  If so I guess we don't cache anything.  Do I have that right?

tests/frontend/org/voltdb/regressionsuites/TestAdHocPlannerCache.java
```diff
@@ -0,0 +1,581 @@
...
+        VoltProjectBuilder project = new VoltProjectBuilder();
+        final String literalSchema =
+                "CREATE TABLE R1 ( " +
+                " ID BIGINT DEFAULT 0 NOT NULL, "
+                + "DESC VARCHAR(300), "
+                + "NUM bigint,"
+                + "RATIO FLOAT, "
+                + "PRIMARY KEY (desc)); "
+
+                + "create procedure proc1 AS select num as number from R1 where id > ? order by num;"
```
wolffcm: Might be good do some testing with an expression index (that contains literals)?
```SQL
CREATE INDEX EXPRIDX ON R1 (MOD(NUM, 3));

SELECT ... FROM R1 WHERE MOD(NUM, 3) = 0 AND ID = ?;
```
How would the caching work for something like that?

tests/frontend/org/voltdb/regressionsuites/TestAdHocPlannerCache.java
```diff
@@ -0,0 +1,581 @@
...
+        sql = "SELECT ID FROM R1 sub1 WHERE ID > 1 order by ID;";
+        vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        validateTableOfScalarLongs(vt, new long[]{2, 3});
+        checkPlannerCache(client, CACHE_MISS2_MISS1);
+
+        vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        validateTableOfScalarLongs(vt, new long[]{2, 3});
+        checkPlannerCache(client, CACHE_HIT1);
+
+        sql = "SELECT ID FROM R1 sub1 WHERE ID > 2 order by ID;";
```
wolffcm: I don't see how it's possible to get a hit in both the caches?  Wouldn't we stop after we got a hit in the "L1" (literal) cache?

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
jxchinaren: I borrow the L1, L2 from the planner statistics, they should mean literal cache and parameterized plan cache specifically. L1 is only for queries with exact literal sql string matching. L2 will cache extracted parameters query and user question mark query. 

Why don't we want to populate the coreCache if the query failed to plan with extracted params?
-- Query will be planned first with literal extracted parameterization. But once found exception (e.g. bad parameters), they will be planned just as the literal query without parameterization step.

For this case, L2 cache may contain non parameterized plan resulting lot of issues later: asserting, ArrayOutOfBoundException, NPE, etc. Remember L2 cache is intended for parameterized plan caching. However for this bad parameter query, it will be added into L1 cache.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
jxchinaren: No. It's not possible to have hasUserQuestionMark and wasBadParameterized to be true at the same time. I can rename "wasBadParameterized" to be "wasBadAutoParameterized", adding an assert, and a comment here to make it better to read.

wolffcm:
End of review round.

tests/frontend/org/voltdb/regressionsuites/TestAdHocPlannerCache.java
jxchinaren: The type here "CACHE_HIT2_HIT1" means HIT L2 cache, adding it to L1 cache also. I can rename the "CACHE_HIT2_HIT1" to be something "CACHE_HIT2_ADDL1". Does it make sense?

tests/frontend/org/voltdb/regressionsuites/TestAdHocPlannerCache.java
jxchinaren: If the query has ?, there is no difference with general ? AdHoc queries. 
If the query does not have ?, there is no behavior changed.

I can add a couple of tests for it also.

jxchinaren:
@wolffcm Add two new commits to resolve your comments.

wolffcm:
Looks good, thanks for adding the comments.

PR #2317, opened 2015-05-07
The rules for associating the names of primary key constraints to index algorithms.

BillWhite:
If the name of a primary key constraint contains the substring "TREE",
we always use a tree hash.  If the name contains the substring "HASH"
but not the substring "TREE", and the column type is unhashable, then we
use a tree index if the name is autogenerated, and throw an error if the
name is user supplied.  If the name does not contain "TREE" but it
does contains "HASH" and the column type is hashable, then we always
use a hash index.


tests/frontend/org/voltdb/regressionsuites/TestIndexesSuite.java
```diff
@@ -1133,6 +1135,43 @@ public void testUpdateRange() throws IOException, ProcCallException, Interrupted
...
+                            // explicitly name the constraint, we use the tree hash.
+                            "create table block ( goodhashname varchar(256) not null, primary key ( goodhashname ) );",
+                            // If we name the column with both a tree and hash name,
+                            // the tree name wins.  This doesn't actually test anything,
+                            // I suppose.
+                            "create table block ( goodtreehashname varchar(256) not null, primary key (goodtreehashname));",
+                            // If we name the constraint with both a tree and hash name, and the
+                            // field is not hashable, the tree name wins.
+                            "create table block ( goodvanilla varchar(256) not null, constraint goodtree_hash_constraint primary key ( goodvanilla ) );");
+        List<String> failing
```
paulmartel: Simpler to use array initialization
String passing[] = { ... };

paulmartel: Probably want to verify that you don't get a false error if the constraint name contains "hash" and the column is indeed hashably typed.
Probably want to verify that you do indeed get a hash implementation in this case.
The easiest way to do this is to define a procedure with 
callProcedure("@AdHoc", "create procedure procedurename as select pk from table order by pk;");
Then callProcedure("@ExplainProc", "procedurename") and scan the result for mentions of "IndexScan".

paulmartel: Actually verifying that you got a tree (IndexScan) for tables where you should would also be worth adding to your validations.

PR #2318, opened 2015-05-07
Do a better job at finding expressions in parsed statements

wolffcm:
Also, add more tests for DML subquery guards.

### DIFF
--- a/src/frontend/org/voltdb/planner/ParsedUpdateStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedUpdateStmt.java
@@ -18,6 +18,7 @@
 package org.voltdb.planner;
 
 import java.util.LinkedHashMap;
+import java.util.List;
 
 import org.hsqldb_voltpatches.VoltXMLElement;
 import org.voltdb.catalog.Column;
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/planner/ParsedUpdateStmt.java
```diff
@@ -73,4 +74,15 @@ public String toString() {
 
         return retval;
     }
+
+    @Override
+    public List<AbstractExpression> findAllSubexpressionsOfClass(Class< ? extends AbstractExpression> aeClass) {
+        List<AbstractExpression> exprs = super.findAllSubexpressionsOfClass(aeClass);
+
+        for (AbstractExpression expr : columns.values()) {
+            exprs.addAll(expr.findAllSubexpressionsOfClass(aeClass));
```
jxchinaren: Is there a case the "expr" that can be NULL ,like the INSERT case?

wolffcm: I don't think so.  For the INSERT case, the expression can be NULL for `INSERT INTO ... SELECT`, because there is no VALUES clause.  I'll verify that `UPDATE tbl SET field = NULL` doesn't generate an actual null in Java for this.

wolffcm: Using NULL in the SET clause for updates won't create a Java null, so I think we are safe.  But we are also dependent on what the HSQL parser does, so to be extra safe I pushed another commit that has the null check anyways.

PR #2323, opened 2015-05-08
ENG-8084 changed crash to info message

dweiss88:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1804,7 +1804,7 @@ void logDebuggingInfo(int adminPort, int httpPort, String httpPortExtraLogMessag
                 String parts[] = sb.toString().split(" ", 2);
                 if (parts.length == 2) {
                     parts[0] = parts[0].trim();
-                    parts[1] = parts[1].trim();
+                    parts[1] = parts[0] + "_" + parts[1].trim();
```
akhanzode: Why combine is this just build string? Looks like you are splitting again for printing.

akhanzode:
Is it possible to write a test for this?

src/frontend/org/voltdb/RealVoltDB.java
dweiss88: Ya, I'm combining so I can print out the version number in validate build string. I wasn't sure how else to get the version number of the cluster without changing VoltZK

akhanzode: Can you write a test for this change in behavior..?

dweiss88:
I'm not sure how we could do a unit test for this, but I think the plan is to create a system test for handling rejoin testing in the future.

akhanzode:
Look at TestMixedVersionClusters

dweiss88:
The issue with including the test is that the buildstring needs to be updated to test the bug that this ticket fixes. I think it might be possible, but I'd need to look deeper into how much I'd need to modify the code to make buildstring settable from the test.

akhanzode:
Sure good to find out if this test can be modified to use diff build string and successful server actions. Feel free to modify local cluster if needed.

PR #2325, opened 2015-05-11
Avoid wrong answers when using subqueries within other expressions

wolffcm:
These changes add a post-pass over parsed expression trees to wrap scalar subqueries in an instance of ScalarValueExpression where needed.  I also pushed the check for single-column schema in to the code that creates the ScalarValueExpression node.

Another approach might be to always create the ScalarValueExpression over subqueries, and then remove the SVE node if its parent is a comparison operator or an EXISTS operator.  We still wouldn't avoid the overhead of a post-pass though: we would need to check that the subquery children of SVE nodes have only one column after parsing is complete.

### DIFF
--- a/src/frontend/org/voltdb/expressions/ExpressionUtil.java
+++ b/src/frontend/org/voltdb/expressions/ExpressionUtil.java
@@ -493,7 +493,7 @@ public abstract class ExpressionUtil {
         return scalarExpr;
     }
 
-    private static boolean isScalarSubqueryContext(AbstractExpression parentExpr) {
+    private static boolean subqueryRequiresScalarValueExpressionFromContext(AbstractExpression parentExpr) {
         if (parentExpr == null) {
             // No context: we are a top-level expression.  E.g, an item on the
             // select list.  In this case, assume the expression must be scalar.
@@ -503,6 +503,11 @@ public abstract class ExpressionUtil {
         // Exists and comparison operators can handle non-scalar subqueries.
         if (parentExpr.getExpressionType() == ExpressionType.OPERATOR_EXISTS
                 || parentExpr instanceof ComparisonExpression) {
+            return false;
+        }
+
+        // There is already a ScalarValueExpression above the subquery.
+        if (parentExpr instanceof ScalarValueExpression) {
             return false;
         }
 
@@ -524,7 +529,7 @@ public abstract class ExpressionUtil {
         }
 
         if (expr instanceof SelectSubqueryExpression
-                && isScalarSubqueryContext(parentExpr)) {
+                && subqueryRequiresScalarValueExpressionFromContext(parentExpr)) {
             expr = addScalarValueExpression((SelectSubqueryExpression)expr);
         }
         return expr;
--- a/src/frontend/org/voltdb/expressions/ExpressionUtil.java
+++ b/src/frontend/org/voltdb/expressions/ExpressionUtil.java
@@ -493,6 +493,18 @@ public abstract class ExpressionUtil {
             }
         }
 
+        // Let's not forget the args, which may also contain subqueries.
+        List<AbstractExpression> args = expr.getArgs();
+        if (args != null) {
+            for (int i = 0; i < args.size(); ++i) {
+                AbstractExpression arg = args.get(i);
+                AbstractExpression newArg = wrapScalarSubqueriesHelper(expr, arg);
+                if (newArg != arg) {
+                    expr.setArgAtIndex(i, newArg);
+                }
+            }
+        }
+
         if (expr instanceof SelectSubqueryExpression
                 && subqueryRequiresScalarValueExpressionFromContext(parentExpr)) {
             expr = addScalarValueExpression((SelectSubqueryExpression)expr);
--- a/src/frontend/org/voltdb/expressions/ExpressionUtil.java
+++ b/src/frontend/org/voltdb/expressions/ExpressionUtil.java
@@ -474,25 +474,39 @@ public abstract class ExpressionUtil {
         return wrapScalarSubqueriesHelper(null, expr);
     }
 
-    /**
-     * Add a ScalarValueExpression on top of the SubqueryExpression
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -461,13 +462,85 @@ public static void generateSubqueryExpressionOutputSchema(AbstractExpression exp
...
+        if (rightChild != null) {
+            AbstractExpression newRight = wrapScalarSubqueriesHelper(expr, rightChild);
+            if (newRight != rightChild) {
+                expr.setRight(newRight);
+            }
+        }
+
+        if (expr instanceof SelectSubqueryExpression
+                && subqueryRequiresScalarValueExpressionFromContext(parentExpr)) {
+            expr = addScalarValueExpression((SelectSubqueryExpression)expr);
```
jxchinaren: call function scalarSubqueryExpr.changeToScalarExprType(); ?

This will address the problem for float type operation. 

The other option is to put scalarSubqueryExpr.changeToScalarExprType() inside of function addScalarValueExpression(). We used not to put it in because the getIndexAccessPlanForTable() may use it. But now the index access code is not addressed, so make sense to put the changeToScalarExprType() inside.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -461,13 +462,85 @@ public static void generateSubqueryExpressionOutputSchema(AbstractExpression exp
...
+            }
+        }
+
+        AbstractExpression rightChild = expr.getRight();
+        if (rightChild != null) {
+            AbstractExpression newRight = wrapScalarSubqueriesHelper(expr, rightChild);
+            if (newRight != rightChild) {
+                expr.setRight(newRight);
+            }
+        }
```
jxchinaren: Do you need to iterate the arguments of the expression? It may contains SelectSubqueryExpression also.

Maybe add a new test case for it also.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -1405,6 +1405,22 @@ public void testSubqueryWithExceptions() throws Exception
...
+      // caused the result of a subquery to be seen as the subquery ID, rather than the contents
+      // of subquery's result table.
+
+      validateTableOfScalarLongs(client, "select (select max(wage) from r1) from r1", new long[] {300});
+      validateTableOfScalarLongs(client, "select (select max(wage) from r1) + 0 from r1", new long[] {300});
+
+      validateTableOfScalarLongs(client, "select wage from r1 where wage = (select max(wage) from r1)", new long[] {300});
+      validateTableOfScalarLongs(client, "select wage from r1 where wage = (select max(wage) - 30 from r1) + 30", new long[] {300});
+  }
+
```
jxchinaren: Add tests for:
1. FLOAT type operation
2. scalar value subquery exists in arguments of a function, not sure which one use the m_args, like TRIM, DECODE maybe?

jxchinaren:
End of review round.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
wolffcm: I had the same thought and the call to changeToScalarExprType is already in addScalarValueExpression.  :+1:

src/frontend/org/voltdb/expressions/ExpressionUtil.java
wolffcm: Yes!  Thank you.  I'll recurse into the arguments as well and add tests.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -461,13 +462,85 @@ public static void generateSubqueryExpressionOutputSchema(AbstractExpression exp
...
-     * return ScalarValueExpression
+     * @return ScalarValueExpression
      */
-    public static AbstractExpression addScalarValueExpression(AbstractSubqueryExpression expr) {
+    private static AbstractExpression addScalarValueExpression(SelectSubqueryExpression expr) {
+        if (expr.getSubqueryScan().getOutputSchema().size() != 1) {
+            throw new PlanningErrorException("Scalar subquery can have only one output column");
+        }
+
+        expr.changeToScalarExprType();
```
paulmartel: Only partly apropos of this code change -- mostly for my own edification -- Do we have guards elsewhere against row subqueries like "WHERE (a,b) >= (select max(c), max(d) ..."? 
Or, is this that very guard, providing all the coverage we need?

wolffcm: This will certainly guard against the case that you describe.  Whether or not there's guards in other places, I'm not sure.

wolffcm: Wait. Actually the query you've described is legal, and we return the right answer.  If there is a mismatch in the number of columns in the LHS and RHS, HSQL will complain.

wolffcm: This produces some weird behavior though:
```SQL
select * from t where (bi, bi2) + 3 > 5;
```
It gives a misleading error message: "data type cast needed for parameter or null literal".  I'll file a bug.

wolffcm:
I found that this fix does not handle the `IN` operator correctly (which uses the `m_args` field of `AbstractExpresion`).  It's not clear to me how `IN` differentiates between its right child and its args field.  I'll need to look at this and then push another commit.

wolffcm:
I added a commit to recurse into the `m_args` field for expressions, and a test case for IN.

I also updated the `validate` method for IN expressions (even though it seems to be unused) because it misled me---the args field is not used for IN as far as I can tell.

jxchinaren:
Thanks for the fix. Looks good to me.

PR #2327, opened 2015-05-11
ENG-8237: Bound PersistentBinaryDeque memory usage

nshi:


src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -111,29 +124,37 @@ private void incrementNumEntries(int size) throws IOException {
...
-            m_buf = DBBPool.wrapMBB(m_fc.map(MapMode.READ_ONLY, 0, size));
-            m_readBuf = m_buf.b().duplicate();
-            m_buf.b().position((int) size);
-            m_readBuf.position(SIZE_OFFSET + 4);
+
+        try {
+            if (!m_file.exists()) {
+                m_syncedSinceLastEdit = false;
+            }
+            if (m_ras != null) {
```
afrogers: This should probably be an assert now, since if it's not null, we should also not have been closed above.

nshi: Good point. I'll make that change.

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -111,29 +124,37 @@ private void incrementNumEntries(int size) throws IOException {
...
+                //If it isn't for write, map read only to the actual size and put the write buf position at the end
+                //so size is reported correctly
+                final long size = m_fc.size();
+                m_buf = DBBPool.wrapMBB(m_fc.map(MapMode.READ_ONLY, 0, size));
+                m_readBuf = m_buf.b().duplicate();
+                m_buf.b().position((int) size);
+                m_readBuf.position(SIZE_OFFSET + 4);
+            }
+        } finally {
+            m_closed = false;
```
afrogers: This seems like a potentially wonky place to mark this open. If we've hit an exception opening/mmapping the file, the segments probably not useable? Maybe close it here and let the caller try again?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -326,19 +344,21 @@ public synchronized BBContainer poll(OutputContainerFactory ocf) throws IOExcept
...
+
+            if (s.hasMoreEntries()) {
+                segment = s;
+                retcont = segment.poll(ocf);
+                break;
             }
         }
-        if (retcont == null) {
+
+        if (segment == null || retcont == null) {
```
afrogers: Not sure I understand the added check here. If retcont != null, we're guaranteed that segment != null

PR #2328, opened 2015-05-12
Eng 8168 alias functions

paulmartel:
Keep indexes on keyword-parameterized functions and their aliases like "WEEK" (a.k.a. "EXTRACT(WEEK...") from corrupting catalog replay, effectively freezing the DDL.

Similarly, fix CAST operators in indexes having the same effect.

Fix a handful of functions that would not allow indexing at all (like "RIGHT").

Implement TRIM variants as separate volt functions -- for consistency with other keyword-parameterized functions like EXTRACT (also reduces runtime overhead). Make them wrap a common NValue method for code reuse. Correct the tricky handling of optional / defaulted string arguments in TRIM and SUBSTRING.

Replace over-general "function alias" concept with a more precise "implied argument" concept.
Unify the model (and json format) for function argument lists to be "possibly empty / never null"

Simplify FunctionExpression hashing and equality to rely on function_id vs the various dependents of function_id like name.

Test round-trip parsing and index-ability of ALL sql functions by using them in fulddl.sql indexes.

Generalize parser support for SQL function calls to ALWAYS also allow use of (all) commas in the place of custom argument separators like "FROM", "FOR", "USING" to allow canonical ddl generation to always use commas. Unit test the new and existing comma-separated variant syntax.


src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -494,7 +494,7 @@ public void toJSONString(JSONStringer stringer) throws JSONException {
             stringer.key(Members.RIGHT.name()).value(m_right);
         }
 
-        if (m_args != null && m_args.size() > 0) {
+        if (m_args != null) {
```
jxchinaren: when m_args is empty, do we need to serialize the JSON string?

paulmartel: I decided on purpose that this is not worth "optimizing/complicating" because A) m_args is almost never empty (only for CURRENT_TIMESTAMP, (someday maybe PI ...) and B) I would rather have the consistency of always having non-null m_args than to be unsure -- testing for both empty and null everywhere -- empty in case the expression came from voltXML and null in case it came from round-tripped JSON (as it does in the indexed expression case).

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
```diff
@@ -312,6 +320,21 @@ protected FunctionSQL(int id) {
                     Tokens.TIMEZONE_MINUTE, Tokens.FROM, Tokens.QUESTION,
                     Tokens.CLOSEBRACKET
                 };
+                // A VoltDB extension to support more flexible call syntax
+                // Allowing commas in the place of the traditional separator
```
jxchinaren: Probably need to talk to Andrew about the new support of syntax?

src/frontend/org/voltdb/expressions/FunctionExpression.java
```diff
@@ -26,23 +25,19 @@
 import org.voltdb.types.ExpressionType;
 
 public class FunctionExpression extends AbstractExpression {
-    public enum Members {
+    private enum Members {
         NAME,
-        ALIAS,
+        IMPLIED_ARGUMENT,
```
jxchinaren: Comments for these "new" members?

"IMPLIED_ARGUMENT" is for KEYWORD in the function?
"RESULT_TYPE_PARAM_IDX" is the index of parameter that affects the return type of the function.

src/frontend/org/voltdb/expressions/FunctionExpression.java
```diff
@@ -291,14 +268,44 @@ public void resolveForTable(Table table) {
...
-        if (FunctionForVoltDB.isUnitFunction(m_functionId)) {
-            result += connector + m_alias.substring(m_name.length()+1);
-            connector = ", ";
-        }
-        if (m_args != null) {
+
+        if ( ! m_args.isEmpty()) {
+            String connector = "(";
+
+            // The purpose of m_impliedArgument is to allow functions with
```
jxchinaren: Nice explanation.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
jxchinaren: One question: Are the new comms support necessary to have for the fix of round trip ddl? Or just nice to have?

jxchinaren:
Sorry, my reviews are not made together, interrupt by various stuffs. 

I realize this pull request also fixes ENG-8167, but does it also fix ENG-8243 by any chance?

Finish my review. Looks good to me.

paulmartel:
Committing to 5.3 release branch from separate squashed branch.

PR #2334, opened 2015-05-14
Add the fix for corelated subquery with group by and having

jxchinaren:


src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1587,25 +1587,35 @@ VoltXMLElement voltGetXML(Session session, List<Expression> displayCols,
...
-                // This mechanism of finding the expression that a SIMPLE_COLUMN
-                // is referring to is inherently fragile---columnIndex is an
-                // offset into different things depending on context!
-                if (otherCol != null && (otherCol.opType != OpTypes.SIMPLE_COLUMN) &&
-                         (otherCol.columnIndex == this.columnIndex)  &&
-                         !(otherCol instanceof ExpressionColumn))
+            if (displayCols == null) {
+                if (this instanceof ExpressionColumn == false) {
+                    throw new org.hsqldb_voltpatches.HSQLInterface.HSQLParseException(
+                            "VoltDB does not support this complex query currently.");
```
wolffcm: Maybe add a planner test that gets this message?

jxchinaren: Not known any query that will hit this problem. I just put there instead of hitting an NPE later.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1587,25 +1587,35 @@ VoltXMLElement voltGetXML(Session session, List<Expression> displayCols,
...
-            for (int ii=startKey+1; ii < displayCols.size(); ++ii)
-            {
-                Expression otherCol = displayCols.get(ii);
-                // This mechanism of finding the expression that a SIMPLE_COLUMN
-                // is referring to is inherently fragile---columnIndex is an
-                // offset into different things depending on context!
-                if (otherCol != null && (otherCol.opType != OpTypes.SIMPLE_COLUMN) &&
-                         (otherCol.columnIndex == this.columnIndex)  &&
-                         !(otherCol instanceof ExpressionColumn))
+            if (displayCols == null) {
```
wolffcm: I wonder if you can explain *why* displayCols is null for this case?

wolffcm:
Looks good to me.  If you can come up with an example that hits the new error message you added, consider adding a test case for it.  But I'm okay with you merging this change as-is assuming Jenkins passes.

The handling of SIMPLE_COLUMN in HSQL code disturbs me.  I wish I understood it better.

jxchinaren:
I wish I can understand it well enough either for the SIMPLE_COLUMN. It turns out the hsqldb also confuses it by himself, see hsqldb backend bug: ENG-8273

PR #2335, opened 2015-05-14
Add Junit tests to verify it works for :Scalar subquery correlative b…

jxchinaren:
…y GROUP BY column

Find hsqldb backend bug~

### DIFF
--- a/tests/frontend/org/voltdb/planner/TestPlansScalarSubQueries.java
+++ b/tests/frontend/org/voltdb/planner/TestPlansScalarSubQueries.java
@@ -28,11 +28,13 @@ import java.util.List;
 import org.voltdb.expressions.AbstractExpression;
 import org.voltdb.expressions.AbstractSubqueryExpression;
 import org.voltdb.expressions.ComparisonExpression;
+import org.voltdb.expressions.ScalarValueExpression;
 import org.voltdb.plannodes.AbstractPlanNode;
 import org.voltdb.plannodes.AbstractScanPlanNode;
 import org.voltdb.plannodes.AggregatePlanNode;
 import org.voltdb.plannodes.IndexScanPlanNode;
 import org.voltdb.plannodes.NodeSchema;
+import org.voltdb.plannodes.ProjectionPlanNode;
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/planner/TestPlansScalarSubQueries.java
```diff
@@ -72,6 +74,25 @@ public void testSelectCorrelatedScalar() {
...
+
+        AbstractPlanNode pn = compile(sql);
+        pn = pn.getChild(0);
+        assertTrue(pn instanceof ProjectionPlanNode);
+        NodeSchema schema = pn.getOutputSchema();
+        assertEquals(3, schema.size());
+        SchemaColumn col = schema.getColumns().get(2);
+        assertTrue(col != null);
+        assertEquals("STORE_CATEGORY", col.getColumnName());
+        assert(col.getExpression() instanceof ScalarValueExpression);
```
jxchinaren: It's better to use "assertTrue" here, will fix it in a new pull request for ENG-8263.

tests/frontend/org/voltdb/planner/testplans-subqueries-ddl.sql
```diff
@@ -98,3 +98,25 @@ COUNT(*) heat
...
+    number integer
+);
+
+create table store_types (
+    category varchar(50),
+    name varchar(50),
+    type_id integer
+);
+
+CREATE PROCEDURE FRANCHISES_BY_NAMED_CATEGORY AS
```
wolffcm: This stored proc seems unnecessary?

wolffcm:
Looks good.

tests/frontend/org/voltdb/planner/testplans-subqueries-ddl.sql
jxchinaren: Because it's the repro case in the ticket, I just copy it here. It should not make a difference from AdHoc query, that's why I change the alias on the AdHoc display columns.

PR #2338, opened 2015-05-14
Eng 8263 groupby subquery

jxchinaren:


src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1828,10 +1827,11 @@ else if (display_col.columnName != null && ! display_col.columnName.equals(""))
         if (m_having != null) {
             exprs.addAll(m_having.findAllSubexpressionsOfType(exprType));
         }
-        if (m_groupByExpressions != null) {
-            for (AbstractExpression groupByExpr : m_groupByExpressions.values()) {
-                exprs.addAll(groupByExpr.findAllSubexpressionsOfType(exprType));
-            }
+        // m_groupByExpressions is replaced all complex expression with TVE already
+        for (ParsedColInfo groupbyCol: m_groupByColumns) {
```
wolffcm: Reading this code, I have lots of questions.

What is the difference between the m_groupByExpressions and m_groupByColumns?  

Does m_groupByExpressions become invalid at some point during planning at which point we need to start using m_groupByColumns instead?  

When do the complex expressions in m_groupByExpressions get replaced?

Are the values in m_groupByExpressions always TVEs, in which case maybe you can change the value type of m_groupByExpressions to TVE?

What if I'm searching for all subexpressions of type TVE, will your change still let me find all of them?

Since these two members are easily confused, would it be good to some refactoring to address this?  If it's too much work to do right now, maybe file a ticket.

wolffcm:
End of review round.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
jxchinaren: m_groupByExpressions are there because of the complex GROUP BY support, as we used to have ONLY support GROUP BY columns. 

All the magics are in function placeTVEsinColumns ().

paulmartel: When Chris asks lots of questions like this, it's usually a good idea to ask "Why does he have to ask? Why is it not more obvious?". Maybe then you will not have to answer the same question for Bill and the new intern, and the one after that, etc. I think the answer to that question (tyopical) is "Because there are no comments where the data member is defined." THAT is what should be fixed in response to Chris' review comment.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -1048,6 +1048,15 @@ public void testSelectScalarSubSelects() throws Exception
             validateTableOfLongs(vt, new long[][] {  {1,3,6}, {2, 2, 5} });
         }
 
+        // ENG-8263: group by scalar value expression
+        vt = client.callProcedure("@AdHoc", "select R1.DEPT, count(*) FROM R1 "
+                + "GROUP BY dept, (SELECT count(dept) FROM R2 where R2.wage = R1.wage) order by dept;").getResults()[0];
+        validateTableOfLongs(vt, new long[][] {  {1,3}, {2, 2} });
```
paulmartel: This is a good planner test but a weak executor test because the subquery result has no effect on the output -- even if the subquery gave completely wrong answers, this query would give a correct result because of the small sample data -- there's no dept that has multiple rows with different subquery results.  Using loadData(true) to get the "extra rows" might fix that.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -1048,6 +1048,15 @@ public void testSelectScalarSubSelects() throws Exception
...
+
+        vt = client.callProcedure("@AdHoc", "select R1.DEPT, count(*) FROM R1 "
+                + "GROUP BY dept, (SELECT count(dept) FROM R2 where R2.wage > 15) order by dept;").getResults()[0];
```
paulmartel: Is this an intentionally silly case where we are grouping by a calculated constant?

A test for non-correlated subquery in group by would be less silly if it was part of a larger expression that referenced a parent column OUTSIDE the subquery like:
    GROUP BY dept, ABS(wage - (select max(wage) from R2))

I like ABS with a subtraction because it potentially gives different (fewer) groups than just subtraction or other math ops, giving the subquery a purpose and making it even less silly.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
jxchinaren: @paulmartel  I also thought about using "loadData(true)", but it took more time to fix the other queries results. I had a fever yesterday afternoon, I guess I paid less attention on the data, hoping SQLCoverage test can cover it.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
wolffcm: Sorry for asking so many questions.  Looking at placeTVEsinColumns, it makes a little more sense to me now.  Yes, comments on the variables would help.

It seems like there's a root cause to this bug that we're not addressing: there's two member variables that are capturing overlapping information.  Is there a way to coalesce them to avoid confusion in the future?

jxchinaren:
@wolffcm @paulmartel 
I add more comments and tests to address the comments.

wolffcm:
Thanks for adding the comments, Xin.  Obviously there's a bunch that I still don't understand about the planner.  The changes look good to me.

jxchinaren:
Close this pull request, squash commit and fix the other issue ENG-8296 together. ENG-8296 seems to be a serious problem either.

PR #2340, opened 2015-05-15
Add tests to ensure we don't use indexes with subquery exprs

wolffcm:


jxchinaren:
The tests looks good to me.

PR #2345, opened 2015-05-18
Add the guards for the HAVING with subquery

jxchinaren:


src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -817,6 +817,10 @@ private void parseHavingExpression(VoltXMLElement havingNode, boolean isDistribu
         assert(havingNode.children.size() == 1);
         m_having = parseExpressionTree(havingNode.children.get(0));
         assert(m_having != null);
+        if (! m_having.findAllSubexpressionsOfClass(SelectSubqueryExpression.class).isEmpty()) {
+            throw new PlanningErrorException(
+                    "SQL HAVING with subquery expression is not supported currently.");
```
paulmartel: We do not say "not supported currently". We say "not allowed".

paulmartel: The fix looks good other than the message text.

PR #2350, opened 2015-05-19
Squash commit the fix for group by scalar value subquery, add more tests

jxchinaren:
1. Added some fixes (hashCode, equals) trying to dedupe the duplicates expression. 

2. Add the guards not to support subquery inside of aggregation function.

3. Add the fix to be able to group by scalar value expression.

### DIFF
--- a/src/frontend/org/voltdb/expressions/ScalarValueExpression.java
+++ b/src/frontend/org/voltdb/expressions/ScalarValueExpression.java
@@ -37,7 +37,9 @@ public class ScalarValueExpression extends AbstractValueExpression {
     @Override
     public boolean equals(Object obj) {
         assert(m_left != null);
-        return m_left.equals(obj);
+        if (obj instanceof ScalarValueExpression == false) return false;
+        ScalarValueExpression expr = (ScalarValueExpression) obj;
+        return m_left.equals(expr.getLeft());
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/expressions/ScalarValueExpression.java
```diff
@@ -41,6 +41,11 @@ public boolean equals(Object obj) {
     }
 
     @Override
+    public int hashCode() {
```
paulmartel: The equals method looks broken, too -- the ScalarValueExpression is equal to an object if its CHILD  is equal to that object ?-- shouldn't we be requiring that the object is a ScalarValueExpression with a CHILD that equals this expression's child? That would at least seem to be traditional/consistent.

paulmartel:
Finished the review round. I'm having doubts that our test cases prove that two identical ScalarSubqueryExpressions will compare as equal.

src/frontend/org/voltdb/expressions/ScalarValueExpression.java
jxchinaren: Yes, it's broken. I realize that while I want to add support of subquery in aggregation functions. 

Still not sure how much it will be fixed. It seems the newly added HashSet is the ONLY place to use the "equals" and "hashCode" function, right?

PR #2360, opened 2015-05-20
ENG-8044: Fix catalog serialization race during initialization.

nshi:
Serialize the catalog once and pass the serialized string to each site
for initialization. Catalog serialization is not thread-safe.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -784,11 +784,13 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback {
              * Configure and start all the IV2 sites
              */
             try {
+                final String serializedCatalog = m_catalogContext.catalog.serialize();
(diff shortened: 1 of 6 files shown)


paulmartel:
This change looks good. Sorry for the regression. Also, sorry that I lost track of this pull request until now -- especially after advocating for it as a 5.3-critical issue.
In review, I was reminded that the original code in Site.java is a little hard to follow -- with two internal initialize methods (initialize and initializeEE) neither of which is private, and with the EE (in the Mock and HSQL cases) being initialized in initialize rather than in initializeEE, with initializeEE returning a new value for m_ee to initialize when it could just set m_ee itself, and with a little additional initializing done inline in the run method (including the cleanup of the catalog reference only used in initializeEE).  So, longer term, Site.java could use a little cleanup.

nshi:
Thanks for the review, @paulmartel . I'll log a ticket for cleanup the initialization methods in Site.java.

PR #2363, opened 2015-05-20
advanced-scalar-subquery.sql: commented out the 'A13' query, since it…

johnTcrawford:
… was causing mismatches (ENG-8292) - now for 5.3.


paulmartel:
Looks good.

PR #2364, opened 2015-05-20
ENG 6131 Count Plan Optimization

yzhang1991:
Now ```COUNT(<constant>)```, ```COUNT(<non-nullable column>)``` will share the same query plan with ```COUNT(*)```, which is index count.
```COUNT(<non-nullable column>)``` with "DISTINCT" keyword will not get a index count plan.

TestReplaceWithIndexCounter is renamed to TestPlansCount for general count plan testing.

### CODE REVIEW

File: tests/frontend/org/voltdb/planner/TestPlansCount.java [hunk 1]
```diff
@@ -179,7 +255,12 @@ public void testCountStar19() {
         List<AbstractPlanNode> pn = compileToFragments("SELECT count(*) from T2 WHERE USERNAME ='XIN' AND POINTS >= 3 AND POINTS <= 600000000000000000000000000");
         AbstractPlanNode p = pn.get(0).getChild(0);
         assertTrue((p instanceof IndexCountPlanNode) == false);
+
+        pn = compileToFragments("SELECT count(*) from T2 WHERE USERNAME ='XIN' AND POINTS >= 3 AND POINTS <= 600000000000000000000000000");
```
paulmartel: typo - still testing count(*) not count(1).

tests/frontend/org/voltdb/planner/TestPlansCount.java
```diff
@@ -179,7 +255,12 @@ public void testCountStar19() {
...
+        p = pn.get(0).getChild(0);
+        assertTrue((p instanceof IndexCountPlanNode) == false);
```
paulmartel: Please replace all cases new and old of "assertTrue(condition == false);" with "assertFalse(condition);"

tests/frontend/org/voltdb/planner/TestPlansCount.java
```diff
@@ -239,43 +355,115 @@ public void testCountStar26() {
...
+    }
 
+    // Testing the count( distinct <nullable column>) on a replicated table.
+    public void testCountColumn08() {
+        List<AbstractPlanNode> pn = compileToFragments("SELECT count(distinct AGE) from T1 WHERE ID=1 AND AGE >= ?");
+        checkIndexCounter(pn, false);
+    }
     /**
      * Check Whether or not the original plan is replaced with CountingIndexPlanNode.
      *
```
paulmartel: Please rename the variable below from isReplaceable to optimizedCount.

tests/frontend/org/voltdb/planner/testplans-count-ddl.sql
```diff
@@ -37,4 +41,4 @@ CREATE TABLE T3 (
     T3_NUM2 INTEGER NOT NULL,
     PRIMARY KEY (T3_ID)
 );
-CREATE INDEX T3_IDX_T3NUM_TREE ON T3 (T3_NUM1,T3_NUM2);
+CREATE INDEX T3_IDX_T3NUM_TREE ON T3 (T3_NUM1,T3_NUM2);
```
paulmartel: Please edit this file to restore the final newline.

tests/frontend/org/voltdb/planner/TestPlansCount.java
```diff
@@ -239,43 +355,115 @@ public void testCountStar26() {
...
+    public void testCountColumn07() {
+        List<AbstractPlanNode> pn = compileToFragments("SELECT count(AGE) from T1 WHERE ID=1 AND AGE >= ?");
+        checkIndexCounter(pn, false);
+    }
 
+    // Testing the count( distinct <nullable column>) on a replicated table.
+    public void testCountColumn08() {
+        List<AbstractPlanNode> pn = compileToFragments("SELECT count(distinct AGE) from T1 WHERE ID=1 AND AGE >= ?");
+        checkIndexCounter(pn, false);
+    }
```
paulmartel: What is the behavior if the user tries a query like "select count( distinct 1 ) ..."?
Please add a test that demonstrates that this gets an exception OR at least that it does not get rewritten as count(*). The best way to test for an exception in junit looks like this:

   try {
        <execute function expected to fail>;
        fail("<describe function call> unexpectedly returned without exception");
    } catch (<specific-exception-class> ex) {
        assertTrue("The exception from <describe function call> did not have the expected message",
                ex.getMessage().contains("<text specific to the condition being tested>");
    }

paulmartel:
End of review round. After making the recommended changes, we should squash this branch's changes onto a different branch before merging to master.

tests/frontend/org/voltdb/planner/TestPlansCount.java
yzhang1991: Fixed.

tests/frontend/org/voltdb/planner/TestPlansCount.java
yzhang1991: Fixed

tests/frontend/org/voltdb/planner/TestPlansCount.java
yzhang1991: Fixed

tests/frontend/org/voltdb/planner/testplans-count-ddl.sql
yzhang1991: Fixed

tests/frontend/org/voltdb/planner/TestPlansCount.java
yzhang1991: Now it will not merge into ```COUNT(*)```
I don't think we need to throw an exception for this because ```COUNT(DISTINCT <constant>)``` can correctly yield the result "1".

PR #2375, opened 2015-05-27
Eng 8294 union subquery

johnTcrawford:
Added new test suite, advanced-scalar-set-subquery, testing scalar subqueries containing set operators (UNION, INTERSECT, EXCEPT [ALL]), per ENG-8294.


tests/scripts/examples/sql_coverage/advanced-scalar-set-subquery.sql
```diff
@@ -0,0 +1,49 @@
...
+
+--- Queries with scalar subqueries containing a set subquery, in the SELECT clause
+SELECT @idcol ID1, (_scalarquerystr) FROM @fromtables AS Z
+SELECT @idcol ID2, (_scalarquerynum) FROM @fromtables AS Z
+
+--- Queries with scalar subqueries containing a set subquery, in the WHERE clause
+SELECT @idcol ID3,                        __[#agg] FROM @fromtables Z WHERE __[#agg] _cmp2 (_scalarquerystr)
+SELECT @idcol ID4, _variable[#num @comparabletype] FROM @fromtables Z WHERE __[#num] _cmp1 (_scalarquerycount)
+SELECT @idcol ID5, _variable[#num @comparabletype] FROM @fromtables Z WHERE __[#num] _cmp3 (_scalarquerynum)
+
```
paulmartel: These WHERE clause uses of scalar setop subquery are surprisingly uniform -- the setop is always nested one level down in the FROM clause of yet another query level.
Arguably, these statements are not exercising at all the syntactic support for setop subqueries (directly) in the where clause, but only in the FROM clause.
What about coverage for other more direct patterns? 

    WHERE __[#num] _cmp3 (_setopquerynum)

Getting useful results (exercising cases other than the runtime exception for getting multiple rows from the setop) could be a little bit of a challenge -- I'd keep tables small, maybe keep some empty, maybe use "INSERT INTO SELECT FROM" to force one-row overlaps, one row differences, etc. between tables. When (if?) Xin has merged ORDER BY ... LIMIT 1 support for set ops, this is another easy way to prevent mostly multi-row return errors.

Speaking of multi-row returns -- I've lost track -- does set op subquery support within IN and EXISTS clauses have a suite? Or is there a ticket for it? That's easily as useful/interesting as these cases.

paulmartel:
End of review round.

PR #2382, opened 2015-05-29
ENG-6950: Use regular file IO for persistent binary deque

nshi:


src/frontend/org/voltcore/utils/DBBPool.java.template
```diff
@@ -353,31 +353,48 @@ public final class DBBPool {
...
-    /*
-     * Allocate a DirectByteBuffer from a global lock free pool
+    /**
+     * Find the closest power of 2 that's larger than or equal to the requested capacity.
+     * @return 0 if the requested capacity is 0, the requested capacity itself if the
+     * next power of 2 overflows, or the next power of 2.
+     */
+    static int roundToClosestPowerOf2(int capacity) {
+        if (capacity == 0) return 0;
+        else if (capacity == 1) return 2;
```
akhanzode: why special case 1 is it common size requested? Can this be in some utility?

src/frontend/org/voltcore/utils/DBBPool.java.template
```diff
@@ -353,31 +353,48 @@ public final class DBBPool {
...
+    }
+
+    /**
+     * Allocate a DirectByteBuffer from a global lock free pool. The allocated buffer may
+     * have a capacity larger than the requested size. The limit will be set to the requested
+     * size.
      */
     public static BBContainer allocateDirectAndPool(final Integer capacity) {
-        ConcurrentLinkedQueue<BBContainer> pooledBuffers = m_pooledBuffers.get(capacity);
+        final int bucket = roundToClosestPowerOf2(capacity);
```
akhanzode: This will affect snapshot restore as diff capacity pool will end up waiting for constant capacity. Is this what we want?

src/frontend/org/voltdb/utils/PBDMMapSegment.java
```diff
@@ -0,0 +1,372 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2015 VoltDB Inc.
+ *
```
akhanzode: I am assuming this is same as current PBD segment.

src/frontend/org/voltcore/utils/DBBPool.java.template
nshi: The reason 1 is special cased here is because the formula below assumes the input is `> 1`.

src/frontend/org/voltcore/utils/DBBPool.java.template
nshi: Snapshot uses fixed 2MB buffers, so it's always using the same pool.

It's actually worse to use different sizes all the time and pool them. You'll end up pooling a lot of unused memory. Hence the rounding to the closest power of two.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
nshi: Yes, this is exactly the same as before.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
```diff
@@ -0,0 +1,372 @@
...
+        m_haveMAdvised = false;
+        m_objectReadIndex = 0;
+        m_bytesRead = 0;
+        m_discardCount = 0;
+    }
+
+    @Override
+    public int getNumEntries() throws IOException {
+        if (m_closed) {
+            open(false);
```
akhanzode: why open when its closed.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
```diff
@@ -0,0 +1,372 @@
...
+        m_closed = false;
+    }
+
+    @Override
+    public void closeAndDelete() throws IOException {
+        close();
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Deleting segment at Index " + m_index + " File: " + m_file.getAbsolutePath());
+        }
+        m_file.delete();
```
akhanzode: will it help to track delete state and throw already deleted. does m_file becomes null?

src/frontend/org/voltdb/utils/PBDMMapSegment.java
```diff
@@ -0,0 +1,372 @@
...
+    @Override
+    public boolean isClosed() {
+        return m_closed;
+    }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            if (m_fc != null) {
+                m_fc.close();
```
akhanzode: if close fails m_buf will not get discarded.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
nshi: One of the changes I made to the `PersistentBinaryDeque` is to close files when they are not in use, so that it doesn't keep all the files open if you have 100 segments. When a segment is ready to be read, it will reopen it.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
```diff
@@ -0,0 +1,372 @@
...
+        final long mBufAddr = m_buf.address();
+        if (!m_haveMAdvised) {
+            final ByteBuffer mbuf = m_buf.b();
+            m_haveMAdvised = true;
+            final long retval = PosixAdvise.madvise(
+                    m_buf.address(),
+                    mbuf.position(),
+                    PosixAdvise.POSIX_MADV_WILLNEED);
+            if (retval != 0) {
+                LOG.warn("madvise will need failed: " + retval);
```
akhanzode: better message maybe? will need fail does not make sense.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
nshi: Good point. I'll have `open()` throw an `IOException` if the file is deleted.

src/frontend/org/voltdb/utils/PBDMMapSegment.java
afrogers: Blindly turning Posix flag names into sentences != semantic sense. Should read something more like madvise(MADV_WILLNEED)

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+        m_readOffset = SEGMENT_HEADER_BYTES;
+        m_discardCount = 0;
+        m_tmpHeaderBuf.clear();
+    }
+
+    @Override
+    public int getNumEntries() throws IOException
+    {
+        // TODO: cache the result
+        if (m_closed) {
```
akhanzode: segment.open followed by getNumEntries would that cause double open?

src/frontend/org/voltdb/utils/PBDMMapSegment.java
nshi: `PBDMMapSegment` is the one currently being used, but renamed. I'm less inclined to fix the existing issues in there because it's not gonna be used after this patch.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+    {
+        return m_closed;
+    }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            if (m_fc != null) {
+                m_fc.close();
+                m_ras = null;
```
akhanzode: move to finally?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+                if (read == -1) {
+                    throw new EOFException();
+                }
+            }
+            m_tmpHeaderBuf.flip();
+            final int length = m_tmpHeaderBuf.getInt();
+            final int flags = m_tmpHeaderBuf.getInt();
+            final boolean compressed = (flags & FLAG_COMPRESSED) != 0;
+
+            if (length < 1) {
```
akhanzode: do we write 0 length?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+        } finally {
+            m_readOffset = m_fc.position();
+            m_fc.position(writePos);
+        }
+    }
+
+    //A white lie, reflects actual disk usage, even if buffers are not taking up that much space.
+    @Override
+    public int sizeInBytes() {
+        return (int) (m_file.length() - SEGMENT_HEADER_BYTES);
```
afrogers: I don't think this is going to work for DR stats, where we call this on rejoin. The version in the mmap segment counts actual (uncompressed, not that it matters to us) bytes of buffer data, sans header bytes.

afrogers: As another note (just checked Jenkins about this), export stats test is timing out, so the DR unit tests aren't running on this branch yet.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+        m_objectReadIndex = 0;
+        m_readOffset = SEGMENT_HEADER_BYTES;
+        m_discardCount = 0;
+        m_tmpHeaderBuf.clear();
+    }
+
+    @Override
+    public int getNumEntries() throws IOException
+    {
+        // TODO: cache the result
```
afrogers: Not saying to have to resolve this now, but agree on the TODO. Weird that the segments don't store any info about themselves in memory.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+        if (m_closed) {
+            open(false);
+        }
+        if (m_fc.size() > 0) {
+            m_tmpHeaderBuf.clear();
```
afrogers: Small DRY complaint: seems like we're always going to clear the dest buffer in readBufferFully, which seems like a reasonable generalization.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+
+    @Override
+    public void incrementNumEntries(int size) throws IOException
+    {
+        //First read the existing amount
+        m_tmpHeaderBuf.clear();
+        PBDUtils.readBufferFully(m_fc, m_tmpHeaderBuf, COUNT_OFFSET);
+
+        //Then write the incremented value
+        final int numEntries = m_tmpHeaderBuf.getInt();
```
afrogers: Especially useful to cache this in memory, since that would spare us the read here. Segment should already know this information.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
nshi: `readBufferFully` doesn't clear the buffer, it just writes to it until it's full. Clearing here to make sure that the position is set to 0.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -0,0 +1,390 @@
...
+                while (retcont.b().hasRemaining()) {
+                    int read = m_fc.read(retcont.b());
+                    if (read == -1) {
+                        throw new EOFException();
+                    }
+                }
+                retcont.b().flip();
+            }
+
+            return new DBBPool.BBContainer(retcont.b()) {
```
afrogers: Really feels like a lot of this shared logic (this anonymous container, for example) could be moved up to a base PBDSegment that's an abstract class instead of an interface. Not a big deal, but would be nice to be able to read the implementation classes more for how they differ.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -228,6 +228,15 @@ public boolean accept(File pathname) {
         assertions();
     }
 
+    private static final boolean USE_MMAP = Boolean.getBoolean("PBD_USE_MMAP");
```
afrogers: After talking to Anish and Stefano about it, the preference from the export perspective is a default of true here. Maybe a better in-person discussion, though.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -254,8 +263,7 @@ public synchronized void offer(BBContainer object, boolean allowCompression) thr
 
         PBDSegment tail = m_segments.peekLast();
         //If we are mostly empty, don't do compression, otherwise compress to reduce space and IO
```
afrogers: Comment no longer applies after fullness heuristic went away

afrogers:
Review pass complete

nshi:
Squash merged externally.

r=arogers

PR #2385, opened 2015-05-29
Eng 8299 no exclude join conditions

BillWhite:
In this commit I do three things.

1. I make it possible to dump SQL statements, HSQLDB ASTs and VoltXML ASTs using the standard Volt logging mechanism.
1. I fix ENG-8299, which caused expressions to seem to ignore the right hand operand of binary operators.
1. I fix ENG-8262, which caused hangs when assertion failures in the PlannerTool circumvented the return of responses from the server to the client.  This needs to be merged to master sooner rather than later.


BillWhite:
Thanks.

PR #2388, opened 2015-06-01
ENG-8313 fixed buffer corruption bug

dweiss88:


akhanzode:
Looks good. Can you make same fix in voltdb-client-cpp repo?

PR #2390, opened 2015-06-02
fixes run.sh demo

benjaminballard:
Several customers have reported problems running the voter demo.  It gets stuck on "... Waiting for VoltDB to start".  It stops printing this because it has actually succeeded, but sqlcmd is waiting for entry or somehow not exiting properly when invoked using the --query parameter and redirecting output.  When it is invoked using STDIN, it exits as expected.  This behavior is only seen on a Mac, on Linux (even old versions) it works as expected.

I filed a ticket for sqlcmd to exit properly in this scenario, but this minor change to the run.sh script fixes the demo on a Mac.  I will be making the same changes to the separate app-* demos on github.

### DIFF
--- a/examples/uniquedevices/run.sh
+++ b/examples/uniquedevices/run.sh
@@ -106,7 +106,7 @@ function init() {
 
 # wait for backgrounded server to start up
 function wait_for_startup() {
-    until sqlcmd  --query=' exec @SystemInformation, OVERVIEW;' > /dev/null 2>&1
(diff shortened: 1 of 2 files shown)


jhugg:
Eeek. I guess so.

akhanzode:
looks fine to me.

benjaminballard:
I've minimized the change at Phil's request.

After further testing, this is repeatable on Linux as well as Mac.

PR #2393, opened 2015-06-03
Notest eng 8181 socketimporter test

pdshaw:
Changes to the Java socket importer client to check that data in the import table matches the data sent to VoltDB via one or more socket importers.


akhanzode:
I am good with java changes.

PR #2399, opened 2015-06-05
ENG-8423: Update cached count and size in PBDSegment after truncation.

nshi:


src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -570,6 +570,7 @@ public synchronized void parseAndTruncate(BinaryDequeTruncator truncator) throws
                                         while (numObjectsBuffer.hasRemaining()) {
                                             fc.write(numObjectsBuffer);
                                         }
+                                        segment.updateCountAndSize(ii, sizeInBytes);
```
afrogers: Since the logic above this call (and the next one) is for writing the count and size into the backing file, would it make sense to move that logic into the segments, too?

Maybe not, since it would be duplicated, and you'd have to pass in the fc. Maybe just renamed this to updateCachedCountAndSize()? Otherwise, kind of misleading, since it only updates the cached copies.

afrogers:
This looks good, other than the above comment.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
nshi: Makes sense, I'll rename this method.

My thinking is that I'll push the whole truncation logic, including the size update, down to the `PBDSegment` on master. That way the bit mangling is all encapsulated in one place.

PR #2402, opened 2015-06-08
Change column name describing segments in use

slowsoul:
Change column name from LOANED_SEGMENT_COUNT to IN_USE_SEGMENT_COUNT so it is more user-friendly.


nshi:
Looks good. I'll merge it once tests pass in Jenkins.

PR #2408, opened 2015-06-09
Eng 8343 round decimal scale

BillWhite:
When we see a string which has more scale than we support, we will now round it up to the
closest supported value.  We round a value equidistant between two supported values up.
This is all done using BigDecimal's round function and the ROUND_UP rounding mode.

I had to make some inessential changes in a C++ file and a java test file.  Apparently
the C++ compiler on my macbook thinks "this" can never be null, and it is offended that 
someone would test for it.  I had to work around that.  Also, Eclipse was unhappy about
a test, so I had to cast an Object to be Comparable.  The test was testing that such a
cast would cause a ClassCastException.  This is really kind of a silly test, and tests
the Java compiler more than the VoltDB code.

### DIFF
--- a/src/ee/expressions/abstractexpression.cpp
+++ b/src/ee/expressions/abstractexpression.cpp
@@ -105,7 +105,9 @@ AbstractExpression::initParamShortCircuits()
 std::string
 AbstractExpression::debug() const
 {
-    if (this == NULL) {
+    // Subvert a g++ error message;
+    const AbstractExpression *thisProxy = this;
+    if (thisProxy == NULL) {
         return "NULL";
     }
     std::ostringstream buffer;
@@ -117,7 +119,9 @@ AbstractExpression::debug() const
 std::string
 AbstractExpression::debug(bool traverse) const
 {
-    if (this == NULL) {
+    // Subvert a g++ error message;
+    const AbstractExpression *thisProxy = this;
+    if (thisProxy == NULL) {
         return "NULL";
     }
     return (traverse ? debug(std::string("")) : debug());
@@ -126,7 +130,9 @@ AbstractExpression::debug(bool traverse) const
 std::string
 AbstractExpression::debug(const std::string &spacer) const
 {
-    if (this == NULL) {
+    // Subvert a g++ error message;
+    const AbstractExpression *thisProxy = this;
+    if (thisProxy == NULL) {
         return "NULL";
     }
     std::ostringstream buffer;
--- a/tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
@@ -24,6 +24,7 @@
 package org.voltdb.regressionsuites;
 
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.net.ConnectException;
 import java.nio.channels.SocketChannel;
 import java.util.ArrayList;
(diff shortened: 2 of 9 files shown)


src/ee/expressions/abstractexpression.cpp
```diff
@@ -105,7 +105,9 @@ AbstractExpression::initParamShortCircuits()
 std::string
 AbstractExpression::debug() const
 {
-    if (this == NULL) {
+    // Subvert a g++ error message;
```
jxchinaren: It should be a warning message in the compiler and we treat the warning as the error.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -386,6 +394,43 @@ static public void validateTableColumnOfScalarVarchar(VoltTable vt, int col, Str
...
+        assertTrue(vt.advanceRow());
+        for (int i=0; i < len; i++) {
+            BigDecimal actual = null;
+            try {
+                actual = vt.getDecimalAsBigDecimal(i);
+            } catch (IllegalArgumentException ex) {
+                ex.printStackTrace();
+                fail();
+            }
+            // Long.MIN_VALUE is like a NULL
```
jxchinaren: This comment is not TRUE for decimal. right?

tests/frontend/org/voltdb/regressionsuites/TestSqlInsertSuite.java
```diff
@@ -73,6 +82,33 @@ public void testInsert() throws Exception
...
+        // Do the same as the last time, but make the last digit equal to 4.
+        // This should round down.
+        validateInsertStmt("insert into decimaltable values 0.9999999999994000;",
+                           new BigDecimal("0.999999999999"));
+        try {
+            validateInsertStmt("insert into decimaltable values 999999999999999999999999999999999999999999.0;",
+                               (BigDecimal)null);
+            fail("Expected an exception here.");
+        } catch (Exception ex) {
+            assertTrue(true);
```
jxchinaren: It's better to verify the error message also here. Your "validateInsertStmt" contains various queries.

You can use the utility function "verifyStmtFails" from regression suite.

tests/frontend/org/voltdb/regressionsuites/TestSqlInsertSuite.java
```diff
@@ -73,6 +82,33 @@ public void testInsert() throws Exception
...
+        }
+        validateInsertStmt("insert into decimaltable values null;", (BigDecimal)null);
```
jxchinaren: Did you test any negative decimal values? Probably also interesting here.

jxchinaren:
Finish review round.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
BillWhite: Ok.  I'll delete it.  We seem to get java nulls for null decimal fields.  This is by testing,
and not by figuring out the design.

tests/frontend/org/voltdb/regressionsuites/TestSqlInsertSuite.java
BillWhite: Fixed.

tests/frontend/org/voltdb/regressionsuites/TestSqlInsertSuite.java
BillWhite: Well, I'm only interested in testing scale here, and scale is after the decimal point.  But I'll
add some negative tests as well.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
paulmartel: Nor by reading the doc?

paulmartel:
What's the expected behavior if I pass a 13-position decimal in a comparison either as a literal or as the value for a ? parameter. In other words, if I insert such a value and then immediately turn around and try to fetch the row that has that value, it seems that I should I be able to find it. There are interesting inequality test cases as well -- if I insert 0.9999999999993 I should not match "where decimal < 0.9999999999993 " OR  "where decimal < 0.9999999999994 ".
Also, inserts using too-precise BigDecimals passed into "?" parameters should be tested.
Also, the specific customer's preferred method of access -- JDBC -- should get some test coverage -- including ad hoc statement with literal decimal value and prepared statement with decimal passed as parameter -- I THINK JDBC uses BigDecimal as its native decimal parameter binding type.

PR #2409, opened 2015-06-09
Eng 8283 file export no type

wijjo:


akhanzode:
looks good squash merge when jenkins is blue.

PR #2410, opened 2015-06-09
ENG-6583 Add bin/rabbitmqloader

wijjo:


akhanzode:
Do u want to let Andrew know to doc this?

PR #2412, opened 2015-06-09
Eng 8190 sqlcov in exists

johnTcrawford:
Additional tests of IN, EXISTS, ANY and ALL, per ENG-8190.

### DIFF
--- a/tests/scripts/examples/sql_coverage/advanced-inexists.sql
+++ b/tests/scripts/examples/sql_coverage/advanced-inexists.sql
@@ -1,8 +1,67 @@
-<configure-for-ints.sql>
+<configure-default.sql>
+
+-- DML: purge and regenerate random data first
+DELETE FROM @dmltable
+INSERT INTO @dmltable VALUES (@insertvals)
+
+-- TEMP, for debugging, just so I can quickly see what data was generated:
+--SELECT * FROM @fromtables ORDER BY @idcol
+--SELECT SUM(ID), SUM(NUM), SUM(RATIO) FROM @fromtables
+
+--- Define "place-holders" used in the queries below
+{_optionaloffset |= ""}
+{_optionaloffset |= "OFFSET 2"}
+{_optionallimitoffset |= ""}
+{_optionallimitoffset |= "LIMIT 4 _optionaloffset"}
+
+{_optionalorderbycollimitoffset |= ""}
+{_optionalorderbycollimitoffset |= "LIMIT 1000"}
+{_optionalorderbycollimitoffset |= "ORDER BY __[#col] _sortorder _optionallimitoffset"}
+
+{_groupbycoloptionalhavingagg |= "GROUP BY __[#col]"}
+{_groupbycoloptionalhavingagg |= "GROUP BY __[#col] HAVING @agg(__[#agg]) _cmp A32.__[#agg]"}
+
+{_anyorall |= "ANY"}
+{_anyorall |= "ALL"}
+
+{_colexpr |= "1+__[#col]"}
+{_colexpr |= "__[#col]-1"}
+{_colexpr |= "__[#col]*2"}
+{_colexpr |= "__[#col]/2"}
 
 --- Test IN/EXISTS Advanced cases
-SELECT * FROM @fromtables A WHERE _variable[@columntype] _maybe IN ( SELECT MIN(_variable[@columntype]) FROM @fromtables B WHERE B._variable[@columntype] _cmp A._variable[@columntype] )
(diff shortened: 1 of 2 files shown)


tests/scripts/examples/sql_coverage/advanced-inexists.sql
```diff
@@ -1,8 +1,67 @@
...
+
+SELECT * FROM @fromtables A33 WHERE _maybe EXISTS (SELECT _variable[#col] FROM @fromtables _setop SELECT __[#col] FROM @fromtables)
+SELECT * FROM @fromtables A34 WHERE _maybe EXISTS (SELECT _variable[#col] FROM @fromtables _setop SELECT __[#col] FROM @fromtables SQ WHERE SQ.__[#col] _cmp A34.__[#col])
+
+SELECT * FROM @fromtables A35 WHERE _maybe EXISTS (SELECT A._variable[#col], B.__[#col] FROM @fromtables A, @fromtables B WHERE A.__[#col] = B.__[#col] AND A.__[#col] _cmp A35.__[#col])
+
+
+--- Additional tests of ANY and ALL (for ENG-8190)
+SELECT * FROM @fromtables A41 WHERE _variable[#col] _cmp _anyorall (SELECT __[#col] FROM @fromtables SQ WHERE SQ.__[#col] _cmp A41.__[#col])
+SELECT * FROM @fromtables A42 WHERE _variable[#col] _cmp _anyorall (SELECT _colexpr FROM @fromtables SQ WHERE SQ.__[#col] _cmp A42.__[#col])
```
paulmartel: Add a similar pattern with _colexpr on the left-hand-side of _cmp instead of in the right-hand-side select.

paulmartel: _colexpr is going to generate some number of invalid queries when you open this up to DDL.sql with its DESC VARCHAR column.
I suppose the work-around is to reference _variable[string] in the select clause to take DESC out of the running for other _variable references.

That would mean requiring a string column in any (other) schema that we (hypothetically) wanted to use with this template -- or else we'd generate 0 statements for that schema.

I suppose we could try to be clever about THAT by using up something more generic like _variable[@incompatible_type] with @incompatible_type variously set in the selectable configure*.sql files associated with the various schema (vs. hard-coded into the top of the template). The effects of choosing different actually in use column types for the "incompatible_type" in schemas other than DDL.sql when there is no obvious VARCHAR bad guy to take out might be "entertaining".

OR we could just live with the single-schema-specific scenario or live with higher invalid percentages or 0 statement counts with other schemas or use new schema variants as workarounds.

tests/scripts/examples/sql_coverage/advanced-inexists.sql
```diff
@@ -1,8 +1,67 @@
...
+SELECT * FROM @fromtables A31 WHERE _maybe EXISTS (SELECT _variable[#col]                        FROM @fromtables SQ WHERE SQ.__[#col] _cmp A31.__[#col] _optionalorderbycollimitoffset)
+SELECT * FROM @fromtables A32 WHERE _maybe EXISTS (SELECT _variable[#col], @agg(_variable[#agg]) FROM @fromtables _groupbycoloptionalhavingagg)
+
+SELECT * FROM @fromtables A33 WHERE _maybe EXISTS (SELECT _variable[#col] FROM @fromtables _setop SELECT __[#col] FROM @fromtables)
+SELECT * FROM @fromtables A34 WHERE _maybe EXISTS (SELECT _variable[#col] FROM @fromtables _setop SELECT __[#col] FROM @fromtables SQ WHERE SQ.__[#col] _cmp A34.__[#col])
+
+SELECT * FROM @fromtables A35 WHERE _maybe EXISTS (SELECT A._variable[#col], B.__[#col] FROM @fromtables A, @fromtables B WHERE A.__[#col] = B.__[#col] AND A.__[#col] _cmp A35.__[#col])
+
+
+--- Additional tests of ANY and ALL (for ENG-8190)
```
paulmartel: Is there long-term utility in referencing this soon-to-be-closed ticket, especially since that ticket covers so many different scenarios?

paulmartel: A comment pointing out the unique aspect of each statement or statement group would be more helpful.

paulmartel:
End of review round. Looks fairly complete except for the noted use of expression like _colexpr as the lhs operand which COULD conceivably exercise different code paths.

tests/scripts/examples/sql_coverage/advanced-inexists.sql
johnTcrawford: I assume you mean to use _colexpr on the LHS of _cmp in the subquery, not the outer query; that's what I'm doing, unless you say otherwise.

As for DESC, isn't it simpler to just define _variable[#col @comparabletype], for those queries that use _colexpr? That seems to work using DDL.sql or int-DDL.sql (since it's defined in configure-default.sql).

PR #2416, opened 2015-06-10
cherry pick Tong's contribution and add guards to fix various stuffs

jxchinaren:
SQL MOD function, for initial support of Integral types ONLY.


src/ee/expressions/numericfunctions.h
```diff
@@ -202,5 +202,34 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+    if (base.isNull() || divisor.isNull()) {
+        return getNullValue(VALUE_TYPE_BIGINT);
+    } else if (divisor.castAsBigIntAndGetValue() == 0) {
+        throw SQLException(SQLException::data_exception_division_by_zero, "division by zero");
+    }
+
+    NValue retval(divisorType);
+    int64_t baseValue = base.castAsBigIntAndGetValue();
+    int64_t divisorValue = divisor.castAsBigIntAndGetValue();
+    return getBigIntValue(baseValue % divisorValue);
```
wolffcm: What are the undefined behaviors in C/C++ regarding `%`?  We don't want to inherit them in our implementation of SQL.  Also, does the SQL standard specify the behavior for interesting cases?

Consider explicitly defining behavior for the following:
- divisor equal to zero
- either operand negative

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
```diff
@@ -826,13 +823,12 @@ Object getValue(Session session, Object[] data) {
...
-                    ((NumberType) nodes[0].dataType).divide(nodes[0],
-                        nodes[1]);
-
-                value = ((NumberType) nodes[0].dataType).subtract(nodes[0],
-                        value, nodes[1].dataType);
-
+                        ((NumberType) nodes[0].dataType).mod(data[0],
+                                data[1]);
+                // End of VoltDB extension
+
```
wolffcm: Extra blank line accidentally added outside of formatted comments.

src/ee/expressions/numericfunctions.h
wolffcm: Oops, I just saw you check for zero above.

src/hsqldb19b3/org/hsqldb_voltpatches/types/NumberType.java
```diff
@@ -1816,6 +1816,55 @@ public Object ceiling(Object a) {
...
+                 }
+
+                 return ValuePool.getLong(al % bl);
+             }
+             default :
+                 throw Error.runtimeError(ErrorCode.U_S0500, "NumberType");
+         }
+     }
+    // End of VoltDB extension
+
```
wolffcm: Extra blank line added outside of formatted comment.

src/ee/expressions/numericfunctions.h
wolffcm: I found this link on Stack Overflow:

http://stackoverflow.com/questions/7594508/modulo-operator-with-negative-values

It looks like having exactly one negative operand results in undefined behavior, meaning different C++ compilers could get different answers here.

wolffcm:
That's all my comments.

src/ee/expressions/numericfunctions.h
jxchinaren: C99: The % operator in C is not the modulo operator but the remainder operator.

Check this:
http://stackoverflow.com/questions/11720656/modulo-operation-with-negative-numbers

I do not find the SQL standard for it. However, for different vendors like MySQL, Oracle, SQL Server, the MOD functions is actually calculating the reminder.

See:
https://dev.mysql.com/doc/refman/5.0/en/mathematical-functions.html#function_mod
http://docs.oracle.com/cd/B19306_01/server.102/b14200/functions088.htm
https://msdn.microsoft.com/en-us/library/ee634767.aspx

jxchinaren: I will add this as the comments for MOD function and ask Andrew to document them.

jxchinaren:
@wolffcm @BillWhite I think the Fortran semantics is the same as C99.
https://gcc.gnu.org/onlinedocs/gfortran/MOD.html

So the current implementation is right, we should put these information as comments. Am I right?

BillWhite:
I think so.

wolffcm:
Sounds good to me.

wolffcm:
Looks good.

PR #2417, opened 2015-06-10
Promote correlation parameters for subqueries with set operators

wolffcm:


tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -1617,6 +1629,151 @@ public void testExistsSimplification() throws NoConnectionsException, IOExceptio
...
+                + "WHERE DESC > ("
+                + "    SELECT DESC "
+                + "    FROM R4 WHERE "
+                + "    DESC <> Z.DESC "
+                + "  INTERSECT ALL "
+                + "    SELECT DESC "
+                + "    FROM R4 "
+                + "    WHERE DESC = Z.DESC)")
+                .getResults()[0];
+        assertFalse(vt.advanceRow());
```
paulmartel: I would feel more confident with tests that signified success with something less generic than an empty result -- too prone to being "right for the wrong reason".

paulmartel:
Thinking ahead a little bit to the next bug, it would be nice if the param array had a bitmap of "previously set params" that had N bits set and the remaining bits cleared when we deserialized N user params and then set additional individual bits when correlation parameters were set -- and then asserted in PVE::eval that the nth bit was previously set.

tests/frontend/org/voltdb/planner/TestPlansScalarSubQueries.java
```diff
@@ -365,6 +366,62 @@ public void testWhereIndexRow() {
...
+            assertEquals(ExpressionType.VALUE_TUPLE, args.get(0).getExpressionType());
+
+            TupleValueExpression tve = (TupleValueExpression)args.get(0);
+            assertEquals("OUTER_TBL", tve.getTableAlias());
+            assertEquals("A", tve.getColumnName());
+
+            AbstractPlanNode subqPlanNode = selSubq.getSubqueryNode();
+            assertNotNull(subqPlanNode);
+            assertTrue(subqPlanNode instanceof UnionPlanNode);
+        }
```
paulmartel: UNIONs have a history/tendency to rely on their "first child" for certain things.

It would probably be worth checking the case where the first child has no correlation but the second child does.

It might also be worth testing a nested case -- a union with a correlated child intersect in it.

There should also be a case where the children use DIFFERENT parent columns -- ideally a partially overlapping pair of columns in each subselect -- to make sure that the params for the subquery account for the "union" of all of its child subselect's requirements.

paulmartel:
End of review round. Good work.

wolffcm:
@paulmartel I've added a few more planner and regression tests (actually modified the regression tests to produce verifiable results).

PR #2418, opened 2015-06-11
ENG-8370: Add TYPE column and draft implementation for SNAPSHOTSTATUS stats

slowsoul:
ENG-8370: Add a TYPE column to help users identify whether a snapshot is a truncation snapshot, a automatic snapshot or a manual snapshot, in the result of SNAPSHOTSTATUS statistics query.


src/frontend/org/voltdb/SnapshotStatus.java
```diff
@@ -17,19 +17,45 @@
...
 import org.voltdb.sysprocs.SnapshotRegistry.Snapshot;
 import org.voltdb.sysprocs.SnapshotRegistry.Snapshot.Table;
-import org.voltcore.utils.Pair;
 
 public class SnapshotStatus extends StatsSource {
 
+    enum SNAPSHOT_TYPE {
+        AUTO_SNAPSHOT,
+        MANUAL_SNAPSHOT,
+        TRUNCATION_SNAPSHOT
```
nshi: You can drop the `_SNAPSHOT` from the enums to make the names shorter in the final output of `@SnapshotStatus`.

src/frontend/org/voltdb/SnapshotDaemon.java
```diff
@@ -205,10 +205,11 @@
         m_path = null;
         m_prefixAndSeparator = null;
 
-
-
         // Register the snapshot status to the StatsAgent
         SnapshotStatus snapshotStatus = new SnapshotStatus();
+        snapshotStatus.setSnapshotPath(
+                VoltDB.instance().getCatalogContext().cluster.getLogconfig().get("log").getInternalsnapshotpath(),
+                VoltDB.instance().getCatalogContext().database.getSnapshotschedule().get("default").getPath());
```
nshi: My only concern over this block is that the catalog context may not be set up at the point the SnapshotDaemon is constructed. Getting the context from the global instance creates an implicit dependency on the ordering of the initialization code that may be error prone.

Since the only place SnapshotDaemon is constructed is in ClientInterface, you can pass the catalog context to the constructor directly.

PR #2420, opened 2015-06-11
This seems like a very simple change which will cure some of the hanging behavior.

BillWhite:
Catch AssertError exceptions in the planner, and return a proper
response.  Don't hang the client waiting for a response.

Note that this is not to be merged to master, but to ENG-8070-hsqldb232-upgrade.  The master version will come soon.

### DIFF
--- a/src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
+++ b/src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
@@ -372,6 +372,8 @@ public class AsyncCompilerAgent {
             }
             catch (Exception e) {
                 errorMsgs.add("Unexpected Ad Hoc Planning Error: " + e);
+            } catch (AssertionError ae) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
```diff
@@ -372,6 +372,8 @@ else if (work.userPartitionKey == null) {
             }
             catch (Exception e) {
                 errorMsgs.add("Unexpected Ad Hoc Planning Error: " + e);
+            } catch (AssertionError ae) {
+                errorMsgs.add("Assertion Error in Ad Hoc Planning: " + ae);
```
wolffcm: I guess that AssertionError is not a subclass of Exception.  

Looks good to me.

Somewhat unrelated, and I'm only mentioning it because I just thought of it: other vendors will sometimes emit a message encouraging users to report unexpected errors bug so we can fix the bug.  We might want to do the same.

PR #2424, opened 2015-06-11
Eng 6841 hold dr data until command log is durable

wweiss-voltdb:


src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -127,6 +127,194 @@ public String toString()
         }
     };
 
+    class SpDurabilityListener implements DurabilityListener {
```
nshi: I suggest separating this into its own class to keep `SpScheduler` as clean as possible since it's the core component of the transaction initiation system.

src/frontend/org/voltdb/CommandLog.java
```diff
@@ -79,7 +81,50 @@ public abstract void logIv2Fault(long writerHSId, Set<Long> survivorHSId,
...
+
+        /**
+         * Assign the listener that we will send SP UniqueId durability notifications to
+         */
+        public void setSpUniqueIdListener(DurableSpUniqueIdListener listener);
+
+        /**
+         * Assign the listener that we will send MP UniqueId durability notifications to
+         */
+        public void setMpUniqueIdListener(DurableMpUniqueIdListener listener);
```
nshi: I would also suggest combining these two into one interface for clarity.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -146,7 +154,7 @@ public static synchronized void pushDRBuffer(
             long lastSequenceNumber,
             long lastUniqueId,
             ByteBuffer buf) {
-        if (log.isTraceEnabled()) {
+        if (startSequenceNumber == lastUniqueId) {
```
nshi: I don't understand why this is equivalent to the original guard and how the sequence number would be equivalent to the unique ID.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -199,14 +195,30 @@ public void setMaxSeenTxnId(long maxSeenTxnId)
...
+            m_durabilityListener.createFirstCompletionCheck(m_cl.isSynchronous(), false);
+        }
+    }
+
+    public void setMpUniqueIdListener(DurableMpUniqueIdListener listener) {
+        m_durabilityListener.setMpUniqueIdListener(listener);
+        if (m_cl != null && !m_durabilityListener.completionCheckInitialized(true)) {
+            m_durabilityListener.createFirstCompletionCheck(m_cl.isSynchronous(), true);
+        }
+    }
```
nshi: When are these two methods being called. I'm confused about the outcome of these two methods. It seems that depending on which one you call first, you'll end up having totally different completion checks.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -157,35 +171,17 @@ public String toString()
...
-                        synchronized (m_lock) {
-                            for (Object o : durableThings) {
-                                m_pendingTasks.offer((TransactionTask)o);
-
-                                // Make sure all queued tasks for this MP txn are released
-                                if (!((TransactionTask) o).getTransactionState().isSinglePartition()) {
-                                    offerPendingMPTasks(((TransactionTask) o).getTxnId());
-                                }
-                            }
-                        }
```
nshi: If this synchronized block is extracted to a method in `SpScheduler`, then you don't have to pass the lock to the durability listener. Maybe that's cleaner?

src/frontend/org/voltdb/PartitionDRGateway.java
afrogers: Oh hey! An accidental commit moment. I remember this check -- it's not equivalent to anything, it's just meant to disable this code block.

src/frontend/org/voltdb/iv2/SpDurabilityListener.java
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        @Override
+        CompletionChecks startNewCheckList(int startSize) {
+            return new AsyncMpCompletionChecks(m_lastSpUniqueId, m_lastMpUniqueId);
+        }
+
+        @Override
+        void addTask(TransactionTask task) {
+            m_lastSpUniqueId = task.m_txnState.uniqueId;
```
nshi: This is deceiving If it is an MP task. It will overwrite the correct `m_lastSpUniqueId` you had.

src/frontend/org/voltdb/iv2/SpDurabilityListener.java
```diff
@@ -0,0 +1,201 @@
...
+                // Make sure all queued tasks for this MP txn are released
+                if (!o.getTransactionState().isSinglePartition()) {
+                    m_spScheduler.offerPendingMPTasks(o.getTxnId());
+                }
+            }
+            super.processChecks();
+        }
+    }
+
+    final ArrayDeque<CompletionChecks> m_writingTransactionLists =
```
nshi: I read the usage of this multiple time but am still failing to understand how it's used.

- Why is it initialized with a capacity of 4? Is it intended to have at most 4 things?
- Why does `onDurability()` only poll the first element out if there might be more than one?
- Why does the `SpSchedule` care about generic completion checks?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -157,35 +164,16 @@ public String toString()
...
-            }
-        };
+        m_durabilityListener = new SpDurabilityListener(this, m_pendingTasks);
         m_uniqueIdGenerator = new UniqueIdGenerator(partitionId, 0);
     }
 
     @Override
+    public void setLock(Object o) {
+        super.setLock(o);
+    }
```
nshi: This can be removed.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1051,6 +1051,23 @@ public CountDownLatch snapshotCompleted(SnapshotCompletionEvent event)
         return new CountDownLatch(0);
     }
 
+    public void processDurabilityChecks(final CompletionChecks currentChecks) {
+        final SiteTaskerRunnable r = new SiteTasker.SiteTaskerRunnable() {
+            @Override
+            void run() {
+                assert(currentChecks != null);
+                synchronized (m_lock) {
+                    currentChecks.processChecks();
```
nshi: Processing all the generic completion checks here means that they'll all run under the initiator lock on the site thread, which could block transaction processing.

afrogers:
It would be really convenient if RealVoltDB could have a method for registering a DurableUniqueIdListener. Even if it just delegated down through the initiator. Maybe we (I) can just do this separately as part of the consumer-side work, but it should probably be done.

PR #2425, opened 2015-06-12
Eng 8428 subquery row null prep

paulmartel:
No production code changes.
This is test-only change, and mostly cosmetic at that.
It adds extra diagnostics and diagnostic support methods to RegressionSuite
and leverages them to clean up TestSubQueriesSuite.

Most of the changes come from reformatting (re-whitespacing, keyword upper/lower casing) of sql query test for readability.

I tried to avoid semantic changes to test cases/queries, but there may have been a handful of silly cases that were just too tempting.

Jenkins passes.


jxchinaren:
Looks good to me.

paulmartel:
pushed via squash branch

PR #2426, opened 2015-06-12
fix the union partition table detection bug

jxchinaren:


src/frontend/org/voltdb/planner/ParsedUnionStmt.java
```diff
@@ -140,6 +141,10 @@ void parseTablesAndParams(VoltXMLElement stmtNode) {
             m_children.add(childStmt);
             // Add statement's tables to the consolidated list
             m_tableList.addAll(childStmt.m_tableList);
+
+            // m_tableAliasList is not interesting for UNION
+            // m_tableAliasMap may have same alias table from different children
+            addStmtTablesFromChildren(childStmt.m_tableAliasMap);
```
wolffcm: Could it be a problem in the future for m_tableAliasList and m_tableAliasMap to be out of sync?  

(I find it a little strange that we have both a table alias map and a table alias list.  Couldn't we just use the keys of the map as the list?  But I guess that question is beyond the scope of your fix.)

src/frontend/org/voltdb/planner/ParsedUnionStmt.java
```diff
@@ -191,6 +196,22 @@ private void parseOrderColumns(VoltXMLElement columnsNode) {
...
+        for (String alias: tableAliasMap.keySet()) {
+            StmtTableScan tableScan = tableAliasMap.get(alias);
+
+            if (m_tableAliasMap.get(alias) == null) {
+                m_tableAliasMap.put(alias, tableScan);
+            } else {
+                // if there is a duplicate table alias in the map,
+                // find a new unique name for the key
+                // the value in the map are more interesting
+                alias += "_" + System.currentTimeMillis();
```
wolffcm: I guess there is a chance that you could hit two tables with the same name in the same millisecond?  In that case you might miss a partitioned table.

Maybe have a static counter that you increment and append to table aliases to create new table names?

Or, it might make sense to use a multi-map (I am not sure if Java has such a thing) if tables with the same alias are allowed in the case of the set operators?

wolffcm:
That's all of my comments.  I'll have access to email tomorrow, but after that I'll be difficult to get a hold of.

src/frontend/org/voltdb/planner/ParsedUnionStmt.java
jxchinaren: Not really, the m_tableAliasList preserves the Join Order specifically. Maybe we could use LinkedHashMap or rename m_tableAliasListInJoinOrder. 

I think the later one is easier and will do that.

src/frontend/org/voltdb/planner/ParsedUnionStmt.java
jxchinaren: I was hard to hit the problem while writing test cases and moved on.

However, I think a recursive function call will be easier to handle this potential duplicated table alias issue. Will do that.

jxchinaren:
@wolffcm You can review my new commit if you are still available. Please have a good vacation.

wolffcm:
Looks good to me.

PR #2427, opened 2015-06-15
ENG-8277 added exporttypes to rest interface

dweiss88:


src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -462,6 +465,8 @@ public void handle(String target,
                     } else {
                         handleGetUsers(jsonp, target, baseRequest, request, response, authResult.m_client);
                     }
+                } else if (baseRequest.getRequestURI().contains("/export/type")) {
```
manjujames: If it is to get all the export types, should it be /export/types (plural)? Do we have some internal standard for the URLs?

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -725,6 +730,28 @@ public void handleGetUsers(String jsonp, String target,
...
+            }
+            JSONObject exportTypes = new JSONObject();
+            HashSet<String> exportList = new HashSet<String>();
+            for (ServerExportEnum type : ServerExportEnum.values()) {
+                exportList.add(type.value());
+            }
+            try {
+                exportTypes.put("types", exportList);
+            } catch (JSONException e) {
+                m_log.warn("Failed to generate exportTypes JSON: ", e);
```
manjujames: Should we be sending error response back as well in this case?

src/frontend/org/voltdb/utils/HTTPAdminListener.java
dweiss88: Yeah the URL is supposed to match our deployment schema, like how we specify the export type through the tags:
<export>
   <configuration enabled="true" stream="log" type="file">

Since we use "type" in the tag it makes sense to use that in the URL.

src/frontend/org/voltdb/utils/HTTPAdminListener.java
dweiss88: That is a good point, I will add an error response "failed to get types"

akhanzode:
squash merged.

PR #2431, opened 2015-06-16
advanced-inexists.sql: per Paul's clarified suggestion, added one more

johnTcrawford:
pattern ('A44'), testing column expressions on the LHS of the comparator
in an ANY/ALL query (in the outer query, not the sub-query as before).


paulmartel:
Looks good.

PR #2434, opened 2015-06-16
ENG-8372: Now including lib files also in voltdb jar and EELibraryLoa…

manjujames:
…der will

pick it up from the jar. It is still possible to use java.library.path instead using
the flag "use.javalib".


src/frontend/org/voltdb/EELibraryLoader.java
```diff
@@ -81,4 +93,119 @@ public synchronized static boolean loadExecutionEngineLibrary(boolean mustSucced
...
+        boolean success = extractedLibFile.setReadable(true) &&
+                extractedLibFile.setWritable(true, true) &&
+                extractedLibFile.setExecutable(true);
+        if (!success) {
+            hostLog.warn("Could not update extracted lib file " + extractedLibFile +
+                    " to be rwx. Falling back to using java.library.path");
+            return null;
+        }
+
+        /* Do we need this check that is in SnappyLoader?
```
akhanzode: Can you remove this commented code.

src/frontend/org/voltdb/EELibraryLoader.java
```diff
@@ -56,7 +63,12 @@ public synchronized static boolean loadExecutionEngineLibrary(boolean mustSucced
                     assert(versionString != null);
                     final String libname = "voltdb-" + versionString;
                     hostLog.info("Loading native VoltDB code ("+libname+"). A confirmation message will follow if the loading is successful.");
-                    System.loadLibrary(libname);
+                    File libFile = getNativeLibraryFile(libname);
+                    if (libFile==null) {
+                        System.loadLibrary(libname);
```
akhanzode: You can do this based on property here. else find path ad load with path. now null has special meaning here.

src/frontend/org/voltdb/EELibraryLoader.java
```diff
@@ -81,4 +93,119 @@ public synchronized static boolean loadExecutionEngineLibrary(boolean mustSucced
...
+            hostLog.error("Error loading library file from jar. Falling back to using java.library.path", e);
+            return null;
+        }
+    }
+
+    private static File loadLibraryFile(String libFolder, String libFileName, String tmpFolder) throws IOException {
+        hostLog.debug("Loading library from jar using path=" + libFolder + "/" + libFileName);
+
+        // Using UUID as in Snappy, but we probably don't need it?
+        String uuid = UUID.randomUUID().toString();
```
akhanzode: if you use uuid how does cleanup happen? Will every run create a new tmp file?

src/frontend/org/voltdb/EELibraryLoader.java
```diff
@@ -81,4 +93,119 @@ public synchronized static boolean loadExecutionEngineLibrary(boolean mustSucced
...
+                }
+            }
+        }
+        finally {
+            if (reader != null) {
+                reader.close();
+            }
+
+            // Delete the extracted lib file on JVM exit.
+            extractedLibFile.deleteOnExit();
```
akhanzode: ah.

src/frontend/org/voltdb/EELibraryLoader.java
akhanzode: What happens if we can't write to the tmp directory. We see the situation in cloud where /tmp is not writable. Can it be in voltdbroot or override by some property without using java.io.tmpdir. What happens in case of snappy?

src/frontend/org/voltdb/EELibraryLoader.java
manjujames: Not sure I understand.. getNativeLibraryFile will return null if use.javalib is set or of it cannot load from the jar for some reason.

src/frontend/org/voltdb/EELibraryLoader.java
manjujames: Snappy has another property also that they use here for tmp location. I was wondering what that was for. Now I see. I will add a property for volt tmp dir and default to java.io.tmpdir.

src/frontend/org/voltdb/EELibraryLoader.java
manjujames: Sure. I just wanted to make sure that you don't see any reason to have this code in here.

akhanzode:
few things:

1. See if you can add test for override in TestEELibraryLoader
2. Can you test VEM and make sure a cluster can be started?

#1 is not that important you can test the override by hand too.

src/frontend/org/voltdb/EELibraryLoader.java
```diff
@@ -56,7 +64,12 @@ public synchronized static boolean loadExecutionEngineLibrary(boolean mustSucced
                     assert(versionString != null);
                     final String libname = "voltdb-" + versionString;
                     hostLog.info("Loading native VoltDB code ("+libname+"). A confirmation message will follow if the loading is successful.");
-                    System.loadLibrary(libname);
+                    if (Boolean.parseBoolean(System.getProperty(USE_JAVA_LIBRARY_PATH, "false"))) {
```
akhanzode: you can use Boolean.getBoolean() here.

manjujames:
I did try override by hand, but it will be good to have an automated test. I'll add one.

manjujames:
I tried to test VEM a bit already, but I will probably get some of your time today.

PR #2435, opened 2015-06-16
ENG-8423: Push truncation into PBDSegment implementations

nshi:
As part of the work to get the cached count and size updated correctly on PersistentBinaryDeque truncation, I pushed the truncation code down into each PBDSegment implementation.


afrogers:
Looks fine, unless you can think of a way to refactor this to avoid at least some of the duplication.

PR #2437, opened 2015-06-16
ENG-8490 Optimize table block size for mview with no group by

yzhang1991:


src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -389,12 +389,33 @@ Table *TableCatalogDelegate::constructTableFromCatalog(catalog::Database const &
...
+        groupbyColumnCount = mvInfo->groupbycols().size();
+      }
+      else {
+        vector<AbstractExpression *> m_groupByExprs;
+        ExpressionUtil::loadIndexedExprsFromJson(m_groupByExprs, expressionsAsText);
+        groupbyColumnCount = m_groupByExprs.size();
+      }
+      if (groupbyColumnCount == 0) {
+        // ENG-8490: If the materialized view came with no group by, set table block size to 16KB
+        // to achiev better space efficiency.
```
paulmartel: Fix spelling in comment.

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -389,12 +389,33 @@ Table *TableCatalogDelegate::constructTableFromCatalog(catalog::Database const &
...
+        // FYI: maximum column count = 1024, largest fixed length data type is decimal (16 bytes)
+        tableAllocationTargetSize = 1024 * 16;
```
paulmartel: I just remembered that there is one case wider than decimal -- we store short varchars inside the tuple to avoid overhead of the separate allocation -- so the maximum column width is actually 64 bytes, not 16.

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -389,12 +389,33 @@ Table *TableCatalogDelegate::constructTableFromCatalog(catalog::Database const &
     SHA1Init(&shaCTX);
     SHA1Update(&shaCTX, reinterpret_cast<const uint8_t *>(catalogTable.signature().c_str()), (uint32_t )::strlen(catalogTable.signature().c_str()));
     SHA1Final(reinterpret_cast<unsigned char *>(signatureHash), &shaCTX);
+    // Persistent table will use default size (2MB) if tableAllocationTargetSize is zero.
+    int tableAllocationTargetSize = 0;
+    if (materialized) {
+      catalog::MaterializedViewInfo *mvInfo = catalogTable.materializer()->views().get(catalogTable.name());
+      const string expressionsAsText = mvInfo->groupbyExpressionsJson();
```
paulmartel: You do not need to examine groupbyExpressionsJson() to detect that there are group by columns.
The first group by column OR expression will always add one or more elements to groupbycols().
Just check whether groupbycols().isEmpty().

paulmartel:
end of review round

paulmartel:
Looks good. Merge if/when it passes jenkins.

PR #2439, opened 2015-06-17
Implemented Rounding for the DECIMAL type.  

BillWhite:
We round up after 12 digits of scale.  Rounding happens in both the client before the wire
protocol and in the server after the wire protocol.  The server
is configured with a pair of system properties, one to enable
or disable, and one to set the rounding mode.  The client is
configured with a ClientConfig function call.  If the server
and client are configured differently, the results may be
unpredictable.

There are tests for JDBC and for the client interface.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
+++ b/tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
@@ -112,6 +112,10 @@ public class LocalCluster implements VoltServerConfig {
     //This is additional process invironment variables that can be passed.
     // This is used to pass JMX port. Any additional use cases can use this too.
     private Map<String, String> m_additionalProcessEnv = null;
+    protected final Map<String, String> getAdditionalProcessEnv() {
+        return m_additionalProcessEnv;
+    }
+
     // Produce a (presumably) available IP port number.
     public final PortGeneratorForTest portGenerator = new PortGeneratorForTest();
     private String m_voltdbroot = "";
@@ -125,6 +129,9 @@ public class LocalCluster implements VoltServerConfig {
     // with the port numbers and command line parameter value specific to that
     // instance.
     private final CommandLine templateCmdLine = new CommandLine(StartAction.CREATE);
+
+    // This is used in generating the name.
+    private String m_prefix = null;
 
     public LocalCluster(String jarFileName,
                         int siteCount,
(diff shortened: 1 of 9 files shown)


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -126,6 +130,9 @@ public void setHasLocalServer(boolean hasLocalServer) {
     // instance.
     private final CommandLine templateCmdLine = new CommandLine(StartAction.CREATE);
 
+    // This is used in generating the name.
```
paulmartel: Add comment detail like "to avoid name conflicts between localCluster instances that are configured the same except for minor settings."

tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
```diff
@@ -0,0 +1,497 @@
...
+        Client client = getClient();
+        boolean success;
+        ClientResponse cr = null;
+        try {
+            cr = client.callProcedure(insertProcName, insertValue);
+            success = true;
+        } catch (Exception ex) {
+            success = false;
+        }
+        assertEquals(getRoundingString("Insert Statement Compilation Failure"), expectSuccess, success);
```
paulmartel: "Compilation" not the right word here -- suggest "Invocation".

tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
```diff
@@ -0,0 +1,497 @@
...
+                ""
+                ;
+        try {
+            project.addLiteralSchema(literalSchema);
+        } catch (IOException e) {
+            assertFalse(true);
+        }
+        int idx = 0;
+        addConfig(idx++, builder, project, null);
+        addConfig(idx++, builder, project, makePropertiesMap());
```
paulmartel: Testing both the null AND the empty map seem to just be doing a meta-test of the test framework.
It's just going to report redundant errors, if any.
This test seems long enough without that.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -54,7 +54,7 @@
 
     static final int VARCHAR_VARBINARY_THRESHOLD = 100;
 
-    public void testTicketEng2250_IsNull() throws Exception
+    public void notestTicketEng2250_IsNull() throws Exception
```
paulmartel: The "notest" hack is not for long-term use. Was this committed by accident?
If these tests are obsolete, drop them. Otherwise, there at least be an explicit comment explaining why they were hacked out and the plan for getting them hacked back in.

paulmartel:
After our discussion today, maybe there should be test variants that use java stored procs in place of your sql stored procs that synthesize their own scaley values from strings serialized to them.
Look at how an example like how MultiPartitionParamSerializationError is used to see how to wire a java stored proc into a junit test schema.

BillWhite:
I'll delete this branch and push another.

PR #2441, opened 2015-06-17
 OP-103: add support for ResultSet.getString() method to retrieve different SQL types.

JadenW:


src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -790,7 +790,14 @@ public Statement getStatement() throws SQLException {
     public String getString(int columnIndex) throws SQLException {
         checkColumnBounds(columnIndex);
         try {
-            return table.getString(columnIndex - 1);
+            VoltType type = table.getColumnType(columnIndex - 1);
```
akhanzode: may be wrap this into getObject().toString()

JadenW: There is a need for special treatment for type VarBinary,
getObject methods will return Byte[], toString() method wont return actual string but only HashValue.
So right now I decode that by creating new String Object and assign byte[] to it.  
Also , is "UTF-8” decoding need here?

akhanzode: What does other database drivers do? We should be consistent. Also it looks like ou found issue with getObject can you fix that in consistent manner?

JadenW: Mysql could specify an "encoding" for each byte[] filed.  
However votldb always use hex-encoded binary value to represent String input. So I think just to use Encoder.hexEncode() from volt.utils package for covert byte[] to String will be fine.  (This is method is also used for putJSONRep() method, which is used for generating string from bytes[] to JSON string)

I think this is just byte[] to String conversion issue, so getObject method should be fine.

akhanzode:
squash merged.

PR #2450, opened 2015-06-19
Eng 8428 subquery null partial fix

paulmartel:
Fixes null handling issues related to row-compare by implementing a fully three-valued (true, false, null/unknown) "left to right" row comparison.
Adds and reorganizes subquery tests --mostly pertaining to row compare and nulls.


src/ee/expressions/comparisonexpression.h
```diff
@@ -106,13 +171,16 @@ class ComparisonExpression : public AbstractExpression {
...
+    {
         VOLT_TRACE("eval %s. left %s, right %s. ret=%s",
-                   typeid(compare).name(), typeid(*(m_left)).name(),
+                   OP::op_name(),
+                   typeid(*(m_left)).name(),
                    typeid(*(m_right)).name(),
-                   compare.cmp(m_left->eval(tuple1, tuple2),
-                               m_right->eval(tuple1, tuple2)).isTrue()
-                   ? "TRUE" : "FALSE");
+                   ((lnv.isNull() || rnv.isNull()) ? "NULL" :
```
BillWhite: Am I confused?  Aren't lnv and rnv defined below on lines 188 and 193?  How does this compile?

src/ee/expressions/vectorcomparisonexpression.cpp
```diff
@@ -22,54 +22,16 @@ namespace voltdb {
...
-    if (!value.isNullTuple()){
-        int schemaSize = m_tuple.getSchema()->columnCount();
-        for (int columnIdx = 0; columnIdx < schemaSize; ++columnIdx)
-        {
-            if (m_tuple.isNull(columnIdx)) {
-                return true;
-            }
-        }
-        return false;
+    if (value.isNullTuple()) {
```
BillWhite: Does this maybe want to be named "hasNullValue"?  I would think "isNullValue" would have all nulls.
But what do I know?

src/ee/expressions/vectorcomparisonexpression.hpp
```diff
@@ -62,28 +62,48 @@
...
+        else {
+            if (OP::implies_false_for_row(value1, value2) &&
+                ! fallback_result.isNull()) {
+                // allow early return on strict inequality
+                return NValue::getFalse();
+            }
         }
     }
-    return false;
+    // The only cases that have not already short-circuited involve all equal columns.
```
BillWhite: Should this comment be "all columns which compare true"?  If you compare (1, 2, 3) < (4, 5, 6),
won't that end up here with no equality comparison anywhere?

src/frontend/org/voltdb/expressions/ComparisonExpression.java
```diff
@@ -92,17 +92,17 @@ public Object clone() {
     protected void loadFromJSONObject(JSONObject obj) throws JSONException {
         super.loadFromJSONObject(obj);
        if (obj.has(Members.QUANTIFIER.name())) {
-           m_qunatifier = QuantifierType.get(obj.getInt(Members.QUANTIFIER.name()));
+           m_quantifier = QuantifierType.get(obj.getInt(Members.QUANTIFIER.name()));
```
BillWhite: Doesn't all the horsing around with the Member enum seem too clever?  Is this what the term "arch" mean?
Couldn't it just be a final static String?  We only seem to use it when marshalling and unmarshalling json.

BillWhite:
This is the end of my review comments.

src/ee/expressions/comparisonexpression.h
paulmartel: It compiles except when TRACE level is enabled -- I guess I should do a test run like that. Good catch.

src/ee/expressions/vectorcomparisonexpression.hpp
paulmartel: No? (1,2,3) < (4,5,6) short-circuits to true as soon as we detect 1 < 4 (a lexicographic sure thing).
It's cases like (1,2,3) <= (1,2,3) that fall through because the equal columns just passed the buck.

src/frontend/org/voltdb/expressions/ComparisonExpression.java
paulmartel: I don't know from arch, I call it "overengineering". I asked myself this same question back in my early days, and ended up not rocking the boat -- it's long overdue for simplification, and I suppose the sanity could either be achieved with a global makeover OR, at the cost of a temporary inconsistency, phased in, module by module, as convenient.

paulmartel:
squashed and merged from other ENG-8501 branch

PR #2453, opened 2015-06-19
get some potential leaks merged from my findings.

akhanzode:


src/ee/voltdbjni.cpp
```diff
@@ -1289,6 +1289,7 @@ SHAREDLIB_JNIEXPORT jobject JNICALL Java_org_voltcore_utils_DBBPool_nativeAlloca
     jobject buffer = jniEnv->NewDirectByteBuffer( memory, size);
     if (buffer == NULL) {
         jniEnv->ExceptionDescribe();
+        delete memory;
```
BillWhite: Isn't this the perfect place for a boost::scoped_ptr?
What if NewDirectByteBuffer thows an exception?  Then the variable
memory is leaked, isn't it?  NewDirectByteBuffer may not throw today,
but who knows what the future will bring.

We already have some boost in the product.  Putting scoped_ptr and shared_ptr
would be a good idea if they are not already available.

PR #2454, opened 2015-06-19
ENG-8465 fixed using null values with spring

dweiss88:
Seems to fix using null regardless of datatype:

1> create table contestants (a tinyint, b smallint, c integer, d bigint, e float, f decimal, g varchar(20), h varbinary(20), i timestamp);
Command succeeded.
2> select * from contestants;
A     B     C     D     E     F     G     H     I    
----- ----- ----- ----- ----- ----- ----- ----- -----
 NULL  NULL  NULL  NULL  NULL  NULL NULL  NULL  NULL 

(Returned 1 rows in 0.02s)

### DIFF
--- a/tests/frontend/org/voltdb/jdbc/TestJDBCDriver.java
+++ b/tests/frontend/org/voltdb/jdbc/TestJDBCDriver.java
@@ -41,6 +41,7 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.sql.Types;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


akhanzode:
Can you write following test

1. For each type set with type specified value=null a get should return same value.
2. set on each type with String null and Regular null and compare two values via get they should be same.

tests/frontend/org/voltdb/jdbc/TestJDBCDriver.java
```diff
@@ -541,6 +544,59 @@ public void testDoubleInsert() throws SQLException {
...
+        cs.setNull(3, Types.SMALLINT);
+        cs.setNull(4, Types.INTEGER);
+        cs.setNull(5, Types.BIGINT);
+        cs.setNull(6, Types.DOUBLE);
+        cs.setNull(7, Types.VARCHAR);
+        cs.setNull(8, Types.VARBINARY);
+        cs.setNull(9, Types.TIMESTAMP);
+        cs.setNull(10, Types.DECIMAL);
+        cs.execute();
+        cs = conn.prepareCall("{call SelectC}");
```
akhanzode: Which row you are selecting the one which was inserted with all Types.NULL or the second one? Shouldn't you test both. Can you add some comments as well?

PR #2455, opened 2015-06-19
get some potential leaks merged from my findings.

rmorgenstein:


paulmartel:
This change looks good. 
Yet there may be more to do, to be on the safe side:

I noticed one other edge case that could theoretically leak.
In the unlikely scenario that fragmentProgressUpdate failed to allocate jTargetTableName after successfully allocating jPlanNodeName, it seems the latter would be leaked.

Also, I notice that neither getQueuedExportBytes nor pushExportBuffer call ExceptionCheck after NewStringUTF -- assuming it won't get an OutOfMemoryError (a documented possibility).

Finally, we do not check for the null return (a documented possibility) for various calls in voltdbjni.cpp to New<PrimitiveType>Array functions. We just proceed to operate on their results -- and we don't always check the result of THOSE follow-on operations either (not sure whether we should be checking for null or an exception (or both?) after each of those. This seems like it risks undefined behaviors.

PR #2456, opened 2015-06-20
Backport some AST printing changes.

BillWhite:
Backport some AST printing changes.  These are really only used
for debugging, and are not enabled ever in anger.  To enable them, enable the
HSQLDB_COMPILER logger to be in DEBUG mode in tests/log4j-allconsole.xml.
This is the easiest mode to enable, as it's the mode that's already there, but
commented out.


paulmartel:
Looks good. Thanks for splitting this out.

PR #2461, opened 2015-06-23
ENG-8437:

manjujames:
Fixes to classes included in client side 1.6 build so that they will compile correctly.
Now compile with -source 1.6 with the required -bootclasspath option as well.

### DIFF
--- a/build-client.xml
+++ b/build-client.xml
@@ -22,6 +22,7 @@
         <javac
             target="1.6"
             source="1.6"
+            bootclasspath="${vendor.lib.dir}/classes-java1.6.jar"
             srcdir="@{srcdir}"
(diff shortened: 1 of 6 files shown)


BillWhite:
Will these changes require Eclipse changes as well?

manjujames:
No, this won't require Eclipse changes.

akhanzode:
This looks ok to me. If @paulmartel is ok with lexer changes and jenkins is happy.

paulmartel:
Looks good to me. I am a little shocked -- do we really only use this feature (client side) for this one rather uninteresting case?

manjujames:
That's right. We also use capture labels in HSQLLexer, but that is server side only.

PR #2464, opened 2015-06-24
Eng 8526 always listen if dr tag present

luyangco:
The proposed new DR UI document requires the master node to always "listen" to network, even the "listen" attribute in the deployment file is not being set.


afrogers:
Reviewed

PR #2469, opened 2015-06-25
Various AdHoc constants value fixes

jxchinaren:
AdHoc query fix:
1. Treat constants in SQL literal numbers either as integer or decimal.
2. Add tests for float E sign.
3. Fix the various of AdHoc planner caching with type issues.
4. Refactor and add more testing utils.
---
Missing stored procedure fix (Can be done separately from this ticket):
5. Stored procedure constants literal value fix either integer or decimal.
6. More tests on out of range of our decimal literal value if needed.

### DIFF
--- a/src/frontend/org/voltdb/planner/ParameterizationInfo.java
+++ b/src/frontend/org/voltdb/planner/ParameterizationInfo.java
@@ -177,16 +177,31 @@ class ParameterizationInfo {
 
                 VoltXMLElement paramIndexNode = new VoltXMLElement("parameter");
                 paramIndexNode.attributes.put("index", String.valueOf(paramIndex));
-                String typeStr = node.attributes.get("valuetype");
-                paramIndexNode.attributes.put("valuetype", typeStr);
                 paramIndexNode.attributes.put("id", idStr);
                 paramsNode.children.add(paramIndexNode);
 
+                // handle parameter value type
+                String typeStr = node.attributes.get("valuetype");
+                VoltType vt = VoltType.typeFromString(typeStr);
+
                 String value = null;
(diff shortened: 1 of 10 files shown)


src/frontend/org/voltdb/planner/ParameterizationInfo.java
```diff
@@ -177,16 +177,31 @@ static void parameterizeRecursively(VoltXMLElement node,
...
+                VoltType vt = VoltType.typeFromString(typeStr);
+
                 String value = null;
-                if (VoltType.typeFromString(typeStr) != VoltType.NULL) {
+                if (vt != VoltType.NULL) {
                     value = node.attributes.get("value");
                 }
                 paramValues.add(value);
+
+                if (vt == VoltType.NUMERIC) {
```
BillWhite: I think you are trying to see if this numeric value is an integer or a decimal.  If that is right,
perhaps you should add a comment that says so.  What if the string is a float?  What if the
string is not properly formatted, say it's "1.3.4"?

BillWhite:
Done with this comment round.

src/frontend/org/voltdb/planner/ParameterizationInfo.java
```diff
@@ -177,16 +177,31 @@ static void parameterizeRecursively(VoltXMLElement node,
...
+                    vt = VoltType.BIGINT;
+                    try {
+                        Long.parseLong(value);
+                    } catch (NumberFormatException e) {
+                        // DECIMAL probably is not bigger/smaller enough than our decimal range.
```
jxchinaren: Should change the comment here also:
"Our DECIMAL may not be bigger/smaller enough to store the constant value"

PR #2483, opened 2015-06-26
Move decimal rounding to the EE (only default behavior).

BillWhite:
When we create a decimal number from a string in the EE,
and the resulting decimal number would have a scale which
is too large, we round the decimal number to 12 decimal digits.
If the resulting rounded number is still too large we throw
and exception.

There are java unit tests and C++ unit tests.

### DIFF
--- a/src/ee/common/NValue.cpp
+++ b/src/ee/common/NValue.cpp
@@ -27,7 +27,6 @@
 #include <set>
 
 namespace voltdb {
-
 Pool* NValue::getTempStringPool() {
     return ExecutorContext::getTempStringPool();
 }
@@ -215,6 +214,7 @@ std::string NValue::createStringFromDecimal() const {
  *   This function does not handle scientific notation string, Java planner should convert that to plan string first.
  */
 void NValue::createDecimalFromString(const std::string &txt) {
+    std::cout << "createDecimalFromString(" << txt << ");\n";
     if (txt.length() == 0) {
         throw SQLException(SQLException::volt_decimal_serialization_error,
                                        "Empty string provided");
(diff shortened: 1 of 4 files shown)


src/ee/common/NValue.cpp
```diff
@@ -215,6 +214,7 @@ std::string NValue::createStringFromDecimal() const {
  *   This function does not handle scientific notation string, Java planner should convert that to plan string first.
  */
 void NValue::createDecimalFromString(const std::string &txt) {
+    std::cout << "createDecimalFromString(" << txt << ");\n";
```
jxchinaren: Do you need this ?

src/ee/common/NValue.cpp
```diff
@@ -259,27 +259,54 @@ void NValue::createDecimalFromString(const std::string &txt) {
...
     while (fractionalString.size() > 0 && fractionalString[fractionalString.size() - 1] == '0')
         fractionalString.erase(fractionalString.size() - 1, 1);
     // check if too many decimal places
-    if (fractionalString.size() > 12) {
-        throw SQLException(SQLException::volt_decimal_serialization_error,
-                           "Maximum scale exceeded. Maximum of 12 digits to the right of the decimal point");
-    }
-    while(fractionalString.size() < NValue::kMaxDecScale) {
-        fractionalString.push_back('0');
+    if (fractionalString.size() > kMaxDecScale) {
```
jxchinaren: // Can you add a comment for using the default rounding here?

tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
```diff
@@ -467,7 +560,7 @@ private static void addConfig(int idx, MultiConfigSuiteBuilder builder, VoltProj
         final String literalSchema =
                 "CREATE TABLE P1 ( id integer );" +
                 "CREATE TABLE DECIMALTABLE ( " +
-                "dec decimal" +
+                    "dec decimal"     +
```
jxchinaren: Do we have any default decimal string value rounding tests? I guess the default value here also use interpreted from string, but probably at Java level? I am not sure.

End of review round.

BillWhite: Actually, we have tests for all the rounding modes, and for rounding
enabled and disabled in the junit tests.  But this exercise is just to
implement the default, which is rounding enabled and HALF_UP.

src/ee/common/NValue.cpp
BillWhite: Done.

BillWhite:
Close to delete this branch and push again with changes.

PR #2491, opened 2015-06-29
ENG 4980 select count distinct on partition key

yzhang1991:


tests/frontend/org/voltdb/planner/TestPushDownAggregates.java
```diff
@@ -124,6 +124,30 @@ public void testAllPushDownAggregates() {
...
+                                              ExpressionType.AGGREGATE_MAX,
+                                              ExpressionType.AGGREGATE_SUM,
+                                              ExpressionType.AGGREGATE_COUNT},
+                        new ExpressionType[] {ExpressionType.AGGREGATE_SUM,
+                                              ExpressionType.AGGREGATE_SUM,
+                                              ExpressionType.AGGREGATE_MIN,
+                                              ExpressionType.AGGREGATE_MAX,
+                                              ExpressionType.AGGREGATE_SUM,
+                                              ExpressionType.AGGREGATE_SUM},
+                                              true);
```
paulmartel: Add test cases that list count(distinct pkey) with other compatible aggs like non-distinct aggs of other columns, trivially/needlessly distinct min or max of other columns.

Add "negative" test case that selects count(distinct pkey) and count(distinct other) to show that it only takes one other non-trivial distinct on another column to disable the pushdown -- I THINK that's what we expect, but is there a compromise that allows us to somewhat reduce the size of the data set that gets sent to the coordinator -- this task MAY make a good follow-on ticket.

Please add "regression suite" test cases that show correct query results for all of the above queries.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -87,6 +87,7 @@
     private NodeSchema m_avgPushdownFinalProjectSchema;
 
     private boolean m_hasPartitionColumnInGroupby = false;
+    private boolean m_partitionColumnAsAggArg = false;
```
paulmartel: The name, lacking a verb part like "has" is difficult to understand.
I don't know what it means.
And I don't think that it is as useful as it could be given how it is currently calculated (more below).

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -257,6 +293,10 @@ void setupForNewPlans(AbstractParsedStmt parsedStmt) {
                 m_parsedSelect.setHasPartitionColumnInGroupby();
             }
 
+            if (partitionColumnAsAggArg(m_parsedSelect.m_aggResultColumns)) {
+                m_parsedSelect.setPartitionColumnAsAggArg();
```
paulmartel: Hide this assignment inside the function. It does not need a return value.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -224,6 +224,42 @@ private boolean isPartitionColumnInGroupbyList(ArrayList<ParsedColInfo> groupbyC
...
+        assert(m_parsedSelect != null);
+
+        if (aggResultColumns == null) {
+            return false;
+        }
+
+        for (ParsedColInfo aggResultCol : aggResultColumns) {
+            if ( ! (aggResultCol.expression instanceof AggregateExpression)) {
+                continue;
+            }
```
paulmartel: I don't think that it is enough to process the ParsedColInfo's that have top-level aggregate expressions.
What about "select count(distinct pkey) + 1 from table;"?
What about "select count(distinct pkey), count(distinct other) + 1 from table;"?
What about "select count(distinct pkey) from table group by grouping having count(distinct other) > 1;"?

I am pretty sure that there is existing code that seeks out all of the aggregates that are used in a query -- find that code and try to reuse it.

paulmartel: Actually, a possibly better idea would be to delay this processing until all the aggs are found and collected in one place and simply process the resulting collection of aggs. Xin can help you find that.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -224,6 +224,42 @@ private boolean isPartitionColumnInGroupbyList(ArrayList<ParsedColInfo> groupbyC
...
+            boolean match = false;
+            for (SchemaColumn pcol : scanTable.getPartitioningColumns()) {
+                if  (pcol != null && pcol.getColumnName().equals(tve.getColumnName()) ) {
+                    match = true;
+                    break;
+                }
+            }
+            if (! match) {
+                return false;
```
paulmartel: We return false here if any agg we find is not on a TupleValueExpression for a partition column for its table scan.
That's correct sometimes, but why do we care in the case that it is a harmless non-distinct agg or a harmless min(distinct ...) or max(distinct ...)?
It looks like we are missing many optimizable cases.

Also, there is the silly edge case of a distinct agg on a GROUP BY column or GROUP BY expression -- these cases can actually be replaced by the constant 1 for count aggs and replaced by the grouping expression with no agg for sum, min, max, or avg. Distinct makes no difference in this case. Even the case of an agg of any expression based on group by expressions and constants (but no other columns) can be simplified by replacing count(...) with 1 or by eliminating the agg. MAYBE Xin has already implemented some of these optimizations? I don't remember.

Side note: maybe HSQL or (at least at the latest) AbstractParsedStmt.parseAggregateExpression is smart enough to drop the distinct qualifier for max and min aggs so that the planner never sees them? That would be nice.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -224,6 +226,39 @@ private boolean isPartitionColumnInGroupbyList(ArrayList<ParsedColInfo> groupbyC
         return false;
     }
 
+    private boolean canPushDownDistinctAggregation(AggregateExpression aggExpr) {
+        assert(m_parsedSelect != null);
+
+        if ( aggExpr == null || aggExpr.isDistinct() == false ) {
```
paulmartel: These should be asserts.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -224,6 +226,39 @@ private boolean isPartitionColumnInGroupbyList(ArrayList<ParsedColInfo> groupbyC
...
+            return false;
+        }
+        if ( aggExpr.getExpressionType() == ExpressionType.AGGREGATE_COUNT_STAR ) {
```
paulmartel: What does this mean? Is COUNT(DISTINCT *) even legal? or can we assert here too?

paulmartel:
End of review round.

src/frontend/org/voltdb/planner/PlanAssembler.java
yzhang1991: Legal.

yzhang1991:
Squash merged to master.

PR #2496, opened 2015-06-30
Eng 8166 sqlcov normalize

johnTcrawford:
Pull request for the normalizer (second) half of my Innovation Week project. Mainly improved the handling of ORDER BY clauses: normalizers will no longer reorder rows that the database has sorted via ORDER BY columns without table references (e.g. ORDER BY ID will be treated the same as ORDER BY R1.ID); because that broke some tests, also made it ignore sub-queries (anything before ')'), and added options for sorting NULL values in various ways, with a new normalizer for each way (nulls-first-normalizer.py, nulls-last-normalizer.py, nulls-highest-normalizer.py, and especially nulls-lowest-normalizer.py - which is how VoltDB sorts NULL values). Also stopped printing 0 (and other Python False) values as None. And created a NotANormalizer class (now called by not-a-normalizer.py), and a StandardNormalizer sub-class (now called by normalizer.py, as well as the new nulls-... normalizers), to reduce code duplication (though this has not yet been applied to fuzzynormalizer.py).


tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    if start < 0 or end < 0 or start > len(rows) or end > len(rows) or start > end:
+        raise ValueError('Illegal value of start (%d) or end (%d): negative, greater than len(rows) (%d), or start > end' % (start, end, len(rows)))
+    elif moveto < 0 or (moveto > start and moveto < end) or moveto > len(rows):
+        raise ValueError('Illegal value of moveto (%d): negative, greater than len(rows) (%d), or between start (%d) and end (%d)' % (moveto, len(rows), start, end))
+    elif moveto == start or moveto == end:
+        return
+
+    newrows = []
+    newrows.extend(rows[0:min(moveto, start)])
```
paulmartel: Would this be easier to understand (without the need for min) as two separate statements in the if and else blocks below?

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+    if moveto < start:
+        newrows.extend(rows[start:end])
+        newrows.extend(rows[moveto:start])
+    elif moveto > end:
```
paulmartel: This elif is redundant -- it could just be an else -- it misleads me into thinking that there is a third (pass-through) case.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+        newrows.extend(rows[end:moveto])
+        newrows.extend(rows[start:end])
+    newrows.extend(rows[max(moveto, end):len(rows)])
```
paulmartel: Same comment about splitting into if/else case-specific statements, getting rid of max.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+                              (sort_nulls == SortNulls.lowest  and not desc[i])):
+                            if i == 0:  # first ORDER BY column, so don't need to check others
+                                move_rows(rows, begin, j, 0)
+                                break
+                            ref = project_sorted(rows[begin], sorted_cols[0:i])
+                            for k in range(begin - 1, -2, -1):
+                                if (k < 0 or ref != project_sorted(rows[k], sorted_cols[0:i])):
+                                    move_rows(rows, begin, j, k + 1)
+                                    break
+                        elif (sort_nulls == SortNulls.last or
```
paulmartel: This redundant elif vs. else -- quite apart from its added complexity -- also misleadingly suggests that there's a third else case that falls through.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+                if prev != tmp:
+                    if prev and prev[i] is None:
+                        if (sort_nulls == SortNulls.first or
+                              (sort_nulls == SortNulls.highest and desc[i]) or
+                              (sort_nulls == SortNulls.lowest  and not desc[i])):
+                            if i == 0:  # first ORDER BY column, so don't need to check others
+                                move_rows(rows, begin, j, 0)
+                                break
+                            ref = project_sorted(rows[begin], sorted_cols[0:i])
+                            for k in range(begin - 1, -2, -1):
```
paulmartel: Could use a comment to explain what is going on here -- where these strange -2, -1 values are coming from.

paulmartel: Would it be clearer to normalize k differently as in "for k in range(begin, -1, -1)" and adjust references to k accordingly? I can't decide.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+    # Sort SQL NULL (Python None) values, in ORDER BY columns, in the
+    # specified order (if any)
+    if sort_nulls != SortNulls.never:
+        for i in xrange(len(sorted_cols)):
+            begin = 0
+            prev = None
+            for j in xrange(len(rows) + 1):
+                tmp = []
+                if j < len(rows):
+                    tmp = project_sorted(rows[j], sorted_cols[0:i + 1])
```
paulmartel: The spacing in "0:i + 1" obscures the meaning -- the precedence of "+" over ":".

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+                if prev != tmp:
+                    if prev and prev[i] is None:
+                        if (sort_nulls == SortNulls.first or
+                              (sort_nulls == SortNulls.highest and desc[i]) or
+                              (sort_nulls == SortNulls.lowest  and not desc[i])):
+                            if i == 0:  # first ORDER BY column, so don't need to check others
+                                move_rows(rows, begin, j, 0)
+                                break
+                            ref = project_sorted(rows[begin], sorted_cols[0:i])
```
paulmartel: 'project_sorted(X, sorted_cols[0:i])' for various definitions of X is applied in so many places that I wonder if it would aid readability to define it as a lambda at the top of the "for i" loop.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
paulmartel: Actually, I'm not quite understanding this algorithm. Maybe(?) comments would help that.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -0,0 +1,258 @@
...
+        last_paren_index = sql_upper.rfind(')')  # will often be -1
+        sort_cols = parse_sql(sql_upper[last_paren_index+1:])
+        indices = []
+        desc = []
+        if sort_cols:
+            # gets the ORDER BY column indices, in the order used in that clause
+            for i in xrange(len(sort_cols)):
+                # Find the table column name that matches the ORDER BY column name
+                for j in xrange(len(table.columns)):
+                    if sort_cols[i] == table.columns[j].name:
```
paulmartel: Are we really only matching "the hard case" of deriving each column index from column name and ignoring "the easy case" where we are given an explicit column index as in "ORDER BY 1, 2, 3"? Does sqlcoverage NOT verify the order resulting from this syntax?

paulmartel:
End of review round. I'd feel better if I understood how the re-sorting of nulls actually works.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Done.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Done.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Done - simplified move_rows per this comment & the ones above.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Done - changed elif to else (but left a comment enumerating the remaining possibilities.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Added copious comments, explaining what is going on (which I normally avoid where not needed, but here I agree it is). Did not change k, which represents a row index; it would be more confusing to change it.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Agreed; fixed.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Done.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: Yes, we really were ignoring "the easy case" (ORDER BY 1, 2, 3) - but I just fixed that, by changing one line in the normalize function (added an 'or' to one 'if' statement).

PR #2497, opened 2015-06-30
Create OptimizedProjector to efficiently move fields across tuples

wolffcm:
This is a squashed version of #2197.


wolffcm:
Had to add another commit for a debug build failure, will re-squash and then merge to master.

PR #2498, opened 2015-06-30
ENG-8491 make get numbers more JDBC compliant

dweiss88:


src/frontend/org/voltdb/jdbc/JDBC4DatabaseMetaData.java
```diff
@@ -1209,17 +1209,30 @@ public boolean supportsConvert(int fromType, int toType) throws SQLException {
...
         case java.sql.Types.DECIMAL:
-        case java.sql.Types.VARCHAR:
-        case java.sql.Types.VARBINARY:
-        case java.sql.Types.TIMESTAMP:
             switch (toType) {
             case java.sql.Types.VARCHAR:
+            case java.sql.Types.TINYINT:
+            case java.sql.Types.SMALLINT:
+            case java.sql.Types.INTEGER:
+            case java.sql.Types.BIGINT:
```
akhanzode: what about other fromType -> toType like BIGINT to INTEGER, SMALLEST, TINYINT in general going from bigger size to smaller?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -257,7 +257,35 @@ public InputStream getAsciiStream(String columnLabel) throws SQLException {
...
+                decimalValue = new BigDecimal(table.getDouble(columnIndex - 1));
+                break;
+            case DECIMAL:
+                decimalValue = table.getDecimalAsBigDecimal(columnIndex - 1);
+                break;
+            default:
+                throw new IllegalArgumentException("Cannot get BigDecimal value for column type '" + type + "'");
+            }
+
+            if (table.wasNull()) {
```
akhanzode: return table.wasNull() ? null : decimalValue;

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -326,6 +354,49 @@ public Blob getBlob(String columnLabel) throws SQLException {
...
+        case BIGINT:
+            longValue = new Long(table.getLong(columnIndex - 1));
+            break;
+        case FLOAT:
+            final Double retDouble = new Double(table.getDouble(columnIndex - 1));
+            if (retDouble.equals(new Double(Math.ceil(retDouble)))) {
+                longValue = retDouble.longValue();
+                break;
+            }
+            throw new IllegalArgumentException("getByte() called on non-integral column.");
```
akhanzode: getByte()? wrong message?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -326,6 +354,49 @@ public Blob getBlob(String columnLabel) throws SQLException {
...
+        case DECIMAL:
+            final BigDecimal retDec = table.getDecimalAsBigDecimal(columnIndex - 1);
+            if (retDec == null)
+                break;
+            try {
+                longValue = new Long(retDec.longValueExact());
+                break;
+            } catch (ArithmeticException e) {
+                throw new IllegalArgumentException("getByte() called on non-integral column.");
```
akhanzode: same here

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -326,6 +354,49 @@ public Blob getBlob(String columnLabel) throws SQLException {
         return getBlob(findColumn(columnLabel));
     }
 
+    private Long getInteger(int columnIndex) throws SQLException {
+        final VoltType type = table.getColumnType(columnIndex - 1);
+        Long longValue = new Long(0);
```
akhanzode: Why initialize and create new it if you are recreating a new Long(0) for return.

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -326,6 +354,49 @@ public Blob getBlob(String columnLabel) throws SQLException {
...
+            if (retDec == null)
+                break;
+            try {
+                longValue = new Long(retDec.longValueExact());
+                break;
+            } catch (ArithmeticException e) {
+                throw new IllegalArgumentException("getByte() called on non-integral column.");
+            }
+        default:
+            throw new IllegalArgumentException("Cannot get integral value for column type '" + type + "'");
```
akhanzode: integral value?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -326,6 +354,49 @@ public Blob getBlob(String columnLabel) throws SQLException {
...
+            break;
+        case INTEGER:
+            longValue = new Long(table.getLong(columnIndex - 1));
+            break;
+        case BIGINT:
+            longValue = new Long(table.getLong(columnIndex - 1));
+            break;
+        case FLOAT:
+            final Double retDouble = new Double(table.getDouble(columnIndex - 1));
+            if (retDouble.equals(new Double(Math.ceil(retDouble)))) {
```
akhanzode: comment what you are trying to do here. how will ceil help here?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -353,7 +424,11 @@ public boolean getBoolean(String columnLabel) throws SQLException {
     public byte getByte(int columnIndex) throws SQLException {
         checkColumnBounds(columnIndex);
         try {
-            return (new Long(table.getLong(columnIndex - 1))).byteValue();
+            Long longValue = getInteger(columnIndex);
```
akhanzode: original code did table.getLong() whats the purpose to do getInteger now?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -499,7 +574,35 @@ public Date getDate(String columnLabel, Calendar cal) throws SQLException {
     public double getDouble(int columnIndex) throws SQLException {
         checkColumnBounds(columnIndex);
         try {
-            return table.getDouble(columnIndex - 1);
+            final VoltType type = table.getColumnType(columnIndex - 1);
+            Double doubleValue = new Double(0);
```
akhanzode: Why initialize?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -555,7 +688,11 @@ public int getHoldability() throws SQLException {
     public int getInt(int columnIndex) throws SQLException {
         checkColumnBounds(columnIndex);
         try {
-            return (new Long(table.getLong(columnIndex - 1))).intValue();
+            Long longValue = getInteger(columnIndex);
```
akhanzode: why call getInteger and not use table.getLong() getInteger() is doing some additional checks right which makes this method hard to track where is truncation happening etc.

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -574,7 +711,8 @@ public int getInt(String columnLabel) throws SQLException {
     public long getLong(int columnIndex) throws SQLException {
         checkColumnBounds(columnIndex);
         try {
-            return table.getLong(columnIndex - 1);
+            Long longValue = getInteger(columnIndex);
```
akhanzode: this should be safe as is right?

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -751,7 +889,11 @@ public RowId getRowId(String columnLabel) throws SQLException {
     public short getShort(int columnIndex) throws SQLException {
         checkColumnBounds(columnIndex);
         try {
-            return (new Long(table.getLong(columnIndex - 1))).shortValue();
+            Long longValue = getInteger(columnIndex);
```
akhanzode: What happens on other dbs?

src/frontend/org/voltdb/jdbc/JDBC4DatabaseMetaData.java
dweiss88: this catches all of those cases and says they are okay, then checks are done later on in getSomething where if the value is too large it throws an exception.

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
dweiss88: This is because of the JDBC specs http://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getBigDecimal(int)

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
dweiss88: ceil is there to check if the float is an integer value or not, if float == ceil(float) then I know that the float is an integer and it's okay to retrieve this value here if not I throw an exception.

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
dweiss88: This is just a common function I created for repeat behavior, not sure of a better name to call it. It is called by all the get integer methods (getByte, getShort, getInt, and getLong)

src/frontend/org/voltdb/jdbc/JDBC4ResultSet.java
```diff
@@ -326,6 +351,46 @@ public Blob getBlob(String columnLabel) throws SQLException {
         return getBlob(findColumn(columnLabel));
     }
 
+    private Long getPrivateInteger(int columnIndex) throws SQLException {
+        final VoltType type = table.getColumnType(columnIndex - 1);
+        Long longValue = null;
```
vtkstef: this is critical code path. Autoboxing is slower than if we would code this up with primitives. use 'long' instead of `Long`

PR #2501, opened 2015-07-01
Always set use.javalib property in LocalCluster. This makes sure that…

manjujames:
… junits from Eclipse will always pass.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
+++ b/tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
@@ -25,6 +25,7 @@ package org.voltdb.regressionsuites;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
@@ -191,7 +192,9 @@ public class LocalCluster implements VoltServerConfig {
         assert (siteCount > 0);
         assert (hostCount > 0);
 
-        m_additionalProcessEnv = env;
+        m_additionalProcessEnv = env==null ? new HashMap<String, String>() : env;
+        // always set use.javalib for LocalCluster so that Eclipse runs will always be OK.
+        m_additionalProcessEnv.put(EELibraryLoader.USE_JAVA_LIBRARY_PATH, "true");
         // get the name of the calling class
         StackTraceElement[] traces = Thread.currentThread().getStackTrace();
         m_callingClassName = "UnknownClass";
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -241,25 +246,14 @@ public LocalCluster(String jarFileName,
...
-                // set the java lib path to the one for this process - Add obj/release/nativelibs
-                javaLibraryPath = System.getProperty("java.library.path");
-                if (javaLibraryPath == null || javaLibraryPath.trim().length() == 0) {
-                    javaLibraryPath = buildDir + "/nativelibs";
-                } else {
-                    javaLibraryPath += ":" + buildDir + "/nativelibs";
-                }
-            }
-        }
+                buildDir + File.separator + m_jarFileName + ":" + buildDir + File.separator + "prod";
```
akhanzode: You only need this if usejavalib is true right?

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -241,25 +246,14 @@ public LocalCluster(String jarFileName,
...
 
-        if (javaLibraryPath==null) {
-            // need this in classpath to find native library. Otherwise, don't add it to classpath to
-            // test override of loading EE lib from jar.
-            classPath = classPath + ":" + buildDir + File.separator + "prod";
+        // set the java lib path to the one for this process - Add obj/release/nativelibs
+        String javaLibraryPath = System.getProperty("java.library.path");
+        if (javaLibraryPath == null || javaLibraryPath.trim().length() == 0) {
+            javaLibraryPath = buildDir + "/nativelibs";
```
akhanzode: same with this.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
manjujames: This was there in code before always. So, I am just bringing it back to what it was to be safe.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
manjujames: Yes, this will be used only if use.javalib is true, but with the issue that Ning found, I thought it might be better to set it always.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
akhanzode: sure...I was reading diff 2 steps before to now.

akhanzode:
squash merged.

PR #2507, opened 2015-07-02
ENG-8576: Cleanup logging in HostMessenger.

jhugg:
Changes the messaging when a node fails to:

WARN: Received remote hangup from foreign host 172.17.0.6
WARN: Host 1 failed. Cluster remains operational.

Other minor log cleanup. Everything is still in the on-disk log for now. Some warnings became infos to make that happen.


akhanzode:
Why not use just network logger for all the logs coming out of HostMessenger. Otherwise look good.

jhugg:
Anish, I thought about it, but the few places its left didn't feel like network messages. Maybe that means the code should be somewhere else?

PR #2509, opened 2015-07-02
Improve AST Printing in HSQLInterface.

BillWhite:
Improve AST Printing in HSQLInterface.  When printing the HSQLDB AST we
get exceptions sometimes, mostly NPEs.  This can cause printing the
VoltXML printing to not be done, even if it would print just fine.


wolffcm:
Looks good to me.

PR #2511, opened 2015-07-02
Eng 8472 single vdb kit

pdshaw:
Changes to build and associated scripts to include Mac and Linux native libs in voltdb-<rel>.jar


tools/copy-lib.sh
```diff
@@ -0,0 +1,31 @@
...
+# For automated builds only; harmless in a dev build but adds no value
+# so enabled only in the "test" user context
+
+if [ "$USER" != "test" ]; then
+    exit 0
+fi
+
+OS=`uname`
+
+if [ "$OS" = "Darwin" -a -e obj/release/nativelibs/libvoltdb*.jnilib ]
```
akhanzode: Can this be specific version than * it would be good to just keep 2 shared objects we are interested in the jar.

tools/copy-lib.sh
```diff
@@ -0,0 +1,31 @@
...
+then
+    # the Mac case...
+    tar cf - obj/release/nativelibs/ | ssh volt0 "cd libs; tar xf -"
+    exit 0
+else
+    # the Linux case...
+    if [ -e ~/libs/obj/release/nativelibs/* ]; then
```
akhanzode: see if * can be avoided.

akhanzode:
Rest I don't know how to review. Please when you merge squash merge too many commits.

tools/copy-lib.sh
pdshaw: Done. In the queue for testing now, behind 29 other builds :-(

PR #2514, opened 2015-07-06
Remove unnecessary synchronization/concurrency handling from DR path

afrogers:


src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -91,7 +89,14 @@ public static PartitionDRGateway getInstance(int partitionId,
         } catch (IOException e) {
             VoltDB.crashLocalVoltDB(e.getMessage(), false, e);
         }
-        m_partitionDRGateways.put(partitionId,  pdrg);
+
+        synchronized (m_gatewayInitLock) {
```
nshi: I wonder if the lock is even necessary here. The only path that calls `getInstance()` is in `SpInitiator.configure()` which runs on the main thread during initialization.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -91,7 +89,14 @@ public static PartitionDRGateway getInstance(int partitionId,
...
+            assert !m_partitionDRGateways.containsKey(partitionId);
+            ImmutableMap.Builder<Integer, PartitionDRGateway> builder = ImmutableMap.builder();
+            builder.putAll(m_partitionDRGateways);
+            builder.put(partitionId, pdrg);
+            m_partitionDRGateways = builder.build();
```
nshi: I'm seeing this pattern more and more, maybe it's worth extracting it into `MiscUtils`.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -139,92 +146,93 @@ protected AtomicLong initialValue() {
...
-
-                DRRecordType recordType = DRRecordType.valueOf(type);
-                switch (recordType) {
-                case INSERT:
-                case DELETE:
-                case DELETE_BY_INDEX: {
-                    //Insert
-                    if (haveOpenTransaction.get() == -1) {
-                        log.error("Have insert/delete but no open transaction");
+            synchronized (traceLock) {
```
nshi: What's the trace lock for? This block of code doesn't touch any state other than the buffer itself.

PR #2516, opened 2015-07-06
ENG-8541 Active-Active DR schema

slowsoul:
1. In sqlcmd/Ad Hoc DDL mode, added **"set dr=active/passive/default;"** statement for setting DR mode, which can only be changed when all of the DRed tables are empty. Also, added **"show config;"** statement for user to inspect current DR mode.
2. In catalog mode, added **"set dr=active/passive/default;"** statement so that DR mode can be set and compiled to catalog JAR file.
3. For developing, there is a **getIsActiveActiveDRed()** method at database level in catalog for both Java and C++ side to retrieve current DR mode.


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -807,19 +809,40 @@ else if (isTableLimitDeleteStmt(suspect)) {
      * String 1 is name of a table if the change could be made if the table of that name had no tuples.
      * String 2 is the error message to show the user if that table isn't empty.
```
nshi: The comment is out-of-date w.r.t. the return type.

src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -1002,7 +1034,12 @@ private void writeDeletion(CatalogType prevType, CatalogType newlyChildlessParen
         if (errorMessage != null) {
             String[] response = checkAddDropIfTableIsEmptyWhitelist(prevType, ChangeType.DELETION);
             // handle all the error messages and state from the modify check
-            processModifyResponses(errorMessage, response);
+            List<String[]> responseList = null;
+            if (response != null) {
+                responseList = new ArrayList<>();
+                responseList.add(response);
+            }
```
nshi: You can just use `Arrays.asList(...)` to create a single-element list.

src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -1030,7 +1067,12 @@ private void writeAddition(CatalogType newType) {
         if (errorMessage != null) {
             String[] response = checkAddDropIfTableIsEmptyWhitelist(newType, ChangeType.ADDITION);
             // handle all the error messages and state from the modify check
-            processModifyResponses(errorMessage, response);
+            List<String[]> responseList = null;
+            if (response != null) {
+                responseList = new ArrayList<>();
+                responseList.add(response);
+            }
```
nshi: Same here.

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -123,7 +125,7 @@ private boolean isUniqueIndex(CatalogType suspect) {
     }
 
     @Override
-    public String[] checkModifyIfTableIsEmptyWhitelist(CatalogType suspect, CatalogType prevType, String field) {
+    public List<String[]> checkModifyIfTableIsEmptyWhitelist(CatalogType suspect, CatalogType prevType, String field) {
```
nshi: You should also check for activeness here so that both the producer and the consumer clusters have the same settings.

nshi:
Review completed. Please also add the ticket name to the commit when you squash.

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -87,6 +90,14 @@ protected String checkModifyWhitelist(final CatalogType suspect, final CatalogTy
             if ("estimatedtuplecount".equals(field)) {
                 return null;
             }
+        } else if (suspect instanceof Database) {
+            if ("isActiveActiveDRed".equalsIgnoreCase(field)) {
+                return "Incompatible DR modes between two clusters";
+            }
+            if ("schema".equalsIgnoreCase(field) ||
+                "securityprovider".equalsIgnoreCase(field)) {
```
nshi: These two are irrelevant to DR. They should not be serialized as commands sent to the replica cluster.

PR #2519, opened 2015-07-07
Improve printing of HSQLDB Syntax Trees.

BillWhite:


src/hsqldb19b3/org/hsqldb_voltpatches/StatementSchema.java
```diff
@@ -1311,9 +1311,9 @@ Result getResult(Session session) {
 
                     // A VoltDB extension to support indexed expressions,
                     // the assume unique attribute, and partial indexes
-                    org.hsqldb_voltpatches.index.Index addedIndex = 
+                    org.hsqldb_voltpatches.index.Index addedIndex = tableWorks.addIndex(indexColumns, name, unique);
                     // End of VoltDB extension
-                    tableWorks.addIndex(indexColumns, name, unique);
+
```
wolffcm: I think the intent behind the original awkward formatting here was to make the Volt-specific changes appear additive and between the `A VoltDB extension ... End of VoltDB Extension` comments, so maybe leave this as-is.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -2799,5 +2802,73 @@ public boolean voltHasSubqueries() {
...
+                    sb.append(' ');
+                }
+                sb.append("]");
+                break;
+        }
+
+        return sb.toString();
+    }
+
+    public static String indentStr(int blanks, boolean startNL, boolean endNL) {
```
wolffcm: Would be good to follow the convention of prefixing methods with `volt-` here.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
```diff
@@ -911,7 +911,7 @@ private static void addConditionsToList(HsqlArrayList list,
     }
 
     /**
-     * Retreives a String representation of this obejct. <p>
+     * Retrieves a String representation of this obejct. <p>
```
wolffcm: I'm not sure how strict we are being about Volt-specific changes at this point, but historically we've tried to make every change appear additive and between the formatted comments (including for minor changes like typos and whitespace adjustments).  I've avoided making small fixes like this as a result.  It might be good to touch base with Paul to see how strict we still need to be.

wolffcm:
Looks really nice overall.

I wonder if you've considered doing a "hardening" run with all our JUnit tests.  That is, run all our junit tests where we execute voltDescribe for everything that runs though the HSQL parser, just to make sure there are no NPEs or other exceptions.  This is especially important for developers since we'll be using it as a debugging tool, so we'll want the new voltDescribe to be robust to malformed ASTs.

I'd touch base with Paul about our current policy for HSQL changes, but other than that, looks good to me.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
BillWhite: Ok.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
BillWhite: I have reverted it.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementSchema.java
BillWhite: I have reverted it.

BillWhite:
I fully expect to get NPEs.  I have a large bunch of changes to add to this, for some cases I didn't get which are also important.  The most common uses of these are: (1) in HSQLInterface.voltGetXML, where we print debugging information, and (2) in Eclipse, hovering over an HSQLDB statement or expression.  In the first we have try/catch to guard against the NPE.  In the latter Eclipse has a guard.  We see InvocationException errors instead of the NPE. These are a big PITA to diagnose and fix, but they are otherwise benign.

I'm going to delete this branch and repush it.  I have addressed all of your comments, though, by reverting the spelling and formatting changes.  It can't hurt to carefully adhere to the separation protocol.

BillWhite:
Closed to push new changes.

PR #2520, opened 2015-07-07
Eng 8589 row subquery

BillWhite:
Fix a bug with subqueries as parameters to aggregates.  We used to fail with an internal error.  Now we fail with a proper error message.


src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -2339,11 +2340,16 @@ VoltXMLElement voltGetXML(Session session, List<Expression> displayCols,
...
                         // Prepare to skip displayCols that are the referent of a SIMPLE_COLUMN."
                         // quit seeking simple_column's replacement.
-                        return otherCol.voltGetXML(session, displayCols, ignoredDisplayColIndexes, startKey, getAlias());
+                        return otherCol.voltGetXML(session, displayCols, ignoredDisplayColIndexes, startKey, getAlias(), parameters);
                     }
                 }
                 assert(false);
             }
+        } else if (exprOp == OpTypes.ROW_SUBQUERY) {
+            VoltXMLElement subquery = new VoltXMLElement("tablesubquery");
```
wolffcm: There is a switch-case below with a case for TABLE_SUBQUERY. 
- I wonder if we could just reuse the same code for both TABLE_SUBQUERY and ROW_SUBQUERY?
- The code for TABLE_SUBQUERY passes an empty parameter list.  I believe that this list is only used to append a parameter list to the output VoltXML, and that that the parameter list should only be on the outermost statement.  Maybe you should be passing a 0-length parameter list here too?

wolffcm:
That's all my comments for now.  I'd like to better understand how ROW_SUBQUERY and TABLE_SUBQUERY are related.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
BillWhite: Could be.  I wasn't sure what to do here.  But it seemed like the right thing to do to propagate the parameters completely.  I think I need to understand this better.

BillWhite:
I'm closing this and deleting the branch.  I reworked this to use Chris' fix, and extended it to treat ROW_SUBQUERY and SCALAR_SUBQUERY like a TABLE_SUBQUERY.  It seems as if HSQLDB disaggregates them.  I added attributes on the VoltXMLElement to retain their separateness, but the planner won't use that yet.

PR #2523, opened 2015-07-07
Eng 8527 enable flag on connection tag

luyangco:
Put a new "enabled" attribute inside DR connection tag. 

Set "enabled" equals to false will cause DR consumer not to connect producer. Changing this attribute at runtime can be done via @UpdateApplicationCatalog or voltadmin update command.


afrogers:
Reviewed

afrogers:
Squash merged externally

PR #2524, opened 2015-07-08
stop sqlcmd exiting from empty file command

yzhang1991:


paulmartel:
Needs a sqlcmdtest script with baselines to validate correct handling for this case -- continuing to a second command. Two identical bad file commands that get two identical error messages would make a perfect baseline.

PR #2525, opened 2015-07-08
ENG-8550 change error message for mview misplaced count(*)

yzhang1991:


yzhang1991:
I feel strange to see the error message got printed 3 times, I think we should avoid this:
18> CREATE VIEW small_towns_fails ( min_pop, max_pop, number ) 
19>     AS SELECT min(population), max(population), count(*) FROM TOWNS WHERE population < 10000; 
ERROR: [null]: Materialized view "SMALL_TOWNS_FAILS" must have count(*) after the GROUP BY columns (if any) but before the aggregate functions (if any).
ERROR: [null]: [null]: Materialized view "SMALL_TOWNS_FAILS" must have count(*) after the GROUP BY columns (if any) but before the aggregate functions (if any).
[null]: Materialized view "SMALL_TOWNS_FAILS" must have count(*) after the GROUP BY columns (if any) but before the aggregate functions (if any).

PR #2527, opened 2015-07-08
Fix regressed ENG-8352 SQL TRIM runtime error - eliminate unused #def…

paulmartel:
In trying to identify TestFunctionsSuite issues, I discovered that the #defines that connect the TRIM function implementations with their HSQL identifiers had gotten out of synch and needed to be cleaned up a bit.
It was causing a runtime failure for missing TRIM function implementations.
With this patch, TestFunctionsSuite on my soon-to-land branch passes its TRIM function tests.


src/ee/expressions/functionexpression.h
```diff
@@ -287,9 +281,9 @@ static const int FUNC_VOLT_HEX                         = 20029;
 static const int FUNC_VOLT_BIN                         = 20030;
 
 // From Tokens.java.
-static const int SQL_TRIM_LEADING                     = 149;
-static const int SQL_TRIM_TRAILING                    = 284;
-static const int SQL_TRIM_BOTH                        = 22;
+static const int SQL_TRIM_LEADING                     = 151;
```
BillWhite: I wonder how these got values got changed.  If they changed in HSQLDB 1.9.3-->2.3.2, then why didn't the merge procedure catch them?

BillWhite:
Looks plausible to me.  But I wonder how these crept in.

PR #2530, opened 2015-07-08
Rework hsql232 volt patches -- imports, commentary, 2 harmless code t…

paulmartel:
…weaks.
One code tweak -- how we override hsql232's new "NOW" alias in favor of our slightly different original "NOW" alias -- should have no real effect -- it's just a tad more logical.

I SERIOUSLY DOUBT that it will have any effect at all on our current mysterious NOW-related failures, but I can always hope.

The other code tweak -- a cleaner patch for an hsql edge case mishandling of BITNOT.


wolffcm:
Looks fine to me.

BillWhite:
Looks right to me, but I'm not really sufficiently expert to speak categorically.

PR #2532, opened 2015-07-09
Bwhite better hsqldb printing

BillWhite:
This commit improves printing of HSQLDB ASTs.


paulmartel:
End of review round. OK to merge.

PR #2538, opened 2015-07-10
Eng 8589 row subquery

BillWhite:
1. Treat ROW_SUBQUERY, TABLE_SUBQUERY and SCALAR_SUBQUERY similarly when generating VoltXML.  We create tablesubquery elements for each of these, but differentiate between them using an attribute.  The planner, of course, will not be using this yet, but it's there in case we can make use of it.
1. Fix some AST printing issues.  This could be separated out, but it was easier to do it this way.
1. Reenable all the tests in TestVoltCompiler.  I think this fixes some failures in TestSubQuerySuite as well, but I'm not sure.


BillWhite:
The Jenkins run completed.  The tests went from 83 PRO junit failures to 67 PRO junit failures.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -2195,9 +2195,16 @@ public void voltCollectAllColumnNames(OrderedHashSet col_set) {
...
-        prototypes.put(OpTypes.ROW_SUBQUERY,  null); // not yet supported subquery feature
-        prototypes.put(OpTypes.TABLE_SUBQUERY,new VoltXMLElement("tablesubquery"));
+        /*
+         * A ROW_SUBQUERY is a subquery which is expected to return a single,
+         * multicolumn row.  A TABLE_SUBQUERY is an unrestricted subquery.  A
+         * single column, single row query is a SCALAR_SUBQUERY.  We treat them
+         * all roughly interchangeably here, but we will add an attribute which
+         * the planner could use if it wanted to.
+         */
+        prototypes.put(OpTypes.SCALAR_SUBQUERY,new VoltXMLElement("tablesubquery").withValue("subquerytype", "scalar"));
```
wolffcm: I like what you've done here.  Does it make sense to take this to its logical conclusion and change `tablesubquery` to just `subquery`?  Would this mean a lot of changes to the planner?

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -1047,18 +1047,19 @@ public void resolveExpressionTypes(Session session) {
             }
         }
 
-        for (int i = 0; i < rangeVariables.length; i++) {
-            Expression e = rangeVariables[i].getJoinCondition();
+        if (rangeVariables != null) {
```
wolffcm: Did you add this guard to avoid a NPE?  This will need the special formatted comments.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -4013,52 +4013,52 @@ private Database checkDDLAgainstScalarSubquerySchema(String errorRegex, String..
...
                                     "create view tview as select cash, count(*), max(( select cash from books as child where books.cash = child.cash )) from books group by cash;\n");
-         */
     }
 
     public void test8290UnboundIdentifiersNotCaughtEarlyEnough() throws Exception {
         // The name parent is not defined here.  This is an
         // HSQL bug somehow.
-        /* hsql232: ENG-8355, NPE when using subquery with unbound name
-        checkDDLAgainstScalarSubquerySchema("Object not found: PARENT",
+        checkDDLAgainstScalarSubquerySchema("data type cast needed for parameter or null literal",
```
wolffcm: This new error message seems wrong (although it's good that we're no longer crashing).  Do we need a new ticket for this new behavior?

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -4013,52 +4013,52 @@ private Database checkDDLAgainstScalarSubquerySchema(String errorRegex, String..
...
-        checkDDLAgainstScalarSubquerySchema("Partial index \"BIDX\" with subquery expression\\(s\\) is not supported.",
+        checkDDLAgainstScalarSubquerySchema(msg,
                                     "create index bidx on books ( title ) where 'ossians ride' < ( select title from books as child where books.cash = child.cash ) ;\n");
-        */
         // Scalar subquery not allowed in indices.
         checkDDLAgainstScalarSubquerySchema("DDL Error: \"unexpected token: SELECT\" in statement starting on lineno: [0-9]*",
                                     "create index bidx on books ( select title from books as child where child.cash = books.cash );");
-        /* not yet hsql232: ENG-8589, NPE when creating index with subquery
-        checkDDLAgainstScalarSubquerySchema("Index BIDX1 with subquery expression\\(s\\) is not supported.",
+        checkDDLAgainstScalarSubquerySchema("data type cast needed for parameter or null literal",
```
wolffcm: As below, this error message seems wrong given the statement.

wolffcm:
End of review comments.  Would this fix any of the junit_other tests that I commented out under ENG-8307?

BillWhite:
I'm not sure what else is fixed by this commit.  Mostly I was fixing tests in TestVoltCompiler, because they are faster to run than the regression tests.  I'm working on categorizing the tests in TestSubQueriesSuite with this changes right now.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
BillWhite: Done.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
BillWhite: I don't know enough about the planner to answer this question.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
BillWhite: I agree.  I still think we are doing the wrong thing here.  But this error message is what we get, and we expect an error message, after all.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
BillWhite: I agree.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -1047,18 +1047,23 @@ public void resolveExpressionTypes(Session session) {
             }
         }
 
-        for (int i = 0; i < rangeVariables.length; i++) {
-            Expression e = rangeVariables[i].getJoinCondition();
+        // A VoltDB extension to avoid an NPE.
+        if (rangeVariables != null) {
+        // End of VoltDB extension
+            for (int i = 0; i < rangeVariables.length; i++) {
```
wolffcm: Changes to whitespace should be considered like other changes, so I think the preferred way of dealing with this change would be to not indent the block inside the if statement even though that will look weird.  

The basic idea is that if you do `git diff` on your changes, then you should see only green lines, with changed regions starting and ending with the special comments.

BillWhite: Fixed.

PR #2542, opened 2015-07-13
Eng 5954 with clause tb

BillWhite:
I added some tests, but did not implement anything.


tests/frontend/org/voltdb/planner/testwithclause.sql
```diff
@@ -0,0 +1,89 @@
...
+   role_id      INTEGER
+);
+
+CREATE TABLE ROLE(
+   role_id      INTEGER NOT NULL,
+   description  VARCHAR(100)
+);
+
+-- Primary Keys
+ALTER TABLE EMPLOYEE
```
paulmartel: Readability: declaring primary keys within the table definition rather than through an explicit ALTER TABLE, keeps the definition pieces "in one place".

tests/frontend/org/voltdb/planner/TestWithClause.java
```diff
@@ -0,0 +1,168 @@
...
+            "       cp.pdesc\n" +
+            "FROM   employee e,\n" +
+            "       dept_count dc,\n" +
+            "       current_proj cp\n" +
+            "WHERE  e.deptno = dc.deptno\n" +
+            "and    cp.empno = e.empno;\n" +
+            ""
+            );
+    }
+    public void notestWithInFrom() throws Exception {
```
paulmartel: I was expecting subtests that DEFINED the "view" with a WITH clause in the usual place at the top of the statement but just USED the definition in the various places; (FROM clause, IN clause, scalar context).
It seems like we are pushing on edge cases before we've adequately explored common cases.

tests/frontend/org/voltdb/planner/TestWithClause.java
```diff
@@ -0,0 +1,168 @@
...
+        compile("WITH dept_count AS (\n" +
+                "  SELECT deptno, COUNT(*) AS dept_count\n" +
+                "  FROM   employee\n" +
+                "  GROUP BY deptno)\n" +
+                "SELECT e.name AS employee_name,\n" +
+                "       dc.dept_count AS emp_dept_count\n" +
+                "FROM   employee e,\n" +
+                "       dept_count dc\n" +
+                "WHERE  e.deptno = dc.deptno;" +
+                "");
```
paulmartel: It looks like all of these test cases happen to wrap the WITH-defined view (dept_count) in an alias (dc).
We should cover the simple case with no alias.

paulmartel: We should cover the case where a column name/alias in the WITH view is referenced in the query without a "dot" qualifier and must be resolved to the WITH view's column by its unambiguous column name alone.

tests/frontend/org/voltdb/planner/TestWithClause.java
```diff
@@ -0,0 +1,168 @@
...
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package org.voltdb.planner;
+
+public class TestWithClause extends PlannerTestCase {
```
paulmartel: I was actually expecting regressionsuite tests rather than "just" planner tests -- I guess that could be another ticket.

tests/frontend/org/voltdb/planner/TestWithClause.java
```diff
@@ -0,0 +1,168 @@
...
+                    "testwithclause", planForSinglePartitionFalse);
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+    }
+
+    public void notestSimpleWith() throws Exception {
+        compile("WITH dept_count AS (\n" +
```
paulmartel: Readability: defining a column alias with the same name as the WITH view alias invites confusion.

tests/frontend/org/voltdb/planner/TestWithClause.java
```diff
@@ -0,0 +1,168 @@
...
+        setupSchema(TestFunctions.class.getResource("testwithclause.sql"),
+                    "testwithclause", planForSinglePartitionFalse);
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+    }
+
+    public void notestSimpleWith() throws Exception {
```
paulmartel: There is benefit over a completely stubbed notest to having an actual live test that validates the current temporary mode of failure. Then, if something gets fixed, the test failures become a trigger to change and flesh out the test case's expectations.

tests/frontend/org/voltdb/planner/TestWithClause.java
BillWhite: Well, since they don't actually compile, having regressionsuite tests seem premature.  I think there is a lot of work yet before the regressiontsuite tests are warranted, but these seem like the basics of such tests.

Also, regressionsuite tests are slow, even if they don't actually test anything.

tests/frontend/org/voltdb/planner/TestWithClause.java
BillWhite: So, in testDoubleWith, the with-views, if I may call them that, are dept_count and current_proj.  We use dept_count in the output columns clause, the from clause and the where clause.  We use current_proj in the from clause and the where clause.  In testSimpleWith, we use the with-view dept_count similarly in the output columns, the from clause and the where clause.  That seems like defining the with-views and then using them in other places in the query, at least to me.

tests/frontend/org/voltdb/planner/TestWithClause.java
BillWhite: Ok and Ok.

tests/frontend/org/voltdb/planner/TestWithClause.java
BillWhite: Fair enough.  But I kind of want to know what HSQLDB makes of it.  I guess I was pushing an edge case here.

tests/frontend/org/voltdb/planner/TestWithClause.java
BillWhite: Done.

PR #2544, opened 2015-07-13
Allow creation of hidden columns in TupleSchema

wolffcm:


tests/ee/common/tupleschema_test.cpp
```diff
@@ -0,0 +1,235 @@
...
+using voltdb::VALUE_TYPE_TIMESTAMP;
+using voltdb::VALUE_TYPE_VARCHAR;
+
+class TupleSchemaTest : public Test
+{
+};
+
+// A class to automatically free TupleSchema instances, which cannot
+// be allocated on the stack due to variable-length data that follows
+// each instance.  Modeled after boost::scoped_ptr.
```
paulmartel: Should mention the reason this is "modeled after" vs. just "typedef'ed to".
Which is that TupleSchema::freeTupleSchema needs to be called instead of "delete".
Which, given that overriding operator delete for a class is not that difficult to get right, makes freeTupleSchema seem like more trouble than it is worth.

Actually, most product code calls to freeTupleSchema could be eliminated by use of ScopedSchema if it were generally available in TupleSchema.h. And that could eventually be a migration path to a proper operator delete with a typedef or switch-out of ScopedSchema to scoped_ptr.

tests/ee/common/tupleschema_test.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+TEST_F(TupleSchemaTest, HiddenColumn)
+{
+    voltdb::TupleSchemaBuilder builder(2,  // 2 visible columns
+                                       1); // 1 hidden column
+    builder.setColumnAtIndex(0, VALUE_TYPE_INTEGER);
+    builder.setColumnAtIndex(1, VALUE_TYPE_VARCHAR,
+                             256,   // column size
+                             false, // do not allow nulls
+                             true); // size is in bytes
```
paulmartel: I can't think of any applications for non-null-constraining hidden columns --  it seems improbable that there would be a use for throwing a user-visible exception regarding a column that the use is not aware of. And it seems like "system code" should know better than to try to put a null where it doesn't belong.
But I can also see the argument that "a complex consistency can be better than a sporadic simplicity" so we may as well support identical feature sets between visible and hidden columns.

paulmartel: Sorry for placing this comment out of any context, but my general point still holds.

paulmartel:
The code looks good. OK to merge.

tests/ee/common/tupleschema_test.cpp
wolffcm: I see your point.  It would be really confusing to return a constraint violation to the user if a NULL appeared in a hidden column where it shouldn't.  We would want to return an internal error if this ever happened.

However, I think that the existing ColumnInfo structure does an okay job for both visible and hidden columns, and for hidden columns we need to put something in the `allowNulls` field.  By allowing clients of TupleSchema to specify a value here, readers of the code will learn something about how the hidden column is used.  For example, the hidden DR timestamp column should never be null if it exists.

tests/ee/common/tupleschema_test.cpp
wolffcm: Good points.  I'd like to take a stab at this, time permitting for the next batch of changes.

wolffcm:
Will squash and merge in a separate PR.

PR #2546, opened 2015-07-14
Added generated byte array as value to partitioned/replicated tables in TxnIdSelfcheck2

liam700:
Also added inserts to make sure the values retrieved from the database are non-zero.

### DIFF
--- a/tests/test_apps/txnid-selfcheck2/src/txnIdSelfCheck/procedures/UpdateBaseProc.java
+++ b/tests/test_apps/txnid-selfcheck2/src/txnIdSelfCheck/procedures/UpdateBaseProc.java
@@ -101,8 +101,8 @@ public class UpdateBaseProc extends VoltProcedure {
                     " for cid " + cid);
         }
 
-        voltQueueSQL(insert, txnid, prevtxnid, ts, cid, cidallhash, rid, cnt, adhocInc, adhocJmp, new byte[0]);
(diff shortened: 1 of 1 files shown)


tests/test_apps/txnid-selfcheck2/src/txnIdSelfCheck/procedures/UpdateBaseProc.java
```diff
@@ -111,6 +111,9 @@ public long run() {
         // not seen by the server, hopefully this will bisect where they're occuring.
         data = retval[3];
         validateCIDData(data, getClass().getName());
+        
+        VoltTableRow row = data.fetchRow(0);
+        assert(row.getVarbinary("value").length != 0);
```
prosegay: actually, need to use VoltAbortException(...some message...) here instead of an assert since we're in a stored procedure and voltdb is probably not going to be running with asserts enabled.

tests/test_apps/txnid-selfcheck2/src/txnIdSelfCheck/procedures/UpdateBaseProc.java
```diff
@@ -169,6 +172,9 @@ public long run() {
         // not seen by the server, hopefully this will bisect where they're occurring.
         data = retval[3];
         validateCIDData(data, getClass().getName());
+        
+        VoltTableRow row = data.fetchRow(0);
+        assert(row.getVarbinary("value").length != 0);
```
prosegay: same as previous comment here too

PR #2550, opened 2015-07-14
Unit Test for Decimal Rounding Overflow.

BillWhite:
ENG-8510: Implement a unit test for overflow (positive and negative) caused by  by Decimal rounding.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
@@ -269,6 +269,53 @@ public class TestDecimalRoundingSuite extends RegressionSuite {
                             "Attempted to subtract 0.000000000001 from -99999999999999999999999999.999999999999 causing overflow/underflow");
             cr = client.callProcedure("@AdHoc", "truncate table decimaltable;");
             assertEquals(getRoundingString("Table Cleanup failure"), ClientResponse.SUCCESS, cr.getStatus());
+
+            //
+            // Try it with unrepresentable numbers.
+            //
+            String positiveTest = "insert into decimaltable values 99999999999999999999999999.9999999999995;";
+            String positiveTestMsg = "Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: " +
+                                     "Error compiling query: java.lang.RuntimeException: " +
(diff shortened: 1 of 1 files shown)


tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
```diff
@@ -269,6 +269,53 @@ private void doTestDecimalScaleInsertion(boolean roundingEnabled,
...
+            // Try it with unrepresentable numbers.
+            //
+            String positiveTest = "insert into decimaltable values 99999999999999999999999999.9999999999995;";
+            String positiveTestMsg = "Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: " +
+                                     "Error compiling query: java.lang.RuntimeException: " +
+                                     "Decimal 100000000000000000000000000.000000000000 has more than 38 digits of precision.";
+            String negativeTest = "insert into decimaltable values -99999999999999999999999999.9999999999995;";
+            String negativeTestMsg = "Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: " +
+                                     "Error compiling query: java.lang.RuntimeException: " +
+                                     "Decimal -100000000000000000000000000.000000000000 has more than 38 digits of precision.";
```
paulmartel: I GUESS this provides test coverage for the decimal helper code path common to client and server, but it doesn't exercise the error propagation back to a client upon a server-detected error -- such as when passing a string argument with unroundable value to a decimal-typed stored proc param.
It also doesn't demonstrate the runtime error in the EE when the unroundable value makes it as a string into an explicit sql "cast as decimal".

paulmartel:
A good start that stops short. End of review round

tests/frontend/org/voltdb/regressionsuites/TestDecimalRoundingSuite.java
BillWhite: There are other tests for error propagation and for overflow and underflow in the EE.  We already test a lot of cases and propagate errors from the server to the client.  That path seemed to be tested pretty thoroughly already.  I was interested in testing exactly the case that the DecimalHelper class does the right thing on positive and negative overflow.

I'm not sure this error condition is any different from any other kind of error condition.

paulmartel:
Points taken -- OK to merge.

PR #2559, opened 2015-07-16
ENG-8648 Remove local NValueArray parameters

yzhang1991:


wolffcm:
I made a comment in the ticket for this.  What's the motivation to remove the NValueArray parameters?

paulmartel:
Good to merge.

BillWhite:
It looks like the NValueArray object is stored in the Executor Context.  Is this to avoid passing parameters to something that is globally available?  (Is the Executor Context in thread local storage?)

PR #2561, opened 2015-07-16
Eng 8638 too strict column names

BillWhite:
The only real change is in RangeVariable.java, where I rummage around to find a table name when tables don't have aliases.  The change in TestSubqueriesSuite and TestWithStatment are just to reflect that this has been fixed.


BillWhite:
This needs to have the 8070 branch merged into it.  I'll close this and repush.

PR #2562, opened 2015-07-16
ENG-3011: Implement SQL LN and LOG functions

hhakimi53:


paulmartel:
I'd like to see test queries at the bottom of subtestLog that try to do LN(0) and LN(1) via an AdHoc sql styatement and catch the resulting ProcCallException from client.callProcedure("@AdHoc", <query>);

src/ee/expressions/numericfunctions.h
```diff
@@ -179,6 +179,20 @@ template<> inline NValue NValue::callUnary<FUNC_EXP>() const {
...
 
 
+/** implement the SQL LOG/LN function for all numeric values */
+template<> inline NValue NValue::callUnary<FUNC_LN>() const {
+    if (isNull()) {
+        return *this;
+    }
+    NValue retval(VALUE_TYPE_DOUBLE);
+    double inputValue = castAsDoubleAndGetValue();
+    double resultDouble = std::log(inputValue);
```
BillWhite: The unix manual page for log says that log(1 + x) is not very accurate for values of x close to zero, and claims that this is a common case.  In this case, there is a special function log1p which is applied to x, and which computes the answer more accurately.  Perhaps we should think about using this function if inputValue is close to 1.0.  One would have to do a small bit of research to find out what "close" means, though.

BillWhite:
This is the end of my review comments.

PR #2563, opened 2015-07-16
Eng 8025

liam700:
Adding capped collections testing to the Transaction ID Selfcheck 2 testing app.


liam700:
Thought of some more things to add

PR #2564, opened 2015-07-16
Add a EE structure benchmark

jxchinaren:
Add control parameters for the benchmark to run
use 10 seconds to sleep time after data inserts and before data gets destroyed


build.py
```diff
@@ -442,6 +442,7 @@
      CompactingMapIndexCountTest
      CompactingHashTest
      CompactingPoolTest
+     CompactingBenchmark
```
paulmartel: Rename to CompactingMapBenchmark -- I would like to have other Compacting*Benchmarks in the future like Pool and Hash.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+        resultPrinter("SCAN without END() factor", NUM_OF_VALUES, result);
+        result.clear();
+    }
+
+
+    //
+    // LOOK UP
+    //
+    if(runLookup) {
```
paulmartel: One space after "if", please.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+        return false;
+    }
+    return true;
+}
+
+int main(int argc, char *argv[]) {
+    std::vector<bool> params;
+    int scale = 10000;
+
+    printf("Input parameters requested: (data_size_number_of_integers, "
```
paulmartel: I think this output is confusing to see in the normal case -- maybe only print this after testing that first argument starts with "-" because it is either something like "-?" or "--help" or it is some kind of error.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+        bool runScan=true,
+        bool runScanNoEndCheck=false,
+        bool runLookup=false,
+        bool runDelete=false,
+        bool runVoltMap = true,
+        bool runStlMap=false,
+        bool runBoostMap=false,
+        bool runVoltHash=false
+        ) {
+    int BIGGEST_VAL = NUM_OF_VALUES;
```
paulmartel: This would be a good place to echo all the enabled/disabled options -- list what will be run -- to give the interactive user a chance to change their mind or correct their input mistakes.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+    boost::unordered_multimap<int,int>::iterator iter_boost_map;
+    voltdb::CompactingHashTable<int,int>::iterator iter_volt_hash;
+
+    // benchmark
+    BenchmarkRecorder benVoltMap(VoltMap);
+    BenchmarkRecorder benStl(STLMap);
+    BenchmarkRecorder benBoost(BoostUnorderedMap);
+    BenchmarkRecorder benVoltHash(VoltHash);
+
+    int SLEEP_10_SECONDS = 10;
```
paulmartel: It would be nice if I could use the command line to set this to some other value like 0 or -1 to avoid the sleep altogether and get a faster unattended batch run.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+    }
+private:
+    int m_name;
+
+    int64_t m_start;
+    int64_t m_duration;
+};
+
+void resultPrinter(std::string name, int scale, std::vector<BenchmarkRecorder> result) {
+    std::cout << "Benchmark: " << name << ", scale size " << scale << std::endl;
```
paulmartel: std::endl does not mean '\n' -- it means '\n' AND FLUSH.
So use '\n' when you don't need to flush and use std::endl and NOT flush below when you do mean to flush.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+void BenchmarkRun(int NUM_OF_VALUES,
+        bool runScan=true,
+        bool runScanNoEndCheck=false,
+        bool runLookup=false,
+        bool runDelete=false,
+        bool runVoltMap = true,
+        bool runStlMap=false,
+        bool runBoostMap=false,
+        bool runVoltHash=false
```
paulmartel: These default values are never actually used -- drop them.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+#include "harness.h"
+#include "structures/CompactingMap.h"
+#include "structures/CompactingHashTable.h"
+
+using namespace voltdb;
+using namespace std;
+
+class IntComparator {
+public:
+    inline int operator()(const int &lhs, const int &rhs) const {
```
paulmartel: In some ways, it is not exactly fair or realistic to be benchmarking CompactingMap that uses a custom comparator with the others that use a native int comparator -- don't the others allow a similar custom comparator.
In fact, it would be interesting to have a resettable global comparison counter that gets incremented here so that the results can report the number of comparison operations required to perform the benchmark. In the real world use of CompactingMap, comparison costs could be significant and a structure with slightly slower navigation but fewer comparisons COULD be the real winner.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+};
+
+int64_t getMicrosNow () {
+    timeval tv;
+    gettimeofday(&tv, NULL);
+    return tv.tv_sec * 1000000 + tv.tv_usec;
+}
+
+std::vector<int> getRandomValues(int size, int max) {
+    std::vector<int> vec;
```
paulmartel: Pre-allocating the fully sized vector and using assignment would reduce overhead.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    void start() {
+        m_start = getMicrosNow();
+        m_duration = 0;
+    }
+
+    void stop() {
+        m_duration += getMicrosNow() - m_start;
+        m_start = getMicrosNow();
```
paulmartel: Not sure resetting m_start is useful.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+        std::cout.flush();
+    }
+private:
+    int m_name;
+
+    int64_t m_start;
+    int64_t m_duration;
+};
+
+void resultPrinter(std::string name, int scale, std::vector<BenchmarkRecorder> result) {
```
paulmartel: Spending a lot of time doing copy constructors/destructors here because result and ben are instances vs. const & references.

tests/ee/structures/CompactingBenchmark.cpp
```diff
@@ -0,0 +1,442 @@
...
+int main(int argc, char *argv[]) {
+    std::vector<bool> params;
+    int scale = 10000;
+
+    printf("Input parameters requested: (data_size_number_of_integers, "
+            "runScan, runScanNoEndCheck, runLookup, runDelete, "
+            "runVoltMap, runStlMap, runBoostMap, runVoltHash)\n");
+    // 0 is FALSE, others are TRUE
+
+    if (argc > 1) {
```
paulmartel: One size per run makes it difficult to judge scalability. Maybe the default should be more than one run where you can specify the size for each one, or an initial size and a scale factor.
Then it would be convenient for the output to directly report on scaling -- "took 3.5X the time for 10X the entries, took 10.5X the time for 100X entries" . Even setting a scale factor of 1would allow multiple runs that might give more accurate stats than a single run or might give an idea of how much random variability there is in the benchmark timings. Maybe there is a warmup cost for the memory management? Actually, I don't even know if it's fair to compare one structure first and another one second. I suppose we could find out by running both combined runs and one-structure-only runs.

tests/ee/structures/CompactingBenchmark.cpp
paulmartel: That last comment assumes that you can configure each structure with the same or similar custom comparator implementation -- you really should.

paulmartel:
End of round.

tests/ee/structures/CompactingBenchmark.cpp
paulmartel: On the other hand, maybe these default values should be used in the special no-argument case that should be run by default as part of eecheck?

paulmartel: I think that eecheck runs by default with valgrind enabled.
So, do these benchmarks all get built by default with valgrind enabled? Is there a way to override that to get a more real-world benchmark?

paulmartel:
End of things that only occurred to me after the end of the last round.

tests/ee/structures/CompactingMapBenchmark.cpp
```diff
@@ -0,0 +1,520 @@
...
+    if (flag) {
+        return "enabled";
+    }
+    return "disabled";
+}
+
+void BenchmarkRun(
+        int DATA_SCALE,
+        int SLEEP_IN_SECONDS,
+        int READON_OPS_REPEAT,
```
paulmartel: READONLY or even just READ would be clearer than READON.

tests/ee/structures/CompactingMapBenchmark.cpp
```diff
@@ -0,0 +1,520 @@
...
+}
+
+int main(int argc, char *argv[]) {
+    std::vector<bool> params;
+    int data_scale = 10000;
+    int sleep_in_seconds = 5;
+    int readon_ops_repeat = 1;
+
+    if ((argc > 1 && *argv[1] == '-') || argc <= 3) {
+        printf("Input parameters, first 3 required: ("
```
paulmartel: printf("To run a benchmark, execute %s with command line arguments. The first 3 are required: (
...", argv[0]);

Note: comma separator missing at invisible line break between readon_ops_repeat" and "runScan.

Also, after each parameter, list its type like 

    data_scale<int>, sleep_in_seconds<int>, read_ops_repeat<int>, runScan<0 or 1>, runScanNoEndCheck<0 or 1>, ...

paulmartel:
End of round. OK to merge after minor fixes.

jxchinaren:
Will squash merge it.

PR #2574, opened 2015-07-20
Warningscleanup

rbetts:
This branch cleans up Java compiler warnings. Most are harmless. However, the change to HostMessenger looks like a potential crashing defect. The change to ZKUtil fixes a file descriptor leak (in utility code that I don't think is ever called).  Both need more careful review.


src/frontend/org/voltcore/zk/ZKUtil.java
```diff
@@ -123,14 +123,21 @@ public static File getUploadAsTempFile( ZooKeeper zk, String path, String prefix
 
     public static boolean uploadFileAsChunks(ZooKeeper zk, String zkPath, File file, boolean ephemeral)
             throws Exception {
-        if (file.exists() && file.canRead()) {
-            FileInputStream fis = new FileInputStream(file);
-            ByteBuffer fileBuffer = ByteBuffer.allocate((int)file.length());
-            while (fileBuffer.hasRemaining()) {
-                fis.getChannel().read(fileBuffer);
+        FileInputStream fis = null;
+        try {
```
BillWhite: This looks like a good catch.

BillWhite:
The only question I have is about the removed loggers.  Are they never used?  It looks like they aren't.  I wonder where they came from.  Were they used at one time and then the uses removed?  It probably doesn't matter, but the ability to do logging like this can really help in debugging things ot which may be difficult to attach a debugger.

rbetts:
I didn't remove any log messages - just unused logger variables.

BillWhite:
Well, I was sort of wondering where the messages went.  But perhaps they just slowed things down.

BillWhite:
End of my review comments.  All seems ok to me.

src/frontend/org/voltdb/AuthSystem.java
```diff
@@ -626,12 +626,10 @@ public AuthUser getUser(String name) {
 
         private final String m_user;
         private final byte [] m_password;
-        private ClientAuthHashScheme m_hashScheme;
 
         public HashAuthenticationRequest(final String user, final byte [] hash, final ClientAuthHashScheme scheme) {
```
nshi: You can also remove the `scheme` from the parameters.

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -66,6 +65,9 @@
      */
     private final long m_tokens;
     private final int m_tokenCount;
+
+    // Provide a hook for the GC
```
nshi: Thank you, Ryan.

src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -735,12 +735,15 @@ void putJSONRep(int columnIndex, JSONStringer js) throws JSONException {
...
             break;
+        // VoltType includes a few values that aren't valid column value types
         case INVALID:
             break;
         case NULL:
             break;
         case NUMERIC:
             break;
+        case BOOLEAN:
+            break;
```
nshi: Interesting. I'm not sure `BOOLEAN` can ever end up in a `VoltTable`. All other switch cases on `VoltType` in `VoltTableRow` don't handle this type.

nshi:
The rest looks good to me. Thanks for cleaning them up, Ryan!

src/frontend/org/voltdb/VoltTableRow.java
paulmartel: fyi, full support for BOOLEAN as a user data type is probably not that far off, so eventually, the other switch cases will have to catch up.

PR #2576, opened 2015-07-21
For A/A DR, tag DR'd rows with timestamp

wolffcm:
When creating a persistent table from the catalog, add a hidden column
for the DR timestamp to the schema if active/active DR is enabled and
the table is being DR'd.

When rows are inserted or updated in such a table, update the hidden
column with the current DR timestamp.

Add unit tests for this work.

### DIFF
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -277,7 +277,7 @@ public:
     inline const NValue getHiddenNValue(const int idx) const {
         assert(m_schema);
         assert(m_data);
-        assert(idx < m_schema->columnCount());
+        assert(idx < m_schema->hiddenColumnCount());
 
         const TupleSchema::ColumnInfo *columnInfo = m_schema->getHiddenColumnInfo(idx);
         const voltdb::ValueType columnType = columnInfo->getVoltType();
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -202,6 +202,13 @@ public:
      * a persistent table!
      */
     void setNValue(const int idx, voltdb::NValue value) const;
+
+    /*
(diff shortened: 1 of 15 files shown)


src/ee/common/tabletuple.h
```diff
@@ -266,6 +273,20 @@ class TableTuple {
         return NValue::initFromTupleStorage(dataPtr, columnType, isInlined);
     }
 
+    /** Like the above method but for hidden columns. */
+    inline const NValue getHiddenNValue(const int idx) const {
+        assert(m_schema);
+        assert(m_data);
+        assert(idx < m_schema->columnCount());
```
jxchinaren: Should it be 
"assert(idx >= m_schema->columnCount());" 
?

jxchinaren:
Does UPSERT make a difference here? 

Is there a Java API that test regression testable against this hidden timestamp column? If yes, that's worth testing.

jxchinaren:
UPSERT also goes through the code path in persistent table, probably fine. 

Other than the assert, it looks good to me.

src/ee/common/tabletuple.h
wolffcm: it should be `idx < m_schema->hiddenColumnCount()`, since idx is an index into the hidden columns into the tuple.  Good catch.

afrogers:
Reviewed. Looks good to me.

wolffcm:
@jxchinaren, I've pushed another commit that addresses the bad assert you found.

As far as regression testing, there's no existing tests, and the ticket only mentions cpp unit tests.  I filed ENG-8686 as a way of testing hidden columns:

https://issues.voltdb.com/browse/ENG-8686

Thanks for the quick reviews, I'll squash and then merge shortly.

PR #2579, opened 2015-07-21
Improve catalog roundtrip failure diagnostics

paulmartel:


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -21,6 +21,7 @@
 
 package org.voltdb.catalog;
 
+import java.io.UnsupportedEncodingException;
```
wolffcm: Nothing in your changes suggests these imports are needed---delete them?

wolffcm:
Looks good to me.  This will be a big help.

Would it be difficult to make the failure message when the round-tripping fails inform the user about setting this new variable to get better diagnostics?

PR #2580, opened 2015-07-21
ENG-8658: Refuse to start if the kernel is configured to always use THP

afrogers:


lib/python/voltcli/checkconfig.py
```diff
@@ -0,0 +1,41 @@
...
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+
+import glob
+
+def check_thp_config():
+    thp_filenames = glob.glob("/sys/kernel/mm/*transparent_hugepage/enabled")
```
akhanzode: This should be OS specific only right on OS X this should not run.

nshi: I think it return `None` if this file doesn't exist, which is the case for OS X.

PR #2581, opened 2015-07-21
Added a guard for aggregate functions in indices.

BillWhite:
I added a guard for aggregate functions in indices, and also added a test for them.  There was one for aggregate functions in partial indices.  I just adapted it.

Jira Ticket Number: ENG-8632

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -1611,7 +1611,7 @@ public class DDLCompiler {
 
         AbstractParsedStmt dummy = new ParsedSelectStmt(null, db);
         dummy.setTable(table);
-        StringBuffer msg = new StringBuffer(String.format("Index \"%s\" ", name));
+        StringBuffer msgsb = new StringBuffer(String.format("Index \"%s\" ", name));
         // "parse" the expression trees for an expression-based index (vs. a simple column value index)
         List<AbstractExpression> exprs = null;
         // "parse" the WHERE expression for partial index if any
@@ -1638,9 +1638,9 @@ public class DDLCompiler {
         }
 
         // Check all the subexpressions we gathered up.
-        if (!AbstractExpression.areIndexableExpressions(checkExpressions, msg)) {
+        if (!AbstractExpression.areIndexableExpressions(checkExpressions, msgsb)) {
             // The error message will be in the StringBuffer msg.
-            throw m_compiler.new VoltCompilerException(msg.toString());
+            throw m_compiler.new VoltCompilerException(msgsb.toString());
         }
         String colList = node.attributes.get("columns");
         String[] colNames = colList.split(",");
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1653,11 +1653,19 @@ void addIndexToCatalog(Database db, Table table, VoltXMLElement node, Map<String
                         String msg = String.format("Index %s with subquery expression(s) is not supported.", name);
                         throw this.m_compiler.new VoltCompilerException(msg);
                     }
+                    // The expr cannot contain time sensitive function calls.
+                    // This seems to mean exactly that the expr cannot contain calls
+                    // to the function voltGetCurrentTimestampId.
```
paulmartel: Slightly misleading comment -- voltGetCurrentTimestampId is NOT the name of the forbidden function -- it is an internal method to get the forbidden function's id value -- the actual forbidden function having that id is called "CURRENT_TIMESTAMP" for which "NOW" is an alias.

paulmartel:
Actually, your observation about the similarity of indexed expression guards and index predicate guards is a point well taken. The guards should actually be identical -- no time-sensitive functions, no aggs, and no subqueries. I'd refactor this battery of validations into a private method with an AbstractExpression argument.

Actually, the VERY SAME validation should be getting applied to the following components of materialized views:
  -- the WHERE clause
  -- the GROUP BY clause (including the tagged referent expression if the GROUP BY uses the SELECT list's tags mysql-style)
  -- the arguments to the SELECT list agg functions -- That is, all AGGs in the view must be top-level expressions in the SELECT list (a.k.a. "display columns") that must NOT contain the forbidden expressions (time-sensitives, aggs, or subqueries) in their one-each child expressions.

I'd advocate also reducing the reliance on HSQL guards for this kind of thing -- rather than increasing it, as advised in my earlier comment near the assert. So, in the common validation logic, replace the assert that HSQL has precluded subquery expressions with a proper VoltCompilerException throw in this case -- even if we DO, for now, continue to let hsql take a first crack at validation in the subcases it recognizes as unsupported.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -4060,6 +4060,29 @@ public void test8290UnboundIdentifiersNotCaughtEarlyEnough() throws Exception {
...
+        checkDDLAgainstGivenSchema(".*Index FAULTY with aggregate expression\\(s\\) is not supported\\.",
+                                   ddl,
+                                   "create index faulty on alpha(id, count(seqnum));");
+        checkDDLAgainstGivenSchema(".*Index FAULTY with aggregate expression\\(s\\) is not supported\\.",
+                                   ddl,
+                                   "create index faulty on alpha(id, count(*));");
+        checkDDLAgainstGivenSchema(".*Index FAULTY with aggregate expression\\(s\\) is not supported\\.",
+                                   ddl,
+                                   "create index faulty on alpha(id, sum(id));");
+    }
```
paulmartel: Will want to augment existing test cases to reflect suggested guard creep -- and cover the matrix of subquery/agg/time-sensitive-function crossed with indexed expression, index predicate, matview GROUP BY, matview AGG subexpression, and matview predicate.

paulmartel:
End of review round.

BillWhite:
Closed to squash.

PR #2592, opened 2015-07-23
Unify and add to the list of legality tests used by indices and materialized views.

BillWhite:
Unify and add to the list of legality tests used by indices and materialized views.  Perhaps partial indices can be included here as well?

This is the commit to master.  There is another commit to the 8070 branch, where there are slight differences.

Jira Ticket Number: ENG-8632


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2547,40 +2532,78 @@ private void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt) throws V
...
-                throw m_compiler.new VoltCompilerException(msg);
+                msg.append("must exactly match the GROUP BY clause at index " + String.valueOf(i) + " of SELECT list.");
+                throw m_compiler.new VoltCompilerException(msg.toString());
+            }
+            // Gather up aggregate expressions.  Check their
+            // arguments.  If this display column is not an aggregate expression,
+            // then check it all.
+            AbstractExpression expr = outcol.expression;
+            if (expr.getClass() == AggregateExpression.class) {
+                addArgumentExpressions(checkExpressions, expr);
```
paulmartel: This branch is dead code - you can not logically group by an aggregate.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2494,6 +2494,21 @@ private AbstractExpression buildPartialIndexPredicate(
         return predicate;
     }
 
+    private void addArgumentExpressions(List<AbstractExpression> checkExpressions, AbstractExpression expr) {
```
paulmartel: This seems like overkill. The only possible non-null/non-empty case of agg children is from getLeft().
The rest is dead code.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -1091,4 +1092,73 @@ public static boolean hasInlineVarType(AbstractExpression expr) {
...
+    /**
+     * Return true iff the given expression usable as part of an index expression.
+     * If false, put the tail of an error message in the string buffer.  The
+     * string buffer will be initialized with the name of the index.
+     *
+     * @param expr The expression to check
+     * @param msg  The StringBuffer to pack with the error message tail.
+     * @return true iff the expression can be part of an index.
+     */
+    public static boolean isIndexableExpression(AbstractExpression expr, StringBuffer msg) {
```
paulmartel: A static function that takes an argument of the class type? not a method?.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -1091,4 +1092,73 @@ public static boolean hasInlineVarType(AbstractExpression expr) {
...
+
+    /**
+     * This function will recursively find any function expression with ID functionId.
+     * If found, return true. Otherwise, return false.
+     *
+     * @param expr
+     * @param functionId
+     * @return
+     */
+    public static boolean containsFunctionById(AbstractExpression expr, int functionId) {
```
paulmartel: If not for the expr == null case --- which COULD be handled externally by guards, this could also be a method.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -1091,4 +1092,73 @@ public static boolean hasInlineVarType(AbstractExpression expr) {
...
+        if (expr == null || expr instanceof TupleValueExpression) {
```
paulmartel: minor: AbstractValueExpression could also be used as a generalization on TupleValueExpression that also covers ConstantValueExpression and other non-FunctionExpression leaf nodes. TVE is going to be the most common  negative case, but I don't think it hurts to cast a slightly wider net.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -4059,17 +4059,24 @@ public void testScalarSubqueriesExpectedFailures() throws Exception {
...
                                     "create index bidx2 on books ( cash + ( select cash from books as child where child.title < books.title ) );");
         // Scalar subquery not allowed in materialize views.
         checkDDLAgainstScalarSubquerySchema("Materialized view \"TVIEW\" with subquery sources is not supported.",
                                     "create view tview as select cash, count(*) from books where 7 < ( select cash from books as child where books.title = child.title ) group by cash;\n");
         checkDDLAgainstScalarSubquerySchema("Materialized view \"TVIEW\" with subquery sources is not supported.",
                                     "create view tview as select cash, count(*) from books where ( select cash from books as child where books.title = child.title ) < 100 group by cash;\n");
+        /*
+         // This query is not actually correct.  The problem is not the existence
+         // of the subquery in the display list, but that the group by
+         // clause is uses alias for the display list subquery.  This is
```
paulmartel: Bad grammar and ambiguity : "is uses alias" means what?
Rather than losing a test case, should we be validating the actual error message for this usage, if not here, then elsewhere -- possibly in a simpler more isolated non-view context?

paulmartel:
End of review round.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: This function used to be called 'containsTimeSensitiveFunction' and was defined in DDLCompiler.  But it didn't actually depend on being time sensitive at all, so I thought it was better with this name.  I needed to move it to AbstractExpression because it's only ever used in isIndexableExpression.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
BillWhite: Sorry.  A typo.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: It's 'f(e, a)' or 'e.f(a)'.  It's hardly more than a syntactic variant one of the other.  But I changed it.

src/frontend/org/voltdb/compiler/DDLCompiler.java
BillWhite: I'm a little unnerved by putting this knowledge in the code here, but I've left eliminated addArgumentExpressions.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: I don't understand this comment.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2547,40 +2494,72 @@ private void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt) throws V
...
             }
             checkExpressions.add(outcol.expression);
         }
 
         AbstractExpression coli = stmt.m_displayColumns.get(i).expression;
         if (coli.getExpressionType() != ExpressionType.AGGREGATE_COUNT_STAR) {
-            msg += "must have count(*) after the GROUP BY columns (if any) but before the aggregate functions (if any).";
-            throw m_compiler.new VoltCompilerException(msg);
+            msg.append("is missing count(*) as the column after the group by columns, a materialized view requirement.");
+            throw m_compiler.new VoltCompilerException(msg.toString());
```
paulmartel: This message appears to have regressed to its former content (apparently backed out a change by Ethan?)

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2547,40 +2494,72 @@ private void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt) throws V
...
 
         for (i++; i < displayColCount; i++) {
             ParsedColInfo outcol = stmt.m_displayColumns.get(i);
-            if ((outcol.expression.getExpressionType() != ExpressionType.AGGREGATE_COUNT) &&
-                    (outcol.expression.getExpressionType() != ExpressionType.AGGREGATE_SUM) &&
-                    (outcol.expression.getExpressionType() != ExpressionType.AGGREGATE_MIN) &&
-                    (outcol.expression.getExpressionType() != ExpressionType.AGGREGATE_MAX)) {
-                msg += "must have non-group by columns aggregated by sum, count, min or max.";
-                throw m_compiler.new VoltCompilerException(msg);
+            if ((outcol.expression.getClass() != AggregateExpression.class)) {
```
paulmartel: AFAIK AggregateExpression is a leaf class, so "( ! outcol.expression instanceof AggregateExpression)" would be equivalent. The latter arguably simpler and more future-proof in case we ever had reason to subtype AggregateExpression.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2547,40 +2494,72 @@ private void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt) throws V
...
+                msg.append("must have non-group by columns aggregated by sum, count, min or max.");
+                throw m_compiler.new VoltCompilerException(msg.toString());
+            }
```
paulmartel: The original test here guarded against AVG by omission -- perhaps a shade too subtly.
 Looks like this broke.  If so, and this was not caught in unit testing, we need a negative unit test for AVG in matview.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2547,40 +2494,72 @@ private void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt) throws V
...
+            if (outcol.expression.getLeft() != null) {
```
paulmartel: If you're uncomfortable with assuming that right and args children are null non-issues, feel free to plant an assert here that they are null.

src/frontend/org/voltdb/expressions/AbstractExpression.java
paulmartel: c) instanceof TupleValueExpression is a fine optimization for a common case, BUT instanceof AbstractValueExpression (its base class, common to other leaf classes, all similarly optimizable, here) would catch a broader set of cases.  Note: AbstractValueExpression the base class of leaf classes NOT to be confused with AbstractExpression -- the base class of them all.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -4059,17 +4059,25 @@ public void testScalarSubqueriesExpectedFailures() throws Exception {
...
+        /*
+         // This query is not actually correct.  The problem is not the existence
+         // of the subquery in the display list, but that the group by
+         // clause uses an alias for the display list subquery.  This is
+         // arguably not standard SQL, but it is frequently implemented
+         // this way.  Until we implement this properly we should just
+         // ignore this here.
         checkDDLAgainstScalarSubquerySchema("Materialized view \"TVIEW\" with subquery sources is not supported.",
                                     "create view tview as select ( select cash from books as child where books.title = child.title ) as bucks, count(*) from books group by bucks;\n");
+        */
```
paulmartel: In what sense is this query not correct?
To clarify, we should determine whether we support this exact "select ... group by ... " syntax outside matviews. If not, then avoiding this particular negative test case  in the matview context seems more acceptable -- yet it would be better if it could cross-reference a JIRA ticket.

For the materialized view context, we should be giving a clear, ideally consistent, error message for any syntax we only support in other contexts.
The fact that this test is commented out suggests that that's not happening.
Your comment describes a precondition of the failure, but it would help to describe here -- or in a cross-referenced ticket -- exactly what IS happening in this case, instead.

src/frontend/org/voltdb/expressions/AbstractExpression.java
paulmartel: I agree with your comment, more or less (see my other recent comment), but I am not seeing a connection to my original comment. Let me rephrase and embellish:
  a) public static with expr arg could just be private (non-static) w/out expr arg. Because idiomatic code and minimal scope ROCK!
  b) expr == null seems out of place -- IF such a guard is needed, I'd push it out into any caller (hypothetical) that needed it.

paulmartel:
End of review round.

BillWhite:
Closed to squash.

PR #2600, opened 2015-07-24
ENG-8590: check wasnull to make sure its null.

akhanzode:


vtkstef:
R.by: stefano

vtkstef:
--squash merged

PR #2603, opened 2015-07-24
Altered the CopyLoadPartitioned procs to use insert into instead of a

liam700:
select followed by an insert, as the select and insert sql commands are
tested in many places besides this.


tests/test_apps/txnid-selfcheck2/src/txnIdSelfCheck/procedures/CopyLoadPartitionedBase.java
```diff
@@ -29,23 +29,16 @@
...
-            throw new VoltAbortException("Failed to find cid that should exist: cid=" + cid);
+        int cnt = (int) data.fetchRow(0).getLong(0);
+        if (cnt != 1) {
+            throw new VoltAbortException("incorrect number of inserted rows=" + cnt + " for cid=" + cid);
         }
-        data.advanceRow();
-        long rcid = data.getLong(0);
-        if (rcid != cid) {
-            throw new VoltAbortException("Failed to find cid does not match. (" + rcid + ":" + cid + ")");
-        }
```
liam700: Even though this cid check is taken out in this pull request, the cids for every insert done are checked by the delete statements called immediately after these copies, look at the LoadTableLoader class (lines 243-252) for where the inserts and deletes are called and the DeleteLoadPartitioned procedures to see how they check for each cid's presence.

liam700:
more changes to be added

PR #2606, opened 2015-07-26
Fix the reverse scan null predicate missing issue for JOIN case

jxchinaren:


src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -580,8 +580,11 @@ private AbstractPlanNode getSelectSubPlanForJoin(BranchNode joinNode,
                 // InnerPlan is an IndexScan. In this case the inner and inner-outer
                 // non-index join expressions (if any) are in the otherExpr. The former should stay as
                 // an IndexScanPlan predicate and the latter stay at the NLJ node as a join predicate
-                List<AbstractExpression> innerExpr = filterSingleTVEExpressions(innerAccessPath.otherExprs);
-                joinClauses.addAll(innerAccessPath.otherExprs);
+                ArrayList<AbstractExpression> otherExprs = new ArrayList<AbstractExpression>();
+                // PLEASE do not update the "innerAccessPath.otherExprs", it may be reused
+                // for other path evaluation on the other outer side join.
```
paulmartel: I don't understand this comment. Can you identify the method where the innerAccessPath.otherExprs element needs to be reused? Can you identify a specific case where we want to extract a single-TVE expression into innerExpr and also keep it available in innerAccessPath.otherExprs?

Is the idea that we extract all single-TVE expressions as just a first pass to get conditions that MIGHT be indexable on the one specific side, and we don't care yet about the TVEs' tables because we will check that we are on the correct side (for that table) later? I could believe that, but, for the sake of query performance, I would want some confirmation that the final plan does not leave behind a completely redundant join condition that was already enforced by the child scan (either as an index-covered predicate or as a scan post-predicate).

I am suspicious that this SEEMs to specifically solve a join-of-reverse-scan-with-null case.
Does this mean that a reverse scan could still have a null-handling bug in the non-join case?
Or can you explain how this join mis-handling was specifically causing a null handling bug --
 was it really a more general bug, and this reverse scan null case just happened to be the one reproducer we found -- one of many possible symptoms of a missing filter?

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -644,19 +647,22 @@ else if (canHaveNLIJ) {
     /**
      * A method to filter out single TVE expressions.
      *
-     * @param expr List of expressions.
+     * @param expr List of single TVE expressions.
+     * @param otherExprs List of expressions other than TVE.
      * @return List of single TVE expressions from the input collection.
      *         They are also removed from the input.
```
paulmartel: This comment is now a lie.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -644,19 +647,22 @@ else if (canHaveNLIJ) {
     /**
      * A method to filter out single TVE expressions.
      *
-     * @param expr List of expressions.
+     * @param expr List of single TVE expressions.
+     * @param otherExprs List of expressions other than TVE.
```
paulmartel: The opposite of "single-TVE expressions" is NOT "expressions other than TVE" it is "multi-TVE expressions".

I am hoping that the other possible case -- zero-TVE expressions (constant conditions) -- are not getting mixed in here!

Also, change all comments to hyphenate "single-TVE". A "single TVE expression" can be mistaken to mean "one that IS just a TVE". A "single-TVE expression" more clearly expresses "one that only HAS one TVE".

PR #2615, opened 2015-07-27
Eng 8446

manjujames:
No tests yet, so this PR is just for early review, not for merging.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1337,9 +1337,19 @@ public void run() {
                 SystemStatsCollector.asyncSampleSystemNow(true, true);
             }
         }, 0, 6, TimeUnit.MINUTES));
+
         GCInspector.instance.start(m_periodicPriorityWorkThread);
     }
 
+    private void startResourceUsageMonitor() {
+        ResourceUsageMonitor resMonitor  = new ResourceUsageMonitor(m_catalogContext.getDeployment().getSystemsettings());
+        if (resMonitor.hasResourceLimitsConfigured()) {
```
akhanzode: What if update deployment is done with resource limits. Will it start monitoring it?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1337,9 +1337,19 @@ public void run() {
...
+            int resourceMonitorInterval = Integer.getInteger(RESOURCE_MONITOR_INTERVAL, 60);
+            m_periodicWorks.add(scheduleWork(resMonitor, resourceMonitorInterval,
+                    resourceMonitorInterval, TimeUnit.SECONDS));
+        }
+    }
+
```
akhanzode: Can you start this monitor only if we have any limits set? on update catalog you can then redirect and spawn and kill if limits are cleared.

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+{
+    private static final VoltLogger m_logger = new VoltLogger("RESOURCE_MONITOR");
+
+    private long m_rssLimit;
+
+    public ResourceUsageMonitor(SystemSettingsType systemSettings)
+    {
+        if (systemSettings != null && systemSettings.getMemorylimit() != null) {
+            // configured value is in GB. Convert it to bytes
+            double dblLimit = systemSettings.getMemorylimit().getSize().doubleValue()*1024*1024*1024;
```
akhanzode: you can collapse this into one *

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+import org.voltdb.utils.SystemStatsCollector;
+import org.voltdb.utils.SystemStatsCollector.Datum;
+
+/**
+ * Used to periodically check if the server's resource utilization is above the configured limits
+ * and pause the server.
+ */
+public class ResourceUsageMonitor implements Runnable
+{
+    private static final VoltLogger m_logger = new VoltLogger("RESOURCE_MONITOR");
```
akhanzode: I guess use HOST logger for this.

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public void run()
+    {
+        if (VoltDB.instance().getMode() != OperationMode.RUNNING) {
+            return;
+        }
+
+        Datum datum = SystemStatsCollector.getRecentSample();
+        if (datum == null) { // this will be null if stats has not run yet
+            m_logger.debug("No stats are available from stats collector. Skipping resource check.");
```
akhanzode: This should not happen often right? log it at warn then so we can at least see why resource monitors didn't kick in.

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+            return;
+        }
+
+        if (m_logger.isDebugEnabled()) {
+            m_logger.debug("RSS=" + datum.rss + " Configured rss limit=" + m_rssLimit);
+        }
+        if (datum.rss >= m_rssLimit && VoltDB.instance().getMode() == OperationMode.RUNNING) {
```
akhanzode: you have already checked for the operation mode so second check is not necessary.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -1011,6 +1019,11 @@ private String writeDeploymentFile(
             query.setTimeout(m_queryTimeout);
             systemSettingType.setQuery(query);
         }
+        if (m_rssLimit != null) {
+            Memorylimit memoryLimit = factory.createSystemSettingsTypeMemorylimit();
```
akhanzode: you want to use same system settings if query timeout is also specified in test and not overwrite them in deployment.

src/frontend/org/voltdb/InternalConnectionHandler.java
```diff
@@ -0,0 +1,161 @@
...
+
+        int counter = 1;
+        int maxSleepNano = 100000;
+        long start = System.nanoTime();
+        while ((backPressureTimeout > 0) && (m_adapter.getPendingCount() > MAX_PENDING_TRANSACTIONS)) {
+            try {
+                int nanos = 500 * counter++;
+                Thread.sleep(0, nanos > maxSleepNano ? maxSleepNano : nanos);
+                //We have reached max timeout.
+                if (System.nanoTime() - start > backPressureTimeout) {
```
vtkstef: these measurements are costly consider approximating

src/frontend/org/voltdb/InternalConnectionHandler.java
```diff
@@ -0,0 +1,161 @@
...
+                m_logger.error("Can not invoke SP procedure from streaming interface partition not found.");
+                m_failedCount.incrementAndGet();
+                tcont.discard();
+                return false;
+            }
+        }
+
+        boolean success;
+        //Synchronize this to create good handles across all ImportHandlers
+        synchronized(InternalConnectionHandler.class) {
```
vtkstef: are we really sure!?!?!

src/frontend/org/voltdb/InternalConnectionHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Returns true if a table with the given name exists in the server catalog.
+     */
+    public boolean hasTable(String name) {
+        Table table = VoltDB.instance().getCatalogContext().tables.get(name);
+        return (table!=null);
+    }
+
+    // Use backPressureTimeout value <= 0  for no back pressure timeout
+    public boolean callProcedure(long backPressureTimeout, String proc, Object... fieldList) {
```
nshi: This method seems largely duplicating the functionality of `ClientInterface.handleRead()`. Is it worth considering refactoring that to make it reusable?

src/frontend/org/voltdb/InternalConnectionHandler.java
nshi: I agree with @vtkstef. It seems overkill to synchronize on the class. All importers on the node will contend on this lock.

Is the intention to create unique client handles for each invocation? If so, an atomic long will do. If ordering is also important, maybe having a single thread submitting the invocations that need to be ordered is enough.

Unless there is some unsafe global state that `m_adapter.createTransaction()` is modifying, I don't think synchronizing on the class is necessary.

src/frontend/org/voltdb/InternalConnectionHandler.java
manjujames: I will look into it. A actually didn't look at these carefully because I was simply moving it from a different class into this new class.

src/frontend/org/voltdb/InternalConnectionHandler.java
manjujames: Again, just something I refactored into this class from an existing one. I will look into this as well.

src/frontend/org/voltdb/InternalConnectionHandler.java
manjujames: Will look into this as well (was refactored into this new class from an existing one, so didn't pay that much attention).

src/frontend/org/voltdb/RealVoltDB.java
manjujames: If it is started with no limits, then it won't monitor it. I hadn't realized that you can change these after the server has been started. Will change this to take that into account.

src/frontend/org/voltdb/RealVoltDB.java
manjujames: This is starting it only if there are limits set (see the if condition).

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: OK.

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: OK. I was making it clear that this is multiplied by 1G since it is done just once.

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: OK.

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: OK.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
manjujames: This is not overwriting SystemSettings. It is only creating memory limit and setting it on SystemSettings.

src/frontend/org/voltdb/InternalConnectionHandler.java
manjujames: Ah, looks like the original class had changed to use an AtomicLong since I had moved the code to this one!

src/frontend/org/voltdb/utils/SystemStatsCollector.java
```diff
@@ -43,6 +42,9 @@
 
     private enum GetRSSMode { MACOSX_NATIVE, PROCFS, PS }
 
+    // Used by tests only to set up fake system statistics numbers
```
vtkstef: why is there a test artifact in the mainline source? Can you use a mock stat collector instead. I don't want mainline code depend in any way fashion on test artifacts like mock libraries and and junit

manjujames: This is not depending on mockito or junit libraries. I will change the class name to make it clear.

src/frontend/org/voltcore/logging/VoltLogger.java
```diff
@@ -348,4 +351,13 @@ public VoltLogger(String classname) {
         // set the final variable for the core logger
         m_logger = tempLogger;
     }
+
+    public void rateLimitedLog(long suppressInterval, Level level, Throwable cause, String format, Object...args) {
```
akhanzode: Can this be totally static with logger passed in?

src/frontend/org/voltdb/InternalConnectionHandler.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public final static long SUPPRESS_INTERVAL = 60;
+    private static final VoltLogger m_logger = new VoltLogger("InternalConnectionHandler");
+
+    // Atomically allows the catalog reference to change between access
+    private final AtomicLong m_failedCount = new AtomicLong();
+    private final AtomicLong m_submitSuccessCount = new AtomicLong();
+    private final InternalClientResponseAdapter m_adapter;
+
+    private static final long MAX_PENDING_TRANSACTIONS = Integer.getInteger("IMPORTER_MAX_PENDING_TRANSACTIONS", 400);
```
akhanzode: this is per partition now IMPORTER_MAX_PENDING_TRANSACTION_PER_PARTITION

src/frontend/org/voltcore/logging/VoltLogger.java
manjujames: I think it's easier this way -
m_logger.rateLimitedLog(...);
vs.
VoltLogger.rateLimitedLog(m_logger, ...);

src/frontend/org/voltdb/InternalConnectionHandler.java
manjujames: OK, thanks.

akhanzode:
We should log the memory limit if enabled and its values.
Do the limit numbers take 1g 1t or any such abbreviations?

akhanzode:
squash merged.

PR #2616, opened 2015-07-27
ENG-8679: Fix dropping of index on a view

hhakimi53:
Updates to HSQL-DB logic to support dropping of index on a view.

### DIFF
--- a/tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
+++ b/tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
@@ -146,4 +146,111 @@ public class TestAdhocCreateDropIndex extends AdhocDDLTestBase {
             teardownSystem();
         }
     }
+
+    public void testCreateDropIndexonView() throws Exception
+    {
+        String pathToCatalog = Configuration.getPathToCatalogForTest("adhocddl.jar");
+        String pathToDeployment = Configuration.getPathToCatalogForTest("adhocddl.xml");
+
+        VoltProjectBuilder builder = new VoltProjectBuilder();
+        builder.addLiteralSchema(
+                "create table FOO (" +
+                "ID integer not null," +
+                "VAL bigint, " +
+                "constraint PK_TREE primary key (ID)" +
+                ");\n" +
+                "create table FOO_R (" +
+                "ID integer not null," +
+                "VAL bigint, " +
+                "constraint PK_TREE_R primary key (ID)" +
+                ");\n"
+                );
+        builder.addPartitionInfo("FOO", "ID");
+        builder.setUseDDLSchema(true);
+        boolean success = builder.compile(pathToCatalog, 2, 1, 0);
+        assertTrue("Schema compilation failed", success);
+        MiscUtils.copyFile(builder.getPathToDeployment(), pathToDeployment);
+
+        VoltDB.Configuration config = new VoltDB.Configuration();
+        config.m_pathToCatalog = pathToCatalog;
+        config.m_pathToDeployment = pathToDeployment;
+
+        try {
+            startSystem(config);
+
+            // create a basic view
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
```diff
@@ -146,4 +146,111 @@ public void testBasicCreateIndex() throws Exception
...
+                "VAL bigint, " +
+                "constraint PK_TREE primary key (ID)" +
+                ");\n" +
+                "create table FOO_R (" +
+                "ID integer not null," +
+                "VAL bigint, " +
+                "constraint PK_TREE_R primary key (ID)" +
+                ");\n"
+                );
+        builder.addPartitionInfo("FOO", "ID");
```
wolffcm: This line seems like it's leftover from older versions of the product when ad hoc DDL was less supported than it is now.  In recent VoltDB versions it's preferable just to include
```SQL
partition table foo on column id;
```
in the literal schema above.

tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
```diff
@@ -146,4 +146,111 @@ public void testBasicCreateIndex() throws Exception
...
+        builder.setUseDDLSchema(true);
```
wolffcm: I believe that "use DDL schema" is now the default, but you can verify this by removing this line and making sure the test still passes.

tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
```diff
@@ -146,4 +146,111 @@ public void testBasicCreateIndex() throws Exception
             teardownSystem();
         }
     }
+
+    public void testCreateDropIndexonView() throws Exception
+    {
+        String pathToCatalog = Configuration.getPathToCatalogForTest("adhocddl.jar");
+        String pathToDeployment = Configuration.getPathToCatalogForTest("adhocddl.xml");
+
+        VoltProjectBuilder builder = new VoltProjectBuilder();
```
wolffcm: Since this code is nearly identical to the code above, consider refactoring it to its own method (e.g, `createSchema`) called from both tests?

tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
```diff
@@ -146,4 +146,111 @@ public void testBasicCreateIndex() throws Exception
...
+            try {
+                m_client.callProcedure("@AdHoc",
+                        "drop index VALDEX if exists;");
+            }
+            catch (ProcCallException pce) {
+                pce.printStackTrace();
+                fail("Should be able to drop a bad index with if exists");
+            }
+            assertFalse(findIndexInSystemCatalogResults("VALDEX"));
+        }
```
wolffcm: I think you should try to add and drop another index whose only component is `total`---seems like the ticket was complaining about dropping an index on an aggregate column specifically.

wolffcm:
That's all the review comments I have for this round.

tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
hhakimi53: Sure, will do that as it will provide more coverage. Also will make changes as highlighted above.

PS: Issue of dropping index is not limited to indexes created on aggregate functions only; it is also seen if index is created on standalone columns too. I will update the ticket for the same. Thanks.

tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
hhakimi53: By default use DDL schema is set to false, so we have to set it explicitly.

PR #2619, opened 2015-07-27
ENG-8541 follow up: license check for using active-active DR

slowsoul:
Add license check logic for changing DR mode to active-active

### DIFF
--- a/src/frontend/org/voltdb/Inits.java
+++ b/src/frontend/org/voltdb/Inits.java
@@ -374,7 +374,6 @@ public class Inits {
 
             String result = CatalogUtil.checkLicenseConstraint(catalog, m_rvdb.getLicenseApi());
             if (result != null) {
-                hostLog.fatal(result);
                 VoltDB.crashLocalVoltDB(result);
             }
 
@@ -384,7 +383,6 @@ public class Inits {
             // a starter context which uses a placeholder catalog.
             result = CatalogUtil.compileDeployment(catalog, m_deployment, false);
             if (result != null) {
-                hostLog.fatal(result);
                 VoltDB.crashLocalVoltDB(result);
             }
 
--- a/src/frontend/org/voltdb/Inits.java
+++ b/src/frontend/org/voltdb/Inits.java
@@ -372,11 +372,17 @@ public class Inits {
             catalog.execute(serializedCatalog);
             serializedCatalog = null;
 
+            String result = CatalogUtil.checkLicenseConstraint(catalog, m_rvdb.getLicenseApi());
+            if (result != null) {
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/Inits.java
```diff
@@ -372,11 +372,17 @@ public void run() {
             catalog.execute(serializedCatalog);
             serializedCatalog = null;
 
+            String result = CatalogUtil.checkLicenseConstraint(catalog, m_rvdb.getLicenseApi());
+            if (result != null) {
+                hostLog.fatal(result);
```
nshi: You can omit the fatal log here. `VoltDB.crashLocalVoltDB(result);` will print the fatal message.

src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
```diff
@@ -55,10 +56,16 @@
 
     // accept work via this mailbox
     Mailbox m_mailbox;
+    private LicenseApi m_licenseApi;
```
nshi: This should be a `final` field.

src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
```diff
@@ -55,10 +56,16 @@
...
+
+    public AsyncCompilerAgent(LicenseApi licenseApi) {
+        m_licenseApi = licenseApi;
+        m_helper = new AsyncCompilerAgentHelper(m_licenseApi);
+    }
 
     // The helper for catalog updates, back after its exclusive three year tour
     // of Europe, Scandinavia, and the sub-continent.
-    AsyncCompilerAgentHelper m_helper = new AsyncCompilerAgentHelper();
+    AsyncCompilerAgentHelper m_helper;
```
nshi: This should also be a `final` field.

src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
nshi: Since the `m_licenseApi` is not used in this class, you don't need to cache it here.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -36,6 +37,11 @@
 public class AsyncCompilerAgentHelper
 {
     private static final VoltLogger compilerLog = new VoltLogger("COMPILER");
+    private LicenseApi m_licenseApi;
```
nshi: Make it `final`.

PR #2626, opened 2015-07-27
ENG-6422 make guava TypeToken compilable with Java 8

vtkstef:


vtkstef:
R.by: @akhanzode

akhanzode:
Can you also run this branch through java8 and test jdbcloader which is only user of type token. There is a unit test for it but that runs under 1.7 or 1.8 not 1.6

vtkstef:
yes TestJDBCLoader works on java8

vtkstef:
passes jenkins running on java8 machines

PR #2628, opened 2015-07-27
 ENG-8609: Fix snapshot converter generates different row count every…

JadenW:
… time

It seems that the conversion ended incorrectly if one ConverterThread get empty content.

Manually test that now the converter will not produce the empty/truncated csv file.


akhanzode:
Please write a test for this now that the utility is documented and available.

PR #2630, opened 2015-07-28
Clear existing children when connecting subqueries to plan

wolffcm:
When I was writing some planner tests, I was encountering failed
assertions because a scan plan node had more than one child.  I
traced the cause to the method `connectChildrenBestPlans` which seems
to have been added as part of in/exists/scalar subquery support.

I wasn't able to recreate this issue using sqlcmd (I tried both ad hoc
and procedures), but seems like this issue might manifest in a
real-world scenario?

The added test was failing assertions in the planner before the fix.

### DIFF
--- a/src/frontend/org/voltdb/planner/PlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/PlanAssembler.java
@@ -812,6 +812,7 @@ public class PlanAssembler {
                 assert (betsCostPlan != null);
                 AbstractPlanNode subQueryRoot = betsCostPlan.rootPlanGraph;
                 subQueryRoot.disconnectParents();
+                scanNode.clearChildren();
                 scanNode.addAndLinkChild(subQueryRoot);
             }
         } else {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -812,6 +812,7 @@ private AbstractPlanNode connectChildrenBestPlans(AbstractPlanNode parentPlan) {
                 assert (betsCostPlan != null);
                 AbstractPlanNode subQueryRoot = betsCostPlan.rootPlanGraph;
                 subQueryRoot.disconnectParents();
```
BillWhite: While we are about it, should this be bestCostPlan, not betsCostPlan?

BillWhite:
With one typo comment, this seems right to me.  I'm not 100% sure I understand it, though.

paulmartel:
Looks like progress to me. Good to go.

wolffcm:
Thanks guys.  Something weird is happening in Jenkins, but I'll merge once it looks good.

PR #2631, opened 2015-07-28
Remove unnecessary synchronization/concurrency handling from DR path

afrogers:


src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -92,7 +92,14 @@ public static PartitionDRGateway getInstance(int partitionId,
         } catch (IOException e) {
             VoltDB.crashLocalVoltDB(e.getMessage(), false, e);
         }
-        m_partitionDRGateways.put(partitionId,  pdrg);
+
+        synchronized (m_gatewayInitLock) {
```
nshi: The synchronization here is unnecessary. `getInstance()` is called serially during initialization now.

PR #2632, opened 2015-07-28
do not count in the second boolean parameter as a differentiator to g…

jxchinaren:
…enerate unique id

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
@@ -2688,10 +2688,11 @@ public class Expression implements Cloneable {
             final List<String> id_list = new ArrayList<>();
             new Object() {
                 public void traverse(Expression exp) {
+                    assert(exp != null);
                     if (exp instanceof ExpressionAggregate) {
                         // do not count in the second boolean child for aggregate expression
                         // introduced from hsqldb 2.3.2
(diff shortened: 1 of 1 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -2688,9 +2688,16 @@ protected String voltGetUniqueId(final Session session) {
             final List<String> id_list = new ArrayList<>();
             new Object() {
                 public void traverse(Expression exp) {
-                    for (Expression expr : exp.nodes) {
-                        if (expr != null)
-                            id_list.add(expr.voltGetUniqueId(session));
+                    if (exp instanceof ExpressionAggregate) {
+                        // do not count in the second boolean child for aggregate expression
+                        // introduced from hsqldb 2.3.2
+                        if (exp != null && exp.nodes.length > 0)
```
paulmartel: The (exp == null) case was not handled in the original code, you don't need to guard for it now -- UNLESS you are trying to fix a new NPE.

paulmartel:
End of review round.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
jxchinaren: Actually, "exp" is never supposed to be NULL ever. I can remove this redundant check and turn it into an assert.

PR #2635, opened 2015-07-29
Unify the semantic checking for testing legality of indices and

BillWhite:
Unify the semantic checking for testing legality of indices and materialized views.

Jira Ticket Number: ENG-8632

### DIFF
--- a/src/frontend/org/voltdb/expressions/AbstractValueExpression.java
+++ b/src/frontend/org/voltdb/expressions/AbstractValueExpression.java
@@ -71,4 +71,9 @@ public abstract class AbstractValueExpression extends AbstractExpression {
         // Nothing to do... This is all about pulling types UP from child expressions.
     }
 
+    @Override
+    protected boolean containsFunctionById(int functionId) {
+        return false;
+    }
+
 }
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/expressions/AbstractValueExpression.java
```diff
@@ -71,4 +71,9 @@ public void finalizeValueTypes()
         // Nothing to do... This is all about pulling types UP from child expressions.
     }
 
+    @Override
+    protected boolean containsFunctionById(int functionId) {
+        return false;
+    }
+
```
paulmartel: Interesting creative, elegant even, response to the review comments.
Chris and I have debated the related issue --
the relative merits of an abstract class like AbstractExpression "having knowledge of / acting on behalf of" its subclasses as a convenience, a kind of isolation of the solution, vs. delegating ANY kind of type-specific behavior via method calls, more of an OO purist approach.
From my perspective, there are clear wins and losses for each side of the debate, and lots of murky middle ground cases. I wish I had better guiding principles or metrics or something to rely on for making these kinds of judgements.

BTW, this changeset looks good to go. Thanks.

PR #2641, opened 2015-07-30
Support for APPROX_COUNT_DISTINCT aggregate function

wolffcm:
There's changes to HSQL, the planner and the EE here.

Also there's a test app that I used for the innovation week demo.


src/ee/common/ValueFactory.hpp
```diff
@@ -92,6 +92,13 @@ class ValueFactory {
         return NValue::getAllocatedValue(VALUE_TYPE_VARBINARY, reinterpret_cast<const char*>(rawBuf), (size_t)rawLength, NValue::getTempStringPool());
     }
 
+    /// Constructs a varbinary value copied into temporary string
+    /// pool.  Arguments provide a pointer to the raw bytes and the
+    /// size of the value.
+    static inline NValue getTempBinaryValue(const char* rawBuf, int32_t rawLength) {
+        return NValue::getAllocatedValue(VALUE_TYPE_VARBINARY, rawBuf, (size_t)rawLength, NValue::getTempStringPool());
+    }
```
paulmartel: minor:  this method could be called to simplify line 92.

src/ee/common/ValuePeeker.hpp
```diff
@@ -131,6 +131,35 @@ class ValuePeeker {
...
+        case VALUE_TYPE_TIMESTAMP:
+        case VALUE_TYPE_DECIMAL:
+        case VALUE_TYPE_BOOLEAN:
+            *length = static_cast<int32_t>(NValue::getTupleStorageSize(vt));
+            return value.m_data;
+
+        case VALUE_TYPE_VARCHAR:
+        case VALUE_TYPE_VARBINARY:
+            *length = static_cast<int32_t>(value.getObjectLength_withoutNull());
+            return static_cast<const char*>(value.getObjectValue_withoutNull());
```
paulmartel: Minor inconsistency: we bother to add the special case code for VALUE_TYPE_VARCHAR and VALUE_TYPE_VARBINARY (vs. just asserting false) even though the sole caller has previously asserted that these are non-cases.

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
```diff
@@ -373,6 +379,24 @@ public void addAggregate(ExpressionType aggType,
...
+        // Create a new aggregate expression which we'll use to update the
+        // output schema (whose exprs are TVEs).
+        AggregateExpression aggExpr = new AggregateExpression(aggType);
+        aggExpr.finalizeValueTypes();
+
+        int outputSchemaIndex = m_aggregateOutputColumns.get(index);
+        SchemaColumn schemaCol = m_outputSchema.getColumns().get(outputSchemaIndex);
+        AbstractExpression schemaExpr = schemaCol.getExpression();
+        schemaExpr.setValueType(aggExpr.getValueType());
+        schemaExpr.setValueSize(aggExpr.getValueSize());
```
paulmartel: Nice catch -- the redundancy of SchemaColumn that forces this kind of upkeep makes me sad.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1522,6 +1534,9 @@ public void collectAllColumnExpressions(HsqlList set) {
         prototypes.put(OpTypes.MATCH_UNIQUE_FULL,    null); // not yet supported ExpressionLogical
         // aggregate functions
         prototypes.put(OpTypes.COUNT,         (new VoltXMLElement("aggregation")).withValue("optype", "count"));
+        // A VoltDB extension APPROX_COUNT_DISTINCT
```
paulmartel: This is a large block of volt code.
VoltDB extension annotations should not be nested --  I'm pretty sure this confuses my scripts.

src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
```diff
@@ -458,6 +467,20 @@ static Type getType(int setType, Type type) {
...
+                switch (dataType) {
+                case Types.TINYINT :
+                case Types.SQL_SMALLINT :
+                case Types.SQL_INTEGER :
+                case Types.SQL_BIGINT :
+                case Types.SQL_DECIMAL :
+                case Types.SQL_TIMESTAMP :
+                    return Type.SQL_DOUBLE;
+                default:
+                    throw Error.error(ErrorCode.X_42565);
```
paulmartel: Add similar comment paraphrasing the error text.

paulmartel: May want a small comment here justifying why not supporting FLOAT/VARCHAR/VARBINARY -- that may provide guidance to the poor soul who tries to add a new data type to Volt -- like, should I expect to be able to APPROX_COUNT_DISTINCT my new geohash?

src/hsqldb19b3/org/hsqldb_voltpatches/Tokens.java
```diff
@@ -100,6 +100,9 @@
     static final String        T_CORR             = "CORR";
     static final String        T_CORRESPONDING    = "CORRESPONDING";
     static final String        T_COUNT            = "COUNT";
+    // A VoltDB extension APPROX_COUNT_DISTINCT
+    static final String        T_APPROX_COUNT_DISTINCT = "APPROX_COUNT_DISTINCT";
+    // End of VoltDB extension
```
paulmartel: In the past, I have chalked up disruption of alphabetical order as a cultural "not raised on ABCs" thing. There goes that theory. Anyway, Let's straighten this out and make Big Bird proud of us.

tests/frontend/org/voltdb/planner/TestPlansApproxCountDistinct.java
```diff
@@ -0,0 +1,400 @@
...
+    // Find all the aggregate nodes in a fragment, whether they are hash, serial or partial.
+    private static List<AbstractPlanNode> findAllAggPlanNodes(AbstractPlanNode node) {
+        List<AbstractPlanNode> aggNodes = node.findAllNodesOfType(PlanNodeType.AGGREGATE);
+        List<AbstractPlanNode> hashAggNodes = node.findAllNodesOfType(PlanNodeType.HASHAGGREGATE);
+        List<AbstractPlanNode> partialAggNodes = node.findAllNodesOfType(PlanNodeType.PARTIALAGGREGATE);
+
+        aggNodes.addAll(hashAggNodes);
+        aggNodes.addAll(partialAggNodes);
+        return aggNodes;
+    }
```
paulmartel: Is this an original helper method or copy/pasted from some other agg planner test -- seems like a candidate for the PlannerTestCase toolbox.

tests/frontend/org/voltdb/planner/TestPlansApproxCountDistinct.java
```diff
@@ -0,0 +1,400 @@
...
+                AGGREGATE_VALS_TO_HYPERLOGLOG,
+                AGGREGATE_COUNT);
+
+        // With an agg that can be pushed down, but only because its argument is a partition key.
+        // Also, with approx count distinct with partition key as argument.
+        pn = compileToFragments(
+                "SELECT desc, approx_count_distinct(id), count(distinct id) "
+                + "from P1 "
+                + "group by desc");
+        assertFragContainsAggWithFunctions(pn.get(COORDINATOR_FRAG),
```
paulmartel: I am surprised that we follow through with approx_count_distinct in this known non-overlap case vs. falling back to count(distinct) or (if we had reason -- that I don't know -- to prefer approx_count_distinct on the partitions) pushing down the whole approx_count_distinct sending back the float result and summing the float on the coordinator. What would be the tradeoff for either of these alternatives. This is not a stop-ship issue, but if I'm right, we should have a follow-on special case optimization ticket for when the user perhaps out of habit or for uniformity asked for approx in this edge case where it would be easier to give exact values.

wolffcm: This is an interesting idea.

If the field being distinctly counted is a partition key and not governed by a primary key or unique constraint, then we know that the values from each partition will not overlap.  However, counter-intuitively, getting the approximate count from each partition, and then summing on the coordinator would lead to a *less* accurate result, since the accuracy of the approximation increases with the number of unique values you add to the hyperloglog.  

In this situation the approximation has the benefit of only constant memory being used on the partitions, since `count(distinct id)` would need to use a hash table to remove duplicates, sized linearly with the number of distinct values.

If the user requests `APPROX_COUNT_DISTINCT` on a unique column, then we can compute an exact answer in linear time and constant space.  In this case we might consider doing so, but using `COUNT` would produce a BIGINT instead of the expected FLOAT value.  This could be something of a surprise?

My preference would be to encourage users to think of `APPROX_COUNT_DISTINCT` as a statistical function that uses a particular algorithm to get its answer, like `STDDEV`.   I wouldn't want users to see that we're giving exact results in some situations and coming to rely on it elsewhere.

I'll file a ticket for this.  It might be a good topic to discuss at a grooming.

src/ee/common/ValuePeeker.hpp
wolffcm: That's true.  I was trying to produce a function that would work for all data types, even though the current use doesn't need them all.

I guess trying to be proactive like this results in unreachable, untested code, which isn't good, so I'll delete the unused types and replace with an assert.

tests/frontend/org/voltdb/planner/TestPlansApproxCountDistinct.java
wolffcm: Actually, I spoke too soon about adding estimates produced by hyperloglogs containing non-overlapping values.  I did a quick experiment with the phone numbers produced by the Voter example, and the estimate got more accurate.  I'll include details in the ticket.

src/hsqldb19b3/org/hsqldb_voltpatches/Tokens.java
```diff
@@ -1652,6 +1658,9 @@
         reservedKeys.put(Tokens.T_CORR, CORR);
         reservedKeys.put(Tokens.T_CORRESPONDING, CORRESPONDING);
         reservedKeys.put(Tokens.T_COUNT, COUNT);
+        // A VoltDB extension APPROX_COUNT_DISTINCT
+        reservedKeys.put(Tokens.T_APPROX_COUNT_DISTINCT, APPROX_COUNT_DISTINCT);
+        // End of VoltDB extension
```
paulmartel: One last alphabet miss.

tests/frontend/org/voltdb/planner/TestPlansApproxCountDistinct.java
paulmartel: Valid points. BTW, is he float return type for APPROX_DISTINCT i"industry standard"? It also surprised me just a little.

tests/frontend/org/voltdb/regressionsuites/TestApproxCountDistinctSuite.java
```diff
@@ -0,0 +1,428 @@
...
+
+        verifyStmtFails(client,
+                "select approx_count_distinct(vc_inline) from unsupported_column_types;",
+                "incompatible data type in operation");
+
+        verifyStmtFails(client,
+                "select approx_count_distinct(vb_inline) from unsupported_column_types;",
+                "incompatible data type in operation");
+
+        verifyStmtFails(client,
```
paulmartel: Add specific comment here -- float is a fixed-width type disqualified for its own reason.

third_party/cpp/murmur3/MurmurHash3.cpp
```diff
@@ -31,11 +32,17 @@ namespace voltdb {
 
 #define	FORCE_INLINE inline __attribute__((always_inline))
 
+inline uint32_t rotl32 ( uint32_t x, int8_t r )
```
paulmartel: I'm guessing all this new 32 code with the eye-straining two-byte indent -- SO wasteful, why double your whitespace when 1-byte indenting is almost exactly as effective -- came from manually tweaking the 64 code
rather than any kind of upstream version upgrade?
A comment somewhere at the top taking/delegating credit for these parts of the code might be of historical interest. It doesn't have to be formal.
Apparently, nobody maintaining other third party code here has been swayed by the elegance of the 
hsql "// Voltdb extension" annotation scheme. Their loss.

paulmartel: Actually, looking at the apparent "originality" of some of this added code has me seriously wondering where it came from -- apparently some strange land where comments are not valued. Or maybe from a book or published paper where all is explained?

tests/frontend/org/voltdb/planner/TestPlansApproxCountDistinct.java
wolffcm: Oops.  No, it's not standard.  Other vendors who have a function with this name return an integer.

Choosing to produce a float was a decision I made early on, and it never occurred to me that other vendors might produce something else.  I like it because it emphasizes the probabilistic nature of the result.  But maybe the "APPROX_" prefix is enough.

I guess I'm leaning more towards BIGINT as the result type after some thought.  Do you have a preference?

third_party/cpp/murmur3/MurmurHash3.cpp
wolffcm: It's from here:

https://github.com/PeterScott/murmur3

At least I think it is.  I got the "32" changes from there.

It seems like we've changed the murmur3 API to return a value rather than write to a supplied pointer.  I don't know why.  I followed suit, for consistency's sake.

paulmartel: Thanks for the link. There should be an attribution like this added somewhere in a header comment block.
Also a source comment characterizing any in-house customizations to date, as you just did, might be helpful.
I'm not asking (yet?) for hsql-style canonical annotation comments on all our changes to third party sources.
A little self-policing can ward off the heavy-handed regulators.

wolffcm:
I added some more commits for review comments, an issue I found with ORDER BY, and changing the return type to BIGINT.

src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
```diff
@@ -458,6 +467,30 @@ static Type getType(int setType, Type type) {
...
+                case Types.SQL_DECIMAL :
+                case Types.SQL_TIMESTAMP :
+                    return Type.SQL_BIGINT;
+                default:
+                    // We only support fixed-width types for this
+                    // aggregate function.
+                    //
+                    // FLOAT is not supported since this function
+                    // relies on different values having different bit
+                    // patterns, and the same values having being the
```
paulmartel: comment grammar "having being"

tests/test_apps/approx-count-distinct/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.class
+*.dat
+*.png
```
paulmartel: missing final newline

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -907,6 +907,44 @@ public boolean hasAnySubexpressionOfClass(Class< ? extends AbstractExpression> a
     }
 
     /**
+     * A predicate class for searching expression trees,
```
paulmartel: Are there other AbstractExpression tree-walking test functions that could be refactored into SubexprFinderPredicates? I seem to recall similar code in many places.

Another common tree-walking idiom is the one that collects qualifying AbstractExpressions in a list.
This COULD be another application of SubexprFinderPredicate, as in "walk the tree and if (predicate.matches(node)) { result.add(node); }

Some tree-walkers have short-circuit logic like "when looking for a node where pred.matches(node) don't bother looking in sub-trees of parent where pred2.matches(parent)" -- maybe we would continue to hard-code those if the optimization seemed worth it.

There are probably a number of candidate use cases for the simple hasAnySubexpressionWithPredicate that are disguised as use cases
of this other list-filler visitor, but that in the end ONLY do an isEmpty() on the resulting list.
hasAnySubexpressionWithPredicate would short-circuit these on the first match.

Given the deadline, maybe you've moved the ball far enough in this play, and
this further refactoring is something to do on the next go-around.

wolffcm:
I've pushed a commit that should make the test_apps example as easy to run as voter is.

wolffcm:
I'll squash and commit on a separate branch.

PR #2648, opened 2015-08-03
Eng 8289 snapshot blocks with large payload

luyangco:
This patch is for the problem that when disk I/O is busy(e.g. large transactions), client transaction per second (tps) might drop to 0 for a while when DR replica starts the sync snapshot, it recovers after the sync snapshot.
This patch addressed the problem by:
1) Don't syncing DR buffers to disk when non-truncation snapshot happens,
2) putting DR sync snapshot to the same directory of snapshot path defined in the deployment.


luyangco:
Send new PR after rebasing (https://github.com/VoltDB/voltdb/pull/2807), close this one.

PR #2649, opened 2015-08-04
Eng 8753 Kafka System Test

pdshaw:
Changes --

Import SP deletes matching rows from Mirror table

Test ends when mirror table is empty, meaning all tuples have made the round trip from client creation to VDB export table to Kafka and back again through VDB import, with successful data match along the way.

Test calculates and prints import TPS by tracking mirror table row delete rate.


tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
```diff
@@ -103,8 +104,8 @@
         @Option(desc = "Benchmark duration, in seconds.")
         int duration = 300;
 
-        @Option(desc = "Warmup duration in seconds.")
-        int warmup = 2;
```
prosegay: why did you take out the warmup?

tests/test_apps/kafkaimporter/deployment.xml
```diff
@@ -7,19 +7,19 @@
...
+            <property name="brokers">volt8a:9092,volt8b:9092,volt8c:9092</property>
+            <property name="topics">T312_KAFKAEXPORTTABLE1</property>
             <property name="procedure">InsertImport</property>
         </configuration>
     </import>
 
     <export enabled="true" target="kafka">
         <configuration>
-            <property name="metadata.broker.list">kafka2:9092</property>
+            <property name="metadata.broker.list">volt8a:9092,volt8b:9092,volt8c:9092</property>
```
akhanzode: Do u really want to point to volt8?

tests/test_apps/kafkaimporter/run.sh
```diff
@@ -160,8 +160,8 @@ function async-benchmark() {
     java -classpath $CLIENTCLASSPATH -Dlog4j.configuration=file://$LOG4J \
         kafkaimporter.client.kafkaimporter.KafkaImportBenchmark \
         --displayinterval=5 \
-        --warmup=10 \
-        --duration=60 \
+        --ratelimit=20000 \
+        --duration=15 \
         --servers=localhost
```
prosegay: does this make it a 15 second test by default? does that make sense (minor)

prosegay:
lets sqaush merge this when it's ready

tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
```diff
@@ -80,10 +81,10 @@
     static Client client;
     // Some thread safe counters for reporting
     AtomicLong linesRead = new AtomicLong(0);
-    AtomicLong rowsAdded = new AtomicLong(0);
+    static AtomicLong rowsAdded = new AtomicLong(0);
     static final AtomicLong finalInsertCount = new AtomicLong(0);
 
-    private static final int END_WAIT = 60000; // wait for 60 seconds at the end
+    private static final int END_WAIT = 10; // wait at the end for import to settle after export completes
```
prosegay: 10 microseconds?

tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
pdshaw: Didn't really do anything useful.

tests/test_apps/kafkaimporter/deployment.xml
pdshaw: Where should this point?

tests/test_apps/kafkaimporter/run.sh
pdshaw: Was set low because the export/import tps ratio is such that 15 seconds export could be a 10 minute import.

However it's looking forward to set it higher, both because import will get faster, and I've added rate limit, which allows balancing export and import.

tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
pdshaw: 10 seconds. I'll fix the comment -- it's the loop time, with the loop repeated until all the import activity is done.

prosegay:
Hi Peter, I think one thing we need to check is that if by some failure or error, zero rows get exported, that the test cannot return success. Please check that case. Then we will do a squash merge.

PR #2653, opened 2015-08-04
Eng 8645 beef up the non-determinism tests.

BillWhite:
I don't think this is merge worthy yet.  I think more testing is possible.  But perhaps this can be reviewed to see if I've gone overboard with the function ParsedSelectStatements.addHonoraryOrderByExpressions.


src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -289,8 +289,18 @@ public String toString() {
 
     private static final String INDENT = "  | ";
 
+    /**
+     * Return a node name to help out toString.  Subclasses
+     * can chime in if they have a notion to.  See TupleValueExpression,
+     * for example.
+     */
+    protected String getNodeNameForToString() {
```
wolffcm: Something that occurs to me about naming: I see "node" and I think "plan node," at least in our code base.  That could just be me though.

src/frontend/org/voltdb/expressions/TupleValueExpression.java
```diff
@@ -403,4 +403,23 @@ else if ( ! tableName.equals(impliedTableName)) {
...
+        } else if (alias != null) {
+            return String.format ("<> AS %s", alias);
+        } else {
+            return "<none>";
+        }
+    }
+
+    protected String getNodeNameForToString() {
+        return String.format("%s.%s",
+                             chooseTwoNames(m_tableName, m_tableAlias),
```
wolffcm: So this might print out `myTable AS mtbl.myCol AS col`?  Seems a little hard to read.

If there's an alias in the FROM clause of a statement, it replaces the original table name in the statement's scope, so maybe omit the original table name if there's a table alias?

Either way, it will be a good improvement to debug output.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1805,105 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
         return result;
     }
 
+    private void addHonoraryOrderByExpressions(
+            HashSet<AbstractExpression> orderByExprs,
+            List<ParsedColInfo> candidateColumns) {
+        HashMap<AbstractExpression, Set<AbstractExpression>> valueEquivalence = analyzeValueEquivalence();
```
wolffcm: Are we re-computing value equivalences for the whole statement each time we invoke this method?  Seems like we might want to cache this somewhere.

wolffcm: Upon further inspection, it looks like this is just called once so it may not matter.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1805,105 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
...
+                        if (addAllColumns) {
+                            // Don't forget to remember to forget the other indices.  (E. Presley, 1955)
+                            addedAllColumns = true;
+                            for (Column addCol : table.getColumns()) {
+                                // We have to convert this to a TVE to add
+                                // it to the orderByExprs.  We will use null
+                                // for the table and column aliases and -1
+                                // for the column index.  These will match
+                                // anything.
+                                TupleValueExpression ntve = new TupleValueExpression(tve.getTableName(),
```
wolffcm: Would this work for unique expression indexes?  I don't know if UNIQUE can be used to qualify an expression index or not.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1805,105 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
...
+                        }
+                        if (addAllColumns) {
+                            // Don't forget to remember to forget the other indices.  (E. Presley, 1955)
+                            addedAllColumns = true;
+                            for (Column addCol : table.getColumns()) {
+                                // We have to convert this to a TVE to add
+                                // it to the orderByExprs.  We will use null
+                                // for the table and column aliases and -1
+                                // for the column index.  These will match
+                                // anything.
```
wolffcm: I'm not sure I understand.  Will this work if you assign an alias to the table in the FROM clause?

Seems like you'd want to find the corresponding TVE in the candidate columns and just add that expression?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1805,105 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
...
+                                                                                     -1);
+                                orderByExprs.add(ntve);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
```
wolffcm: This method doesn't strike me as being overly slow.  How would you characterize its algorithmic complexity?  Seems like `O(n*m + n*k)`, where `n` is the number of candidate expressions, `m` is the average number of value equivalences for a candidate expression, and `k` is the number of unique indexes for each table used in the statement.  The other methods concerned with determinism seem to have similar complexity.

It seems like what this method is doing is finding all the columns that will have a deterministic ordering in the output.  Do I have this right?  In my mind this is a little different than an "honorary order by expression."  When this method returns, it won't be accurate to say "this statement produces results ordered by `orderByExprs`".  The output won't be ordered by the unique indexes you use to infer determinism.

Method naming aside, what you have here looks good to me overall.

You could imagine getting more sophisticated here, by looking for expressions that are composed of other deterministic expressions, and things like that---if you know that `a` and `b` have a deterministic ordering, then `a * b` is also okay, and so on.  Just mentioning this as possible future work.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: Ok.  I'll change it.

src/frontend/org/voltdb/expressions/TupleValueExpression.java
BillWhite: I was thinking, since it's for debugging, it would be better to see both the defined name and the alias.  I don't really care that much about the scope rules in this context, I just want to know what the expression is.  I think we can usually tell which expression is being worked on from the names of the leaves.  But just printing TupleValueExpression[VALUE_EXPRESSION] is not all that expressive.  I suppose I was thinking that if we had two tables, with different aliases, it might be useful to show the aliases, so we could quickly tell which table was which.  I tried the " AS " notation to try to mimic the SQL notation.  I don't have any aliases in my problem now, so I don't know if it's useful or just confusing.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: I think I need to compute it for each select statement, or at least for each select list, which is the same thing.  Otherwise, I don't know which tables and order-by expressions are at issue.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: I had thought not.  But apparently it's possible.  The documentation is incorrect about this.

Apparently it's not possible to create primary keys with expressions, but unique indices both
defined in the table creation or else defined separately can have expressions.  The documentation is unclear here as well.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: I was thinking you wanted to compare columns, independent of their aliases.  I was thinking a column is denoted unambiguously by it's table name and column name.  However, maybe table aliases are the right way to go with this.  The problem is self joins, I think.  I'll think about it harder.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1802,129 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
...
+                }
+                // If one of the indices is completely covered
+                // we will not have to process any other indices.
+                // So, we remember this and early-out.
+                boolean addedAllColumns = false;
+                for (Constraint constraint : indices) {
+                    if (addedAllColumns) {
+                        break;
+                    }
+                    Index index = constraint.getIndex();
```
wolffcm: There are constraints that are not enforced via indexes, like the LIMIT PARTITION ROWS constraint.  is getIndex() guaranteed to return a non-null value?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1802,129 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
...
+        // same table scan.  If we see a table which has a different name from
+        // the other tables, or which has no table name, since it is a sub-query,
+        // then we have nothing else to do here.  We need to finish the entire
+        // select list, even if we see multiple tables.
+        String scanName = null;
+        String scanAlias = null;
+        boolean earlyOut = false;
+        for (ParsedColInfo colInfo : candidateColumns) {
+            AbstractExpression colExpr = colInfo.expression;
+            if (colExpr instanceof TupleValueExpression) {
```
wolffcm: If the objective is not to proceed to the unique index step if there are multiple tables in the FROM clause for the statement, wouldn't it be simplest just to check that m_tableList (in AbstractParsedStmt) has just one item?

BillWhite: Yes.  I think the weight of the evidence is that I need to rethink this all.  I need to somehow merge together addHonoraryOrderByExpressions and orderByColumnsCoverUniqueKeys.

BillWhite: I'm going to close this pull request and try to combine
addHonoraryOrderByExpressions and orderByColumnsCoverUniqueKeys.  I'll also
verify that DELETE FROM SELECT has similar determinism semantics as SELECT.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1799,6 +1802,129 @@ private static boolean orderByColumnsDetermineAllColumns(List<ParsedColInfo> ord
...
+                    if (addAllColumns) {
+                        // Don't forget to remember to forget the other indices.  (E. Presley, 1955)
+                        addedAllColumns = true;
+                        for (Column addCol : table.getColumns()) {
+                            // We have to convert this to a TVE to add
+                            // it to the orderByExprs.  We will use null
+                            // for the table and column aliases and -1
+                            // for the column index.  These will match
+                            // anything.
+                            TupleValueExpression ntve = new TupleValueExpression(tve.getTableName(),
```
wolffcm: This code still seems wrong to me.  A table alias (in the FROM clause) should replace the referred table in the scope of the statement.  For example, this is not a valid statement, because the name "foo" has been superseded by "fooalias":
```SQL
select foo.* from foo as fooalias
```
So the table alias seems important when you're creating TVEs to add to the honorary order by list, even for a statement that accesses a single table.  I think you need to make sure that the table alias of the TVE you create here matches whatever the table's alias is for the statement.  And it would be good to have tests that exercise this.

BillWhite:
This need more thought.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: I think this is probably wrong over all.  It will be right if there is only
one table, though.  The table names are the same, foo, the aliases are
different, foo and fooalias, but you don't really care, because there's
only one table.

But I think you are right overall.  I didn't understand how m_tableAliasMap
works, and I may still not understand it.

BillWhite: Thanks for all this review.  It's really helpful.

PR #2655, opened 2015-08-05
ENG-8641 Add max and indexes to the benchmark

yzhang1991:
Let the benchmark cover both MIN and MAX.
Let optimal and sub-optimal cases have same number of indexes with similar structure.


paulmartel:
Looks good.

PR #2661, opened 2015-08-05
Vmc 223 fix sql queries

johnTcrawford:
Fixes getDefaultStoredProcedures & getUserStoredProcedures, correctly updating the Navigators they rely on, which were changed by VMC-210; this will fix several SqlQueriesTest methods, as well as allowing FullDdlSqlTest to clean up for itself properly (deleting Stored Procs and Tables it created). And simplifies these methods, plus getSystemStoredProcedures, by having them call a new private method, getSpecifiedStoredProcedures. Also fixes a problem (introduced by my fix for ENG-8509) where an unexpected error causes an exception, rather than a graceful comparison failure.


johnTcrawford:
VMC-223-squashed-sql-queries, which is the squashed version of this branch, was just merged to master, so this branch is obsolete.

PR #2662, opened 2015-08-05
Eng 6685 add insert into to txnid selfcheck app

liam700:
Altered the CopyLoadPartitioned procs to use insert into instead of a select followed by an insert, as the select and insert sql commands are tested in many places besides this. Took random number generation out of stored procedure, so that the hash check does not fail.


prosegay:
this has been merged manually, close this pull request

liam700:
merged manually

PR #2663, opened 2015-08-05
ENG-8723: snapshot hidden column (also ENG-8665)

slowsoul:
1. Make the serializer used in snapshot in EE include hidden columns by default (FullTupleSerializer).

2. Make table loading in EE in snapshot restore count for hidden columns (direct load for active-active, another ticket for restore passive to active DR).

3. Add support for hidden columns in frontend when DR is in active-active mode, by 1) augmenting the DR table schema data written along with each snapshotted tuple block during snapshot save. 2) augmenting target DR table schema during snapshot restore.

4. Added unit tests which goes over save/restore/save/compare process.

5. ENG-8665: When restoring a snapshot that does not contain hidden DR column to a Active DR table, add this column, set its value in all rows to NULL_BIGINT. When EE sees NULL_BIGINT in this column, it will correctly sets it based on the uniqueId of current snapshot restore transaction before loading the tuple into the table, otherwise EE will directly load it.

### DIFF
--- a/src/ee/common/FullTupleSerializer.h
+++ b/src/ee/common/FullTupleSerializer.h
@@ -24,6 +24,10 @@ namespace voltdb {
 class ReferenceSerializeOutput;
 class TupleSchema;
 
+/**
+ * FullTupleSerializer provides delegate methods to serialize visible and hidden columns
+ * of the given tuple. It also gives corresponding max serialization size for buffer allocation.
+ */
 class FullTupleSerializer : public TupleSerializer {
 public:
     /**
--- a/src/ee/common/FullTupleSerializer.h
+++ b/src/ee/common/FullTupleSerializer.h
@@ -15,8 +15,8 @@
  * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef DEFAULTTUPLESERIALIZER_H_
-#define DEFAULTTUPLESERIALIZER_H_
+#ifndef FULLTUPLESERIALIZER_H_
+#define FULLTUPLESERIALIZER_H_
 #include "common/TupleSerializer.h"
 #include "common/tabletuple.h"
 
@@ -24,7 +24,7 @@ namespace voltdb {
 class ReferenceSerializeOutput;
 class TupleSchema;
 
-class DefaultTupleSerializer : public TupleSerializer {
+class FullTupleSerializer : public TupleSerializer {
 public:
     /**
      * Serialize the provided tuple to the provide serialize output
@@ -36,8 +36,8 @@ public:
      */
     size_t getMaxSerializedTupleSize(const TupleSchema *schema);
 
-    virtual ~DefaultTupleSerializer() {}
+    virtual ~FullTupleSerializer() {}
 };
 }
 
-#endif /* DEFAULTTUPLESERIALIZER_H_ */
+#endif /* FULLTUPLESERIALIZER_H_ */
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -320,8 +320,6 @@ public:
     void serializeTo(voltdb::SerializeOutput &output, bool includeHiddenColumns = false);
     void serializeToExport(voltdb::ExportSerializeOutput &io,
                           int colOffset, uint8_t *nullArray);
-    void serializeAllColumnsToDR(ExportSerializeOutput &io,
-                          int colOffset, uint8_t *nullArray);
     void serializeToDR(voltdb::ExportSerializeOutput &io,
                        int colOffset, uint8_t *nullArray,
                        const std::vector<int>* interestingColumns);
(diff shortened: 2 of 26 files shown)


src/ee/common/FullTupleSerializer.h
```diff
@@ -0,0 +1,43 @@
...
+#ifndef FULLTUPLESERIALIZER_H_
+#define FULLTUPLESERIALIZER_H_
+#include "common/TupleSerializer.h"
+#include "common/tabletuple.h"
+
+namespace voltdb {
+class ReferenceSerializeOutput;
+class TupleSchema;
+
+class FullTupleSerializer : public TupleSerializer {
```
nshi: Probably worth explaining how this serializer is different from the default one in a comment here.

src/ee/common/tabletuple.h
```diff
@@ -785,6 +787,14 @@ inline void TableTuple::deserializeFrom(voltdb::SerializeInputBE &tupleIn, Pool
         NValue::deserializeFrom(tupleIn, dataPool, dataPtr, columnInfo->getVoltType(),
                 columnInfo->inlined, static_cast<int32_t>(columnInfo->length), columnInfo->inBytes);
     }
+
+    for (int j = 0; j < hiddenColumnCount; ++j) {
+        const TupleSchema::ColumnInfo *columnInfo = m_schema->getHiddenColumnInfo(j);
+
+        char *dataPtr = getWritableDataPtr(columnInfo);
+        NValue::deserializeFrom(tupleIn, dataPool, dataPtr, columnInfo->getVoltType(),
```
nshi: Is it possible that `tupleIn` does not have values for the hidden column but the table has? Will this line throw an exception if that's possible?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1686,7 +1686,7 @@ int64_t VoltDBEngine::tableStreamSerializeMore(
         char *ptr = reinterpret_cast<char*>(serializeIn.readLong());
         int offset = serializeIn.readInt();
         int length = serializeIn.readInt();
-        outputStreams.add(ptr + offset, length - offset);
+        outputStreams.add(ptr + offset, length);
```
nshi: Good catch! I can't believe we lived with this horrible bug for so long.

src/ee/storage/persistenttable.cpp
```diff
@@ -1152,6 +1156,13 @@ void PersistentTable::processLoadedTuple(TableTuple &tuple,
                                          size_t &tupleCountPosition,
                                          bool shouldDRStreamRows) {
     try {
+        // if the value passed from fronted is NULL, it means it's not yet initialized,
+        // so fill it with current DR timestamp
+        if (hasDRTimestampColumn() && tuple.getHiddenNValue(getDRTimestampColumnIndex()).isNull()) {
```
nshi: I wonder if it'll be easier to leave it in `insertTupleCommon()` and have `setDRTimestampForTuple()` update the value if it's NULL. That way, all insertions are guaranteed to run through the check and have the value populated and still works for the restore path.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -142,6 +142,8 @@
     static final String JSON_DATA_KEY = "data";
     static final String JSON_STATUS_KEY = "status";
 
+    public static final String DR_HIDDEN_COLUMN_NAME = "dr_clusterid_timestamp";
```
nshi: `VoltTable` is a public API, I prefer hiding this in a private API somewhere related to DR.

src/frontend/org/voltdb/sysprocs/saverestore/SavedTableConverter.java
```diff
@@ -35,10 +33,16 @@
...
 
     public static Boolean needsConversion(VoltTable inputTable,
-                                          Table outputTableSchema) {
+                                          Table outputTableSchema,
+                                          boolean shouldPreserveDRHiddenColumn) {
         if (inputTable.getColumnCount() != outputTableSchema.getColumns().size()) {
             return true;
         }
+        if (shouldPreserveDRHiddenColumn &&
+            !inputTable.getColumnName(inputTable.getColumnCount() - 1).equalsIgnoreCase(VoltTable.DR_HIDDEN_COLUMN_NAME)) {
```
nshi: If there are other hidden columns in the input table, this may break. A safer way to check is probably to try getting that column by name. If it doesn't find the column, then return `true`.

src/frontend/org/voltdb/sysprocs/saverestore/SavedTableConverter.java
```diff
@@ -60,15 +64,45 @@ public static Boolean needsConversion(VoltTable inputTable,
...
 
+        // if original table does not have hidden column present, we need to add
+        boolean addDRHiddenColumn = shouldPreserveDRHiddenColumn &&
+                !column_copy_index_map.containsKey(new_table.getColumnCount() - 1);
+        Column catalogColumnForHiddenColumn = null;
+        if (addDRHiddenColumn) {
+            catalogColumnForHiddenColumn = new Column();
+            catalogColumnForHiddenColumn.setName(VoltTable.DR_HIDDEN_COLUMN_NAME);
+            catalogColumnForHiddenColumn.setType(VoltType.BIGINT.getValue());
+            catalogColumnForHiddenColumn.setSize(Long.SIZE / 8);
```
nshi: You can call `VoltType.BIGINT.getLengthInBytesForFixedTypes()` to get this.

src/frontend/org/voltdb/sysprocs/saverestore/SavedTableConverter.java
```diff
@@ -60,15 +64,45 @@ public static Boolean needsConversion(VoltTable inputTable,
...
+            catalogColumnForHiddenColumn.setInbytes(false);
+            // small hack here to let logic below fill VoltType.NULL_BIGINT in for the hidden column
+            // actually this column is not nullable in EE, but it will be set to correct value before
+            // insert(restore) to the corresponding table
+            catalogColumnForHiddenColumn.setNullable(true);
```
nshi: What will happen if we don't create this column at all and pass the VoltTable directly down to the EE? Can the EE detect it and add the hidden column there?

src/frontend/org/voltdb/sysprocs/saverestore/SavedTableConverter.java
```diff
@@ -60,15 +64,45 @@ public static Boolean needsConversion(VoltTable inputTable,
...
+        // if the DR hidden column should be preserved in conversion, append it to the end of target schema
+        // TODO if a previous passive DR table is restored in active mode, do we provide default value to this
+        // TODO hidden column now or just fill with 0 and leave it to EE to fill this column if found
+        if (shouldPreserveDRHiddenColumn) {
+            int columnCount = new_table.getColumnCount();
+            VoltTable.ColumnInfo[] augmentedSchema = new VoltTable.ColumnInfo[columnCount + 1];
+            for (int i = 0; i < columnCount; i++) {
+                augmentedSchema[i] = new VoltTable.ColumnInfo(new_table.getColumnName(i), new_table.getColumnType(i));
+            }
+            augmentedSchema[columnCount] = new VoltTable.ColumnInfo(VoltTable.DR_HIDDEN_COLUMN_NAME, VoltType.BIGINT);
```
nshi: You can consider pulling the logic into `CatalogUtil.getVoltTable()`, so the `new_table` has the right schema of what you want without augmenting it here.

PR #2664, opened 2015-08-05
ENG-8803: Fix Batched DR after CREATE TABLE bug, add related tests

slowsoul:
In CatalogDiffEngine, if DR is set to active-active mode, demand all tables in the previous catalog instead of the current in-batch catalog to be checked to be empty.

It is because we can safely assume that DDL will not be mixed with SQL query in a batch submitted to @AdHoc, so tables newly created in the batch DDL are guaranteed to be empty and there is no need to check them.

The tests test for both active DR and passive DR for executing a batched DDL which set DR mode, create table, partition this table and finally DR this table.


nshi:
Merged externally

PR #2665, opened 2015-08-06
Eng 8455 orderby merge sort

alexeevm:


build.py
```diff
@@ -230,8 +230,9 @@
  indexscanexecutor.cpp
  insertexecutor.cpp
  limitexecutor.cpp
- materializedscanexecutor.cpp
  materializeexecutor.cpp
+ materializedscanexecutor.cpp
```
paulmartel: materializedscanexecutor.cpp got bumped slightly out of alphabetical order by the merge.

src/ee/executors/executorutil.cpp
```diff
@@ -57,6 +57,7 @@
 #include "executors/limitexecutor.h"
 #include "executors/materializeexecutor.h"
 #include "executors/materializedscanexecutor.h"
+#include "executors/mergereceiveexecutor.h"
```
paulmartel: I noticed another alphabetical glitch with materializedscanexecutor, though this one pre-existed.

src/ee/executors/executorutil.cpp
```diff
@@ -106,4 +116,31 @@ AbstractExecutor* getNewExecutor(VoltDBEngine *engine,
     return NULL;
 }
 
+bool TupleComparer::operator()(TableTuple ta, TableTuple tb) const
```
paulmartel: I see executorutil (as minimal as it has become) as really more "externally focused" for callers outside the AbstractExecutors class tree that encapsulates access TO the entirety of that class tree.

I'd have moved TupleComparer into abstractexecutor.h/cpp instead, possibly as a nested class of AbstractExecutor.
This avoids the need to propagate includes of executorutil.h

src/ee/executors/executorutil.cpp
```diff
@@ -106,4 +116,31 @@ AbstractExecutor* getNewExecutor(VoltDBEngine *engine,
...
+        {
+            if (cmp < 0) return true;
+            if (cmp > 0) return false;
+        }
+        else if (dir == SORT_DIRECTION_TYPE_DESC)
+        {
+            if (cmp < 0) return false;
+            if (cmp > 0) return true;
+        }
+        else
```
paulmartel: A pre-existing issue: 
This supposed handling of sort directions other than the allowed ones seems overly expensive. It should be handled ONCE when m_dirs is initialized -- and via an assert that gets compiled out of production code.
That would allow less branching here:

        int cmp = ...
        if (cmp < 0) return (dir == SORT_DIRECTION_TYPE_ASC);
        if (cmp > 0) return (dir == SORT_DIRECTION_TYPE_DESC);

src/ee/executors/mergereceiveexecutor.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace voltdb {
+
+    class TempTable;
+    class OrderByPlanNode;
+    class LimitPlanNode;
+    class AggregateExecutorBase;
+
+    /**
+     * The ORDER BY executor to be used at the coordinator node
```
paulmartel: replace "ORDER BY" with "optimized replacement for an ORDER BY"

src/ee/executors/mergereceiveexecutor.h
```diff
@@ -0,0 +1,87 @@
...
+     * to merge-sort results from multiple partitions. The assumption is
+     * that the individual partitions results are already sorted in the right order
```
paulmartel: replace "in the right order" with "the order specified by the inlined OrderByPlanNode".

src/ee/executors/orderbyexecutor.cpp
```diff
@@ -69,19 +70,22 @@ OrderByExecutor::p_init(AbstractPlanNode* abstract_node,
 
     OrderByPlanNode* node = dynamic_cast<OrderByPlanNode*>(abstract_node);
     assert(node);
-    assert(node->getInputTableCount() == 1);
 
-    assert(node->getChildren()[0] != NULL);
+    if (!node->isInline()) {
```
paulmartel: The inlined case should have its own parallel assert that there are no children.
Also, the inline limit node child should be asserted as null in the inline case and (in production) the limit_node member initialization can be brought into the !node->isInline() block.

src/ee/plannodes/receivenode.cpp
```diff
@@ -43,31 +43,79 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
-#include "receivenode.h"
-
 #include <sstream>
 
+#include "boost/foreach.hpp"
```
paulmartel: My preference for order of #includes is:
1) local like "receivenode.h"
2) other product-specific headers like "common/..."
3) third-party headers like "boost/..."
4) system headers

This helps validate early that our own header files come equipped with the headers they NEED and that they
don't contain overly broad definitions that could interfere with third party headers etc.
This validation makes it easier to add an include of one of our own headers into another cpp without having to reconstruct its dependencies there.

src/ee/plannodes/receivenode.cpp
```diff
@@ -43,31 +43,79 @@
...
+
+#include "receivenode.h"
+
 namespace voltdb {
 
-ReceivePlanNode::~ReceivePlanNode() { }
+namespace {
```
paulmartel: Is there some reason to prefer anonymous namespace over good old static declaration?
I've never been a fan.

src/ee/plannodes/receivenode.h
```diff
@@ -46,20 +46,30 @@
...
     std::string debugInfo(const std::string& spacer) const;
+    bool isMergeReceive() const { return m_mergeReceive; }
+    TupleSchema* allocateTupleSchemaPreAgg() const;
 
 protected:
     void loadFromJSONObject(PlannerDomValue obj);
+
+private:
+    bool m_mergeReceive;
+    // output schema pre inline aggregation and projection
```
paulmartel: Add a comment describing that this is only non-empty when m_mergeReceive is true.
Also explain what it means for this to be empty when m_mergeReceive is true.

src/ee/plannodes/receivenode.cpp
```diff
@@ -43,31 +43,79 @@
...
+}
+
+
+ReceivePlanNode::ReceivePlanNode() :
+    m_mergeReceive(false), m_outputSchemaPreAgg()
+{ }
+
+ReceivePlanNode::~ReceivePlanNode()
+{
+    if (m_mergeReceive) {
```
paulmartel: This flag test is not actually required.

src/ee/plannodes/receivenode.h
```diff
@@ -46,20 +46,30 @@
...
+    std::vector<SchemaColumn*> m_outputSchemaPreAgg;
```
paulmartel: Do we gain much from MergeReceivePlanNode not having its own class?

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+import org.voltdb.expressions.AbstractExpression;
+import org.voltdb.plannodes.AbstractPlanNode;
+import org.voltdb.plannodes.AbstractScanPlanNode;
+import org.voltdb.plannodes.AggregatePlanNode;
+import org.voltdb.plannodes.HashAggregatePlanNode;
+import org.voltdb.plannodes.OrderByPlanNode;
+import org.voltdb.plannodes.ReceivePlanNode;
+import org.voltdb.types.PlanNodeType;
+
+public class ReplaceOrderBy extends MicroOptimization {
```
paulmartel: The name is a bit vague -- how about "InlineOrderByIntoMergeReceive"?

build.py
alexeevm: corrected

src/ee/executors/executorutil.cpp
alexeevm: corrected

src/ee/executors/executorutil.cpp
alexeevm: yes, moved to the AbstarctExecutor

src/ee/executors/executorutil.cpp
alexeevm: very nice.

src/ee/executors/orderbyexecutor.cpp
```diff
@@ -69,19 +70,22 @@ OrderByExecutor::p_init(AbstractPlanNode* abstract_node,
...
 
-    //
-    // Our output table should look exactly like out input table
-    //
-    node->
-        setOutputTable(TableFactory::
+        assert(node->getChildren()[0] != NULL);
+
+        //
+        // Our output table should look exactly like out input table
```
paulmartel: Pre-existing typo: "out input" should be "our input".

src/ee/plannodes/receivenode.cpp
```diff
@@ -43,31 +43,79 @@
...
-        SchemaColumn* col = getOutputSchema()[ctr];
-        buffer << spacer << "  [" << ctr << "] ";
-        buffer << "name=" << col->getColumnName() << " : ";
-        buffer << "size=" << col->getExpression()->getValueSize() << " : ";
-        buffer << "type=" << col->getExpression()->getValueType() << "\n";
+    if (!m_mergeReceive) {
+        schemaDebugInfo(buffer, getOutputSchema(), "Incoming", spacer);
+    } else {
+        buffer << spacer << "Merge Receive\n";
+        schemaDebugInfo(buffer, m_outputSchemaPreAgg, "Incoming", spacer);
```
paulmartel: Maybe want to output some reassurance that the empty m_outputSchemaPreAgg, IS OK like printing "Effectively the same as Outgoing".

src/ee/executors/mergereceiveexecutor.h
alexeevm: corrected both comments

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2091,6 +2105,28 @@ else if (agg_expression_type != ExpressionType.AGGREGATE_MIN &&
...
+        Index index = root.getCatalogIndex();
+        if ( ! IndexType.isScannable(index.getType())) {
+            return;
+        }
+        if (! index.getPredicatejson().isEmpty()) {
+            // do not try to look at Partial/Sparse index
+            return;
+        }
+
+        ArrayList<AbstractExpression> bindings = new ArrayList<AbstractExpression>();
```
paulmartel: Note: use of java 1.7 diamond operator aids readability e.g.

        ArrayList<AbstractExpression> bindings = new ArrayList<>();

src/ee/executors/orderbyexecutor.cpp
alexeevm: agree

src/ee/executors/orderbyexecutor.cpp
alexeevm: corrected

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+    protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
+    {
+        assert(planNode != null);
+
+        Queue<AbstractPlanNode> children = new LinkedList<AbstractPlanNode>();
+        children.add(planNode);
+
+        while(!children.isEmpty()) {
+            AbstractPlanNode plan = children.remove();
+            if (PlanNodeType.RECEIVE == plan.getPlanNodeType()) {
```
paulmartel: cache plan.getPlanNodeType() for readability.

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+                // continue. We are after the coordinator ORDER BY node.
+                return planNode;
+            } else if (PlanNodeType.ORDERBY == plan.getPlanNodeType()) {
```
paulmartel: Don't need an else after a return.

src/ee/plannodes/receivenode.cpp
alexeevm: corrected, plus few other files like mergereceiveexecutor

src/ee/plannodes/receivenode.cpp
alexeevm: In this particular case no. It may be helpful to group multiple "static" in one place to avoid typing the qualifier for each of them. Also they allows to do typedefs if needed, maybe few other things.
Changed to static

src/ee/plannodes/receivenode.cpp
alexeevm: yes, removed

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+        receive.clearParents();
+        if (rootNode == null) {
+            rootNode = receive;
+        } else {
+            rootNode.clearChildren();
+            rootNode.addAndLinkChild(receive);
+        }
+
+        // Add inline ORDER BY node
+        AbstractPlanNode orderByNode = inlineCandidates.get(PlanNodeType.ORDERBY);
```
paulmartel: You don't need two variables named orderbyNode and orderByNode to hold the same value.

src/ee/plannodes/receivenode.cpp
alexeevm: added the message Incoming "Incoming Table effectively the same as Outgoing" if it's empty

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+        // the ORDER BY node does. Alternatively, it could belong to a distributed subquery.
+        // Walk up the tree starting at the receive node until we hit either a scan node
+        // (distributed subquery) or the original order by node (distributed order by)
+        // Collect all nodes that are currently in between ORDER BY and RECEIVE nodes
+        // If the optimization is possible, they will be converted to inline nodes of
+        // the MERGE RECEIVE node. The expected node types are:
+        //      LIMIT, AGGREGATE/PARTIALAGGREGATE/HASHAGGREGATE
+        // The HASHAGGREGATE must be convertible to AGGREGATE or PARTIALAGGREGATE for optimization
+        // to be applicable.
+        Map<PlanNodeType, AbstractPlanNode> inlineCandidates = new HashMap<PlanNodeType, AbstractPlanNode>();
```
paulmartel: I think the hashmap is getting in the way here.
It would be simpler just to have one local variable for each of the ORDER BY, the AGG, the LIMIT -- these exist in the current code anyway, so just widen their scopes as needed and initialize to null.

An alternative -- one which which I think I like less -- is to populate the actual list of inline children as you go.
It's the required addAll that makes the initial hashmap seem especially redundant to me. I think that populating from the three local variables (if not null) in one short block of code that follows the block of code that tries to set them in various ways would be most concise.

src/ee/plannodes/receivenode.h
alexeevm: Added comments

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+        Set<Integer> coveredGroupByColumns = new HashSet<Integer>();
+
+        Iterator<AbstractExpression> orderbyIt = orderbys.iterator();
+        while (orderbyIt.hasNext()) {
+            AbstractExpression orderby = orderbyIt.next();
+            int idx = 0;
+            for (AbstractExpression groupby : groupbys) {
+                if (!coveredGroupByColumns.contains(idx)) {
+                    // This groupby expression is not bind to any orderby yet
+                    if (orderby.bindingToIndexedExpression(groupby) != null) {
```
paulmartel: Advice: use equals here instead of bindingToIndexedExpression(groupby) != null.
The matching of indexed expressions to GROUP BY and ORDER BY expressions uses bindingToIndexedExpression for good reason and I HOPE properly retains and propagates those bindings up to the cached compiled plan. It just seems needless IN THIS CASE to try to cleverly match constants and parameters -- which is ALL that bindings do above and beyond simpler equality checks. This is especially so since we are not propagating any bindings we might find here up  to the cached compiled plan.

The code as it stands only works reliably if 

(A) it's a common equality case and so the bindings list is non-null but empty and there is no constraint to propagate -- expression equality would have worked just as well in this case
OR 
(B) we are generating a non-empty binding list here (where a strict equality would fail) AND we don't happen to suffer any ill effects from failing to propagate the binding up to the cached compiled plan BECAUSE that particular binding was already identified in the earlier match of the indexable expression, causing it to get propagated, anyway. 

Yet because of the hsql-enforced restriction that display columns and order by columns for a group by query MUST be based on either aggs OR expressions that are EQUAL TO the group by expressions, I don't see room for ORDER BY and GROUP BY expressions that differ only in their use of a parameter in place of a constant.

If the user tried to mix things up like "select year / 10 as decade, max(GDP) ... group by year / ? order by year / 10", the plan would get rejected by hsql.
If a query like "select year / 10 as decade, max(GDP) ... group by year / 10 order by year / 10"
got parameterized into  "select year / ? as decade, max(GDP) ... group by year / ? order by year / ?"
the resulting plan would be using 3 different parameters and so would not pass either equality tests or the 
bindingToIndexedExpression(groupby) != null test. I HOPE (but I don't know for sure) that the parameterized plan with its attempt to reference non-agg non-group-by expressions will ungracefully  throw some kind of error vs. say try to assume that these are additional (redundant, as it turns out) group by columns?). If we do get a throw, the planner will fall back to the unparameterized plan and all will be well for the most-part.

By that token, it should be just as impossible to generate here an original never-seen-before-so-never-propagated binding which could lead to a bad plan for a slightly different query executed down the line.

Even so, it seems simpler and maybe? safer and slightly more efficient just to limit ourselves to strict equality checking here.

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
```diff
@@ -0,0 +1,234 @@
...
+                        break;
+                    }
+                }
+                ++idx;
+            }
+        }
+        if (orderbys.isEmpty() && groupbys.size() == coveredGroupByColumns.size()) {
+            // All GROUP BY expressions are also ORDER BY - Serial aggregation
+            return AggregatePlanNode.convertToSerialAggregatePlanNode(hashAggr);
+        } else if (orderbys.isEmpty() && !coveredGroupByColumns.isEmpty() ) {
```
paulmartel: elses not reqired after return.

src/frontend/org/voltdb/plannodes/AbstractJoinPlanNode.java
```diff
@@ -270,6 +270,20 @@ public SortDirectionType getSortDirection() {
...
+    @Override
+    public boolean isOutputOrdered (List<AbstractExpression> sortExpressions, List<SortDirectionType> sortDirections) {
+        AbstractPlanNode outerTable = m_children.get(0);
+        AbstractPlanNode aggrNode = AggregatePlanNode.getInlineAggregationNode(this);
+        if (aggrNode != null && aggrNode.getPlanNodeType() == PlanNodeType.HASHAGGREGATE) {
+            return false;
+        }
+        if (outerTable.getPlanNodeType() == PlanNodeType.INDEXSCAN || outerTable instanceof AbstractJoinPlanNode) {
+            return outerTable.isOutputOrdered(sortExpressions, sortDirections);
+        } else {
```
paulmartel: else after return.

src/frontend/org/voltdb/plannodes/AbstractJoinPlanNode.java
```diff
@@ -270,6 +270,20 @@ public SortDirectionType getSortDirection() {
...
     }
 
+    @Override
+    public boolean isOutputOrdered (List<AbstractExpression> sortExpressions, List<SortDirectionType> sortDirections) {
+        AbstractPlanNode outerTable = m_children.get(0);
+        AbstractPlanNode aggrNode = AggregatePlanNode.getInlineAggregationNode(this);
+        if (aggrNode != null && aggrNode.getPlanNodeType() == PlanNodeType.HASHAGGREGATE) {
+            return false;
+        }
+        if (outerTable.getPlanNodeType() == PlanNodeType.INDEXSCAN || outerTable instanceof AbstractJoinPlanNode) {
```
paulmartel: Add a comment here along the lines of "not yet handling ORDER BY expressions based on more than just the left-most table".

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -307,6 +308,24 @@ public boolean isOrderDeterministic() {
     }
 
     /**
+     * Does the output is sorted according to the order implied by the (expression, sort order) pairs.
```
paulmartel: grammar "Does ... is"

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -307,6 +308,24 @@ public boolean isOrderDeterministic() {
...
+     *@param sortExpressions list of ordering columns expressions
+     *@param sortDirections list of corresponding sort orders
+     *
+     * @return TRUE if the node's output table is ordered. FALSE otherwise
+     */
+    public boolean isOutputOrdered (List<AbstractExpression> sortExpressions, List<SortDirectionType> sortDirections) {
+        assert(sortExpressions.size() == sortDirections.size());
+        if (m_children.size() == 1) {
+            return m_children.get(0).isOutputOrdered(sortExpressions, sortDirections);
+        } else {
```
paulmartel: else after return.

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
```diff
@@ -320,6 +320,71 @@ public SortDirectionType getSortDirection() {
...
+        List<ColumnRef> indexedColRefs = new ArrayList<ColumnRef>();
+        List<Integer> indexedColIds = new ArrayList<Integer>();
+        CatalogUtil.getCatalogIndexExpressions(getCatalogIndex(), getTableScan(),
+                indexedExprs, indexedColRefs, indexedColIds);
+        int indexExprCount = (indexedExprs.isEmpty()) ? indexedColRefs.size() : indexedExprs.size();
+        if (indexExprCount < sortExpressions.size()) {
+            // Not enough index expressions to cover all of the sort expressions
+            return false;
+        }
+        for (int idxToCover = 0; idxToCover < sortExpressions.size(); ++idxToCover) {
```
paulmartel: Duplicate the loop inside the if/else and simplify each loop body to:

            AbstractExpression sortExpression = sortExpressions.get(idxToCover);
            if (!isSortExpressionCovered(...) {
                return false;
            }

About the same # LOC, more efficient, easier to step through.

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
```diff
@@ -320,6 +320,71 @@ public SortDirectionType getSortDirection() {
...
+            }
+        }
+        return true;
+    }
+
+    private boolean isSortExpressionCovered(AbstractExpression sortExpression, List<AbstractExpression> indexedExprs,
+            int idxToCover) {
+        assert(idxToCover < indexedExprs.size());
+        AbstractExpression indexExpression = indexedExprs.get(idxToCover);
+        return sortExpression.bindingToIndexedExpression(indexExpression) != null;
```
paulmartel: Is discarding (vs. propagating) bindings here safe? Only if they were previously discovered in selecting this indexscan -- do we know that -- or could the sort be using an accidentally like-valued but different parameter which should be getting its own binding propagated but is not?

paulmartel: possible counter-example. For:
    create index on t (year / 100, country);

    select * from t where year / 100 < 20 order by year / 100, country;

should NOT require an ORDER BY.

    select * from t where year / 100 < 20 order by year / 10, country;

SHOULD require an ORDER BY.

Neither query should use a plan cached for the other.

src/frontend/org/voltdb/plannodes/OrderByPlanNode.java
```diff
@@ -204,4 +213,22 @@ protected String explainPlanForNode(String indent) {
...
+        assert(sortExpressions.size() == sortDirections.size());
+        assert (m_sortExpressions.size() == m_sortDirections.size());
+        if (sortExpressions.size() != m_sortExpressions.size()) {
+            return false;
+        }
+        for(int idx = 0; idx < m_sortExpressions.size(); ++idx) {
+            if (sortDirections.get(idx) != m_sortDirections.get(idx)) {
+                return false;
+            }
+            if (m_sortExpressions.get(idx).bindingToIndexedExpression(sortExpressions.get(idx)) == null) {
```
paulmartel: Use expression equals.
Actually, AFAIK there are no false returns from this function and it could all be asserts -- why would we ever construct an ORDER BY node that did not use the statement's ORDER BY expressions and sort directions?

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
```diff
@@ -69,22 +99,67 @@ public void resolveColumnIndexes()
...
+            OrderByPlanNode opn = (OrderByPlanNode) orderNode;
+            opn.resolveSortIndexesUsingSchema(m_outputSchemaPreInlineAgg);
+
+            AggregatePlanNode aggrNode = AggregatePlanNode.getInlineAggregationNode(this);
+            if (aggrNode != null) {
+                aggrNode.resolveColumnIndexesUsingSchema(m_outputSchemaPreInlineAgg);
+                m_outputSchema = aggrNode.getOutputSchema().clone();
+                m_outputSchema.sortByTveIndex();
+            } else {
+                m_outputSchema = m_outputSchemaPreInlineAgg;
```
paulmartel: These else blocks look redundant -- is m_outputSchema not already good to go in these cases?

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
```diff
@@ -69,22 +99,67 @@ public void resolveColumnIndexes()
...
+            }
+        } else {
+            m_outputSchema = m_outputSchemaPreInlineAgg;
+        }
     }
 
     @Override
     public void toJSONString(JSONStringer stringer) throws JSONException {
         super.toJSONString(stringer);
+        if (m_mergeReceive == true) {
```
paulmartel: Here and elsewhere, "== true" is noise.

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
```diff
@@ -69,22 +99,67 @@ public void resolveColumnIndexes()
...
 
     @Override
     public void loadFromJSONObject( JSONObject jobj, Database db ) throws JSONException {
         helpLoadFromJSONObject(jobj, db);
+        if (jobj.has(Members.MERGE_RECEIVE.name())) {
+            m_mergeReceive = jobj.getBoolean(Members.MERGE_RECEIVE.name());
+            if (m_mergeReceive == true) {
+                if (jobj.has(Members.OUTPUT_SCHEMA_PRE_AGG.name())) {
+                    m_outputSchemaPreInlineAgg = new NodeSchema();
+                    m_hasSignificantOutputSchema = true;
```
paulmartel: Isn't m_hasSignificantOutputSchema more properly established elsewhere?

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
```diff
@@ -116,4 +191,25 @@ public boolean reattachFragment( SendPlanNode child  ) {
...
+    public boolean isMergeReceive() {
+        return m_mergeReceive;
+    }
+
+    @Override
+    public boolean isOutputOrdered (List<AbstractExpression> sortExpressions, List<SortDirectionType> sortDirections) {
+        if (isMergeReceive()) {
+            AbstractPlanNode orderBy = getInlinePlanNode(PlanNodeType.ORDERBY);
+            assert(orderBy != null);
+            return orderBy.isOutputOrdered(sortExpressions, sortDirections);
```
paulmartel: again, is a non-compliant orderby a real thing?
likewise, is a non-compliant merge-receive a real thing?
Otherwise just return isMergeReceive().

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
```diff
@@ -320,6 +320,71 @@ public SortDirectionType getSortDirection() {
...
+                return false;
+            }
+        }
+        // Verify that all sort expressions are covered by the consecutive index expressions
+        // starting from the first one
+        List<AbstractExpression> indexedExprs = new ArrayList<AbstractExpression>();
+        List<ColumnRef> indexedColRefs = new ArrayList<ColumnRef>();
+        List<Integer> indexedColIds = new ArrayList<Integer>();
+        CatalogUtil.getCatalogIndexExpressions(getCatalogIndex(), getTableScan(),
+                indexedExprs, indexedColRefs, indexedColIds);
```
paulmartel: indexColIds unused, fetched for nothing? See review comments on getCatalogIndexExpressions.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2140,4 +2141,31 @@ public static String getLimitPartitionRowsDeleteStmt(Table table) {
...
+            List<AbstractExpression> indexedExprs, List<ColumnRef> indexedColRefs, List<Integer> indexedColIds) {
+        String exprsjson = index.getExpressionsjson();
+        if (exprsjson.isEmpty()) {
+            indexedColRefs.addAll(CatalogUtil.getSortedCatalogItems(index.getColumns(), "index"));
+            for (ColumnRef cr : indexedColRefs) {
+                indexedColIds.add(cr.getColumn().getIndex());
+            }
+        } else {
+            try {
+                indexedExprs.addAll(AbstractExpression.fromJSONArrayString(exprsjson, tableScan));
```
paulmartel: This helper could benefit from:
  less List-building -- a variant of AbstractExpression.fromJSONArrayString that filled a list vs. constructed a new list would help. The copying of the ColRef list into the list provided is a little clumsy -- I suppose that getSortedCatalogItems could have a similar list-loading variant.
 a boolean return that indicated which list was filled.
 dropping the indexedColIds output which is trivially derivable as needed (seldom) by potential callers
 more callers -- there are plenty of candidates, especially if you don't mind trading a little extra list processing for the benefit of hiding the catalog and json details -- most direct callers of AbstractExpression.fromJSONArrayString could be calling this instead.

src/ee/plannodes/receivenode.h
alexeevm: During the initial implementation, the only difference between the MergeReceivePlanNode and the ReceivePlanNode was the flag, and I opted to keep them combined. Now, that the MergeReceivePlanNode can handle aggregation (and possible future enhancements), the two nodes diverged and could be separated into two classes. The new MergeReceivePlanNode can be a subclass of the existing ReceivePlanNode to make sure all the existing logic on the planner side related to the multiple fragments and distributed quires would work for the new plan node.

tests/frontend/org/voltdb/planner/TestPlansDistinct.java
```diff
@@ -401,8 +441,12 @@ protected void checkDistinctWithGroupbyPlans(String distinctSQL, String groupbyS
             assertFalse(hasTopProjection2);
         }
 
-        // check the rest plan nodes
-        assertEquals(apn1.toExplainPlanString(), apn2.toExplainPlanString());
+        // check the rest plan nodes. In case of applied 'MERGE RECEIVE' optimization
+        // the whole top fragment is a single Receive node with other inline nodes which were
+        // already checked above
+        if (distinctMergeReceive == false && groupByMergeReceive == false) {
```
paulmartel: Is there some reason we don't test for plan equality when both flags are true?

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
```diff
@@ -586,7 +586,7 @@ public void testGroupByWithLimit() {
 
         // In future, it may use the serial aggregate: ENG-6586
```
paulmartel: Is this comment now obsolete? Is ENG-6586 fixed -- at least in part?

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
```diff
@@ -1515,12 +1515,54 @@ private void checkMVReaggreateFeature(
...
+        assertTrue(pns.size() == 2);
+        AbstractPlanNode p = pns.get(0);
+        assertTrue(p instanceof SendPlanNode);
+        p = p.getChild(0);
+
+        if (p instanceof ProjectionPlanNode) {
+            p = p.getChild(0);
+        }
+
+        AbstractPlanNode receiveNode = (p instanceof ProjectionPlanNode) ? p.getChild(0) : p;
```
paulmartel: Redundant code -- I don't think we ever need to skip TWO projection nodes.

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
```diff
@@ -1515,12 +1515,54 @@ private void checkMVReaggreateFeature(
...
+
+            assertEquals(numGroupbyOfReaggNode, reAggNode.getGroupByExpressionsSize());
+            assertEquals(numAggsOfReaggNode, reAggNode.getAggregateTypesSize());
+        }
+
+        p = pns.get(1);
+        assertTrue(p instanceof SendPlanNode);
+        p = p.getChild(0);
+
+        assertTrue(p instanceof AbstractScanPlanNode);
```
paulmartel: The merge receive case should be able to assert something a bit stronger, concerning an index scan with a specific sort direction.

tests/frontend/org/voltdb/planner/TestPlansLimit.java
```diff
@@ -189,8 +192,22 @@ private void checkInlineLimitWithOrderby(List<AbstractPlanNode> pns, boolean pus
...
-        assertNotNull(p.getInlinePlanNode(PlanNodeType.LIMIT));
+        boolean mergereceive = false;
+        if (p instanceof ReceivePlanNode) {
+            assertTrue(((ReceivePlanNode)p).isMergeReceive());
+            assertNotNull(p.getInlinePlanNode(PlanNodeType.ORDERBY));
+            mergereceive = true;
+        } else {
+            assertTrue(p instanceof OrderByPlanNode);
+        }
+        if (mergereceive) {
```
paulmartel: Simplify overly complex flow control that falls out of and right back into corresponding if/else blocks.
Eliminate the mergereceive flag and either merge the if blocks and the else blocks or reset p (the parent of the inline limit) in each of the if and else blocks and fall out to check p's inline child.

tests/frontend/org/voltdb/planner/TestPlansLimit.java
```diff
@@ -203,4 +220,25 @@ private void checkInlineLimitWithOrderby(List<AbstractPlanNode> pns, boolean pus
         }
     }
 
+    private void checkInlineLimitAndOrderbyWithReceive(List<AbstractPlanNode> pns, boolean pushdown) {
+        AbstractPlanNode p;
+
+        p = pns.get(0).getChild(0);
+        assertTrue(p instanceof ProjectionPlanNode);
```
paulmartel: This raises an interesting question -- whether a MergeReceivePlanNode should be able to absorb a parent projection as an inline child for temptable space savings.
If this makes sense, we should create a follow-on ticket for it.

tests/frontend/org/voltdb/regressionsuites/testorderby-ddl.sql
```diff
@@ -46,3 +47,31 @@ create table b
...
+);
+
+PARTITION TABLE P1 ON COLUMN P1_D0;
+
+CREATE PROCEDURE TruncateP as DELETE FROM P;
+CREATE PROCEDURE TruncateP1 as DELETE FROM P1;
+
+CREATE VIEW V_P (V_P_D1, V_P_D2, V_CNT, V_SUM_P_D3, V_SUM_P_D0)
+    AS SELECT P_D1, P_D2, COUNT(*), SUM(P_D3), COUNT(P_D0) 
+    FROM P  GROUP BY P_D1, P_D2;
```
paulmartel: Missing final newline.

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
alexeevm: Refactored

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
alexeevm: done

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
alexeevm: done

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
alexeevm: I went with the first approach

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
alexeevm: Examples like "select year / ? as decade, max(GDP) ... group by year / ? order by year / ?" were my motivation to use bindingToIndexedExpression method instead of simple equals. But I haven't realize the former would return null in this case anyway. What is interesting, that the index scan for the ""select year / 10 as decade, max(GDP) ... group by year / 10 order by year / 10" has sort direction INVALID with index either on year column or year /10 expression. The INVALID sort order prohibit the optimization anyway. 

I will use the equality checking instead

src/frontend/org/voltdb/planner/microoptimizations/ReplaceOrderBy.java
alexeevm: removed

src/frontend/org/voltdb/plannodes/AbstractJoinPlanNode.java
alexeevm: added

src/frontend/org/voltdb/plannodes/AbstractJoinPlanNode.java
alexeevm: removed

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
alexeevm: Thanks!

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
alexeevm: removed

src/frontend/org/voltdb/utils/CatalogUtil.java
alexeevm: Refactored. 
The primary motivation for this helper was the multitude of places where index expressions were constructed from an index. I will circle back to try to use this helper instead once I am done with the review.

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
alexeevm: Refactored

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
alexeevm: yes indeed

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
alexeevm: The code suffers from the same problem as the ReplaceOrderBy.convertToSerialAggregation.

I replace the sortExpression.bindingToIndexedExpression with the equality check

src/frontend/org/voltdb/plannodes/OrderByPlanNode.java
alexeevm: The only situation I can imagine is a union with its own ORDER BY not matching the individual selects ORDER BY

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
alexeevm: Yes indeed. generateOutputSchema guarantees that.

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
alexeevm: removed

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
alexeevm: yes, generateOutputSchema does it. Not sure, why I have this line in the first place. removed.

src/frontend/org/voltdb/plannodes/OrderByPlanNode.java
alexeevm: On a second thought, my example doesn't fly -  the union's own results are always unordered, so propagating the isOutputOrdered question to is children is pointless.

src/frontend/org/voltdb/plannodes/ReceivePlanNode.java
alexeevm: I can't come up with an example where it is not. Changed to return isMergeReceive()

tests/frontend/org/voltdb/planner/TestPlansDistinct.java
alexeevm: This is where I cut conners - The validation logic is tailored for two 'conventional' plans where it traverses both of them top-down in parallel and both trees has similar structures. After a certain point both plans are identical and their 'explain' string can be compared. In case of applied optimization, the whole top fragment is reduced to a single Receive node (apart from the top send and projection nodes) with inline nodes. I was struggling to come-up with a single variant that covers all possible scenarios - both plans are optimized, not optimized, only one is optimized. It is on my TODO list to find a better solution

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
alexeevm: Yes, with the corrected PlanAssembler.switchToIndexScanForGroupBy the winner plan does use Serial aggregate
 LIMIT 5
  RECEIVE FROM ALL PARTITIONS
    INDEX SCAN of "T3" using "T3_TREE1"
    range-scan on 1 of 2 cols from (A3 > 1) to end, filter by (A3 > 1)
     inline Serial AGGREGATION ops: COUNT(
      inline LIMIT 5

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
alexeevm: removed.

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
alexeevm: added
        assertTrue(p instanceof IndexScanPlanNode);
        assertEquals(sortDirection, ((IndexScanPlanNode)p).getSortDirection());

tests/frontend/org/voltdb/planner/TestPlansLimit.java
alexeevm: simplified by merging relevant if/else blocks

tests/frontend/org/voltdb/planner/TestPlansLimit.java
alexeevm: I think we can. The only complication I see is  an aggregation and the projection together - the MergeReceive node would need two temp tables in this case instead of the current one. If it's only either one, then it shouldn't be too difficult to add.

tests/frontend/org/voltdb/regressionsuites/testorderby-ddl.sql
alexeevm: added

src/ee/executors/mergereceiveexecutor.cpp
```diff
@@ -0,0 +1,320 @@
...
+        assert(pre_agg_schema != NULL);
+        m_tmpInputTable.reset(allocate_temp_table(m_abstractNode->databaseId(),
+                                                  pre_agg_schema,
+                                                  limits,
+                                                  "tempInput"));
+    } else {
+        m_tmpInputTable.reset(TableFactory::getTempTable(m_abstractNode->databaseId(),
+                                                         "tempInput",
+                                                         m_abstractNode->generateTupleSchema(),
+                                                         m_tmpOutputTable->getColumnNames(),
```
paulmartel: Pushing the output column names down to the dependency table is not that useful.
I'd favor simplifying this code by having the (m_agg_exec == NULL) if/else just decide what schema to use.
Then unconditionally call TableFactory::getTempTable with the correctly-sized vector of "" column names.

alexeevm: simplified

src/ee/executors/mergereceiveexecutor.cpp
```diff
@@ -0,0 +1,320 @@
...
+        pmp.countdownProgress();
+    }
+}
+
+TempTable* allocate_temp_table(CatalogId databaseId, TupleSchema* schema, TempTableLimits* limits, const char* tempTableName) {
+    assert(limits);
+    assert(schema);
+    int column_count = schema->columnCount();
+    std::vector<std::string> column_names;
+    column_names.reserve(column_count);
```
paulmartel: As the column names don't really matter, here, we can replace all this initialization with simple initialization to all "" values with:
std::vector<std::string> column_names(schema->columnCount());

See related comment in MergeReceiveExecutor::p_init

alexeevm: yes, this is exactly what I have done. I actually eliminated this helper method completely:
    // Create a temp table to collect tuples from multiple partitions
    TupleSchema* pre_agg_schema = (pre_agg_schema != NULL) ?
        merge_receive_node->allocateTupleSchemaPreAgg() : m_abstractNode->generateTupleSchema();
    std::vector<std::string> column_names(pre_agg_schema->columnCount());
    m_tmpInputTable.reset(TableFactory::getTempTable(m_abstractNode->databaseId(),
                                                         "tempInput",
                                                         m_abstractNode->generateTupleSchema(),
                                                         column_names,
                                                         limits));

src/ee/executors/mergereceiveexecutor.cpp
```diff
@@ -0,0 +1,320 @@
...
+        assert( i != nonEmptyPartitions -1 || end == tuples.end());
+    }
+
+    int tupleCnt = 0;
+    int tupleSkipped = 0;
+
+    while (limit == -1 || tupleCnt < limit) {
+        range_iterator rangeIt = partitions.begin();
+        // remove partitions that are empty
+        while (rangeIt != partitions.end()) {
```
paulmartel: This empty partition culling should be applied more selectively.
The previous culling in p_execute already assures that "partitions" entries don't start empty.
The only process that MAY cause a partition to become empty is when it produces the min tuple.
So it's only AFTER ++minRangeIt->first that THIS ONE partition (not all) need to be considered for erasure from partitions.

src/ee/executors/mergereceiveexecutor.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+namespace voltdb {
+
+namespace {
+
+typedef std::vector<TableTuple>::const_iterator tuple_iterator;
+typedef std::pair<tuple_iterator, tuple_iterator> tuple_range;
+typedef std::list<tuple_range>::iterator range_iterator;
+
+range_iterator min_tuple_range(std::list<tuple_range>& partitions, TupleComparer comp) {
```
paulmartel: std::list seems like a less-than-optimal structure for partitions.
I think that much of the work of min_tuple_range could be optimally delegated to std::heap calls.
I am picturing partitions being maintained as a std::heap of (non-empty) tuple_ranges ordered by applying TupleComparer  on their "first tuples".
On each iteration, the head (min range) of the heap is popped off, the min range's first tuple is popped off  -- via first++ -- for return and if not at end, min range is inserted back into the heap (to be sorted by its new first tuple).

src/ee/executors/mergereceiveexecutor.cpp
paulmartel: This comment is superceded (subsumed) by the idea of using std::heap.

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
```diff
@@ -757,6 +758,157 @@ private void subtestEng5021() throws Exception
...
+        client.callProcedure("InsertO1", 7, 8,"dummy","dummy");
+
+        client.callProcedure("InsertO3", 1, 1, 7, 7);
+        client.callProcedure("InsertO3", 2, 2, 7, 7);
+        client.callProcedure("InsertO3", 3, 3, 8, 8);
+        client.callProcedure("InsertO3", 4, 4, 1, 1);
+        client.callProcedure("InsertO3", 10, 10, 10, 10);
+
+        // Partitions Result sets are ordered by index. No LIMIT/OFFEST
+        VoltTable vt = client.callProcedure("@AdHoc", "SELECT PKEY FROM O1 ORDER BY PKEY DESC").getResults()[0];
```
paulmartel: Assign a query to a String variable in these tests to make it clear that/when you mean to refer to the exact same statement -- such as in corresponding calls to @AdHoc and @Explain.
Note how the NLIJ example adds a "limit 3" between the @AdHoc and the @Explain -- I'm not sure if that was intentional or accidental. I'm not sure if they should be separate test cases.

src/ee/executors/mergereceiveexecutor.cpp
```diff
@@ -0,0 +1,320 @@
...
+                rangeIt = partitions.erase(rangeIt);
+                --nonEmptyPartitions;
+            } else {
+                ++rangeIt;
+            }
+        }
+        if (nonEmptyPartitions == 1) {
+            // copy the remaining tuples to the output
+            range_iterator rangeIt = partitions.begin();
+            while (rangeIt->first != rangeIt->second && (limit == -1 || tupleCnt < limit)) {
```
paulmartel: Can the redundancy between this inner loop and the outer loop be eliminated.
All the inner loop is doing differently from the outer loop is short-circuiting minRangeIt to be "partitions.begin()" instead of calculating a min from multiple candidates -- easily fixed with a ?: .

src/ee/executors/mergereceiveexecutor.cpp
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    return true;
+}
+
+bool MergeReceiveExecutor::p_execute(const NValueArray &params) {
+    int loadedDeps = 0;
+
+    // iterate over dependencies and merge them into the temp table.
+    // The assumption is that the dependencies results are are sorted.
```
paulmartel: "the dependencies results are are sorted" -> "each dependency result is already sorted"

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
```diff
@@ -757,6 +758,157 @@ private void subtestEng5021() throws Exception
...
+
+        vt = client.callProcedure("@AdHoc",
+                "SELECT V_P_D1 FROM V_P order by V_P_D1").getResults()[0];
+        expected = new long[][] {{1}, {1}, {6}, {6}, {11}, {11}};
+        validateTableOfLongs(vt, expected);
+        vt = client.callProcedure("@Explain",
+                "SELECT V_P_D1 FROM V_P order by V_P_D1").getResults()[0];
+        assertTrue(vt.toString().contains("MERGE RECEIVE"));
+    }
+
```
paulmartel: MergeReceive test cases should include cases where there are no tuples, where all the tuples happen to come from one partition, and where the tuples come from some but not all partitions.
For completeness in testing the mergesort, it should also include cases where the order of the tuples from each partition happen not to overlap each others' ranges, and cases where tuples on multiple partitions have identical sort keys.

It would be good to have C++ unit tests of the merge_sort helper functions applied to these kinds of cases -- at least exercising the agg_exec == NULL case to build a temptable.

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
```diff
@@ -757,6 +758,157 @@ private void subtestEng5021() throws Exception
 
     }
 
+    private void subtestOrderByMP() throws Exception
```
paulmartel: Add a test case for querying a view, ordering by its non-partition-key grouping columns.

tests/frontend/org/voltdb/planner/TestPlansOrderBy.java
```diff
@@ -289,4 +296,321 @@ public void testOrderbyColumnsNotInDisplayList() {
...
+            // Select from the aggregate view
+            List<AbstractPlanNode> frags =  compileToFragments("Select * from V_P1 ORDER BY V_G1, V_G2");
+            AbstractPlanNode pn = frags.get(0).getChild(0).getChild(0);
+            validateAggregatedMergeReceive(pn, true, false, false, false);
+        }
+        {
+            // Two aggregates at the coordinator. ORDER BY
+            List<AbstractPlanNode> frags =  compileToFragments("Select V_G1, sum(V_CNT), max(v_sum_age) from V_P1_ABS GROUP BY V_G1 ORDER BY V_G1");
+            AbstractPlanNode pn = frags.get(0).getChild(0).getChild(0);
+            assertEquals(PlanNodeType.ORDERBY, pn.getPlanNodeType());
```
paulmartel: It would help to have the negative test cases segregated to the end of the subtest or into a separate subtest.
Also, I'm not at all clear on why exactly some of these are not getting optimized.
Here in particular, I think you mean two aggregate NODES at the coordinator -- "two aggregates" was ambiguous -- "two aggregates"? like "sum and max"? does not sound like a problem.

tests/frontend/org/voltdb/planner/TestPlansOrderBy.java
```diff
@@ -289,4 +296,321 @@ public void testOrderbyColumnsNotInDisplayList() {
...
+            // Not supported yets
+            //            select indexed_non_partition_key, sum(col), sum(col)+1
+            //            from partitioned
+            //            group by indexed_non_partition_key
+            //            order by indexed_non_partition_key
+            //            limit 2;
+            List<AbstractPlanNode> frags =  compileToFragments(
+                    "select P_D1, sum(P_D1), sum(P_D1)+1 from P group by P_D1 order by P_D1 limit 2");
+            AbstractPlanNode pn = frags.get(0).getChild(0);
+            assertEquals(PlanNodeType.ORDERBY, pn.getPlanNodeType());
```
paulmartel: This case seems like low-hanging fruit. I'm not sure how it is interfering with the optimization.
Where does the extra projection appear in the pre-optimized plan -- is there no obvious place for it in the optimized plan?

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
paulmartel: I don't think you'll like the result of using equality for cases like this. I expect that it will by-pass the optimization.

paulmartel:
End of review round.

src/ee/executors/mergereceiveexecutor.cpp
alexeevm: corrected

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
alexeevm: Accidental. Implemented

tests/frontend/org/voltdb/planner/TestPlansOrderBy.java
alexeevm: Correct, "two aggregate nodes at the coordinator". I separated the test cases as you suggested and added better comments why the optimization was/wasn't applied

tests/frontend/org/voltdb/planner/TestPlansOrderBy.java
alexeevm: The winner coordinator fragment in this case is
 ORDER BY (SORT)
   PROJECTION
     Hash AGGREGATION ops: SUM(VOLT_TEMP_TABLE.column#0)
      RECEIVE FROM ALL PARTITIONS

My assumption was that this plan would require merge-receive node to have two temp tables.. The first one is to collect partitions input (the existing pre_aggregate table), and the second one (a new pre_projection table) as an aggregate node output to apply the projection because it can't be inlined with the aggregation yet.

Is it your understanding as well?

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
alexeevm: I think it's already there inside the subtestOrderByMP_Agg
SELECT V_P_D1 FROM V_P order by V_P_D1
where V_P is
CREATE VIEW V_P (V_P_D1, V_P_D2, V_CNT, V_SUM_P_D3, V_SUM_P_D0)
    AS SELECT P_D1, P_D2, COUNT(*), SUM(P_D3), COUNT(P_D0) 
    FROM P  GROUP BY P_D1, P_D2;
 and P_D0 is the partition column.
I also added one more select
SELECT V_P_D1, V_P_D2 FROM V_P order by V_P_D1 DESC , V_P_D2 DESC;

tests/frontend/org/voltdb/planner/TestPlansOrderBy.java
paulmartel: That sounds reasonable. The interesting thing about this particular case is that there is no strict order dependency between the ORDER BY and the PROJECTION, so, in theory the PROJECTION could float to the top of the plan and remain as a final stage, allowing the ORDER BY, AGG, and RECEIVE (and even an optional limit) to be optimized into a mergereceive.

This seems like a win even without the projection inlined.
The trick is to determine that the ORDER BY really doesn't use any of the projected expressions -- the optimization would not apply to any ORDER BY that has such a dependency.

src/ee/executors/mergereceiveexecutor.cpp
alexeevm: Merged two loops into a single flow

src/ee/executors/mergereceiveexecutor.cpp
alexeevm: Yes, replacing std::list with the std::make_heap/pop_heap/push_heap algorithms did simplify the logic.

I am adding the C++ unit tests for the merge sort executor

src/ee/executors/mergereceiveexecutor.cpp
alexeevm: Using heap completely eliminated the need for this loop

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
alexeevm: Created tests/ee/executors/MergeReceiveExecutorTest.cpp to test the merge sort. The merge sort itself is guaranteed to get non-empty partitions only. 
     [exec] MergeReceiveExecutorTest:
     [exec] 	emptyResultSetTest: PASSED.
     [exec] 	singlePartitionTest: PASSED.
     [exec] 	singlePartitionLimitOffsetTest: PASSED.
     [exec] 	singlePartitionBigOffsetTest: PASSED.
     [exec] 	twoNonOverlapPartitionsTest: PASSED.
     [exec] 	multipleOverlapPartitionsTest: PASSED.

I added the scenario where one of the partitions is empty to the regression suite.

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
alexeevm: Yes, you are right. Both 'explain' plans contains ORDER BY
12> CREATE INDEX P_D32_10_IDX ON P (P_D3 / 10, P_D2);
Command succeeded.
13> explain select P_D1 from P where P.P_D3 / 10 > 0 order by P_D3 / 10, P_D2;
EXECUTION_PLAN                 
  ORDER BY (SORT)
   RECEIVE FROM ALL PARTITIONS
     INDEX SCAN of "P" using "P_D32_10_IDX"
     range-scan on 1 of 2 cols from ((column#3 / 10) > ?1) to end, filter by ((column#3 / ?0) > ?1)

What is interesting though, the planner does produce the correct plan in this case
   MERGE RECEIVE FROM ALL PARTITIONS
      inline ORDER BY (SORT)
       INDEX SCAN of "P" using "P_D32_10_IDX"
        range-scan on 1 of 2 cols from ((P_D3 / 10) > 0) to end, filter by ((P_D3 / 10) > 0)

I guess, this is where the equality comparison is coming into the play.

alexeevm: I think I corrected the code by capturing the bindings now.

The above two examples are planned and executed as expected 
select * from t where year / 100 < 20 order by year / 100, country; -- uses MERGE RECEIVE
select * from t where year / 100 < 20 order by year / 10, country; -- uses ORDE BY

Added the corresponding test cases to the planner and regression test suite

alexeevm:
Paul, I think I've addressed  all your review comments. 
I opted to create a new  AbstarctReceivePlanNode class and to have both ReceivePlanNode and MergeReceivePlanNode inherit from it instead of the ReceivePlanNode <- MergeReceivePlanNode hierarchy mainly for consistency - to follow existing AbstarcJoin/ScanPlanNode pattern.

Mike

paulmartel:
End of review round. I like the AbstractReceiveNode refactoring.

PR #2666, opened 2015-08-06
Fix view delete with limit bug and add corresponding test case

yzhang1991:


tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -245,31 +245,33 @@ private void runAndVerifyENG6511(Client client, String query) throws IOException
     }
 
     // Test the correctness of min/max when choosing an index on both group-by columns and aggregation column/exprs.
-    private void subtestENG6511SinglePartition() throws IOException, ProcCallException
+    private void subtestENG6511(boolean singlePartition) throws IOException, ProcCallException
     {
         Client client = getClient();
         truncateBeforeTest(client);
+        int gid = singlePartition ? 1 : 2;
```
paulmartel: Rename "gid" to "pid" as this value is called everywhere else.

paulmartel:
End of review round. Good catch.

PR #2667, opened 2015-08-06
ENG-8809: Add logging and rate limit when needed.

akhanzode:


akhanzode:
squashed r=manjujames

PR #2668, opened 2015-08-06
ENG-8708: Upgrade JNA library to version 4.1.0.

nshi:
Also updated the third-party license text of JNA from LGPL to Apache.


luyangco:
Merged.

Yang Lu, Software Engineer
*VoltDB, Inc.*
*E: *ylu@voltdb.com

PR #2685, opened 2015-08-11
ENG-8554: Enable TestCommandLog JUnit test and isolate more DRv1 logic

wweiss-voltdb:


src/frontend/org/voltcore/messaging/TransactionInfoBaseMessage.java
```diff
@@ -139,7 +139,7 @@ public boolean isForReplay() {
         return m_isForReplay;
     }
 
-    public boolean isForDR() {
```
nshi: We can probably get rid of this API.

src/frontend/org/voltdb/iv2/SpProcedureTask.java
```diff
@@ -50,7 +50,7 @@
         HOST_TRACE_ENABLED = hostLog.isTraceEnabled();
     }
 
-    SpProcedureTask(Mailbox initiator, String procName, TransactionTaskQueue queue,
+    public SpProcedureTask(Mailbox initiator, String procName, TransactionTaskQueue queue,
```
nshi: This seems like an unnecessary change.

wweiss-voltdb: It needs to be public because the TestCommandLog is not part of the iv2 package.

src/frontend/org/voltcore/messaging/TransactionInfoBaseMessage.java
wweiss-voltdb: I agree. I came across it and wanted to make sure we did not use it with an assert for a couple of weeks before removing it.

nshi:
Merged externally

PR #2689, opened 2015-08-12
Eng 8835

akhanzode:


vtkstef:
merged into ENG-8817-counter which was --squash merged into master

PR #2690, opened 2015-08-12
Eng 6833 sqlcov non ascii

johnTcrawford:
Optionally (& by default), adds various non-ASCII characters to the list of randomly chosen characters used to generate VARCHAR values.


tests/scripts/SQLGenerator.py
```diff
@@ -775,7 +803,8 @@ def generate_statements_from_text(self, text):
 
 
 class SQLGenerator:
-    def __init__(self, catalog, template, subversion_generation):
+    def __init__(self, catalog, template, subversion_generation, ascii_only = False):
```
paulmartel: Is the default value necessary/helpful? I'm thinking that the value is available to the main caller as a boolean value -- I don't see much harm in forcing the hand of any other callers (existing or hypothetical future) to force them to explicitly decide on a value.

paulmartel:
Looks good.

paulmartel:
Good to go.

johnTcrawford:
Thanks, but it already went, on Friday - that is, I merged the squashed version of this branch (after checking with you, Paul). I'll close this Pull Request, and Branch, to avoid any further confusion, though.

PR #2691, opened 2015-08-12
Fix catgen's bug

yzhang1991:
If one catalogType class has more than one catalogMaps with same type signature, catalog.py will generate wrong C++ code.


paulmartel:
Looks good to merge.

PR #2693, opened 2015-08-12
ENG-8348: Implement SQL Is Distinct from comparison operator and juni…

hhakimi53:
Review request for implementation of SQL "is distinct from" operator.  
Changes include: 
- Logic update to hsql, frontend and ee components to support "is distinct from" operator
- Update query plan logic in hsql to not parameterize null constant for volt null type in query plan and use the plan with null constant instead.
- Add junit tests for is distinct from operator and moved case/when to the comparison operator junit test file.
- Add "is distinct from" operator in sql coverage tests.

Testing: Passed all Jenkins tests against the changes except SQL coverage test which is still in progress.


wolffcm:
I'll have a look at this one.

src/ee/common/types.h
```diff
@@ -276,6 +276,7 @@ enum ExpressionType {
     EXPRESSION_TYPE_COMPARE_GREATERTHANOREQUALTO    = 15, // greater than equal operator between left and right
     EXPRESSION_TYPE_COMPARE_LIKE                    = 16, // LIKE operator (left LIKE right). both children must be string.
     EXPRESSION_TYPE_COMPARE_IN                      = 17, // IN operator [left IN (right1, right2, ...)]
+    EXPRESSION_TYPE_COMPARE_NOTDISTINCT             = 19, // Not distinct operator between left and right
```
wolffcm: Why not use 18, since it's the next available number?

hhakimi53: I had chosen 18 earlier though found that it's already in use by "EXPRESSION_TYPE_OPERATOR_EXISTS". So chose 19 for is distinct operator as that was next available number. I agree if it sequence order was maintained in there, it's much better  to read through.

src/ee/expressions/comparisonexpression.h
```diff
@@ -235,5 +253,40 @@ class InlinedComparisonExpression : public ComparisonExpression<C> {
     {}
 };
 
+
+class NotDistinctExpression : public AbstractExpression {
```
wolffcm: I'm wondering why you have a `NotDistinctExpression` class when the other comparison ops use class templates like `InlinedComparisonExpression<CmpGt>`.  Count NOT DISTINCT FROM follow a similar pattern?

src/ee/expressions/comparisonexpression.h
```diff
@@ -62,85 +62,103 @@ namespace voltdb {
 
 // Each of these OP classes implements a standard static function interface
 // for a different comparison operator assumed to apply to two non-null-valued
-// NValues.
-// "compare_withoutNull" delegates to an NValue method implementing the specific
+// NValues (**except CmpNotDistinct, which can take two null values).
```
wolffcm: If we never expect the compare methods for most of the OP classes to be passed a null, it might be worth adding asserts to that effect?

src/ee/common/types.h
wolffcm: How long did it take you to realize that 18 was already taken?  I think Mike A. must have chosen 18 for EXISTS because there was no more room in the EXPRESSION_TYPE_OPERATOR category.  Looks like the next EXPRESSION_TYPE_COMPARE will have to be number 22.

Maybe file a ticket to renumber these so there's a little more elbow room to add new stuff?

src/ee/expressions/comparisonexpression.h
```diff
@@ -62,85 +62,103 @@ namespace voltdb {
...
-// "implies_true_for_row" returns true if a prior true return from compare_withoutNull
+// "implies_true_for_row" returns true if a prior true return from compare
 // applied to a row's prefix column implies a true result for the row comparison.
 // This may require a recheck for strict inequality.
-// "implies_false_for_row" returns true if a prior false return from compare_withoutNull
+// "implies_false_for_row" returns true if a prior false return from compare
 // applied to a row's prefix column implies a false result for the row comparison.
 // This may require a recheck for strict inequality.
 // "includes_equality" returns true if the comparison is true for (rows of) equal values.
+// isNullRejecting() returns true if the comparison does not consider NULL values as valid ones
```
wolffcm: I think by "reject" you mean that it will fail an assertion at run time in debug builds, is that right?

src/frontend/org/voltdb/planner/ParameterizationInfo.java
```diff
@@ -164,6 +164,14 @@ static void parameterizeRecursively(VoltXMLElement node,
             String idStr = node.attributes.get("id");
             assert(idStr != null);
 
+            // handle parameter value type
+            String typeStr = node.attributes.get("valuetype");
+            VoltType vt = VoltType.typeFromString(typeStr);
+            // don't parameterize the plan for NULL constants of type "VoltType.NULL"
```
wolffcm: For what kind of SQL syntax will VoltType.NULL occur?  Just NULL literals?

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+
+        VoltServerConfig config = null;
+        MultiConfigSuiteBuilder builder =
+            new MultiConfigSuiteBuilder(TestComparisonOperatorsSuite.class);
+        boolean success;
+
+        VoltProjectBuilder project = new VoltProjectBuilder();
+        try {
+            setUpSchema(project);
+            // CONFIG #1: Local Site/Partitions running on JNI backend
```
wolffcm: By convention, most of our regressionsuite tests have all the configs listed last.  Frequently, we might edit a test so that it runs on just one configuration (maybe IPC because we need to debug the EE, or with (1, 1, 0) if we want the test to run faster).  So putting the suite() method last in the file will make the configs easier to find.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+                "VC1 VARCHAR(6)," +     // inlined
+                "VC2 VARCHAR(16)," +    // not inlined
+                "VB1 VARBINARY(6)," +   // inlined
+                "VB2 VARBINARY(64));" + // not inlined
+                "";
+
+            project.addLiteralSchema(literalSchema);
+    }
+
+    private void populateTableForIsDistinctFromTests(Client client)
```
wolffcm: I would put this method higher up in this file, near the methods that use this data, so the tests are easier to read and understand what the results should be.

tests/scripts/examples/sql_coverage/grammar.sql
```diff
@@ -49,6 +49,7 @@
 {_cmp |= "<="}
 {_cmp |= ">="}
 {_cmp |= "!="} -- Apparently, an HSQL-supported alias for the standard <>
+{_cmp |= "IS DISTINCT FROM"}
```
wolffcm: This should be "IS NOT DISTINCT FROM"?

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+
+public class TestComparisonOperatorsSuite  extends RegressionSuite {
+    //
+    // JUnit / RegressionSuite boilerplate
+    //
+    public TestComparisonOperatorsSuite(String name) {
+        super(name);
+    }
+
+    public void testIsDistinctFrom() throws Exception
```
wolffcm: I think you mean "IsNotDistinctFrom" here and in other places?

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+            throws Exception
+    {
+        String sql;
+        VoltTable vt;
+
+        // NULL constant - results in non-parameterized plan
+        sql = "SELECT * FROM S2 A WHERE  A.WAGE is not distinct from NULL;";
+        vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertEquals(vt.getRowCount(), 1);
+        if(!isHSQL()) {
```
wolffcm: I'm not sure why this is guarded.  Does this not work for the HSQL backend?

src/ee/expressions/comparisonexpression.h
hhakimi53: In current implementation, by reject, it meant do not take into account null values. I will be adding asserts to  comparators compare(). So for upcoming review it will reject will mean assertion in debug build if compare is called with null values.

src/frontend/org/voltdb/planner/ParameterizationInfo.java
hhakimi53: Yes, at least for NULL literals, type gets resolved to VoltType.NULL (I have refered Null literals as Null constant here. if that's not correct terminology, I can update the comment. pls let me know).
later when type is tried to be resolved to it's value, it fails to resolve value for VoltType.Null. So that's the intend here, let's not have parameterize the plan for Null constants/literals.

src/ee/expressions/comparisonexpression.h
hhakimi53: will do. thnxs

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: will do, thnxs

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: sure, thnxs

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+            validateTableOfLongs(client, sql, new long[][] {{4,     Long.MIN_VALUE,     2}});
+        }
+
+        // Non-Null constant results in parameterized plan
+        sql = "SELECT * FROM S2 A WHERE  A.WAGE is  distinct from 1000.01;";
```
wolffcm: Oh, I guess we're supporting both "is distinct" and "is not distinct".  I didn't realize it until just now.

wolffcm: I guess IsDistinct gets written as NOT(isNotDistinct), and that's how the EE sees it?

hhakimi53: Yes, we are supporting both "is distinct from" and "is not distinct from". HSQL does all that job for us and in EE we see "isDistinctFrom", as u mentioned, as NOT(isDistinctFrom).

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+                                                            {9,  5152,  2,  5253,           3}});
+        }
+
+        // left join on column that has null values
+        sql = "Select S2.wage, S2.ID, count (*) from S1 left Join S2 On S2.WAGE is not distinct from S2.wage group by S2.wage, S2.ID;";
+        vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertEquals(vt.getRowCount(), 3);
+
+        if(!isHSQL()) {
+            validateTableOfLongs(client, sql, new long[][] {{5253,              5,   5},
```
wolffcm: Left join is an interesting case.  Maybe do another version of this query without the group by?  You might hit one of the existing join bugs we have on the HSQL branch though.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+            validateTableOfLongs(client, sql, new long[][] {{Long.MIN_VALUE,    1}});
+        }
+    }
+
+    private void subTestIsDistinctFromInCompatibleTypes(Client client) throws Exception {
+        try {
+            String sql;
+            sql = "SELECT * FROM S1 A WHERE  A.WAGE is distinct from \'Z\';";
+            client.callProcedure("@AdHoc", sql);
+            fail("Expected incompatible data types exception");
```
wolffcm: You can also use verifyStmtFails in the RegressionSuite class to do this more concisely.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: With HSQL backend, validation test fails as it computes to a different value instead of Long.MIN_VALUE. So for all the places we have null, which is seen as Long.MIN_VALUE, they are gaurded to not be executed against hsql backend. if there is another way to execute, it will be better. that way we will get stricter result checks. pls let me know. thnxs

src/frontend/org/voltdb/planner/ParameterizationInfo.java
wolffcm: No, your terminology is fine.  I'm not so familiar with this code.  I didn't phrase my question very well.  I meant to ask are NULL constants the *only* place where we get VoltType.NULL from?  

I guess I'm curious about why we didn't hit this issue before.  But I guess "is null" is its own operator, maybe that's why.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
wolffcm: Hmm.  Can we really not check for NULLs in result sets produced by the HSQL backend?  I'd check with Paul or Xin on this one.  I'd feel a lot better if we had validation on the HSQL side.

src/frontend/org/voltdb/planner/ParameterizationInfo.java
hhakimi53: For Is_Null operator, it takes different path and I have tested that path to see how it behaves w.r.t. Null types.

To resolve types (resolveTypes()), for most of the comparison operators like "=, >, ..." resolveTypesForComparison() is used to resolve left and right child of a node of an expression tree (does a lot more work in here and handles child with null constants by converting/morphing them to types of its sibling). That's how in getCleanParameters() in ProcedureRun class it's able to fetch appropriate values fine. So if there is null constant in SQL with other comparators, type is resolved right here (resolveTypes()). So later it does not get effected in parameterization for Null constants for most of the comparator operators. 

For isNotDistinctm in resolveTypes(), child elements are converted to row expression first and row expression type is resolved though after processing of this function still childs with Null constants and leads to issue if we parameterize it. It's bit convoluted in there how the flow happens, atleast for me :-).

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+
+    private void subTestIsDistinctFromUsingSubqueries(Client client) throws Exception
+    {
+        String sql;
+        VoltTable vt;
+
+        // test cases below test different subquery condition paths in EE like LHS NULL, RHS NOT NULL and so forth
+        sql = "SELECT wage salary, count(*) from S2 "+
```
wolffcm: In this query you're exercising the new operator in the WHERE clause, the HAVING clause and in the WHERE clause in an inner query.  That's good!

It can be difficult to think about what the result should be though, reading a query like this.

My usual approach for something like this is to write simple queries that exercise the new functionality in just one place in the statement, so it's pretty obvious to someone reading the statement what the answer should be if they know how the tables are populated.

This is just my approach, others might choose to do things differently.

Is it possible to use "all" and "any" with your new operator?  Would be good to demonstrate that here if so.

wolffcm:
I've finished this review round.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: I was wrong in here, please ignore my earlier statement. I was confusing/mixing it with something else, sorry. 

I removed gaurd from one of places and we were able to validate hsql as well as ee result set with null it. I will walk through all of them again and start taking off the guards. Thanks for highlighting it.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: That's good to know, will use that.

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: Will give it a try for all and any with the new operator.

For me, the queries with nested sub-queries are bit complex too to deduce it's result. For validation purpose, used postgres to verify the results match. hope is it stresses a lot of different combinational paths in there. thnxs

hhakimi53:
Closing the pull request. Changes to support is distinct from has been merged into master.

PR #2695, opened 2015-08-13
Eng 8780 export conflicts

luyangco:
Generate export table for every DR table when A/A DR is enabled, the export table contains all columns of its correspondent DR table, plus four additional columns for logging DR conflicts purpose.

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -38,7 +38,6 @@ import java.util.TreeMap;
 import java.util.regex.Matcher;
 
 import org.apache.commons.lang3.StringUtils;
-import org.hsqldb_voltpatches.FunctionSQL;
 import org.hsqldb_voltpatches.HSQLDDLInfo;
 import org.hsqldb_voltpatches.HSQLInterface;
 import org.hsqldb_voltpatches.HSQLInterface.HSQLParseException;
@@ -51,15 +50,15 @@ import org.voltdb.VoltType;
 import org.voltdb.catalog.CatalogMap;
 import org.voltdb.catalog.Column;
 import org.voltdb.catalog.ColumnRef;
-import org.voltdb.catalog.IndexRef;
 import org.voltdb.catalog.Constraint;
 import org.voltdb.catalog.Database;
+import org.voltdb.catalog.DatabaseConfiguration;
 import org.voltdb.catalog.Group;
 import org.voltdb.catalog.Index;
+import org.voltdb.catalog.IndexRef;
 import org.voltdb.catalog.MaterializedViewInfo;
 import org.voltdb.catalog.Statement;
 import org.voltdb.catalog.Table;
-import org.voltdb.catalog.DatabaseConfiguration;
 import org.voltdb.common.Constants;
 import org.voltdb.common.Permission;
 import org.voltdb.compiler.ClassMatcher.ClassNameMatchStatus;
@@ -71,8 +70,6 @@ import org.voltdb.expressions.AbstractExpression;
 import org.voltdb.expressions.AbstractSubqueryExpression;
 import org.voltdb.expressions.AggregateExpression;
 import org.voltdb.expressions.ExpressionUtil;
-import org.voltdb.expressions.FunctionExpression;
-import org.voltdb.expressions.SelectSubqueryExpression;
 import org.voltdb.expressions.TupleValueExpression;
 import org.voltdb.groovy.GroovyCodeBlockCompiler;
 import org.voltdb.parser.HSQLLexer;
@@ -139,6 +136,14 @@ public class DDLCompiler {
     final ClassLoader m_classLoader;
 
     private final Set<String> tableLimitConstraintCounter = new HashSet<>();
+
+    // Addition columns for DR conflicts table
+    private static final String [] ADDITIONAL_COLUMNS = {
+        "TABLE_NAME VARCHAR",  // actual size of this column will be calculated at runtime
+        "CLUSTER_ID TINYINT",
+        "TIMESTAMP BIGINT",
+        "OPERATION_TYPE TINYINT"
+    };
 
     private class DDLStatement {
         public DDLStatement() {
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -140,6 +137,14 @@
 
     private final Set<String> tableLimitConstraintCounter = new HashSet<>();
 
+    // Addition columns for DR conflicts table
+    private static final String [] ADDITIONAL_COLUMNS = {
```
afrogers: `ADDITIONAL_COLUMNS` is a kind of generic name, given that this is sitting in the global scope of the DDLCompiler. Should be qualified with something telling us that it's for DR active-active

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+
+        sb.append("CREATE TABLE " + tableName + " (");
+        for (VoltXMLElement subnode : node.children) {
+            // Indexes and constraints are not needed for export table
+            if (subnode.name.equals("columns")) {
+                // Insert additional columns to DR conflicts table first
+                for (String column : ADDITIONAL_COLUMNS) {
+                    sb.append(column);
+                    if (column.startsWith("TABLE_NAME")) {
+                        sb.append("(" + node.attributes.get("name").length() + " BYTES" + ")");
```
afrogers: Why is the column name specified in bytes? Fine if we don't allow arbitrary unicode, but could just specify in characters.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                    }
+                    sb.append(", ");
+                }
+                // Then duplicate all columns of the original table
+                int columnIdx = 0;
```
afrogers: Always fun to see the variety of ways people add a comma after all but the last thing in some list...

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+        if (node.attributes.containsKey("partitioncolumn")) {
+            sb.append("PARTITION TABLE " + tableName + " ON COLUMN " + node.attributes.get("partitioncolumn") + ";\n");
+        }
+        sb.append("EXPORT TABLE " + tableName + " TO STREAM " + CatalogUtil.DR_CONFLICTS_TABLE_EXPORT_GROUP + ";\n");
+    }
+
+    // Generate DDL to create or drop the DR conflict table
+    // TODO:When DR table supports dynamic schema change, we need also change the schema of conflict table,
+    //       maybe dropping the old one and recreating a new one is the easiest way.
+    private String generateDdlForDRConflictsTable(Database currentDB, Database previousDBIfAny) {
```
afrogers: Minor nitpick: s/Ddl/DDL/

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                String drConflictExportTableName = CatalogUtil.DR_CONFLICTS_TABLE_PREFIX + node.attributes.get("name");
+                // Does the conflict export table already existed?
+                if (previousDBIfAny != null) {
+                    tableAlreadyExist = previousDBIfAny.getTables().get(drConflictExportTableName) != null;
+                } else {
+                    for (VoltXMLElement element : xmlRoot.children) {
+                        if (element.name.equals("table")
+                                && element.attributes.containsKey("export")
+                                && element.attributes.get("name").equals(drConflictExportTableName)) {
+                            tableAlreadyExist = true;
```
afrogers: break here

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                        }
+                    }
+                }
+                if (!tableAlreadyExist && currentDB.getIsactiveactivedred() == true) {
```
afrogers: If we're not active-active, we should simply not enter this loop. Otherwise we're doing a bunch of redundant work

src/frontend/org/voltdb/compiler/DDLCompiler.java
luyangco: It is for the purpose of conserving space, Hsql only allows ASCII characters table name, using BYTES instead of characters could save extra space, especially for a table has thousands of hundred rows.
In the VOLTDB CREATE TABLE documentation, it says, 

>Specifying the VARCHAR length in characters is recommended because UTF-8 characters can require a variable number of bytes to store. By specifying the length in characters you can be sure the column has sufficient space to store any string of the specified length. Specifying the length in bytes is only recommended when all values contain only single byte (ASCII) characters or when conserving space is required and the strings are less than 64 bytes in length

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                    createConflictExportTableDDL(sb, node, drConflictExportTableName);
+                }
+            }
+        }
+
+        // Drop the dr conflicts table if corresponding dr table is not existed or A/A is disabled or dr is disabled.
+        for (VoltXMLElement node : xmlRoot.children) {
+            if (node.name.equals("table")
+                    && node.attributes.get("name").startsWith(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX)) {
+                String drTableName = node.attributes.get("name").replaceAll(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX, "");
```
afrogers: Theoretically unsafe; I mean, I seriously doubt anyone's table name contains the literal 'AUTOGEN_VOLTDB_DR_CONFLICTS__', but it's technically allowable. Better to create a regex outside the loop like: `Pattern conflictsPattern = Pattern.compile(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX + "([a-zA-Z][a-zA-Z0-9]*);`. Then, inside the loop, you can replace the if check with:
```
Matcher matcher;
if (node.name.equals("table") && (matcher = 
        conflictsPattern.matcher(node.attributes.get("name"))).matches()) {
    String drTableName = matcher.group(1);
    ...
```

src/frontend/org/voltdb/compiler/DDLCompiler.java
luyangco: Will change to DR_CONFLICTS_EXPORT_TABLE_ADDITIONAL_COLUMNS, makes it more specific.

src/frontend/org/voltdb/compiler/DDLCompiler.java
afrogers: Makes sense, then.

src/frontend/org/voltdb/compiler/DDLCompiler.java
luyangco: Yeah, annoying comma.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                }
+                if (!tableAlreadyExist && currentDB.getIsactiveactivedred() == true) {
+                    // If the conflict export table doesn't existed yet, create a new one.
+                    createConflictExportTableDDL(sb, node, drConflictExportTableName);
+                }
+            }
+        }
+
+        // Drop the dr conflicts table if corresponding dr table is not existed or A/A is disabled or dr is disabled.
+        for (VoltXMLElement node : xmlRoot.children) {
```
afrogers: Again, a lot of logic could be spared here if active-active is disabled. If that's the case, we could instead just put directly at the top of this whole method a loop like:
```
if (!currentDB.getIsactiveactivedred()) {
    for (VoltXMLElement node : xmlRoot.children) {
        if (node.name.equals("table") && 
            node.attributes.get("name").startsWith(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX)) {
            sb.append("DROP TABLE " + node.attributes.get("name") + ";\n");
        }
    }
    return;
}
```
Then we don't have to deal any further with the no active-active case.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +225,118 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                    && node.attributes.get("name").startsWith(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX)) {
+                String drTableName = node.attributes.get("name").replaceAll(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX, "");
+                boolean remove = true;
+                for (VoltXMLElement element : xmlRoot.children) {
+                    if (element.name.equals("table")
+                            && element.attributes.containsKey("drTable")
+                            && element.attributes.get("name").equals(drTableName)) {
+                        remove = false;
+                    }
+                    if (element.name.equals("table")
```
afrogers: Isn't this missing a check that we're talking about the right table? Also, couldn't this case be elided into the above one by adding an extra condition `&& !element.attributes.get("drTable").equalsIgnoreCase("DISABLE")``?

src/frontend/org/voltdb/compiler/DDLCompiler.java
luyangco: Done.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2140,4 +2153,50 @@ public static String getLimitPartitionRowsDeleteStmt(Table table) {
...
+     * @param export   list of export configuration
+     */
+    public static ExportType addExportConfigToDRConflictsTable(Catalog catalog, ExportType export) {
+        if (export == null) {
+            export = new ExportType();
+        }
+        boolean userDefineStream = false;
+        for (ExportConfigurationType exportConfiguration : export.getConfiguration()) {
+            if (exportConfiguration.getStream().equals(DR_CONFLICTS_TABLE_EXPORT_GROUP)) {
+                userDefineStream = true;
```
afrogers: break here

tests/frontend/org/voltdb/compiler/TestDDLCompiler.java
```diff
@@ -729,4 +730,28 @@ void testFailure(String ddl) {
...
+        File schemaFile = VoltProjectBuilder.writeStringToTempFile(
+        "SET " + DatabaseConfiguration.DR_MODE_NAME + "=" + DatabaseConfiguration.ACTIVE_ACTIVE + ";\n" +
+        "CREATE TABLE T (D1 INTEGER NOT NULL, D2 INTEGER, D3 VARCHAR(32), VAL1 INTEGER, VAL2 INTEGER, VAL3 INTEGER, PRIMARY KEY (D1), LIMIT PARTITION ROWS 1000);\n" +
+        "DR TABLE T;\n" +
+        "PARTITION TABLE T ON COLUMN D1;\n");
+        String schemaPath = schemaFile.getPath();
+
+        try {
+            assertTrue(compiler.compileFromDDL(jarOut.getPath(), schemaPath));
+            assertNotNull(compiler.getCatalogDatabase().getTables().get(CatalogUtil.DR_CONFLICTS_TABLE_PREFIX + "T"));
```
afrogers: Also validate that the export table has the expected schema

afrogers:
Review complete

src/frontend/org/voltdb/compiler/DDLCompiler.java
luyangco: The above one should check to see if DR is on, but here the behavior is intended to drop the any conflicts export table when DR is off.

luyangco: never mind, I misunderstood your comment, you are right.

luyangco:
There is merge conflicts between my local and remote branch. Address all comments in another PR.

PR #2699, opened 2015-08-13
ENG-8840 plan based view catch up

yzhang1991:


src/ee/storage/persistenttable.h
```diff
@@ -210,6 +212,9 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
   public:
     virtual ~PersistentTable();
 
+    void setViewsUpdateEnabled(bool enabled) { m_viewsUpdateEnabled = enabled; }
+    bool isViewsUpdateEnabled() { return m_viewsUpdateEnabled; }
```
paulmartel: Is there code missing from this change set?
Where's the caller of setViewsUpdateEnabled?

Is this public getter function actually needed?

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -303,13 +303,6 @@ void setupForNewPlans(AbstractParsedStmt parsedStmt) {
             return;
         }
 
-        // @TODO
-        // Need to use StmtTableScan instead
-        // check that no modification happens to views
-        if (tableListIncludesView(parsedStmt.m_tableList)) {
-            throw new PlanningErrorException("Illegal to modify a materialized view.");
-        }
-
```
paulmartel: Was this exception never thrown previously -- is/was an end-user's attempt to update a materialized view caught somewhere else/earlier? Is there a junit test validating that such a guard still exists?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1971,33 +1972,42 @@ else if (type != ConstraintType.PRIMARY_KEY &&  type != ConstraintType.UNIQUE) {
         catalog_const.setType(type.getValue());
     }
 
+    void compileMatViewQueriesToStatement(Statement stmt, VoltXMLElement xmlquery,
```
paulmartel: Combination of plurals and singulars make the function name hard to understand. How about  compileMatViewQueryToStatement

paulmartel:
Do we have any existing unit tests that populate a table with a view -- we really should.
Then we should add a test case that populates two identical tables, one with view update enabled, one with it disabled, and validates that I can call the buildUp query to bring the two views into synchronization.

src/frontend/org/voltdb/compiler/MatViewQueriesXMLGenerator.java
```diff
@@ -56,6 +67,38 @@ private String lastElementId() {
         return String.valueOf( m_maxElementId );
     }
 
+    private void adjustColumnAliases() {
+        ArrayList<String> realColumnNames = new ArrayList<String>(Collections.nCopies(m_destTable.getColumns().size(), ""));
```
paulmartel: What's the purpose of these "" values vs. null? Is it really OK for these "" values to survive the first loop and get used as alias values? What does an alias value of "" mean?

paulmartel:
End of review round.

yzhang1991:
Need to remove all the debug outputs.
(This ticket and the ENG-8641 ticket)

src/frontend/org/voltdb/planner/PlanAssembler.java
yzhang1991: We have a guard against modifying materialized view in hsql.

src/frontend/org/voltdb/compiler/MatViewQueriesXMLGenerator.java
```diff
@@ -56,6 +67,38 @@ private String lastElementId() {
...
+        Collections.fill(realColumnNames, null);
+        for (Column info : m_destTable.getColumns()) {
+            realColumnNames.set(info.getIndex(), info.getName());
+        }
+        List<VoltXMLElement> viewColumns = VoltXMLElementHelper.getFirstChild(m_xml, "columns").children;
+        for (int i=0; i<viewColumns.size(); ++i) {
+            VoltXMLElement columnref = viewColumns.get(i);
+            String realColumnName = realColumnNames.get(i);
+            columnref.attributes.put("alias", realColumnName);
```
paulmartel: add assert( ! "".equals(realColumnName))

PR #2700, opened 2015-08-14
Add disabling threads in txnid2

liam700:
Adding option to txnid2 app that allows disabling different threads when running the client.


liam700:
resolving conflict

PR #2703, opened 2015-08-14
Added turning off threads in TxnId2

liam700:
Added the ability to turn off threads in TxnId2 app with a run parameter.


liam700:
corrected some minor things

PR #2706, opened 2015-08-14
Eng 8780 export conflicts prime

luyangco:
Incorporated Alex's review comments.

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -38,7 +38,6 @@ import java.util.TreeMap;
 import java.util.regex.Matcher;
 
 import org.apache.commons.lang3.StringUtils;
-import org.hsqldb_voltpatches.FunctionSQL;
 import org.hsqldb_voltpatches.HSQLDDLInfo;
 import org.hsqldb_voltpatches.HSQLInterface;
 import org.hsqldb_voltpatches.HSQLInterface.HSQLParseException;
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -140,6 +138,22 @@
...
+    private static final String [] DR_CONFLICTS_EXPORT_TABLE_ADDITIONAL_COLUMNS = {
+        "VOLTDB_AUTOGEN_TABLE_NAME VARCHAR",  // actual size of this column will be calculated at runtime
+        "VOLTDB_AUTOGEN_CLUSTER_ID TINYINT",
+        "VOLTDB_AUTOGEN_TIMESTAMP BIGINT",
+        "VOLTDB_AUTOGEN_OPERATION_TYPE TINYINT"
+    };
+
+    private static final Pattern drConflictsTablePattern = Pattern.compile(
+            CatalogUtil.DR_CONFLICTS_TABLE_PREFIX +             /* table prefix */
+            "(?i)" +                                            /* ignore case instruction */
```
afrogers: Ignore case seems weird, given that we're just matching any word character. Harmless, but weird.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +234,119 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+        // Drop the dr conflicts table if corresponding dr table is not existed or A/A is disabled or dr is disabled.
+        Matcher matcher;
+        for (VoltXMLElement node : xmlRoot.children) {
+            if (node.name.equals("table")
+                    && (matcher = drConflictsTablePattern.matcher(node.attributes.get("name"))).matches()) {
+                boolean remove = true;
+                if (currentDB.getIsactiveactivedred()) {
+                    for (VoltXMLElement element : xmlRoot.children) {
+                        if (element.name.equals("table")
+                                && element.attributes.containsKey("drTable")
```
afrogers: It would be better to capture the `matcher.group(1)` just once outside this loop rather than calling it with each pass. Every call to `group()` will make a new substring.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -220,6 +234,119 @@ public void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProc
...
+                                && element.attributes.get("drTable").equalsIgnoreCase("ENABLE")
+                                && element.attributes.get("name").equals(matcher.group(1))) {
+                            remove = false;
+                            break;
+                        }
+                    }
+                } else {
+                    remove = true;
```
afrogers: `else` is unnecessary -- value is already `true` by default.

afrogers: Seeing this whole thing again, I still feel like it might be clearer to have two methods -- one called something like `dropAllDRConflictTables()`, and then this one, but with the `currentDb.getIsactivedred()` stuff stripped out.

afrogers:
Second review complete.

afrogers:
No further comments. I'll merge once Jenkins is happy.

afrogers:
Merged externally

PR #2709, opened 2015-08-17
Init the global variables once per host, not each time for every site

jxchinaren:


src/ee/common/executorcontext.cpp
```diff
@@ -32,6 +35,33 @@ static pthread_once_t static_keyOnce = PTHREAD_ONCE_INIT;
 
 static void createThreadLocalKey() {
     (void)pthread_key_create( &static_key, NULL);
+
```
wolffcm: This seems like a good place to put stuff that has to happen exactly once for the whole process, but maybe the name of this method should be changed to "processWideInit" or something?

Also, pthread_key_create might call malloc under the hood, do we want call it after we've made the calls to mallopt?

wolffcm:
That's all my comments.

jxchinaren:
@wolffcm I just resolved your comments. Thanks.

wolffcm:
Looks good to me.

jxchinaren:
I will close this pull request to squash merge with one single commit, it's better to cherry pick and patch to previous version for Ruth.

PR #2710, opened 2015-08-17
Fix crash when creating expression index

wolffcm:
This issue was in code called from Expression.getSQL.  It seems like we only use this method when creating expression indexes.  We use the expression text to create a hash for the index in case we need an auto-generated name.  So the SQL text generated via this method isn't actually executed, it's just used to generate unique names.  So this is a low-risk change.

We don't use this method for other places where we capture SQL, e.g., views.  In those cases we capture text at the token level.


paulmartel:
Looks good. Please notify Andrew to add this to the 5.6 release notes list of fixes.

PR #2711, opened 2015-08-17
Eng 8469 dont fsync unchanged command log

wweiss-voltdb:


src/frontend/org/voltdb/CommandLog.java
```diff
@@ -124,6 +126,12 @@ public abstract void logIv2Fault(long writerHSId, Set<Long> survivorHSId,
          * @param completionChecks
          */
         void processDurabilityChecks(CompletionChecks completionChecks);
+
+        /**
+         * Make sure Read-Only SysProcs get executed in with Synchronous Command Logging
```
nshi: They could be read-write sysprocs. The `isDurable()` flag on the sysprocs just indicates whether that sysproc should be logged or not, doesn't have a strong correlation to the read-onliness of the procedure.

PR #2715, opened 2015-08-18
ENG-8779: Prevent UpdateApplicationCatalog from changing hostcount

wweiss-voltdb:


nshi:
Reviewed, will merge once master is open.

PR #2716, opened 2015-08-18
Update THP error message to include instructions for setting the corr…

afrogers:
…ect value

### DIFF
--- a/lib/python/voltcli/checkconfig.py
+++ b/lib/python/voltcli/checkconfig.py
@@ -60,8 +60,8 @@ def _check_thp_config():
         return "The kernel is configured to use transparent huge pages (THP). " \
             "This is not supported when running VoltDB. Use the following commands " \
             "to disable this feature for the current session:\n\n" \
-            "bash -c \"echo never > /sys/kernel/mm/{0}transparent_hugepage/enabled\"\n" \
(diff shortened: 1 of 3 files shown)


lib/python/voltcli/checkconfig.py
```diff
@@ -25,17 +25,36 @@
...
+    has_error = check_thp_files(file_prefix)
+    if has_error is None:
+        file_prefix = ""
+        has_error = bool(check_thp_files(file_prefix))
+    if has_error:
+        return "The kernel is configured to use transparent huge pages (THP). " \
+            "This is not supported when running VoltDB. Use the following commands " \
+            "to disable this feature for the current session:\n\n" \
+            "bash -c \"echo never > /sys/kernel/mm/{0}transparent_hugepage/enabled\"\n" \
+            "bash -c \"echo never > /sys/kernel/mm/{0}transparent_hugepage/defrag\"\n\n" \
```
nshi: Based on the recent feedback, I think prepending `sudo` will help most people.

nshi:
I also noticed that there is another script that does a bunch of checks, including the THP, in the voltdb.d subdirectory. That script should at least call the functions in this script instead of doing its own checks.

https://github.com/VoltDB/voltdb/blob/master/lib/python/voltcli/voltdb.d/check.py#L141

PR #2719, opened 2015-08-19
Eng 8645 fewer NDOs

BillWhite:
Jenkins is all blue with this.


src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -289,8 +289,18 @@ public String toString() {
 
     private static final String INDENT = "  | ";
 
+    /**
+     * Return a node name to help out toString.  Subclasses
+     * can chime in if they have a notion to.  See TupleValueExpression,
+     * for example.
+     */
+    protected String getExpressionNodeNameForToString() {
+        return null;
```
paulmartel: It's clumsy to return null only to test for null in the caller.
Why not have this function return this.getClass().getSimpleName() and let TupleValueExpression call this as its super method and append its details to the result.

src/frontend/org/voltdb/expressions/TupleValueExpression.java
```diff
@@ -403,4 +403,24 @@ else if ( ! tableName.equals(impliedTableName)) {
...
+                return String.format("%s AS %s", name, alias);
+            } else {
+                return name;
+            }
+        } else if (alias != null) {
+            return String.format ("<> AS %s", alias);
+        } else {
+            return "<none>";
+        }
+    }
```
paulmartel: I think it's the whitespace in this format that really hurts readability: "<> AS x.<> AS y" makes me think too much.
It does not need to be beautiful but it also does not have to be quite so ugly.
How about the following formats: "tablename", "tablename(aliasname), or "(aliasname)" or "<none>" or maybe use [ or { or < in place of ( , maybe use empty empty punctuation in place of "<none>"?
Use "maybe just use an empty string when name is missing, so we'd have tablename.columnname or tablename(tablealias).columnname
I think I remember that Chris had similar issues in an earlier review.

paulmartel: I don't understand why the function is called chooseTwo... when it mostly doesn't.
Maybe call it something more like "aliasedName"?

src/frontend/org/voltdb/expressions/TupleValueExpression.java
```diff
@@ -403,4 +403,24 @@ else if ( ! tableName.equals(impliedTableName)) {
...
+
+    @Override
+    protected String getExpressionNodeNameForToString() {
+        return String.format("%s.%s",
+                             chooseTwoNames(m_tableName, m_tableAlias),
+                             chooseTwoNames(m_columnName, m_columnAlias));
```
paulmartel: Is String.format more readable than the "+" operator?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1385,16 +1387,27 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+        // different instances of the same table in self joins.
         HashMap<String, List<AbstractExpression> > baseTableAliases =
                 new HashMap<String, List<AbstractExpression> >();
         for (ParsedColInfo col : orderByColumns()) {
             AbstractExpression expr = col.expression;
             List<AbstractExpression> baseTVEs = expr.findBaseTVEs();
+            // This check is not actually right.  The statement:
+            //    select id from alpha order by alpha.x + alpha.y;
+            // is not table spanning, nor is it constant.  But the
+            // set of base TVEs will have two elements.
```
paulmartel: Good catch.
Is the right thing to be counting here the number of table aliases referenced by baseTVEs?
Can you write a repro case showing how/where this actually fails and either fix it now or write a ticket to fix it?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1385,16 +1387,27 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
             if (baseTVEs.size() != 1) {
                 // Table-spanning ORDER BYs -- like ORDER BY A.X + B.Y are not helpful.
                 // Neither are (nonsense) constant (table-less) expressions.
                 continue;
             }
+            // I'm not sure what this comment means.  - Bill W. 2015.08.06.
```
paulmartel: Me neither -- remove the comment.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     * <ol>
+     *   <li>Queries may have limits and offsets.  These just clutter up the
+     *       argument, so we ignore them here.</li>
+     *   <li>Let R be the relation formed by doing the joins and where-filtering,
+     *       and perhaps grouping, but before ordering and selecting.  The columns
+     *       of R are the columns of the tables in the FROM list, or else the
+     *       columns defined by the group-by expressions.  We assume that elements
+     *       of R all have unique indices drawn from a finite index set I.  We
+     *       can refer to elements of R as ri, rj and so forth.  The set I is not
+     *       necessarily ordered.  <b>Note</b> that are depends on the state of
```
paulmartel: What exactly would constitute a "column defined by the group-by expression"?
Why does it need to be called out here?
"that are depends"?

The terms "index" and "unique index" here mean what? "integer identifier" rather than "value-based table access method" I am GUESSING?" What are these indexes and where are they coming from? Are they positions in a list?
This bit about the state of the database tables is not telling me anything. What's "it" -- what's the "input variable"? (I've lost the precedent).

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     * <p>
+     * The set of order-by expressions is a set, so it doesn't have an order.
+     * But each expression in it comes from a list of columns, so there is
+     * an implicit order. Let's set some notation.  Our goal in the sequence
+     * of definitions is to define an <em>honorary order-by expression</em>,
+     * but getting there requires some work.
+     * <p>
+     * <ol>
+     *   <li>Queries may have limits and offsets.  These just clutter up the
+     *       argument, so we ignore them here.</li>
```
paulmartel: Not clear at all why these particular irrelevant details of the universe of possibles are being pointed out here as particularly irrelevant -- it seems to go without saying.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     *   <li>Let the select list be C1, ..., Cn, let CE1, ..., CEn
+     *       be the list of expressions of the columns and let CN1, ..., CNn be
+     *       the names of the columns.  These
+     *       CEks are expressions in the columns of R.  If there is a
+     *       group-by clause, the CEks must be factorable as CEk = CETk(GE1, ..., GEl),
+     *       where each GEk is a group-by expression.  We say this factorization
+     *       makes the select expressions <em>possibly grouped.</em>
+     *       (This is actually stronger
+     *       than the true requirement, but this will do for the purposes of this
+     *       note.)</li>
```
paulmartel: This sounds like "the story you are about to hear is a lie, but it's short". This is not helpful.
I don't know if the irrelevant inaccuracy you are owning up to is the same one I see.
Can you be clearer?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     *   <li>Let the original list of order-by expressions be E = E1, ..., Em.  Then
+     *       the Eks are functions of the columns of the
+     *       select list CE1, possibly grouped.</li>
```
paulmartel: I don't believe this to be the case. e.g. "select A from T order by B;".

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     *   <li>Given a list of expressions, E, and an expression, H, we write
+     *       E+H for the list E with H added to the end.  So, if E = E1, ..., Em, then
+     *       E+H = E1, ..., Em, H.</li>
+     *   <li>For each row of R,
+     *       say ri, we can calculate a tuple of values ce=[ce1, ..., cen] using the values
+     *       in row ri.  We then sort the set of ce tuples.  Since each such ce-tuple
+     *       is associated with an index, i, this sorting induces a partial ordering, [E[, on
```
paulmartel: Is "index i" supposed to be a row number (position) in R?
I don't understand the notation [E[. Typo?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     *       I, the index set of R.
+     *       <ul>
+     *         <li>We include the set of order-by expressions, E,
+     *             in the notation because the ordering relation depends on it.</li>
+     *         <li>We use square brackets for ordering to avoid needing HTML entities
+     *             in the JavaDoc.)</li>
```
paulmartel: Doh. NOW you tell me. Maybe it would be better to use English or to define the notation BEFORE using it.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     *       </ul>
+     *       Note that [E[ is only a partial order.  It may be that two rows, say ri and
+     *       rj, both have the same tuple value.  In this case i and j are incomparable.</li>
```
paulmartel: Don't know what incomparable means in this context -- equal I guess? The most interesting case is probably NOT when j and i have the same tuple value, but when they have some differences on which E does not depend.
I probably don't care about the relative handling of two identical rows.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+     *       be equal to G, then H is an honorary order-by expression, for similar reasons.
+     *       Similarly, for all expressions in the transitive closure of such constrained
+     *       equalities.  Not all of these are helpful.</li>
+     *   <li>Let's consider unique indices.  Each row, ri, of R can be written as ri = [ri1, ..., rin], where rik is
+     *       a tuple from table scan TSk.  Each table scan TSk has a a table, Tk, and
+     *       an alias, TAk.  Note that Tk and Tl may be the same table for k != l, but
+     *       TAk != TAl if k != l.  The table scan may have a table name, or the table may be a subquery,
+     *       which has only an alias.  Let table Tk have an index constraint IX, on expressions
+     *       IXE1, ..., IXEm.  Then each IXEk must be an expression on columns in
+     *       table Tk.  Let C be a column in Tk.
```
paulmartel: This comment seems to have run out of gas mid-thought.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+        // and there is nothing more to do with subqueries.  The processing
+        // of joins is the content of ticket ENG-8677.
+        if (m_tableAliasMap.size() != 1) {
+            return;
+        }
+        // We know there's exactly one.
+        StmtTableScan scan = m_tableAliasMap.values().iterator().next();
+        // Get the table.  There's only one.
+        Table table = getTableFromDB(scan.getTableName());
+        // If we can't find the table, something is probably wrong.
```
paulmartel: Is this null not the simple result of a subquery in the FROM clause?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+        // But don't crash, just return and hope everything sorts
+        // itself out.
+        if (table == null) {
+            return;
+        }
+        // Indices have columns, but the order by list has expressions.
```
paulmartel: Incorrect comment. Indices can have expressions, too.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                TupleValueExpression tve = (TupleValueExpression) expr;
+                Column col = table.getColumns().get(tve.getColumnName());
+                orderByColumns.add(col);
+            }
+        }
+        for (ParsedColInfo colInfo : candidateColumns) {
+            AbstractExpression expr = colInfo.expression;
+            if (expr instanceof TupleValueExpression) {
+                TupleValueExpression tve = (TupleValueExpression) expr;
+                CatalogMap<Constraint> indices = table.getConstraints();
```
paulmartel: Initializing indices inside the loop, why? Wasteful especially if indices is null.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                if (indices == null) {
+                    continue;
+                }
+                // If one of the indices is completely covered
+                // we will not have to process any other indices.
+                // So, we remember this and early-out.
+                boolean addedAllColumns = false;
+                for (Constraint constraint : indices) {
+                    if (addedAllColumns) {
```
paulmartel: would be simpler to break immediately instead of setting the flag to test later.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                        break;
+                    }
+                    Index index = constraint.getIndex();
+                    // Ignore non-unique indices.
+                    if (index.getUnique() == false) {
+                        continue;
+                    }
```
paulmartel: This culling of unusable indexes could also be done up front.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                    CatalogMap<ColumnRef> columns = index.getColumns();
```
paulmartel: A unique expression index does NOT guarantee uniqueness of its INPUT columns which is what you would be  fetching here.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,195 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                    }
+                    if (addAllColumns) {
+                        // Don't forget to remember to forget the other indices.  (E. Presley, 1955)
+                        addedAllColumns = true;
+                        for (Column addCol : table.getColumns()) {
+                            // We have to convert this to a TVE to add
+                            // it to the orderByExprs.  We will use null
+                            // for the table and column aliases and -1
+                            // for the column index.  These will match
+                            // anything.
```
paulmartel: This will work more generally, and better in the future if we use the same table alias that was used in the expressions that got us to this particular table/index.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1494,111 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                Set<AbstractExpression> tveEquivs = valueEquivalence.get(colExpr);
+                if (tveEquivs != null) {
+                    for (AbstractExpression expr : tveEquivs) {
+                        if (expr instanceof ParameterValueExpression
+                                || expr instanceof ConstantValueExpression) {
+                            orderByExprs.add(colExpr);
+                        }
+                    }
+                }
+            }
```
paulmartel: So this handles "select x where x = ?" but not 
"select abs(x) where x = ?"
or more generally "select x * y where x = ? and y = ?"?
(these seem like low-hanging fruit)
not to mention "select abs(x) where abs(x) = ?" -- a little more challenging.
If we are calling any of these out of scope for the current ticket, we'll need new ticket(s).

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
paulmartel: If your intent is not to use expression indexes in this go - around, then there is an easy guard to prevent their use (and getColumns() WILL give the unique columns as you expect).
Create a ticket if use of expression indexes are out of scope of the current ticket (unless covered by the tickets I requested above).

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
paulmartel: Also, want to add temporary negative tests for any cases not covered by the ticket.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
paulmartel: Also, want to add negative tests for any cases not covered by the ticket (because the index we found is an expression index and won't establish determinism for its input columns).

paulmartel:
End of review round.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: Ok.  Apples and apples.

src/frontend/org/voltdb/expressions/TupleValueExpression.java
BillWhite: I tried tablename(aliasname) and found it required too much thinking.  Maybe it's just me, though, and I changed it back to parenthesized names.  I still like the SQL syntax, though, but not enough to argue about it.

src/frontend/org/voltdb/expressions/TupleValueExpression.java
BillWhite: Isn't it?  Isn't it just like C's sprintf?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: Actually, I'm not sure how one would test this.  Maybe a clearly deterministic query would
be made to be non-deterministic.  I think it's faster to just fix it to reflect what is clearly
wanted.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: 1. Yes.  The new tickets are to be thrown off by ENG8677, which is a more extensive analysis.
2. I'll add the negative cases.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: Done.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: This is important.  Let's imagine a table with N integer columns. c1, c2, c3, c4 and c5.  If we have a unique index on columns T.c1 and T.c2, and a two rows, one with values ck=ak and the others with ck=bk, where a1=b1 and a2=b2.  Then it must be that all the ak=bk.  The two rows have identical values.  We don't care what order they appear in, since we can't distinguish between them anyway.

PR #2720, opened 2015-08-19
Eng 8772 - Disk space monitoring and pausing the server

manjujames:
Junit is in pro because the features are enabled in Pro only.


src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+import org.voltdb.compiler.deploymentfile.PathsType;
+import org.voltdb.compiler.deploymentfile.SystemSettingsType;
+import org.voltdb.licensetool.LicenseApi;
+import org.voltdb.utils.CatalogUtil;
+import org.voltdb.utils.MiscUtils;
+
+/**
+ * Disk space monitoring related functionality of resource monitoring.
+ */
+public class DiskResourceChecker
```
akhanzode: Can resource checker share some interface with disk and memory checker?

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+        switch(featureName)
+        {
+        case COMMANDLOG:
+        case COMMANDLOGSNAPSHOT:
+            return licenseApi.isCommandLoggingAllowed();
+        case DROVERFLOW:
+            return licenseApi.isDrReplicationAllowed();
+        case EXPORTOVERFLOW:
+            return MiscUtils.isPro();
+        case SNAPSHOTS:
```
akhanzode: this can be combined in 1st case.

tests/frontend/org/voltdb/TestResourceMonitor.java
```diff
@@ -98,7 +98,7 @@ public void notestNoMonitoringInterval() throws Exception
 
         // Wait for monitoring interval time and verify server is still in running mode
         m_mockStatsProducer.m_rss = 2048L*1024*1024;
-        resumeAndWait(ResourceUsageMonitor.DEFAULT_MONITORING_INTERVAL+1);
+        resumeAndWait(60+1);
```
akhanzode: why change from constant to hard coded value?

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+                configuredFeatures.add(feature.getName());
+                if (!isSupportedFeature(feature.getName())) {
+                    m_logger.warn("Ignoring unsupported feature " + feature.getName());
+                    continue;
+                }
+                String size = feature.getSize();
+                m_configuredLimits.put(feature.getName(),
+                        new FeatureDiskLimitConfig(feature.getName(), pathsConfig, size));
+                if (m_logger.isDebugEnabled()) {
+                    m_logger.debug("Added disk usage limit configuration " + size + " for feature " + feature.getName());
```
akhanzode: why debug? can be useful info if dumped only once.

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+            EnumSet<FeatureNameType> forDefault = EnumSet.complementOf(configuredFeatures);
+            for (FeatureNameType featureName : forDefault)
+            {
+                if (!isSupportedFeature(featureName)) {
+                    continue;
+                }
+                m_configuredLimits.put(featureName,
+                        new FeatureDiskLimitConfig(featureName, pathsConfig, defaultSize));
+                if (m_logger.isDebugEnabled()) {
+                    m_logger.debug("Added disk usage limit configuration " + defaultSize + " for feature " + featureName);
```
akhanzode: same as above

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+            return licenseApi.isCommandLoggingAllowed();
+        default: return false;
+        }
+    }
+
+    private boolean isDiskAvailable(File filePath, int percThreshold, double sizeThreshold)
+    {
+        boolean canWrite = (s_testFileCheck==null) ? filePath.canWrite() : s_testFileCheck.canWrite(filePath);
+        if (!canWrite) {
+            m_logger.rateLimitedLog(60, Level.WARN, null, "Invalid or readonly file path " + filePath);
```
akhanzode: use a constant for interval. Does this need to be rate limited? if rate limiting use a format.

tests/frontend/org/voltdb/TestMemoryResourceMonitor.java
```diff
@@ -0,0 +1,225 @@
...
+import org.voltdb.client.ClientFactory;
+import org.voltdb.client.ClientUtils;
+import org.voltdb.common.Constants;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.utils.FakeStatsProducer;
+import org.voltdb.utils.MiscUtils;
+import org.voltdb.utils.SystemStatsCollector;
+import org.voltdb.utils.SystemStatsCollector.Datum;
+
+public class TestMemoryResourceMonitor extends TestCase
```
akhanzode: It seems its only testing memory limit? Can you add tests for disk limit cases?

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+
+    public boolean isOverLimitConfiguration()
+    {
+        for (FeatureDiskLimitConfig config : m_configuredLimits.values())
+        {
+            if (config.m_diskSizeLimit <= 0 && config.m_diskSizeLimitPerc <= 0) {
+                continue;
+            }
+            if (!isDiskAvailable(config.m_path, config.m_diskSizeLimitPerc, config.m_diskSizeLimit)) {
+                m_logger.warn("Disk is over configred limits for feature " + config.m_featureName);
```
wweiss-voltdb: misspelled configured

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,251 @@
...
+            return MiscUtils.isPro();
+        case SNAPSHOTS:
+            return licenseApi.isCommandLoggingAllowed();
+        default: return false;
+        }
+    }
+
+    private boolean isDiskAvailable(File filePath, int percThreshold, double sizeThreshold)
+    {
+        boolean canWrite = (s_testFileCheck==null) ? filePath.canWrite() : s_testFileCheck.canWrite(filePath);
```
wweiss-voltdb: It might be better to pass in a mocked file rather than having a static test variable that is checked in the main path.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: Yes, that is a good idea. May be something to do if we add more resources to the list that is monitored though? As of now, this is the only one that is called from ResourceChecker (memory one is simple and integrated into the main resource checker class).

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: Thanks.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: This is just adding to an internal structure and might be useful information while debugging. We do show the limits configured separately as info message.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: Same as before.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: Oh yes, thanks.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: That is a really good idea. But, the File is created by the class based on paths, so we probably can't do it. I will look into it.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: Actually, this does not need to be rate limited. Rate limited logging is leftover from when this wasn't returning false. Thanks.

tests/frontend/org/voltdb/TestMemoryResourceMonitor.java
manjujames: Yes, I mentioned it in the note with the pull request. The test has to be Pro because these features are enabled only in pro. There is a separate pull request with the test in pro repo.

tests/frontend/org/voltdb/TestResourceMonitor.java
manjujames: The default is now in the xsd. But, I can still define a constant in this test class, I guess..

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: It does mess up my alphabetic ordering though ;-)

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * Disk space monitoring related functionality of resource monitoring.
+ */
+public class DiskResourceChecker
+{
+    private static final VoltLogger m_logger = new VoltLogger("HOST");
+
+    static FileCheckForTest s_testFileCheck; // used only for testing
+    private final Map<FeatureNameType, FeatureDiskLimitConfig> m_configuredLimits = new HashMap<>();
```
vtkstef: this is a good ImmutableMap candidate. as it is wholly built in `findDiskLimitConfiguration` ImmutableMaps take up less memory, and keep the put order on key or value iterations

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -0,0 +1,249 @@
...
+            if (sizeConfig==null || sizeConfig.trim().isEmpty()) {
+                m_diskSizeLimit = 0;
+                m_diskSizeLimitPerc = 0;
+            } else {
+                String str = sizeConfig.trim();
+                try {
+                    if (str.charAt(str.length()-1) == '%') {
+                        m_diskSizeLimit = 0;
+                        m_diskSizeLimitPerc = Integer.parseInt(str.substring(0, str.length()-1));
+                        if (m_diskSizeLimitPerc > 100 || m_diskSizeLimitPerc < 0) {
```
vtkstef: is 0% a valid request?

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: OK, thanks.

src/frontend/org/voltdb/DiskResourceChecker.java
manjujames: 0 will disable disk limit monitoring. Also, I didn't want to get into what value is a valid request. Is 1% valid, for example..

PR #2721, opened 2015-08-21
Eng 8564 sqlcov adv sc subq slow

johnTcrawford:
Makes the advanced-scalar-subquery test suite (of sqlCoverage) much faster, mainly by reusing the same aggregate function multiple times in the same query, and similarly, using the same comparison operator multiple times in the same query (instead of every possible combination of aggregate functions and comparison operators); with code changes to support this.

### DIFF
--- a/tests/scripts/SQLGenerator.py
+++ b/tests/scripts/SQLGenerator.py
@@ -33,6 +33,7 @@ from voltdbclient import * # for VoltDB types
 from optparse import OptionParser # for use in standalone test mode
 # Need these to print non-ascii characters:
 import codecs
+import sys
 UTF8Writer = codecs.getwriter('utf8')
 sys.stdout = UTF8Writer(sys.stdout)
 
@@ -280,7 +281,7 @@ class BaseGenerator:
     #                   |       |             |
     TYPE_PATTERN_GROUP  =                                           "type" # optional type for columns, values
     #                   |       |             |                      |
-    __EXPR_TEMPLATE = r"%s" r"(\[\s*" r"(#(?P<label>\w+)\s*)?" r"(?P<type>[\w=<>!]+)?\s*" \
+    __EXPR_TEMPLATE = r"%s" r"(\[\s*" r"(#(?P<label>\w+)\s*)?" r"(?P<type>\w+|[=<>!]{1,2})?\s*" \
                       r"(:(?P<min>(-?\d*)),(?P<max>(-?\d*)))?\s*" r"(null(?P<nullpct>(\d*)))?" r"\])?"
     #                         |                |                             |                   |
     #                         |                |                             |       end of [] attribute section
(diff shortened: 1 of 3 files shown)


tests/scripts/SQLGenerator.py
```diff
@@ -482,6 +482,31 @@ def configure_from_schema(self, schema, prior_generators):
...
+       Attributes only matter on the first occurrence of "_text" for a given label.
+       As a convenience, forward references can use the __[#label] syntax instead of _text[#label]
+       to defer locking in attribute settings until a later _text occurrence.
+    """
+
+    def __init__(self):
+        BaseGenerator.__init__(self, "_text")
+
+    def prepare_params(self, attribute_groups):
+        self.__supertype = attribute_groups[BaseGenerator.TYPE_PATTERN_GROUP]
```
paulmartel: The term "TYPE" in TYPE_PATTERN_GROUP seems to have been stretched beyond any real meaning in this particular case, but I don't see an alternative that wouldn't muddy the intent of the other use cases. It's more of a generic symbol in this use case than it is a type -- and maybe this class would be more correctly named  SymbolGenerator than TextGenerator since (AFAICT) its alphabet is limited to "word-and-comparison" characters -- arguably, you could tighten up the regexp so that it only allowed words (sequence of word characters) OR comparisons (sequence of operator characters) but not a mix of the two -- and still cover your current use cases.

The use of the name __supertype throughout this class seems totally unwarranted for this case -- I would probably replace it with __symbol. I THINK that the value of attribute_groups[BaseGenerator.TYPE_PATTERN_GROUP] has already been macro-resolved to some kind of single "symbol" value. Right? I have to admit to being surprised by the elegance of this solution, making me wonder how well I really remember this code. Nice work.

tests/scripts/examples/sql_coverage/grammar.sql
```diff
@@ -14,8 +14,11 @@
 --{_stringfun |= "LOWER"}
 --{_stringfun |= "UPPER"}
 
-{_genericagg |= "MIN"}
-{_genericagg |= "MAX"}
+-- Aggregate functions that accept a string (or numeric) column and return the same type
+{_stringagg |= "MIN"}
+{_stringagg |= "MAX"}
```
paulmartel: Nice refinement.

tests/scripts/examples/sql_coverage/advanced-scalar-subquery.sql
```diff
@@ -58,30 +58,30 @@ SELECT @idcol, (SELECT @agg(_variable[#agg]) FROM @fromtables WHERE A2.__[#agg]
 SELECT @idcol, (SELECT @agg(_variable[#agg]) FROM @fromtables WHERE A3._variable[@comparabletype] _cmp2 _variable[@comparabletype]) FROM @fromtables    A3 _optionalorderbyidlimitoffset
 
 --- Queries with scalar subqueries in the WHERE clause (with optional ORDER BY, LIMIT, OFFSET, GROUP BY or HAVING clauses)
-SELECT _variable[#ord]         FROM @fromtables A11 WHERE __[#ord] _cmp1 (SELECT @agg(       __[#ord]) FROM @fromtables                                      ) _grouporderbyvarlimoffhaving
-SELECT _variable[#ord numeric] FROM @fromtables A12 WHERE __[#ord] _cmp2 (SELECT @agg(_variable[#agg]) FROM @fromtables                                      ) _grouporderbyvarlimoffhaving
+SELECT _variable[#ord]         FROM @fromtables A11 WHERE __[#ord] _text[#cmp _cmp] (SELECT _text[#agfcn @agg](       __[#ord]) FROM @fromtables                                         )    _grouporderbyvarlimoffhaving
```
paulmartel: The new scheme seems to have superceded my previous attempts at a similar optimization using the lower-cardinality generator macros _cmp1, _cmp2, _cmp3, and _cmp13. At  this point, AT LEAST _cmp1 and _cmp3 look like overkill -- used only to define _cmp13. ANd _cmp2 and _cmp13 are only being used once each in statements A2 and A3. I don't know how prodigious A2 and A3 currently are -- can we afford to let them use the more general _cmp and eliminate these refinements? If that would cost too much, then I'd just replace _cmp13 with _cmp1 in statement A2, expand the _cmp1 definition to include the values currently in _cmp3 and drop the definitions of _cmp3 and _cmp13.

tests/scripts/examples/sql_coverage/advanced-scalar-subquery.sql
```diff
@@ -58,30 +58,30 @@ SELECT @idcol, (SELECT @agg(_variable[#agg]) FROM @fromtables WHERE A2.__[#agg]
...
-SELECT (SELECT _genericagg(_variable[#agg string])   FROM @fromtables WHERE _variable[@comparabletype] _cmp3 A33._variable[@comparabletype]) C0, @agg(__[#agg]) C1 FROM @fromtables A33 GROUP BY C0 HAVING @agg(__[#agg]) _cmp 'Z'
+SELECT (SELECT _text[#agfcn @agg](_variable[#agg])                 FROM @fromtables WHERE _variable[@comparabletype]            _cmp  A31._variable[@comparabletype]) C0, __[#agfcn](__[#agg]) C1 FROM @fromtables A31 GROUP BY C0
+SELECT (SELECT _text[#agfcn @agg](_variable[#agg @comparabletype]) FROM @fromtables WHERE _variable[@comparabletype] _text[#cmp _cmp] A32._variable[@comparabletype]) C0, __[#agfcn](__[#agg]) C1 FROM @fromtables A32 GROUP BY C0 HAVING __[#agfcn](__[#agg]) __[#cmp] 12
+SELECT (SELECT _text[#agfcn _stringagg](_variable[#agg string])    FROM @fromtables WHERE _variable[@comparabletype] _text[#cmp _cmp] A33._variable[@comparabletype]) C0, __[#agfcn](__[#agg]) C1 FROM @fromtables A33 GROUP BY C0 HAVING __[#agfcn](__[#agg]) __[#cmp] 'Z'
 --- These do not currently work (meanwhile, commented out to save execution time)
```
paulmartel: I just happened to notice this comment -- not sure whether it is referring to  ENG-8239 as sited above or if there is some other reason "these do not work". Is it worth doing a quick run with them enabled to get a clarification of what their specific issue may be?

paulmartel:
End of review round.

PR #2722, opened 2015-08-21
ENG-8870: Combine Partitioned and Replicated DR Stream guard to fix s…

wweiss-voltdb:
…coping issues

### DIFF
--- a/src/ee/storage/DRTupleStream.h
+++ b/src/ee/storage/DRTupleStream.h
@@ -129,15 +129,43 @@ public:
 
 class DRTupleStreamDisableGuard {
 public:
-    DRTupleStreamDisableGuard(DRTupleStream *stream) : m_stream(stream), m_oldValue(stream->m_enabled) {
-        stream->m_enabled = false;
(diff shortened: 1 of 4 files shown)


src/ee/storage/DRTupleStream.h
```diff
@@ -129,15 +129,43 @@ class MockDRTupleStream : public DRTupleStream {
 
 class DRTupleStreamDisableGuard {
 public:
-    DRTupleStreamDisableGuard(DRTupleStream *stream) : m_stream(stream), m_oldValue(stream->m_enabled) {
-        stream->m_enabled = false;
+    DRTupleStreamDisableGuard(DRTupleStream *drStream, DRTupleStream *drReplicatedStream, bool ignore) :
+            m_drStream(drStream), m_drReplicatedStream(drReplicatedStream),
+            m_drStreamOldValue(drStream->m_enabled), m_drReplciatedStreamOldValue(false)
```
nshi: Typo in `m_drReplciatedStreamOldValue`.

It can be initialized here with the correct value and you can declare it as a const member.

PR #2728, opened 2015-08-24
Eng 8884

pdshaw:
Use passed dist.version string in copy operations instead of wildcard;

~test/libs moved to ~test/nativelibs


tools/copy-lib.sh
```diff
@@ -15,19 +15,18 @@ echo '$1 is' $1
 echo '$2 is' $2
```
akhanzode: put something concrete $2 is $2 not much helpful echo "Version is : $2" or something like that. Easy for someone else to understand.

akhanzode:
Looks good to me. squash merge if jenkins is happy and manually tested.

PR #2729, opened 2015-08-24
ENG-7968: Tune DR buffer sizes to mitigate impact on consumer cluster

afrogers:


tests/ee/storage/DRTupleStream_test.cpp
```diff
@@ -694,6 +694,65 @@ TEST_F(DRTupleStreamTest, BigBufferAfterExtendOnBeginTxn) {
...
+    boost::shared_ptr<StreamBlock> results = m_topend.blocks.front();
+    EXPECT_EQ(results->uso(), 0);
+    EXPECT_EQ(results->offset(), MAGIC_TUPLE_PLUS_TRANSACTION_SIZE);
+
+    m_topend.blocks.pop_front();
+    m_topend.receivedDRBuffer = false;
+
+    appendTuple(2, 3);
+    m_wrapper.endTransaction();
+
```
wweiss-voltdb: Shouldn't you inject 2 transactions before the flush to verify that they row limit is being enforced?

afrogers: Well, this test just verifies that it's not over-enforced. The above test makes sure that it goes in the list of completed blocks when we append too many tuples, thus testing row limit enforcement

PR #2731, opened 2015-08-24
Adds txnid2 testing for missing cmdlogs after a recover.

liam700:
This feature can be configured by disabling all threads besides the client threads by copying the list of threads in the async-benchmark function of the run.sh script, pasting it to the disabled threads parameter, and deleting the "clients" item from the list.


liam700:
Decided to expand statistics slightly

PR #2734, opened 2015-08-24
ENG-8724, serialize and deserialize hidden column to/from DR binary log.

luyangco:
Review only. The code breaks a test case in DRBinaryLog_test, as looking up tuple contains hidden column always fails now. This will be addressed in another ticket.

### DIFF
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -320,6 +320,9 @@ public:
     void serializeTo(voltdb::SerializeOutput &output);
     void serializeToExport(voltdb::ExportSerializeOutput &io,
                           int colOffset, uint8_t *nullArray);
+    void serializeHiddenColumnsToDR(ExportSerializeOutput &io);
+    void serializeAllColumnsToDR(ExportSerializeOutput &io,
+                          int colOffset, uint8_t *nullArray);
     void serializeToDR(voltdb::ExportSerializeOutput &io,
                        int colOffset, uint8_t *nullArray,
                        const std::vector<int>* interestingColumns);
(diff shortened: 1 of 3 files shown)


src/ee/common/tabletuple.h
```diff
@@ -320,6 +320,9 @@ class TableTuple {
     void serializeTo(voltdb::SerializeOutput &output);
     void serializeToExport(voltdb::ExportSerializeOutput &io,
                           int colOffset, uint8_t *nullArray);
+    void serializeHiddenColumnsToDR(ExportSerializeOutput &io);
```
afrogers: This probably doesn't need to be part of the public interface

src/ee/common/tabletuple.h
```diff
@@ -794,6 +797,16 @@ inline void TableTuple::deserializeFromDR(voltdb::SerializeInputLE &tupleIn,  Po
     int nullMaskLength = ((columnCount + 7) & -8) >> 3;
     const uint8_t *nullArray = reinterpret_cast<const uint8_t*>(tupleIn.getRawPointer(nullMaskLength));
 
+    const int32_t hiddenColumnCount = m_schema->hiddenColumnCount();
+    for (int i = 0; i < hiddenColumnCount; i++) {
+        const TupleSchema::ColumnInfo * hiddenColumnInfo = m_schema->getHiddenColumnInfo(i);
```
afrogers: This appears to be an exact duplicate of the code below for deserializing an ordinary column, except now we're calling `getHiddenColumnInfo()`. Would be nice to factor this out into an inlined method.

afrogers: Actually, maybe not. You're call; code duplication here isn't that bad.

afrogers:
Review complete

src/ee/common/tabletuple.h
luyangco: Yes you're right, some level of duplicity exists here. But since we still need to differentiate deserializing normal columns or hidden columns, factor the code into a new method won't reduce the duplicity.

PR #2735, opened 2015-08-24
ENG-8781, write DR conflict to export table.

luyangco:
Write a stub function to export DR conflict, when DR conflict detection is done, need to rewrite the test to make it more functional.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -103,6 +103,8 @@ ENABLE_BOOST_FOREACH_ON_CONST_MAP(Table);
 static const size_t PLAN_CACHE_SIZE = 1000;
 // how many initial tuples to scan before calling into java
 const int64_t LONG_OP_THRESHOLD = 10000;
+// table name prefix of DR conflict table
+const std::string DR_CONFLICT_TABLE_PREFIX = "VOLTDB_AUTOGEN_DR_CONFLICTS__";
 
 namespace voltdb {
 
@@ -472,10 +474,19 @@ Table* VoltDBEngine::getTable(std::string name) const
     return findInMapOrNull(name, m_tablesByName);
 }
 
-Table* VoltDBEngine::getDRConflictTable(PersistentTable* drTable) const
+Table* VoltDBEngine::getDRConflictTable(PersistentTable* drTable)
 {
-    // Caller responsible for checking null return value.
-    return findInMapOrNull(drTable, m_cachedDRConflictLookupTable);
+    Table* exportTable;
+    boost::unordered_map<PersistentTable*, Table*>::iterator it = m_cachedDRConflictLookupTable.find(drTable);
+    if (it == m_cachedDRConflictLookupTable.end()) {
+        exportTable = getTable(DR_CONFLICT_TABLE_PREFIX + drTable->name());  // cache table miss, back to full search
+        if (exportTable) {
+            m_cachedDRConflictLookupTable[drTable] = exportTable;
+        }
+    } else {
+        exportTable = it->second;
+    }
+    return exportTable;
 }
 
 TableCatalogDelegate* VoltDBEngine::getTableDelegate(std::string name) const
@@ -1277,10 +1288,6 @@ void VoltDBEngine::rebuildTableCollections()
             }
         }
     }
-}
-
-void VoltDBEngine::addToDRConflictTableMap(PersistentTable* drTable, Table* exportTable) {
-    m_cachedDRConflictLookupTable[drTable] = exportTable;
 }
 
 void VoltDBEngine::setExecutorVectorForFragmentId(int64_t fragId)
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -472,6 +472,12 @@ Table* VoltDBEngine::getTable(std::string name) const
     return findInMapOrNull(name, m_tablesByName);
 }
 
+Table* VoltDBEngine::getDRConflictTable(PersistentTable* drTable) const
+{
+    // Caller responsible for checking null return value.
+    return findInMapOrNull(drTable, m_cachedDRConflictLookupTable);
+}
+
 TableCatalogDelegate* VoltDBEngine::getTableDelegate(std::string name) const
 {
     // Caller responsible for checking null return value.
@@ -1241,6 +1247,7 @@ void VoltDBEngine::rebuildTableCollections()
     m_tables.clear();
     m_tablesByName.clear();
     m_tablesBySignatureHash.clear();
+    m_cachedDRConflictLookupTable.clear();
 
     // need to re-map all the table ids / indexes
     getStatsManager().unregisterStatsSource(STATISTICS_SELECTOR_TYPE_TABLE);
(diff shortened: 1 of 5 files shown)


src/ee/storage/BinaryLogSink.cpp
```diff
@@ -63,6 +68,14 @@ class CachedIndexKeyTuple {
 
 BinaryLogSink::BinaryLogSink() {}
 
+int8_t BinaryLogSink::getClusterIdFromDRId(int64_t drId) {
```
afrogers: More properly, clusterId should be `uint8_t`, which is why we treat it as an `int` in Java.

src/ee/execution/VoltDBEngine.h
```diff
@@ -499,6 +502,12 @@ class __attribute__((visibility("default"))) VoltDBEngine {
          */
         boost::unordered_map<int64_t, PersistentTable*> m_tablesBySignatureHash;
 
+        /*
+         * Map of DR tables to DR conflict tables
+         */
+        //boost::unordered_map<Table*, Table*> m_cachedDRConflictLookupTable;
```
afrogers: Remove commented code; although, actually, why are we paying the overhead of storing these in a balanced tree rather than an `unordered_map`?

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -148,12 +151,31 @@ class DRBinaryLogTest : public Test {
                                                                                                           singleColumnName,
                                                                                                           tableHandle + 1, false, 0));
         m_singleColumnTable->setDR(true);
+
+        std::vector<ValueType> exportColumnType;
+        std::vector<int32_t> exportColumnLength;
+        std::vector<bool> exportColumnAllowNull(6, false);
+        exportColumnType.push_back(VALUE_TYPE_VARCHAR);     exportColumnLength.push_back(std::strlen("OTHER_TABLE_2"));
+        exportColumnType.push_back(VALUE_TYPE_TINYINT);     exportColumnLength.push_back(NValue::getTupleStorageSize(VALUE_TYPE_TINYINT));
```
afrogers: I know such comments exist in many places, but would be nice if these are accompanied by descriptive comments.

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -403,6 +431,28 @@ TEST_F(DRBinaryLogTest, PartitionedTableNoRollbacks) {
...
+    // single row write transaction
+    TableTuple otherTempTuple = m_otherTableWithoutIndex->tempTuple();
+    otherTempTuple.setNValue(0, ValueFactory::getTinyIntValue(123));
+    otherTempTuple.setNValue(1, ValueFactory::getBigIntValue(9999999999));
+
+    m_sink.exportDRConflict(m_otherTableWithoutIndex, m_exportTable, DR_RECORD_INSERT, tempTuple);
+    m_sink.exportDRConflict(m_otherTableWithoutIndex, m_exportTable, DR_RECORD_INSERT, otherTempTuple);
+
+    // Really have no idea what to test for now.
+    EXPECT_EQ(2, m_exportTable->activeTupleCount());
```
afrogers: I suppose you could manually construct the expected tuple and try to look it up by values, although that's a little pointless.

afrogers: On second thought, not entirely pointless; I'd do just that -- make sure you get tuples w/ table name, etc., and the right tuple values

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -200,6 +213,38 @@ void BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_t,
     }
 }
 
+void BinaryLogSink::exportDRConflict(PersistentTable *drTable, Table *exportTable, const DRRecordType &type, TableTuple &exportTuple) {
+//    if (!engine) {
```
afrogers: Remove commented code, unless the point is that this will go back in once we actually detect conflicts.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -200,6 +213,38 @@ void BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_t,
...
+        int64_t drId = ValuePeeker::peekAsBigInt(hiddenColumn);
+
+        tempTuple.setNValue(0, ValueFactory::getStringValue(drTable->name()));  // Table Name
+        tempTuple.setNValue(1, ValueFactory::getTinyIntValue(getClusterIdFromDRId(drId)));       // Cluster Id
+        tempTuple.setNValue(2, ValueFactory::getBigIntValue(getSequenceNumberFromDRId(drId)));   // Timestamp
+        tempTuple.setNValue(3, ValueFactory::getTinyIntValue(type));            // Type of Operation
+        tempTuple.setNValues(4, exportTuple, 0, exportTuple.sizeInValues());    // rest of columns
+
+        /*************************Debug************************/
+        std::ostringstream os;
```
afrogers: This looks like this isn't meant to be committed uncommented/unguarded?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -472,6 +472,12 @@ Table* VoltDBEngine::getTable(std::string name) const
     return findInMapOrNull(name, m_tablesByName);
 }
 
+Table* VoltDBEngine::getDRConflictTable(PersistentTable* drTable) const
+{
+    // Caller responsible for checking null return value.
+    return findInMapOrNull(drTable, m_cachedDRConflictLookupTable);
```
afrogers: Given that this is just a caching lookup, would probably be nicer if that lookup here did the logic to add this to the cache itself instead of relying on the caller.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -200,6 +213,38 @@ void BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_t,
...
+        os << tempTuple.getSchema()->debug();
+        LogManager::getThreadLogger(LOGGERID_HOST)->log(LOGLEVEL_ERROR, os.str().c_str());
```
afrogers: Or if this was meant to be committed, it probably shouldn't be logging itself as ERROR

afrogers:
Review complete

src/ee/execution/VoltDBEngine.cpp
luyangco: Nice suggestion, it makes code clearer to read.

src/ee/execution/VoltDBEngine.h
luyangco: A performance vs memory issue. I believe the number of conflict export tables won't be large, so probably the memory is not a issue. Plus there is no need to keep the order here, I will change to unordered_map.

tests/ee/storage/DRBinaryLog_test.cpp
luyangco: The problem is that exportTable is not a PersistentTable, I don't find a good way to retrieve tuple from a StreamedTable.

src/ee/storage/BinaryLogSink.cpp
luyangco: The valid input range for clusterId is from 0 to 127, so It should be both fine for signed and unsigned int8_t. And we write clusterId into a TINYINT column, which is also a int8_t type.

PR #2740, opened 2015-08-25
Eng 8645 fewer NDOs

BillWhite:
Jenkins is mostly blue with this.  The failures, I think, are unrelated to these changes.


wolffcm:
Would it make sense for me to code review this?  I know that Paul had reviewed the last batch of changes.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1385,19 +1387,44 @@ protected boolean orderByColumnsCoverUniqueKeys()
         // the result is deterministic.
         // This holds regardless of whether the associated index is actually used in the selected plan,
         // so this check is plan-independent.
+        //
+        // baseTableAliases associates table aliases with the order by
+        // expressions which reference them.  Presumably by using
+        // table aliases we will map table scans to expressions rather
+        // than tables to expressions, and not confuse ourselves with
+        // different instances of the same table in self joins.
```
wolffcm: Thanks for adding this comment.  This is my understanding as well.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1385,19 +1387,44 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
-            List<AbstractExpression> baseTVEs = expr.findBaseTVEs();
-            if (baseTVEs.size() != 1) {
+            //
+            // Compute the set of tables mentioned in the expression.
+            //   1. Search out all the TVEs.
+            //   2. Throw the aliases of the tables of each of these into a HashSet.
+            //      The table must have an alias.  It might not have a name.
+            //   3. If the HashSet has size > 1 we can't use this expression.
+            //
+            List<AbstractExpression> baseTVEExpressions = expr.findBaseTVEs();
```
wolffcm: So I guess that a statement like
```SQL
SELECT id FROM t ORDER BY abs(id);
```
would consider the primary key `id` to be covered?  Wouldn't it still be possible to have non-deterministic output if `id` has both positive and negative values?  Seems like we need to be a bit more selective when deciding whether to add the alias.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1513,116 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+                        }
+                    }
+                }
+            }
+        }
+        // If there is not exactly one table scan we will not proceed.
+        // We don't really know how to make indices work with joins,
+        // and there is nothing more to do with subqueries.  The processing
+        // of joins is the content of ticket ENG-8677.
+        if (m_tableAliasMap.size() != 1) {
```
wolffcm: You could move this check to the beginning of this method to avoid unnecessary work.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -30,7 +30,12 @@
 
 import org.hsqldb_voltpatches.VoltXMLElement;
 import org.voltdb.VoltType;
+import org.voltdb.catalog.CatalogMap;
```
wolffcm: Many of these imports could probably go away now, since you moved most of the code over to AbstractParsedStmt.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
wolffcm: If my comment is correct then it was broke when you got there, I know.  But if it's easy to fix, consider fixing it.

wolffcm:
At some point it would be good to make sure this enhanced determinism detection is also used for DELETE statements.  It's especially important for DML because we'll actually fail to compile instead of just warn if the planner thinks there's non-determinism.  Can you file a ticket for this?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1481,6 +1513,116 @@ protected boolean orderByColumnsCoverUniqueKeys()
...
+        for (ParsedColInfo colInfo : candidateColumns) {
+            AbstractExpression expr = colInfo.expression;
+            if (expr instanceof TupleValueExpression) {
+                TupleValueExpression tve = (TupleValueExpression) expr;
+                // If one of the indices is completely covered
+                // we will not have to process any other indices.
+                // So, we remember this and early-out.
+                for (Index index : indices) {
+                    CatalogMap<ColumnRef> columns = index.getColumns();
+                    // If all the columns in this index are in the current
```
wolffcm: If all the columns in a unique index are already in the order by list, doesn't that mean the plan is deterministic?  Why do you need to add all the columns in the table to the order by list?  I feel like I'm missing something here.

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -465,6 +465,26 @@ public void testMPDeterminismOfSelectOneKeyValue() {
         assertMPPlanDeterminism("select a, z from ppk where a = 1 and b < 10", "order by z, a");
     }
 
+    public void testMPDeterminismImpliedByParameter() {
+        assertPlanDeterminismCore("select * from ttree_with_key where b = ? order by a, c limit 1;", true, true, DeterminismMode.FASTER);
+        assertPlanDeterminismCore("select * from ttree_with_key where abs(b) = ? order by a, c limit 1;", false, false, DeterminismMode.FASTER);
+        // If ENG-8677 and follow on tickets are closed,
+        // we can reenable this test.
+        final boolean ENG8677IsFixed = false;
```
wolffcm: Is there enough testing here?  Maybe a test case where detecting determinism depends on adding all the columns in the table to the honorary order by list?

wolffcm:
That's all my comments.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: I didn't explain my plan.

I'm not trying to fix everything with this commit.  I'm trying to fix the particular problem in the ticket.

Everything will take another pass, with a better understanding of what an order deterministic query would be.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: Moved.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: This may be true.  I hadn't thought about this.

If this was for a multiple table query this would not be true, I don't think.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: Done.  Eclipse is very good at keeping the import list minimal.  We could do this
automatically on save if we didn't have to retain HSQLDB's import list.

tests/frontend/org/voltdb/planner/TestDeterminism.java
BillWhite: Fair enough.  I'm added some more tests.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
wolffcm: Yeah, I can see it being useful for the multi-table case.

BillWhite: Also, even in the single table case, the honorary order by list has been expanded at this point.  The plan won't know about the expanded list, I think, and it may actually be planned when this is called.  So, the planner can't tell if the query is deterministic or not.

The bug in the ticket is an example of this.  There is a primary key on columns a, b and c, the order by list has a and c, and b is fixed to be a constant.  We select *.  In this case adding b to the order by list will not change the ordering, so we could add it.

BillWhite: I filed a ticket for the delete case.

BillWhite:
Squash merged.

PR #2741, opened 2015-08-25
Remove -Xcheck:jni from junit JVM options

wolffcm:
It slows down execution too much on Java 8.  There are also lots of
JNI warnings that are issued when we run with Java 8.  I'll file a
separate ticket to address them.


build.xml
```diff
@@ -1690,7 +1690,6 @@ TEST CASES
         >
             <classpath refid='project.classpath' />
             <jvmarg value="-server" />
-            <jvmarg value="-Xcheck:jni" />
```
jxchinaren: Is it reasonable to have a comment here and comment out the -Xcheck:jni flag other than delete it completely? I wonder someone in future will add it back without any notice.

This is my ONLY comment.

wolffcm: Sure, I'll add a comment explaining why.

jxchinaren:
Looks good to go.

akhanzode:
Do we really want to do this? If yes I would still create a target that does check:jni and let jenkins do that in addition.

wolffcm:
@akhanzode I filed ENG-8898 for running JUnit tests in Jenkins with check:jni turned on for selected jobs.  I don't think we'll want it turned on normally if we ever switch to Java 8 by default---the test I looked at took five times longer under Java 8 with check:jni turned on.

https://issues.voltdb.com/browse/ENG-8898

Does this ticket seem sufficient to you?  If you still have concerns, let's figure what would work best.

akhanzode:
That ticket works for me. I hope it gets done and doesn't get prioritized down.

PR #2742, opened 2015-08-25
Eng 8896 Add all possible VDB nullable and not nullable data types.

pdshaw:


tests/test_apps/kafkaimporter/run.sh
```diff
@@ -138,8 +142,9 @@ function async-benchmark() {
     java -classpath $CLIENTCLASSPATH \
         kafkaimporter.client.kafkaimporter.KafkaImportBenchmark \
         --displayinterval=5 \
-        --ratelimit=20000 \
-        --duration=60 \
+        --ratelimit=2000 \
+        --duration=20 \
```
prosegay: did you really mean to change the default ratelimit and duration values?

tests/test_apps/kafkaimporter/run.sh
```diff
@@ -138,8 +142,9 @@ function async-benchmark() {
...
+        --allvalues=false \
```
prosegay: "allvalues" is not very discriptive, maybe "usealltypes"???

tests/test_apps/kafkaimporter/deployment.xml
```diff
@@ -8,7 +8,8 @@
     <import>
         <configuration module="kafkastream.jar" type="custom" enabled="true">
             <property name="brokers">kafka2:9092</property>
-            <property name="topics">T1_KAFKAEXPORTTABLE1</property>
+            <property name="topics">A7_KAFKAEXPORTTABLE1</property>
+            <property name="groupid">A7</property>
             <property name="procedure">InsertImport</property>
```
prosegay: we talked about this: use the same export topic prefix, and use two different import tables pointing to the different import procedures, see if that is possible.

tests/test_apps/kafkaimporter/db/procedures/InsertImport2.java
```diff
@@ -0,0 +1,253 @@
...
+            voltExecuteSQL(true);
+        } else {
+            voltQueueSQL(insertCounts, key, deletedCount);
+            voltExecuteSQL(true);
+        }
+        return 0;
+    }
+
+    private boolean reportMismatch(String typeName, String mirrorVal, String importVal) {
+        System.out.println("Mirror " + typeName + "not equal to import " + typeName + ":" + mirrorVal + " != " + importVal);
```
prosegay: use volt abort exception if there is a mismatch and cause the testapp to fail with a non-zero return code so that apprunner can pick it up and fail.

tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
```diff
@@ -202,12 +209,13 @@ public synchronized void printStatistics() {
             log.info(String.format("Export Throughput %d/s, Total Rows %d, Aborts/Failures %d/%d, Avg/95%% Latency %.2f/%.2fms",
                     thrup, rows, stats.getInvocationAborts(), stats.getInvocationErrors(),
                     stats.getAverageLatency(), stats.kPercentileLatencyAsDouble(0.95)));
-        } catch (Exception e) {
-            log.error("Exception in printStatistics", e);
+        } catch (Exception ex) {
+            log.error("Exception in printStatistics", ex);
+            log.error(ex.getStackTrace());
         }
```
prosegay: don't need the stack trace if you use log4j second parameter

tests/test_apps/kafkaimporter/pounder.groovy
```diff
@@ -0,0 +1,88 @@
...
+        cl.await()
+
+        int id = rlimit.getAndIncrement()
+        while ( id < rows ) {
+            awaiter << kp.send(rec(topic, id,rnd.nextInt(Integer.MAX_VALUE)))
+            id = rlimit.getAndIncrement()
+        }
+        awaiter.flush()
+    }
+}.each { it.join() }
```
prosegay: my understanding is that pounder is not used in this app if that is correct pls remove this file

tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
```diff
@@ -266,8 +272,9 @@ public void runBenchmark() throws Exception {
             }
             // check for export completion
             //exportMon.waitForStreamedAllocatedMemoryZero();
-        } catch (Exception e) {
-            log.error("Exception in Benchmark", e);
+        } catch (Exception ex) {
+            log.error("Exception in Benchmark", ex);
+            log.error(ex.getStackTrace());
```
vtkstef: the second one is superfluous as the first one will print the stack trace

tests/test_apps/kafkaimporter/deployment.xml
prosegay: In any case, the supplied depoloyment file should work with all program options without change, it seems that this one does not.

tests/test_apps/kafkaimporter/run.sh
pdshaw: Was convenient for development but I'm undo those changes for this request.

tests/test_apps/kafkaimporter/deployment.xml
pdshaw: Added 2 topic/import sections. Works good!

pdshaw: Btw, 2 topics makes the local test and execution via run.sh more flexible, along with covering new functionality.

But Kafka setup via apprunner still needs to align the topic creation. Should be a easy change though. One quick adaptation is to always create both topics.

PR #2743, opened 2015-08-25
Eng 8824 Speed up java restore times

jhugg:


src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,22 +727,336 @@ public final VoltTableRow fetchRow(int index) {
...
+                case DECIMAL:
+                    VoltDecimalHelper.serializeNull(m_buffer);
+                    break;
+
+                default:
+                    throw new VoltTypeException("Unsupported type: " +
+                            columnType);
+                }
+            } else {
+
```
paulmartel: For clarity mostly, the code above this line could run before/outside the try block.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,22 +727,336 @@ public final VoltTableRow fetchRow(int index) {
...
+    /**
      * Append a {@link VoltTableRow row} from another <tt>VoltTable</tt>
      * to this VoltTable instance. Technically, it could be from the same
      * table, but this isn't the common usage.
      * @param row {@link VoltTableRow Row} to add.
      */
     public final void add(VoltTableRow row) {
+        if (m_readOnly) {
+            throw new IllegalStateException("Table is read-only. Make a copy before changing.");
+        }
```
paulmartel: A bit unconventional for "assert(verifyTableInvariants());" not to be the very first line in a method.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,22 +727,336 @@ public final VoltTableRow fetchRow(int index) {
...
+                    }
+                    else {
+                        Object inboundValue = row.get(i, inboundType);
+                        addColumnValue(inboundValue, outboundType, i);
+                    }
+                }
+            }
+
+            //
+            // Note, there is some near-identical code in both row add methods.
```
paulmartel: Clarify: "both row add methods" -> "the add and addRow methods" -- I was searching needlessly for an overload of add.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,22 +727,336 @@ public final VoltTableRow fetchRow(int index) {
...
+
+            m_rowCount++;
+            m_buffer.putInt(m_rowStart, m_rowCount);
+            final int rowsize = m_buffer.position() - pos - 4;
+
+            // check for too big rows
+            if (rowsize > VoltTableRow.MAX_TUPLE_LENGTH) {
+                throw new VoltOverflowException(
+                        "Table row total length larger than allowed max " + VoltTableRow.MAX_TUPLE_LENGTH_STR);
+            }
```
paulmartel: Seems like we could assume a valid row size in the canDoRawCopy case and push the row size validation into the !canDoRawCopy case -- that leaves 3 lines of code between the close of the "else" above and the retesting of the !canDoRawCopy else condition -- consider copying the three lines into the if and else blocks and avoiding the canDoRawCopy variable/test.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,22 +727,336 @@ public final VoltTableRow fetchRow(int index) {
...
+            // (It would be nice to re-factor, but I couldn't make a clean go at
+            //  it quickly - Hugg)
+            //
+
+            m_rowCount++;
+            m_buffer.putInt(m_rowStart, m_rowCount);
+            final int rowsize = m_buffer.position() - pos - 4;
+
+            // check for too big rows
+            if (rowsize > VoltTableRow.MAX_TUPLE_LENGTH) {
```
paulmartel: Would it make any sense to limit m_buffer to its capacity OR an additional VoltTableRow.MAX_TUPLE_LENGTH, whichever comes first, so we can test for the too-long tuple only as a special case in the BufferOverflowException catch block? Seems pointless to potentially expand a buffer to accommodate a too-long tuple that will fail in the end.
Also, this makes it cheap to check for too-long tuple as we go.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,22 +727,336 @@ public final VoltTableRow fetchRow(int index) {
...
+            // buffer additions
+            m_buffer.position(pos);
+            throw vte;
+        }
+        catch (BufferOverflowException e) {
+            m_buffer.position(pos);
+            expandBuffer();
+            add(row);
+        }
+        catch (IllegalArgumentException e) {
```
paulmartel: Somebody seems to know exactly where and likely why IllegalArgumentException is getting thrown and yet it's not easily preventable (to avoid throwing in the common code path)?

At the least, the comments should let us all in on the details of the hack.

src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -373,6 +386,73 @@ public final boolean wasNull() {
...
+            // all of these types are fixed length, so easy to get raw type
+            retval = new byte[length];
+            pos = m_buffer.position();
+            m_buffer.position(offset);
+            m_buffer.get(retval);
+            m_buffer.position(pos);
+            return retval;
+        case STRING:
+        case VARBINARY:
+            // all of these types are variable lenght with a prefix
```
paulmartel: spelling lenght

src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -373,6 +386,73 @@ public final boolean wasNull() {
...
+            length = m_buffer.getInt(offset);
+            if (length == VoltTable.NULL_STRING_INDICATOR) {
+                length = 0;
+            }
+            length += 4;
+            retval = new byte[length];
+            pos = m_buffer.position();
```
paulmartel: Are we caching the pos too late, here, after getInt() so off by +4.
Wouldn't it be safer to init pos at the top of the function and always restore just before exit?
OTOH, should we really be bothering to restore pos after every column fetch?
What's the usefulness of leaving pos at the start of the already fetched column?

src/frontend/org/voltdb/VoltType.java
```diff
@@ -720,51 +720,39 @@ public static boolean isNullVoltType(Object obj)
...
         else if (obj == VoltType.NULL_TIMESTAMP ||
-                obj == VoltType.NULL_STRING_OR_VARBINARY ||
-                obj == VoltType.NULL_DECIMAL)
+                 obj == VoltType.NULL_STRING_OR_VARBINARY ||
+                 obj == VoltType.NULL_DECIMAL)
         {
-            retval = true;
+            return true;
         }
         else
```
paulmartel: Redundant elses and indents after returns.

tests/frontend/org/voltdb/TestVoltTable.java
```diff
@@ -998,4 +999,130 @@ public void test300Columns() {
...
+                while (pt.advanceRow()) {
+                    t2.add(pt);
+                }
+            }
+
+            assertTrue(t2.equals(t));
+        }
+    }
+
+    public void testSchemaChangeAddRow() {
```
paulmartel: Could use overflow coverage, like test that the data initialized with the narrow schema and then copied to the wider schema can get copied back to another narrow table without triggering overflow.
Test that corrupting a wide row by inserting a large bigint value thwarts adding it back to the narrow table.

Also, could use null value coverage -- does your data randomizer do nulls?

paulmartel: Are we vouching for the health of a VoltTable after a failed add -- that is, do we guarantee a clean rollback so you can catch the exception from the add and continue?
If so, we should have before and after comparisons of a table that has seen some failed adds.

paulmartel: BTW, null coverage should exercise the no-nulls failure case supported by that fancy metadata option.

Should row/column blit be forbidden when the source table/column is the same but a little more null-friendly than the target?

paulmartel:
End of review round.

src/frontend/org/voltdb/VoltType.java
jhugg: Fixed.

src/frontend/org/voltdb/VoltTable.java
jhugg: Fixed I think.

src/frontend/org/voltdb/VoltTable.java
jhugg: BWIFI, but I changed this in two places.

src/frontend/org/voltdb/VoltTable.java
jhugg: Clarified.

src/frontend/org/voltdb/VoltTable.java
jhugg: I think I did this. Code looks better.

src/frontend/org/voltdb/VoltTable.java
jhugg: My hunch is this is not a common miss and there isn't much performance change here for the branch comparing two constants that's always false. I personally prefer the explicit check. Willing to reconsider if you push.

src/frontend/org/voltdb/VoltTable.java
jhugg: I think there's a comment right afterward explaining this gets thrown sometimes if the buffer is full.

src/frontend/org/voltdb/VoltTableRow.java
jhugg: Good catch

src/frontend/org/voltdb/VoltTableRow.java
jhugg: Yeah... not sure we should bother, but working with buffers, it's nice to not change pos whenever a method takes a param that specifies a location. If the method doesn't specify, then usually you change pos to the end of the chunk you just dealt with.

Good catch on the +4. I cleaned this up slightly.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -703,6 +727,223 @@ public final VoltTableRow fetchRow(int index) {
...
+                default:
+                    throw new VoltTypeException("Unsupported type: " + columnType);
+                }
+            }
+            catch (ClassCastException cce) {
+                throw new VoltTypeException("Value for column " + col + " (" +
+                        getColumnName(col) + ") is type " +
+                        value.getClass().getSimpleName() + " when type " + columnType +
+                        " was expected.");
+            }
```
nshi: It looks like a large part of the type coercion code above is somewhat covered by the ParameterConverter. Can they share the same piece of code?

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -710,15 +951,109 @@ public final VoltTableRow fetchRow(int index) {
...
+        if (row.getColumnCount() != m_colCount) {
+            throw new IllegalArgumentException(row.getColumnCount() + " arguments but table has " + m_colCount + " columns");
+        }
+
+        // memoize the start of this row in case we roll back
+        final int pos = m_buffer.position();
+
+        try {
+            // Allow the buffer to grow to max capacity
+            m_buffer.limit(m_buffer.capacity());
```
nshi: Any specific reason to do this? This makes it hard to use a shared buffer to serialize stuff that assumes only a certain size to be used.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -710,15 +951,109 @@ public final VoltTableRow fetchRow(int index) {
...
+            byte[] inboundSchemaString = row.getSchemaString();
+            byte[] mySchemaString = getSchemaString();
+
+            // The way this works is that when two schema strings are found to have
+            // the same value, the target table's reference is pointed at the source
+            // table's reference. This allows the copying of multiple rows from one
+            // table to another to only do a deep comparison once, and to do reference
+            // equivalence checks for subsequent rows.
+            boolean canDoRawCopy = (inboundSchemaString == mySchemaString) ||
+                    Arrays.equals(inboundSchemaString, mySchemaString);
```
nshi: `Arrays.equals()` already handles the case where they are referencing the same object.

src/frontend/org/voltdb/VoltTable.java
nshi: Never mind, I see why you do this. Each row insertion sets the limit back to the position. It's kind of weird, though.

src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -373,6 +386,71 @@ public final boolean wasNull() {
...
+     * same schema.
+     */
+    final byte[] getRawRow() {
+        long rowSize = m_buffer.getInt(m_position - ROW_HEADER_SIZE);
+        byte[] fullRow = new byte[(int) (ROW_HEADER_SIZE + rowSize)];
+        int pos = m_buffer.position();
+        m_buffer.position(m_position - ROW_HEADER_SIZE);
+        m_buffer.get(fullRow);
+        m_buffer.position(pos);
+        return fullRow;
```
nshi: You can just return a slice of the ByteBuffer to avoid making a copy each time.

src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -373,6 +386,71 @@ public final boolean wasNull() {
...
+            length = m_buffer.getInt(offset);
+            if (length == VoltTable.NULL_STRING_INDICATOR) {
+                length = 0;
+            }
+            length += 4;
+            retval = new byte[length];
+            m_buffer.position(offset);
+            m_buffer.get(retval);
+            m_buffer.position(pos);
+            return retval;
```
nshi: Same here, you can return a slice of the ByteBuffer with a proper limit to avoid the copy.

src/frontend/org/voltdb/VoltTable.java
jhugg: Not sure I mind this being explicit?

src/frontend/org/voltdb/VoltTableRow.java
jhugg: Hmm... It's not clear to me how slow this method is for smaller values, but could be worth looking into.

src/frontend/org/voltdb/VoltTable.java
jhugg: I don't know? We're much stricter here... probably faster too.

src/frontend/org/voltdb/VoltTableRow.java
nshi: I'm not particularly worried about the speed of the copy itself. This just seem unnecessarily causing more minor GCs to happen because it generates a ton of these small objects in the young gen, but the GCs are probably also fast.

PR #2744, opened 2015-08-25
Eng 8781 write dr conflicts

luyangco:


src/ee/storage/BinaryLogSink.cpp
```diff
@@ -290,6 +303,24 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
     return static_cast<int64_t>(rowCount);
 }
 
+void BinaryLogSink::exportDRConflict(PersistentTable *drTable, Table *exportTable, const DRRecordType &type, TableTuple &exportTuple) {
+    if (exportTable != NULL && exportTable->isExport()) {
```
afrogers: I would hope you could assert that `exportTable->isExport()`. Seems like an odd thing to guard for, since that sounds like a fundamentally invalid scenario for it to be false.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -31,6 +32,17 @@
 
 namespace voltdb {
 
+const uint8_t MAX_CLUSTER_ID = static_cast<uint8_t>((1 << 8) - 1);
```
afrogers: This is a little bit of an odd place for these. Might be nicer to factor them out into their own header.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -290,6 +303,24 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
     return static_cast<int64_t>(rowCount);
 }
 
+void BinaryLogSink::exportDRConflict(PersistentTable *drTable, Table *exportTable, const DRRecordType &type, TableTuple &exportTuple) {
+    if (exportTable != NULL && exportTable->isExport()) {
+        TableTuple tempTuple = exportTable->tempTuple();
+        NValue hiddenColumn = exportTuple.getHiddenNValue(drTable->getDRTimestampColumnIndex());
+        int64_t drId = ValuePeeker::peekAsBigInt(hiddenColumn);
```
afrogers: The hidden column doesn't contain a drId. It contains the hidden column type as constructed at line 90 of ExecutorContext.hpp. Lowest 49 bits are a timestamp; 8 after that are a clusterId; rest are unused.

src/ee/storage/BinaryLogSink.cpp
afrogers: Or maybe don't define them at all, given the below comment.

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -490,6 +515,28 @@ TEST_F(DRBinaryLogTest, PartitionedTableNoRollbacks) {
...
+    // single row write transaction
+    TableTuple otherTempTuple = m_otherTableWithoutIndex->tempTuple();
+    otherTempTuple.setNValue(0, ValueFactory::getTinyIntValue(123));
+    otherTempTuple.setNValue(1, ValueFactory::getBigIntValue(9999999999));
+
+    m_sink.exportDRConflict(m_otherTableWithoutIndex, m_exportTable, DR_RECORD_INSERT, tempTuple);
+    m_sink.exportDRConflict(m_otherTableWithoutIndex, m_exportTable, DR_RECORD_INSERT, otherTempTuple);
+
+    // Really have no idea what to test for now.
+    EXPECT_EQ(2, m_exportTable->activeTupleCount());
```
afrogers: Is there a reason you're not validating what these tuples are? Would be nice to look them up and make sure that table name/cluster ID/etc. got set as you expected.

src/ee/storage/BinaryLogSink.cpp
luyangco: Thanks for pointing out, I have to say at the time when I wrote these lines, I frequently misused the concept of `drId`, `uniqueId` and bits stored in hidden column.

tests/ee/storage/DRBinaryLog_test.cpp
luyangco: Since `m_exportTable` is a export table not persistent table, I don't find a good method to get the value of tuple and test the equality. Do you have any ideas?

afrogers: The best thing I can think of is to create a dummy wrapper for the export table that then immediately deserializes the exported tuple and adds it to a list that you can match against expectations.

Start with `exportEnabled=false` (but still `exportOnly=true`) in the call to the `TableFactory`, and then set `m_wrapper` with a dummy like we use for the DR stream. Probably would need to expose some test method to do this, since `m_wrapper` is private on `StreamedTable`. Sounds like a lot of work, honestly.

Could also just make a `PersistentTable` lie and say it's an export table? That's ugly, though.

src/ee/storage/BinaryLogSink.cpp
luyangco: We still need the `getTimestampFromHiddenNValue()` and `getClusterIdFromHiddenColumn()` anyway, maybe define in `TableTuple.h` as static functions is a better idea? Logic behind that is `getHiddenNValue()` is also at this header file, let people more easily to find out these handy functions.

afrogers: I wouldn't necessarily put them in `TableTuple.h`; this is just one specific hidden column value. There are plans to use hidden columns for a variety of purposes in the future.

If anywhere, maybe put them into `ExecutorContext.hpp`, and extract the logic for packing the column values as well. That seems like a reasonably unified place for it.

afrogers:
Merged externally.

PR #2745, opened 2015-08-26
Eng 8876 - support % value for RSS limit configuration

manjujames:


src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -90,7 +90,7 @@ public boolean isOverLimitConfiguration()
                 continue;
             }
             if (!isDiskAvailable(config.m_path, config.m_diskSizeLimitPerc, config.m_diskSizeLimit)) {
-                m_logger.warn("Disk is over configured limits for feature " + config.m_featureName);
+                m_logger.error("Disk is over configured limits for feature " + config.m_featureName);
```
akhanzode: Should be better message like "Your disk capacity for feature has crossed the configured limit of xxx. and some action....

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -88,7 +92,7 @@ public void run()
         }
 
         if (isOverMemoryLimit() || m_diskLimitConfig.isOverLimitConfiguration()) {
-            m_logger.warn("Pausing the server");
+            m_logger.error("Resource monitor detected limit reached. Pausing the server.");
```
akhanzode: Nice to log which limit is reached.

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -109,7 +113,7 @@ private boolean isOverMemoryLimit()
             m_logger.debug("RSS=" + datum.rss + " Configured rss limit=" + m_rssLimit);
         }
         if (datum.rss >= m_rssLimit) {
-            m_logger.warn(String.format("RSS %d is over configured limit value %d.", datum.rss, m_rssLimit));
+            m_logger.error(String.format("RSS %d is over configured limit value %d.", datum.rss, m_rssLimit));
```
akhanzode: Again good user friendly message.

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -127,4 +131,32 @@ public void setBackPressure(boolean hasBackPressure)
     {
         // nothing to do here.
     }
+
+    // package-private for junit
+    double getMemoryLimitSize(String sizeStr)
+    {
+        sizeStr = sizeStr.trim();
+        if (sizeStr==null || sizeStr.length()==0) {
```
akhanzode: null check is not useful as trim would bomb with NPE

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -127,4 +131,32 @@ public void setBackPressure(boolean hasBackPressure)
...
+            return 0;
+        }
+
+        try {
+            if (sizeStr.charAt(sizeStr.length()-1)=='%') { // size as a percentage of total available memory
+                int perc = Integer.parseInt(sizeStr.substring(0, sizeStr.length()-1));
+                if (perc<0 || perc > 100) {
```
akhanzode: 0 should not come here? or if it does will it get set to 0 and disabled? we should not include 100

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -127,4 +131,32 @@ public void setBackPressure(boolean hasBackPressure)
...
+                    throw new IllegalArgumentException("Invalid memory limit percentage: " + sizeStr);
+                }
+                return PlatformProperties.getPlatformProperties().ramInMegabytes*1048576L*perc/100.0;
+            } else { // size in GB
+                double size = Double.parseDouble(sizeStr)*1073741824L;
+                if (size<0) {
```
akhanzode: u can check 0 before multiplication.

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -127,4 +131,32 @@ public void setBackPressure(boolean hasBackPressure)
...
+                    throw new IllegalArgumentException("Invalid memory limit value: " + sizeStr);
+                }
+                return size;
+            }
+        } catch(NumberFormatException e) {
+            throw new IllegalArgumentException("Invalid memory limit value " + sizeStr +
+                    ". Memory limit must be configued as a percentage of total available memory or as GB value");
+        }
+    }
```
akhanzode: also think how will VMC know the unit and if its a % or not. We may have to add jackson mixin I would rather avoid it. Can GB be part of what user specifies that way deployment can just be returned as is to VMC

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: The trim is no longer required because I have to do it at the caller. There is a default value, so null is not really required. I just have it there..

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: It will get here if user explicitly sets 0. If we want to disallow 100%, I will make 99 the maximum value.

src/frontend/org/voltdb/ResourceUsageMonitor.java
manjujames: It will add one more line of code ;-)

PR #2751, opened 2015-08-27
Add AdHoc queries timeout feature

jxchinaren:


src/frontend/org/voltdb/compiler/AdHocPlannerWork.java
```diff
@@ -117,7 +122,7 @@ public static AdHocPlannerWork makeStoredProcAdHocPlannerWork(long replySiteId,
             // should be no correlation inferred or assumed between the partitioning and the
             // statement's constants or parameters.
             false, (singlePartition ? new Object[1] /*any vector element will do, even null*/ : null),
-            "@AdHoc_RW_MP", ProcedureInvocationType.ORIGINAL, 0, 0,
+            "@AdHoc_RW_MP", ProcedureInvocationType.ORIGINAL, 0, 0, BatchTimeoutOverrideType.NO_TIMEOUT,
```
paulmartel: Is this really the code path used by the experimental stored proc dynamic sql feature?
If so, AND we happen to be in a RO proc, we should probably be passing in the current override of that proc's invocation -- also seems like we should not be calling "@AdHoc_RW_MP" from read-only or SP procs -- but that's a separate issue. Maybe it's just the comment before these methods that's wrong? 
We should test whether an AdHoc called FROM a stored proc uses the proper timeout override value.
If it does not, I am willing to release note this as a known limitation for 5.6 and fix later.

jxchinaren: @paulmartel  "AdHoc_RW_MP" makes me think that this is not relevant to read only query timeout. That's why I use the BatchTimeoutOverrideType.NO_TIMEOUT.

I've pushed another commit to add these tests also.

jxchinaren:
@paulmartel I will do squash merge this is finally passed through review.

paulmartel:
This looks good except that we want to use Andrew's wording in the javadoc:

    WARNING: Use of a queryTimeout value that is greater than the global timeout value for your VoltDB configuration will temporarily override that safeguard. Currently, non-privileged users (requiring only SQLREAD permissions) can invoke this method, potentially degrading system performance with an uncontrolled long-running procedure.

jxchinaren:
Squash merge to master instead.

PR #2758, opened 2015-08-28
ENG-8872: Update persistent table logic so that it will have atleast one block assigned to it 

hhakimi53:
Logic updates include:
- Allocate the block to hold tuples gets assigned to persistent table when one requested from table factory instead of doing so on first tuple insertion.
- When freeing up tuple storage during tuple delete, don't release the empty block if it's the only remaining block.

### DIFF
--- a/src/ee/storage/persistenttable.h
+++ b/src/ee/storage/persistenttable.h
@@ -886,7 +886,9 @@ inline void PersistentTable::deleteTupleStorage(TableTuple &tuple, TBPtr block)
 
     bool transitioningToBlockWithSpace = !block->hasFreeTuples();
 
+    //uint32_t activeTuples = block->activeTuples();
     int retval = block->freeTuple(tuple.address());
+    //std::cout<<"Active Tuples: before free: "<< activeTuples << " after free: "<<block->activeTuples() << " number of blocks: " << m_data.size()<<std::endl;
     if (retval != -1) {
         //Check if if the block is currently pending snapshot
         if (m_blocksNotPendingSnapshot.find(block) != m_blocksNotPendingSnapshot.end()) {
@@ -901,7 +903,10 @@ inline void PersistentTable::deleteTupleStorage(TableTuple &tuple, TBPtr block)
         }
     }
 
-    if (block->isEmpty()) {
+    // if the current block is empty and there are more than one blocks, release the existing
+    // empty block.
+    if (block->isEmpty() && m_data.size() > 1) {
+        //std::cout<<"release the block"<<std::endl;
         m_data.erase(block->address());
         m_blocksWithSpace.erase(block);
         m_blocksNotPendingSnapshot.erase(block);
--- a/src/ee/storage/tablefactory.cpp
+++ b/src/ee/storage/tablefactory.cpp
@@ -92,6 +92,16 @@ Table* TableFactory::getPersistentTable(
     // initialize stats for the table
     configureStats(databaseId, name, table);
 
+    if(!exportOnly) {
+        PersistentTable *persistentTable = dynamic_cast<PersistentTable*>(table);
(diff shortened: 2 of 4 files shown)


src/ee/storage/persistenttable.h
```diff
@@ -886,7 +886,9 @@ inline void PersistentTable::deleteTupleStorage(TableTuple &tuple, TBPtr block)
 
     bool transitioningToBlockWithSpace = !block->hasFreeTuples();
 
+    //uint32_t activeTuples = block->activeTuples();
```
wolffcm: Commented out code looks messy.  If there's debug code that you think might be useful at a later time, please use `VOLT_DEBUG`, `VOLT_TRACE`, etc macros instead.

src/ee/storage/persistenttable.h
```diff
@@ -901,7 +903,10 @@ inline void PersistentTable::deleteTupleStorage(TableTuple &tuple, TBPtr block)
         }
     }
 
-    if (block->isEmpty()) {
+    // if the current block is empty and there are more than one blocks, release the existing
```
wolffcm: It would be worth saying why you always keep one block around, describing the performance benefit to doing so.  I could imagine someone else coming along a year from now, and reverting your change to save memory.

src/ee/storage/tablefactory.cpp
```diff
@@ -92,6 +92,16 @@ Table* TableFactory::getPersistentTable(
     // initialize stats for the table
     configureStats(databaseId, name, table);
 
+    if(!exportOnly) {
+        PersistentTable *persistentTable = dynamic_cast<PersistentTable*>(table);
```
wolffcm: I don't think this code is on a performance-critical path, but I would still use a `static_cast` here, because `dynamic_cast` has more overhead.  It's also pretty obvious that the table will be persistent, since it's allocated right above.  If you want to be extra defensive, you can `assert(dynamic_cast<PersistentTable*>(table))`.

src/ee/storage/tablefactory.cpp
```diff
@@ -92,6 +92,16 @@ Table* TableFactory::getPersistentTable(
...
+        assert(persistentTable);
+        if(persistentTable) {
```
wolffcm: This condition seems trivially true given this method, consider removing it.

src/ee/storage/tablefactory.cpp
```diff
@@ -92,6 +92,16 @@ Table* TableFactory::getPersistentTable(
...
+            TBPtr block = persistentTable->allocateNextBlock();
+            if (block->hasFreeTuples()) {
```
wolffcm: Will this always be true?  If so, consider changing it to an assert.  If it won't always be true, then it seems like we might be leaking the block here?

tests/ee/storage/CompactionTest.cpp
```diff
@@ -507,7 +509,7 @@ TEST_F(CompactionTest, CompactionWithCopyOnWrite) {
 
     }
     m_table->doForcedCompaction();
-    ASSERT_EQ( m_table->m_data.size(), 0);
+    ASSERT_EQ( m_table->m_data.size(), 1);
```
wolffcm: Maybe add a test that ensures that a block still exists even after executing a `TRUNCATE TABLE`?

wolffcm:
That's all my comments.

src/ee/storage/persistenttable.h
hhakimi53: It was for debugging purpose; I will take it out

src/ee/storage/persistenttable.h
hhakimi53: sure, will do.

PR #2763, opened 2015-08-28
ENG-7685: Adding materialized view to the Txnid2-selfcheck app

liam700:


tests/test_apps/txnid-selfcheck2/src/txnIdSelfCheck/procedures/UpdateBaseProc.java
```diff
@@ -210,7 +225,35 @@ public static void validateCIDData(VoltTable data, String callerId) {
...
+                throw new VoltAbortException("View has multiple entries of the same cid, that should not happen.");
+            VoltTableRow row0 = view.fetchRow(0);
+            long v_entries = row0.getLong("entries");
+            long v_max = row0.getLong("maximum");
+            long v_min = row0.getLong("minimum");
+            long v_sum = row0.getLong("summation");
+
+            if (v_entries != entries)
+                throw new VoltAbortException(
+                        "The count(*):"+v_entries+" materialized view aggregation does not match the number of cnt entries:"+entries+" for client:"+cid);
```
liam700: The reason that I included the client and the cid in this error message is so that the tester can look at the table and materialized view and find the client that caused the exception.

PR #2769, opened 2015-08-31
Eng 8849 full table scan warning

luyangco:


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1525,6 +1525,15 @@ void addTableToCatalog(Database db, VoltXMLElement node) throws VoltCompilerExce
             }
         }
 
+        // Warn user if DR table don't have any unique index.
+        if (db.getIsactiveactivedred() &&
+                node.attributes.get("drTable") != null &&
+                node.attributes.get("drTable").equalsIgnoreCase("ENABLE") &&
+                table.getIndexes().isEmpty()) {
```
nshi: This only catches the case where the DR table doesn't have any index. If the table has non-unique index, this won't detect it.

luyangco: Fixed. I have to say at the time when I wrote these lines, my knowledge about unique index is really misleading :)

nshi:
Merged externally.

PR #2770, opened 2015-08-31
ENG-8709: Remove stx btree

nshi:
* Replaced use of stx btree with CompactingMap in TupleBlock.
* Implemented CompactingSet on top of CompactingMap.
* Replaced use of stx btree with CompactingSet in ElasticIndex.


src/ee/storage/ElasticIndex.h
```diff
@@ -89,19 +89,21 @@ class ElasticIndexKey
...
  */
 class ElasticIndexComparator
 {
   public:
-    bool operator()(const ElasticIndexKey &a, const ElasticIndexKey &b) const;
+    int operator()(const ElasticIndexKey &a, const ElasticIndexKey &b) const;
 };
 
+struct vdb_empty_struct {};
+typedef struct vdb_empty_struct data_type;
```
luyangco: Dead code?

nshi: No. CompactingSet uses this as the data type to put into the underlying
map so that it doesn't take up additional memory.

luyangco: In `CompactingSet.h` it uses emptyStruct for this purpose, and I searched data_type this is no other place use it. Maybe the code is duplicated?

nshi: Oh, yes, you are right. I thought this was the one I added in CompactingSet. I'll remove this instance. Thanks.

tests/ee/storage/CompactionTest.cpp
```diff
@@ -334,19 +334,19 @@ TEST_F(CompactionTest, BasicCompaction) {
         pkeysFoundAfterDelete.insert(pkey);
     }
 
-    std::vector<int32_t> diff;
-    std::insert_iterator<std::vector<int32_t> > ii( diff, diff.begin());
-    std::set_difference(pkeysNotDeleted.begin(), pkeysNotDeleted.end(), pkeysFoundAfterDelete.begin(), pkeysFoundAfterDelete.end(), ii);
-    for (int ii = 0; ii < diff.size(); ii++) {
-        printf("Key that was not deleted, but wasn't found is %d\n", diff[ii]);
-    }
+    // std::vector<int32_t> diff;
```
luyangco: Commented code should be removed, unless we need it later to for getting debug output, same as other commented code in this file. Maybe the good practice is protecting them by some macros.

src/ee/structures/CompactingMap.h
```diff
@@ -91,10 +112,16 @@ class NormalKeyValuePair : public std::pair<Key, Data> {
  *
  */
 
-template<typename KeyValuePair, typename Compare, bool hasRank=false>
+template<typename KeyValuePair, typename Compare, bool hasRank=false, typename ValueType=KeyValuePair>
```
jxchinaren: Why adding the extra typename here? 
If there is no other places using other than the default value "KeyValuePair", maybe add a comment here?

src/ee/structures/CompactingMap.h
```diff
@@ -197,12 +336,40 @@ class CompactingMap {
...
+        return z;
+    }
+
+    void copyRecursive(const TreeNode *fromNIL, const TreeNode *from, TreeNode *to);
+
     void erase(TreeNode *z);
     TreeNode *lookup(const Key &key) const;
     TreeNode *lookupRank(int64_t ith) const;
 
+    TreeNode *findLowerBound(const Key &key) const;
```
jxchinaren: As they are internal functions, maybe should make them private other than protected? 

I understand we have not done so for other existing functions. So this may be filed as a new ticket to do code refactoring.

jxchinaren:
Those are all my comments, I like this change.

src/ee/structures/CompactingMap.h
nshi: CompactingSet uses a different ValueType because it only cares about the key part of the KeyValuePair.

src/ee/structures/CompactingMap.h
nshi: Making them private would prevent any descendants of CompactingMap from using these functions. I'm not sure that they have to be so strict.

tests/ee/storage/CompactionTest.cpp
nshi: I'll clean this test up.

src/ee/structures/CompactingMap.h
nshi: I'll add the comment.

luyangco:
End of my code review.

BillWhite:
I actually looked through this yesterday.  I don't know about the performance implications, but it all looked plausible and straightforward to me.

nshi:
Thank you, Bill.

nshi:
Squash merged externally.

PR #2771, opened 2015-08-31
Eng 8872 new

hhakimi53:
Logic updates include:
-  Allocation of the tuple storage block to persistent table happens when creating persistent table via table factory method instead of doing so on demand - first tuple insertion.
- When freeing up tuple storage during tuple delete, caller can specify whether to release the empty tuple storage block if it's the only remaining block.
- Added unit test for truncate table case.

### DIFF
--- a/tests/ee/storage/CompactionTest.cpp
+++ b/tests/ee/storage/CompactionTest.cpp
@@ -363,7 +363,9 @@ TEST_F(CompactionTest, BasicCompaction) {
         m_table->deleteTuple(tuple, true);
     }
     m_table->doForcedCompaction();
-    ASSERT_EQ( m_table->m_data.size(), 0);
+    // ENG-8872: at any given point, persistent table will have atleast one block available with m_data
+    // even if that block is empty
+    ASSERT_EQ( m_table->m_data.size(), 1);
     ASSERT_EQ( m_table->activeTupleCount(), 0);
 }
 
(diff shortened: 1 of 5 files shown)


tests/ee/storage/CompactionTest.cpp
```diff
@@ -364,7 +364,9 @@ TEST_F(CompactionTest, BasicCompaction) {
         m_table->deleteTuple(tuple, true);
     }
     m_table->doForcedCompaction();
-    ASSERT_EQ( m_table->m_data.size(), 0);
+    // ENG-8872: at any given point, persistent table will have atleast one block available with m_data
+    // even if that block is empty
+    ASSERT_EQ( m_table->m_data.size(), 1);
```
BillWhite: This is somewhat minor, but I don't like putting ticket numbers in the code, unless it's
in a unit test, the ticket is open and ticket is associated with the failure of the test somehow.
After ENG-8872 is finished, it will not be important to the code.

If you want to docuement that the table has one block, even if it's empty, then just document
that fact.

BillWhite:
That's all I have to say.

tests/ee/storage/CompactionTest.cpp
hhakimi53: Will update the comment, thnxs.

wolffcm:
Looks good.  Thanks Huzefa.  This is a nice improvement---the customer will be glad this is fixed.

PR #2774, opened 2015-09-02
add isDebug function to disable query timeout test with debug build

jxchinaren:


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -1360,6 +1360,19 @@ public boolean isValgrind() {
         return templateCmdLine.m_backend == BackendTarget.NATIVE_EE_VALGRIND_IPC;
     }
 
+    public static boolean isDebugDefined() {
+        final String buildType = System.getenv().get("BUILD");
+        if (buildType == null) {
+            return false;
+        }
+        return buildType.toLowerCase().startsWith("debug");
```
BillWhite: As we discussed, this only works if the envariable BUILD is set appropriately.  While
this is true in Jenkins and when using the ant junit target, it would not be true in Eclipse
or in the other GUIs.  Eclipse has its own set of .class files.  I don't know about how
Eclipse finds the jni implementations.  This seems fragile and, at the least, not well
understood or documented.

BillWhite:
That's all I have to say.

PR #2775, opened 2015-09-02
Port Is Distinct from operator support to master

hhakimi53:
Review request for port of SQL "is distinct from" operator to master from HSQL upgrade branch (ENG-8348-IsDistinctFrom).

Changes include:
- Updated logic to support "is distinct from" operator
- Update query plan logic in hsql to not parameterize null constant for volt null type in query plan and use the plan with null constant instead. Currently this logic is not exercised as NULL constants for comparison operators does not seem to supported: ENG-8946.
- Add junit tests for is distinct from operator and moved case/when to the comparison operator junit test file. Currently these tests gets executed against native (VoltDB) backend only and is not executed against HSQL backend as HSQL backend does not seem to have full support for "Is Distinct from" implemented to execute the compiled "is distinct from" plan and results in run time exception - "unsupported internal operation: Expression".

Testing: 
- Passed jenkins jobs.


src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLogical.java
```diff
@@ -697,6 +697,11 @@ private void resolveTypesForComparison(Session session,
 
     private void resolveRowTypes() {
 
+        // NOTE: in hsql upgrade the logic below takes in account if the left
```
wolffcm: I'm having some trouble parsing this comment.  Are you saying that we can't support IS DISTINCT FROM in HSQL 2.3 since there is a bug here?

hhakimi53: The intent of comment in here was if we have to bring over some of the changes from the upgrade branch, in that case we should not port as is as there is bug in the resolverowtype() where it decides how to resolves left/right child whose datatype is null. Fix for that is living on IsDistinctFrom branch based of HSQL upgrade branch. And strictly it does not have to with is Distinct operator even.
I will remove the comment from here as at present I dont think we are cherry picking the HSQL changes otherwise this will cause confusion. Thnxs

src/frontend/org/voltdb/planner/ParameterizationInfo.java
```diff
@@ -164,6 +164,14 @@ static void parameterizeRecursively(VoltXMLElement node,
             String idStr = node.attributes.get("id");
             assert(idStr != null);
 
+            // handle parameter value type
+            String typeStr = node.attributes.get("valuetype");
+            VoltType vt = VoltType.typeFromString(typeStr);
+            // don't parameterize the plan for NULL constants of type "VoltType.NULL"
+            if (vt == VoltType.NULL) {
```
wolffcm: This change seems like it affects lots of different kinds of statements, rather than just statements that contain IS [NOT] DISTINCT FROM.  I'd like to understand the whole implications of this change.

I just tried `SELECT * FROM t WHERE i = NULL;` and it fails with "data type cast needed for parameter or null literal".  This seems like a a bogus message since we should be able to infer the type of the NULL from the LHS of the `=`.  Does this change here fix that issue as well?

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,465 @@
...
+        client.callProcedure("S2.insert", 5, 5253, 3);
+    }
+
+    private void subTestIsDistinctFrom(Client client)
+            throws Exception
+    {
+        // results for HSQL backend below are based on HSQL 2.3.2/3 upgrade branch.
+        // is Distinct from operator does not on HSQL backend based on 1.9.2/3. So
+        // assert in case if these test cases are executed on HSQL backend
+        assert(!isHSQL());
```
wolffcm: I'm having trouble parsing this comment.  There must be IS DISTINCT FROM support in HSQL 1.9 since you haven't made any changes to the HSQL parser itself.  So why can't we run this test with the HSQL back end?

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
```diff
@@ -0,0 +1,465 @@
...
+        //System.out.println("\n\nsql: " + sql + "\nResult: " + (vt.toString()) + " row Count: "+ vt.getRowCount() + "\n\n");
+        validateTableOfLongs(client, sql, new long[][] {{1, 1000, 1, 1000, 2}, {5, 2553, 3, 5253, 3}});
+
+
+
+        // case 2.1: on column that can have null values; result set does not has null values
+        sql = "Select S1.ID ID," +
+                "S1.Wage, S1.Dept, " +
+                "S2.WAGE, S2.Dept " +
+                "from S1, S2 where S1.WAGE is not distinct from S2.WAGE;";
```
wolffcm: Part of the point of IS NOT DISTINCT FROM is that it acts like `=` but that NULLs compare as equal, right?  Maybe I'm missing something but I don't see any comparisons of that type here, where there are NULLs on both sides of the operator.

src/frontend/org/voltdb/planner/ParameterizationInfo.java
hhakimi53: This change intent is to disallow parameterization of NULL parameter and let it be treated as constants themselves. Otherwise it can result into path where it tries to obtain value for NULL of type NULL if let the NULL constants to be parameterized. This prevents doing it so.

When compiling queries which have NULL constants in comparison (>, <, <>, =), it will not compile as it fails to resolve the data type of child which is null in expression. I had discovered this when working on is distinct from operator and we have a ticket for this - ENG-8946, to track the issue.

wolffcm: I guess I'm not understanding what this change actually does.  What wasn't working before that works now as a result of this change?

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: Yes, we don't have NULL in both tables and is distinct will not have any such matches due to that. This test case intent is to make sure that it does not match NULL to a non-NULL column (false evaluation).

tests/frontend/org/voltdb/regressionsuites/TestComparisonOperatorsSuite.java
hhakimi53: When we try to run it against is distinct from against HSQL backend, we get unsupported operation run time exception. Based on stack trace, I did not see support for is distinct from in there. So does not seem that it is fully implemented.

org.voltdb.client.ProcCallException: VOLTDB ERROR: HSQL-BACKEND ERROR
  General error java.lang.RuntimeException: unsupported internal operation: Expression
    at org.voltdb.sysprocs.AdHoc_RO_SP.run(AdHoc_RO_SP.java:52)
	at org.voltdb.client.ClientImpl.callProcedure(ClientImpl.java:297)
	at org.voltdb.client.ClientImpl.callProcedureWithClientTimeout(ClientImpl.java:251)
	at org.voltdb.client.ClientImpl.callProcedure(ClientImpl.java:206)
	at org.voltdb.regressionsuites.RegressionSuite.validateTableOfLongs(RegressionSuite.java:392)
	at org.voltdb.regressionsuites.TestComparisonOperatorsSuite.subTestIsDistinctFrom(TestComparisonOperatorsSuite.java:137)
	at org.voltdb.regressionsuites.TestComparisonOperatorsSuite.testIsDistinctFrom(TestComparisonOperatorsSuite.java:99)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at junit.framework.TestCase.runTest(TestCase.java:168)
	at junit.framework.TestCase.runBare(TestCase.java:134)
	at junit.framework.TestResult$1.protect(TestResult.java:110)
	at junit.framework.TestResult.runProtected(TestResult.java:128)
	at junit.framework.TestResult.run(TestResult.java:113)
	at junit.framework.TestCase.run(TestCase.java:124)
	at junit.framework.TestSuite.runTest(TestSuite.java:243)
	at junit.framework.TestSuite.run(TestSuite.java:238)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:83)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:50)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:467)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:683)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:390)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:197)

wolffcm:
That's all my comments.  Looks good otherwise.

wolffcm:
Looks good, thanks for updating the tests.

PR #2777, opened 2015-09-02
Add query timeout sqlcmd tests

jxchinaren:


tests/sqlcmd/scripts/querytimeout/timeout.config
```diff
@@ -0,0 +1 @@
+--query-timeout=1000
```
wolffcm: Missing newline at end of file.

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -272,15 +272,28 @@ def do_main():
...
                 if not inpath.endswith(".in"):
                     continue
-                print "Running ", os.path.join(parent, inpath)
                 prefix = inpath[:-3]
+
+                config_params = ""
+                prompt = "Running " + os.path.join(parent, inpath)
+                if os.path.isfile(os.path.join(parent, prefix + '.config')):
+                    with open (os.path.join(parent, prefix + '.config'), "r") as configFile:
+                        config_params=configFile.read().replace('\n', '')
```
wolffcm: Will this work with more than one option in the .config file?  Seems like you would want to split the contents of the file on the `\n` character (producing a list), and the concatenate that to `['../../sqlcmd']` in the call to `subprocess.call` below.

wolffcm:
That's all my comments.

tests/sqlcmd/sqlcmdtest.py
jxchinaren: Yes, it's only working for 1 configuration. I added the multiple configuration support.

wolffcm:
Looks good to me.

PR #2779, opened 2015-09-03
ENG-8962: Now able to add status/error code also into AsyncCompilerRe…

manjujames:
…sult.

### DIFF
--- a/src/frontend/org/voltdb/ClientInterface.java
+++ b/src/frontend/org/voltdb/ClientInterface.java
@@ -2388,7 +2388,7 @@ public class ClientInterface implements SnapshotDaemon.DaemonInitiator {
                 else {
                     ClientResponseImpl errorResponse =
                         new ClientResponseImpl(
-                                ClientResponseImpl.GRACEFUL_FAILURE,
+                                (result.errorCode == ClientResponse.UNINITIALIZED_APP_STATUS_CODE) ? ClientResponse.GRACEFUL_FAILURE : result.errorCode,
                                 new VoltTable[0], result.errorMsg,
                                 result.clientHandle);
                     writeResponseToConnection(errorResponse);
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2388,7 +2388,7 @@ else if (result instanceof CatalogChangeResult) {
                 else {
                     ClientResponseImpl errorResponse =
                         new ClientResponseImpl(
-                                ClientResponseImpl.GRACEFUL_FAILURE,
+                                (result.errorCode == ClientResponse.UNINITIALIZED_APP_STATUS_CODE) ? ClientResponse.GRACEFUL_FAILURE : result.errorCode,
```
akhanzode: why use app status code for status? why not RESPONSE_UNKNOWN -> GRACEFUL_FAILURE otherwise known status which is errorCode

akhanzode:
other than the one comment rest looks fine.

src/frontend/org/voltdb/ClientInterface.java
manjujames: RESPONSE_UNKNOWN means - Status code indicating that the response was lost, and the outcome of the invocation is unknown.
The one that I use currently is the one for a status code byte value that means 'uninitialized'

akhanzode: But its used for appStatus and not status in response.RESPONSE_UNKNOWN is used internally for failover and such not necessarily for sending response to user always. You can always have different code for AsyncCompilerResult and does not have to use ClientResponse codes. Thats actually more cleaner.

akhanzode:
Looks fine now...next time we touch this we should drive everything by errorCode

manjujames:
OK, thanks.

manjujames:
squash merged.

PR #2781, opened 2015-09-03
Need to have name for hidden columns for snapshot restore to work

luyangco:
serialize/deserialize to DR has no knowledge of hidden columns, normal serialize/deserialize now has; native & stream snapshot can now save/restore hidden columns, index snapshot needs investigation

Make restore procedure DR-aware, add related tests

Consider hidden columns when loading table in EE

Move setDRTimestamp outside insertTupleCommon for snapshot restore purpose

Fix missing declaration

Refactor add hidden column logic in table converter

Add logic for EE to set uninitialized DR timestamp column during snapshot restore

Clean up unnecessary code

ENG-8723, address remaining review comments, also rebase with master branch.

Change-Id: I34d50894d866b7f1174dea5fcaab86300cf323e5

ENG-8723, if schema contains hidden DR timestamp, tempTuple creation shoule initialize the hidden column value to NULL.
Update hidden DR timestamp is always allowed but insert new value to DR timestamp column need to check if it is NULL.

Change-Id: Ia27e6050212b10afb5378a6f876b67cd8ab93843

### DIFF
--- a/build.py
+++ b/build.py
@@ -204,6 +204,7 @@ CTX.INPUT['common'] = """
  RecoveryProtoMessage.cpp
  RecoveryProtoMessageBuilder.cpp
  DefaultTupleSerializer.cpp
+ FullTupleSerializer.cpp
  executorcontext.cpp
  serializeio.cpp
  StreamPredicateList.cpp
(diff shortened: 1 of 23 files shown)


luyangco:
I reviewed Ray's PR and fixed the problems I found, when Jenkins passes all tests then I gonna merge.

PR #2783, opened 2015-09-04
Add EE to frontend callback interface for logging DR conflicts

luyangco:
Add test for new function in TupleSchema

Delete global refs in destructor of JNITopend

Add NULL check for output table parameter

Count length placeholder for variable length columns

ENG-8596, rename function to make the intention more obiviously.

Change-Id: I48a68c0e22dff20baecc11a7f74eb86182695a8a

### DIFF
--- a/src/ee/common/TupleSchema.h
+++ b/src/ee/common/TupleSchema.h
@@ -27,6 +27,7 @@
 
 #include "common/FatalException.hpp"
 #include "common/types.h"
+#include "common/NValue.hpp"
 
 #define UNINLINEABLE_OBJECT_LENGTH 64
 #define UNINLINEABLE_CHARACTER_LENGTH 16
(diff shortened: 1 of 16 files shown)


src/ee/common/TupleSchema.h
```diff
@@ -27,6 +27,7 @@
 
 #include "common/FatalException.hpp"
 #include "common/types.h"
+#include "common/NValue.hpp"
```
luyangco: Remove unnecessary header file.

src/ee/common/tabletuple.h
```diff
@@ -442,6 +452,40 @@ class TableTuple {
             return (size_t)0;
         }
     }
+
+    inline size_t serializedColumnSize(int colIndex) const {
```
luyangco: Should reuse the `maxExportSerializedColumnSize` to reduce code redundancy.

src/ee/common/types.cpp
```diff
@@ -83,11 +83,24 @@ NValue getRandomValue(ValueType type) {
...
+            }
+            characters[i] = '\0';
+            return ValueFactory::getDecimalValueFromString(string(characters));
+        }
         case VALUE_TYPE_DOUBLE:
             return ValueFactory::getDoubleValue((rand() % 10000) / (double)(rand() % 10000));
         case VALUE_TYPE_VARCHAR: {
-            int length = (rand() % 10);
-            char characters[11];
+            int length = (rand() % 8);
```
luyangco: I don't understand why we need to change the constant here. Same as below.

PR #2784, opened 2015-09-04
advanced-scalar-subquery.sql: for those GROUP BY and HAVING queries that

johnTcrawford:
are commented out because they do not currently work (i.e., return
errors), cleaned them up, and their associated comments, to make it
clearer why they don't work (related to which Jira tickets), and what
fixes may make them work, as a follow-up to Paul's comments, and
a brief discussion we had, about ENG-8564.


paulmartel:
Looks good. Thanks.

PR #2785, opened 2015-09-04
Add a user-visible POINT type to VoltDB SQL

wolffcm:


src/frontend/org/voltdb/ParameterSet.java
```diff
@@ -167,6 +171,10 @@ else if (obj == VoltType.NULL_STRING_OR_VARBINARY) {
             else if (obj == VoltType.NULL_DECIMAL) {
                 size += 16;
                 continue;
+            }
+            else if (obj == VoltType.NULL_POINT) {
+                size += 8;
+                continue;
```
BillWhite: Doesn't this want to be a named constant somehow?  It seems like, if the representation of point changes, and in particular, if the size changes, you might have to ferret out a lot of places where the representation of points is known.

src/frontend/org/voltdb/ParameterSet.java
```diff
@@ -564,6 +575,11 @@ else if (nextType == VoltType.STRING) {
                     }
                     break;
                 }
+                case POINT :
+                    float lat = in.getFloat();
+                    float lng = in.getFloat();
+                    value = new PointType(lat, lng);
+                    break;
```
BillWhite: Places like here.

BillWhite: That's all I have for the moment.  I'd like to look at the tests in more detail, but
if I don't get to it don't wait on me.  This all seems straightforward enough.

src/ee/common/NValue.hpp
```diff
@@ -945,6 +946,18 @@ class NValue {
         return *reinterpret_cast<bool*>(m_data);
     }
 
+    PointType& getPoint() {
+        assert(getValueType() == VALUE_TYPE_POINT);
+        assert(sizeof(PointType) == 8); // 2*sizeof(float)
```
paulmartel: Seems like the wrong place for this assert -- one more place to have to find/change when we re-implement -- I'd expect this kind of assert closer to where we depend on the size or expose the internals. Maybe a more apropos assert here would be that sizeof(PointType) <= sizeof(m_data), which is presumed by the cast.

src/ee/common/NValue.hpp
```diff
@@ -2836,6 +2912,10 @@ template <TupleSerializationFormat F, Endianess E> inline void NValue::deseriali
     case VALUE_TYPE_DOUBLE:
         *reinterpret_cast<double* >(storage) = input.readDouble();
         break;
+    case VALUE_TYPE_POINT:
+        reinterpret_cast<float*>(storage)[0] = input.readFloat();
+        reinterpret_cast<float*>(storage)[1] = input.readFloat();
```
paulmartel: I think I'd have gone abstract here -- constructed a PointType instance and done a cast of storage and a PointType object assignment. Not sure it matters.

src/ee/common/PointType.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <limits>
+#include <sstream>
+
+namespace voltdb {
+/**
+ * A class for representing instances of geo-spatial points.
+ * Stored as a pair of floats, so it fits into an 8-byte word.
+ */
+class PointType {
```
paulmartel: I find the Type suffix redundant and confusing on classes that are not some kind of metadata "Type" class or type tag enum. Why not just Point? I'd even find "PointValue" clearer since it's a value class as opposed to an object class.

src/ee/common/executorcontext.cpp
```diff
@@ -218,8 +218,18 @@ void ExecutorContext::cleanupExecutorsForSubquery(int subqueryId) const
...
     typedef std::map<int, std::vector<AbstractExecutor*>* >::value_type MapEntry;
+
+    // if we're recovering from an error, the executors map may never
+    // have been initialized.
+    if (m_executorsMap == NULL) {
+        // if there's no executors, there's no temp tables to check,
+        // so return true.
+        return true;
+    }
+
```
paulmartel: This change seems off-topic. Consider making a separate pull request -- possibly for an expedited merge with master.

src/frontend/org/voltdb/ParameterSet.java
paulmartel: OR call VoltType.POINT.getLengthInBytesForFixedTypesWithoutCheck() ?

src/hsqldb19b3/org/hsqldb_voltpatches/types/VoltPointType.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    public Type getAggregateType(Type other) {
+        // incompatible types in operation
+        throw Error.error(ErrorCode.X_42561);
+    }
+
+    @Override
+    public Type getCombinedType(Type other, int operation) {
+        // The 'combined type' seems to refer to operators other than
+        // comparisons, which always return BOOLEAN, and are valid
```
paulmartel: had trouble parsing this comment -- the comma after "comparisons" made me think that "which always return BOOLEAN" applied to "operators" vs. "comparisons"

paulmartel:
End of review round.

src/ee/common/NValue.hpp
wolffcm: I was thinking that I should probably use BOOST_STATIC_ASSERT in the header for PointType.

src/ee/common/NValue.hpp
wolffcm: I thought about this.  Object assignment would be nicer to read.  But C++ doesn't guarantee that function arguments are evaluated left to right, so I would have had to hoist the calls to readFloat into temporary variables.  I chose this way to avoid having to hoist.  This is definitely premature optimization on my part (and one the compiler might do anyway), so I'll use object assignment.

src/ee/common/PointType.hpp
wolffcm: On the Java side, I chose `PointType` to be consistent with `TimestampType`, which is another user-facing class.  I figured I'd use the same name on the C++ side too.  Consistency aside, I like plain old `Point` too.  I'll change it.

src/ee/common/executorcontext.cpp
wolffcm: Ok, I'll pull it out and make a pull request for master.

wolffcm:
@BillWhite @paulmartel I've incorporated the feedback for your comments.  Thanks for your insights.

I've also tried to take Bill's suggestion to it's logical conclusion and push all representation-specific code into PointType and Point classes in Java and the EE respectively.  Hopefully this will ease any pain in changing representation later.  This change was not so trivial so it could probably do with a glance before it's made canon on the geospatial integ branch.

BillWhite:
I skimmed through this.  It seems better, though it was not bad before.
I'm happy with this.

wolffcm:
I squashed on a separate branch and will merge it to sql-geospatial.

PR #2786, opened 2015-09-04
ENG-8699 wait for initial truncation snapshot to complete before we s…

vtkstef:
…tart polling export persistent binary deques


akhanzode:
There is nothing interesting here. If pro is good this one is good to merge.

PR #2794, opened 2015-09-09
Check that executors map is not null in temp tables emptiness check

wolffcm:
This caused a hard EE crash during development.  (I had new functions
added to the planner not yet implemented in the EE.  This should have
just returned an exception to the user.)


hhakimi53:
looks good.

PR #2803, opened 2015-09-10
Change the python script buildtools.py to generate dependences better.

BillWhite:
There are two problems with the buildtools.py script.  Both have to do with how the script calculates include file dependences.  It currently calculates these by compiling all source files with -MM, which extracts the dependences and just lists them. The buildtool.py script then pastes these dependences into the makefile.  The two problems with this are:

1. It takes a long time to preprocess all the source files.  This is a minor annoyance.
2. If an include file changes, the makefile will not know it. So, we may compile too much, if an
    include file is deleted, or too little, if an include file is added.  This is a major problem.  We
    might get inconsistent builds.

This commit fixes both of these problems.  This is all blue on jenkins except for the Pro Junit
other tests.  I believe the failure there is environmental, and I'm running it again.

Jira Ticket Number: ENG-8988
Jira Ticket Url: https://issues.voltdb.com/browse/ENG-8988


paulmartel:
Has this been tested on mac and linux?
Do you have "manual reproducers"? examples I can try of changes to the include structure (adding, dropping headers and/or includes in sources or headers) under which the old scheme fails to properly rebuild and the new scheme works?
Do you have any manual "build from a clean directory" benchmarks showing the total build time (for makefile AND final build artifacts) impact of this change?
I'd like to "test drive" this change before pushing to master. I suppose we could use the geo integration branch as a sandbox for this if the team doesn't object.

BillWhite:
I've built with it on both Linux and the Mac.  This is the usual way
include file dependences have been done since GCC 3.0 was released, which
was June 18, 2001, so it's been tested pretty extensively in other projects.

You can run build.py without any arguments and see that the makefile
construction is almost instantaneous with the new way.  It takes about half
a minute on master.

I don't have any examples of places where it failed before.  I was just
working from thought experiments.

BillWhite:
FWIW, this is all blue on Jenkins.

BillWhite:
I'm going to merge this work informally into my S2 Building branch.  So, building it in jenkins is not really all that helpful.

PR #2807, opened 2015-09-10
Eng 8289 after rebase

luyangco:
This patch is for the problem that when disk I/O is busy(e.g. large transactions), client transaction per second (tps) might drop to 0 for a while when DR replica starts the sync snapshot, it recovers after the sync snapshot.

Community Branch contains
1) Don't syncing DR buffers to disk when non-truncation snapshot happens,
2) putting DR sync snapshot to the same directory of snapshot path defined in the deployment.


afrogers:
Reviewed

afrogers:
Merged externally

PR #2808, opened 2015-09-10
in matview code, use pre-existing hex decoder, disable output except …

paulmartel:
…at trace level


jxchinaren:
Looks good to me.

PR #2809, opened 2015-09-11
ENG-8987: Give an error if a procedure returns more than Short.MAX_VA…

manjujames:
…LUE number of tables.

Revert previous change that allowed up Integer.MAX_VALUE number of tables because it
broke backward compatibility for client protocol.

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -398,7 +398,16 @@ public class ProcedureRunner {
             // don't leave empty handed
             if (results == null) {
                 results = new VoltTable[0];
+            } else if (results.length > Short.MAX_VALUE) {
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -398,8 +398,17 @@ else if (m_language == Language.GROOVY) {
             // don't leave empty handed
             if (results == null) {
                 results = new VoltTable[0];
+            } else if (results.length > Short.MAX_VALUE) {
+                String statusString = "Output from procedure exceeded the limit of " + Short.MAX_VALUE + " tables";
```
akhanzode: is this same as other error? Ask Andrew if he cares. otherwise code changes look good.

manjujames:
Squash merged.

PR #2811, opened 2015-09-11
ENG-8945: Return conflicting row when attempting to insert into uniqu…

wweiss-voltdb:
…e indexes

### DIFF
--- a/src/ee/indexes/CompactingHashMultiMapIndex.h
+++ b/src/ee/indexes/CompactingHashMultiMapIndex.h
@@ -72,7 +72,7 @@ class CompactingHashMultiMapIndex : public TableIndex
         return *reinterpret_cast<MapIterator*> (cursor.m_keyIter);
     }
 
-    bool addEntryDo(const TableTuple *tuple)
+    const void* const* addEntryDo(const TableTuple *tuple)
     {
         ++m_inserts;
         return m_entries.insert(setKeyFromTuple(tuple), tuple->address());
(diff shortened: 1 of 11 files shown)


src/ee/indexes/CompactingHashMultiMapIndex.h
```diff
@@ -72,7 +72,7 @@ class CompactingHashMultiMapIndex : public TableIndex
         return *reinterpret_cast<MapIterator*> (cursor.m_keyIter);
     }
 
-    bool addEntryDo(const TableTuple *tuple)
+    const void* const* addEntryDo(const TableTuple *tuple)
```
nshi: I don't think it's a good idea to expose the internals of the index to its caller. The API of the index deals with tuples, so we should try to be consistent here. A more consistent approach is to return `MapConstIterator` here or even the `std::pair<MapConstIterator, bool>` following the STL standard. Then the caller can get the tuple using the iterator.

paulmartel: I'm going to disagree here because this class is not one of those that has recently caught on to the stl religion, so -- at least for now -- it's doing a pretty good job of hiding what's under it.

A better way to make the API more consistent with the pre-existing calls:
Change this to a void function and either add a TableTuple* or TableTuple& argument OR make the current argument a two-way non-const (not sure why it's a * not an &). I would use the return of the map insert to set the address of the TableTuple argument -- and here I don't care much whether or not we complicate THAT into an iterator that we have to milk stl-style for our void*.  If using a separate TableTuple argument from the input argument, the caller could pre-initialize the TableTuple (possibly on the stack) to be a "null tuple" with the correct table schema. The caller would test isNullTuple after returning.

If it's time to go all out and stl-ize this whole class, too, then let's do it, but dripping it in so that a new dev can't predict what's exposed via stl style vs. the original style just seems like slow torture.

nshi: I have no strong preference over whether this is STL-style or not. I was trying to point out that returning `const void* const*` is a). exposing the internals of the index to the caller, b). it's inconsistent with the rest of the existing API, and c). it's not self-describing at all. I don't think I did a good job of making these three points clear in my original comment.

Paul's proposal of taking an additional argument would also work pretty well here.

PR #2812, opened 2015-09-12
Migrate SnapshotRestore details down out of VoltSystemProcedure.

paulmartel:
A cut & paste into SnapshotRestore.java (see the matching pro changeset)
except for a ltweak to use the separate TRACE_LOG more consistently tor trace output.


luyangco:
Reviewed. It looks good to me.

PR #2813, opened 2015-09-14
Eng 8945 include conflicting row

wweiss-voltdb:


src/ee/indexes/CompactingTreeMultiMapIndex.h
```diff
@@ -63,6 +63,7 @@ class CompactingTreeMultiMapIndex : public TableIndex
 {
     typedef typename KeyValuePair::first_type KeyType;
     typedef typename KeyType::KeyComparator KeyComparator;
+    typedef typename KeyValuePair::second_type ValueType;
```
nshi: I don't see where this is used. Unintended change?

PR #2816, opened 2015-09-15
ENG-8469: Only notify Durability Listeners when the command log has c…

wweiss-voltdb:
…hanged


src/frontend/org/voltdb/iv2/SpDurabilityListener.java
```diff
@@ -122,6 +127,18 @@ public void processChecks() {
             }
             super.processChecks();
```
nshi: This is a duplicate call because `processChecks()` also calls it, which makes `processChecks()` to be the same as `checkForSyncLoggedSysProcs()`.

PR #2817, opened 2015-09-15
Created a schema for a possible application.

BillWhite:
This was my take on our conversation.  I wrote notes on the white board, but got kicked out of the room before I remembered to write my notes down, so this is from memory.

1. I renamed the point table to be taxis.  You can change it to be drivers if you want.  
1. I also added cities.  I think it's easy to add them, and I don't think it hurts anything.
1. I replicated the polygon table to be counties and states, since we wanted some nested structure.  Each has a cellid2id join table, and a couple of indices.
1. The tables ending in _map and the indices ending in _idx will together be our geospatial index, and will not be visible to the users, I think.

### DIFF
--- a/tests/test_apps/s2/taxi_ddl.sql
+++ b/tests/test_apps/s2/taxi_ddl.sql
@@ -1,13 +1,12 @@
 
-load classes procs.jar;
+-- load classes procs.jar;
 
 file -inlinebatch EOB
 
 create table taxis (
     id              bigint not null primary key,
     location        point not null,
-    cellid          bigint not null,
-    geoid           biging not null
+    cellid          bigint not null
 );
 
 create index taxis_cellid_idx on taxis(cellid);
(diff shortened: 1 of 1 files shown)


tests/test_apps/s2/taxi_ddl.sql
```diff
@@ -0,0 +1,60 @@
+
+load classes procs.jar;
+
+file -inlinebatch EOB
+
+create table taxis (
+    id              bigint not null primary key,
+    location        point not null,
+    cellid          bigint not null,
+    geoid           biging not null
```
wolffcm: What's the geoid column for?

BillWhite: It's a combination of  the state and county FIPs id.
I'm going to push this again.  I think this is not clearly thought out.

I'm thinking states and counties need to be the in the same table, with the
kind (county/state) as a column.  Otherwise we aren't able to test nesting.

BillWhite:
I took out the fips stuff, which is not helpful, I merged the counties and states.  I also
commented out the create procedure and load proc.jar lines, so that I could verify that
the syntax is right.

wolffcm:
Looks good, thanks.

PR #2818, opened 2015-09-15
Start of a new S2 Demo Program.

BillWhite:
I added a new demo program in tests/test_apps/s2 called iwdemo.  This is roughly
parallel to s2demo.  I left s2demo alone to use it as a pattern.

If you run taxi_run.sh it should run the demo and propulate the cities table with
about 30000 cities.  Don't look at all of them, or you'll be all day at it.

There's the beginnings of creating tables for counties, but I don't have it all
done yet.  Since the routines to populate the counties tables are never called,
it's ok that they are broken for the time being.


wolffcm:
Looks good to me.  counties_geometry.csv is more than a million lines.  How does git handle files as large as this?

BillWhite:
It looks like a large file to github is one which is greater than 50MB.
The largest of these is 33MB, so I think it's not really a large file.

Maybe we should put it on dropbox, though.

PR #2820, opened 2015-09-16
Add a TaxiManager class

wolffcm:
Create a taxi manager class that populates the taxis table, and periodically updates locations of taxis.


tests/test_apps/s2/client/iwdemo/TaxiManager.java
```diff
@@ -0,0 +1,165 @@
...
+                        m_random.nextInt((int)NUM_CITIES)).getResults()[0];
+                vt.advanceRow();
+                PointType pt = vt.getPoint(0);
+                System.out.println("  " + vt.getString(1));
+
+                // Allocate 8 cabs to this point
+                for (int j = 0; j < NUM_DIRECTIONS; ++j) {
+                    m_taxiLocations[taxiId] = pt;
+                    S2LatLng ll = S2LatLng.fromDegrees(pt.getLatitude(), pt.getLongitude());
+                    S2CellId cell = S2CellId.fromLatLng(ll);
```
BillWhite: I wonder if this S2 stuff should be in the shared procedure?  I think it makes a better demo if the client doesn't know much about S2.  I use S2LatLng in the Benchmark class, but that's pretty trivial, and could be a Pair<Double, Double> for all that.  Extracting a CellID in the client seems wrong.

tests/test_apps/s2/client/iwdemo/TaxiManager.java
```diff
@@ -0,0 +1,165 @@
...
+
+        S2LatLng ll = S2LatLng.fromDegrees(
+                origPt.getLatitude(),
+                origPt.getLongitude());
+        S2CellId cell = S2CellId.fromLatLng(ll);
+
+        // Go up to a coarser level of granularity,
+        // choose an adjacent cell.
+        S2CellId parent = cell.parent(S2_UPDATE_LEVEL);
+        List<S2CellId> parentNeighbors = new ArrayList<>();
```
BillWhite: Here too.

BillWhite:
Seems ok to me.  The comment I had should not stop committing.

tests/test_apps/s2/client/iwdemo/TaxiManager.java
wolffcm: Yeah.  I was thinking of the DB as receiving only updated lat/lng pairs from each taxi, kind of like a GPS-enabled taxi might send in real life.  So it's up to the application to figure out what the updated lat/lng should be.  S2 seemed to be an easy way to do this, but I could have just incremented or decremented lat/lng as well.

wolffcm:
I have a separate branch that includes these changes and a lot more.  I'll close this and create a new PR.

PR #2821, opened 2015-09-17
Lots of changes to IW branch

wolffcm:
Added a taxi manager class that creates taxis and moves them around.

Added a periodic task that shows places near taxis every second.

Added a web demo using Google Maps.


BillWhite:
Seems right to me.

PR #2822, opened 2015-09-17
Added county and state boundaries.

BillWhite:
The county and state boundaries are added.  I don't actually know if they work, but the data I can see seems to be right.

Note that this is on top of Chris' lastest commit.


BillWhite:
Sorry, I made a mistake.  Wait an instant and I'll redo this.

PR #2823, opened 2015-09-17
Insert state and county metadata and boundaries.

BillWhite:
Really insert county and state metadata and boundaries.


wolffcm:
Looks good.

PR #2826, opened 2015-09-17
Add stored procedure calls to marker listeners

wolffcm:


BillWhite:
Seems right to me.  I will delete the extra stuff I added.

PR #2827, opened 2015-09-18
Fix InsertRegion stored procedure

wolffcm:
Add county/state to map marker info windows.
Choose more evenly dispersed cities.


BillWhite:
Looks right.

PR #2829, opened 2015-09-18
ENG-9037, fix a bug that causes non-nulled default value of hidden ti…

luyangco:
…mestamp column.

Change-Id: I0c4d737d4c7c8013f5556cbf37effb1d9bd4ad12


nshi:
Reviewed. Pending test verification.

PR #2831, opened 2015-09-21
Eng 3147 varbinary index

jxchinaren:


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1824,28 +1824,20 @@ void addIndexToCatalog(Database db, Table table, VoltXMLElement node, Map<String
         if (exprs == null) {
             for (int i = 0; i < colNames.length; i++) {
                 VoltType colType = VoltType.get((byte)columns[i].getType());
-                if (colType == VoltType.DECIMAL || colType == VoltType.FLOAT || colType == VoltType.STRING) {
+                if (colType == VoltType.DECIMAL || colType == VoltType.FLOAT ||
```
wolffcm: Perhaps this test could be implemented as a static method---`VoltType.isIntType()`?

For the POINT type work I wanted to make POINT not indexable (for now), so I added a static method `VoltType.isIndexableType`, we'll need to update this method when this work is merged to the geo topic branch.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -842,7 +843,10 @@ static protected void checkQueryPlan(Client client, String query, String...patte
         String vtStr = vt.toString();
 
         for (String pattern : patterns) {
-            assertTrue(vtStr.contains(pattern));
+            if (! vtStr.contains(pattern)) {
+                System.err.println(vtStr);
+                fail("The above explain plan is expected to contain pattern: " + pattern);
```
wolffcm: Why not just put the explained plan into the "fail" message itself?  That's where I usually look for diagnostics.

src/frontend/org/voltdb/compiler/DDLCompiler.java
paulmartel: @wolffcm I don't think you meant _static_ method? Also, since int is supported as a common shorthand for INTEGER, I'd keep getting confused about whether isIntType included BIGINT, TINYINT, etc. maybe VoltType.isAnIntegerType would be a better name?

wolffcm: I was thinking that a static method would be easiest.  Would there be a way to encapsulate this logic that didn't involve touching each member of the VoltType enum?  Maybe there's a different way to do it that I'm not seeing.

I *do* like your name better, it definitely reduces the ambiguity.

wolffcm: I guess you could have a VoltType member method that just does an `instanceof` check for the 4 integer types, that would be cleaner.  Java enum classes are still new to me.

tests/testprocs/org/voltdb_testprocs/regressionsuites/indexes/indexes-ddl.sql
```diff
@@ -117,3 +117,12 @@ CREATE TABLE tableY (
 PARTITION TABLE tableY ON COLUMN keyA;
 
 CREATE INDEX idx_y_keyI ON tableY(keyH,keyI);
+
+-- tree index on varbinary type
+CREATE TABLE blobTableTree (
+    id INTEGER NOT NULL,
+    blob2 VARBINARY(2),
```
wolffcm: When I think of a "blob", I think of "Binary Large OBject", which in some databases is a different data type that VARBINARY and can be really big, as in several MB or GB.  So using "blob" to refer to VARBINARY(2) seems wrong to me.

tests/frontend/org/voltdb/regressionsuites/TestIndexesSuite.java
```diff
@@ -1230,7 +1332,7 @@ public TestIndexesSuite(String name) {
         // end of easy-to-disable code section */
 
         /*/ CONFIG #3: IPC -- keep this normally disabled with / * vs. //
-        config = new LocalCluster("testindexes-threesite.jar", 1, 1, 0, BackendTarget.NATIVE_EE_IPC);
+        config = new LocalCluster("testindexes-threesite.jar", 1, 1, 0, BackendTarget.NATIVE_EE_JNI);
```
wolffcm: Should probably change this back to IPC before you merge.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -84,7 +85,8 @@ def normalize_values(tuples, columns):
     # I assume t is a voltdbclient.VoltTable.
     if hasattr(tuples, "__iter__"):
         for i in xrange(len(tuples)):
-            if hasattr(tuples[i], "__iter__"):
+            # varbinary is array.array type and has __iter__ defined, but should a single value to be compareds
```
wolffcm: I'm having trouble parsing this comment.  Maybe you mean, "...but should be considered as a single value for comparison"?

tests/scripts/examples/sql_coverage/index-varbinary-schema.py
```diff
@@ -0,0 +1,51 @@
...
+                    ("B", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("C", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("D", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("E", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("F", FastSerializer.VOLTTYPE_VARBINARY)
+                    ),
+        "partitions": (),
+        "indexes": ()
+    }
+#,
```
wolffcm: Delete this code if it's not used?

tests/scripts/examples/sql_coverage/index-varbinary-schema.py
```diff
@@ -0,0 +1,51 @@
...
+                    ("ID", FastSerializer.VOLTTYPE_INTEGER),
+                    ("A", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("B", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("C", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("D", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("E", FastSerializer.VOLTTYPE_VARBINARY),
+                    ("F", FastSerializer.VOLTTYPE_VARBINARY)
+                    ),
+        "partitions": (),
+        "indexes": ()
```
wolffcm: I'm a little fuzzy on sql_coverage in general.  I guess this file is used to tell the query generator about which columns are in which tables, and which columns are indexed and used as partition keys?  Why not include the indexed columns here?

tests/scripts/examples/sql_coverage/index-DDL.sql
```diff
@@ -81,3 +81,17 @@ CREATE TABLE P_SCAN2 (
...
+	B VARBINARY(256),
+	C VARBINARY(50),
+    D VARBINARY(256),
+    E VARBINARY(256),
+    F VARBINARY(256)
+);
+CREATE INDEX BLOB_TABLE_TREE_1 ON BLOB_TABLE (A);
+CREATE INDEX BLOB_TABLE_TREE_2 ON BLOB_TABLE (B, C);
+--- Column D,E,F are reserved for HASH index testing in future
+
```
wolffcm: Do we need coverage for indexes on varbinary columns in partitioned tables?

wolffcm:
Finished review round.

wolffcm:
Regarding using "blob" and "varbinary" interchangeably throughout your changes, I think you really mean "varbinary" and not "blob."  It's possible this is just me though.

tests/testprocs/org/voltdb_testprocs/regressionsuites/indexes/indexes-ddl.sql
jxchinaren: When I read the code, I learn "BLOB" is just our VARBINARY, like the junit test "TestBlobType" other than "TestVarbinaryType". 

I can change the test table name as "varbinaryTableTree", does this name sounds better?

tests/scripts/examples/sql_coverage/StandardNormalizer.py
jxchinaren: Yes, that's what I mean, I should update this comment clearly. 

We used to rely on this "if hasattr(tuples[i], "__iter__")" check to make sure that it's a single value for a column to be compared. However, the python client returns an Array.array for varbinary type, which will break that IF check.

tests/scripts/examples/sql_coverage/index-varbinary-schema.py
jxchinaren: "this file is used to tell the query generator about a)which columns are in which tables, and b)which columns are indexed and c)used as partition keys? "

I think the a) is true, b) and c) are not true. Sadly, the indexes and partitions sections are not used. I guess I should remove them, other than simply coping&paste.

tests/testprocs/org/voltdb_testprocs/regressionsuites/indexes/indexes-ddl.sql
wolffcm: Yes, that sounds better to me.

src/frontend/org/voltdb/compiler/DDLCompiler.java
jxchinaren: I had the same though as you guys to refactor these types check into a function for VoltType, and I also found some similar functions there already, like: isExactNumeric(), isInteger(), isNumber()... None of them match my needs here. 

So I probably will add a new variant now: "isAnIntegerType()"...

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
jxchinaren: I though it's easier to look at the console when it failed, but that's just me. 

I will add the explain plan into the failure message with new line in between for better reading.

tests/scripts/examples/sql_coverage/index-DDL.sql
jxchinaren: I do not think is there any difference for the indexes with partition table, but I do think we should test partition table on VARBINARY column with indexes. 

I will simple copy the Replicated table cases then.

jxchinaren:
@wolffcm I added two new commits to resolve your comments, thanks for reviewing that.

wolffcm:
Looks good to me, thanks for making those changes.

jxchinaren:
Will close it and squash merge then.

PR #2832, opened 2015-09-21
[For review only] Materialized views on export tables

wolffcm:
I'm creating this PR to keep track of review comments in the materialized views on export tables work that Anish did during last innovation week.  The ticket for this code review task is ENG-8911.

This work isn't to be merged until we've completed the remaining tasks mentioned in ENG-8506.


examples/voter/ddl.sql
```diff
@@ -66,6 +66,7 @@ AS
 ;
 
 END_OF_BATCH
+export table votes to stream anish2;
```
wolffcm: This `export table` table statement could be inside the batch.

However, it seems like we'll probably want an all-new new example of export tables with materialized views, since we probably don't want to make the voter example have `votes` as an export table.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -933,11 +935,50 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp)
...
             m_catalogDelegates[catalogTable->path()] = tcd;
             m_delegatesByName[tcd->getTable()->name()] = tcd;
+//            std::cout << "Creating export table " << tcd->getTable()->name() << "\n";
+//            std::cout.flush();
 
             // set export info on the new table
             if (tcd->exportEnabled()) {
                 tcd->getTable()->setSignatureAndGeneration(catalogTable->signature(), timestamp);
                 m_exportingTables[catalogTable->signature()] = tcd->getTable();
+                StreamedTable *streamedtable = dynamic_cast<StreamedTable *>(tcd->getTable());
```
wolffcm: This and the next set of changes immediately below are nearly identical.  We should merge this into a common function.

Also, we should check to see if there's code here in common with persistent tables that could also be re-used.

With this change, we're removing another thing that makes persistent and streamed tables different, so we should consider pushing some code into the base class, or otherwise refactoring to reflect this new commonality, where it makes sense to do so.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1125,6 +1202,8 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp)
                 }
                 // This is not a leak -- the view metadata is self-installing into the new table.
                 // Also, it guards its targetTable from accidental deletion with a refcount bump.
+//    std::cout << "Creating reg view 2\n";
```
wolffcm: We should remove this commented-out instrumentation.

src/ee/storage/streamedtable.cpp
```diff
@@ -115,6 +236,9 @@ bool StreamedTable::deleteTuple(TableTuple &tuple, bool fallible)
 {
     size_t mark = 0;
     if (m_wrapper) {
+        for (int i = 0; i < m_views.size(); i++) {
+            m_views[i]->processTupleDelete(tuple, false);
```
wolffcm: I thought that deleting from export tables wasn't allowed?  I'm obviously missing something here.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -1690,11 +1690,11 @@ void addExportTableToConnector(final String targetName, final String tableName,
             throw new VoltCompilerException("While configuring export, table " + tableName + " was not present in " +
             "the catalog.");
         }
-        if (CatalogUtil.isTableMaterializeViewSource(catdb, tableref)) {
-            compilerLog.error("While configuring export, table " + tableName + " is a source table " +
-                    "for a materialized view. Export only tables do not support views.");
-            throw new VoltCompilerException("Export table configured with materialized view.");
-        }
+//        if (CatalogUtil.isTableMaterializeViewSource(catdb, tableref)) {
```
wolffcm: We can delete this unused code.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1369,6 +1452,31 @@ void VoltDBEngine::initMaterializedViewsAndLimitDeletePlans() {
             }
         }
     }
+
+    //for streamed tables;
+    BOOST_FOREACH (LabeledTable labeledTable, m_database->tables()) {
```
wolffcm: This loop is very similar to the loop above it for persistent tables.  Perhaps they could be merged?

src/ee/storage/streamedtable.cpp
wolffcm: We must delete exported rows if the transaction is rolled back.

wolffcm:
Closing this pull request, but will leave branch around so we can circle back to these comments later.

PR #2837, opened 2015-09-22
ENG-9022: Track both MpUniqueId and SpUniqueId in InvocationBuffers

wweiss-voltdb:


src/ee/storage/DRTupleStream.cpp
```diff
@@ -380,51 +382,66 @@ void DRTupleStream::beginTransaction(int64_t sequenceNumber, int64_t uniqueId) {
...
-                              m_currBlock->remaining());
-     io.writeByte(DR_VERSION);
-     io.writeByte(static_cast<int8_t>(DR_RECORD_END_TXN));
-     io.writeLong(m_openSequenceNumber);
-     uint32_t crc = vdbcrc::crc32cInit();
-     crc = vdbcrc::crc32c( crc, m_currBlock->mutableDataPtr(), END_RECORD_SIZE - 4);
-     crc = vdbcrc::crc32cFinish(crc);
-     io.writeInt(crc);
-     m_currBlock->consumed(io.position());
+    if (txnId == spHandle) {
```
afrogers: Could just inspect the uniqueId directly; it's got the partitionId right there in its lowest 14 bits. `uniqueId & 0x3fff == 16383` or whatnot. Seems a little much to pass these just for this check.

src/ee/storage/TupleStreamBase.cpp
```diff
@@ -241,7 +241,7 @@ void TupleStreamBase::rollbackTo(size_t mark)
         if (m_currBlock == NULL) {
             extendBufferChain(m_defaultCapacity);
         }
-        m_currBlock->recordCompletedTxnForDR(m_committedSequenceNumber, m_committedUniqueId);
+        m_currBlock->recordCompletedSequenceNumForDR(m_committedSequenceNumber);
```
afrogers: Why aren't we rolling back the uniqueIds here? It looks like you're just circumventing this by only setting them in endTxn? We do the same thing with the sequence number, though, right?

I guess the question I'm asking is: does this line even make sense? Or can we remove it altogether?

src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -0,0 +1,83 @@
...
+    public static int getClusterIdFromDRId(long drId) {
+        return (int)((drId >> 55) & MAX_CLUSTER_ID);
+    }
+
+    public static long getSequenceNumberFromDRId(long drId) {
+        return drId & MAX_SEQUENCE_NUMBER;
+    }
+
+    public static long getSentinelOrSeqNumFromDRId(long drId) {
+        return (drId<0?drId:(drId & MAX_SEQUENCE_NUMBER));
```
afrogers: (Warning: style complaint) Somehow the lack of whitespace just makes me cringe.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -156,14 +158,15 @@ public static synchronized long pushDRBuffer(
...
+            long lastSpUniqueId,
+            long lastMpUniqueId,
             ByteBuffer buf) {
         if (log.isTraceEnabled()) {
             log.trace("Received DR buffer size " + buf.remaining());
             AtomicLong haveOpenTransaction = haveOpenTransactionLocal.get();
             buf.order(ByteOrder.LITTLE_ENDIAN);
-            //Magic header space for Java for implementing zero copy stuff
-            buf.position(8 /* stream block header */ + 69 /* txn metadata padding */);
+            //Magic header space for Java for implementing zero copy stuff (match with InvocationBuffer.MAGIC_DR_TRANSACTION_PADDING)
```
afrogers: Not that it really matters, but would probably rather have this comment in the pro class rather than referring to the pro class here

src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -0,0 +1,83 @@
...
+
+import com.google_voltpatches.common.base.Preconditions;
+
+public class DRLogSegmentId {
+    public static final short MAX_CLUSTER_ID = (1 << 8) - 1;
+    public static final long MAX_SEQUENCE_NUMBER = (1L << 55) - 1L;
+
+    public static final long UNINITIALIZED_SEQUENCE_NUMBER = -1L;
+
+    public long drId;
```
afrogers: I'm not sure I'm entirely comfortable with this class being built entirely on state that is both public and mutable...

src/ee/storage/DRTupleStream.h
```diff
@@ -88,11 +88,15 @@ class DRTupleStream : public voltdb::TupleStreamBase {
     void beginTransaction(int64_t sequenceNumber, int64_t uniqueId);
     // If a transaction didn't generate any binary log data, calling this
     // would be a no-op because it was never begun.
-    void endTransaction();
+    void endTransaction(int64_t txnId, int64_t spHandle, int64_t uniqueId);
 
     bool checkOpenTransaction(StreamBlock *sb, size_t minLength, size_t& blockSize, size_t& uso);
 
-    std::pair<int64_t, int64_t> getLastCommittedSequenceNumberAndUniqueId() { return std::pair<int64_t, int64_t>(m_committedSequenceNumber, m_committedUniqueId); }
+    void getLastCommittedSequenceNumberAndUniqueIds(int64_t &seqNum, int64_t &spUniqueId, int64_t &mpUniqueId) {
```
afrogers: Maybe worth encapsulating these out params in a struct?

PR #2838, opened 2015-09-23
Eng 8859 geo merge

BillWhite:
This is mostly a merge from master.  However, it has a corresponding pro branch.

Also, the build tools are updated to use fast dependency generation.  The compiler creates makefile-compatible files which the makefile includes.  If the files do not exist it is not an error, as the inclusion is done with "-include".  Making the makefile used to take minutes.  Now it takes less than a second.

I also had to work around some build issues with the most recent Xcode version.

The S2 library is not in this commit.  This is just a merge and fixing the aftermath.

Note that there is no pull request for the master merge, as there is no ENG-8859-geospatial pro branch.


wolffcm:
I was able to build this branch on both Linux and OS X (with the newest version of Xcode).  Looks very nice.  One minor thing: in the generated makefile, some paths appear with double slashes in them, as in `static_objects//voltdbjni.d`.

Also, it'd be nice to include some comments that describe how dependency generation works now.  Maybe this method is pretty standard, but it was new to me.

I definitely appreciate the improved build times!  I'm sure others will too.

wolffcm:
@BillWhite I don't want my comments to hold up this merge, please go ahead and merge it as soon as you're ready.

BillWhite:
I wasn't holding up for your comments.  I was holding up because I wanted
to get to a stopping place in building S2.  I'll get it merged before
lunch, I think.

PR #2842, opened 2015-09-24
ENG--9078, change the conflict resolution API.

luyangco:
Change-Id: I09b6b5d47ef66c25b4176e7062275ccf26e229c5

### DIFF
--- a/src/frontend/org/voltdb/PartitionDRGateway.java
+++ b/src/frontend/org/voltdb/PartitionDRGateway.java
@@ -59,6 +59,26 @@ public class PartitionDRGateway implements DurableUniqueIdListener {
         }
     }
 
+    // Keep sync with EE DRConflictType at types.h
+    public static enum DRConflictType {
+        DR_CONFLICT_UNIQUE_CONSTRIANT_VIOLATION,
+        DR_CONFLICT_MISSING_TUPLE,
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -59,6 +59,26 @@ public static DRRecordType valueOf(int ordinal) {
...
+        public static final ImmutableMap<Integer, DRConflictType> conversion;
+        static {
+            ImmutableMap.Builder<Integer, DRConflictType> b = ImmutableMap.builder();
+            for (DRConflictType t : DRConflictType.values()) {
+                b.put(t.ordinal(), t);
+            }
+            conversion = b.build();
+        }
+
+        public static DRConflictType valueOf(int ordinal) {
```
nshi: You don't need to build the conversion map and write this method. You can just use `DRConflictType.values()[ordinal]` to get the type.

luyangco: Should `DRRecordType` also use `DRRecordType.values()[ordinal]` to get the type at PartitionDRGateway.java:180?

nshi: Yes.

luyangco: Fixed.

PR #2844, opened 2015-09-25
Eng 9042 query timeout

jxchinaren:


src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -1804,6 +1807,21 @@ final ClientResponseImpl handleRead(ByteBuffer buf, ClientInputHandler handler,
...
 
+        //Check individual query timeout value settings with privilege
+        int batchTimeout = task.getBatchTimeout();
+        if (BatchTimeoutOverrideType.isUserSetTimeout(batchTimeout)) {
+            if (! user.hasPermission(Permission.ADMIN)) {
+                int systemTimeout = catalogContext.cluster.getDeployment().
+                        get("deployment").getSystemsettings().get("systemsettings").getQuerytimeout();
+                if (systemTimeout != ExecutionEngine.NO_BATCH_TIMEOUT_VALUE &&
+                        (batchTimeout > systemTimeout || batchTimeout == ExecutionEngine.NO_BATCH_TIMEOUT_VALUE)) {
+                    log.info("The attempted individual query timeout " + batchTimeout + " override was ignored "
```
wolffcm: Is it possible that this message could be written to the log very, very frequently?  If so, is there a mechanism in place to avoid lots and lots of redundant messages and filling up all available disk space?

wolffcm: Also, is this logic on the client machine or the server machine?  If it's on the client machine, doing the guard here could be a security risk.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1027,8 +1027,8 @@ private static void printUsage(int exitCode)
         + "  In interactive mode, a value of \"true\" discards any unprocessed input\n"
         + "  and returns to the command prompt. Default: true.\n"
         + "\n"
-//        + "[--query-timeout=millisecond_number]\n"
-//        + "  Read only queries timeout setting. Default: no timeout.\n"
+        + "[--query-timeout=millisecond_number]\n"
+        + "  Read only queries timeout setting. Default: no timeout.\n"
```
wolffcm: This could be a little more verbose?  Maybe, "Read-only queries that take longer than this number of milliseconds will abort.  Default: no timeout."

tests/frontend/org/voltdb/regressionsuites/TestQueryTimeout.java
```diff
@@ -95,6 +97,8 @@ public void testReplicatedProcTimeout() throws IOException, ProcCallException, I
         }
         System.out.println("test replicated table procedures timeout...");
 
+        m_username = "userWithAllProc";
```
wolffcm: Since these are members in the superclass, is it possible that setting these in tests might cause the new values to "leak" out into other tests?  This could cause strange failures given that the test methods may be run in any order.

Another approach might be to pass a username and password to an overloaded `getClient` method which would temporarily change these members to get a connected client, and always set them back to their original values.

tests/frontend/org/voltdb/regressionsuites/TestQueryTimeout.java
```diff
@@ -207,19 +212,61 @@ private void checkTimeoutIncreasedProcSucceed(boolean sync, Client client, Strin
...
     }
 
-    private void checkTimeoutDecreaseProcFailed(boolean sync, Client client, String procName, Object...params)
+    private void checkTimeoutIncreasedProcFailed(boolean sync, Client client, String procName, Object...params)
             throws IOException, ProcCallException, InterruptedException {
         checkDeploymentPropertyValue(client, "querytimeout", Integer.toString(TIMEOUT));
 
         try {
             client.callProcedure(procName, params);
+            fail(procName + " is supposed to timed out, but not actually!");
```
wolffcm: What do you mean by "but not actually" here?  It seems like you *do* want it to time out to pass the test?

tests/frontend/org/voltdb/regressionsuites/TestQueryTimeout.java
```diff
@@ -207,19 +212,61 @@ private void checkTimeoutIncreasedProcSucceed(boolean sync, Client client, Strin
...
         checkDeploymentPropertyValue(client, "querytimeout", Integer.toString(TIMEOUT));
 
         // increase the individual timeout value in order to succeed running this long procedure
+        // However, for non-admin user, timeout value can not override system timeout value.
+        if (sync) {
+            try {
+                client.callProcedureWithTimeout(TIMEOUT*50, procName, params);
+                fail(procName + " is supposed to be timed out!");
+            } catch(Exception ex) {
+                assertTrue(ex.getMessage().contains(ERRORMSG));
```
wolffcm: I guess here is where you would check for a message like this "the attempted timeout override ... was ignored"?  But you said that you haven't implemented that message (yet) during stand up.

wolffcm:
That's all my comments.  I didn't see any showstoppers, I just had the few questions.

tests/frontend/org/voltdb/regressionsuites/TestQueryTimeout.java
jxchinaren: Yes, printing our this error message other than in the log needs more adding extra information into procedure invocation which may change our client procedure invocation protocol. There may be tricky ways to do that, but I am not sure if that's worth doing.

tests/frontend/org/voltdb/regressionsuites/TestQueryTimeout.java
jxchinaren: "supposed to timed out, but not eventually".

src/frontend/org/voltdb/ClientInterface.java
jxchinaren: This is on server side. All the other user privilege check are checked just before this block. 

If there is a user keep calling this procedure with individual timeout larger than system timeout asynchronously, I guess the info message will be printed out frequently. Although it's a rare case, but I think the server has the client back pressure to prevent it happening too frequently eventually (however this is just my guess).  Does it make sense?

wolffcm: I'm not sure how back pressure works.  I know that we have ways of preventing the log being flooded with a certain message.  An engineer on the Elastic or ECO might be able to clarify?

jxchinaren: I asked Ning and get the logger information. It's "RateLimitedLogger", and I will use it instead. 

Print it once for every 60 seconds, is that OK?

wolffcm: Sure, that works.  Thanks for looking into that.

tests/frontend/org/voltdb/regressionsuites/TestQueryTimeout.java
jxchinaren: That's why every tests in this suites has to include the username and password at the beginning when the security is turned on like this. 

If not setting in this way, the "getClient()" will use the default client which has little privilege with call procedure.

jxchinaren:
@wolffcm I added a new commit to resolve your comments, thanks for reviewing this pull request.

PR #2858, opened 2015-09-30
ENG-8712: Update tuple streams to work with batch rollback

afrogers:


src/ee/storage/TupleStreamBase.cpp
```diff
@@ -243,9 +245,11 @@ void TupleStreamBase::rollbackTo(size_t mark)
         }
         m_currBlock->recordCompletedTxnForDR(m_committedSequenceNumber, m_committedUniqueId);
```
nshi: I'm curious why this is only in the branch where we rollback more than one blocks. In fact, it's not possible to rollback more than one block for DR.

src/ee/storage/streamedtable.cpp
```diff
@@ -161,7 +161,7 @@ void StreamedTable::setSignatureAndGeneration(std::string signature, int64_t gen
 void StreamedTable::undo(size_t mark)
 {
     if (m_wrapper) {
-        m_wrapper->rollbackTo(mark);
+        m_wrapper->rollbackTo(mark, 0);
```
nshi: Why not use `SIZE_MAX` so that it's consistent with the other place.

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -773,6 +773,29 @@ TEST_F(DRBinaryLogTest, UpdateWithUniqueIndex) {
...
+    m_undoLog.undo(m_undoToken + 1);
+
+    endTxn(true);
+
+    flushAndApply(99);
+
+    EXPECT_EQ(1, m_tableReplica->activeTupleCount());
+    TableTuple tuple = m_tableReplica->lookupTupleByValues(source_tuple);
+    ASSERT_FALSE(tuple.isNullTuple());
+}
```
nshi: It'll be more interesting to see a test case that commits a txn and tries to partial rollback the next txn.

nshi:
Review completed.

PR #2860, opened 2015-10-01
use a smaller length csvStr name instead

jxchinaren:


jxchinaren:
As this is just a test name change, I will just merge it now.

PR #2868, opened 2015-10-02
Introduce the Google S2 library to the EE.

BillWhite:
Most of the files in this changeset are the Google Test, Google Mock and Google S2 libraries.
The bulk of the changes to VoltDB files are in build.py and buildtools.py.  There are also some
small changes in c++ files, but only a handful, and not significant changes.

1. Added the sources to the Google S2 library to the EE.
2. Made build.py and buildtools.py generate makefiles that call cmake and then make.
3. Removed "using namespace std" in voltdbipc.cpp.  It causes misinterpretation of
   the symbol "bind", and is unnecessary anyway.
4. Add a phony call to an S2 function in geofunctions.cpp, to verify linking.
5. The S2 include files are installed into obj/${build}/include/s2geo.  The
   library is installed into obj/${build}/lib/libs2geo.a.
6. Added the sources to google test and google mock.  The first is needed
   to make the s2 tests.


BillWhite:
I created ENG-9178 to help me remember to update the README.thirdparty file.

build.py
```diff
@@ -53,22 +52,51 @@
 if CTX.compilerName() == 'gcc':
     CTX.CPPFLAGS += " -pthread"
     CTX.LDFLAGS += " -rdynamic"
+    if (CTX.compilerMajorVersion() >= 4):
+        CTX.CPPFLAGS += " -Wno-deprecated-declarations  -Wno-unknown-pragmas"
```
wolffcm: You must be suppressing these warnings because S2 or Google Test builds fail with fail-on-warnings?

buildtools.py
```diff
@@ -211,10 +222,20 @@ def buildMakefile(CTX):
     makefile.write("JNILIBFLAGS += %s\n" % (JNILIBFLAGS))
     makefile.write("JNIBINFLAGS += %s\n" % (JNIBINFLAGS))
     makefile.write("JNIEXT = %s\n" % (JNIEXT))
-    makefile.write("SRC = ../../%s\n" % (INPUT_PREFIX))
-    makefile.write("THIRD_PARTY_SRC = ../../%s\n" % (THIRD_PARTY_INPUT_PREFIX))
     makefile.write("NM = %s\n" % (NM))
     makefile.write("NMFLAGS = %s\n" % (NMFLAGS))
+    makefile.write("#\n# Capture the name of the root directory, so that we can give\n")
```
wolffcm: I don't think we ever concerned ourselves with absolute paths before.  I guess this must be needed for cmake to do its work?

buildtools.py
```diff
@@ -17,6 +17,8 @@ def __init__(self, args):
         self.OUTPUT_PREFIX = ""
         self.TEST_PREFIX = ""
         self.INPUT = {}
+        self.SRC_INCLUDE_DIRS = []
+        self.OBJ_INCLUDE_DIRS = []
```
wolffcm: Why the "OBJ_" prefix for this variable?  I can see it's for S2, so maybe "THIRD_PARTY_" or even "S2_" would be more appropriate?

third_party/cpp/google-s2-geometry/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "googletest"]
```
wolffcm: I'm curious, what does this file do?

wolffcm:
I can't comment on the source line because it's not one of the files shown in this PR, but in `thirdparty/cpp/googletest/README.md`, the top of this document links to the build status in a separate repo located on travis-ci.org.  This doesn't seem relevant to us, maybe remove it?

wolffcm:
I'm done reviewing this.  I had some questions but nothing serious.  I'm okay with you merging this to our topic branch when you're ready.

build.py
BillWhite: Yes.  The -Wno-deprecated-declarations is there because on the MAC the system openssl headers are deprecated.  At one time I set things up to use homebrew ssl headers anyway on the MAC, so maybe it's no longer required.  When we eliminate openssl it will no longer be required in any case.  The -Wno-unknown-pragmas is there because there is a small routine in geospatial.cpp which calls S2.  It doesn't really do anything with it, it just demonstrates that the linking works.  But I wanted to disable the unused-variable warning in that function only, so I used a pragma.  On the MAC this is ok, but gcc complains when it sees pragmata it does not understand.  So I needed to tell gcc it was ok.

buildtools.py
BillWhite: I wanted to install the S2 headers someplace.  It makes more sense to install them in obj/include, but I need to name that.  It's more general than S2, so I made two directory variables.

buildtools.py
BillWhite: This is actually misleading.  I used to use absolute paths, but I changed back to relative paths.  I don't really like it, but it seems like it works.  I'll remove the comment.

third_party/cpp/google-s2-geometry/.gitmodules
BillWhite: I didn't see this.  I could have included google test as a submodule of google-s2-geometry.  Then git submodule init pulls the source from github and we would not have to keep it in our repository.  People range from very negative to mildly negative about submodules in git.  I personally think it's better to just put it into the VDB repository the way I have and update it
manually from time to time, as time goes by.

I'll delete this file.  I don't think it's helping anything, and causes confusion.

PR #2870, opened 2015-10-05
Update junit tests to be order deterministic so that it produces result set in same order

hhakimi53:
Update "is distinct from" TestComparisonOperatorsSuite junit tests to be order deterministic so that it produces result set in same deterministic order.

Verified after the changes "TestComparisonOperatorsSuite" junit test passes with release, debug and memcheck builds.


BillWhite:
I have two comments.

1. Is ID a primary key, or some other kind of unique key?

1. Tell me if I'm wrong, but I thought that IS NOT DISTINCT FROM is like equality, but stronger than equality, in that NULL = NULL is NULL and not TRUE, but NULL IS NOT DISTINCT FROM NULL is TRUE and not NULL.

    There is code in the planner which remembers which expressions are declared equal by where clauses or inner joins.  It might be a good idea to add IS NOT DISTINCT FROM to this mechanism.

hhakimi53:
Thanks Bill for looking at it. 

1. Yes, ID is primary key.
2. True. 

Sure, we can add that if it's not covered.

PR #2872, opened 2015-10-05
Update result set of the query in junit to be order deterministic

hhakimi53:
Updated result set of a query to be order deterministic. 

Testing:
Verified the junit TestSubQueriesSuite passes with memcheck, debug and release builds.


paulmartel:
Looks good.

PR #2873, opened 2015-10-05
GEOGRPAHY type support

wolffcm:
Created a GeographyValue class in Java for Java client access (we can change the name later if we want).

Created a Geography class in the EE with helper classes for iterating over the parts of a polygon.

Added a `polygonfromtext` scalar function to the EE for creating polygons.

Note that I had to implement WKT to polygon parsing twice, once in the EE and once in Java.


src/ee/common/Geography.hpp
```diff
@@ -0,0 +1,377 @@
...
+        size_t m_numRemainingLoops;
+    };
+
+    LoopContainer(const char* data)
+        : m_data(data)
+    {
+    }
+
+    // return the number of loops
+    // (using size_t here to follow convention of std containers)
```
paulmartel: I think that std containers actually use std:size_t which is NOT strictly speaking identical to the unqualified size_t used here.

src/ee/common/Geography.hpp
```diff
@@ -0,0 +1,377 @@
...
+                 return VALUE_COMPARE_LESSTHAN;
+             }
+             else if (loop.numVertices() > (*rhsLoopIt).numVertices()) {
+                 return VALUE_COMPARE_GREATERTHAN;
+             }
+             ++rhsLoopIt;
+        }
+
+        // Each loop has the same number of vertices.
+        // Compare the vertices themselves.
```
paulmartel: I'm not sure whether there are practical pros or cons to two polygons identifying as not equal when they may simply differ in some arbitrary way that does not change the covered area.
Some examples:
    two "identical" polygons might have the same vertices, just listed in reversed orders.
    they may start their loops on different points but cover all the same points in the same order.
    they may contain secondary loops that have these same differences.
    they may contain identical sets of secondary loops but listed in a different order.

We MAY want to canonicalize polygons on input, like by shifting each loop to start at its lowest-sorting point and then proceed to the lowest-sorting adjacent point and around the loop in that direction, and then sort the secondary loops by their initial point(s).

src/hsqldb19b3/org/hsqldb_voltpatches/types/VoltGeographyType.java
```diff
@@ -0,0 +1,114 @@
...
+        // Incompatible type in operation
+        throw Error.error(ErrorCode.X_42565);
+    }
+
+    @Override
+    public Object convertToType(SessionInterface session, Object a, Type type) {
+        // We come here when parsing default values.
+        if (type instanceof VoltGeographyType) {
+            // This is currently unreachable, since there's no way to
+            // create a POINT object in a DEFAULT clause given allowable syntax.
```
paulmartel: Scrub this file for cut & paste errors -- this is not the POINT type.

src/hsqldb19b3/org/hsqldb_voltpatches/types/VoltGeographyType.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public Type getAggregateType(Type other) {
+        // incompatible types in operation
+        throw Error.error(ErrorCode.X_42561);
+    }
+
+    @Override
+    public Type getCombinedType(Type other, int operation) {
+        // The 'combined type' seems to refer combining types with
```
paulmartel: refer combining --> refer to combining

tests/frontend/org/voltdb/regressionsuites/TestGeographyValue.java.orig
```diff
@@ -0,0 +1,239 @@
+package org.voltdb.regressionsuites;
```
paulmartel: TestGeographyValue.java.orig.pushed.by.mistake?

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -723,6 +726,102 @@ protected void assertTablesAreEqual(String prefix, VoltTable expectedRows, VoltT
...
+                fail("Unexpected type in expected row: " + expectedObj.getClass().getSimpleName());
+            }
+        }
+    }
+
+    /**
+     * Accept a "table" as 2-dimensional array of objects, to make it easier to write tests.
+     * Currently only handles some data types.  Feel free to add more as needed.
+     */
+    public static void assertTablesAreEqual(Object[][] expectedTable, VoltTable actualTable) {
```
paulmartel: The naming of these functions seems inconsistent and possibly counteruntuitive given the signatures.
assertTablesAreEqual with "Tables" plural sets me up a little bit to expect that it will accept two VoltTable arguments.
It also sets me up to expect the row-level method to be called assertRowsAreEqual, when it's actually assertEqualRows. Alternative suggestion "assertContentOfTable" / "assertContentOfRow"?

src/ee/common/Geography.hpp
```diff
@@ -0,0 +1,377 @@
...
+        return iterator();
+    }
+
+private:
+    const char* const m_data;
+};
+
+/**
+ * A class for representing instances of geo-spatial geographies.
+ * (Currently only polygons can be represented here.)
```
paulmartel: Half-baked idea: I'm half-heartedly considering that the various Geography representations (C++, java, wire protocol, etc.) should be "planning ahead" and define a "discriminator field". For now, it would be set always to a constant denoting POLYGON, but in the future having other possible values and possibly being the basis for a "tagged union" kind of class. How much would we gain by having that foot in the door of a more polymorphic type vs. how much does it clutter up the current work for nothing, and how hard would it be to retrofit this as we need it? I really don't know.

src/ee/common/Geography.hpp
paulmartel: And yet I am NOT suggesting that we should go so far as to make any attempt to do any normalization where the geometry would REALLY matter -- like where the holes of two polygons (un)cover exactly the same areas but in one polygon they overlap differently (assuming that's even allowed?) than in the other. So, there seems like there's a possibly subjective "practical limit" to this kind of thing.

src/ee/common/MiscUtil.h
```diff
@@ -41,6 +43,46 @@ class MiscUtil
...
+};
+
+template<>
+struct MiscUtil::FloatToSameSizeInt<float> {
+    typedef int32_t IntType;
+};
+
+template<>
+struct MiscUtil::FloatToSameSizeInt<double> {
+    typedef int64_t IntType;
```
paulmartel: I'm gonna guess that this code with commentary was cribbed from somewhere -- a bibliographical comment might prove useful.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -772,6 +773,7 @@ private void addColumnValue(Object value, VoltType columnType, int col) {
             case STRING:
                 m_buffer.putInt(NULL_STRING_INDICATOR);
                 break;
+            case GEOGRAPHY:
```
paulmartel: A minor inconsistency here -- code duplicated for STRING but shared for GEOGRAPHY and VARBINARY -- switch to having all types sharing code?

src/frontend/org/voltdb/VoltType.java
```diff
@@ -239,7 +240,17 @@
             "POINT",
             new Class[] {PointType.class},
             PointType[].class,
-            'P'); // 'p' was taken by timestamp
+            'P'), // 'p' was taken by timestamp
+
+    /**
+     * Geography type, for a geographical objects (polygons, etc)
```
paulmartel: typo "for a" --> "for"

src/ee/common/Geography.hpp
wolffcm: My limited understanding is that different geometry libraries handle this stuff differently.

S2 seems to want vertices to be specified in CCW order, where the last point is same as the first, and inner loops should be in CW order.  (IIRC, loops whose edges cross are not allowed.)  They also provide a PolygonBuilder class to help do some canonicalization.  (Bill and I used this class in Java S2 to get the state and county boundaries to load correctly.)  There's some nuance here that I don't understand yet.

I think we need a ticket to understand how much canonicalization is necessary and when it should be done.

I've heard that folks who do computational geometry figure out how similar two polygons are by computing the area of intersection divided by the area of overlap `(P1 ^ P2) / (P1 U P2)`, but operations like that seem really expensive.

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -0,0 +1,286 @@
...
+            for (PointType pt : loop) {
+                pt.flattenToBuffer(buf);
+            }
+        }
+    }
+
+    /**
+     * Deserialize a GeographyValue from a ByteBuffer.
+     * (Assumes that the 4-byte length prefix has already been deserialized, and that
+     * offset points the start of data just after the prefix.)
```
paulmartel: points --> point to

src/ee/common/Geography.hpp
wolffcm: In writing this code, I've been trying to limit the places where representation details are implemented to just two files (one in Java and one in C++).  My hope is that adding a "type of geography" prefix to the serialized representation could just as easily be done later as now.

So I'm leaning toward "retrofitting wouldn't be too bad."  But I'm open to arguments otherwise.

src/ee/common/MiscUtil.h
wolffcm: This is code that I wrote---I did't see anything like this in Boost so I put it in.  When Point goes to being doubles, I'll remove it.

tests/frontend/org/voltdb/types/TestGeographyValue.java
```diff
@@ -0,0 +1,99 @@
...
+                geog.toString());
+
+        // round trip
+        geog = new GeographyValue("POLYGON((32.305 -64.751, 25.244 -80.437, 18.476 -66.371, 32.305 -64.751), "
+                + "(28.066 -68.874, 25.361 -68.855, 28.376 -73.381, 28.066 -68.874))");
+        assertEquals("POLYGON((32.305 -64.751, 25.244 -80.437, 18.476 -66.371, 32.305 -64.751), "
+                + "(28.066 -68.874, 25.361 -68.855, 28.376 -73.381, 28.066 -68.874))",
+                geog.toString());
+
+        // PointType will change size, from (float, float) to (double, double)
```
paulmartel: There should be test cases to exercise at least all lower-case "polygon" and capitalized "Polygon" in the WKT (is WKT case sensitive? if so, then add negative cases). If lower case input is OK, I assume that toString normalizes to upper case? 

Also to test: sensitivity to extra or alternative whitespace -- extra spaces before and after punctuation, use of tabs and newlines.

paulmartel:
End of review round.

src/ee/common/Point.hpp
```diff
@@ -30,15 +31,18 @@ namespace voltdb {
  */
 class Point {
 public:
+
+    typedef float Coord;
+
     /** Constructor for a null point,
```
BillWhite: I kind of like the name CoordType rather than Coord here.  But maybe that's too picky.

src/ee/expressions/geofunctions.cpp
```diff
@@ -92,4 +118,107 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POINTFROMTEXT>() const
     return returnValue;
 }
 
+static void readLoop(const std::string &wkt,
```
BillWhite: Can the WKT have white space?  Does this skip it?

src/frontend/org/voltdb/ParameterSet.java
```diff
@@ -153,6 +154,14 @@ public static ParameterSet fromArrayNoCopy(Object... params) {
                     case POINT:
                         size += VoltType.POINT.getLengthInBytesForFixedTypesWithoutCheck() * ((PointType[])obj).length;
                         break;
+                    case GEOGRAPHY:
+                        for (GeographyValue gv : (GeographyValue[])obj) {
+                            size += 4; // length prefix
```
BillWhite: Does this work on Centos 6?  If so, we should completely abandon all new code with BOOST_FOREACH.

src/frontend/org/voltdb/ParameterSet.java
```diff
@@ -578,6 +595,15 @@ else if (nextType == VoltType.STRING) {
                 case POINT :
                     value = PointType.unflattenFromBuffer(in);
                     break;
+                case GEOGRAPHY :
+                    len = in.getInt();
+                    if (len == VoltType.NULL_STRING_LENGTH) {
+                        value = VoltType.NULL_GEOGRAPHY;
+                    }
+                    else {
+                        value = GeographyValue.unflattenFromBuffer(in);
```
BillWhite: The description of (de)serialization of the Geography type needs to be in the Wire Protocol Specification.

Also with Point.

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -0,0 +1,286 @@
...
+        int length = 4; // number of loops prefix
+        for (List<PointType> loop : m_loops) {
+            length += 4; // number of vertices prefix;
+            length += (loop.size() * PointType.getLengthInBytes());
+        }
+
+        return length;
+    }
+
+    /**
```
BillWhite: I think the format of the polygons should be documented right here.

BillWhite:
That's the end of my plaintive bleatings for this round.

src/frontend/org/voltdb/ParameterSet.java
wolffcm: This Java code, so it should work everywhere.

I did have range for loops in C++ for iterating over the loops and vertices of polygons.  It was elegant.  I took them out, since the page Phil linked to says it not supported for compiler version we're using on CentOS 6.

wolffcm:
Thanks Paul and Bill.  I'll make changes based on your comments and push another commit to this PR.

src/ee/common/Point.hpp
paulmartel: IPronoun disagreeVerb withAdjective BillName onPreposition thisMadness. I'm not AT ALL seeing a benefit to naming types with the suffix Type except for metatypes like ExpressionType ValueType, PlanNodeType. A good old initial capital letter seems to cover thngs nicely.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
wolffcm: I like your suggested names, I'll use them.

tests/frontend/org/voltdb/types/TestGeographyValue.java
wolffcm: Reading the WKT [spec](http://www.opengeospatial.org/standards/sfa), keywords are not case sensitive. Yes, `toString` always produces uppercase text.

I'll add test cases with white space and mixed case keywords.

src/frontend/org/voltdb/types/GeographyValue.java
wolffcm: Done.

src/frontend/org/voltdb/ParameterSet.java
wolffcm: Where does this spec live?  If it's not a lot of trouble to doc this, I'll just do it.  Otherwise, I'll file a ticket.

wolffcm:
@paulmartel @BillWhite I've pushed another commit that addresses your review comments.  This branch is ready for a second round of review.

src/frontend/org/voltdb/ParameterSet.java
BillWhite: I can't find the spec's sources.  It needs a lot of copy editing.  Andrew Gent disavows all responsibility for it, or at least he did the last time I asked.  It's possible Ruth knows.

I think there should be a ticket.

BillWhite:
These look fine to me.

src/ee/expressions/geofunctions.cpp
```diff
@@ -92,4 +117,107 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POINTFROMTEXT>() const
...
+            break;
+        }
+        else {
+            throwInvalidWktPoly("unexpected token: '" + (*it) + "'");
+        }
+    }
+
+    if (it != end) {
+        // extra stuff after input
+        throwInvalidWktPoly("unrecognized input after WKT");
```
paulmartel: Minor. This error message would be more practically helpful if it contained (at least some of) the remaining unrecognized text.
I am thinking of the case where the user accidentally inserts two closing parens instead of one to end their initial loop -- seeing the text of their "hole" loops in the message would hint that the problem is actually in the middle of rather than AFTER the WKT that they had INTENDED to define.

I have no other comments this round.

wolffcm:
Resolved conflicts and squash merged on a separate branch.

PR #2874, opened 2015-10-05
Update result set of the query in junit to be order deterministic

hhakimi53:
Additional changes to make result set of another query in junit framework to be deterministic.


paulmartel:
Looks good.

PR #2876, opened 2015-10-06
Eng 8683 send full tuple in update records

luyangco:


src/ee/execution/VoltDBEngine.h
```diff
@@ -134,6 +134,7 @@ class __attribute__((visibility("default"))) VoltDBEngine {
         TableCatalogDelegate* getTableDelegate(std::string name) const;
         catalog::Database* getDatabase() const { return m_database; }
         catalog::Table* getCatalogTable(std::string name) const;
+        virtual bool getIsActiveActiveDREnabled() const { return getDatabase()->isActiveActiveDRed(); }
```
afrogers: Probably preferable to put this in the .cpp file and leave the forward declaration.

src/ee/storage/persistenttable.cpp
```diff
@@ -629,7 +629,8 @@ bool PersistentTable::updateTupleWithSpecificIndexes(TableTuple &targetTupleToUp
         const int64_t currentSpHandle = ec->currentSpHandle();
         const int64_t currentUniqueId = ec->currentUniqueId();
         std::pair<const TableIndex*, uint32_t> uniqueIndex = getSmallestUniqueIndex();
-        drMark = drStream->appendUpdateRecord(lastCommittedSpHandle, m_signature, currentTxnId, currentSpHandle, currentUniqueId, targetTupleToUpdate, sourceTupleWithNewValues, uniqueIndex);
+        bool isActiveAcitve = ec->getEngine()->getIsActiveActiveDREnabled();
```
afrogers: Really, a trivial way to do this entire thing would just be to a) rename `getSmallestUniqueIndex()` to `getUniqueIndexForDR()` or something of the like--this is the only place we use it, after all, and b) just change that function not to select an index if active-active is enabled. (Oh, and then just set that there isn't an index in that case, and clear that flag on catalog update/dr active state change)

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -842,8 +857,8 @@ TEST_F(DRBinaryLogTest, BasicUpdate) {
 
 TEST_F(DRBinaryLogTest, UpdateWithUniqueIndex) {
     createIndexes();
-    std::pair<const TableIndex*, uint32_t> indexPair = m_table->getSmallestUniqueIndex();
-    std::pair<const TableIndex*, uint32_t> indexPairReplica = m_tableReplica->getSmallestUniqueIndex();
+    std::pair<const TableIndex*, uint32_t> indexPair = m_table->getUniqueIndexForDR();
+    std::pair<const TableIndex*, uint32_t> indexPairReplica = m_tableReplica->getUniqueIndexForDR();
```
afrogers: I'd like to see a test where you reset the engine with active-active enabled and show that the update/delete optimizations aren't used.

afrogers:
Reviewed

afrogers:
Closing to push rebased version.

PR #2877, opened 2015-10-06
Make makefiles faster.

BillWhite:
The current makefile tools call the C++ preprocessor on all source files to find dependences, and then write the dependences to the makefile itself.  This makes the makefile creation slower than necessary.  It may also cause some dependences to be ignored.  If a dependence is added, the makefile will not know that.  There is some ant logic to compare the dependence of the makefile and the files it uses, but this is really unnecessary.

This change generates makefiles with the same gay abandon.  But the makefiles are now generated almost immediately.  Dependences are discovered by the compiler and written to special files which the makefile includes, if they exist.  This means that the makefile has no dependences when it is generated.

This is the way dependences have been managed in C++ sources for several years, since egcc, as I recall.  This is not an experimental technique.


wolffcm:
This looks good to me.  How similar is this to what we currently have on our GEO topic branch?  Should we expect a lot of conflicts when we need to merge to master?

BillWhite:
This should be almost exactly what we have on the GEO topic branch.  I just ripped out the inclusion of the S2 sources.

PR #2881, opened 2015-10-08
ENG-9116, change the index array iteration order so now check the pri…

luyangco:
…mary key first.

If the insertion is going to trigger constraint violation, fails early to reduce the occurences of rewinding.
This is also useful for A/A conflict detection, which needs to report primary key constraint violation first.

Change-Id: Ieda75da8743387e7282f856e76c0aa60348f7a65

### DIFF
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -969,13 +969,13 @@ void PersistentTable::deleteFromAllIndexes(TableTuple *tuple) {
 }
 
 void PersistentTable::tryInsertOnAllIndexes(TableTuple *tuple, TableTuple *conflict) {
-    for (int i = static_cast<int>(m_indexes.size()) - 1; i >= 0; --i) {
+    for (int i = 0; i < static_cast<int>(m_indexes.size()); ++i) {
         m_indexes[i]->addEntry(tuple, conflict);
         FAIL_IF(!conflict->isNullTuple()) {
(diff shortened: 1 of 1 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -969,13 +969,13 @@ void PersistentTable::deleteFromAllIndexes(TableTuple *tuple) {
 }
 
 void PersistentTable::tryInsertOnAllIndexes(TableTuple *tuple, TableTuple *conflict) {
-    for (int i = static_cast<int>(m_indexes.size()) - 1; i >= 0; --i) {
+    for (int i = 0; i < static_cast<int>(m_indexes.size()); ++i) {
```
nshi: Is the primary key guaranteed to be at the head of the array, even across catalog updates?

luyangco: * If the table already has data in it, catalog update is not allowed to this table. 
* If the table is empty, for any schema change, EE always drops the table first then recreate  the schema. 

Because of above two reasons, in `TableCatalogDelegate.cpp:427` the primary key is guaranteed to be the first of the indices array.

nshi: Makes sense. Thanks, Yang.

nshi: I think it's allowed to add non-unique indexes to an existing table with data in it.

PR #2882, opened 2015-10-08
ENG-9117, rearrange the index array into deterministic order (primary…

luyangco:
…_key first, unique indices afterwards, non-unique indices at the end).

Change-Id: I48960782a119d97a8e73f9590579fd3910b9f3e4


nshi:
To make it fully deterministic, it should also guarantee that the indexes of the same type (unique, non-unique) are ordered.

luyangco:
The order of unique and non-unique indexes are sorted by `index_map`, it's a std::map which defined at line 291, and the index name is the sort key. So as long as we don't change the index name, the order is always deterministic.

nshi:
Great! That'll do.

PR #2887, opened 2015-10-09
Fix makefile merge errors.

BillWhite:
I found some merge errors made when I moved these changes to master from the geo branch.  They didn't cause built failures, but they will cause merge problems when we merge the branch, and they were confusing.  Also, they didn't fix the problem caused by upgrades to the newest Xcode in Yosemite.

Jenkins is mostly blue with this.  The only jUnit test failure is one of the usual ones.  Furthermore, this does not affect the Java code, just the EE.

### DIFF
--- a/build.py
+++ b/build.py
@@ -33,6 +33,7 @@ CTX = BuildContext(sys.argv)
 #  and how the build will go down. It also checks the platform and parses
 #  command line args to determine target and build level.
 
+# print("Compiler: %s %d.%d.%d" % (CTX.compilerName(), CTX.compilerMajorVersion(), CTX.compilerMinorVersion(), CTX.compilerPatchLevel()))
 ###############################################################################
 # SET GLOBAL CONTEXT VARIABLES FOR BUILDING
 ###############################################################################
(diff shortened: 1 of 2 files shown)


buildtools.py
```diff
@@ -205,126 +245,162 @@ def buildMakefile(CTX):
...
+        makefile.write("\n")
     makefile.write("\n")
+    makefile.write("#\n# This target lets us print makefile variables, for debugging\n")
+    makefile.write("#\n# the makefile.\n#\n")
+    makefile.write("echo_makefile_config:\n")
+    makefile.write('\t@echo "ROOTDIR = $(ROOTDIR)"\n')
+    makefile.write('\t@echo "OBJDIR = $(OBJDIR)"\n')
+    makefile.write('\t@echo "SRCDIR = $(SRCDIR)"\n')
+    makefile.write('\t@echo "THIRD_PARTY_SRC = $(THIRD_PARTY_SRC)"\n')
+    makefile.write('\t@echo "GOOGLE_S2_SRC = $(GOOGLE_S2_SRC)"\n')
```
wolffcm: I guess GOOGLE_S2_SRC GOOGLE_S2_OBJ will be unset, so these variables will show up as blanks.  It could be confusing, so maybe leave these two lines out?

wolffcm:
These changes look good to me.    I think folks will appreciate this---I can build the EE much faster than before and it helps keep me focused.  

This is still a subset of changes that are on the GEO branch, right?  It looks like it.

I think you might have already done this, but maybe send an email to volt-developers about this change, so people know who to talk to if they run into problems?

build.py
```diff
@@ -78,7 +90,7 @@
 
 # this is where the build will look for header files
 # - the test source will also automatically look in the test root dir
-CTX.INCLUDE_DIRS = ['src/ee']
+CTX.SRC_INCLUDE_DIRS += ['src/ee' ]
```
jxchinaren: Is this space intended?

buildtools.py
```diff
@@ -47,8 +28,14 @@ def __init__(self, args):
         self.NMFLAGS = "-n"    # specialized by platform in build.py
         self.COVERAGE = False
         self.PROFILE = False
+        # Compiler Configuration
         self.CC = "gcc"
         self.CXX = "g++"
+        self.COMPILER_NAME = "Unknown"
```
jxchinaren: Nice cleanup and refactoring.

buildtools.py
```diff
@@ -66,6 +53,62 @@ def __init__(self, args):
...
+        vvector = [0, 0, 0]
+        # Apple now uses clang and has its own versioning system.
+        # The version 7 compiler needs special compilation options.
+        # This is pretty dumb code that could be improved as we support more
+        # compilers and versions.
+        #   Xcode versions before 10.10
+        #   Xcode version 10.11
+        #   Clang built not from apple.
+        #   gcc
+        patterns = [
```
jxchinaren: I guess we only have one single version of MAC OS X on Jenkins, so some of these regx have not been tested. Maybe we do not care too much as MAC OS X is not a production ready platform. 

However, nice covering for us to upgrade for OS X EI Capitan.

jxchinaren:
I have understood all the changes well enough, but it looks good to go. Yes, we need this better efficient compiling work.

PR #2892, opened 2015-10-10
Change underlying type for POINT from float to double

wolffcm:
Also, use a different representation for null---not (NaN, NaN), but
(360.0, 360.0).

Made the Java client return a null object reference instead of returning an
object with the lat and long set to magic numbers.

Added POINT and GEOGRAPHY to the round trip DDL tests.


paulmartel:
Looks good to merge.

wolffcm:
Thanks Paul.  I'm still waiting on Jenkins to qualify.

src/ee/common/Point.hpp
```diff
@@ -46,13 +45,24 @@ class Point {
         : m_latitude(latitude)
         , m_longitude(longitude)
     {
+        assert (m_latitude >= -90.0 && m_latitude <= 90.0);
```
BillWhite: Should this be an assertion or an exception?  If it's an assertion you can't recover, really.  It might be a result of an arithmetic calculation which happened to move a longitude out of range.

src/ee/common/Point.hpp
```diff
@@ -46,13 +45,24 @@ class Point {
...
 
-    /** Null values are represented as either lat or lng being NaN. */
+    // Use the number 360.0 for the null coordinate.
+    static Coord nullCoord() {
+        // A static const member could be used for this, but clang
+        // wants the constexpr keyword to be used with floating-point
+        // constants, and constexpr isn't supported until gcc 4.6.
+        // Creating a static function seems nicer than suppressing the
+        // warning or using the conditional compilation.
+        return 360.0;
```
BillWhite: This could also be a preprocessor macro.  If the static function is explicitly declared inline, isn't that effectively the same as a constexpr constant?  Maybe this is a reason to move to full C++11?

tests/frontend/org/voltdb/types/TestPointType.java
```diff
@@ -23,31 +23,77 @@
...
-        assertEquals(20.666f, point.getLongitude(), 0.001);
+        assertEquals(16, PointType.getLengthInBytes());
+
+        PointType point = new PointType(10.333, 20.666);
+        assertEquals(10.333, point.getLatitude(), 0.001);
+        assertEquals(20.666, point.getLongitude(), 0.001);
 
         assertTrue(point.equals(point));
-        assertFalse(point.isNull());
+        assertFalse(point.equals(new PointType(0.0, 10.0)));
```
BillWhite: Where did the "isNull" test go?  I would think it's probably useful.

tests/frontend/org/voltdb/types/TestPointType.java
```diff
@@ -23,31 +23,77 @@
...
+
+        // Null serialization puts 360.0 in both lat and long
+        bb.position(0);
+        PointType.serializeNull(bb);
+        bb.position(0);
+        assertEquals(360.0, bb.getDouble());
+        assertEquals(360.0, bb.getDouble());
     }
 
     private void testOnePointFromFactory(String aWKT, float aLatitude, float aLongitude, float aEpsilon, String aErrMsg) {
```
BillWhite: These floats should be changed to be doubles, shouldn't they?  I'm not sure it 100% matters, since they will be lengthened by java wherever necessary.  But it seems more consistent to change them all.

BillWhite:
That's all I've got.  There is nothing really pressing, and it's probably better to merge this sooner.  But there is one "isNull" test which seems to have disappeared, and some floats in testOnePointFromFactory which should probably be changed to doubles.

src/ee/common/Point.hpp
wolffcm: Right now there's no operations that produce points other than `pointfromtext`.  So If we ever get a point that's out of range, we know that there's data corruption.

I'm not certain but I think that S2 will convert `(lat, lng)` to `(x, y, z)` for most of its calculations, and for operations that return points, we'll want to convert back to lat/lng.  This conversion shouldn't produce out-of-range longitude.  

I'd like to revisit this once we have more geo functionality to understand the practical implications here.

tests/frontend/org/voltdb/types/TestPointType.java
wolffcm: There is no `isNull` method, because no null `PointType` instance can be created.  We just use a null reference for this.

wolffcm: I mean I guess we could have an `isNull` method that always returns true?

BillWhite: Ok.  Now I understand.

PR #2894, opened 2015-10-13
ENG-8789 reject loading classes compiled with higher Java version

dweiss88:


src/frontend/org/voltdb/sysprocs/UpdateApplicationCatalog.java
```diff
@@ -180,10 +190,35 @@ public DependencyPair executePlanFragment(
...
+                        m_javaClass.forName(classname, true, testjarloader);
                     }
                     // LinkageError catches most of the various class loading errors we'd
                     // care about here.
+                    catch (UnsupportedClassVersionError e) {
+                        String msg = "Cannot load classes compiled with a higher version of Java than currently" +
+                                     " in use.";
+                        Integer major = 0;
+                        try {
+                            major = Integer.parseInt(e.getMessage().split("version")[1].trim().split("\\.")[0]);
```
manjujames: We probably should have null checks and array index checks and all that? It won't be as clean looking code with the checks, but will protect us in case the error message format changes.

src/frontend/org/voltdb/sysprocs/UpdateApplicationCatalog.java
```diff
@@ -180,10 +190,35 @@ public DependencyPair executePlanFragment(
...
+                        } catch (Exception ex) {
+                            // Could not extract major version number from error message.
+                        }
+                        HashMap<Integer, String> versionMap = new HashMap<Integer, String>();
+                        versionMap.put(45, "Java 1.1");
```
manjujames: This map could be defined as a constant at the same place as the JavaClassForTest class.

src/frontend/org/voltdb/sysprocs/UpdateApplicationCatalog.java
```diff
@@ -180,10 +190,35 @@ public DependencyPair executePlanFragment(
...
+                        versionMap.put(46, "Java 1.2");
+                        versionMap.put(47, "Java 1.3");
+                        versionMap.put(48, "Java 1.4");
+                        versionMap.put(49, "Java 5");
+                        versionMap.put(50, "Java 6");
+                        versionMap.put(51, "Java 7");
+                        versionMap.put(52, "Java 8");
+                        if (versionMap.containsKey(major)) {
+                            msg = msg.concat(" Class " + classname + " was compiled with " + versionMap.get(major) +
```
manjujames: I think this message with the exact classname is useful even if we can't find 'major'. So, may be show the error with the version #, but with the other details if we cannot find the java version that was used for compilation.

tests/frontend/org/voltdb/sysprocs/TestMockUpdateApplicationCatalog.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    public void testVersionCheck() throws IOException, ClassNotFoundException {
+        String newCatalogURL = Configuration.getPathToCatalogForTest("catalogupdate-cluster-expanded.jar");
+        String deploymentURL = Configuration.getPathToCatalogForTest("catalogupdate-cluster-base.xml");
+        try {
+            m_client.updateApplicationCatalog(new File(newCatalogURL), new File(deploymentURL));
+            fail("Update catalog should fail with version error.");
+        } catch (ProcCallException e) {
+            assertTrue(e.getMessage().contains("Cannot load classes compiled with a higher version"));
```
manjujames: You could probably check that "Java 8" shows up in the error message as well. That will test that the parsing on the message in UnsupportedClassVersionError also works fine.

src/frontend/org/voltdb/sysprocs/UpdateApplicationCatalog.java
```diff
@@ -180,10 +202,30 @@ public DependencyPair executePlanFragment(
...
                     // care about here.
+                    catch (UnsupportedClassVersionError e) {
+                        String msg = "Cannot load classes compiled with a higher version of Java than currently" +
+                                     " in use. Class " + classname + " was compiled with ";
+
+                        Integer major = 0;
+                        try {
+                            major = Integer.parseInt(e.getMessage().split("version")[1].trim().split("\\.")[0]);
+                        } catch (Exception ex) {
+                            log.debug("Unable to parse compile version number from UnsupportedClassVersionError.");
```
manjujames: I think we should log the exception also. And use 'if log.isDebugEnabled()'

manjujames:
Other than the minor debug comment, this looks good to me. Feel free to merge if the branch build looks OK.

PR #2895, opened 2015-10-13
Simplify ThreadLocalPool memory management API

paulmartel:
This changeset is mostly refactoring for simplified interfaces and adding comments.
The two small intended behavior changes are:
    a bug fix specific to MEMCHECK builds where strings that SHOULD be leaked into the temp string pool were being explicitly deleted -- not sure why memcheck runs seemed to only sometimes complain.
    an odd failure mode when freeing an invalid allocation in ThreadLocalPool that arose from too much
shared code between the allocate and free code paths. If the ThreadLocalPool could not find a pool from which to deallocate an object, based on the claimed size of the object, instead of failing, it tried to allocate a new pool for that object size (DOH!) before complaining that the pool did not contain the object (DUH!). This should never happen in production, but it has historically been a great way to turn a small bug (corruption of an object's self-recorded allocation size) into a big mess - an attempt to needlessly allocate a potentially huge pool.

### DIFF
--- a/src/ee/common/StringRef.h
+++ b/src/ee/common/StringRef.h
@@ -49,11 +49,12 @@ namespace voltdb
         /// allocated out of the ThreadLocalPool.
         static StringRef* create(std::size_t size, Pool* dataPool);
 
-        /// Destroy the given StringRef object and free any memory, if
-        /// any, allocated from pools to store the object.
+        /// Destroy the given StringRef object and free any memory
+        /// allocated from persistet pools to store the object.
         /// sref must have been allocated and returned by a call to
-        /// StringRef::create() and must not have been created in a
-        /// temporary Pool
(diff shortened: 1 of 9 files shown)


src/ee/common/StringRef.h
```diff
@@ -49,11 +49,12 @@ namespace voltdb
         /// allocated out of the ThreadLocalPool.
         static StringRef* create(std::size_t size, Pool* dataPool);
 
-        /// Destroy the given StringRef object and free any memory, if
-        /// any, allocated from pools to store the object.
+        /// Destroy the given StringRef object and free any memory
+        /// allocated from persistet pools to store the object.
```
BillWhite: Sorry, a typo: Is this "persistent" pools, or maybe "persisted" pools?

BillWhite:
I'm happy with this, I guess.

PR #2899, opened 2015-10-15
Fix EE crash when selecting GEOGRAPHY from partitioned tables

wolffcm:


BillWhite:
Looks right to me.  I like the test changes.

PR #2900, opened 2015-10-15
Eng 9094 planner error stack trace

BillWhite:
Add a couple of log entries in exception handlers.

### DIFF
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -108,6 +108,10 @@ public class PlannerTool {
         return planSql(sqlIn, infer, false, null);
     }
 
+    private void logException(Exception e, String fmtLabel) {
+        hostLog.error(fmtLabel + ": ", e);
+    }
+
     /**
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -255,6 +260,7 @@ synchronized AdHocPlannedStatement planSql(String sqlIn, StatementPartitioning p
                     partitioning = plan.getStatementPartitioning();
                 }
             } catch (Exception e) {
+                logException(e, "Error compiling query");
                 throw new RuntimeException("Error compiling query: " + e.toString(), e);
```
paulmartel: I'd find helpful an appendage to the RuntimeException's message saying "A java stack trace has been logged."

Anyway, this seems like a positive step. I suppose that it may turn out that there are other paths not covered and/or that these RuntimeExceptions feed into yet another catch which MIGHT be a better, more general place to log the stack, but maybe we should just try this as is and see what happens.

BillWhite: I suppose I could step past the rethrow and see where this propagates to.

PR #2906, opened 2015-10-19
ENG-9287: Make DRLogSegmentId serializable

afrogers:


nshi:
Want to add a simple sanity check unit test to make sure we can round-trip the same object?

PR #2908, opened 2015-10-19
Add testing for save/restore snapshots with geo columns

wolffcm:
There is a companion pro pull request:

https://github.com/VoltDB/pro/pull/1044


BillWhite:
This all seems plausible as well.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -921,6 +924,17 @@ private void addColumnValue(Object value, VoltType columnType, int col) {
...
                 }
 
+                case POINT: {
+                    if (value instanceof PointType) {
+                        PointType pt = (PointType)value;
+                        pt.flattenToBuffer(m_buffer);
+                    }
+                    else {
+                        throw new ClassCastException();
+                    }
```
paulmartel: This is a strangely verbose way of getting a ClassCastException.
You get the same behavior if you take out the "if (value instanceof PointType) {".
Is there a rationale for being purposely more verbose?
It looks like this was cribbed from a different case that wanted to try multiple object formats before failing with the ClassCastException -- it HAD a rationale THERE -- wanting to have parallel code paths for the different input types and not wanting to have to catch the first ClassCastException for the chance to retry.

src/frontend/org/voltdb/VoltType.java
```diff
@@ -742,7 +742,8 @@ public static boolean isNullVoltType(Object obj)
         if ((obj == null) ||
             (obj == VoltType.NULL_TIMESTAMP) ||
             (obj == VoltType.NULL_STRING_OR_VARBINARY) ||
-            (obj == VoltType.NULL_DECIMAL))
+            (obj == VoltType.NULL_DECIMAL) ||
+            (obj == VoltType.NULL_POINT))
```
paulmartel: Added handling for NULL_POINT, but what about GEOGRAPHY?

paulmartel:
Either I missed something or the only thing being tested here is save/restore for NULL POINT/GEOGRAPHY values. SO, seems like it could use more coverage.

paulmartel:
Now I see the pro coverage for the pro feature. DOH! nevermind. This looks good.

src/frontend/org/voltdb/VoltTable.java
wolffcm: You are right, of course.  I simplified the code like you suggested.

wolffcm:
Rebased branch...

PR #2909, opened 2015-10-20
ENG-9103 improve channel assignment trigger in channel distributer

vtkstef:


akhanzode:
Looks good to me merge when blue.

vtkstef:
squash merged into master

PR #2911, opened 2015-10-21
ENG=9136: Make sure that connections are cleaned up correctly on upda…

manjujames:
…te application catalog.

### DIFF
--- a/src/frontend/org/voltdb/HTTPClientInterface.java
+++ b/src/frontend/org/voltdb/HTTPClientInterface.java
@@ -20,7 +20,8 @@ package org.voltdb;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.servlet.http.HttpServletResponse;
@@ -47,8 +48,8 @@ public class HTTPClientInterface {
     private static final RateLimitedLogger m_rate_limited_log = new RateLimitedLogger(10 * 1000, m_log, Level.WARN);
 
     AtomicReference<AuthenticatedConnectionCache> m_connections = new AtomicReference<AuthenticatedConnectionCache>();
+    private AuthenticatedConnectionCache m_oldCache;
     static final int CACHE_TARGET_SIZE = 10;
-    private final AtomicBoolean m_shouldUpdateCatalog = new AtomicBoolean(false);
 
     public static final String PARAM_USERNAME = "User";
     public static final String PARAM_PASSWORD = "Password";
@@ -56,6 +57,7 @@ public class HTTPClientInterface {
     public static final String PARAM_ADMIN = "admin";
     int m_timeout = 0;
     final String m_timeoutResponse;
+    private ExecutorService m_closeAllExecutor;
 
     public final static int MAX_QUERY_PARAM_SIZE = 2 * 1024 * 1024; // 2MB
 
@@ -104,6 +106,8 @@ public class HTTPClientInterface {
         final ClientResponseImpl r = new ClientResponseImpl(ClientResponse.CONNECTION_TIMEOUT,
                 new VoltTable[0], "Request Timeout");
         m_timeoutResponse = r.toJSONString();
+        // TODO: how many threads should this require? Only the first closeAll will take time, so 1 or 2 should be good enough.
+        m_closeAllExecutor = Executors.newFixedThreadPool(2);
     }
 
     public void process(Request request, HttpServletResponse response) {
@@ -272,18 +276,10 @@ public class HTTPClientInterface {
         String admin = request.getParameter(PARAM_ADMIN);
 
         AuthenticatedConnectionCache connection_cache = m_connections.get();
-
-        // first check for a catalog update and purge the cached connections
-        // if one has happened since we were here last
-        if (m_shouldUpdateCatalog.compareAndSet(true, false))
-        {
-            if (connection_cache != null) {
-                connection_cache.closeAll();
-                connection_cache = null;
+        while (connection_cache == null) {
+            if (m_oldCache!=null) {
+                closeAllAsync(m_oldCache);
             }
-        }
-
-        if (connection_cache == null) {
             Configuration config = VoltDB.instance().getConfig();
             int port = config.m_port;
             int adminPort = config.m_adminPort;
@@ -301,7 +297,13 @@ public class HTTPClientInterface {
             if (config.m_adminInterface.length() > 0) {
                 adminInterface = config.m_adminInterface;
             }
-            m_connections.set(new AuthenticatedConnectionCache(10, clientInterface, port, adminInterface, adminPort));
+            AuthenticatedConnectionCache newCache = new AuthenticatedConnectionCache(10, clientInterface, port, adminInterface, adminPort);
+            boolean setNewValue = m_connections.compareAndSet(null, newCache);
+            if (setNewValue) {
+                connection_cache = newCache;
+            } else {
+                connection_cache = m_connections.get();
+            }
         }
 
         // check for admin mode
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -346,17 +348,26 @@ private AuthenticationResult getAuthenticationResult(Request request) {
...
-            return new AuthenticationResult(null, null, adminMode, username, ex.getMessage());
+            return new AuthenticationResult(null, null, null, adminMode, username, ex.getMessage());
         }
     }
 
+    private void closeAllAsync(final AuthenticatedConnectionCache cache) {
+        m_closeAllExecutor.submit(new Runnable() {
+            @Override
+            public void run() {
+                cache.closeAll();
```
akhanzode: catch exceptions so submit does not blow up server

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -104,6 +106,8 @@ public HTTPClientInterface() {
         final ClientResponseImpl r = new ClientResponseImpl(ClientResponse.CONNECTION_TIMEOUT,
                 new VoltTable[0], "Request Timeout");
         m_timeoutResponse = r.toJSONString();
+        // TODO: how many threads should this require? Only the first closeAll will take time, so 1 or 2 should be good enough.
+        m_closeAllExecutor = Executors.newFixedThreadPool(2);
```
akhanzode: Use coreutils and a single thread.

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -369,13 +380,17 @@ public AuthenticationResult authenticate(Request request) {
...
-            m_connections.get().releaseClient(authResult.m_client, authResult.m_scheme, force);
+            assert(authResult.m_connectionCache != null);
+            authResult.m_connectionCache.releaseClient(authResult.m_client, authResult.m_scheme, force || (authResult.m_connectionCache != m_connections.get()));
         }
     }
 
     public void notifyOfCatalogUpdate()
     {
-        m_shouldUpdateCatalog.set(true);
+        AuthenticatedConnectionCache cache = m_connections.get();
```
akhanzode: comment this part and the while loop

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -369,13 +380,17 @@ public AuthenticationResult authenticate(Request request) {
     //Must be called by all who call authenticate.
     public void releaseClient(AuthenticationResult authResult, boolean force) {
         if (authResult != null && authResult.m_client != null) {
-            assert(m_connections != null);
-            m_connections.get().releaseClient(authResult.m_client, authResult.m_scheme, force);
+            assert(authResult.m_connectionCache != null);
+            authResult.m_connectionCache.releaseClient(authResult.m_client, authResult.m_scheme, force || (authResult.m_connectionCache != m_connections.get()));
```
akhanzode: null check.

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -369,13 +380,17 @@ public AuthenticationResult authenticate(Request request) {
     //Must be called by all who call authenticate.
     public void releaseClient(AuthenticationResult authResult, boolean force) {
         if (authResult != null && authResult.m_client != null) {
-            assert(m_connections != null);
-            m_connections.get().releaseClient(authResult.m_client, authResult.m_scheme, force);
+            assert(authResult.m_connectionCache != null);
```
akhanzode: remove assert as this can be null and valid.

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -104,6 +107,13 @@ public HTTPClientInterface() {
         final ClientResponseImpl r = new ClientResponseImpl(ClientResponse.CONNECTION_TIMEOUT,
                 new VoltTable[0], "Request Timeout");
         m_timeoutResponse = r.toJSONString();
+        m_closeAllExecutor = CoreUtils.getSingleThreadExecutor("HttpClientInterface-closeAll");
+    }
+
+    public void stop() {
+        // This is called on server shutdown.
+        // So, it is OK if we interrupt threads trying to close client connections.
+        m_closeAllExecutor.shutdownNow();
```
akhanzode: Do you intend to now await for es to shutdown?

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -56,6 +58,7 @@
     public static final String PARAM_ADMIN = "admin";
     int m_timeout = 0;
     final String m_timeoutResponse;
+    private ExecutorService m_closeAllExecutor;
```
akhanzode: Can be final right?

src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -346,14 +354,32 @@ private AuthenticationResult getAuthenticationResult(Request request) {
...
+
+        try {
+            m_closeAllExecutor.submit(new Runnable() {
+                @Override
+                public void run() {
+                    cache.closeAll();
+                }
+            });
+        } catch(RejectedExecutionException e) {
+            m_rate_limited_log.log(System.currentTimeMillis(), Level.WARN, e,
```
akhanzode: Should be just INFO or DEBUG as nothing bad has happened.

src/frontend/org/voltdb/client/AuthenticatedConnectionCache.java
```diff
@@ -51,6 +51,7 @@
     final int m_port;
     final int m_adminPort;
     final int m_targetSize; // goal size of the client cache
+    private boolean m_isClosing;
```
akhanzode: Is it closing or closed true or false?

tests/frontend/org/voltdb/TestJSONInterface.java
```diff
@@ -1991,4 +1992,174 @@ public void testProfile() throws Exception {
...
+                    String responseJSON = callProcOverJSON("@AdHoc", pset, m_username, m_password, false, false);
+                    //System.out.println("Insert response: " + responseJSON);
+                    if (!responseJSON.contains("\"data\":[[1]]")) {
+                        System.out.println("Insert should have returned 1. Got: " + responseJSON);
+                        s_success = false;
+                        return;
+                    }
+                    s_insertCount.incrementAndGet();
+                    Thread.sleep(200);
+                    pset = ParameterSet.fromArrayNoCopy("select count(*) from test1");
```
akhanzode: Do you need this ParameterSet sounds not correct.. should be just string param to @AdHoc

src/frontend/org/voltdb/HTTPClientInterface.java
manjujames: I am calling shutdownNow because if we get here it is because the server is getting shut down. So, there isn't really a need to wait for termination.

src/frontend/org/voltdb/HTTPClientInterface.java
manjujames: We don't ever expect this to happen. So, if it does happen, we need to know so that we can fix it, correct?

src/frontend/org/voltdb/HTTPClientInterface.java
manjujames: eh?

manjujames: Never mind, I get it now. You want me to add comments. Makes sense.

src/frontend/org/voltdb/client/AuthenticatedConnectionCache.java
manjujames: Closing. This is set when closeAll is called so that additional closeAlls will not need to do anything.

tests/frontend/org/voltdb/TestJSONInterface.java
manjujames: It is just a string parameter.

src/frontend/org/voltdb/HTTPClientInterface.java
akhanzode: Do u want this to generate a support call/ticket when this happens is the question?

src/frontend/org/voltdb/client/AuthenticatedConnectionCache.java
akhanzode: What I mean is you don't have closing and closed states so closed can be the variable name and isClosed is what will guard. m_closed can be set at the end of closeAll I believe its synchronized right?

manjujames: We need to set it at the beginning on closeAll method because we don't want to start a job to close the connections in the cache, if one has already started (HttpClientInterface.closeAllAsync).

src/frontend/org/voltdb/HTTPClientInterface.java
manjujames: I would definitely like to hear about it..

PR #2912, opened 2015-10-21
Eng 9162 sqlcov postgres

johnTcrawford:
Various changes to allow the SQLCoverage tests to run against PostgreSQL, rather than HSqlDB (optionally - HSqlDB is still the default). The biggest part (among many) of this is the split of the old HsqlBackend class into GenericBackend and its two sub-classes, HsqlBackend and PostgreSQLBackend.
Note: before merging to master, I will merge master to this branch (& watch Jenkins results), to confirm that the memcheck-lite and pro-pythontests failures are unrelated to any changes on this branch.


src/frontend/org/voltdb/HsqlBackend.java
```diff
@@ -17,44 +17,28 @@
...
-import org.voltdb.types.TimestampType;
 import org.voltdb.utils.Encoder;
 import org.voltdb.utils.LogKeys;
 
 /**
  * A wrapper around the HSQLDB engine. This class can be used to execute SQL
- * statements instead of the C++ ExecutionEngine. It is currently not used.
- *
+ * statements instead of the C++ ExecutionEngine. It is currently used only
+ * by the SQL Coverage tests.
```
paulmartel: should be "SQL Coverage and junit regressionsuite tests"

src/frontend/org/voltdb/GenericBackend.java
```diff
@@ -0,0 +1,322 @@
...
+import org.voltdb.utils.Encoder;
+import org.voltdb.utils.LogKeys;
+
+/**
+ * A wrapper around another database server (and JDBC connection), such as
+ * HSQLDB or PostgreSQL. This class, with its sub-classes can be used to
+ * execute SQL statements instead of the C++ ExecutionEngine. It is currently
+ * used only by the SQL Coverage tests.
+ */
+public abstract class GenericBackend {
```
paulmartel: As a name "Generic" is too generic. The name should convey that it is not so generic as to include the usual default EE backend. It should contain some kind of descriptive term like "Alternative" or "NonNative" or "External" or "NonVoltDB".
At one point, I had the idea of a truly generic interface that covered both volt (JNI, IPC) and non-volt (HSQL) backends but I ran out of innovation week. Maybe someday, but that's not this and we should make that clear.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -529,6 +538,13 @@ public HsqlBackend getHsqlBackendIfExists() {
         return m_site.getHsqlBackendIfExists();
     }
 
+    /**
+     * If returns non-null, then using PostgreSQL backend
+     */
+    public PostgreSQLBackend getPostgreSQLBackendIfExists() {
+        return m_site.getPostgreSQLBackendIfExists();
+    }
+
```
paulmartel: The GenericBackend class provides an opportunity not only to eliminate redundancy in the implementation of HsqlBackend and PostgreSQLBackend but also to define the entire interface to these classes and eliminate redundancies in the callers, like this.

In other words, instead of duplicating HsqlBackendxxx into HsqlBackendxxx and PostgreSQLBackendxxx, you can mutate HsqlBackendxxx into GenericBackendxxx,
leaving the callers more abstract but no more complex than they were.

The only part of the callers that should need to be tweaked is the object instantiation which could be delegated to a factory (class or method) which then becomes the only code (except for the BackendTarget enum) that has to change to add or drop a new backend. Everything else should be dealing with the abstract generic.

src/frontend/org/voltdb/GenericBackend.java
```diff
@@ -0,0 +1,322 @@
...
+import org.voltdb.messaging.FastSerializer;
+import org.voltdb.types.TimestampType;
+import org.voltdb.utils.Encoder;
+import org.voltdb.utils.LogKeys;
+
+/**
+ * A wrapper around another database server (and JDBC connection), such as
+ * HSQLDB or PostgreSQL. This class, with its sub-classes can be used to
+ * execute SQL statements instead of the C++ ExecutionEngine. It is currently
+ * used only by the SQL Coverage tests.
```
paulmartel: SQL Coverage and regression JUnit

src/frontend/org/voltdb/HsqlBackend.java
```diff
@@ -17,44 +17,28 @@
...
     /** java.util.logging logger. */
     @SuppressWarnings("unused")
     private static final VoltLogger log = new VoltLogger(HsqlBackend.class.getName());
-    private static final VoltLogger hostLog = new VoltLogger("HOST");
-    private static final VoltLogger sqlLog = new VoltLogger("SQL");
 
-    private final static Object backendLock = new Object();
-    private static HsqlBackend m_backend = null;
+    private static final Object backendLock = new Object();
+    protected static HsqlBackend m_backend = null;
```
paulmartel: private makes more sense than protected here.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -0,0 +1,175 @@
...
+        else if (typeName.equals("int8"))
+            return new VoltTable.ColumnInfo(colName, VoltType.BIGINT);
+        else if (typeName.equalsIgnoreCase("NUMERIC"))
+            return new VoltTable.ColumnInfo(colName, VoltType.DECIMAL);
+        else if (typeName.equals("float8"))
+            return new VoltTable.ColumnInfo(colName, VoltType.FLOAT);
+        else if (typeName.equals("bytea"))
+            return new VoltTable.ColumnInfo(colName, VoltType.VARBINARY);
+        else if (typeName.equals("char") || typeName.equals("text"))
+            return new VoltTable.ColumnInfo(colName, VoltType.STRING);
```
paulmartel: Would it be simpler, more readable just to optionally reset typeName and then just unconditionally call super?
You could even do this with a "one liner" and a static HashMap<String, String>.
Lookup typeName in the map and use its translated value, or if it's null, use typeName,

paulmartel:
End of review round.

johnTcrawford:
Paul, I believe I've addressed all of your concerns. Would you like to review again, or shall I merge to master (once Jenkins finishes running tests and is blue)?

paulmartel:
Looks good.

PR #2914, opened 2015-10-21
ENG-9248: With the changes to include lib in the jar, symbol generati…

manjujames:
…on could be called twice for 'dist' target. When it is called a second time, the symbol file was being overwritten by symbol file generated from stipped lib file. Fixed build to use overwrite when preparing lib file.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -938,7 +938,7 @@ DISTRIBUTION
 </target>
 
 <target name="prep_libvoltdb">
-    <copy todir='voltdb'>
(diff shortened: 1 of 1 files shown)


akhanzode:
Looks good to me. Does this mean customers will get lib with debug symbols?

manjujames:
The symbol file available on volt0 will have the necessary info like before. The .so file will be as before.

akhanzode:
ok merge when blue.

PR #2916, opened 2015-10-22
Detect content determinism in Read/Write stored procedures.

BillWhite:
We detect some kinds of content determinism in Read/Write stored
procedures during planning.  After we compute the best plan, we look at
the AbstractParsedExpression to find expressions which may be content
non-deterministic.  Right now we only detect aggregate functions of
floating point columns, but the mechanism is deliberately general.  This
determinism determination is propagated up to the Procedure.  In the
ProcedureCompiler, both in compileJavaProcedure and
compileSingleStmtProcedure, we check for content non-determinism and
cause a warning to be logged to the console and the log file.

Note that limit-order deterministic queries, by which I mean queries
which are order deterministic or else have no limit or offset, may be
inherently content non-deterministic, and inherently content
deterministic queries may be limit-order non-deterministic.  These are
mostly orthogonal notions.  If a stored procedure is non-deterministic,
either by limit-order or inherent non-determinism, we throw an
exception.

Note also that Read-only stored procedures are not affected by this at
all.

### DIFF
--- a/src/frontend/org/voltdb/planner/CompiledPlan.java
+++ b/src/frontend/org/voltdb/planner/CompiledPlan.java
@@ -89,12 +89,24 @@ public class CompiledPlan {
     /**
      * Whether the plan's statement mandates a result with nondeterministic content;
      */
-    private boolean m_statementHasLimitOrOffset = false;
+    private boolean                   m_statementHasLimitOrOffset       = false;
 
     /**
      * Whether the plan's statement mandates a result with deterministic content and order;
      */
-    private boolean m_statementIsOrderDeterministic = false;
+    private boolean                   m_statementIsOrderDeterministic   = false;
+
+    /**
+     * This is false iff the statement has some calculation which is in itself
+     * content non-deterministic. The most typical example is an aggregate of a
+     * column whose type is floating point. The floating point arithmetic may be
+     * slightly different with different plans or different row orders.
+     */
+    private boolean                   m_statementIsContentDeterministic = true;
+    /**
+     * This string describes the reason a plan is not content deterministic.
+     */
+    private String                    m_contentDeterminismDetail        = null;
 
     /** Which extracted param is the partitioning object (assuming parameterized plans) */
     public int partitioningKeyIndex = -1;
@@ -129,8 +141,9 @@ public class CompiledPlan {
     }
 
     /**
-     * Mark the level of result determinism imposed by the statement,
-     * which can save us from a difficult determination based on the plan graph.
+     * Mark the level of result determinism imposed by the statement, which can
+     * save us from a difficult determination based on the plan graph. Note that
+     * we don't set m_statementIsContentDeterministic here.
      */
     public void statementGuaranteesDeterminism(boolean hasLimitOrOffset, boolean order) {
         m_statementHasLimitOrOffset = hasLimitOrOffset;
@@ -149,13 +162,20 @@ public class CompiledPlan {
         return rootPlanGraph.isOrderDeterministic();
     }
 
+    public boolean isContentDeterministic() {
+        return m_statementIsContentDeterministic;
+    }
+
     /**
-     * Accessor for flag marking the original statement as guaranteeing an identical result/effect
-     * when "replayed" against the same database state, such as during replication or CL recovery.
+     * Accessor for flag marking the original statement as guaranteeing an
+     * identical result/effect when "replayed" against the same database state,
+     * such as during replication or CL recovery. If
+     * m_statementIsContentDeterministic is false we want to check this. This is
+     * the one area in which content and limit-order determinism interact.
      */
     public boolean hasDeterministicStatement()
     {
-        return m_statementIsOrderDeterministic;
+        return m_statementIsOrderDeterministic && m_statementIsContentDeterministic;
     }
 
     /**
@@ -168,10 +188,15 @@ public class CompiledPlan {
     }
 
     /**
-     * Accessor for description of plan non-determinism.
+     * Accessor for description of plan non-determinism. Note that we prefer the
+     * content determinism message to the rootPlanGraph's message.
+     *
      * @return the corresponding value from the first fragment
      */
     public String nondeterminismDetail() {
+        if (!m_statementIsContentDeterministic) {
+            return m_contentDeterminismDetail;
+        }
         return rootPlanGraph.nondeterminismDetail();
     }
 
@@ -321,5 +346,13 @@ public class CompiledPlan {
         else {
             return "CompiledPlan: [null plan graph]";
         }
+    }
+
+    public void setIsContentDeterministic(boolean contentDeterministic) {
+        m_statementIsContentDeterministic = contentDeterministic;
+    }
+
+    public void setNondeterminismDetail(String contentDeterminismMessage) {
+        m_contentDeterminismDetail = contentDeterminismMessage;
     }
 }
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -1968,4 +1968,21 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
         }
         return exprs;
     }
+
+    @Override
+
+    public String isContentDeterministic() {
+        String ans = null;
+        for (ParsedColInfo displayCol : m_displayColumns) {
+            AbstractExpression displayExpr = displayCol.expression;
+            for (AbstractExpression absExpr : displayExpr.findAllSubexpressionsOfClass(AggregateExpression.class)) {
+                AggregateExpression aggExpr = (AggregateExpression) absExpr;
+                if (aggExpr.getValueType() == VoltType.FLOAT) {
+                    return "Aggregate functions of floating point columns may not be deterministic.  We suggest converting to DECIMAL.";
+                }
+            }
+        }
+        return null;
+    }
+
 }
--- a/src/frontend/org/voltdb/planner/PlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/PlanAssembler.java
@@ -377,14 +377,20 @@ public class PlanAssembler {
             return;
         }
 
-        if (parsedStmt instanceof ParsedInsertStmt  && !plan.isOrderDeterministic()) {
+        boolean contentDeterministic = plan.isContentDeterministic();
+        if (parsedStmt instanceof ParsedInsertStmt && !(plan.isOrderDeterministic() && contentDeterministic)) {
             ParsedInsertStmt parsedInsert = (ParsedInsertStmt)parsedStmt;
             boolean targetHasLimitRowsTrigger = parsedInsert.targetTableHasLimitRowsTrigger();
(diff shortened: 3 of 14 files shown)


BillWhite:
Jenkins is uncharacteristically blue, just so you know.

pdshaw:
Kool!

src/frontend/org/voltdb/planner/CompiledPlan.java
```diff
@@ -89,12 +89,24 @@
...
+     * This is false iff the statement has some calculation which is in itself
+     * content non-deterministic. The most typical example is an aggregate of a
+     * column whose type is floating point. The floating point arithmetic may be
+     * slightly different with different plans or different row orders.
+     */
+    private boolean                   m_statementIsContentDeterministic = true;
+    /**
+     * This string describes the reason a plan is not content deterministic.
+     */
+    private String                    m_contentDeterminismDetail        = null;
```
paulmartel: Not sure if there is an official coding standard against it, but the whitespace looks excessive, ugly even.

src/frontend/org/voltdb/planner/CompiledPlan.java
```diff
@@ -322,4 +347,12 @@ public String toString() {
...
         }
     }
+
+    public void setIsContentDeterministic(boolean contentDeterministic) {
+        m_statementIsContentDeterministic = contentDeterministic;
+    }
+
+    public void setNondeterminismDetail(String contentDeterminismMessage) {
+        m_contentDeterminismDetail = contentDeterminismMessage;
+    }
```
paulmartel: Having and setting a boolean flag and a detail string whose nullity coincides exactly with the flag value is redundant and possibly misleading -- as if the two states are or can be managed separately. Why not eliminate the boolean and its setter and redefine the boolean getter (if wanted) as a null test on the string.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1968,4 +1968,21 @@ else if (display_col.columnName != null && ! display_col.columnName.equals(""))
...
+            for (AbstractExpression absExpr : displayExpr.findAllSubexpressionsOfClass(AggregateExpression.class)) {
+                AggregateExpression aggExpr = (AggregateExpression) absExpr;
+                if (aggExpr.getValueType() == VoltType.FLOAT) {
+                    return "Aggregate functions of floating point columns may not be deterministic.  We suggest converting to DECIMAL.";
+                }
+            }
+        }
+        return null;
+    }
+
```
paulmartel: This test just seems to be scratching the surface -- offending agg expressions could also influence content when they occur in the having clause.
Also, they can occur in subqueries -- we should be recursing into subquery statements.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -377,14 +377,20 @@ private static void failIfNonDeterministicDml(AbstractParsedStmt parsedStmt, Com
...
+            if (!contentDeterministic) {
+                contentDeterministicMsg = "  " + plan.nondeterminismDetail();
+            }
 
             if (parsedStmt.m_isUpsert) {
                 throw new PlanningErrorException(
                         "UPSERT statement manipulates data in a non-deterministic way.  "
-                        + "Adding an ORDER BY clause to UPSERT INTO ... SELECT may address this issue.");
+                        + "Adding an ORDER BY clause to UPSERT INTO ... SELECT may address this issue."
+                        + contentDeterministicMsg);
```
paulmartel: It seems like the advice to add an ORDER BY clause is not useful when the root of the problem is an agg of floats.
Is it possible to distinguish these cases and give better context-specific advice?

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -50,30 +50,100 @@
...
         String[] lines = captured.split("\n");
-        // Output should include a line suggesting replacement of float with double.
-        assertTrue(foundLineMatching(lines, ".*FloatParamToGetNiceComplaint.* float.* double.*"));
+        // Output should include a line suggesting replacement of float with
+        // double.
+        for (String oneMessagePattern : errorMessages) {
+            assertTrue(foundLineMatching(lines, oneMessagePattern));
+        }
+    }
+    public void testFloatParamComplaint() throws Exception {
```
paulmartel: Normally, there's a space between methods.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -50,30 +50,100 @@
...
         builder.setCompilerDebugPrintStream(capturing);
         builder.addLiteralSchema(simpleSchema);
-        builder.addPartitionInfo("floatie", "ival");
-        builder.addProcedures(FloatParamToGetNiceComplaint.class);
+        if (partitionInfo != null) {
+            for (String[] pair : partitionInfo) {
+                assertTrue(pair.length == 2);
+                builder.addPartitionInfo(pair[0], pair[1]);
+            }
+        }
```
paulmartel: addPartitionInfo is something of an anachronism and rarely worth using any more, and it's causing problems here, so I'd dump it.
You get the equivalent effect more easily by just tacking

    PARTITION TABLE %s ON COLUMN %s;

onto your literal schema which is most easily done outside this function.

The similar "literal schema" approach to defining statement procedures would also simplify this test suite.

Java procedures still need to use the special VoltProjectBuilder call, but that code becomes much simpler here when you don't have to weed out statement procedures.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -50,30 +50,100 @@
...
+                               null,
+                               FloatingInsertAggregates.class,
+                               new String[] { ".*FloatingInsertAggregates.*" });
+    }
+
+    public void testFloatingAggregatesDDL() throws Exception {
+        String simpleSchema = "create table floatingaggs_input ( alpha float );" + "create table floatingaggs_output ( beta float );" + "";
+        testCompilationFailure(simpleSchema,
+                               null,
+                               new String[] { "FLoatingDDLInsertAggregates", "create procedure FloatingDDLInsertAggregates as insert into floatingaggs_output select sum(alpha) from floatingaggs_input;",
```
paulmartel: Odd embedded upper case L in name?

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -50,30 +50,100 @@
...
+            "PRIMARY KEY(ival)" +
+            ");";
+        String[][] partitionInfo = new String[][] { { "floatie", "ival" } };
+        testCompilationFailure(simpleSchema,
+                               partitionInfo,
+                               FloatParamToGetNiceComplaint.class,
+                               new String[] { ".*FloatParamToGetNiceComplaint.* float.* double.*" });
+    }
+
+    public void testFloatingInsertAggregates() throws Exception {
```
paulmartel: This punny use of "Floating" is unhelpfully anti-descriptive.

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -35,8 +35,18 @@ protected void setUp() throws Exception {
 
     private final static boolean m_staticRetryForDebugOnFailure = /**/ false; //*/ = true;//to debug
 
-    private void assertPlanDeterminismCore(String sql, boolean order, boolean content,
-            DeterminismMode detMode)
+    private void assertPlanDeterminismCore(String sql,
```
paulmartel: Could use some commenting about the difference between Full and non-Full.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -111,6 +187,7 @@ public void testSimple() throws Exception {
         builder.addProcedures(DeterministicROSeqProc.class);
         builder.addProcedures(DeterministicRWProc1.class);
         builder.addProcedures(DeterministicRWProc2.class);
+        builder.addProcedures(FloatingUpdateAggregates.class);
```
paulmartel: Needs test cases for aggs of float only inside subqueries and for aggs of float only in the HAVING clause.

paulmartel:
End of review round.

src/frontend/org/voltdb/planner/CompiledPlan.java
BillWhite: It's an Eclipse thing.  The variables are lined up.   At least some of them are.  It would
be better if all the file was formatted this way.  I'll change it to be one space.

src/frontend/org/voltdb/planner/CompiledPlan.java
BillWhite: I was copying the original, which has a detail string and a boolean.  But the detail
string originally is never null, because it is written to the catalog.  So this is a different case.
I've changed it.

src/frontend/org/voltdb/planner/PlanAssembler.java
BillWhite: Maybe.  But both cases can occur independently, I think.  So I can't really eliminate one of them.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
BillWhite: Ok.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
BillWhite: I can't add a literal schema outside this function.  The builder is local to this function.

I was trying to factor out the compilation failure algorithm to be general enough to use for
the three cases.

I don't see that this is confusing anyway.  The documentation seems incorrect, though.
I'll fix it.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
BillWhite: A Typo.  That was actually causing a test to misbehave, though.  I'm glad you caught it.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
BillWhite: Sorry, I figured out what you meant.  I'm making this change.

BillWhite:
I added a lot of code to propagate properly from subqueries.  This looks different enough that it should go through another code review round.  So I'm closing this and deleting the github branch.  I'll push it later.

PR #2917, opened 2015-10-22
ENG-9090: Catch NoNodeException and fail starting the server with app…

manjujames:
…ropriate error message.


akhanzode:
Looks good. If jenkins is blue , merge.

PR #2918, opened 2015-10-22
Fix the language standard for gcc 4.6.3 (Ubuntu 12.04).

BillWhite:
The python rules for choosing the C++ language setting were correctly
stated but incorrectly coded.  On Ubuntu 12.04 the system compiler is gcc 4.6.3.  We need
to compile with --std=03 there.  Also, we need to include -Wno-unused-set-variables.

### DIFF
--- a/build.py
+++ b/build.py
@@ -83,18 +83,18 @@ if CTX.compilerName() == 'gcc':
 	print("GCC Version %d.%d.%d is too old\n" 
 	       % (CTX.compilerMajorVersion(), CTX.compilerMinorVersion(), CTX.compilerPatchLevel())); 
 	sys.exit(-1);
-    if CTX.compilerMinorVersion() == 4:
-	CTX.CXX_VERSION_FLAG = "--std=c++0x"
-	print("Building with C++ 0x\n")
+    if 4 <= CTX.compilerMinorVersion() <= 6:
+	   CTX.CXX_VERSION_FLAG = "--std=c++0x"
+	   print("Building with C++ 0x\n")
     else:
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -94,7 +96,7 @@
 CTX.CPPFLAGS += " " + CTX.CXX_VERSION_FLAG
 
 # linker flags
-CTX.LDFLAGS += """ -g3"""
+CTX.LDFLAGS += """ -g3 """
```
wolffcm: Any idea why this flag is here as an LD flag?

build.py
```diff
@@ -83,7 +85,7 @@
 	print("GCC Version %d.%d.%d is too old\n" 
 	       % (CTX.compilerMajorVersion(), CTX.compilerMinorVersion(), CTX.compilerPatchLevel())); 
 	sys.exit(-1);
-    if CTX.compilerMinorVersion() == 4:
+    if 4 <= CTX.compilerMinorVersion() <= 6:
```
wolffcm: I guess this is legal python syntax?  Cool.

wolffcm:
Looks good.  You've verified that this builds on U12?  Please go ahead and merge if so.

BillWhite:
This does not build on the macmini2, and probably not on the macmini.  Phil says we can upgrade these two machines.  The problem, I think, is that xcode is too old there.  The gcc version on macmini2 is based on llvm 3.3, which may not be sufficiently grand to use C++11.

I'm going to merge this anyway.  The problem is understood sufficiently well that we don't want the SW version on a non-deploy platform to stand in the way of making forward progress.

PR #2919, opened 2015-10-22
Eng 8842 rebased

afrogers:


src/ee/execution/JNITopend.cpp
```diff
@@ -490,82 +490,76 @@ int64_t JNITopend::pushDRBuffer(int32_t partitionId, StreamBlock *block) {
...
-    assert(existingRows != NULL);
-    assert(expectedRows != NULL);
-    assert(newRows != NULL);
+bool JNITopend::reportDRConflict(int32_t partitionId, int64_t timestamp, std::string tableName, DRRecordType action,
+        DRConflictType deleteConflict, Table *existingTableForDelete, Table *expectedTableForDelete,
+        DRConflictType insertConflict, Table *existingTableForInsert, Table *newTableForInsert) {
+
+    if (action == DR_RECORD_DELETE || action == DR_RECORD_UPDATE) {
+        assert(existingTableForDelete);
+        assert(expectedTableForDelete);
```
nshi: Is it true that for a delete or update you'll always have an delete conflict? If not, will these two tables be NULL?

src/ee/execution/JNITopend.cpp
```diff
@@ -490,82 +490,76 @@ int64_t JNITopend::pushDRBuffer(int32_t partitionId, StreamBlock *block) {
...
+    }
+    if (action == DR_RECORD_INSERT || action == DR_RECORD_UPDATE) {
+        assert(existingTableForInsert);
+        assert(newTableForInsert);
```
nshi: Similar question here for update.

src/ee/indexes/tableindex.h
```diff
@@ -531,6 +538,7 @@ class TableIndex
     virtual bool replaceEntryNoKeyChangeDo(const TableTuple &destinationTuple,
                                          const TableTuple &originalTuple) = 0;
     virtual bool existsDo(const TableTuple* values) const = 0;
+    virtual bool existsDo(const TableTuple* values, TableTuple *conflictTuple) const = 0;
```
nshi: This is very similar to `moveToKey()`. The usage sounds like they are the same. Why not just call `moveToKey()`?

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -103,7 +115,16 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
 
             ReferenceSerializeInputLE rowInput(rowData, rowLength);
             tempTuple.deserializeFromDR(rowInput, pool);
-            table->insertPersistentTuple(tempTuple, true);
+            try {
+                table->insertPersistentTuple(tempTuple, true);
+            } catch (ConstraintFailureException &e) {
+                if (isActiveActiveDREnabled && table->isDREnabled()) {
```
nshi: The check for `table->isDREnabled()` is redundant here. If a table is not DRed, it won't generate any binary log.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -127,9 +148,29 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
 
             TableTuple deleteTuple = table->lookupTupleByValues(tempTuple);
             if (deleteTuple.isNullTuple()) {
+                if (isActiveActiveDREnabled && table->isDREnabled()) {
```
nshi: Same here, `table->isDREnabled()` is redundant.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -127,9 +148,29 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+                        continue;
+                    }
+                }
                 throwSerializableEEException("Unable to find tuple for deletion: binary log type (%d), DR ID (%jd), unique ID (%jd), tuple %s\n",
-                                             type, (intmax_t)sequenceNumber, (intmax_t)uniqueId, tempTuple.debug(table->name()).c_str());
+                                                 type, (intmax_t)sequenceNumber, (intmax_t)uniqueId, tempTuple.debug(table->name()).c_str());
+            }
+
+            // we still run in risk of having timestamp mismatch, need to check.
+            if (isActiveActiveDREnabled && table->isDREnabled()) {
```
nshi: And here.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,14 +197,55 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
 
             TableTuple oldTuple = table->lookupTupleByValues(tempTuple);
             if (oldTuple.isNullTuple()) {
+                if (isActiveActiveDREnabled && table->isDREnabled()) {
```
nshi: And here.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,14 +197,55 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+                    // create tuple to accommodate new row
+                    TableTuple oldTuple(table->schema());
```
nshi: It's somewhat confusing here with the new `oldTuple` here shadowing the `oldTuple` defined on line 198. Proper names should be given to each one of them to avoid confusion.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,14 +197,55 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+                    boost::shared_array<char> data = boost::shared_array<char>(new char[tempTuple.tupleLength() + TUPLE_HEADER_SIZE]);
+                    oldTuple.move(data.get());
+                    oldTuple.copyForPersistentInsert(tempTuple, pool);
+                    ReferenceSerializeInputLE newRowInput(newRowData, newRowLength);
+                    tempTuple.deserializeFromDR(newRowInput, pool);
```
nshi: Since the new row is going to be deserialized in all cases, why not deserialize it once above the checks and reuse it?

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,14 +197,55 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+                        continue;
+                    }
+                }
                 throwSerializableEEException("Unable to find tuple for update: binary log type (%d), DR ID (%jd), unique ID (%jd), tuple %s\n",
-                                             type, (intmax_t)sequenceNumber, (intmax_t)uniqueId, tempTuple.debug(table->name()).c_str());
+                                         type, (intmax_t)sequenceNumber, (intmax_t)uniqueId, tempTuple.debug(table->name()).c_str());
+            }
+
+            // Timestamp mismatch conflict
+            if (isActiveActiveDREnabled && table->isDREnabled()) {
```
nshi: Redundant check of `table->isDREnabled()`.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,14 +197,55 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+                NValue remoteHiddenColumn = tempTuple.getHiddenNValue(table->getDRTimestampColumnIndex());
+                int64_t remoteTimestamp = ExecutorContext::getDRTimestampFromHiddenNValue(remoteHiddenColumn);
+                if (localTimestamp != remoteTimestamp) {
+                    // create tuple to accommodate new row
+                    TableTuple expectedTuple(table->schema());
+                    boost::shared_array<char> data = boost::shared_array<char>(new char[tempTuple.tupleLength()]);
+                    expectedTuple.move(data.get());
+                    expectedTuple.copyForPersistentInsert(tempTuple, pool);
+                    ReferenceSerializeInputLE newRowInput(newRowData, newRowLength);
+                    tempTuple.deserializeFromDR(newRowInput, pool);
```
nshi: Same here, deserialize it once.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,14 +197,55 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
             }
 
             ReferenceSerializeInputLE newRowInput(newRowData, newRowLength);
             tempTuple.deserializeFromDR(newRowInput, pool);
 
-            table->updateTupleWithSpecificIndexes(oldTuple, tempTuple, table->allIndexes());
+            try {
+                table->updateTupleWithSpecificIndexes(oldTuple, tempTuple, table->allIndexes());
+            } catch (ConstraintFailureException &e) {
+                if (isActiveActiveDREnabled && table->isDREnabled()) {
```
nshi: Redundant `table->isDREnabled()`.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -292,22 +375,108 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+   */
+static void findConflictTuple(Table *table, TableTuple *searchTuple, TableTuple *expectedTuple, std::vector< LabeledTableTuple > &conflictRows) {
+    boost::unordered_set<char*> redundancyFilter;
+    BOOST_FOREACH(TableIndex* index, table->allIndexes()) {
+        if (index->isUniqueIndex()) {
+            TableTuple conflictTuple;
+            if (index->exists(searchTuple, &conflictTuple)) {
+                if (expectedTuple && expectedTuple->equals(conflictTuple)) {
+                    continue; // skip the expected tuple
+                } else if (searchTuple->equals(conflictTuple)) {
```
nshi: What if there is a tuple that's exactly the same as the search tuple? It should be reported to the resolver, too.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -292,22 +375,108 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+    TableTuple tempTuple = outputTable->tempTuple();
+    NValue hiddenValue = tupleToBeWrote->getHiddenNValue(drTable->getDRTimestampColumnIndex());
+    tempTuple.setNValue(0, ValueFactory::getTinyIntValue(rowType));
+    tempTuple.setNValue(1, ValueFactory::getTinyIntValue(actionType));
+    tempTuple.setNValue(2, ValueFactory::getTinyIntValue(conflictType));
+    tempTuple.setNValue(3, ValueFactory::getTinyIntValue(conflictOnPKType));
+    switch (rowType) {
+    case EXISTING_ROW:
+    case EXPECTED_ROW:
+        tempTuple.setNValue(4, ValueFactory::getTinyIntValue(KEEP_ROW));   // decision
```
nshi: Probably worth commenting that these are the default decisions, not final decisions.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -292,22 +375,108 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
+        break;
+    case NEW_ROW:
+        tempTuple.setNValue(4, ValueFactory::getTinyIntValue(DELETE_ROW));     // decision
+        break;
+    default:
+        break;
+    }
+    tempTuple.setNValue(5, ValueFactory::getTinyIntValue((ExecutorContext::getClusterIdFromHiddenNValue(hiddenValue))));    // clusterId
+    tempTuple.setNValue(6, ValueFactory::getBigIntValue(ExecutorContext::getDRTimestampFromHiddenNValue(hiddenValue)));     // timestamp
+    tempTuple.setNValue(7, ValueFactory::getTinyIntValue(NOT_DIVERGE));
```
nshi: Worth commenting that this field may be set by the resolver.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -319,6 +488,116 @@ void BinaryLogSink::validateChecksum(uint32_t checksum, const char *start, const
...
+                                                                                      insertConflict,
+                                                                                      existingTableForInsert.get(),
+                                                                                      newTableForInsert.get());
+
+    TableTuple tempTuple(conflictExportTable->schema());
+    if (deleteConflict != NO_CONFLICT) {
+        if (existingTuple) {
+            TableIterator iterator = existingTableForDelete->iterator();
+            iterator.next(tempTuple);
+            DRRowDecision decision = static_cast<DRRowDecision>(ValuePeeker::peekTinyInt(tempTuple.getNValue(4)));
```
nshi: The metadata column indices in the DR export table should be made constants to avoid potential errors if we ever change the schema.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -319,6 +488,116 @@ void BinaryLogSink::validateChecksum(uint32_t checksum, const char *start, const
...
+            if (decision == DELETE_ROW) {
+                drTable->deleteTuple(*existingTuple, true);
+            }
+        }
+    }
+    if (insertConflict != NO_CONFLICT) {
+        TableIterator iterator = existingTableForInsert->iterator();
+        for (int i = 0; iterator.next(tempTuple); i++) {
+            DRRowDecision decision = static_cast<DRRowDecision>(ValuePeeker::peekTinyInt(tempTuple.getNValue(4)));
+            if (decision == DELETE_ROW) {
```
nshi: Based on the discussion on Friday, it's worth asserting here that if there are both delete conflict and insert conflict in an update, it should never delete any existing rows if the delete conflict is not resolved.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -319,6 +488,116 @@ void BinaryLogSink::validateChecksum(uint32_t checksum, const char *start, const
...
+                drTable->deleteTuple(*existingRows[i].first.get(), true);
+            }
+        }
+        iterator = newTableForInsert->iterator();
+        iterator.next(tempTuple);
+        DRRowDecision decision = static_cast<DRRowDecision>(ValuePeeker::peekTinyInt(tempTuple.getNValue(4)));
+        if (decision == KEEP_ROW) {
```
nshi: Also worth asserting here that if the existing rows are kept, the decision for the new row should **not** be `KEEP_ROW`.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -319,6 +488,116 @@ void BinaryLogSink::validateChecksum(uint32_t checksum, const char *start, const
...
+        existingTableForDelete.get()->deleteAllTuples(true);
+    }
+    if (expectedTableForDelete.get()) {
+        expectedTableForDelete.get()->deleteAllTuples(true);
+    }
+    if (existingTableForInsert.get()) {
+        existingTableForInsert.get()->deleteAllTuples(true);
+    }
+    if (newTableForInsert.get()) {
+        newTableForInsert.get()->deleteAllTuples(true);
```
nshi: I wonder if we can save a copy if the tuples were shallow copied.

src/ee/storage/ConstraintFailureException.h
```diff
@@ -55,6 +55,9 @@ class ConstraintFailureException: public SQLException {
 
     virtual const std::string message() const;
     virtual ~ConstraintFailureException();
+
+    TableTuple* getConflictTuple() { return &m_tuple; }
+    TableTuple* getOriginalTuple() { return &m_otherTuple; }
```
nshi: Const correctness.

src/ee/storage/table.cpp
```diff
@@ -238,6 +238,7 @@ int Table::getApproximateSizeToSerialize() const {
  * performance critical path if table is large.
  */
 size_t Table::getAccurateSizeToSerialize(bool includeTotalSize) {
+    std::cout << "getAccurateSizeToSerialize() being called" << std::endl;
```
nshi: Debug output should be removed.

src/ee/storage/table.cpp
```diff
@@ -257,8 +258,10 @@ size_t Table::getAccurateSizeToSerialize(bool includeTotalSize) {
 }
 
 size_t Table::getColumnHeaderSizeToSerialize(bool includeTotalSize) const {
+    std::cout << "getColumnHeaderSizeToSerialize() being called" << std::endl;
```
nshi: Debug output should be removed.

src/ee/storage/table.cpp
```diff
@@ -257,8 +258,10 @@ size_t Table::getAccurateSizeToSerialize(bool includeTotalSize) {
...
     // table size
     size_t bytes = includeTotalSize ? sizeof(int32_t) : 0;
+    std::cout << "getColumnHeaderSizeToSerialize &m_columnHeaderData=" << &m_columnHeaderData << std::endl;
```
nshi: Debug output should be removed.

src/ee/execution/JNITopend.cpp
luyangco: Here is problematic, I'm going to rewrite the assertions.

src/ee/indexes/tableindex.h
luyangco: The only difference is `existsDo(const TableTuple* values, TableTuple *conflictTuple)` takes a whole tuple as parameter, and `bool moveToKey(const TableTuple *searchKey, IndexCursor& cursor)` takes the key tuple as the input parameter. 

Maybe I should make `existsDo()` as a wrapper and call `moveKey()` internally.

src/ee/storage/BinaryLogSink.cpp
luyangco: Will delete it.

src/ee/storage/BinaryLogSink.cpp
luyangco: Will rename it to `expectedTuple`.

src/ee/storage/BinaryLogSink.cpp
luyangco: Will do.

src/ee/indexes/tableindex.h
nshi: I would prefer not expanding the public API just for the convenience of passing in a slightly different parameter but essentially does the same thing.

src/ee/storage/BinaryLogSink.cpp
luyangco: Will do it.  My first thought was that we should not include the new tuple itself, but in this case apparently the new tuple hasn't been in the table yet.

src/ee/storage/BinaryLogSink.cpp
luyangco: Will comment on both this place and above.

src/ee/storage/BinaryLogSink.cpp
luyangco: Yes I will do it. Actually it took me some time to debug when I accidentally changed the schema.

src/ee/storage/BinaryLogSink.cpp
luyangco: Good suggestion.

src/ee/storage/BinaryLogSink.cpp
luyangco: What we can save here is not copying all non-inlined data to temp table (just copy the pointer to non-inlined data), for the inlined data it is still copied when we call `tempTable->insertTuple()`.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -319,6 +496,128 @@ void BinaryLogSink::validateChecksum(uint32_t checksum, const char *start, const
...
+            }
+        }
+
+        newTableForInsert.reset(TableFactory::getCopiedTempTable(0, NEW_TABLE, conflictExportTable, NULL));
+        if (newTuple) {
+           createConflictExportTuple(newTableForInsert.get(), drTable, pool, newTuple, NOT_CONFLICT_ON_PK, actionType, insertConflict, NEW_ROW);
+        }
+    }
+
+    bool diverge = ExecutorContext::getExecutorContext()->getTopend()->reportDRConflict(static_cast<int32_t>(UniqueId::pid(uniqueId)),
```
afrogers: This is switched from what I have in Java; the API returns `false` if it's flagged divergent, `true` otherwise

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -319,6 +496,128 @@ void BinaryLogSink::validateChecksum(uint32_t checksum, const char *start, const
...
+
+    TableTuple tempTuple(conflictExportTable->schema());
+    if (deleteConflict != NO_CONFLICT) {
+        if (existingTuple) {
+            TableIterator iterator = existingTableForDelete->iterator();
+            iterator.next(tempTuple);
+            DRRowDecision decision = static_cast<DRRowDecision>(ValuePeeker::peekTinyInt(tempTuple.getNValue(DR_ROW_DECISION_COLUMN_INDEX)));
+            // It should never delete any existing rows if conflict is not resolved.
+            if (diverge) {
+                assert(decision != DELETE_ROW);
```
afrogers: Nitpick: wrap check within the assert. Granted, it will compile out anyway, but `assert(!(diverge && decision == DELETE_ROW))` makes it clear that it's the intent.

afrogers:
Merged externally.

PR #2920, opened 2015-10-23
Handle geospatial data from the VMC.

BillWhite:
Both sqlcmd and the VMC parameters to the server to use with stored
procedures.  For sqlcmd these parameters are serialized PointType and
GeographyValue values, and the user types WKT strings.  For VMC the user
types WKT strings as well, but VMC didn't know how to convert the
strings to geospatial types, and didn't even know the types of
parameters.  So, the strings are sent to the server and converted there.
Geospatial values in tables are serialized as WKT strings and sent back
to the VMC.  So, from the standpoint of the VMC user, the data starts
and end as WKT strings.  This is less than completely desirable.  The
user should be able to send GEOJson as parameters, and the server should
be able to reply in GEOJson values.  But this is not completely
necessary now, and probably needs a second ticket.

### DIFF
--- a/src/frontend/org/voltdb/ParameterConverter.java
+++ b/src/frontend/org/voltdb/ParameterConverter.java
@@ -494,10 +494,36 @@ public class ParameterConverter {
                 throw new VoltTypeException(String.format("deserialize BigDecimal from string failed. (%s to %s)",
                         inputClz.getName(), expectedClz.getName()));
             }
-        } else if (expectedClz == PointType.class && inputClz == PointType.class) {
-            return param;
-        } else if (expectedClz == GeographyValue.class && inputClz == GeographyValue.class) {
-            return param;
+        } else if (expectedClz == PointType.class) {
+            // Is it a point already?  If so, just return it.
+            if (inputClz == PointType.class) {
+                return param;
+            }
+            // Is it a string from which we can construct a point?
+            // If so, return the newly constructed point.
+            if (inputClz == String.class) {
+                try {
+                    PointType pt = PointType.pointFromText((String)param);
+                    return pt;
+                } catch (IllegalArgumentException e) {
+                    throw new VoltTypeException(String.format("deserialize PointType from string failed (string %s)",
+                                                              (String)param));
+                }
+            }
+        } else if (expectedClz == GeographyValue.class) {
+            if (inputClz == GeographyValue.class) {
+                return param;
+            }
+            if (inputClz == String.class) {
+                String paramStr = (String)param;
+                try {
+                    GeographyValue gv = GeographyValue.geographyValueFromText(paramStr);
+                    return gv;
+                } catch (IllegalArgumentException e) {
+                    throw new VoltTypeException(String.format("deserialize GeographyValue from string failed (string %s)",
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -869,6 +879,57 @@ void putJSONRep(int columnIndex, JSONStringer js) throws JSONException {
         }
     }
 
+    /**
```
wolffcm: Is there a reason that these methods can't be in `PointType` and `GeographyValue`, instead of here?  Also, these methods aren't actually called from anywhere.  Maybe we should defer submitting them until we've a better handle on ENG-9209?

src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -494,10 +494,36 @@ else if (expectedClz == BigDecimal.class) {
                 throw new VoltTypeException(String.format("deserialize BigDecimal from string failed. (%s to %s)",
                         inputClz.getName(), expectedClz.getName()));
             }
-        } else if (expectedClz == PointType.class && inputClz == PointType.class) {
-            return param;
-        } else if (expectedClz == GeographyValue.class && inputClz == GeographyValue.class) {
-            return param;
+        } else if (expectedClz == PointType.class) {
```
wolffcm: Since you're changing this general method, it would be good to add tests to the regressionsuite tests TestPointType and TestGeographyValue.

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -66,6 +66,10 @@ public GeographyValue(String wkt) {
         }
     }
 
+    public static GeographyValue geographyValueFromText(String text) {
```
wolffcm: I like this, thanks for adding it.

wolffcm:
Also, we do have GEB tests for VMC.  John Crawford manages them.  There should be tests for geo-typed parameters there.  I suspect it won't be trivial to add, so we should probably make a ticket for it.

wolffcm:
That's all my comments this round.

src/frontend/org/voltdb/VoltTableRow.java
BillWhite: My first thought was to put them in the class definition.  But they don't really need any private state, and they really didn't seem like they were fundamental to the type.  So I put them here.  I decided not to use them, but I suspect we will need them, and it will save me an hour of looking at the geojson and JSONStringer documentation to leave them here, as I will forget about those two things.  So I left them in in a comment.

src/frontend/org/voltdb/ParameterConverter.java
BillWhite: That's a good thought.  I'll do it.

src/frontend/org/voltdb/VoltTableRow.java
wolffcm: I guess that most of the JSON conversion happens here, so I can see your point.

BillWhite:
Closing to squash this.

PR #2921, opened 2015-10-23
Don't print too many stack traces.

BillWhite:
PlanningErrorExceptions and HSQLParseExcptions are extected behavior in
the planner.  So, we don't want stack traces for them.  Also, log to the
COMPILE logger, not the HOST logger.

I'm not 100% sure this solves the problem.  I'm not sure how to reproduce it.

### DIFF
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -47,6 +47,7 @@ import org.voltdb.utils.Encoder;
  */
 public class PlannerTool {
     private static final VoltLogger hostLog = new VoltLogger("HOST");
+    private static final VoltLogger compileLog = new VoltLogger("COMPILE");
 
     private final Database m_database;
     private final Cluster m_cluster;
@@ -109,7 +110,7 @@ public class PlannerTool {
     }
 
     private void logException(Exception e, String fmtLabel) {
-        hostLog.error(fmtLabel + ": ", e);
+        compileLog.error(fmtLabel + ": ", e);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -131,8 +132,16 @@ public synchronized CompiledPlan planSqlCore(String sql, StatementPartitioning p
...
         }
         catch (Exception e) {
-            logException(e, "Error compiling query");
-            throw new RuntimeException("Error compiling query: " + e.toString() + " (Stack trace has been logged).", e);
+            /*
+             * Don't log PlanningErrorExceptions or HSQLParseExceptions, as they
+             * are at least somewhat expected.
+             */
+            String loggedMsg = "";
+            if (!(e instanceof PlanningErrorException || e instanceof HSQLParseException)) {
```
wolffcm: You can express this `instanceof` logic a little more concisely using Java's catch syntax:
```Java
catch (PlanningErrorException|HSQLParseException e) {
 // do nothing
} catch (Exception e) {
  // ...
}

File: src/frontend/org/voltdb/compiler/PlannerTool.java [hunk 2]
```diff
@@ -47,6 +47,7 @@
  */
 public class PlannerTool {
     private static final VoltLogger hostLog = new VoltLogger("HOST");
+    private static final VoltLogger compileLog = new VoltLogger("COMPILE");
```
wolffcm: It's really a planning log, isn't it?  When I see "compile" I think of the messages that we get from compiling DDL, and I don't think that your change affects them, does it?

wolffcm:
As for testing, if these changes qualify, I think it's fine to merge to master and just verify that the apprunner test is now passing.

src/frontend/org/voltdb/compiler/PlannerTool.java
wolffcm: Or maybe that catch syntax isn't appropriate here.  I'm always looking for places that I can use it.

src/frontend/org/voltdb/compiler/PlannerTool.java
BillWhite: Well, there is a COMPILER log and not a PLANNER log. If we add a PLANNER log we need
to change the log4j configuration file, which means that customers may need to do the same
thing, or at least merge our change into their configuration file.  It didn't seem important
enough to make that change.

BillWhite:
Do you know how to run the apprunner test (or tests?)  Phil doesn't.

src/frontend/org/voltdb/compiler/PlannerTool.java
wolffcm: Oh, I see.  You weren't creating a new category of log message here.  What you've done here makes sense.

wolffcm:
I don't know either.  Someone must know.  Maybe John C or Ruth know?

PR #2924, opened 2015-10-26
Remove unused table name and table size parameters

wolffcm:
This removes unused parameters from our Java/C++ interface for reporting long-running queries.  Among them is a table name parameter of type std::string which may be implicated in the crash described in ENG-9290.

Xin recommended doing this in ENG-8903, I'll close that ticket as a duplicate.

### DIFF
--- a/src/ee/common/Topend.cpp
+++ b/src/ee/common/Topend.cpp
@@ -27,9 +27,12 @@ namespace voltdb {
         return 0;
     }
 
-    int64_t DummyTopend::fragmentProgressUpdate(int32_t batchIndex, std::string planNodeName,
-            std::string targetTableName, int64_t targetTableSize, int64_t tuplesFound,
-            int64_t currMemoryInBytes, int64_t peakMemoryInBytes) {
+    int64_t DummyTopend::fragmentProgressUpdate(
+            int32_t batchIndex,
+            const std::string& planNodeName,
+            int64_t tuplesFound,
+            int64_t currMemoryInBytes,
+            int64_t peakMemoryInBytes) {
         return 1000000000; // larger means less likely/frequent callbacks to ignore
     }
 
--- a/src/ee/common/Topend.h
+++ b/src/ee/common/Topend.h
@@ -46,9 +46,12 @@ class Topend {
     // query to stop.
     // Return 0 if the Topend wants the EE to stop processing the current fragment
     // or the number of tuples the EE should process before repeating this call.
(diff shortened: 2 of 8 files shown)


src/ee/common/Topend.cpp
```diff
@@ -27,9 +27,12 @@ namespace voltdb {
...
 
-    int64_t DummyTopend::fragmentProgressUpdate(int32_t batchIndex, std::string planNodeName,
-            std::string targetTableName, int64_t targetTableSize, int64_t tuplesFound,
-            int64_t currMemoryInBytes, int64_t peakMemoryInBytes) {
+    int64_t DummyTopend::fragmentProgressUpdate(
+            int32_t batchIndex,
+            const std::string& planNodeName,
+            int64_t tuplesFound,
+            int64_t currMemoryInBytes,
+            int64_t peakMemoryInBytes) {
```
BillWhite: Thanks for reformatting the parameters.  It's much easier to read.

BillWhite:
I've nothing more to say about this.

paulmartel:
I think that this change falls short AND misses the mark. I think that the missed mark comes from the focus on the dummy table name and count values pushed out through the jni. These are just constants and probably not the source of the error -- Xin had already de-fanged these.  The "falls short" part is that it does not eliminate the unused m_lastAccessedTable member or the pointless inline code that maintains it.
Also, it does nothing about the m_lastAccessedExec member which is also associated with a string parameter and has more complexity to its maintenance -- making it a more likely suspect in my book.
I would EXPECT a string error if SOMEHOW m_lastAccessedExec failed to be properly reset to null and then an executor/plannode that it was referencing got its plan decached. We were not able to figure out a code path where that could still happen -- the destructor ~ProgressMonitorProxy seems to reliably call VoltDBEngine::pushFinalTuplesProcessedForProgressMonitoring which would only fail to null out the pointer if the serialization of an exception ALSO threw an exception? Could that be it?

I think that we should be erring on the side of slightly less information returned through the jni to avoid reliance on this pointer. Maybe we could replace the cached pointer with a safer cached PlanNodeType member and possibly serialize it as an int rather than a string for extra safety. But, actually, short term, I'm favoring radical surgery -- eliminating all string processing from anywhere near this code path.

wolffcm:
Thanks @paulmartel.  I think the customer's statement might have been a long-running insert into select statement.  It's possible that they hit a constraint violation.  This is total speculation without a reproducible.

I will push another commit that I think addresses the issues you've highlighted here.

wolffcm:
I'll close this pull request and create another one.

PR #2929, opened 2015-10-27
Remove some complexity from long-running fragment progress report API

wolffcm:
Here's another try at simplifying how we log long-running fragment messages.


paulmartel:
This will do the trick, I hope.

PR #2933, opened 2015-10-28
ENG-9057, add k-safe export property to the default setting of DR con…

luyangco:
…flict exporter.

Change-Id: Ifb785e743a583bffa5ee420e48ca42d2a70375e8


nshi:
Wow, you are fast! Looks good to me.

PR #2936, opened 2015-10-29
Calculate non-determinism of aggregates of floating point types.

BillWhite:
If some aggregate functions are applied to floating point parameters the
result may be non-deterministic, due to floating point numeric
differences.  We need to causes errors or warnings in this case.  This
is complicated by the fact that the error messages need to propagated up
from the aggregate call expression through the containing expressions,
into the subquery statement object.  From there it may need to propagate
into a join tree or in ParsedUnionStatement objects if the expression is
in a subquery of a joined table or in a subquery of a union or
intersection expression.

### DIFF
--- a/src/frontend/org/voltdb/compiler/ProcedureCompiler.java
+++ b/src/frontend/org/voltdb/compiler/ProcedureCompiler.java
@@ -17,8 +17,6 @@
 
 package org.voltdb.compiler;
 
-import groovy.lang.Closure;
-
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -60,6 +58,8 @@ import org.voltdb.utils.CatalogUtil;
 import org.voltdb.utils.InMemoryJarfile;
 
 import com.google_voltpatches.common.collect.ImmutableMap;
+
+import groovy.lang.Closure;
 
 /**
  * Compiles stored procedures into a given catalog,
@@ -504,31 +504,7 @@ public abstract class ProcedureCompiler implements GroovyCodeBlockConstants {
 
         procedure.setHasseqscans(procHasSeqScans);
 
-        for (Statement catalogStmt : procedure.getStatements()) {
-            if (catalogStmt.getIscontentdeterministic() == false) {
-                String potentialErrMsg =
-                    "Procedure " + shortName + " has a statement with a non-deterministic result - statement: \"" +
-                    catalogStmt.getSqltext() + "\" , reason: " + catalogStmt.getNondeterminismdetail();
-                // throw compiler.new VoltCompilerException(potentialErrMsg);
-                compiler.addWarn(potentialErrMsg);
-            }
-            else if (catalogStmt.getIsorderdeterministic() == false) {
-                String warnMsg;
-                if (procHasWriteStmts) {
-                    String rwPotentialErrMsg = "Procedure " + shortName +
-                            " is RW and has a statement whose result has a non-deterministic ordering - statement: \"" +
-                            catalogStmt.getSqltext() + "\", reason: " + catalogStmt.getNondeterminismdetail();
-                    // throw compiler.new VoltCompilerException(rwPotentialErrMsg);
-                    warnMsg = rwPotentialErrMsg;
-                }
-                else {
-                    warnMsg = "Procedure " + shortName +
-                        " has a statement with a non-deterministic result - statement: \"" +
-                        catalogStmt.getSqltext() + "\", reason: " + catalogStmt.getNondeterminismdetail();
-                }
-                compiler.addWarn(warnMsg);
-            }
-        }
+        checkForDeterminismWarnings(compiler, shortName, procedure, procHasWriteStmts);
 
         // set procedure parameter types
         CatalogMap<ProcParameter> params = procedure.getParameters();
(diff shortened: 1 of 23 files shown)


BillWhite:
FWIW, the one failing test, TestCLReplayLiveDDLSwitch, passes on my mac when I try it by hand.

src/frontend/org/voltdb/compiler/ProcedureCompiler.java
```diff
@@ -637,6 +613,35 @@ else if (catalogStmt.getIsorderdeterministic() == false) {
...
     }
 
+    private static void checkForDeterminismWarnings(VoltCompiler compiler, String shortName, final Procedure procedure,
+                                         boolean procHasWriteStmts) {
+        for (Statement catalogStmt : procedure.getStatements()) {
+            if (catalogStmt.getIscontentdeterministic() == false) {
+                String potentialErrMsg =
+                    "Procedure " + shortName + " has a statement with a non-deterministic result - statement: \"" +
+                    catalogStmt.getSqltext() + "\" , reason: " + catalogStmt.getNondeterminismdetail();
+                // throw compiler.new VoltCompilerException(potentialErrMsg);
```
wolffcm: You can delete this commented out code.

src/frontend/org/voltdb/compiler/ProcedureCompiler.java
```diff
@@ -637,6 +613,35 @@ else if (catalogStmt.getIsorderdeterministic() == false) {
...
+                compiler.addWarn(potentialErrMsg);
+            }
+            else if (catalogStmt.getIsorderdeterministic() == false) {
+                String warnMsg;
+                if (procHasWriteStmts) {
+                    String rwPotentialErrMsg = "Procedure " + shortName +
+                            " is RW and has a statement whose result has a non-deterministic ordering - statement: \"" +
+                            catalogStmt.getSqltext() + "\", reason: " + catalogStmt.getNondeterminismdetail();
+                    // throw compiler.new VoltCompilerException(rwPotentialErrMsg);
```
wolffcm: You can delete this commented out code.

BillWhite: Ok.  That was there before, though I factored this out into a separate procedure.  I wanted to use it with single statement procedures, but apparently it's not necessary there.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -58,7 +58,37 @@
...
+     * function applied to a floating point expression.
+     */
+    private String m_contentDeterminismMessage = null;
+
+    /**
+     * Note that this expression is inherently non-deterministic. This may be
+     * called if the expression is already known to be non-deterministic, even
+     * if the value is false, because we are careful to never go from true to
+     * false here. Perhaps we should concatenate the messages. But since we only
+     * have one now it would result in unnecessary duplication.
```
wolffcm: I guess in the future you could have a set of messages to avoid duplication, but this seems fine for now.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -58,7 +58,37 @@
     protected VoltType m_valueType = null;
     protected int m_valueSize = 0;
     protected boolean m_inBytes = false;
+    /*
+     * We set this to true iff the expression has a non-deterministic operation.
```
wolffcm: Minor nit: you don't set it to true, you set it to a message describing the operation.

src/frontend/org/voltdb/planner/CompiledPlan.java
```diff
@@ -168,10 +190,15 @@ public boolean hasLimitOrOffset() {
     }
 
     /**
-     * Accessor for description of plan non-determinism.
+     * Accessor for description of plan non-determinism. Note that we prefer the
+     * content determinism message to the rootPlanGraph's message.
```
wolffcm: Why do we prefer the former over the latter?  How are they different?

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -93,11 +93,15 @@
...
+        public final String m_isContentDeterministic;
 
-        public ParsedResultAccumulator(boolean orderIsDeterministic, boolean hasLimitOrOffset)
+        public ParsedResultAccumulator(boolean orderIsDeterministic,
+                                       boolean hasLimitOrOffset,
+                                       String isContentDeterministic)
         {
             m_orderIsDeterministic = orderIsDeterministic;
             m_hasLimitOrOffset  = hasLimitOrOffset;
+            m_isContentDeterministic = isContentDeterministic;
```
wolffcm: Don't you want to assign to `m_isContentDeterministic` only if it's null?  Otherwise an expected warning might disappear.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -50,30 +51,258 @@
...
+    }
+
+    /**
+     * Test that a statement procedure with an expression which has a
+     * subexpression which is an aggregate whose type is float and is in a
+     * subquery causes a compilation error.
+     *
+     * @throws Exception
+     */
+    public void testAggreagtesOfFloatInLeftOfJoin() throws Exception {
```
wolffcm: Oops, a typo in the word "aggregates," here and in a couple other places.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
```diff
@@ -157,7 +398,7 @@ public void testSimple() throws Exception {
         assertTrue(foundLineMatching(lines, ".*\\[WRITE].*NondeterministicRWProc.*"));
         assertTrue(foundLineMatching(lines, ".*\\[WRITE].*DeterministicRWProc.*"));
         assertTrue(foundLineMatching(lines, ".*\\[TABLE SCAN].*select ival, sval from indexed_replicated_blah.*"));
-
+        // Find FloatingUpdate error lines.
```
wolffcm: Is this a comment indicating that you want to add some tests here?

src/frontend/org/voltdb/compiler/ProcedureCompiler.java
BillWhite: Ok.  It was there before, and I didn't know if it was important or not.

src/frontend/org/voltdb/planner/PlanAssembler.java
BillWhite: This is the constructor.  So, m_isContentDeterministic is not initialized yet.  I guess it's null because that's somehow the default for java, but I don't know if member data have their default value before they are initialized or not.  In any case, this little class is just a triple of values, so I always want to set it.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
BillWhite: Got it.  Copy and paste strikes again.

tests/frontend/org/voltdb/compiler/TestVoltCompilerAnnotationsAndWarnings.java
BillWhite: No, it means I moved some tests out into the compilation failure tests, and forgot to remove
the comment.  Everything in this test looks for determinism state of stored procedures which
compile just fine.  To write a test for a stored procedure which did not compile because it was non-deterministic requires calling the function testCompilationFailure above.

wolffcm:
That's all my comments.

wolffcm:
I'm okay with you merging this when you're ready.

src/frontend/org/voltdb/planner/CompiledPlan.java
BillWhite: Well, I have two messages and only one slot for them.  I suppose I could concatenate them somehow, but it wasn't clear if that was a better idea or not.

BillWhite:
Thanks.  These are all cosmetic changes, right?  Some commented out code is deleted and some spelling errors are changed.  I'll make this changes, but I don't think I need to run this through Jenkins again, do I?

src/frontend/org/voltdb/planner/CompiledPlan.java
wolffcm: How are the contents of the two messages different?

wolffcm:
If you're only making cosmetic changes I would just be sure that your branch still builds.

BillWhite:
Oh, sure.  I realized after I left the comment that Jenkins is going to build whether I wait or not, so I might as well wait for it.  It's not like TMW, where a build cycle can take 12-18 hours.

PR #2937, opened 2015-10-29
Implement a CONTAINS function for point-in-polygon

wolffcm:
Now that we have S2 linked and have implemented the SQL types, not a lot of our own code was needed for this.

We'll be re-visiting how we instantiate S2Polygons, so the code here that does it temporary.

I'll hold off on merging this until we've refreshed the topic branch from master.


BillWhite:
I can review this in about half an hour.

src/ee/common/Geography.hpp
```diff
@@ -215,6 +218,34 @@ class Geography {
         return LoopContainer(m_data);
     }
 
+    std::unique_ptr<S2Polygon> toS2Polygon() const {
```
BillWhite: Is our current C++ standard grand enough to use unique_ptr?  Apparently so.

src/ee/common/Geography.hpp
```diff
@@ -215,6 +218,34 @@ class Geography {
...
+        std::vector<S2Loop*>* loops = new std::vector<S2Loop*>();
+        auto loopIt = this->loops().begin();
+        auto loopEnd = this->loops().end();
+        for (; loopIt != loopEnd; ++loopIt) {
+            std::vector<S2Point> verts;
+            auto vertIt = (*loopIt).begin();
+            // S2 considers the final closing vertex
+            // (which should be identical to the first vertex)
+            // to be implicit.
+            auto vertEnd = (*loopIt).end() - 1;
```
BillWhite: Do we do something similar?  I don't see a check for this in the code, but I've only glanced at it.

BillWhite:
Well, that was simpler than I thought it would be.  That's all I have here.

We should have a ticket, maybe a timebox, to create better test examples for unit tests, both Java and C++.

src/ee/common/Geography.hpp
wolffcm: Right now in the Java client class and in the EE storage, we store the redundant last vertex.  Maybe we shouldn't.

src/ee/common/Geography.hpp
wolffcm: I think the only magic item needed to support unique_ptr is rvalue references, which is in gcc 4.4.  (As is "auto".)  I took a chance and it seems to be working.

wolffcm:
I squash-merged this in a separate branch.

@BillWhite, what did you mean by "better test examples"?  We get some coverage from the S2 unit tests, we'll have a demo app, and also a benchmarking JUnit test (ENG-9327) that will randomly generate (rectangular, at least) polygons.  Do you think there's more that we need to cover?

BillWhite:
I'm thinking non-rectangular regions, overlapping regions and perhaps
somewhat irregular regions,
like states or islands.

PR #2947, opened 2015-10-29
Eng 9277 per node conflict export table

luyangco:


nshi:
Cherry-picked onto Walter's branch. Will land it from there.

PR #2949, opened 2015-10-30
Min and Max of float are deterministic.

BillWhite:
The aggregate functions min and max of floating point expressions are
deterministic.  In the previous commit, 758efd3, we incorrectly called
them non-deterministic.

https://issues.voltdb.com/browse/ENG-9358


hhakimi53:
looks good.

PR #2950, opened 2015-10-30
ENG-9283 & ENG-9362 allow ad hoc ddl after an index with timestamp in…

paulmartel:
… predicate

ENG-9283 is fixed by fixing conversion to SQL of a timestamp-valued ConstantValueExpression
ENG-9362 is fixed by making hsql preserve index predicates on table redefinition.

The DDLCompiler changes are intended to limit visibility of DDLCompiler functions and variables.
I liberally applied "private", "static" and "final" modifiers and removed "public" wherever possible.
This helped me make sense of how/when these other pieces of code came into play.

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -17,9 +17,7 @@
 
 package org.voltdb.compiler;
 
-import java.io.FileReader;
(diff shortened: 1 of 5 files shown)


wolffcm:
Looks good to me.

wolffcm:
Also, I like the refactoring you did to DDLCompiler.

PR #2952, opened 2015-10-30
Eng 9360 - conflict detection and resolution

luyangco:
Passed manual test.

### DIFF
--- a/src/ee/execution/JNITopend.cpp
+++ b/src/ee/execution/JNITopend.cpp
@@ -479,39 +479,38 @@ int64_t JNITopend::pushDRBuffer(int32_t partitionId, StreamBlock *block) {
 int JNITopend::reportDRConflict(int32_t partitionId, int32_t remoteClusterId, int64_t remoteTimestamp, std::string tableName, DRRecordType action,
         DRConflictType deleteConflict, Table *existingTableForDelete, Table *expectedTableForDelete,
         DRConflictType insertConflict, Table *existingTableForInsert, Table *newTableForInsert) {
+
     // prepare tablename
     jstring tableNameString = m_jniEnv->NewStringUTF(tableName.c_str());
 
     // prepare input buffer for delete conflict
     jobject existingRowsBufferForDelete = NULL;
-    char* existingArrayForDelete = NULL;
+    boost::shared_ptr<char> existingArrayForDelete;
     if (existingTableForDelete) {
         size_t serializeSize = existingTableForDelete->getAccurateSizeToSerialize(false);
-        existingArrayForDelete = new char[serializeSize];
-        ReferenceSerializeOutput conflictSerializeOutput(existingArrayForDelete, serializeSize);
+        existingArrayForDelete = boost::shared_ptr<char>(new char[serializeSize]);
+        ReferenceSerializeOutput conflictSerializeOutput(existingArrayForDelete.get(), serializeSize);
         existingTableForDelete->serializeToWithoutTotalSize(conflictSerializeOutput);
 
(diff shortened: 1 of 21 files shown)


src/ee/execution/JNITopend.cpp
```diff
@@ -476,52 +476,84 @@ int64_t JNITopend::pushDRBuffer(int32_t partitionId, StreamBlock *block) {
...
+        existingArrayForDelete = boost::shared_ptr<char>(new char[serializeSize]);
+        ReferenceSerializeOutput conflictSerializeOutput(existingArrayForDelete.get(), serializeSize);
+        existingTableForDelete->serializeToWithoutTotalSize(conflictSerializeOutput);
+
+        existingRowsBufferForDelete = m_jniEnv->NewDirectByteBuffer(static_cast<void*>(existingArrayForDelete.get()),
+                                                    static_cast<int32_t>(serializeSize));
+        if (existingRowsBufferForDelete == NULL) {
+            m_jniEnv->ExceptionDescribe();
+           throw std::exception();
+        }
```
nshi: This block of code used to be wrapped inside a method, any reason it's being duplicated now?

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -459,6 +477,7 @@ void BinaryLogSink::exportDRConflict(Table *exportTable, bool resolved, TempTabl
     if (expectedTableForDelete) {
         TableIterator iterator = expectedTableForDelete->iterator();
         while (iterator.next(tempTuple)) {
+            //TODO: decision column on expected tuple is meanless.
```
nshi: Why is this a TODO? It reads just like a normal comment.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -45,9 +45,9 @@
         NEW_ROW
     }
 
-    public static enum DRRowDecision {
-        KEEP_ROW,
-        DELETE_ROW;
+    public static enum DRConflictResolutionFlag {
+        ACCEPT_CHANGE,
+        CONVERGENT
```
nshi: It feels weird that these two things are in the same enum. Don't they have overlaps?

tests/ee/storage/table_and_indexes_test.cpp
```diff
@@ -417,7 +417,7 @@ TEST_F(TableAndIndexTest, DrTest) {
     *reinterpret_cast<int32_t*>(&data.get()[startPos]) = htonl(static_cast<int32_t>(sb->offset()));
     drStream.m_enabled = false;
     districtTable->setDR(false);
-    sink.apply(&data[startPos], tables, &pool, mockEngine);
+    sink.apply(&data[startPos], tables, &pool, mockEngine, 0);
```
nshi: The table_and_indexes_test test suite uses cluster ID 0 as the local cluster ID. The remote cluster ID passed in here should different from that.

src/frontend/org/voltdb/PartitionDRGateway.java
nshi: After reading the code that uses this enum, I think I understand it now. These enums are indices, not values.

src/ee/storage/BinaryLogSink.cpp
luyangco: I'll remove TODO mark, it is just a comment.

src/ee/execution/JNITopend.cpp
luyangco: There is just my laziness not encapsulate them into a function, I'll wrap them.

nshi:
Squash merged externally

PR #2953, opened 2015-10-30
ENG-9350 fixed multi-node cluster latency logger issues

dweiss88:


src/frontend/org/voltdb/utils/LatencyLogger.java
```diff
@@ -125,11 +126,18 @@ public double getValueAtPercentile (double percentile) {
...
+        final String hostname;
         if (args.length != 3) {
-            System.out.println("Usage server port reportIntervalSeconds");
-            return;
-        }
+            if (args.length == 4)
+                hostname = args[3];
+            else {
+                System.out.println("Usage server port reportIntervalSeconds [hostname]");
+                return;
```
manjujames: What is the difference between server and hostname here?

src/frontend/org/voltdb/utils/LatencyLogger.java
```diff
@@ -137,7 +145,7 @@ public static void main(String[] args) throws Exception {
                 throw new NumberFormatException();
         } catch (NumberFormatException e) {
             System.out.println("reportIntervalSeconds should be greater than or equal to 1");
-            System.out.println("Usage server port reportIntervalSeconds");
+            System.out.println("Usage server port reportIntervalSeconds [hostname]");
```
manjujames: If you make a method called usage() that prints out the usage, you don't have to change this in multiple places when you change usage.

src/frontend/org/voltdb/dtxn/LatencyStats.java
```diff
@@ -47,25 +47,26 @@
...
         }
 
         @Override
         public Object next() {
-            return null;
+            if (returnRow) {
+                returnRow = false;
+                return new Object();
+            } else {
+                return null;
```
manjujames: I don't understand what this change is.. I'll talk to you on Monday.

src/frontend/org/voltdb/utils/LatencyLogger.java
dweiss88: server can be designated with an ip address while hostname is a specific name that the cluster recognizes as the identifier for that machine.

manjujames:
Looks good to me now. Feel free to merge when build is blue. Use squash-merge so that the commit on LatencyStats can be avoided.

PR #2955, opened 2015-11-02
Fix plannertester baselines using mergereceive and prevent noisy exce…

paulmartel:
…ptions in hsql describe.


wolffcm:
This looks good to me.  I guess this failure was happening because we didn't run the planner tester tests before merging the merge-receive improvements?  I didn't check the changes to the planner tester baselines, it's hard to make heads or tails of the diffs.

I'm a little confused as to why we would have new HSQL exceptions as a result of the merge-receive work.  The SQL statements being processed wouldn't have changed.  Can you explain?

PR #2960, opened 2015-11-02
ENG-9372, change the "decision" column of conflict export log from th…

luyangco:
…e per-row based "decision" to action

level "decision", the value could be either "A" (Apply remote change) or "R" (Reject remote change).

Change-Id: I65fc08a5990a08d8fa69f7bbb6e3bae5f0089e3a


nshi:
Merged.

PR #2961, opened 2015-11-02
sqlcmdtest.py, timeout.errbaseline: fix sqlcmdtest: Bill's fix for

johnTcrawford:
ENG-9246 added an extra digit to the query-timeout error message (now
showing milliseconds), which broke sqlcmdtest. This fixes that, by
changing the expected message from "1.00 seconds" to "1.00# seconds",
and alters the clean_output method, to ignore that final millisecond
digit, replacing it with "#", so they now match.

### DIFF
--- a/tests/sqlcmd/sqlcmdtest.py
+++ b/tests/sqlcmd/sqlcmdtest.py
@@ -150,7 +150,7 @@ def clean_output(parent, path):
     memory_check_matcher = re.compile(r"""
             ^(WARN:\s)?Strict\sjava\smemory\schecking.*$  # Match the start.
             """, re.VERBOSE)
-    # 2. Allow different latency numbers to be reported like
+    # 2. Allow different latency numbers to be reported, like
     # "(Returned 3 rows in 9.99s)" vs. "(Returned 3 rows in 10.01s)".
     # These both get "fuzzed" into the same generic string "(Returned 3 rows in #.##s)".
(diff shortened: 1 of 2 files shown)


tests/sqlcmd/sqlcmdtest.py
```diff
@@ -160,6 +160,20 @@ def clean_output(parent, path):
                                  # survives as \g<1>
             [0-9]+\.[0-9]+s      # also required, replaced with #.##s
             """, re.VERBOSE)
+    # 3. Allow different query timeout periods to be reported, like
```
BillWhite: Don't you need parenthesis around the "[0-9]+\.[0-9]+s"?  Otherwise, it isn't a group.  I always have to reread about python REs, so maybe I'm wrong.

tests/sqlcmd/sqlcmdtest.py
```diff
@@ -168,6 +182,7 @@ def clean_output(parent, path):
             continue
         cleanedline = memory_check_matcher.sub("", line)
         cleanedline = latency_matcher.sub("\g<1>#.##s", cleanedline)
+        cleanedline = query_timeout_matcher.sub("\g<1>#\g<2>", cleanedline)
```
BillWhite: Doesn't this replace the "rows in" part with #.##s?

BillWhite:
Seems right to me.

PR #2964, opened 2015-11-03
ENG-8426: Make sequential calls to fetchRow() constant time per call,…

jhugg:
… not linear.

This makes scanning with fetchrow a linear time thing, but still as much as twice as slow as advancerow.

Also remove an unused VoltTable variable.


src/frontend/org/voltdb/VoltTable.java
```diff
@@ -128,13 +127,18 @@
...
 
-    static final AtomicInteger expandCountDouble = new AtomicInteger(0);
-
     boolean m_readOnly = false;
     int m_rowStart = -1; // the beginning of the row data (points to before the row count int)
     int m_rowCount = -1;
     int m_colCount = -1;
 
+    // non-positive value that probably shouldn't be -1 to avoid off-by-one errors
+    static final int NO_MEMOIZED_ROW_OFFSET = Integer.MIN_VALUE;
```
wolffcm: All these new members could be made private.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -712,16 +716,23 @@ public final int getColumnIndex(String name) {
...
+
+        // check bounds
         if ((index < 0) || (index >= m_rowCount)) {
             throw new IndexOutOfBoundsException("index = " + index + "; rows = " + m_rowCount);
         }
 
-        int pos = m_rowStart + 4;
-        for (int i = 0; i < index; i++) {
+        // if no memoized value or looking in front of the memoized value, reset
+        if ((m_memoizedRowOffset == NO_MEMOIZED_ROW_OFFSET) || (index < m_memoizedRowOffset)) {
```
wolffcm: Can you verify that VoltTable can never shrink?  If it could shrink, and grow again, the memoized buffer offset could point to the middle of a row.

wolffcm:
This looks good to me except for the things I mentioned above.

Note that master is closed right now until we branch for 5.8.

src/frontend/org/voltdb/VoltTable.java
jhugg: VoltTables are always append only. There's no way to shrink them.

src/frontend/org/voltdb/VoltTable.java
jhugg: I will make this change.

jhugg:
Thanks for the review Chris. I will merge once master is open.

PR #2970, opened 2015-11-04
ENG-9099, rename @Statistics selector DR to DRPRODUCER, and make DR a…

luyangco:
… synonym of DRPRODUCER.

Change-Id: Iae04a5510c2f060718f0201b1dcfb88a225a29b0


nshi:
TestStatisticsSuite also needs to be fixed.

PR #2972, opened 2015-11-04
Fix a test error.

BillWhite:
The test TestUnionSuite.testUnionStar had a query which returned order
non-deterministic results.  This sometimes caused test failures.  This
commit works around this by adding an order-by to the query.

https://issues.voltdb.com/browse/ENG-9395.


hhakimi53:
looks good.

BillWhite:
Jenkins is all blue.

PR #2973, opened 2015-11-04
Optimize table truncate path for fewer numbers of rows

hhakimi53:
Changes are to optimize truncate operation where instead of performing table swap of existing table with newly created empty persistent table, logic will perform deletion of all the tuple entries in the table iteratively and retain existing objects associated to table instead of creating afresh.This helps truncate operation for table with few rows to be as fast as when performing delete all rows using predicate that is true (result in deletion of all iteratively internally does).

### DIFF
--- a/src/ee/executors/deleteexecutor.cpp
+++ b/src/ee/executors/deleteexecutor.cpp
@@ -96,7 +96,6 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
     int64_t modified_tuples = 0;
 
     if (m_truncate) {
-        bool truncate = false;
         VOLT_TRACE("truncating table %s...", targetTable->name().c_str());
         // count the truncated tuples as deleted
         modified_tuples = targetTable->visibleTupleCount();
@@ -107,15 +106,8 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
                    (int)targetTable->visibleTupleCount(),
                    (int)targetTable->allocatedTupleCount());
 
-        truncate = targetTable->isTruncateFavorable();
-        if (truncate) {
-            // delete using table swap: undo by table not by each tuple.
-            targetTable->truncateTable(m_engine);
-        }
-        else {
-            // delete tuple by tuple
-            targetTable->deleteAllTuples(true);
-        }
+        // empty the table either by table swap or iteratively deleting tuple-by-tuple
+        targetTable->truncateTable(m_engine);
     }
     else {
         assert(m_inputTable);
--- a/src/ee/executors/deleteexecutor.cpp
+++ b/src/ee/executors/deleteexecutor.cpp
@@ -96,6 +96,7 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
     int64_t modified_tuples = 0;
 
     if (m_truncate) {
+        bool truncate = false;
         VOLT_TRACE("truncating table %s...", targetTable->name().c_str());
         // count the truncated tuples as deleted
         modified_tuples = targetTable->visibleTupleCount();
@@ -106,11 +107,17 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
                    (int)targetTable->visibleTupleCount(),
                    (int)targetTable->allocatedTupleCount());
 
-        // actually delete all the tuples: undo by table not by each tuple.
-        targetTable->truncateTable(m_engine);
+        truncate = targetTable->isTruncateFavorable();
+        if (truncate) {
+            // delete using table swap: undo by table not by each tuple.
+            targetTable->truncateTable(m_engine);
+        }
+        else {
+            // delete tuple by tuple
+            targetTable->deleteAllTuples(true);
+        }
     }
-    else
-    {
+    else {
         assert(m_inputTable);
         assert(m_inputTuple.sizeInValues() == m_inputTable->columnCount());
         assert(targetTuple.sizeInValues() == targetTable->columnCount());
--- a/src/ee/storage/TupleBlock.h
+++ b/src/ee/storage/TupleBlock.h
@@ -101,7 +101,7 @@ public:
     { return ThreadLocalPool::freeExactSizedObject(sizeof(TupleBlock), object); }
 
     double loadFactor() {
-        return m_activeTuples / m_tuplesPerBlock;
+        return (double) m_activeTuples / m_tuplesPerBlock;
     }
 
     inline bool hasFreeTuples() {
--- a/src/ee/storage/TupleBlock.h
+++ b/src/ee/storage/TupleBlock.h
@@ -101,7 +101,7 @@ public:
     { return ThreadLocalPool::freeExactSizedObject(sizeof(TupleBlock), object); }
 
     double loadFactor() {
-        return (double) m_activeTuples / m_tuplesPerBlock;
+        return static_cast <double> (m_activeTuples) / m_tuplesPerBlock;
     }
 
     inline bool hasFreeTuples() {
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -339,13 +339,13 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
         // - tables with indexes
 
         // cut-off for table with no views
-        const double TABLE_LF_CUTOFF_FOR_TRUNC = 0.105666;
+        const double tableLFCutoffForTrunc = 0.105666;
         //cut-off for table with views
-        const double TABLE_WITH_VIEWS_LF_CUTOFF_FOR_TRUNC = 0.015416;
+        const double tableWithViewsLFCutoffForTrunc = 0.015416;
 
         const double blockLoadFactor = m_data.begin()->getValue()->loadFactor();
-        if ((blockLoadFactor <= TABLE_LF_CUTOFF_FOR_TRUNC) ||
-            (m_views.size() > 0 && blockLoadFactor <= TABLE_WITH_VIEWS_LF_CUTOFF_FOR_TRUNC)) {
+        if ((blockLoadFactor <= tableLFCutoffForTrunc) ||
+            (m_views.size() > 0 && blockLoadFactor <= tableWithViewsLFCutoffForTrunc)) {
             return deleteAllTuples(true);
         }
     }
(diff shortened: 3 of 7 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -107,6 +107,12 @@ class SetAndRestorePendingDeleteFlag
     TableTuple &m_target;
 };
 
+
+// cut-off for table with no views
+const double PersistentTable::TABLE_LF_CUTOFF_FOR_TRUNC = 0.105666;
+//cut-off for table with views
+const double PersistentTable::TABLE_WITH_VIEWS_LF_CUTOFF_FOR_TRUNC = 0.015416;
+
```
BillWhite: Were these values determined empirically?

BillWhite:
Seems plausible to me.  You might want to add a bit of commentary about how the
constants TABLE_LF_CUTOFF_FOR_TRUNC and TABLE_WITH_VIEWS_LF_CUTOFF_FOR_TRUNK
were determined, but other than that it seems ok to me.

src/ee/storage/persistenttable.cpp
hhakimi53: The cut-off numbers were obtained based on the delete-truncate tests we ran with under following scenarios:
- Small and medium sized tables
- Table with varying views and how much populated the views
- Table having varying size and multiple varchar columns .
- Table having indexes on it.

src/ee/executors/deleteexecutor.cpp
```diff
@@ -96,6 +96,7 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
     int64_t modified_tuples = 0;
 
     if (m_truncate) {
+        bool truncate = false;
```
paulmartel: This "false" setting has no effect. It is simpler to declare and initialize the variable in one statement where possible. Actually, I'm not convinced that having the local variable AT ALL really helps readability.

src/ee/storage/persistenttable.h
```diff
@@ -658,6 +659,17 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
...
+    // threshold cutoff in terms of block load factor at which truncate is better
+    // than tuple-by-tuple delete. Cut-off values are picked based on worst case
+    // scenarios with intent to improve performance and to avoid performance
+    // regression by not getting too greedy for performance by favoring truncate
+    // truncate instead of tuple-by-tuple delete.
+
+    // cut-off for table with no views
+    static const double TABLE_LF_CUTOFF_FOR_TRUNC;
+    //cut-off for table with views
+    static const double TABLE_WITH_VIEWS_LF_CUTOFF_FOR_TRUNC;
```
paulmartel: These constants could also be hidden in persistenttable.cpp if isTruncateFavorable were inlined into the truncateTable method.

src/ee/storage/persistenttable.h
```diff
@@ -951,6 +963,24 @@ inline TableTuple PersistentTable::lookupTupleForUndo(TableTuple tuple) {
...
+    // is below certain threshold value. the threshold value is
+    // different if the table has views on it
+    if (m_data.size() == 1) {
+        const double blockLoadFactor = m_data.begin()->getValue()->loadFactor();
+        //const double blockLoadFactor = (double) visibleTupleCount() / m_tuplesPerBlock;
+        if ((blockLoadFactor <= TABLE_LF_CUTOFF_FOR_TRUNC) ||
+            (m_views.size() > 0 && blockLoadFactor <= TABLE_WITH_VIEWS_LF_CUTOFF_FOR_TRUNC)) {
+            return false;
+        }
+    } else if (isPersistentTableEmpty() == true) {
```
paulmartel: If the table is empty, we can skip BOTH deleteAllTuples and the table swapping -- another reason to absorb this logic into truncateTable.

src/ee/executors/deleteexecutor.cpp
```diff
@@ -106,11 +107,17 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
...
-        targetTable->truncateTable(m_engine);
+        truncate = targetTable->isTruncateFavorable();
+        if (truncate) {
+            // delete using table swap: undo by table not by each tuple.
+            targetTable->truncateTable(m_engine);
+        }
+        else {
+            // delete tuple by tuple
+            targetTable->deleteAllTuples(true);
+        }
```
paulmartel: A case could be made that all of this complexity could be would be better hidden inside one PersistentTable function which could just be truncateTable -- I think this is the only truncateTable caller. That seems to make more sense than 3 calls on targetTable.

paulmartel:
Over all, this change could be much simpler/smaller if it were isolated into truncateTable.

src/ee/executors/deleteexecutor.cpp
hhakimi53: It was done this way as I thought this will keep truncateTable() call specifically to table swap work only and let delete executor decide which version to call deletAllTuples() or truncateTable(). I will look into consolidating above logic into truncateTable() if that simplifies the logic here. Thnxs

src/ee/storage/persistenttable.h
hhakimi53: Will do that way as it will eliminate deleteAllTuple() call if table is empty.

src/ee/executors/deleteexecutor.cpp
hhakimi53: BinaryLogSink::BinaryLogSink() also calls truncateTable(). Currently looking into it and will follow up with DR team/Ning.

src/ee/storage/persistenttable.cpp
```diff
@@ -316,6 +316,40 @@ void PersistentTable::truncateTableRelease(PersistentTable *originalTable) {
...
 
 void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
+    if (isPersistentTableEmpty() == true) {
+        return;
+    }
+
+    // if is the only tuple-storage block table has, it may be better to truncate
+    // table by iteratively deleting table rows. Evalute if that's the case
+    // based on the block and tuple block load factor
+    if (m_data.size() == 1) {
```
BillWhite: Maybe a better way to word this comment would be "If the table has only one tuple-storage
block, it may be better to truncate the table by iteratively deleting table rows.  Evaluate if this
is the case based on the block and tuple block load factor."

hhakimi53: Ok, will reword the comment. thnxs

src/ee/storage/TupleBlock.h
```diff
@@ -100,7 +100,7 @@ class TupleBlock {
     { return ThreadLocalPool::freeExactSizedObject(sizeof(TupleBlock), object); }
 
     double loadFactor() {
-        return m_activeTuples / m_tuplesPerBlock;
+        return (double) m_activeTuples / m_tuplesPerBlock;
     }
```
BillWhite: Is it better to use a static_cast<double>(m_activeTuples)/m_tuplesPerBlock?  I don't know.
The results are the same, I suppose.

BillWhite:
That's all I have.  Nothing major, just presentation mostly.

src/ee/storage/TupleBlock.h
hhakimi53: Ya, that would be appropriate way to cast in C++ realm (representation wise). Will update it, thnxs

src/ee/storage/persistenttable.cpp
```diff
@@ -316,6 +316,40 @@ void PersistentTable::truncateTableRelease(PersistentTable *originalTable) {
 
 
 void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
+    if (isPersistentTableEmpty() == true) {
+        return;
+    }
+
+    // If the table has only one tuple-storage block, it may be better to truncate
+    // table by iteratively deleting table rows. Evalute if this is the case
+    // based on the block and tuple block load factor
```
jxchinaren: You are checking the size of m_data to decide whether there is on tuple block or not. Why the comment does different saying the load factor?

jxchinaren: My bad. I see you checked them inside of the function. Those two constant factors are really important here.

tests/frontend/org/voltdb/regressionsuites/TestSQLFeaturesNewSuite.java
```diff
@@ -71,12 +73,21 @@ public void testTruncateTable() throws Exception {
...
+
+        // Populate table with large # of rows (using SP) to exercise swap path for truncate also.
+        // Perform row insertion in chunks as there is upper limit on # on calls that be queued
+        // and executed in single SP call.
+        int rowsToInsert = 50000;
+        final int rowsInsertionEachChunk = 10000;
+        for (int rowsInserted = 0; rowsInserted < rowsToInsert;) {
+            // Insert data
+            client.callProcedure("PopulateTruncateTable", rowsInserted + 1, rowsInsertionEachChunk);
+            rowsInserted = rowsInserted + rowsInsertionEachChunk;
```
paulmartel: Readability: Use the idiom for incrementing the loop counter like this in the third component of the "for" statement.

paulmartel:
End of review round.

src/ee/storage/persistenttable.cpp
```diff
@@ -316,6 +316,40 @@ void PersistentTable::truncateTableRelease(PersistentTable *originalTable) {
...
+        // - tables with indexes
+
+        // cut-off for table with no views
+        const double tableLFCutoffForTrunc = 0.105666;
+        //cut-off for table with views
+        const double tableWithViewsLFCutoffForTrunc = 0.015416;
+
+        const double blockLoadFactor = m_data.begin()->getValue()->loadFactor();
+        if ((blockLoadFactor <= tableLFCutoffForTrunc) ||
+            (m_views.size() > 0 && blockLoadFactor <= tableWithViewsLFCutoffForTrunc)) {
```
nshi: It looks like that the second condition will never be true because it will only check for the table-with-view case if the load factor is greater than `tableLFCutoffForTrunc`, which means that it is also greater than `tableWithViewsLFCutoffForTrunc`, so the second condition is going to be false or not checked at all.

nshi:
For tables with a lot of non-inlined strings, I'm not sure this optimization is as efficient as expected. It may potentially has GBs of non-inlined data for tuples that fit into a single tuple block. Deleting them one by one also generates huge DR binary log whereas a table truncation just generate a single record of type TRUNCATE.

PR #2982, opened 2015-11-09
build.xml: removed my kludge of setting the PostgreSQL JDBC .jar file in

johnTcrawford:
project.classpath; it's simpler to just set the CLASSPATH environment
variable of the machine, when needed (including in Jenkins).


paulmartel:
Looks good.

PR #2983, opened 2015-11-09
Eng 9363 geo func

hhakimi53:
# Added geo functions:
- numInteriorRings(): returns number of loops/rings for the given polygon 
- numPoints(): returns number of points the given polygon has. This is sum of number of points on different loops of polygon
- Longitude() and Latitude():  for longtitude and latitude for the given point
# Added junit tests for the above changes

### DIFF
--- a/src/ee/expressions/geofunctions.cpp
+++ b/src/ee/expressions/geofunctions.cpp
@@ -246,4 +246,31 @@ template<> NValue NValue::call<FUNC_VOLT_CONTAINS>(const std::vector<NValue>& ar
     return ValueFactory::getBooleanValue(s2Poly->Contains(s2Point));
 }
 
+template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_NUM_INTERIOR_RINGS>() const {
+	const Geography polygon = getGeography();
+	NValue retVal(VALUE_TYPE_INTEGER);
+	retVal.getInteger() = polygon.numLoops();
+	return retVal;
+}
+
+template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_NUM_POINTS>() const {
+    const Geography polygon = getGeography();
+    NValue retVal(VALUE_TYPE_INTEGER);
+    retVal.getInteger() = polygon.numVertices();
+    return retVal;
+}
+
+template<> NValue NValue::callUnary<FUNC_VOLT_POINT_LATITUDE>() const {
+	const Point point = getPoint();
+	NValue retVal(VALUE_TYPE_DOUBLE);
+	retVal.getDouble() = point.getLatitude();
+	return retVal;
+}
+template<> NValue NValue::callUnary<FUNC_VOLT_POINT_LONGITUDE>() const {
+	const Point point = getPoint();
+	NValue retVal(VALUE_TYPE_DOUBLE);
+	retVal.getDouble() = point.getLongitude();
+	return retVal;
+}
+
 } // end namespace voltdb
--- a/src/ee/expressions/geofunctions.cpp
+++ b/src/ee/expressions/geofunctions.cpp
@@ -247,13 +247,20 @@ template<> NValue NValue::call<FUNC_VOLT_CONTAINS>(const std::vector<NValue>& ar
 }
 
 template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_NUM_INTERIOR_RINGS>() const {
-	const Geography polygon = getGeography();
+	if (isNull()) {
+	    return NValue::getNullValue(VALUE_TYPE_INTEGER);
+	}
+    const Geography polygon = getGeography();
 	NValue retVal(VALUE_TYPE_INTEGER);
-	retVal.getInteger() = polygon.numLoops();
+	// exclude exterior ring
+	retVal.getInteger() = polygon.numLoops() - 1;
 	return retVal;
 }
 
 template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_NUM_POINTS>() const {
+    if (isNull()) {
+        return NValue::getNullValue(VALUE_TYPE_INTEGER);
+    }
     const Geography polygon = getGeography();
     NValue retVal(VALUE_TYPE_INTEGER);
     retVal.getInteger() = polygon.numPoints();
@@ -261,13 +268,19 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_NUM_POINTS>() const {
 }
 
 template<> NValue NValue::callUnary<FUNC_VOLT_POINT_LATITUDE>() const {
+    if (isNull()) {
+        return NValue::getNullValue(VALUE_TYPE_DOUBLE);
+    }
 	const Point point = getPoint();
 	NValue retVal(VALUE_TYPE_DOUBLE);
 	retVal.getDouble() = point.getLatitude();
 	return retVal;
 }
 template<> NValue NValue::callUnary<FUNC_VOLT_POINT_LONGITUDE>() const {
-	const Point point = getPoint();
+    if (isNull()) {
+        return NValue::getNullValue(VALUE_TYPE_DOUBLE);
+    }
+    const Point point = getPoint();
 	NValue retVal(VALUE_TYPE_DOUBLE);
 	retVal.getDouble() = point.getLongitude();
 	return retVal;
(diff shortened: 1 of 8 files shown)


src/ee/expressions/geofunctions.cpp
```diff
@@ -246,4 +246,31 @@ template<> NValue NValue::call<FUNC_VOLT_CONTAINS>(const std::vector<NValue>& ar
     return ValueFactory::getBooleanValue(s2Poly->Contains(s2Point));
 }
 
+template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_NUM_INTERIOR_RINGS>() const {
+	const Geography polygon = getGeography();
+	NValue retVal(VALUE_TYPE_INTEGER);
+	retVal.getInteger() = polygon.numLoops();
```
wolffcm: You should subtract one from this answer, since there will always be one outer loop.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -100,6 +100,89 @@ public void testContains() throws Exception {
...
     }
 
+    public void testPolygonInteriorRings() throws Exception {
+        Client client = getClient();
+        populateTables(client);
+
+        // polygon with no holes
+        VoltTable vt = client.callProcedure("@AdHoc",
+                "select borders.name from borders "
+                + "where numInteriorRing(borders.region) = 1 "
```
wolffcm: This numInterior ring should return zero for Colorado and Wyoming here.

src/ee/expressions/geofunctions.cpp
hhakimi53: Oh ok, yes as they are interior rings. Will change it. thnxs

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: Yes, this also needs to be changed as it's taking into account exterior rings. will update the tests, thnxs.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -100,6 +100,89 @@ public void testContains() throws Exception {
...
+                + "order by borders.pk").getResults()[0];
+        assertContentOfTable(new Object[][]
+                {{"Colorado"},
+                 {"Wyoming"}
+                }, vt);
+
+        // polygon with holes
+        vt = client.callProcedure("@AdHoc",
+                "select borders.name from borders "
+                + "where numInteriorRing(borders.region) > 1 "
```
wolffcm: Please lock down the actual number that the function produces here with the equality operator.  This query will still appear to pass even if the function returned a number larger than two.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -100,6 +100,89 @@ public void testContains() throws Exception {
...
+                + "order by borders.pk").getResults()[0];
+        assertContentOfTable(new Object[][]
+                {{"Colorado with a hole around Denver"}
+                }, vt);
+
```
wolffcm: Also, you should add a query that verifies that passing a null polygon produces a null value.  It might be easier to do all of this with a single query, like `select numInteriorRings(region) from borders order by pk`.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -100,6 +100,89 @@ public void testContains() throws Exception {
...
+        populateTables(client);
+
+        // polygon with no holes have one ring only
+        // number of points will be as that on the only ring
+        VoltTable vt = client.callProcedure("@AdHoc",
+                "select borders.name, numPoints(borders.region) from borders "
+                + "where numInteriorRing(borders.region) = 1 "
+                + "order by borders.pk").getResults()[0];
+        assertContentOfTable(new Object[][]
+                {{"Colorado", 5},
```
wolffcm: The WKT for polygons require a closing vertex for each ring, which is why the answer is five here, even though there's just unique 4 vertices in the polygon.

I'm not sure which answer is right from the perspective of the OGC docs.  Can you have a look at the doc and verify that either 4 or 5 is the right answer?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -100,6 +100,89 @@ public void testContains() throws Exception {
...
+        // polygon with holes will have multiple interior rings
+        // number of points will be sum of points on all the rings
+        vt = client.callProcedure("@AdHoc",
+                "select borders.name, numPoints(borders.region) from borders "
+                + "where numInteriorRing(borders.region) > 1 "
+                + "order by borders.pk").getResults()[0];
+        assertContentOfTable(new Object[][]
+                {{"Colorado with a hole around Denver", 10}
+                }, vt);
+
```
wolffcm: It would be  good to also explicitly verify that null polygons produce a null result for this function.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -100,6 +100,89 @@ public void testContains() throws Exception {
...
+        VoltTable vt = client.callProcedure("@AdHoc",
+                "select places.name, LATITUDE(places.loc), LONGITUDE(places.loc) " +
+                "from places " +
+                "order by places.pk").getResults()[0];
+        assertContentOfTable(new Object[][]
+                {{"Denver",         39.70399856567383,  -104.95899963378906},
+                 {"Albuquerque",    35.112998962402344, -106.5989990234375},
+                 {"Cheyenne",       41.13399887084961,  -104.81300354003906},
+                 {"Fort Collins",   40.584999084472656, -105.0770034790039},
+                 {"Neverwhere",     360.0,              360.0},
```
wolffcm: The answers for Neverwhere shouldn't be 360.0, they should be null.

wolffcm:
That's all my comments this round.  Overall, it looks very nice.  I'll let you fix the stuff I mentioned and I'll have another look.

wolffcm:
One more thing.  It looks like the official name in the OGC docs is `numInteriorRing` (with no 's' at the end), but PostGIS spells the function *with* an s at the end.  It turns out this is due to the legacy of the standard:

http://osgeo-org.1560.x6.nabble.com/NumInteriorRing-s-td5119060.html

Can you find out what other vendors do?  If most of them just alias one to name to the other, we should probably do the same.

hhakimi53:
Most probably, we should alias one to another.

PostGIS supports both accessors function: http://postgis.net/docs/reference.html#Geometry_Accessors

MySql supports NumInteriorRings() (http://dev.mysql.com/doc/refman/5.7/en/gis-polygon-property-functions.html) and there is a ticket to add alias https://dev.mysql.com/worklog/task/?id=2388

SpatialLite also supports both NumInteriorRing() and NumInteriorRings() https://www.gaia-gis.it/spatialite-2.0/spatialite2_sql.html

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: There is not specific clarification in OGC SFA spec atleast, how should numPoints() should be treated for ploygon or even closed line strings. 

Post GIS includes the closing point (duplicate as start) when evaluating numPoints() on closed line string: http://postgis.net/docs/ST_NumPoints.html . So if consider that approach, than 5 would be correct answer considering each ring as closed line-string.

wolffcm:
I agree about the aliasing of numInteriorRings to numInteriorRing, given the links you found.

wolffcm:
This looks very nice.  Go ahead and merge when you're ready and Jenkins looks good.

hhakimi53:
Thanks Chris for reviewing it. Jenkins is having problems with some jobs and with failure message GCC too old. Will talk to Phil about it as it looks related to Jenkins and merge the changes when Jenkins is all set with the changes.

BillWhite:
This is a build problem.  Let me look at it.

Ignore the mac build failure (OSX10.6).  Those machines should be taken out
of the rotation.

The eecheck build failure is on U10.04, which has been working pretty
reliably up until now.  But I don't think it's your problem.

hhakimi53:
Thanks Bill.

Yes, it was originally build problem. I talked to Phil and it was getting pushed to wrong machine. Currently Phil is looking into it.

hhakimi53:
Jenkins is passing for eecheck (except OSX - known issue), community and pro-junit tests. Will merge in the changes to 8859 integration branch.

PR #2984, opened 2015-11-09
Refactor ee code xin

jxchinaren:


src/ee/storage/TupleBlock.h
```diff
@@ -35,6 +35,7 @@
 #include <deque>
 
 namespace voltdb {
+#define INVALID_NEW_BUCKET_INDEX -1
```
paulmartel: #defines are for C hackers. Use "static const int".

src/ee/storage/TupleBlock.cpp
```diff
@@ -32,7 +32,7 @@ TupleBlock::TupleBlock(Table *table, TBBucketPtr bucket) :
         m_activeTuples(0),
         m_nextFreeTuple(0),
         m_lastCompactionOffset(0),
-        m_tuplesPerBlockDivNumBuckets(m_tuplesPerBlock / static_cast<double>(TUPLE_BLOCK_NUM_BUCKETS)),
+        m_tuplesPerBucket(m_tuplesPerBlock / static_cast<double>(TUPLE_BLOCK_NUM_BUCKETS)),
```
paulmartel: This member is not actually needed at all -- neither is double math.

src/ee/storage/TupleBlock.h
```diff
@@ -114,37 +115,35 @@ class TupleBlock {
         return false;
     }
 
+    /**
+     * If the next free tuple is not able to be in the current bucket, return the new bucket index,
```
paulmartel: It's not the "next free tuple" in the buckets, it's the "tuple block with its current fullness".

src/ee/storage/TupleBlock.h
```diff
@@ -114,37 +115,35 @@ class TupleBlock {
...
-        }
-        else {
-            int index = static_cast<int>(::floor(m_activeTuples / m_tuplesPerBlockDivNumBuckets));
-            assert(index < TUPLE_BLOCK_NUM_BUCKETS);
-            assert(index >= 0);
-            return index;
+            return INVALID_NEW_BUCKET_INDEX;
         }
+
+        int index = static_cast<int>(::floor(m_activeTuples / m_tuplesPerBucket));
```
paulmartel: Better just to use integer math:
int index = TUPLE_BLOCK_NUM_BUCKETS * m_activeTuples / m_tuplesPerBlock;

src/ee/storage/TupleBlock.h
```diff
@@ -114,37 +115,35 @@ class TupleBlock {
...
+        assert(index < TUPLE_BLOCK_NUM_BUCKETS);
+        assert(index >= 0);
```
paulmartel: We don't need to assert that multiplying and dividing non-negative numbers doesn't give a negative number.

src/ee/storage/TupleBlock.h
```diff
@@ -153,6 +152,9 @@ class TupleBlock {
 
     std::pair<int, int> merge(Table *table, TBPtr source, TupleMovementListener *listener = NULL);
 
+    /**
+     * Find next free tuple storage address and its bucket index position.
```
paulmartel: not really "its bucket index position", more like "the tupleblock's bucket index".

src/ee/storage/TupleBlock.h
```diff
@@ -167,12 +169,14 @@ class TupleBlock {
         }
         m_activeTuples++;
         int newBucketIndex = calculateBucketIndex();
-        if (newBucketIndex != m_bucketIndex) {
-            m_bucketIndex = newBucketIndex;
-            return std::pair<char*, int>(retval, newBucketIndex);
-        } else {
-            return std::pair<char*, int>(retval, -1);
+        if (newBucketIndex == m_bucketIndex) {
+            // have space in current bucket
```
paulmartel: It's not about "having space in the bucket", it's about "tupleblock is not too full for its current bucket".

src/ee/storage/persistenttable.cpp
```diff
@@ -230,7 +230,7 @@ void PersistentTable::nextFreeTuple(TableTuple *tuple) {
     /**
      * Check to see if the block needs to move to a new bucket
      */
-    if (retval.second != -1) {
+    if (retval.second != INVALID_NEW_BUCKET_INDEX) {
```
paulmartel: You missed a bunch more magic "-1"s still in this module.

src/ee/storage/persistenttable.cpp
```diff
@@ -1422,15 +1423,15 @@ bool PersistentTable::doCompactionWithinSubset(TBBucketMap *bucketMap) {
...
         for (int ii = 0; ii < TUPLE_BLOCK_NUM_BUCKETS; ii++) {
-            lightestIterator = (*bucketMap)[ii]->begin();
-            if (lightestIterator != (*bucketMap)[ii]->end()) {
+            lightestIterator = (*bucketVector)[ii]->begin();
+            if (lightestIterator != (*bucketVector)[ii]->end()) {
                 lightest = lightestIterator.key();
                 if (lightest != fullest) {
                     foundLightest = true;
                     break;
                 } else {
```
paulmartel: Don't need an else after a break;

src/ee/storage/TupleBlock.h
```diff
@@ -167,12 +169,14 @@ class TupleBlock {
...
+            return std::pair<char*, int>(retval, INVALID_NEW_BUCKET_INDEX);
         }
+
+        // needs a new bucket and update bucket index
+        m_bucketIndex = newBucketIndex;
+        return std::pair<char*, int>(retval, newBucketIndex);
```
paulmartel: Even simpler?:

    if (newBucketIndex == m_bucketIndex) {
        newBucketIndex = INVALID_NEW_BUCKET_INDEX;
    }
    else {
         m_bucketIndex = newBucketIndex;
    }
    return std::pair<char*, int>(retval, newBucketIndex);

src/ee/storage/TupleBlock.h
paulmartel: "INVALID" may be a little misleading. A lot of times it just means that the bucket index has not changed, so there is not a new value.  maybe call this "NO_NEW_BUCKET_INDEX" instead?

paulmartel:
End of review round. The code IS much more readable, now. Thanks.

jxchinaren:
@paulmartel Thanks for reviewing that, I added another commit to resolve your comments.

PR #2985, opened 2015-11-10
ENG-9406 shrink StringRef overhead to 1/3, shrink string memory code …

paulmartel:
…path.


src/ee/common/NValue.hpp
```diff
@@ -716,6 +715,16 @@ class NValue {
         return copy;
     }
 
+    std::size_t getAllocationSizeForObject() const
```
wolffcm: I'm having trouble understanding how the code has changed here.  I see you've deleted two prototypes, and added one method with a definition.  There should be two method bodies that were deleted, but I don't see them---was this method completely unused before?

paulmartel: Yes, two declarations, no callers, no definition -- but the name was a keeper!

It's actually a new function, now public, that "perfects" NValue's usual isolation of StringRef
from TableTuple -- also it changes a needlessly hypothetical "what would have been allocated for this object's length?" function into a "what was actually allocated for this object".

wolffcm:
This looks good.  I guess I don't have many comments this time.

I verified that we're pooling 1/3rd of what we were before, by running Ben's benchmark code with instrumented pools.  This is a very nice improvement.

wolffcm:
We should keep any eye on the performance graphs with this change, it might show an improvement?

PR #2986, opened 2015-11-10
Eng 9380 row guards

paulmartel:


paulmartel:
The root of the problem is an assumption in the EE that the columns in a row used in a row comparison with a subquery correspond one-to-one with column references (TVEs) in the parent query.

wolffcm:
I'm looking at this now.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -295,6 +295,7 @@ protected AbstractExpression parseConditionTree(VoltXMLElement root) {
         // The assumption is that FALSE boolean results can replace NULL
```
wolffcm: I've a comment unrelated to your changes.  In an AND expression NULL is equivalent to FALSE, but NULL OR TRUE should return true, which seems to violate the assumption that FALSE can replace NULL.

Is there a latent bug here for cases like `WHERE f = 3 OR g IN (select ...)`?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -664,6 +665,58 @@ private AbstractExpression parseOperationExpression(VoltXMLElement exprNode) {
...
+            rejectDisallowedRowOpExpressions(expr.getLeft());
+        }
+
+        // The problem cases are all comparison ops.
+        if ( ! (expr instanceof ComparisonExpression)) {
+            return;
+        }
+
+        // The problem cases all have row expressions as left hand operands.
+        AbstractExpression rowExpression = expr.getLeft();
```
wolffcm: Since `expr` must be a comparison expression here, and both `=` and `<>` are symmetric, is it possible that there are problem cases where a row expression is on the right and a select statement is on the left?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -664,6 +665,58 @@ private AbstractExpression parseOperationExpression(VoltXMLElement exprNode) {
...
+        }
+
+        // Verify that the ROW OP SUBQUERY expression's ROW operand only
+        // contains column arguments that reference exactly one column.
+        for (AbstractExpression arg : rowExpression.getArgs()) {
+            List<AbstractExpression> tves = arg.findBaseTVEs();
+            if (tves.size() != 1) {
+                if (tves.isEmpty()) {
+                    throw new PlanningErrorException(
+                            "Unsupported use of a constant value in a row column expression.");
```
wolffcm: Seems like the absence of TVEs doesn't guarantee that constant values are in the row expression.  What about `current_timestamp` or other nullary functions (if there are any)?  

Is a column argument of `col * 2` okay here?

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -1333,6 +1357,74 @@ public void testRowNotOrIsNullInOrOpAnyWithInnerNull() throws Exception
...
+            }
+            // A subquery with a limit does not rewrite as EXISTS,
+            // so the multi-column-based row column expression
+            // is rejected.
+            sql =   "select ID from R1 " +
+                    "where (WAGE, ID+DEPT-ID) IN " +
+                    "      (select WAGE, DEPT from R2 " +
+                    "       where ID < 104 limit 1) " +
+                    "order by ID;";
+            try {
```
wolffcm: Maybe add a test case where the row expression is `(WAGE * 2, ID * 10)`, just to show that it works, if indeed it does?

wolffcm:
End of review round.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
paulmartel: So, should I say "Unsupported use of a column-independent value" instead of "column value"?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
paulmartel: @wolffcm Symmetry is not even required. I should just err on the side of guarding against any operand of type RowSubqueryExpression and not care about the opposite operand.  Row **Subquery** Expression seems to imply a subquery context. I don't know whether or not some other operand type is used for edge case syntax like "row(a, b, c) > row(1, 2, 3)" and whether that currently works, and I'm not sure I care.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
paulmartel: @wolffcm Seemed to work in informal testing -- will add a test case.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
wolffcm: Sure, that works.  Maybe it's a little confusing though, so I'd be okay with you leaving the message as-is too.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
paulmartel: @wolffcm yes, NULL OR TRUE should return true just like FALSE OR TRUE returns true.
I think that there may be a latent edge case bug here but it's not quite that simple.
I think it's more like `not nullable_f IN ( select non_null_g from t )` --> `not exists ( select 1 from t where non_null_g = nullable_f )`. When nullable_f is NULL, I think the IN operation returns NULL and NOT NULL returns NULL interpreted as FALSE. The exists form calculates an empty result so the exists predicate is false and not exists is TRUE.
(and yet I think I have tests for this?).

paulmartel: @wolffcm - yes. apparently we avoid the NOT IN --> NOT EXISTS transformation for this very reason.

wolffcm:
@paulmartel Post-review response, these changes look good.

paulmartel:
separately squash merged

PR #2992, opened 2015-11-12
synch memory sizing worksheet to new ENG-9406 string allocation code

paulmartel:


src/frontend/org/voltdb/utils/CatalogSizing.java
```diff
@@ -222,7 +222,10 @@ private static int getVariableColumnSize(int capacity, int dataSize, boolean for
             }
             bufferSize += increment;
         }
-        return bufferSize + 8 + 24;
+        // The rounded buffer size has an additional 4-byte allocation size and
+        // 8-byte back pointer overhead. There is also has an 8-byte pointer
```
wolffcm: The 8-byte back pointer and buffer size should be included in the content that sizes the buffer, right?  Or maybe I'm confused about how we size buffers.

wolffcm:
I guess there's no unit tests for this. It would be good to have some.  (I am not suggesting that you do this now, but feel free to file a ticket.)

A general comment, not related to your change: it seems silly to duplicate the buffer sizing logic from the EE here, but I guess the alternative is a JNI call.  I'm repeatedly reminded of design challenges of implementing a DB in two languages.

That's all my comments.

src/frontend/org/voltdb/utils/CatalogSizing.java
paulmartel: @wolffcm This calculation accurately reflects the reality of the sizing logic. We pad then round up then pad some more. I have to admit that I never did achieve the effect I was going for here -- to minimize the cross-exposure among the different allocator layers WRT how much overhead they add to the entry.  What I settled for was a slight minimization in the size of the code change and the number of places I had to reference the overhead from different layers -- admittedly not a major consideration. Maybe I should have maintained the status quo where the power-of-two contained "the whole allocation" now with the addition of the embedded 4-byte allocation length. In retrospect, that does seem like a simpler goal. Maybe I should change the half dozen lines of code it would take to fix this in ThreadLocalPool?

paulmartel: @wolffcm a unit test would do what? call through special test-only JNI hooks to validate the calculation?

wolffcm: A unit test could at least lock down a hard-coded value and then we would see a test failure if we ever introduced a bug into the code.  That's better than no test.

My tendency would be to avoid making any more product changes at this point so close to shipping 5.8.  It might make sense to make follow-up tickets for the refactoring you still want to do.

PR #2995, opened 2015-11-12
Eng 8972 geo small crypto here

BillWhite:
This is to embed openssl into our source tree, and use the libcrypto which
results from this build.  The only interesting files in the really big list are
buildtools.py, build.py and the Makefile.org file from openssl.  The rest are
just imported from openssl.

There are three problems not addressed by this commit, two minor and one major.
1. There are some files here which are really auto configured, so they should not be in git at all.
2. There are some directories which should be listed in the .gitignore file, so that they will not show up uselessly in git status.
3. The big problem is that this build does not make all the S2 unit tests pass.  The problem really is with this build, as I have built using the system build and everything passes just fine.


wolffcm:
2450 is a lot of files.  Or existing source directory only has 1952 files.  Our third_party source directory has 4628 files though.  Is there any way to whittle down the number of files we need?

For the short term, this gets us a bluer topic branch and lets us feel confident about merging our changes, so I'm okay with merging this if it looks good in Jenkins.

BillWhite:
Probably we can get rid of a lot of this.  But I reckoned that this is
impeding geo progress, and it was better pushing everything and optimizing
later.

PR #3000, opened 2015-11-13
cpptest changes to pass stricter valgrind checks

paulmartel:


BillWhite:
All makes sense to me.

PR #3001, opened 2015-11-13
ENG-8998: Remove DR restart/rejoin events

afrogers:


nshi:
Squash merged externally.

PR #3006, opened 2015-11-14
Fix processing of null geography and point values for JSON 

hhakimi53:


wolffcm:
Oops, I think I introduced this one.  This looks good.  Thanks a lot for jumping on this.

PR #3014, opened 2015-11-17
ENG-9431: Update DR binary log unit tests.

nshi:
Separate the execution engines, one for master, one for replica, so they
can have two cluster IDs and such. Check for the DR timestamp after
updates.

Change-Id: I45b57d1742ae021092528c64ae980398997e6d3b

### DIFF
--- a/tests/ee/storage/DRBinaryLog_test.cpp
+++ b/tests/ee/storage/DRBinaryLog_test.cpp
@@ -51,7 +51,8 @@ using namespace voltdb;
 
 const int COLUMN_COUNT = 7;
 const int HIDDEN_COLUMN_COUNT = 1;
-const int CLUSTER_ID = 0;
+const int CLUSTER_ID = 1;
+const int CLUSTER_ID_REPLICA = 2;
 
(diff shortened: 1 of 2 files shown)


tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -709,22 +745,22 @@ TEST_F(DRBinaryLogTest, PartitionedTableNoRollbacks) {
...
     // delete the second row inserted in the last write
-    beginTxn(112, 102, 101, 73);
+    beginTxn(m_engine, 112, 102, 101, 73);
     deleteTuple(m_table, second_tuple);
     // Tick before the delete
     ASSERT_TRUE(flush(101));
-    endTxn(true);
-    // Apply the binary log after endTxn() to get a valid undoToken.
+    endTxn(m_engine, true);
+    // Apply the binary log after endTxn(m_engine, ) to get a valid undoToken.
```
afrogers: Kind of an odd find-replace into the comment?

PR #3017, opened 2015-11-17
Add gcc 5+ version to build script

hhakimi53:
Updated build script to build code using gcc 5.* in same way as it does today for gcc versions 4.3 or higher.

### DIFF
--- a/src/ee/common/types.cpp
+++ b/src/ee/common/types.cpp
@@ -112,7 +112,7 @@ NValue getRandomValue(ValueType type, uint32_t maxLength) {
             int length = (rand() % maxLength);
             unsigned char bytes[maxLength];
             for (int ii = 0; ii < length; ii++) {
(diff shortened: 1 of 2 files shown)


src/ee/common/types.cpp
```diff
@@ -112,7 +112,7 @@ NValue getRandomValue(ValueType type, uint32_t maxLength) {
             int length = (rand() % maxLength);
             unsigned char bytes[maxLength];
             for (int ii = 0; ii < length; ii++) {
-                bytes[ii] = (unsigned char)rand() % 256; //printable characters
+                bytes[ii] = static_cast<char> (rand() % 256); //printable characters
```
wolffcm: It would be better to cast to "unsigned char" since that is the type of the thing actually being assigned to.

Looks good otherwise.

hhakimi53: Sorry, that was an oversight, have fixed it. thnxs

PR #3019, opened 2015-11-18
ENG-8559 planner NPE on partitioned matviews and subqueries

paulmartel:
renamed getStmtTableScanByAlias to resolveStmtTableScanByAlias, so I could use the former name as a simpler accessor, allowing m_tableAliasMap to be hidden better.

The wide visibility of m_tableAliasMap had me chasing ghosts looking for root causes of the NPE.

Moved some AbstractPlanNode dependencies out of StmtSubqueryScan where they don't belong and into PlanAssembler where they do belong and a little into StatementPartitioning (where maybe it really doesn't, but the alternative was not obvious).

### DIFF
--- a/src/frontend/org/voltdb/planner/PlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/PlanAssembler.java
@@ -37,6 +37,7 @@ import org.voltdb.catalog.Database;
 import org.voltdb.catalog.Index;
 import org.voltdb.catalog.Table;
 import org.voltdb.expressions.AbstractExpression;
+import org.voltdb.expressions.AbstractSubqueryExpression;
 import org.voltdb.expressions.AggregateExpression;
 import org.voltdb.expressions.ConstantValueExpression;
 import org.voltdb.expressions.ExpressionUtil;
@@ -61,6 +62,7 @@ import org.voltdb.plannodes.IndexScanPlanNode;
 import org.voltdb.plannodes.InsertPlanNode;
 import org.voltdb.plannodes.LimitPlanNode;
 import org.voltdb.plannodes.MaterializePlanNode;
+import org.voltdb.plannodes.MergeReceivePlanNode;
 import org.voltdb.plannodes.NestLoopPlanNode;
 import org.voltdb.plannodes.NodeSchema;
 import org.voltdb.plannodes.OrderByPlanNode;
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -806,30 +809,85 @@ private CompiledPlan getNextUnionPlan() {
...
+            AbstractPlanNode parent = current.getParent(0);
+            parent.unlinkChild(current);
+            parent.addAndLinkChild(child);
+            return root;
+        }
+        if (current.getChildCount() == 1) {
+            // This is still a coordinator node
+            return removeCoordinatorSendReceivePairRecursive(root, current.getChild(0));
+        }
+        // We have hit a nestloop join.
```
wolffcm: Anything with more than one child is a nestloop join?  Couldn't you get here for union too?

wolffcm:
I've started reviewing this, and will finish up later tonight.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -860,6 +918,43 @@ private AbstractPlanNode connectChildrenBestPlans(AbstractPlanNode parentPlan) {
...
+                // For now, this is just a special case violation of the limitation on
+                // use of expression subqueries in MP queries on partitioned data.
+                // That special case was going undetected when we didn't flag it here.
+                m_recentErrorMsg = IN_EXISTS_SCALAR_ERROR_MESSAGE;
+                return null;
+            }
+            // // Something more along these lines would have to be enabled
+            // // to allow expression subqueries to be used in multi-partition
+            // // matview queries.
+            // if (!getBestCostPlanForExpressionSubQueries(subqueryExprs)) {
```
wolffcm: Is uncommenting this code sufficient for queries of this type to work, or would there be additional changes needed?

wolffcm:
This looks good to me.  And the refactoring is nice.

src/frontend/org/voltdb/planner/PlanAssembler.java
paulmartel: This logic seems like it would do the trick -- that is, it seems to be not too different from the code that handles the normal code path, and it at least passed some light planner testing before I realized that I needed to enforce the guard for consistency, but I would not say for sure without some serious testing. In general, I think that there is a mismatch between the ambition of the original scalar subquery implementation and the guards that were placed on it pre-release. The tricky cases probably got SOME amount of exercise before the guards took the pressure off, but they may have had undetected faults or they may be suffering from bit rot as the live code evolved around them.

src/frontend/org/voltdb/planner/PlanAssembler.java
paulmartel: Yeah -- I forgot that union is also a possibility -- I'll change the comment.

PR #3020, opened 2015-11-18
ENG-9446 add string allocation CatalogSizing junit test case

paulmartel:


BillWhite:
This seems right to me.  I like extracting the allocation size calculation.

PR #3022, opened 2015-11-19
Use S2 polygon encoding format for serialization

wolffcm:


BillWhite:
I'll look into this right now.

src/ee/common/Geography.hpp
```diff
@@ -18,251 +18,220 @@
...
-        bool operator==(const iterator& that) const {
-            return m_pos == that.m_pos;
-        }
+/**
+ * A class for representing instances of geo-spatial geographies.
+ * (Currently only polygons can be represented here.)
+ *
+ * Note that variable length data in the EE is typically prefixed with a length.
+ * The pointer here should be to the start of the data just after the length.
+ */
```
BillWhite: Is this a length in bytes?  I suspect it's not in loops.

If it's in bytes, what's the m_length value for?

src/ee/common/Geography.hpp
```diff
@@ -272,153 +241,378 @@ class Geography {
      *   3. Finally, if all loops have the same number of vertices, sort on the
      *        points themselves (which will involve doing floating-point comparison)
      */
```
BillWhite: How do we know what order the loops will be in?  Are the always maintained in the WKT input order?

src/ee/common/Geography.hpp
```diff
@@ -272,153 +241,378 @@ class Geography {
      *   3. Finally, if all loops have the same number of vertices, sort on the
```
BillWhite: I think (2) is not exactly right.  If the loops are L={L1, ..., Lk} and R={R1, ..., Rk}, then the first m such that
size(Lm) != size(Rm) determines the order.  If size(Lm) < size(Rm) and size(L(m+1)) > size(R(m+1)), then
L < R.  I don't know how to say that in English very clearly, though.

As it is, it seems like you are comparing the sum of the vertices of all the loops, which is not what you are
doing, and would be nonsense if you did.

src/ee/common/serializeio.h
```diff
@@ -68,6 +68,8 @@ namespace voltdb {
 
 #ifdef __DARWIN_OSSwapInt64 // for darwin/macosx
 
+#undef htonll
```
BillWhite: Do you need to surround these with "#ifdef htonll"?  If htonll is not defined, and you undef it, does this fail to compile?

Or do you know it's always defined?

Byte swapping is evil.

src/ee/expressions/geofunctions.cpp
```diff
@@ -137,22 +123,23 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POINTFROMTEXT>() const
...
         Point::Coord lat = stringToCoord(POLY, wkt, *it);
-        loop->push_back(lat);
         ++it;
 
         Point::Coord lng = stringToCoord(POLY, wkt, *it);
-        loop->push_back(lng);
         ++it;
 
+        points.push_back(S2LatLng::FromDegrees(lat, lng).ToPoint());
+
```
BillWhite: It's a shame we can't use emplacement here.

src/ee/expressions/geofunctions.cpp
```diff
@@ -171,6 +158,24 @@ static void readLoop(const std::string &wkt,
...
+    if (first != last) {
+        throwInvalidWktPoly("A polygon ring's first vertex must be equal to its last vertex");
+    }
+
+    // S2 format considers the closing vertex in a loop to be
+    // implicit, while in WKT it is explicit.  Remove the closing
+    // vertex here to reflect this.
+    points.pop_back();
+
+    loop->Init(points);
```
BillWhite: This is exactly where you want move semantics.  Of course, we may use move semantics when we compile on C++11, but we will never do this to release.

BillWhite:
That's pretty much all I've got.

As far as representing points with three dimensions, I wonder if we could make S2point abstract, represent longitude and latitude, and calculate the x, y and z values.  There is a relatively simple formula to calculate these, but it involves some trigonometric functions.  I wonder how slow that would
be and whether the space savings is worth the computation time.  I'm not even sure how to determine what "worth it" would mean.

src/ee/common/Geography.hpp
wolffcm: Yes, it's in bytes.  The m_length value is mostly for assertions to make sure that we don't read past the end of the byte array that makes up the polygon, and that we also read everything we're supposed to.

src/ee/common/Geography.hpp
wolffcm: S2Polygon wants to reorder the loops in depth-first order, so the outer loop ends up last.  So the WKT order cannot be maintained.

src/ee/common/Geography.hpp
wolffcm: You're right.  The comment is unclear.  I have tried to make it clearer.

src/ee/common/serializeio.h
wolffcm: Hmm... There are similarly unguarded `#undef`s just below, and this seems to be building on OS X and in Jenkins.  The only reason I added this is because there are S2 headers that redefine those symbols too.

src/ee/expressions/geofunctions.cpp
wolffcm: Would it do any good in this case, since S2Point wouldn't seem to have any special move semantics?  It's just static array of three doubles.

src/ee/expressions/geofunctions.cpp
wolffcm: I'm not sure I follow you.  I've been using unique_ptr with success, and I think that needs move semantics.  Is there any reason I shouldn't be using it?  It seems to be available in the compilers we're using.

For S2Loop, it's not using a vector under the hood, and it doesn't have a convenient way to just give it a point array without dong a copy.  I've been a little averse to making too many changes to S2.

There's definitely an extra copy here.  Do you think it's a problem?

wolffcm:
I've pushed a commit that tries to make comments a little clearer.

As far as time/space tradeoff in lat/lng vs xyz points, I'm not sure we can understand the penalty of converting to xyz on the fly until we have a demo with a sizable workload.  I think that it's probably worth the experiment.

src/ee/common/Geography.hpp
BillWhite: I was just thinking it should be spelled out explicitly.  If I wondered
about it, someone else might as well.

src/ee/expressions/geofunctions.cpp
BillWhite: Isn't there an extra copy?  S2LatLng::ToPoint() creates an object, then the vector object is constructed, then the object is copied in to the vector.  Emplacement would allow you to construct the temporary object directly.  But maybe, since this is a little bit of a POD object it doesn't matter.

I didn't think the C++ on C6 implemented move semantics.  I could be misremembering, though.

BillWhite:
I'm fine with merging this as it is.  I wasn't objecting to anything, more than just kibitzing.

PR #3026, opened 2015-11-19
ENG-9432: Allow generation of a zero-table native snapshot

afrogers:


luyangco:
Looks good to me, but two test cases fail on Jenkins which I guess needs to be modified accordingly.

PR #3027, opened 2015-11-19
ENG-9432: Base DR conflict timestamps off Unix epoch; move VOLT_EPOCH…

afrogers:
… definition to executorcontext

### DIFF
--- a/src/ee/common/executorcontext.hpp
+++ b/src/ee/common/executorcontext.hpp
@@ -29,12 +29,12 @@
 
 namespace voltdb {
 
+extern const int64_t VOLT_EPOCH;
 const int64_t HIDDEN_VALUE_TIMESTAMP_MASK = (1LL << 49) - 1LL;
 
 class AbstractExecutor;
 class DRTupleStream;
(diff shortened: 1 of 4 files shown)


src/ee/common/executorcontext.hpp
```diff
@@ -88,7 +82,7 @@ class ExecutorContext {
         m_spHandle = spHandle;
         m_txnId = txnId;
         m_lastCommittedSpHandle = lastCommittedSpHandle;
-        m_currentTxnTimestamp = (m_uniqueId >> 23) + m_epoch;
+        m_currentTxnTimestamp = (m_uniqueId >> 23) + VOLT_EPOCH;
```
luyangco: I'm not sure if I asked the same question before, is there a specific reason to use the m_uniqueId before assigning the new value to it? Seems weird to me.

nshi: That's a very good point. Seems like this is always wrong?

afrogers:
Merged externally

PR #3028, opened 2015-11-19
 ENG-8980: an override to let VoltDB start even if THP is enabled

JadenW:
manually test:
 on thp enabled machine, voltdb  create --skip-check=thp will start voltdb with warning

voltdb check won't be affected

### DIFF
--- a/lib/python/voltcli/verbs.py
+++ b/lib/python/voltcli/verbs.py
@@ -425,9 +425,9 @@ class ServerBundle(JavaBundle):
     def initialize(self, verb):
         JavaBundle.initialize(self, verb)
 	verb.add_options(
-            cli.StringListOption(None, '--skip-check', 'skip_requirements',
+            cli.StringListOption(None, '--ignore', 'skip_requirements',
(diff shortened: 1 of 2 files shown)


lib/python/voltcli/verbs.py
```diff
@@ -424,6 +424,11 @@ def __init__(self, subcommand,
 
     def initialize(self, verb):
         JavaBundle.initialize(self, verb)
+	verb.add_options(
+            cli.StringListOption(None, '--skip-check', 'skip_requirements',
```
akhanzode: I thought we wanted --ignore=thp

lib/python/voltcli/verbs.py
```diff
@@ -424,6 +424,11 @@ def __init__(self, subcommand,
...
+                             '''requirements to skip when start voltdb:
+			     thp - ignore transparent huge page check at your peril, you could run out of memory''',
```
akhanzode: Check this message with Andrew more customer friendly but stern.

PR #3029, opened 2015-11-19
ENG-9461, as spUniqueId or mpUniqueId of the invocation buffer may be…

luyangco:
… 0 if the buffer contains only sp or mp txn,

fix the incorrect statistics of lastQueuedSegmentId and lastAckedSegmentId.

Change-Id: I91207c20dc72339bdbe2d6aaf78650dcfaffc65b

### DIFF
--- a/src/frontend/org/voltdb/DRLogSegmentId.java
+++ b/src/frontend/org/voltdb/DRLogSegmentId.java
@@ -92,14 +92,14 @@ public class DRLogSegmentId implements Serializable {
 
         public MutableBinaryLogInfo() {
             drId = Long.MIN_VALUE;
-            spUniqueId = Long.MIN_VALUE;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -96,6 +96,12 @@ public MutableBinaryLogInfo() {
             mpUniqueId = Long.MIN_VALUE;
         }
 
+        public void reset() {
+            drId = Long.MIN_VALUE;
+            spUniqueId = Long.MIN_VALUE;
```
afrogers: As a matter of default, it's a little odd that these are using `Long.MIN_VALUE` while the rest of DR uses 0, a.k.a. `InvocationBuffer.UNSET_UNIQUE_ID`

nshi:
Squash merged.

PR #3031, opened 2015-11-20
Test for precision loss converting from latitude/longitude to S2Point.

BillWhite:
We store points as S2Point values, which are three dimensional values.
When converting from latitude to longitude to S2Point there is some
arithmetic and trigonometry, and this can cause precision loss.  This
test checks that this precision loss is bounded.

https://issues.voltdb.com/browse/ENG-9388


src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -249,7 +249,7 @@ public static GeographyValue unflattenFromBuffer(ByteBuffer inBuffer) {
      * But the interface we expose to users is (lat, lng).  This class is the
      * internal representation for vertices.
      */
-    private static class XYZPoint {
+    public static class XYZPoint {
         private final double m_x;
```
BillWhite: I made this public so that I could test it.  Perhaps package would have been possible.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -37,6 +36,66 @@ public TestGeospatialFunctions(String name) {
         super(name);
     }
 
+    /*
+     * We want to store the borders table once and for all, and insert and test
+     * without repeating ourselves. This class holds geometry values for us for
+     * inserting and for testing later on.
+     */
+    private static class Borders {
```
BillWhite: This  class is entirely so that we don't have to repeat the polygon loops.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -178,10 +213,87 @@ public void testLongitudeLatitude() throws Exception {
...
 
+    public void testPolygonFloatingPrecision() throws Exception {
+        final double EPSILON = -1.0;
+        Client client = getClient();
+        populateTables(client);
+
+        String sql = "select name, region "
+                        + "from borders order by borders.pk";
+        VoltTable vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertApproximateContentOfTable(new Object[][]
```
wolffcm: If I'm understanding the code correctly, this EPSILON parameter will only get used for comparisons of floating point column values, but not for polygons or points.  I think that's why the -1.0 here (to specify exact comparison) is working---the polygon vertices themselves will be compared with `RegressionSuite.assertEquals(PointType ...)`, which has a hard-coded epsilon of 0.001.

I think it would be good to have an assertApproximateEquals method that accepts an epsilon for points.

wolffcm: I.e., my initial testing for point equality was just lazy with respect to epsilons.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -178,10 +213,87 @@ public void testLongitudeLatitude() throws Exception {
...
+     *
+     * The error bound is 1.0e-13, which is the value of EPSILON below.  We tested 1.0e-14,
+     * but that fails.
+     *
+     * Note that no conversions from text to floating point happen anywhere here, so that
+     * is not a source of precision loss.  Only calculation cause these precision losses.
+     *
+     * @throws Exception
+     */
+    public void testXYZPoint() throws Exception {
```
wolffcm: I don't think that this particular test needs anything from RegressionSuite.  If that's the case, it would be good to move it to `org/voltdb/types/TestGeographyValue`, so XYZPoint can have package-level (and not public) visibility.  GeographyValue is a class that users will be working with, so its public interface will need to be well-documented.  I don't want to have to expose the details of XYZPoint to users.

  (There are two identically-named classes in different packages.  Perhaps that wasn't such a good idea.)

wolffcm:
For the most part, this looks good.  I think you should make XYZPoint non-public before merging though.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
BillWhite: Yes.  This test is not actually that useful.  But I think it would be good to move the epsilon into the assert as well, so that's what I'll do.

BillWhite:
This was squashed elsewhere.

PR #3034, opened 2015-11-23
ENG-9489 sql functions. DO NOT MERGE! FOR REVIEW ONLY!

paulmartel:
Apply contributed patch to 1a0c54a master.

### DIFF
--- a/build.py
+++ b/build.py
@@ -74,7 +74,7 @@ if CTX.PROFILE:
     CTX.CPPFLAGS += " -fvisibility=default -DPROFILE_ENABLED"
 
 # linker flags
-CTX.LDFLAGS += """ -g3"""
+CTX.LDFLAGS += """ -g3 -lboost_regex"""
 CTX.LASTLDFLAGS = """ """
 CTX.LASTIPCLDFLAGS = """ -ldl """
 
--- a/src/ee/expressions/datefunctions.h
+++ b/src/ee/expressions/datefunctions.h
@@ -20,6 +20,8 @@
 #include "boost/date_time/posix_time/posix_time_duration.hpp"
 #include "boost/date_time/posix_time/ptime.hpp"
 #include "boost/date_time/posix_time/conversion.hpp"
+#include "boost/date_time/posix_time/posix_time.hpp"
+#include "boost/date_time/gregorian/gregorian.hpp"
 #include <ctime>
 #include "common/SQLException.h"
 #include "common/executorcontext.hpp"
@@ -29,6 +31,13 @@ static const boost::posix_time::ptime EPOCH(boost::gregorian::date(1970,1,1));
 static const int64_t GREGORIAN_EPOCH = -12212553600000000;  // 1583-01-01 00:00:00
 static const int8_t QUARTER_START_MONTH_BY_MONTH[] = {
         /*[0] not used*/-1,  1, 1, 1,  4, 4, 4,  7, 7, 7,  10, 10, 10 };
+
+static const int64_t PTIME_MAX_YEARS = std::numeric_limits<int>::max();
+static const int64_t PTIME_MIN_YEARS = std::numeric_limits<int>::min();
+static const int64_t PTIME_MAX_MONTHS = std::numeric_limits<int>::max();
+static const int64_t PTIME_MIN_MONTHS = std::numeric_limits<int>::min();
+static const int64_t PTIME_MAX_QUARTERS = PTIME_MAX_MONTHS / 3;
+static const int64_t PTIME_MIN_QUARTERS = PTIME_MIN_MONTHS / 3;
 
 /** Convert from epoch_micros to date **/
 static inline void micros_to_date(int64_t epoch_micros_in, boost::gregorian::date& date_out) {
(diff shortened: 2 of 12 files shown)


build.py
```diff
@@ -74,7 +74,7 @@
     CTX.CPPFLAGS += " -fvisibility=default -DPROFILE_ENABLED"
 
 # linker flags
-CTX.LDFLAGS += """ -g3"""
```
paulmartel: This linker option and the boost/regex.hpp include in stringfunctions.h need to be supported in a general way that works for both VoltDB's cross-platform build environment AND for VoltDB's community and commercial deployments.

The VoltDB code base to date has avoided dependency on boost library code by only using boost "header-only" features.

Any boost version dependencies need to be identified for the purpose of our internal cross-platform build and test procedures. Our build and test system tend to work in two modes:
  1. build and test on each supported platform. (including Centos and RedHat 5 and later and Ubuntu 12.04 and later, and OSX 10.6 and later)
  2. build on Centos 5 and test on compatible linux platforms (including Centos and RedHat 5 and later, and Ubuntu 10.04 and later)

Centos/RedHat 6 and Ubuntu 12.04 may become prerequisites for building or deploying VoltDB release 6 or greater.

Normally, for safety in product deployment, external open source library dependencies are resolved by version-controlled components included in the VoltDB source and binary packaging.

src/ee/expressions/datefunctions.h
```diff
@@ -393,4 +402,280 @@ template<> inline NValue NValue::callConstant<FUNC_CURRENT_TIMESTAMP>() {
...
+    if (number.isNull()) {
+        return getNullValue();
+    }
+
+    const NValue& date = arguments[1];
+    if (date.isNull()) {
+        return getNullValue();
+    }
+
+    int64_t interval = number.castAsBigIntAndGetValue();
```
paulmartel: The remainder of this function -- except for multiplying interval by 3 -- should be factored into a static function that is also called by <FUNC_VOLT_DATEADD_MONTH> for maximized code reuse.

src/ee/expressions/datefunctions.h
```diff
@@ -393,4 +402,280 @@ template<> inline NValue NValue::callConstant<FUNC_CURRENT_TIMESTAMP>() {
...
+    if (interval > PTIME_MAX_YEARS || interval < PTIME_MIN_YEARS) {
+        throw SQLException(SQLException::data_exception_numeric_value_out_of_range, "interval is too large for DATEADD function");
+    }
+
+    int64_t epoch_micros = date.getTimestamp();
+    boost::gregorian::date as_date;
+    boost::posix_time::time_duration as_time;
+    micros_to_date_and_time(epoch_micros, as_date, as_time);
+    boost::posix_time::ptime ts(as_date, as_time);
```
paulmartel: These five lines are often repeated -- they could be factored out into a module-static inline function.

src/ee/expressions/functionexpression.cpp
```diff
@@ -202,6 +202,9 @@ ExpressionUtil::functionFactory(int functionId, const std::vector<AbstractExpres
         case FUNC_CURRENT_TIMESTAMP:
             ret = new ConstantFunctionExpression<FUNC_CURRENT_TIMESTAMP>();
             break;
+        case FUNC_PI:
+            ret = new ConstantFunctionExpression<FUNC_PI>();
```
paulmartel: call<FUNC_PI> could be avoided entirely by substituting a constant value for pi in the expression tree.

    ret = new ConstantValueExpression(getDoubleValue(boost::math::constants::pi<double>()));

src/ee/expressions/functionexpression.cpp
```diff
@@ -424,6 +427,36 @@ ExpressionUtil::functionFactory(int functionId, const std::vector<AbstractExpres
         case FUNC_VOLT_SUBSTRING_CHAR_FROM:
             ret = new GeneralFunctionExpression<FUNC_VOLT_SUBSTRING_CHAR_FROM>(*arguments);
             break;
+        case FUNC_VOLT_DATEADD_YEAR:
```
paulmartel: This list of cases is intended to be organized alphabetically to avoid code merge conflicts.

src/ee/expressions/stringfunctions.h
```diff
@@ -636,6 +637,81 @@ template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_CURRENCY>(const std::vect
...
+    if (pat.isNull()) {
+        return getNullValue();
+    }
+    if (pat.getValueType() != VALUE_TYPE_VARCHAR) {
+        throwCastSQLException(pat.getValueType(), VALUE_TYPE_VARCHAR);
+    }
+
+    int32_t lenSource = source.getObjectLength_withoutNull();
+    char* sourceChars = reinterpret_cast<char*>(source.getObjectValue_withoutNull());
+    std::string sourceStr(sourceChars, lenSource);
```
paulmartel: The std::string sourceStr requires a buffer copy and is not required to get the equivalent of sourceStr.begin(), sourceStr.end(), and sourceStr.c_str(). Observe how call<FUNC_VOLT_SUBSTRING_CHAR_FROM> uses valueChars and valueEnd as an alternative to constructing a std::string when the underlying API does not require one.

src/ee/expressions/stringfunctions.h
```diff
@@ -636,6 +637,81 @@ template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_CURRENCY>(const std::vect
...
+            position = NValue::getCharLength(sourceStr.c_str(), rit->position()) + 1;
+        } else {
+            position = 0;
+        }
+    } catch (boost::regex_error& e) {
+        // TODO Return more specific error messages?
+        throw SQLException(SQLException::data_exception_invalid_parameter, "illegal pattern string");
+    }
+
+    return getIntegerValue(static_cast<int32_t>(position));
```
paulmartel: A sql function that is configured with a BIGINT result value should use getBigIntValue rather than getIntegerValue for consistency. I'd need to consider the minor trade-offs of whether to standardize on a BIGINT or INTEGER return type. As long as this sql function is made self-consistent, I'd expect the trade-offs between BIGINT and INTEGER to be minor.

src/frontend/org/voltdb/jdbc/JDBC4DatabaseMetaData.java
```diff
@@ -614,7 +614,7 @@ public int getMaxUserNameLength() throws SQLException
     public String getNumericFunctions() throws SQLException
     {
         checkClosed();
-        return "ABS,BITAND,BITNOT,BITOR,BIT_SHIFT_LEFT,BIT_SHIFT_RIGHT,BITXOR,CEILING,EXP,FLOOR,LN,LOG,POWER,SQRT";
+        return "ABS,BITAND,BITNOT,BITOR,BIT_SHIFT_LEFT,BIT_SHIFT_RIGHT,BITXOR,CEILING,EXP,FLOOR,LN,LOG,PI,POWER,SQRT";
```
paulmartel: The other two new function names should be added to other function lists in this module.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -224,6 +236,17 @@ private FunctionId(String name, Type type, int id, int typeParameter, Type[] par
                     new Type[] { Type.SQL_BIGINT },
                     new short[] {  Tokens.OPENBRACKET, Tokens.QUESTION, Tokens.CLOSEBRACKET }),
 
+           new FunctionId("regexp_position", Type.SQL_BIGINT, FUNC_VOLT_REGEXP_POSITION, -1,
```
paulmartel: What is the history and rationale for this function name?
Has a function by this exact name been previously implemented by some other popular database vendor, or is it a new invention? I had trouble finding one online.
If so, what exact regexp format(s) do they support?

I believe that there is a similar function in the SQL 2011 standard named "position_regex"
(words transposed and no final 'p'). 
It has a slightly different (keyword based) syntax and more extensive options. It uses XQuery regexp format which may not be 100% compatible with the boost normal (Perl format) regexp that is used here -- maybe that in itself is a reason not to use that standard name.

src/ee/expressions/stringfunctions.h
```diff
@@ -636,6 +637,81 @@ template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_CURRENCY>(const std::vect
...
+
+    int32_t lenSource = source.getObjectLength_withoutNull();
+    char* sourceChars = reinterpret_cast<char*>(source.getObjectValue_withoutNull());
+    std::string sourceStr(sourceChars, lenSource);
+
+    int32_t lenPat = pat.getObjectLength_withoutNull();
+    char* patChars = reinterpret_cast<char*>(pat.getObjectValue_withoutNull());
+    std::string patStr(patChars, lenPat);
+
+    boost::regex_constants::syntax_option_type matchOpts = boost::regex_constants::normal;
```
paulmartel: What is the rationale for using Perl format regexp's -- are there better alternatives?

Release 5.8 does not support C++ 11 features, but a future release (possibly 6.0) MAY support some C++ 11 features. C++ 11 specifically has some native regexp support functions. Would it be possible for us to eliminate the boost dependency and migrate to a native implementation at that point?
Would there be any noticeable behavior differences -- such as the supported regexp syntax.
It might make sense to "future-proof" the implementation by using a boost configuration that we know can be consistently re-implemented with the planned C++ 11 native support.

paulmartel:
I have reviewed this changeset as a delta from master for expedience in providing feedback.
But my understanding is that the intended short-term target for these changes is a patch release to v5.8 from which master has already diverged. 
To avoid delays in making these changes available in a v5.8 patch release, it would be more appropriate to submit code changes to the latest release_5.8.x branch (currently 4f437d0).

paulmartel:
The VoltDB build system enforces a small number of our coding standards.
One of them is the use of spaces instead of tabs in source code.
Another is the absence of any kind of whitespace at the end of source code lines.
There is an ant rule (in the voltdb directory, run "ant licensecheck") that checks for violations of these standards and lists the offending source files. This rule will immediately halt our automated test suites when it detects violations.
There is also an automated source code cleanup script that will correctly solve these problems, leaving neatly formatted source code and allowing "ant licensecheck" to then succeed in most cases. In the voltdb directory, run "tools/licensescheck.py --fixws".
Other useful ant rules to run prior to submitting changesets are "ant eecheck" which builds and runs a quick set of C++ unit tests -- and "ant check" which runs a longer battery of regression tests including the junit tests.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -2068,4 +2068,271 @@ public void testBin() throws NoConnectionsException, IOException, ProcCallExcept
             validateTableColumnOfScalarVarchar(result, new String[]{binString});
         }
     }
+    
+    public void testDateadd() throws NoConnectionsException, IOException, ProcCallException {
+        System.out.println("STARTING test DATEADD function tests");
```
paulmartel: Needs test coverage for NULL passed to either or both of bigint and timestamp argument.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -2068,4 +2068,271 @@ public void testBin() throws NoConnectionsException, IOException, ProcCallExcept
...
+        try {
+            client.callProcedure("@AdHoc", "SELECT REGEXP_POSITION(DESC, '[a-z](\\d+)[A-Z]', 'k') FROM P1 WHERE REGEXP_POSITION(DESC, '[a-z](\\d+)[A-Z]', 'k') > 0");
+        } catch (ProcCallException e) {
+            assertEquals(ClientResponse.GRACEFUL_FAILURE, e.getClientResponse().getStatus());
+            assertTrue(e.getClientResponse().getStatusString().contains("illegal match flags"));
+        }        
+        
+        cr = client.callProcedure("@AdHoc", "DELETE FROM P1 WHERE ID = 200");
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+    }
```
paulmartel: needs test coverage for null arguments.
needs test coverage for non-ascii characters in DESC and/or in the pattern.
verify that positions are returned as character positions, not byte positions after multi-byte characters.

src/ee/expressions/stringfunctions.h
```diff
@@ -636,6 +637,81 @@ template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_CURRENCY>(const std::vect
...
+            for(std::string::iterator it = flagStr.begin(); it != flagStr.end(); ++it) {
+                switch (*it) {
+                    case 'c':
+                        break;
+                    case 'i':
+                        matchOpts |= boost::regex_constants::icase;
+                        break;
+                    default:
+                        throw SQLException(SQLException::data_exception_invalid_parameter, "illegal match flags");
+                }
```
paulmartel: It should probably be an error to specify contradictory flags like 'ci' or 'ic'.
Possibly, it should also be an error to repeat flags like 'cc' or 'iii'
Where do the specifications come from for the set of supported match flag characters and their meanings? Are they intended to be case sensitive?

paulmartel:
I have finished my initial review. Please let me know via e-mail or by commenting here if you have further questions.

paulmartel:
Correction: one last review comment. These functions need some basic cases covered by c++ unit tests. See tests/ee/common/nvalue_test.cpp where there are tests for other sql functions.

PR #3037, opened 2015-11-23
A DDL file for geospatial example app

wolffcm:


BillWhite:
Well, the only thing I would add is that it would be better to have some documentation
for the tables.  I can guess what you have in mind, but I might misinterpret them.  Since
this is all green field development, and we don't have the same kind of common vocabulary,
there is a lot of room for misinterpretation.

wolffcm:
You're right, I'll add some comments to the DDL that clarify.

wolffcm:
I've added more verbosity, some `partition table` commands and also a short README.  @BillWhite, could you have a look and let me know what you think?

examples/geospatial/README.md
```diff
@@ -0,0 +1,16 @@
...
+devices.  Advertisers periodically make bids of how much they will pay
+to have an ad served to a user for a specified duration and within a
+particular region.
+
+The demo simulates users opening up ad-serving apps on their mobile
+devices which then request an ad based on the device's location and
+the current time.  This operation will require executing a geospatial
+query to see if the location of the device is contained by any of the
+regions for which a bid is active.  Many thousands of such requests
+can be performed per second.
```
BillWhite: We hope many thousands of such requests can be performed per second.

examples/geospatial/ddl.sql
```diff
@@ -0,0 +1,78 @@
...
+)
+AS
+  select
+    truncate(minute, ts) as ts_minute
+  , advertiser_id
+  , count(*)
+  , sum(bid_amount)
+  from impressions
+  group by ts_minute, advertiser_id;
+
```
BillWhite: If the impressions table is volatile, isn't this view volatile as well?  Can we keep up?

I'm not a view expert.  Maybe this is a foolish question.

BillWhite:
This seems much easier to understand.  I only have one question, and it's really just a question.

examples/geospatial/ddl.sql
```diff
@@ -0,0 +1,78 @@
...
+CREATE VIEW impressions_by_minute_by_advertiser
+(
+  ts_minute
+, advertiser_id
+, impression_count
+, sum_of_impression_revenue
+)
+AS
+  select
+    truncate(minute, ts) as ts_minute
```
wolffcm: This should be seconds, it's more exciting.

examples/geospatial/README.md
wolffcm: I'm trying to be optimistic.

examples/geospatial/ddl.sql
wolffcm: There will be some overhead to update the view when we update the source table.  So it might slow down our throughput a little.  The upside is that we'll have aggregations that are pre-computed and fetched much faster.

PR #3039, opened 2015-11-24
ENG-9348: Fix CopyOnWriteContext and COWIterator to maintain only a s…

wweiss-voltdb:
…ingle m_blocks map.

Also make failed snapshots non-fatal.

### DIFF
--- a/src/ee/storage/CopyOnWriteContext.cpp
+++ b/src/ee/storage/CopyOnWriteContext.cpp
@@ -82,7 +82,7 @@ CopyOnWriteContext::handleActivation(TableStreamType streamType)
 
     m_surgeon.activateSnapshot();
 
-    m_iterator.reset(new CopyOnWriteIterator(&getTable(), &m_surgeon, m_surgeon.getData()));
+    m_iterator.reset(new CopyOnWriteIterator(&getTable(), &m_surgeon));
 
     return ACTIVATION_SUCCEEDED;
 }
@@ -217,12 +217,10 @@ int64_t CopyOnWriteContext::handleStreamMore(TupleOutputStreamProcessor &outputS
                 // at least returned it's currentBlock to the lists.
                 if (allPendingCnt > 0) {
                     // We have orphaned or corrupted some tables. Let's make them pristine.
-                    TBMap allBlocks(m_surgeon.getData());
-                    TBMapI iter = allBlocks.begin();
-                    while (iter != allBlocks.end()) {
+                    TBMapI iter = m_surgeon.getData().begin();
+                    while (iter != m_surgeon.getData().end()) {
                         m_surgeon.snapshotFinishedScanningBlock(iter.value(), TBPtr());
-                        allBlocks.erase(iter);
-                        TBMapI iter = allBlocks.begin();
+                        iter++;
                     }
                 }
                 if (!m_surgeon.blockCountConsistent()) {
--- a/src/ee/storage/CopyOnWriteContext.cpp
+++ b/src/ee/storage/CopyOnWriteContext.cpp
@@ -44,7 +44,6 @@ CopyOnWriteContext::CopyOnWriteContext(
                                                                "COW of " + table.name(),
                                                                &table, NULL)),
              m_pool(2097152, 320),
-             m_blocks(surgeon.getData()),
              m_tuple(table.schema()),
              m_finishedTableScan(false),
              m_totalTuples(totalTuples),
@@ -52,6 +51,7 @@ CopyOnWriteContext::CopyOnWriteContext(
              m_blocksCompacted(0),
              m_serializationBatches(0),
              m_inserts(0),
+             m_deletes(0),
              m_updates(0)
 {
 }
@@ -82,7 +82,7 @@ CopyOnWriteContext::handleActivation(TableStreamType streamType)
 
     m_surgeon.activateSnapshot();
 
-    m_iterator.reset(new CopyOnWriteIterator(&getTable(), &m_surgeon, m_blocks));
+    m_iterator.reset(new CopyOnWriteIterator(&getTable(), &m_surgeon, m_surgeon.getData()));
 
     return ACTIVATION_SUCCEEDED;
 }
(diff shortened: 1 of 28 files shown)


src/ee/storage/CopyOnWriteContext.cpp
```diff
@@ -82,7 +82,7 @@ CopyOnWriteContext::handleActivation(TableStreamType streamType)
 
     m_surgeon.activateSnapshot();
 
-    m_iterator.reset(new CopyOnWriteIterator(&getTable(), &m_surgeon, m_blocks));
+    m_iterator.reset(new CopyOnWriteIterator(&getTable(), &m_surgeon, m_surgeon.getData()));
```
nshi: The iterator constructor can grab the blocks directly from the table surgeon.

src/ee/storage/CopyOnWriteContext.cpp
```diff
@@ -203,32 +187,58 @@ int64_t CopyOnWriteContext::handleStreamMore(TupleOutputStreamProcessor &outputS
...
+                // at least returned it's currentBlock to the lists.
+                if (allPendingCnt > 0) {
+                    // We have orphaned or corrupted some tables. Let's make them pristine.
+                    TBMap allBlocks(m_surgeon.getData());
+                    TBMapI iter = allBlocks.begin();
+                    while (iter != allBlocks.end()) {
+                        m_surgeon.snapshotFinishedScanningBlock(iter.value(), TBPtr());
+                        allBlocks.erase(iter);
+                        TBMapI iter = allBlocks.begin();
+                    }
```
nshi: It is a relatively expensive operation to iterate through all the blocks and modifying the map for each block it looks at. What you want is just an iteration over all buckets and swap them to the right bucket. Something like the following would be more efficient. It avoids the deep copy and rebalancing the tree all the time.

```c++
TBMapI iter = m_surgeon.getData();
while (iter != m_surgeon.getData().end()) {
    m_surgeon.snapshotFinishedScanningBlock(iter.value(), TBPtr());
    iter++;
}
```

src/ee/storage/CopyOnWriteIterator.cpp
```diff
@@ -42,6 +42,42 @@ CopyOnWriteIterator::CopyOnWriteIterator(
 }
 
 /**
+ * When a tuple is "dirty" it is still active, but will never be a "found" tuple
+ * since it is skipped. The tuple may be dirty because it was deleted (this is why it is always skipped). In that
+ * case the CopyOnWriteContext calls this to ensure that the iteration finds the correct number of tuples
+ * in the used portion of the table blocks and doesn't overrun to the uninitialized block memory because
+ * it skiped a dirty tuple and didn't end up with the right found tuple count upon reaching the end.
+ */
```
nshi: The second part of the comment seems to describe the COW iteration rather than this method. The purpose is method is just to determine whether a tuple is before or after the current scanning point.

nshi: Ah, I see. This was an old comment that got copied over.

src/ee/storage/CopyOnWriteIterator.h
```diff
@@ -35,31 +35,42 @@ class CopyOnWriteIterator : public TupleIterator {
...
-     * since it is skipped. The tuple may be dirty because it was deleted (this is why it is always skipped). In that
-     * case the CopyOnWriteContext calls this to ensure that the iteration finds the correct number of tuples
-     * in the used portion of the table blocks and doesn't overrun to the uninitialized block memory because
-     * it skiped a dirty tuple and didn't end up with the right found tuple count upon reaching the end.
-     */
-    bool needToDirtyTuple(const char *blockAddress, const char *tupleAddress) {
-        if (blockAddress < m_currentBlock->address()) {
-            return false;
-        }
+    bool needToDirtyTuple(char *tupleAddress);
```
nshi: `const`?

src/frontend/org/voltdb/DefaultSnapshotDataTarget.java
```diff
@@ -383,6 +388,10 @@ public void run() {
         if (m_onCloseHandler != null) {
             m_onCloseHandler.run();
         }
+        if (m_reportedSerializationFailure != null) {
```
nshi: 11 lines above where the completed flag is written to the file, you also need to check for serialization failure. Otherwise the file will be incorrectly marked as completed.

src/frontend/org/voltdb/SnapshotSiteProcessor.java
```diff
@@ -618,14 +619,13 @@ public void run()
...
                         } catch (Throwable t) {
-                            if (m_lastSnapshotSucceded) {
+                            if (m_perSiteLastSnapshotSucceded) {
                                 if (t instanceof StreamSnapshotTimeoutException ||
                                         t.getCause() instanceof StreamSnapshotTimeoutException) {
                                     //This error is already logged by the watchdog when it generates the exception
                                 } else {
                                     SNAP_LOG.error("Error while attempting to write snapshot data", t);
                                 }
-                                m_lastSnapshotSucceded = false;
```
nshi: You still need this here or it will print the same error many times in a roll.

src/frontend/org/voltdb/TableStreamer.java
```diff
@@ -107,7 +109,13 @@ public boolean activate(SystemProcedureExecutionContext context, boolean undo, b
 
         Pair<Long, int[]> serializeResult = context.tableStreamSerializeMore(m_tableId, m_type, outputBuffers);
         if (serializeResult.getFirst() == SERIALIZATION_ERROR) {
-            VoltDB.crashLocalVoltDB("Failure while serializing data from table " + m_tableId, false, null);
+            // Cancel the snapshot here
+            for (DBBPool.BBContainer container : outputBuffers) {
+                container.discard();
+            }
+            StreamSnapshotFailedException ex = new StreamSnapshotFailedException("Snapshot of table " + m_tableId + " failed to complete.");
```
nshi: Stream snapshot is only used for rejoin and elastic join. It's kind of weird to throw a `StreamSnapshotFailedException` here.

src/frontend/org/voltdb/TableStreamer.java
```diff
@@ -107,7 +109,13 @@ public boolean activate(SystemProcedureExecutionContext context, boolean undo, b
...
+            m_tableTasks.get(m_tableId).m_target.reportSerializationFailure(ex);
```
nshi: The tasks are not indexed by table IDs. They are for different output targets. You need to iterate through all of them and report the failure.

PR #3041, opened 2015-11-24
Enhance basic validation for polygon in frontend

hhakimi53:
Add more validation to frontend to check if polygon is valid
- Start and end point of polygon are equal
- Each loop in polygon are composed for 4 vetrices  = First 3 vertices +  end vertex which is same as start vertex.
- There is atleast one loop in the polygon.

### DIFF
--- a/src/frontend/org/voltdb/types/GeographyValue.java
+++ b/src/frontend/org/voltdb/types/GeographyValue.java
@@ -52,9 +52,11 @@ public class GeographyValue {
         m_loops = new ArrayList<List<XYZPoint>>();
         for (List<PointType> loop : loops) {
             List<XYZPoint> oneLoop = new ArrayList<XYZPoint>();
-            for (int i = 0; i < (loop.size() - 1); ++i) {
+            for (int i = 0; i < loop.size(); ++i) {
                 oneLoop.add(XYZPoint.fromPointType(loop.get(i)));
             }
+            diagnoseLoop(oneLoop, "Invalid loop for GeographyValue: ");
+            oneLoop.remove(oneLoop.size() - 1);
             m_loops.add(oneLoop);
         }
     }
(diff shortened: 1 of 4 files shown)


tests/frontend/org/voltdb/types/TestGeographyValue.java
```diff
@@ -126,6 +126,14 @@ public void testWktParsingNegative() {
         assertWktParseError("premature end of input", "POLYGON ((80 80, 60 60, 70 70,");
         assertWktParseError("missing closing parenthesis", "POLYGON ((80 80, 60 60, 70 70, (30 15, 15 30, 15 45)))");
         assertWktParseError("unrecognized token", "POLYGON ((80 80, 60 60, 70 70, 80 80)z)");
-        assertWktParseError("unrecognized input after WKT", "POLYGON ((80 80, 60 60, 70 70, 90 90))blahblah");
+        assertWktParseError("unrecognized input after WKT", "POLYGON ((80 80, 60 60, 70 70, 90 90, 80 80))blahblah");
```
wolffcm: GeographyValue also has a public constructor that accepts a List<List<PointType>> (a list of loops).  You should add tests for this constructor too.

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -447,10 +461,21 @@ private static int unflattenLoopFromBuffer(ByteBuffer inBuffer, List<XYZPoint> l
                     break;
                 case ')':
                     if (currentLoop == null) {
-                        throw new IllegalArgumentException(msgPrefix + "missing opening parenthesis");
+                        throw new IllegalArgumentException(msgPrefix + "polygon should contain atleast one loop, " +
```
wolffcm: Actually I'm not sure this is the right place for these checks because you want to enforce these conditions for the constructor that accepts List<List<PointType>> as well.  Maybe add a private isValid() method that can be called from everywhere an instance is created?

tests/frontend/org/voltdb/types/TestGeographyValue.java
hhakimi53: I did not check for that one, that also needs to be covered. Thanks for pointing that out.

wolffcm:
This looks good.  Thanks for making these changes.  I guess you'll have some conflicts, maybe with Bill's recent work?

hhakimi53:
Thanks Chris for reviewing the changes. Will merge in the recent changes and check-in once Jenkins is blue.

PR #3042, opened 2015-11-24
Eng 8559 planner npe

paulmartel:


src/frontend/org/voltdb/planner/parseinfo/StmtSubqueryScan.java
```diff
@@ -344,16 +346,6 @@ public boolean canRunInOneFragment() {
...
                 return false;
             }
-
-            //
-            // (2) Contains a partition column:
-            //     This is the interesting case that we are going to support.
-            //     At this point, subquery does not contain LIMIT/OFFSET.
-            //     But if the statement has a distinct aggregate, retain the
-            //     send/receive and compute the distinct result on the coordinator.
-            if ( selectStmt.hasAggregateDistinct() ) {
```
wolffcm: Nice.

src/frontend/org/voltdb/planner/parseinfo/StmtSubqueryScan.java
```diff
@@ -291,12 +291,14 @@ public void processTVE(TupleValueExpression expr, String columnName) {
      */
     public boolean canRunInOneFragment() {
         assert(m_subqueriesPartitioning != null);
-        if (! m_subqueriesPartitioning.requiresTwoFragments()) {
```
wolffcm: I think I see the issue here.  Why didn't you just remove the exclamation point here to solve this issue?

wolffcm: I guess if you are in "forceSP" mode, then you might return false here when you could actually return true?

The logic in StatementPartitioning never fails to make me scratch my head.

wolffcm:
Those are my comments.

src/frontend/org/voltdb/planner/parseinfo/StmtSubqueryScan.java
paulmartel: I tried the smaller change, but then I got a regression that I did not completely understand, and I could not follow exactly what about requiresTwoFragments was triggering it. That's when I decided to try `getCountOfPartitionedTables() == 0` which seemed to cover the various regression cases I was seeing but with simpler logic.

wolffcm: Ok.  The logic you have here seems safe to me.  I'm fine with you merging this when you're ready.

paulmartel:
squash-merged this

PR #3044, opened 2015-11-24
Eng 8575 no java heap warning for most standard configurations - better wording in others

JadenW:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2914,9 +2917,9 @@ private void checkHeapSanity(boolean isPro, int tableCount, int sitesPerHost, in
     {
         long megabytes = 1024 * 1024;
         long maxMemory = Runtime.getRuntime().maxMemory() / megabytes;
-        long drRqt = isPro ? 128 * sitesPerHost : 0;
+        // DRv2 now is off heap
+        // long drRqt = isPro ? 128 * sitesPerHost : 0;
```
jhugg: Don't comment out code. Just remove code that isn't used anymore. Add the comment in the commit message.

manjujames:
This will show the warning message every time you update the catalog, regardless of whether new tables were added or not. Otherwise the changes look good.

src/frontend/org/voltdb/RealVoltDB.java
JadenW: Removed. thanks .

JadenW:
Yes. Once the heap estimation is above the max heap avaible, this warning will continue to reappear to console log every time during UAC. It should disappear when the heap estimation drop below the threshold.

JadenW:
Squashed commit

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2914,9 +2917,8 @@ private void checkHeapSanity(boolean isPro, int tableCount, int sitesPerHost, in
     {
         long megabytes = 1024 * 1024;
         long maxMemory = Runtime.getRuntime().maxMemory() / megabytes;
-        long drRqt = isPro ? 128 * sitesPerHost : 0;
+        // DRv2 now is off heap
```
jhugg: What does this comment add here? It only makes sense if you knew the old code.

PR #3045, opened 2015-11-24
Eng 9443 string reduction

paulmartel:


wolffcm:
@paulmartel I have begun reviewing this.  There's a lot here---there's too many changes to NValue.hpp for the web interface to show them.  But I realize many of these changes are very simple.

There will certainly be a few conflicts with the geospatial branch, but I don't see anything too terrible so far.

It will be interesting to see if there's any performance implications to these changes (especially adding `&` to NValue parameters).

I'll finish up reviewing before you get back from vacation.

wolffcm:
In NValue.hpp (GitHub won't show the diff with so many changes) in function serializerInlineObjectToTupleStorage, this function seems to assume that both the source and the destination are inlined.  Is this always the case?

wolffcm:
In the header comment for serializeToTupleStorage, I had hard time following the discussion of the "allocateForObjects".  Should the last sentence end, "*because* it is not in the required object pointer format"?

wolffcm:
In the same method, serializeToTupleStorage, depending on the setting of `allocateForObjects` and `m_sourceInlined`, we will either allocate into the temp string pool, or into the dataPool passed as an argument to the method.

I'm guessing that when we store a string value to a persistent table, we use the dataPool passed in to the method, but that the value of this input is NULL?  Do I have this right?

src/ee/common/ThreadLocalPool.cpp
```diff
@@ -106,140 +106,66 @@ ThreadLocalPool::~ThreadLocalPool() {
...
 #ifdef MEMCHECK
 /// Persistent string pools with their compaction are completely bypassed for
 /// the memcheck build. It just does standard C++ heap allocations and
 /// deallocations.
-char* ThreadLocalPool::allocateRelocatable(char** referrer_ignored, int32_t sz)
+ThreadLocalPool::Sized* ThreadLocalPool::allocateRelocatable(char** referrer_ignored, int32_t sz)
 {
-    char* data = new char[sz];
-    return data;
+    return new (new char[sizeof(Sized) + sz]) Sized(sz);
```
wolffcm: I had to think about this for a minute.  This is placement new, where the place your specifying is just a character array allocated using regular new, correct?  This makes sense for memcheck mode.

wolffcm:
That's all my comments.  This seems like a nice reduction in technical debt in this area.

The EE and JUnit tests all pass in Valgrind in memcheck mode, right?  Just making sure.

PR #3048, opened 2015-11-30
Eng 9491 boostregexp 5.8

hhakimi53:
Update third party boost regex library file - it was build using PIC flag. 

Testing:
- With this the code builds on Centos 5.8.


build.py
```diff
@@ -106,6 +106,8 @@
 # where to find the tests
 CTX.TEST_PREFIX = "tests/ee/"
 
+CTX.LDFLAGS += """ -L../../%s/boost/lib/%s -lboost_regex""" % (CTX.THIRD_PARTY_INPUT_PREFIX, CTX.PLATFORM)
+
 ###############################################################################
```
BillWhite: We looked at this, and this seems to be the right thing.

BillWhite:
That's all I know about this.

PR #3049, opened 2015-11-30
Eng 9489 contributions - For review  purpose, don't merge

hhakimi53:
Creating this as integration for SQL function integration work. 

This branch incorporates boost regex library needed for implementing regex SQL function and applying latest SQL function changes provided in patch.


src/ee/expressions/datefunctions.h
```diff
@@ -394,4 +425,238 @@ template<> inline NValue NValue::callConstant<FUNC_CURRENT_TIMESTAMP>() {
...
+    int64_t interval = number.castAsBigIntAndGetValue();
+    if (interval > PTIME_MAX_YEARS || interval < PTIME_MIN_YEARS) {
+        throw SQLException(SQLException::data_exception_numeric_value_out_of_range, "interval is too large for DATEADD function");
+    }
+
+    boost::posix_time::ptime ts;
+    micros_to_ptime(date.getTimestamp(), ts);
+
+    try {
+        ts += boost::gregorian::years((int) interval);
```
hhakimi53: Suggest to use C++ style cast here (static_cast<int>).

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -223,9 +235,19 @@ private FunctionId(String name, Type type, int id, int typeParameter, Type[] par
...
-        };
+                    
+            new FunctionId("dateadd", Type.SQL_TIMESTAMP, FUNC_VOLT_DATEADD, -1, 
+                    new Type[] { Type.SQL_VARCHAR, Type.SQL_BIGINT, Type.SQL_TIMESTAMP }, 
+                    new short[] { Tokens.OPENBRACKET, Tokens.X_KEYSET, 11, Tokens.YEAR,
+                                  Tokens.QUARTER, Tokens.MONTH, Tokens.DAY, Tokens.HOUR, Tokens.MINUTE, Tokens.SECOND,
+                                  Tokens.MILLIS, Tokens.MILLISECOND, Tokens.MICROS, Tokens.MICROSECOND, Tokens.COMMA,
+                                  Tokens.QUESTION, Tokens.COMMA, Tokens.QUESTION, Tokens.CLOSEBRACKET }),
+
+           new FunctionId("regexp_position", Type.SQL_BIGINT, FUNC_VOLT_REGEXP_POSITION, -1,
```
hhakimi53: Earlier Paul's comment:What is the history and rationale for this function name?
Has a function by this exact name been previously implemented by some other popular database vendor, or is it a new invention? I had trouble finding one online.
If so, what exact regexp format(s) do they support?

I believe that there is a similar function in the SQL 2011 standard named "position_regex"
(words transposed and no final 'p').

hhakimi53:
I don't see junit tests for the SQL function changes. 

Please provide updated junit tests based on review feedback in https://github.com/VoltDB/voltdb/pull/3034.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
hhakimi53: Huawei's response: 

The motivation of regexp_position is that our customers want to do regular-expression based pattern matching. However, VoltDB does not have such functions until now. So we decided to add one. 

Our customers want a function similar to Oracle’s regexp_like (https://docs.oracle.com/cd/B12037_01/server.101/b10759/conditions018.htm). As VoltDB does not support BOOLEAN type, we decided to return the start position of a matched sub-expression. This is why we name the function regexp_position. It means to get the start position of the first sub-string in a source string that matches the given pattern. 

We adopted the syntax of PostgreSQL, as we found that the name of most VoltDB functions are same as PostgreSQL. PostgreSQL has a function regexp_match(http://www.postgresql.org/docs/9.3/static/functions-matching.html), which returns all matched sub-string given the pattern. However, we cannot introduce the function directly as VoltDB function does not support return of a list. Thus, we adopt the same function arguments, but changed the return value. As our customers request this function in urgency, we decided to provide a workable version quickly and improve it in future.

tests/ee/common/nvalue_test.cpp
```diff
@@ -3328,6 +3328,55 @@ TEST_F(NValueTest, TestTimestampStringParseWithLeadingAndTrailingSpaces)
     }
 }
 
+TEST_F(NValueTest, TestDateadd) {
```
hhakimi53: Can you add additional negative tests cases. Example for date add year, quarter and month, select an interval which is large enough that it generates out of range exception.

tests/ee/common/nvalue_test.cpp
```diff
@@ -3328,6 +3328,55 @@ TEST_F(NValueTest, TestTimestampStringParseWithLeadingAndTrailingSpaces)
...
+    assert(ExecutorContext::getExecutorContext() == NULL);
+    Pool* testPool = new Pool();
+    getExecutorContextForTest(testPool);
+
+    NValue result;
+    NValue interval = ValueFactory::getBigIntValue(1);
+    NValue midSeptember = ValueFactory::getTimestampValue(1000000000000000);
```
hhakimi53: Will suggest to either add a comment on what the timestamp midSeptember translates to or a check the timestamp value is as expected will be helpful in here. It will make the test more readable.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -1798,6 +1798,303 @@ public void testBin() throws NoConnectionsException, IOException, ProcCallExcept
         }
     }
 
+    public void testDateadd() throws NoConnectionsException, IOException, ProcCallException {
+        System.out.println("STARTING test DATEADD function tests");
+
```
hhakimi53: Overall test cases look elaborate and good. 

Please add some additional test cases:
- Input number to DateAdd() be a bigger valid interval and verify the result.
- Input parameters to DateAdd() being Null. (for example: "Select DateAdd(Null, 2, TM) from P2 where ID=1001;  Select DateAdd(year, null, TM) from P2 where ID=1001);"
- Input number parameter to DateAdd() being out-0f-bound.

src/ee/expressions/stringfunctions.h
```diff
@@ -636,6 +637,75 @@ template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_CURRENCY>(const std::vect
...
+        if (!flags.isNull()) {
+            if (flags.getValueType() != VALUE_TYPE_VARCHAR) {
+                 throwCastSQLException(flags.getValueType(), VALUE_TYPE_VARCHAR);
+            }
+
+            int32_t lenFlags = flags.getObjectLength_withoutNull();
+            const char* flagChars = reinterpret_cast<const char*>(flags.getObjectValue_withoutNull());
+
+            for(int i = 0; i < lenFlags; i++) {
+                switch (*flagChars) {
```
hhakimi53: Should it be an error to specify contradictory flags like 'ci' or 'ic'?  Possibly, it should also be an error to repeat flags like 'cc' or 'iii' 
Where do the specifications come from for the set of supported match flag characters and their meanings? Are they intended to be case sensitive?

src/ee/expressions/stringfunctions.h
```diff
@@ -636,6 +637,75 @@ template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_CURRENCY>(const std::vect
...
+
+        if (regex_search(sourceChars, what, patExpr, matchFlags)) {
+            return getBigIntValue(getCharLength(sourceChars, what.position()) + 1);
+        } else {
+            return getBigIntValue(0);
+        }
+    } catch (boost::regex_error& e) {
+        throw SQLException(SQLException::data_exception_invalid_parameter, "illegal pattern string");
+    } catch (std::runtime_error& e) {
+        throw SQLException(SQLException::dynamic_sql_error, "regular pattern is too complicate.");
```
wolffcm: This message should be, "Pattern input to REGEXP_POSITION is too complicated."

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
```diff
@@ -2420,6 +2420,62 @@ VoltXMLElement voltAnnotateFunctionXML(VoltXMLElement exp)
...
+            case Tokens.MILLISECOND :
+                implied_argument = "MILLISECOND";
+                dateadd_func = FunctionForVoltDB.FunctionId.FUNC_VOLT_DATEADD_MILLISECOND;
+                break;
+            case Tokens.MICROS:
+            case Tokens.MICROSECOND :
+                implied_argument = "MICROSECOND";
+                dateadd_func = FunctionForVoltDB.FunctionId.FUNC_VOLT_DATEADD_MICROSECOND;
+                break;
+            default:
```
hhakimi53: Minor nit: The other keywords that can be supplied to datepart argument of dateadd() are dayofyear, week and weekday, which are supported by other vendors. For those keywords, error message stating the keywords are not supported would be more helpful to user.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -1798,6 +1798,303 @@ public void testBin() throws NoConnectionsException, IOException, ProcCallExcept
...
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+
+        vt = client.callProcedure("@AdHoc", "SELECT DATEADD(YEAR, 1, TM) FROM P2 WHERE ID = 20006;").getResults()[0];
+        assertTrue(vt.advanceRow());
+        assertEquals(null, vt.getTimestampAsTimestamp(0));
+    }
+
+    public void testRegexpPosition() throws Exception {
+        System.out.println("STARTING testRegexpPosition");
+
```
hhakimi53: What would be expected behavior for regexp_position() with input flag being 'ci', 'ic', 'ii' and invalid flags? Can you add additional tests to cover those scenarios.

hhakimi53: That's end of this review round.

tests/ee/common/nvalue_test.cpp
junhuazhu: Added and sent it via email

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
junhuazhu: Added.

src/ee/expressions/stringfunctions.h
junhuazhu: fixed.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
junhuazhu: We return the exact message as FUNC_VOLT_TRUNCATE_TIMESTAMP as we don't know error message in SQL parser return to clients.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
junhuazhu: We expect the last value will take effect. tests are added.

junhuazhu: We expect the last value will take effect. tests are added.

src/ee/expressions/stringfunctions.h
junhuazhu: We expect the last value will take effect.

hhakimi53: Ok, looks like that's how PostGRESql is also using flag in regular expression function.

tests/ee/common/nvalue_test.cpp
junhuazhu: add a comment to show the timestamp in string

tests/ee/common/nvalue_test.cpp
hhakimi53: Negative test for out-of-range is missing. Please incorporate it those.

src/ee/expressions/datefunctions.h
```diff
@@ -30,6 +32,27 @@ static const int64_t GREGORIAN_EPOCH = -12212553600000000;  // 1583-01-01 00:00:
...
 
+static const int64_t PTIME_MAX_YEARS = 10000;
+static const int64_t PTIME_MIN_YEARS = boost::gregorian::date(boost::gregorian::min_date_time).year();
+static const int64_t PTIME_MAX_YEAR_INTERVAL = PTIME_MAX_YEARS - PTIME_MIN_YEARS;
+static const int64_t PTIME_MIN_YEAR_INTERVAL = -PTIME_MAX_YEAR_INTERVAL;
+static const int64_t PTIME_MAX_QUARTER_INTERVAL = PTIME_MAX_YEAR_INTERVAL * 4;
+static const int64_t PTIME_MIN_QUARTER_INTERVAL = -PTIME_MAX_QUARTER_INTERVAL;
+static const int64_t PTIME_MAX_MONTH_INTERVAL = PTIME_MAX_YEAR_INTERVAL * 12;
+static const int64_t PTIME_MIN_MONTH_INTERVAL = -PTIME_MAX_MONTH_INTERVAL;
+static const int64_t PTIME_MAX_DAY_INTERVAL = PTIME_MAX_YEAR_INTERVAL * 366;
```
hhakimi53: Above value uses days in a leap year. Should max # days for 10000 years be calculated as
PTIME_MAX_DAY_INTERVAL = PTIME_MAX_YEAR_INTERVAL * 365 + (PTIME_MAX_YEARS / 4); ?

junhuazhu: The MAX/MIN Intervals are used to prevent int64_t overflow when constructing boost::posix_time::time_duration objects. boost::posix_time::time_duration constructors do not check int64_t overflow when computing ticks from given hours, minutes, seconds, and fractional seconds. So these Intervals do not be be accurate.

PR #3052, opened 2015-12-03
Don't create inlined project in inlined scan node

wolffcm:
This is a fix for an issue where the planner was resolving TVE column indices incorrectly, based on a project node that's never used in the planner.

For inlined scan nodes, we don't need a project node since there will be no temp table created, the enclosing plan node will access the data directly.  So the schema used by the enclosing node to access the data should be the same as the table being scanned.


wolffcm:
It might help to append `?w=0` to the url above to see changes compared whitespace-insensitively.

src/frontend/org/voltdb/plannodes/AbstractScanPlanNode.java
```diff
@@ -275,48 +275,49 @@ public void generateOutputSchema(Database db)
...
+            m_outputSchema = projectionNode.getOutputSchema().copyAndReplaceWithTVE();
+            m_hasSignificantOutputSchema = false; // It's just a cheap knock-off of the projection's
+        }
+        else {
+            // We come here if this node is an inline scan node, or if m_tableScanSchema is empty.
+            //
+            // If this scan node is inlined into another node, then there is no output temp
+            // table for this node, so the output schema will be the same as the
+            //
+            // When would m_tableScanSchema be empty?  I'm not sure (--cwolff, 12/02/15)
```
jxchinaren: Like: select NOW from table;
Maybe there are other cases.

jxchinaren:
Looks good to me.

wolffcm:
Squash-merged on a separate branch.

PR #3054, opened 2015-12-03
Eng 9163 sqlcov postgres stabilize

johnTcrawford:


wolffcm:
I'll have a look at this.  I am woefully unfamiliar with sqlcoverage, so I might have some questions.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -56,10 +58,63 @@
         m_PostgreSQLTypeNames.put("int8", "BIGINT");
         m_PostgreSQLTypeNames.put("float8", "FLOAT");
         m_PostgreSQLTypeNames.put("numeric", "DECIMAL");
-        m_PostgreSQLTypeNames.put("bytea", "VARBINARY");
+        m_PostgreSQLTypeNames.put("varbit", "VARBINARY");
         m_PostgreSQLTypeNames.put("char", "CHARACTER");
         m_PostgreSQLTypeNames.put("text", "VARCHAR");
     }
+    // Captures up to 6 order-by columns; beyond those will be ignored
```
wolffcm: I'm unfamiliar with this class.  I guess it replaces an analogous class for HSQL.  It seems like it's main job is to translate various Volt-isms into PostgreSQL-isms?

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -121,13 +176,378 @@ private PostgreSQLBackend(Connection dbconn) {
...
+     * (int-valued) group by the <i>multiplier</i>; may be <b>null</b>, in
+     * which case it is ignored.
+     * @param groups - zero or more groups found within the <i>queryPattern</i>
+     * (e.g. "column").
+     * @return the <i>query</i>, transformed in the specified ways (possibly
+     * unchanged).
+     * @throws NullPointerException if <i>query</i>, <i>queryPattern</i>,
+     * <i>initText</i>, <i>prefix</i>, or <i>suffix</i> is <b>null</b>.
+     */
+    @SuppressWarnings("unused")
```
wolffcm: Using regular expressions to parse SQL has been notoriously difficult to maintain.  (I am thinking of sqlcmd here.)  We're in a little better position here since we can control the tool that is generating the input SQL (sqlcoverage).

I wonder how this translation compares to what was in the HSQL backend class, is there more or less tweaks needed to get equivalent statements?

I don't have a suggestion about what to do here, I'm just curious.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -121,13 +176,378 @@ private PostgreSQLBackend(Connection dbconn) {
...
+                "DOW FROM", ")+1", null, false, false, "column");
+    }
+
+    /** Modify a query containing an EXTRACT(DAY_OF_YEAR FROM ...) function,
+     *  which PostgreSQL does not support, and replace it with
+     *  EXTRACT(DOY FROM ...), which is an equivalent that PostgreSQL
+     *  does support. */
+    static private String transformDayOfYearQuery(String dml) {
+        return transformQuery(dml, dayOfYearQuery, "EXTRACT ( ",
+                "DOY FROM", ")", null, false, false, "column");
```
wolffcm: DOY --> DAY?

tests/scripts/examples/sql_coverage/advanced-template.sql
```diff
@@ -135,12 +138,12 @@ SELECT * FROM @fromtables Q29 WHERE Q29._variable[@columntype] _maybe LIKE '!%'
...
 SELECT __[#numone]        Q36,      CASE WHEN   A._variable[#arg @columntype] _cmp @comparableconstant THEN   A._variable[#numone @columntype]            ELSE   A.__[#arg] @aftermath END FROM @fromtables A WHERE @columnpredicate
 SELECT __[#arg]           Q37,      CASE WHEN   A._variable[#arg @columntype] _cmp @comparableconstant THEN   A.__[#arg]                                                               END FROM @fromtables A WHERE @columnpredicate
 --- CASE WHEN like DECODE
-SELECT * FROM @fromtables Q38 WHERE CASE      Q38._variable[#arg @columntype] WHEN @comparableconstant THEN Q38._variable[#numone @columntype] @aftermath ELSE Q38.__[#arg] @aftermath END _cmp @comparableconstant + 10
-SELECT * FROM @fromtables Q39 WHERE CASE      Q39._variable[#arg @columntype] WHEN @comparableconstant THEN Q39._variable[#numone @columntype] @aftermath                              END _cmp @comparableconstant + 10
+SELECT * FROM @fromtables Q38 WHERE CASE      Q38._variable[#arg @columntype] WHEN @comparableconstant THEN Q38._variable[#numone @columntype] @aftermath ELSE Q38.__[#arg] @aftermath END _cmp (@comparableconstant + 10)
```
wolffcm: I wonder, what was the issue with the parentheses here?

wolffcm:
End of review round.  The most pressing comment I had was the DAY/DOY one.  Otherwise I just had questions.  

This looks very nice, I'm glad we're getting closer to being able to use PostGIS in sqlcoverage.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: Yes, to both questions: it is analogous to HsqlBackend.java (there is also a relavitely new NonVoltDBBackend.java, which both now inherit from, to capture the things that are similar in both); and yes, it "translates" Volt SQL into PostgreSQL SQL.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: There is a lot more here than in HsqlBackend.java, since HSQL and VoltDB syntax are very similar; in fact, HsqlBackend.java mainly just stops and starts the (HSQL) database; there is none of the "transforming" of SQL needed there.

As for using regular expressions, I don't see a good alternative. In my mind, the only major question was whether to do all this on the Java or Python side; I could make arguments either way. (The main arguments for Java are that all this could also be used for JUnit tests, some day, and I'm better at it; the main arguments for Python are that it's closer to the tests that actually make use of this stuff, and perhaps a little better at handling regular expressions - though once you get used to using double-backslashes (\\) to escape, there's not that much difference.)

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: DOY is PostgreSQL's equivalent of VoltDB's DAY_OF_YEAR. (Ours is a lot clearer!) So this is correct.

src/frontend/org/voltdb/PostgreSQLBackend.java
wolffcm: I like the choice you've made to use Java, because I like the idea of using PostgreSQL as the baseline in our JUnit suites.  More people use it, and it seems more tried-and-true than HSQL.

tests/scripts/examples/sql_coverage/advanced-template.sql
johnTcrawford: Yeah, I'll add a comment to explain that. Without the parens, PostgreSQL would get confused, when some but not all comparison operators were used, about how to parse the statement: basically, I think (some) comparison operators have higher precedence, so it would compare the left side with the "@comparableconstant" value, producing a boolean, and then try and fail to add 10 to that (or something like that). In contrast, both VoltDB and HSQL treat it the same, with or without the parens, and regardless of which operator is used.

johnTcrawford:
Thanks. I'll add that explanatory comment, and get this squash-merged.

PR #3056, opened 2015-12-03
sql_coverage_test.py, SQLCoverageReport.py: keep track separately of

johnTcrawford:
NPE's from VoltDB and from the comparison database (HSqlDB or
PostgreSQL), and now fail (& red, in the HTML "SQL Coverage Test
Summary") for the former, and a strong warning (in orange) for the
latter. Also, add a command-line message giving the total number of
mismatches, when that is greater than zero (matching the ones for NPE's
- 2 separate ones, now - and for crashes).

### DIFF
--- a/tests/scripts/SQLCoverageReport.py
+++ b/tests/scripts/SQLCoverageReport.py
@@ -382,7 +382,7 @@ def getTopSummaryLines(cmpdb, includeAll=True):
         topLines += "<td rowspan=2 align=center>Test Suite</td>"
     topLines += """
 <td colspan=5 align=center>SQL Statements</td>
-<td colspan=4 align=center>Test Failures</td>
+<td colspan=5 align=center>Test Failures</td>
 <td colspan=4 align=center>SQL Statements per Pattern</td>
 <td colspan=5 align=center>Time (min:sec)</td>
 </tr><tr>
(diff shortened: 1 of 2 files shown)


wolffcm:
It's probably worth noting that NullPointerExceptions are just one kind of Java exception that might happen in the planner.  I think it makes sense to talk about exceptions produced in Java more generally, to cover cases like ArrayOutOfBoundsException, which aren't NPEs but seem like they should be handled similarly.

wolffcm:
This looks good to me.  Is there any behavior change here in how NPEs in sqlcoverage are reported as red/blue orbs in Jenkins?  I guess that Jenkins just looks at the return code of ant to determine "success" or "failure."  But how does ant figure out if a sqlcoverage run didn't turn up any issues?

johnTcrawford:
Re. ArrayOutOfBoundsException: that's a good suggestion; I'll file a ticket. Possibly other exceptions could be included. Note, however, that we cannot simply fail whenever we receive any type of exception. For example, this is a normal result, for certain queries, which we see often in the results:
Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: Error compiling query: org.voltdb.planner.PlanningErrorException: incompatible data types in combination

Re. reporting in Jenkins: the way it works now is, if running queries against VoltDB produces an NPE, that element of the HTML "SQL Coverage Test Summary" is red, and the build fails, both in the sense that it says "BUILD FAILED" on the command line, and Jenkins is red. (It used to be orange, and would not fail.) If running queries against the comparison database (HSQL or PostgreSQL) produces an NPE, that element of the HTML "SQL Coverage Test Summary" is orange (used to be yellow), and this fact in itself does not cause the build to fail - however, it probably will fail anyway, since either the result matches, which implies an error in VoltDB (probably NPE, which would fail), or it does not match, which implies a failure do to result mismatch. It is conceivable that an NPE in the comparison DB would not fail, when paired with a non-NPE error in VoltDB, but this seems very unlikely in practice.

johnTcrawford:
Ticket filed: ENG-9572: Modify SQLCoverage tests to fail upon encountering ArrayOutOfBoundsException (& others?)

PR #3059, opened 2015-12-03
points.in, points.sql, points.outbaseline: drop procedure make_point and

johnTcrawford:
table points at the end (to clean up); also, drop both "if exists", at
the beginning (this was already done for table points, so just added
procedure make_point).

### DIFF
--- a/tests/sqlcmd/scripts/geospatial/points.sql
+++ b/tests/sqlcmd/scripts/geospatial/points.sql
@@ -1,3 +1,4 @@
+drop procedure make_point if exists;
 drop table points if exists;
 
 create table points (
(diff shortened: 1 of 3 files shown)


tests/sqlcmd/scripts/geospatial/points.sql
```diff
@@ -1,3 +1,4 @@
+drop procedure make_point if exists;
```
hhakimi53: Minor nit: Do we still want to drop procedure and table as it's getting cleaned up at the at the end of test in points.in?

hhakimi53:
Overall changes look fine to me.

tests/sqlcmd/scripts/geospatial/points.sql
johnTcrawford: Yeah, as we discussed, this only helps if another test had a procedure with the same name and failed to drop it, so it's not essential; but it does no harm.

PR #3060, opened 2015-12-04
Use the PCRE2 library for regular expressions.

BillWhite:
Use the PCRE2 library for regular expressions.

We keep the library as a tarball, and expand it as needed.


BillWhite:
I forgot a push.  I will push the real code and then reopen the pull request.

PR #3062, opened 2015-12-04
Replace Boost regular expressions with PCRE2 regular expressions.

BillWhite:
We replaced dependence on the Boost regular expression
library with dependence on the PCRE2 library.  The latter is much
smaller and easier to build than Boost.  It is alleged that Boost regex
can be built without building all of Boost via bjam and b2, but we have
not investigated this.

The PCRE2 sources are added as one single tarball, to reduce our github
repository size.

The regular expression matching function, regexp_position, was reworked
to use the PCRE2 library.  Some C++ unit tests were added as well.

The bulk of the changes are the deletion of the boost regular expression include
files, which are mostly at the end of the pull request's file list.  The important changes
are in the python build scripts and in the file third_party/cpp/expressions/stringfunctions.hpp.
The unit test tests/ee/expressions/function_tests.cpp also has some amusing changes.

https://issues.voltdb.com/browse/ENG-9556


src/ee/expressions/stringfunctions.h
```diff
@@ -677,40 +688,57 @@ template<> inline NValue NValue::call<FUNC_VOLT_REGEXP_POSITION>(const std::vect
...
+        throw SQLException(SQLException::data_exception_invalid_parameter, emsg.c_str());
+    }
+    boost::shared_ptr<pcre2_match_data> match_data(pcre2_match_data_create_from_pattern(pattern.get(), NULL),
+                                                   pcre2_match_data_free);
+    if (match_data.get() == NULL) {
+        throw SQLException(SQLException::data_exception_invalid_parameter, "Internal error: Cannot create PCRE2 match data.");
+    }
+    unsigned int matchFlags = 0;
+    error_code = pcre2_match(pattern.get(),
+                      reinterpret_cast<const unsigned char *>(sourceChars),
```
hhakimi53: Is cast of the subject string to unsigned char necessary?

buildtools.py
```diff
@@ -314,16 +339,16 @@ def buildMakefile(CTX):
         makefile.write("########################################################################\n")
         makefile.write("#\n# %s\n#\n" % filename)
         makefile.write("########################################################################\n")
-        makefile.write("%s: %s \n" % (jni_objname, filename))
+        makefile.write("%s: %s | build-third-party-tools \n" % (jni_objname, filename))
```
wolffcm: What does the `|` character do in Makefiles?

tests/ee/expressions/function_test.cpp
```diff
@@ -103,6 +103,13 @@ struct FunctionTest : public Test {
         template <typename LEFT_INPUT_TYPE, typename RIGHT_INPUT_TYPE, typename OUTPUT_TYPE>
         int testBinary(int operation, LEFT_INPUT_TYPE left_input, RIGHT_INPUT_TYPE right_input, OUTPUT_TYPE output, bool expect_null = false);
 
+        /**
+         * A template for calling unary functions.  This follows the pattern
```
hhakimi53: I think you meant "A template for calling ternary functions."

src/ee/expressions/stringfunctions.h
```diff
@@ -677,40 +688,57 @@ template<> inline NValue NValue::call<FUNC_VOLT_REGEXP_POSITION>(const std::vect
                     case 'c':
                         break;
                     case 'i':
-                        syntaxOpts |= boost::regex_constants::icase;
+                        syntaxOpts |= PCRE2_CASELESS;
```
wolffcm: This was already here, but Huzefa had asked the question, "what if there are multiple flags?"  I think the intention was for the last flag to take effect.  I think we ought to add a line for `case 'c'` to force a case-sensitive match explicitly, in case the user passes a weird options string (e.g., "cciicc").

src/ee/expressions/stringfunctions.h
```diff
@@ -677,40 +688,57 @@ template<> inline NValue NValue::call<FUNC_VOLT_REGEXP_POSITION>(const std::vect
...
-        } else {
+    const char* sourceChars = reinterpret_cast<const char*>(source.getObjectValue_withoutNull());
+    unsigned int lenSource = source.getObjectLength_withoutNull();
+    const unsigned char* patChars = reinterpret_cast<const unsigned char*>(pat.getObjectValue_withoutNull());
+    unsigned int lenPat = pat.getObjectLength_withoutNull();
+    // Compile the pattern.
+
+    int error_code = 0;
+    PCRE2_SIZE error_offset = 0;
+    boost::shared_ptr<pcre2_code> pattern ( pcre2_compile(patChars,
```
wolffcm: Just thinking aloud here: I wonder if there's a way to not have to do this pattern compilation for every row in a table when this is actually called in a SELECT statement.  This is fine for now, though.

wolffcm: Any reason to use `shared_ptr` here and below instead of `scoped_ptr` which is a little more efficient (and perhaps more informative to developers reading the code)?

wolffcm: I mean that it would be more informative if the pattern is not actually shared.  But now I'm looking at `pcre2_code_free` below and thinking maybe I'm not understanding everything.

tests/ee/expressions/function_test.cpp
```diff
@@ -391,6 +461,57 @@ TEST_F(FunctionTest, RepeatTooBig) {
...
+    ASSERT_EQ(testBinary(FUNC_VOLT_REGEXP_POSITION, testString, std::string("[a-z](\\d+)[a-z]"), 0), 0);
+    ASSERT_EQ(testBinary(FUNC_VOLT_REGEXP_POSITION, testString, std::string("[a-z](\\d+)[A-Z]"), 20), 0);
+    ASSERT_EQ(testTernary(FUNC_VOLT_REGEXP_POSITION, testString, std::string("[a-z](\\d+)[a-z]"), "i", 20), 0);
+
+    // Test an illegal pattern.
+    sawexception = false;
+    try {
+        ASSERT_EQ(testBinary(FUNC_VOLT_REGEXP_POSITION, testString, std::string("[a-z](a]"), 0), 0);
+    } catch (voltdb::SQLException &ex) {
+        sawexception = true;
```
hhakimi53: Minor: It may be good to verify the generated exception is what we are looking for. For this test may be something like:

EXPECT_TRUE(ex.message().find("Regular Expression Compilation Error") != string::npos);

buildtools.py
BillWhite: That's an order only dependence.  It says build the third party tools before bulding this target, but only build them once.  Without this notation the third party tools will be built once for each occurrence of the rule, which is to say, once for each source file and test source file.

I had to search a long time to find that out.

tests/ee/expressions/function_test.cpp
BillWhite: Ah yes, right.  I'll fix it.

wolffcm:
We should definitely run the EE unit test and the relevant JUnit in memcheck mode if you haven't already.

src/ee/expressions/stringfunctions.h
```diff
@@ -677,40 +688,57 @@ template<> inline NValue NValue::call<FUNC_VOLT_REGEXP_POSITION>(const std::vect
                     case 'c':
                         break;
                     case 'i':
-                        syntaxOpts |= boost::regex_constants::icase;
+                        syntaxOpts |= PCRE2_CASELESS;
                         break;
                     default:
-                        throw SQLException(SQLException::data_exception_invalid_parameter, "illegal match flags");
+                        throw SQLException(SQLException::data_exception_invalid_parameter, "Regular Expression Compilation Error: Illegal Match Flags");
```
hhakimi53: The u[pdated message is more descriptive and helpful thn earlier one.

src/ee/expressions/stringfunctions.h
BillWhite: Well, in the absence of a specification, we don't really know what the 'c' character means.  I suspect they are trying to match postgres, which is implemented with Spencer regular expressions.  In this case, the 'c' does mean case sensitivity, and the 'i' means case insensitivity.  But the code didn't do that, and I was not sure I wanted to change it.

I'll change it to make 'c' be case sensitivity, 'i' means case insensitivity  and the last one wins.

tests/ee/expressions/function_test.cpp
```diff
@@ -391,6 +461,57 @@ TEST_F(FunctionTest, RepeatTooBig) {
...
+    // Test a null pattern.
+    ASSERT_EQ(testBinary(FUNC_VOLT_REGEXP_POSITION, testString, NULL_STRING, 0, true), 1);
+
+    // Test a null subject.
+    ASSERT_EQ(testBinary(FUNC_VOLT_REGEXP_POSITION, NULL_STRING, "TEST", 0, true), 1);
+#endif /* 0 */
+
+     // Test utf-8 strings.
+    ASSERT_EQ(testBinary(FUNC_VOLT_REGEXP_POSITION, testUTF8String, "[A-Z]贾", 9), 0);
+    ASSERT_EQ(testTernary(FUNC_VOLT_REGEXP_POSITION, testUTF8String, "[a-z]贾", "i", 9), 0);
```
wolffcm: Glad to see tests using UTF-8 chars here.  Maybe add a negative test that uses another character in the pattern, like "家" or something, to show that it doesn't match?

BillWhite:
Ok.  I haven't yet, but I will now.

wolffcm:
That's all my comments.  Overall, this looks very good.

src/ee/expressions/stringfunctions.h
hhakimi53: Yes, Huwaei had responded the ending/last flag character will win. For example "cic" should be computed  as case-sensitive search and 'cici' should be treated as case-insensitive search. So in that case for flag. Huwaei's logic also had bug there; good catch.

That's how PostGRESQL regex function also follows similar pattern in regex computation. So, I think  we can follow the same approach here.

postgres=# select regexp_matches('foobarbequebazilbarfbonk', 'B', 'cicc');
 regexp_matches 
----------------
(0 rows)

postgres=# select regexp_matches('foobarbequebazilbarfbonk', 'B', 'cici');
 regexp_matches 
----------------
 {b}
(1 row)

hhakimi53:
Overall it looks good.

src/ee/expressions/stringfunctions.h
BillWhite: My first thought was to use unique_ptr, which is what I really want.  But that's not available to use.  My second thought was to use boost::scoped_ptr.  But that's too simple.  It doesn't allow for a custom deleter.  So, I fell back to boost::shared_ptr.  I don't think it's all that expensive to increment and decrement a reference counter, compared to doing a regular expression compilation and match.  So, I used it.

There was some loose talk about using boost::checked_delete with boost::scoped_ptr, but I didn't understand it well enough to know if it was real or not.

BillWhite:
When you say run the EE memcheck tests, do you mean to run the memcheck-lite Jenkins test?  That's run automatically by Jenkins, but that's probably not what you had in mind.

Is ther

BillWhite:
Sorry, I really didn't want to close this pull request.  It was only accidental.

src/ee/expressions/stringfunctions.h
wolffcm: I'm confused.  I guess PCRE will use pcre2_code_free to de-allocate allocated memory.  But if it's doing the freeing, then why do we need to use shared_ptr?

src/ee/expressions/stringfunctions.h
BillWhite: No. I don't think so.  I'll remove it, and make the variable unsigned char *.

src/ee/expressions/stringfunctions.h
BillWhite: RAII.

PCRE2 won't call pcre2_code_free to release the memory unless we call it.  I could call it explicitly at the end of the function, and hope that no exceptions are thrown, and that I get it right, and that everybody else who ever works on this gets it right.  Since PCRE2 is basically a C library the first is not unlikely.  But the second is a week reed.

I'll try to explain why this is here.

tests/ee/expressions/function_test.cpp
BillWhite: We test this in the jUnit test.  I didn't think it added anything to test it here.

tests/ee/expressions/function_test.cpp
BillWhite: Done.

src/ee/expressions/stringfunctions.h
BillWhite: I've added a C++ unit test and a jUnit test for this.

tests/ee/expressions/function_test.cpp
hhakimi53: Yes. I was thinking in terms of EE being the generator of these exception and error messages, so could had validated error message here. Most of the cpp tests are modeled on junit test, so that would have provided the similar coverage in EE. It's a minor and I agree with It should be fine not testing it again here.

hhakimi53:
Changes look good. Thanks Bill.

PR #3066, opened 2015-12-04
ENG-9568, fix A-A replication deadlock when setting DR=ACTIVE on a ru…

luyangco:
…nning cluster.

Change-Id: I3bc1b7cd4d955f2b1b9f0bccfd2cb9a70bda9814

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -2245,6 +2245,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback {
                 for (Initiator iv2init : m_iv2Initiators.values()) {
                     iv2init.setConsumerDRGateway(m_consumerDRGateway);
                 }
+                m_consumerDRGateway.initialize(false);
             }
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2709,8 +2710,9 @@ private void prepareReplication() {
     }
 
     private boolean createDRConsumerIfNeeded() {
-        if (!m_config.m_isEnterprise ||
-                !(m_consumerDRGateway instanceof ConsumerDRGateway.DummyConsumerDRGateway)) {
+        if (!m_config.m_isEnterprise
+                || !(m_consumerDRGateway instanceof ConsumerDRGateway.DummyConsumerDRGateway)
+                || !m_catalogContext.cluster.getDrconsumerenabled()) {
```
afrogers: So if I understand this right, we're going to delay creating the consumer until when enable it. The `initialize()` call in `prepareReplication()` will be harmlessly called on the `DummyConsumerDRGateway` if it's not already enabled.

This seems to me like it should work; I just want to confirm that I'm fully following your intent.

PR #3070, opened 2015-12-07
Swap latitude and longitude in geo points.

BillWhite:
Many vendors list points as (longitude, latitude).  This matches
our normal (x, y) intuition.  This is not universal, but it seems
to be more common than (latitude, longitude).  So, we have switched
these.

https://issues.voltdb.com/browse/ENG-9493

### DIFF
--- a/src/ee/common/Point.hpp
+++ b/src/ee/common/Point.hpp
@@ -44,7 +44,7 @@ public:
     {
     }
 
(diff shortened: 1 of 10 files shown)


BillWhite:
This pull request is premature.  I need to merge with the Geo Name Changes, which will be harder than I expected.  I'm closing this.

PR #3074, opened 2015-12-08
Update Planner Tester baselines for recent planner change

wolffcm:
I verified that the "explain plan" output for all queries is
identical to what we got before.  The only change here must be that
there are fewer inlined project nodes (which was confusing column
offset calculation in rare cases), which changed how nodes were
numbered.  Explain plans don't show the project nodes, so it's clear
that there is no functional change with these queries.


hhakimi53:
I was looking at the changes and read through couple of the plans and they appeared to be the same. Though looking at it careful seems like the inline node ids for the plans. Though in diff it looks the complete plan have been replaced. Is there a way or switch that we can use to filter out to see what has changed?

I do not understand the query plans much though. As you have already verified the explain plans for all queries is same as before and if Jenkins passes, I am good with the changes.

tests/frontend/org/voltdb/planner/plannerTester.java
```diff
@@ -540,7 +540,16 @@ else if (m_treeSizeDiff > 0) {
                 }
 
                 if (m_reportDiffExplainedPlan) {
-                    m_reportWriter.write("\nExplained plan:\n" + pn1.toExplainPlanString() + "\n==>\n" + pn2.toExplainPlanString() + "\n");
+                    m_reportWriter.write("\nExplained plan:");
+                    String pn1Plan = pn1.toExplainPlanString();
+                    String pn2Plan = pn2.toExplainPlanString();
+                    if (pn1Plan.compareTo(pn2Plan) == 0) {
+                        m_reportWriter.write("  [contents of explained plans are identical]\n"
```
hhakimi53: This will be very helpful in identifying the reported failure.

PR #3080, opened 2015-12-09
updates for NOT contains()

hhakimi53:
Updated condition expression evaluation logic to take into Function expression can also return boolean when evaluating NOT operator expression.

- Tests in jenkins are currently underway. Community Junits have completed and is blue

### DIFF
--- a/src/frontend/org/voltdb/expressions/ExpressionUtil.java
+++ b/src/frontend/org/voltdb/expressions/ExpressionUtil.java
@@ -615,7 +615,11 @@ public abstract class ExpressionUtil {
             }
         } else if (ExpressionType.OPERATOR_NOT == expr.getExpressionType()) {
             AbstractExpression leftExpr = expr.getLeft();
-            if (VoltType.BOOLEAN == leftExpr.getValueType()) {
+            // functions can also return boolean. so the left child expression can be
+            // function expression and there by don't evaluate every left child expr
+            // as constant value expression
+            if ((VoltType.BOOLEAN == leftExpr.getValueType()) &&
+                    (leftExpr instanceof ConstantValueExpression)) {
                 if (ConstantValueExpression.isBooleanTrue(expr.getLeft())) {
                     return ConstantValueExpression.getFalse();
                 } else {
--- a/src/frontend/org/voltdb/expressions/ExpressionUtil.java
+++ b/src/frontend/org/voltdb/expressions/ExpressionUtil.java
@@ -615,15 +615,12 @@ public abstract class ExpressionUtil {
             }
         } else if (ExpressionType.OPERATOR_NOT == expr.getExpressionType()) {
             AbstractExpression leftExpr = expr.getLeft();
-            if (VoltType.BOOLEAN == leftExpr.getValueType()) {
-                // functions can also return boolean. so the left child expression can be
-                // function expression and there by don't evaluate every left child expr
-                // as constant value expression
-                if (leftExpr instanceof FunctionExpression) {
-                    return expr;
-                }
-
-                if (ConstantValueExpression.isBooleanTrue(expr.getLeft())) {
+            // function expressions can also return boolean. So the left child expression
+            // can be expression which are not constant value expressions, so don't
+            // evaluate every left child expr as constant value expression
+            if ((VoltType.BOOLEAN == leftExpr.getValueType()) &&
+                    (leftExpr instanceof ConstantValueExpression)) {
+                if (ConstantValueExpression.isBooleanTrue(leftExpr)) {
                     return ConstantValueExpression.getFalse();
                 } else {
                     return ConstantValueExpression.getTrue();
--- a/src/frontend/org/voltdb/expressions/ExpressionUtil.java
+++ b/src/frontend/org/voltdb/expressions/ExpressionUtil.java
@@ -615,11 +615,15 @@ public abstract class ExpressionUtil {
             }
         } else if (ExpressionType.OPERATOR_NOT == expr.getExpressionType()) {
             AbstractExpression leftExpr = expr.getLeft();
-            // functions can also return boolean. so the left child expression can be
-            // function expression and there by don't evaluate every left child expr
-            // as constant value expression
-            if ((VoltType.BOOLEAN == leftExpr.getValueType()) &&
-                    (leftExpr instanceof ConstantValueExpression)) {
+
+            if (VoltType.BOOLEAN == leftExpr.getValueType()) {
+                // functions can also return boolean. so the left child expression can be
+                // function expression and there by don't evaluate every left child expr
+                // as constant value expression
+                if (leftExpr instanceof FunctionExpression) {
+                    return expr;
+                }
+
                 if (ConstantValueExpression.isBooleanTrue(expr.getLeft())) {
                     return ConstantValueExpression.getFalse();
                 } else {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -616,6 +616,13 @@ public static AbstractExpression evaluateExpression(AbstractExpression expr) {
...
             AbstractExpression leftExpr = expr.getLeft();
             if (VoltType.BOOLEAN == leftExpr.getValueType()) {
+                // functions can also return boolean. so the left child expression can be
+                // function expression and there by don't evaluate every left child expr
+                // as constant value expression
+                if (leftExpr instanceof FunctionExpression) {
+                    return expr;
+                }
+
                 if (ConstantValueExpression.isBooleanTrue(expr.getLeft())) {
```
BillWhite: Are there any other kinds of expressions that can occur here?

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -616,6 +616,13 @@ public static AbstractExpression evaluateExpression(AbstractExpression expr) {
...
                     return ConstantValueExpression.getFalse();
                 } else {
```
BillWhite: What if leftExpr isn't a ConstantValueExpression at all?  Does isBooleanTrue return false?  One would think so.  But then we return false always, even if leftExpr is something else.

Shouldn't we be testing for more cases here?

BillWhite:
That's all I've got.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
hhakimi53: Any expression whose return type is bool can get in here. At present I am only aware of function expression that return bool, does not work properly in conjunction with not operator. SQL function returning bool was recently added. So the change is specific to function expression when used with not operator. I

src/frontend/org/voltdb/expressions/ExpressionUtil.java
wolffcm: I agree with Bill on this.  The if statement at line 626 (in the new revision) seems obviously incorrect.  Anything that produces a boolean value that is not a function will fall into the same trap.  There might not be anything else like that now, but someday there might be, so it will pay to be proactive here.

The if statement at line 626 should be something more like `if (leftExpr instanceof ConstantValueExpression && ConstantValueExpression.isBooleanTrue(leftExpr)) { ... }`, which should cover the function case as well as anything else that comes along.

hhakimi53: Originally I had taken approach on same line: https://github.com/VoltDB/voltdb/commit/57a293e11248008049fd50dac2353557c0874f7e though later had backed it out with thought that at present lets limit this change to function expression to avoid any regression. I can try it out tomorrow and see how it goes through jenkins. Check has to be made at level above where the check for the expressions that returns boolean is made and let the expression only of ConstantValueExpression to fall through (currently how updated logic is looking for functionexpression) and for any thing else it will have to go through current checks.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -615,8 +615,12 @@ public static AbstractExpression evaluateExpression(AbstractExpression expr) {
...
             AbstractExpression leftExpr = expr.getLeft();
-            if (VoltType.BOOLEAN == leftExpr.getValueType()) {
-                if (ConstantValueExpression.isBooleanTrue(expr.getLeft())) {
+            // function expressions can also return boolean. So the left child expression
+            // can be expression which are not constant value expressions, so don't
+            // evaluate every left child expr as constant value expression
+            if ((VoltType.BOOLEAN == leftExpr.getValueType()) &&
+                    (leftExpr instanceof ConstantValueExpression)) {
+                if (ConstantValueExpression.isBooleanTrue(leftExpr)) {
                     return ConstantValueExpression.getFalse();
```
BillWhite: I can't seem to comment on the line I want to comment, but if this is NOT NOT E, don't you want to recurse into E before you return it?  I mean NOT NOT NOT FALSE should be TRUE.  I think this code would make it NOT FALSE.

Also, if expr.getExpressionType() == OPERATOR_NOT, shouldn't the leftExpr.getValueType() always be BOOLEAN?  Otherwise HSQL should give us a failure, shouldn't it?

Do we have to check for the constant NULL here?  What if the expression is NOT NULL?

BillWhite:
This seems almost completely right to me.  I may misunderstand the code, in which case it's entirely right.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
hhakimi53: Interesting thoughts :). 

1. Which line number in current code are you referring to and is it in this file itself? Is it in isNullRejectingExpression()? 

I tried running query with two or more not operators in there and it's not currently flags as not supported with our without the changes: Query: select places.name, borders.name from places, borders where not not contains(borders.region, places.loc) order by places.pk, borders.pk"


Fails to compile the query 
org.voltdb.client.ProcCallException: Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: Error compiling query: org.voltdb.planner.PlanningErrorException: SQL Syntax error in "select places.name, borders.name from places, borders where not not contains(borders.region, places.loc) order by places.pk, borders.pk" unexpected token: NOT
	at org.voltdb.client.ClientImpl.callProcedure(ClientImpl.java:297)
...
	at org.voltdb.regressionsuites.TestGeospatialFunctions.testContains(TestGeospatialFunctions.java:210)


Theoretically, for the current function evaluateExpression(), evaluation path should work fine for "NOT NOT exp" and "NOT NOT NOT exp" case:
For example Not be applied to function expression that returns a boolean
a. expression: Not - left child <function expression>  => return Not <function expression>
b. expresion NOT - leftchild <NOT> -left child <function expression> => return left child -> left child => function expression

2. for any expression other than boolean return type is handled and flagged by HSQL.

3. I don't think NULL constants are supported in current HSQL we are using and there may be a ticket related to that.

4. "is not NULL" and "is NULL" works fine, in case you were thinking about it. I will add unit test for "is not null" also.

wolffcm: It seems like NOT could be left-associative in HSQL, so `NOT NOT <expr>` is interpreted as `((NOT NOT) <expr>)`?  If it was treated as right-associative it should compile fine.  I wonder what PostgreSQL does?

hhakimi53: Could be, I  can try to check on that one or do we want to create separate ticket for it?

In PostGreS "NOT NOT" works fine. 
postgres=# select * from foo where NOT NOT (id > 1); 
 id | value 
----+-------
  2 |    11
  4 |    51
(2 rows)

postgres=# select * from foo where NOT (id > 1); 
 id | value 
----+-------
  1 |    10
(1 row)

postgres=# select * from foo;
 id | value 
----+-------
  1 |    10
  2 |    11
  4 |    51
(3 rows)
-----------------------------

wolffcm: Just a separate ticket would be fine.  "NOT NOT" just produces a syntax error, not a wrong answer.  Except for auto-generated SQL, customers are unlikely to ever hit this, and there's an obvious workaround.

hhakimi53: I agree; I will create a ticket for the issue. Thanks

PR #3081, opened 2015-12-10
Schema of the null_tuple for outer joins should match inner_tuple

wolffcm:
There's three issues here that are solved by these commits:
- ENG-9389 (EE assertion on query with outer join and is null predicate)
- ENG-9533 (erroneous plan produced for outer join query with an OR predicate)
- ENG-9573 (Q19 runs out of temp table space in the aggbench perf tests)

I believe that the fix for ENG-9389 caused ENG-9533, and the fix for ENG-9533 caused ENG-9573.  The change here is just to manually rollback the fixes for the first two tickets, and add a different fix for ENG-9389 that solves the issue in a different way.

The issue in ENG-9389 is related to the nested loop index executor in the EE.  The null_tuple (an immutable tuple used to produce the null-padded results for unmatching rows in a left outer join) didn't have enough columns and the plan was looking for columns that weren't there.  My fix for this is to give the null_tuple the same schema as the inner_tuple (the tuple that is used to produce output if there is a match).

I've verified that these changes fix all three of these issues.

### DIFF
--- a/src/ee/executors/nestloopindexexecutor.cpp
+++ b/src/ee/executors/nestloopindexexecutor.cpp
@@ -245,6 +245,28 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
     ProgressMonitorProxy pmp(m_engine, this);
 
     TableTuple join_tuple;
+    // It's not immediately obvious here, so there's some subtlety to
+    // note with respect to the schema of the join_tuple.
+    //
+    // The inner_tuple is used to represent the values from the inner
+    // table in the case of the join predicate passing, and for left
+    // outer joins, the null_tuple is used if there is no match.  Both
+    // of these tuples include the complete schema of the table being
+    // scanned.  The inner table is being scanned via an inlined scan
+    // node, so there is no temp table corresponding to it.
+    //
+    // Predicates that are evaluated against the inner table should
+    // therefore use the complete schema of the table being scanned.
+    //
+    // The join_tuple is the tuple that contains the values that we
+    // actually want to put in the output of the join (or to aggregate
+    // if there is an inlined agg plan node).  This tuple needs to
+    // omit the unused columns from the inner table.  The inlined
+    // index scan itself has an inlined project node that defines the
+    // columns that should be output by the join, and omits those that
+    // are not needed.  So the join_tuple contains the columns we're
+    // using from the outer table, followed by the "projected" schema
+    // for the inlined scan of the inner table.
     if (m_aggExec != NULL) {
         VOLT_TRACE("Init inline aggregate...");
         const TupleSchema * aggInputSchema = node->getTupleSchemaPreAgg();
--- a/src/ee/executors/nestloopindexexecutor.cpp
+++ b/src/ee/executors/nestloopindexexecutor.cpp
@@ -166,9 +166,7 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
 
     // NULL tuple for outer join
     if (node->getJoinType() == JOIN_TYPE_LEFT) {
-        Table* inner_out_table = m_indexNode->getOutputTable();
-        assert(inner_out_table);
-        m_null_tuple.init(inner_out_table->schema());
+        m_null_tuple.init(inner_table->schema());
     }
 
     //outer_table is the input table that have tuples to be iterated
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/plannodes/NestLoopIndexPlanNode.java
```diff
@@ -109,12 +109,9 @@ public void resolveColumnIndexes()
         resolvePredicate(inline_scan.getSkipNullPredicate(), outer_schema, index_schema);
         resolvePredicate(inline_scan.getSearchKeyExpressions(), outer_schema, index_schema);
 
-        final NodeSchema inline_scan_output_schema = inline_scan.getOutputSchema();
-
-        // resolve other predicates based on inline index scan node output schema
-        resolvePredicate(m_preJoinPredicate, outer_schema, inline_scan_output_schema);
-        resolvePredicate(m_joinPredicate, outer_schema, inline_scan_output_schema);
-        resolvePredicate(m_wherePredicate, outer_schema, inline_scan_output_schema);
+        resolvePredicate(m_preJoinPredicate, outer_schema, index_schema);
```
jxchinaren: Can you add a comment here saying something like: ENG-9389 inline plan node does not have output temp table and we user its target table schema here on purpose.

src/ee/executors/nestloopindexexecutor.cpp
```diff
@@ -523,7 +520,19 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
...
+                //
+                // Try to put the tuple into our output table
+                // Append the inner values to the end of our join tuple
+                //
+                for (int col_ctr = num_of_outer_cols;
+                     col_ctr < join_tuple.sizeInValues();
+                     ++col_ctr) {
+                    // For the sake of consistency, we don't try to do
+                    // output expressions here with columns from both tables.
+                    join_tuple.setNValue(col_ctr,
```
jxchinaren: Can you explain the join_tuple, outer_tuple, inner_tuple here?

I may miss something and have a question:

Looks like outer_tuple is from the child table of NestLoopIndexPlanNode using its output table schema. Inner_tuple is using the inline index's table schema. john_tuple is the output table of NestLoopIndexPlanNode. 

for example:
select t1.a + t2.b from t1 left outer join t2 on t1.c=t2.c where t1.b > 3;
assume there is a index on t1 (b), and another index on t2(c);

How does the join_tuple here make sense?

jxchinaren:
That's all my comments right now.

src/frontend/org/voltdb/plannodes/NestLoopIndexPlanNode.java
wolffcm: Yes, I'll add the comment, and also add a comment to IndexScanPlanNode.

src/ee/executors/nestloopindexexecutor.cpp
wolffcm: The join_tuple gets its schema from the join node field "tuple schema pre-agg".  In the planner, in `AbstractJoinPlanNode.generateOutputSchema`, this is defined to be a concatenation of the outer table's output and the inner table's output table (with the projection applied, so that we remove unused columns).

However, any access of the fields in the inner table (such as to evaluate predicates, or compute the output of the join) must use the complete schema for the being scanned.  It's the job of the `m_outputExpressions` to do the correct translation from the complete schema of the table being scanned to the output of the join.

In your example, I think we would see the addition on the select list appear in `m_outputExpressions`, and the TVEs in the add expression should use column offsets consistent with t2's complete schema, even though the output of the join should omit other columns in t2 (e.g., t2.a).

Does that make sense?

I'll add your example to the JUnit test, and verify that it's doing the right thing.

wolffcm: I should add that the schema of inner_tuple in the EE join code uses the complete schema of the inner table being scanned.

jxchinaren: EE join_tuple contains two parts: outer_tuple + inner_tuple.

Your second comment is what I want to stress here: 
inner_tuple in the EE join codes use schema of inner table. 

But in planner NestLoopIndexPlanNode::generateOutputSchema, inner_tuple use inner table's output schema. 

Should Java and EE be equal on the join_tuple's schema?

wolffcm: Inner tuple in the EE does use the complete schema of the table being scanned.  But inner columns in join_tuple actually use the projected schema from the inlined scan.  So the join_tuple schema is *not* the same as the schema of outer_tuple + inner_tuple.

Moving data from the complete schema of the inner table to the join_tuple is done by using the m_outputExpressions, which takes care of the relative difference in column offset.

This seems to match what the planner does in `NestLoopIndexPlanNode::generateOutputSchema`:
```Java
 m_outputSchemaPreInlineAgg =
            m_children.get(0).getOutputSchema().
            join(m_children.get(1).getOutputSchema()).copyAndReplaceWithTVE();
```
The "output schema" in this case includes the scan's inlined projection, which is what we want to keep the temp tables small.  So think that Java and the EE agree on the schema of join_tuple.

wolffcm:
@jxchinaren I just pushed another commit that adds comments and the test you suggested.

jxchinaren:
Looks good to me, thanks for adding those comments.

wolffcm:
Thanks for the review Xin.  I'll squash and merge on a separate branch.

PR #3086, opened 2015-12-11
Schema of the null_tuple for outer joins should match inner_tuple

wolffcm:
Also rolled back fixes for ENG-9389 and ENG-9533, which were
problematic; I left the tests though.


wolffcm:
This was reviewed when the same fix was merged to release-5.8.x.

PR #3089, opened 2015-12-14
Schedule BidGenerator runnable to run during demo

wolffcm:


examples/geospatial/client/geospatial/AdBrokerBenchmark.java
```diff
@@ -337,17 +360,9 @@ public void runBenchmark() throws Exception {
...
         }
 
-        // cancel periodic stats printing
-        timer.cancel();
-
-        // block until all outstanding txns return
-        client.drain();
-
-        // print the summary results
         printResults();
```
hhakimi53: Can printResults() without draining all the I/Os result not taking into account the IOs that are still in flight?

hhakimi53:
Changes look good overall.

examples/geospatial/client/geospatial/AdBrokerBenchmark.java
wolffcm: Yes, you're right.  I'll make sure that this is fixed in the next ticket, which adds the dynamic device behavior.

PR #3092, opened 2015-12-16
Eng 9304 geo astext

hhakimi53:
Add support for geo function asText() - converts geo types to wkt. WKT output presents longitude and latitude to decimal numbers to significant 12 digits.

Testing:
- Currently running Jenkins job


src/ee/common/GeographyValue.hpp
```diff
@@ -348,6 +349,44 @@ inline std::string GeographyValue::toString() const {
     return oss.str();
 }
 
+inline std::string GeographyValue::toWKT() const {
+    std::ostringstream oss;
+
+    if (isNull()) {
+        // polygon with no rings/loops
+        oss << "Polygon()";
```
wolffcm: I think we ought to fail an assertion here or throw an exception or something.  We wouldn't ever want to return this to the user, since `asText(null)` should just return a null.

src/ee/common/GeographyPointValue.hpp
```diff
@@ -150,6 +150,45 @@ class GeographyPointValue {
...
 
+    std::string formatLngLat() const{
+        std::ostringstream oss;
+        oss << toString(m_longitude) << " " << toString(m_latitude);
+        return oss.str();
+    }
+
+    std::string toWKT() const {
+        std::ostringstream oss;
+        oss <<"Point (" << formatLngLat() << ")";
```
wolffcm: We should standardize our output with respect to uppercase and lowercase and whitespace.  I think I prefer all caps (the WKT Wikipedia page uses this syntax) with a space after the POINT or POLYGON keyword, but I don't have a strong preference as long as the WKT produced by Java and C++ are as close as possible.

src/ee/common/GeographyPointValue.hpp
```diff
@@ -150,6 +150,45 @@ class GeographyPointValue {
...
+        return oss.str();
+    }
+
+private:
+    // converts double value to string with specified precision displaying
+    // only significant decimal value.
+    // Output pattern is similar to "...##0.0##..."
+    std::string toString(double number, int decimalPrecision=12) const {
```
wolffcm: Will this ever be called with some other value for decimalPrecision?  If no, then I'd suggest removing this parameter.

src/ee/common/GeographyPointValue.hpp
```diff
@@ -150,6 +150,45 @@ class GeographyPointValue {
...
+
+        // provided precision should be less than that can survive roundtrip of double -> text -> double
+#if __cplusplus > 199711L
+        // if platform supports C++11
+        assert(decimalPrecision <= std::numeric_limits< double >::max_digits10);
+#else
+        assert(decimalPrecision <= 17);
+#endif
+
+        decimalNumber = number - floor(number);
```
wolffcm: Doesn't boost::format do a lot of this for us?

src/ee/common/GeographyValue.hpp
```diff
@@ -348,6 +349,44 @@ inline std::string GeographyValue::toString() const {
...
+    }
+    else {
+        Polygon poly;
+        poly.initFromGeography(*this);
+        int numLoops = poly.num_loops();
+        assert (numLoops > 0);
+        GeographyPointValue point;
+        oss << "Polygon (";
+        // capture all the loops
+        for (int i = 0; i < numLoops; ++i) {
```
wolffcm: S2 will reorder the loops such that the outer shell may not be first.  I think you'll need to check the depth value of each loop, and put the loop with depth 0 first.

Are there tests for polygons with interior loops?

tests/frontend/org/voltdb/fullddlfeatures/fullDDL.sql
```diff
@@ -1105,3 +1105,8 @@ CREATE INDEX ENG_8168_INDEX_USES_YEAR ON T55 (YEAR(TS) - 1900);
 -- Wrap up with a minor variant of an index already tried to show that
 -- the previous line didn't corrupt catalog replay.
 CREATE INDEX ENG_8168_INDEX_USES_ABS_AGAIN ON T55 (ABS(BIG - 1));
+-- geo functions
```
wolffcm: Thank you for doing this!

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -470,6 +470,77 @@ public void testPolygonPointDistance() throws Exception {
...
+                "select loc, asText(loc) from places order by pk").getResults()[0];
+
+        while (vt.advanceRow()) {
+            GeographyPointValue gpv = vt.getPoint(0);
+            if (gpv == null) {
+                assertEquals(null, vt.getString(1));
+            }
+            else {
+                String wkt = vt.getString(1);
+                assertEquals(gpv.toString().toLowerCase(), wkt.toLowerCase());
```
wolffcm: Please test without lower casing so that we can be sure that the WKT produced from Java and C++ is the same.

src/ee/common/GeographyValue.hpp
hhakimi53: Yes, we will not enter in this path ever how the logic flow happens. I will add assert in there.

src/ee/common/GeographyValue.hpp
wolffcm: I see that you have tests that should cover this.  But now I'm wondering why they're passing...

src/ee/common/GeographyPointValue.hpp
hhakimi53: Case for Point was a miss, I will update it so we have it in one format - Uppercase.

wolffcm:
That's all my comments.  Looks very nice overall.

src/ee/common/GeographyPointValue.hpp
hhakimi53: No, at present no one is using decimalPrecision. I will remove the parameter. Had done it with intent if we need to change it or resuse the function in future some place else but with different precision, the function will be flexible enough.

src/ee/common/GeographyValue.hpp
hhakimi53: Yes, we have junit for polygons which have loops/holes in them. 

With the changes for isValid() that Bill has, some of the logic in here may have to be redone. I will look into it.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: Sure, will have both frontend and EE produce result in upper-case.

src/ee/common/GeographyPointValue.hpp
hhakimi53: I will look into boost boost::format, I did not look at that one. It will be good if boost can deliver same functionality. 

I had looked at STL and boost stream setprecision() with fixed format and that required parsing output to truncate zeros from insignificant places. So went with above for simplicity and better performance.

hhakimi53:
Thanks for the review feedback.

src/ee/common/GeographyPointValue.hpp
hhakimi53: Looked into boost::format api's and there is equivalent version of printf() with some restrictions. With boost::format() also we will have to handle case whether number has any significant decimal digit or not  (like 3.0) so that we can represent co-ordinates format equivalent to Java's DecimalFormat(##0.0###########);. If we choose DecimalFormat(##0.############) as our representation, than we don't have to this logic in here and can simplify above logic. 

Do we want to switch our co-ordinate format to DecimalFormat(##0.############) from DecimalFormat(##0.0###########)?

src/ee/common/GeographyPointValue.hpp
```diff
@@ -150,6 +150,45 @@ class GeographyPointValue {
...
+        if (decimalNumber == 0) {
+            snprintf(buffer, sizeof(buffer), "%3.1f", number);
+        }
+        else {
+            int wholeNumberDigits = log10(abs(number)) + 1;
+            snprintf(buffer, sizeof(buffer), "%.*g", (wholeNumberDigits + decimalPrecision), number);
+        }
+        return buffer;
+    }
+
```
BillWhite: What's the difference between this and GeographyPointValue.toString?

BillWhite: Sorry.  I'm slow this morning.

src/ee/common/GeographyPointValue.hpp
hhakimi53: Tried couple of different ways with boost format and even with DecimalFormat(##0.############), could not find a way that could have forced boost:format only to decimal places excluding whole number part of it. There is a limitation that format() does not support '*' (http://www.boost.org/doc/libs/1_60_0/libs/format/doc/format.html#printf_differences). So on run time can't tune the precision of double.

Will use the current logic for outputting the format and will go with DecimalFormat(##0.0###########).

wolffcm: Ok, sounds good.

hhakimi53: I added couple of extra junit tests and sometimes back only found a corner case for which the above logic is unable to handle. 
If supplied longtitude, latitude are a whole number, for that due to round-trip of (longitude, latitude) -> (x,y,z) ->  (longitude, latitude), in EE longitude or latitude may be stored as a value less than 5 * 10 ^ -13 and for that case it gets rounded into whole number to certain precision. So in there only instead of presenting the longtitude/latitude as "<wholeNumber>.0" it's getting presented as "<wholeNumber>". So will have to patch up logic for that condition. Currently looking into it.

src/ee/expressions/functionexpression.h
```diff
@@ -256,6 +256,9 @@ static const int FUNC_VOLT_POLYGON_AREA                = 20039;
 static const int FUNC_VOLT_DISTANCE                    = 20040;     // wrapper id for distance between all geo types
 static const int FUNC_VOLT_DISTANCE_POINT_POINT        = 20041;     // distance between point and point
 static const int FUNC_VOLT_DISTANCE_POLYGON_POINT      = 20042;     // distance between polygon and point
+static const int FUNC_VOLT_ASTEXT                      = 20045;     // wrapper function id for converting geography types into wkt
+static const int FUNC_VOLT_ASTEXT_GEOGRAPHY_POINT      = 20046;     // asText(<point value>)
```
hhakimi53: Note: Function asText() has been renumbered starting from 20045 onwards to reserve 20043 and 20044 for polygon is valid and reason geo-functions. 

We will have to renumber the geo-functions again with merge from master gets integrated . I was thinking of renumbering geo-function numbers to start at 21000 with this change, as we will be doing so when merge from master gets integrated in geospatial branch, but it may cause unnecessary merge conflicts, so did not do it.

hhakimi53:
Jenkins run is underway

wolffcm:
I'll have a look after the innovation week show and tell.

On Fri, Dec 18, 2015 at 7:28 AM, hhakimi53 <notifications@github.com> wrote:

> Jenkins run is underway
>
> —
> <https://github.com/VoltDB/voltdb/pull/3092#issuecomment-165804048>.
>



-- 
Chris Wolff, Software Engineer
*VoltDB, Inc.*
*E:* cwolff@voltdb.com

*voltdb.com* <http://www.voltdb.com>
Follow VoltDB  <https://twitter.com/VoltDB>
<https://www.facebook.com/VoltDB>  <https://www.linkedin.com/company/voltdb>

src/ee/common/GeographyPointValue.hpp
```diff
@@ -150,7 +150,59 @@ class GeographyPointValue {
...
+        const int8_t decimalPrecision = 12;
+
+        // check if it's whole number
+        if (number == floor(number)) {
+            return true;
+        }
+
+        // check if rounded value is a whole number
+        double shiftNum = pow(10, decimalPrecision);
+        double roundedNumber = ceil((number * shiftNum) - 0.49 ) / shiftNum;
```
wolffcm: I don't quite understand why this works.  Is this a standard way of rounding to a certain number of decimal places?

wolffcm:
This looks good.  I had rounding question, but that's all.

wolffcm:
I'm in favor of merging this sooner rather than later, to free up John C.'s sqlcoverage work.

src/ee/common/GeographyPointValue.hpp
hhakimi53: Don't know if this is standard way of rounding the numbers but it will provide us a rounded number upto 12 decimal places with effect of 14 decimal places at least (http://www.cplusplus.com/forum/beginner/3600/#msg15518). I will increase the precision to 19 th decimal when computing the above rounding number. The rounded number obtained above and comparing it with it's floor value with provide us answer whether rounded number is whole number or not and based on it use whole number or decimal number printing format. When outputting the result for decimal number, we are using standard printf format rounding to obtain the rounded number.

hhakimi53:
Added few more junit test cases and running it through Jenkins. Will check in changes once it passes Jenkins.

PR #3093, opened 2015-12-16
Eng 9205 geo validate polygons

BillWhite:
Validate polygons.

This commit adds two different sets of functionality, and completes the work on two tickets.

1.  We add a boolean SQL function isValid(), which takes a geography value and tells if it is valid or not.
2. We add a varchar valued SQL function polygon_invalid_reason which takes a geography value and tells why it is invalid if it is, in fact, invalid.  If the geography is, in fact, valid, the function returns the string "Valid polygon".

In the EE, polygons are always made up of counter-clockwise loops.  Holes are loops which are contained in shells.  The SQL function isValid rejects multi-polygons and polygons whose area is more than half the earth.  It also rejects polygons whose rings cross or intersect in bad ways.

In Java, polygons may be constructed two ways. 

1. As OGC well known text. These polygons are given as a sequence of loops.  The first loop is the shell, and subsequent loops are holes.  The shell's vertices are counter-clockwise, and the holes' vertices are clockwise.  This is different from the EE representation, so we reverse the order of all loops which are holes.  Note that we do no validation on well-known-text, though the SQL validation function isvalid can be used to filter out valid or invalid polygons, and the SQL function polygon_invalid_reason can be used to tell why a polygon is invalid.
2. As a list of lists of loops.  In this case the first loop is a shell and subsequent loops are holes.  All loops in this form must be in counter-clockwise order.

Note that the transformation WKT -> EE -> WKT may not be the identity.  The geometry package S2 has its own ideas of the order of holes, and reorders loops to serve its own purposes.  We arrange for the unique shell to be first in the list of WKT.  If there is more than one hole, however, they may be in unpredictable order.


BillWhite:
This pull request is not that helpful.  I'm going to squash it and merge in the new 8859 branch before I continue with this.  Consider this pull request closed.

PR #3094, opened 2015-12-16
Eng 9205 geo validate polygons

BillWhite:
Validate polygons.

This commit adds two different sets of functionality, and completes the work on two tickets.

We add a boolean SQL function isValid(), which takes a geography value and tells if it is valid or not.
We add a varchar valued SQL function polygon_invalid_reason which takes a geography value and tells why it is invalid if it is, in fact, invalid. If the geography is, in fact, valid, the function returns the string "Valid polygon".

In the EE, polygons are always made up of counter-clockwise loops. Holes are loops which are contained in shells. The SQL function isValid rejects multi-polygons and polygons whose area is more than half the earth. It also rejects polygons whose rings cross or intersect in bad ways.

In Java, polygons may be constructed two ways.

As OGC well known text. These polygons are given as a sequence of loops. The first loop is the shell, and subsequent loops are holes. The shell's vertices are counter-clockwise, and the holes' vertices are clockwise. This is different from the EE representation, so we reverse the order of all loops which are holes. Note that we do no validation on well-known-text, though the SQL validation function isvalid can be used to filter out valid or invalid polygons, and the SQL function polygon_invalid_reason can be used to tell why a polygon is invalid.
As a list of lists of loops. In this case the first loop is a shell and subsequent loops are holes. All loops in this form must be in counter-clockwise order.
Note that the transformation WKT -> EE -> WKT may not be the identity. The geometry package S2 has its own ideas of the order of holes, and reorders loops to serve its own purposes. We arrange for the unique shell to be first in the list of WKT. If there is more than one hole, however, they may be in unpredictable order.


BillWhite:
FWIW, this passes all but the python tests in Jenkins.

wolffcm:
Interesting about the EE reordering inner rings.  Is the order of the rings at least deterministic?

BillWhite:
Yes, but it's expensive to compute, and we probably don't want to compute it in Java.  So, we can't trivially compute equality in Java on polygons with more than two holes if one has been sent into the database and back again.  We could try an N^2 algorithm, where N is the number of holes.  We might choose some other sort of order in Java on loops, say ordering by lower left corner.  That might be linear in the number of vertices.

In the EE we can always just call S2.

src/ee/expressions/geofunctions.cpp
```diff
@@ -369,4 +403,64 @@ template<> NValue NValue::call<FUNC_VOLT_DISTANCE_POINT_POINT>(const std::vector
...
+    for (int idx = 0; idx < nloops; idx += 1) {
+        S2Loop *loop = poly.loop(idx);
+        switch (loop->depth()) {
+        case 0:
+            nouters += 1;
+            break;
+        case 1:
+            break;
+        default:
+            VMLOG(2, msg) << "Polygons can only be shells or holes." << std::endl;
```
wolffcm: What is VMLOG?  Maybe add a comment describing it.

src/ee/expressions/geofunctions.cpp
```diff
@@ -369,4 +403,64 @@ template<> NValue NValue::call<FUNC_VOLT_DISTANCE_POINT_POINT>(const std::vector
...
+            return false;
+        }
+        if (!loop->IsNormalized(msg)) {
+            return false;
```
wolffcm: No VMLOG message for this?

src/ee/expressions/geofunctions.cpp
```diff
@@ -369,4 +403,64 @@ template<> NValue NValue::call<FUNC_VOLT_DISTANCE_POINT_POINT>(const std::vector
...
+        return NValue::getNullValue(VALUE_TYPE_BOOLEAN);
+    }
+    // Be optimistic.
+    bool returnval = true;
+    // Extract the polygon and check its validity.
+    Polygon poly;
+    poly.initFromGeography(getGeography());
+    if (!poly.IsValid()
+            || !isMultiPolygon(poly)) {
+        returnval = false;
```
wolffcm: I'm having trouble parsing this logic.  We return false if the polgon's isValid method returns false, or if the polygon is *not* a multipolygon?  Don't you want to get rid of the `!` before `isMultiPolygon`?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -420,6 +433,211 @@ public void testPolygonPointDistance() throws Exception {
...
+       new Border(105, "IntersectingHoles", "Loop 1 crosses loop 2",
+                  GeographyValue.fromText(INTERSECTING_HOLES)),
+       new Border(106, "OuterInnerIntersect", "Loop 1 crosses loop 2",
+                  GeographyValue.fromText(OUTER_INNER_INTERSECT)),
+       new Border(108, "TwoNestedSunwise", "Loop 0 encloses more than half the sphere",
+                  GeographyValue.fromText(TWO_NESTED_SUNWISE)),
+       new Border(109, "TwoNestedWiddershins", "Loop 0 encloses more than half the sphere",
+                  GeographyValue.fromText(TWO_NESTED_WIDDERSHINS)),
+      /*
+       * These are apparently legal. Should they be?
```
wolffcm: I don't have a strong opinion either way.  If they are legal, then the area function should return 0.

third_party/cpp/google-s2-geometry/geometry/include/s2geo/base/logging.h
```diff
@@ -52,13 +52,14 @@ using std::endl;
 #endif
 
 #define LOG_INFO LogMessage(__FILE__, __LINE__)
+#define LOG_INFO_MSG(STREAM) LogMessage(__FILE__, __LINE__, (STREAM))
```
wolffcm: I have been annotating our changes to S2, a la what we did with HSQL.  Although S2 is not being actively developed anymore, so I'm not sure how much value there is to this.  At least if we find a bug, annotations might help figure out if it's due to something we changed.

third_party/cpp/google-s2-geometry/geometry/include/s2geo/base/logging.h
```diff
@@ -52,13 +52,14 @@ using std::endl;
...
 #define LOG_ERROR LOG_INFO
 #define LOG_WARNING LOG_INFO
 #define LOG_FATAL LogMessageFatal(__FILE__, __LINE__)
 #define LOG_QFATAL LOG_FATAL
 
 #define VLOG(x) if((x)>0){} else LOG_INFO.stream()
-
+#define VMLOG(x, msg) if (((msg) == NULL) && ((x)>0)) {} else LOG_INFO_MSG(msg).stream()
```
wolffcm: What does the first argument to this macro do?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -283,6 +285,14 @@ private FunctionId(String name, Type type, int id, int typeParameter, Type[] par
                     new Type[] { Type.SQL_ALL_TYPES, Type.SQL_ALL_TYPES },
                     new short[] {  Tokens.OPENBRACKET, Tokens.QUESTION, Tokens.COMMA,
                                    Tokens.QUESTION, Tokens.CLOSEBRACKET }),
+
+            new FunctionId("isvalid", Type.SQL_BOOLEAN, FUNC_VOLT_VALIDATE_POLYGON, -1,
+                    new Type[] { Type.VOLT_GEOGRAPHY },
+                    new short[] { Tokens.OPENBRACKET, Tokens.QUESTION, Tokens.CLOSEBRACKET }),
+
+            new FunctionId("invalid_polygon_reason", Type.SQL_VARCHAR, FUNC_VOLT_POLYGON_INVALID_REASON, -1,
```
wolffcm: OpenGIS has a function called `ST_isValidReason`.  It behaves like what you've implemented here.  It's kind of an awkward name though, and I don't know if it's OGC standard or not.

I might be okay with diverging from PostGIS here if the function name is not standard, but most of our other geo functions don't use underscores between words, OGC seems to prefer camel-case.

http://postgis.net/docs/ST_IsValidReason.html

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -42,6 +43,10 @@
      *     so that the area inside the ring is on the left side of the line segments
      *     formed by adjacent vertices.
      *
+     * Note that this is different from the order expected by the OGC standard's
```
wolffcm: Should we ask the user to provide *all* the rings in CCW orientation, to match what WKT does?  It would be good for this user-facing interface as simple and consistent as possible.

wolffcm: Oops, I meant outer ring as CCW and inner rings as CW.  Should we require the user to match what WKT requires?

wolffcm:
That's all my comments for this round.  I had a couple of points for discussion:
- I'm thinking that the point list constructor to GeographyValue should also follow WKT conventions.
- The name of the "reason" function seems like it could be improved.

wolffcm:
One thing that just occurred to me, could you add these two new functions
to fullddl.SQL? There's limited utility to indexing on a Boolean function,
but it would be good to have our bases covered.

src/ee/expressions/geofunctions.cpp
BillWhite: No.  It's in IsNormalized.

src/ee/expressions/geofunctions.cpp
BillWhite: Done.

src/ee/expressions/geofunctions.cpp
BillWhite: That was confusing.  The function started out as isNotMultiPolygon, but it seemed
easier to make it isMultiPolygon.  I never switched the boolean sense, though.

src/frontend/org/voltdb/types/GeographyValue.java
BillWhite: I was thinking this was not really user visible.  But I guess this could be called from a
stored procedure, couldn't it?  If the user somehow constructed the loops from some other
source, perhaps they would be in OGC orientation, if not as WKT strings.  That seems
quite likely, in fact.

I'll change this.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
BillWhite: Well, it's not really a valid reason, it's an invalid reason.  I could go with isInvalidReason.

third_party/cpp/google-s2-geometry/geometry/include/s2geo/base/logging.h
BillWhite: I added a comment explaining this.

VLOG and VMLOG, which latter I added, are logging macros.  VLOG(x) returns a std::ostream which can be used to log a message.  It uses the usual C Preprocessor magic to allow this stream to be used immediately as 'VLOG(2) << "My Message"'.  This expands to
    'if ((2) > 0) {} else LogMessage(__file__, __line__) << "MyMessage"
which has no effect.  If one write 'VLOG(0) << "MyMessage"' then the message would be logged to stderr.

Presumably x is most typically a compile time macro.  But it doesn't seem to be here.

BillWhite:
Sure.  I'll add them to fullddl.SQL.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
BillWhite: You know, camel case is not really that helpful in SQL.

wolffcm: Yeah, I could live with isInvalidReason.  It diverges from the equivalent PostGIS function but their name is awkward.

About camel case, you're right.  I think we're kind of stuck with it now though.  The OGC standard doesn't seem to favor underbars for some reason.

src/ee/expressions/geofunctions.cpp
```diff
@@ -369,4 +403,64 @@ template<> NValue NValue::call<FUNC_VOLT_DISTANCE_POINT_POINT>(const std::vector
...
+    return ValueFactory::getBooleanValue(returnval);
+}
+
+template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_INVALID_REASON>() const {
+    assert(getValueType() == VALUE_TYPE_GEOGRAPHY);
+    if (isNull()) {
+        return NValue::getNullValue(VALUE_TYPE_VARCHAR);
+    }
+    // Extract the polygon and check its validity.
+    std::stringstream msg;
```
hhakimi53: May be ostringstream in here?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: That's interesting. 

I quickly checked for one of collinear's area and it's area is returned as 0.

I was looking around net for that, found http://www.agile-online.org/Conference_Paper/CDs/agile_2012/proceedings/papers/Paper_Ledoux_Automatically_repairing_invalid_polygons_with_a_constrained_triangulation_2012.pdf,  Check section 2.

BillWhite: I wonder if this is an instance of the parallel vs. great circle phenomenon we have seen.  For example, a sequence of k points all with the same latitude but different longitudes will be slightly south of a single edge from the last point to the first point of the sequence.  So, these are not actually collinear.

I wonder if it's possible to make a truly collinear polynomial in the sense we imagine.

third_party/cpp/google-s2-geometry/geometry/include/s2geo/base/logging.h
BillWhite: I made many changes to the file layouts before I checked it into git at all.  I got it from a source
which had made other changes to make it compile on the mac.  I verified that the original sources
and the mac-compilable sources were identical, except for the changes needed to compile on the mac.   I also changed the CMakeLists.txt file, which is what cmake uses to describe the project.

It seems to me that we could just as well compare this with the initial version which we checked into github and see the differences.  I'm not sure annotating them in the source is helpful.

BillWhite:
Squash merged.

PR #3095, opened 2015-12-17
Updated to_timestamp calls to use the new DATEADD function

ssomagani:
Replaced instances of date computation using TO_TIMESTAMP with DATEADD in SQL. 
Had to modify the Java code in Reporter.java since I couldn't negate the input parameter (amount to add) in SQL.


wolffcm:
Looks good to me, and the SQL reads much nicer.  Thanks for doing this.

I ran into the same issue when I tried to negate a parameter to DATEADD a while ago.  Since you ran into it too, I filed a ticket for it:

https://issues.voltdb.com/browse/ENG-9622

PR #3098, opened 2015-12-21
ENG-9630 fix BigDecimal ArrayOutOfBounds

paulmartel:
caused by specifying BigDecimal values that are multiples of 10, 100, etc. using E notation in a string initializer.


wolffcm:
I'll have a look at this.

wolffcm:
Looks good.

PR #3101, opened 2015-12-23
Add new functions to expression indexes in fullDDL.sql

wolffcm:
Also, I fixed an issue with regexp_position and its optional
parameter.


BillWhite:
Seems plausible to me.

hhakimi53:
Looks good.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -521,7 +521,15 @@ public String getSQL() {
...
-            sb.append(Tokens.T_COMMA).append(nodes[ii].getSQL());
+            if (nodes[ii] != null) {
+                sb.append(Tokens.T_COMMA).append(nodes[ii].getSQL());
+            }
+            else {
+                // regexp_position has an optional third parameter.
+                // No other VoltDB functions have optional params.
+                assert (ii == 2);
+                assert (m_def.getId() == FunctionId.FUNC_VOLT_REGEXP_POSITION);
+            }
```
paulmartel: It's not clear that the exception is warranted here -- is it strange and inconsistent for FUNC_VOLT_REGEXP_POSITION to define an extra nodes[] element and set it to null vs. just defining one fewer element? In other words, this fix works, but is there a cleaner fix?

wolffcm: I agree with you, it seems strange and inconsistent.  There might be a cleaner fix.  I had trouble finding the place where the child node list is allocated based on the entered input syntax.  I want to get this into 5.9, but maybe this warrants a little more time investigating.

wolffcm: I spent some time investigating this today.  The function's node list seems to intentionally include null values for optional parameters that are omitted.  This seems weird but I feel like something somewhere might need the null values in there, so I'm reticent to change it.

It seems safe for the VoltDB functions just to skip over the nulls.  I updated the comment to reflect the result of my investigation.

I thought that I might look at the FunctionSQL.getSQL method to get an idea of the "right way" to dump SQL for a function... it turns out this method is just a huge switch statement with explicit code for every function.  Crazy.

PR #3102, opened 2015-12-23
Eliminate persistenttable_test flakiness (especially for -Dbuild=memc…

paulmartel:
…heck on mac) from over-dependence on iterator order determinism.

With -Dbuild=memcheck, iterators over even the simplest persistent tables can deviate from insertion order.


jxchinaren:
Looks good to me.

PR #3105, opened 2015-12-23
ENG-9495: Test to validate snapshot forward compatibility

afrogers:


afrogers:
Everything not covered specifically in the unit test has been manually verified; due to current API limitations, it was not feasible to automate these checks.

PR #3109, opened 2015-12-24
Eng 9485 sqlcov geo basic

johnTcrawford:
Covers ENG-9485, ENG-9496, and part of ENG-9497.


lib/python/voltdbclient.py
```diff
@@ -113,6 +113,8 @@ class FastSerializer:
     VOLTTYPE_MONEY = 20     # 8 byte long
     VOLTTYPE_VOLTTABLE = 21
     VOLTTYPE_VARBINARY = 25
+    VOLTTYPE_GEOGRAPHY_POINT = 26
+    VOLTTYPE_GEOGRAPHY = 27
```
wolffcm: What is the relationship between this file and the corresponding one in the repo VoltDB/voltdb-client-python?

https://github.com/VoltDB/voltdb-client-python/commits/master/voltdbclient.py

These two files seem to be kept roughly in sync?

lib/python/voltdbclient.py
```diff
@@ -742,6 +751,58 @@ def writeDecimalString(self, num):
             raise TypeError("num must be of type decimal.Decimal")
         self.writeString(num.to_eng_string())
 
+    # point (GEOGRAPHYPOINT)
+    def readGeographyPoint(self):
+        # length preceded (4 byte value) string
+        length = self.readInt32()
```
wolffcm: These changes don't seem quite right.  GEOGRAPHY_POINT is a fixed-size type, so it's not length-prefixed.

lib/python/voltdbclient.py
```diff
@@ -742,6 +751,58 @@ def writeDecimalString(self, num):
...
+        cnt = self.readInt16()
+
+        for i in xrange(cnt):
+            retval.append(self.readGeographyPoint())
+
+        return tuple(retval)
+
+    def writeGeographyPoint(self, value):
+        if value is None:
+            self.writeInt32(self.NULL_GEOGRAPHY_POINT_INDICATOR)
```
wolffcm: For passing a null point to a stored procedure, we should be writing (360.0, 360.0) as doubles, rather than a single 32-bit integer.  Unless I'm missing some context here?

lib/python/voltdbclient.py
```diff
@@ -742,6 +751,58 @@ def writeDecimalString(self, num):
...
+        self.writeInt32(len(encoded_value))
+        self.wbuf.extend(encoded_value)
+
+    # polygon (GEOGRAPHY)
+    def readGeography(self):
+        # length preceded (4 byte value) string
+        length = self.readInt32()
+        if self.NullCheck[self.VOLTTYPE_GEOGRAPHY](length) == None:
+            return None
+        return self.readStringContent(length)
```
wolffcm: I don't think that reading the string content will "just work" since we do some byte-swapping to put the data onto the wire.  Unfortunately this looks like another place we need to reimplement the protocol for geography.

lib/python/voltdbclient.py
```diff
@@ -742,6 +751,58 @@ def writeDecimalString(self, num):
...
+
+    def writeGeography(self, value):
+        if value is None:
+            self.writeInt32(self.NULL_GEOGRAPHY_INDICATOR)
+            return
+
+        encoded_value = value.encode("utf-8")
+        self.writeInt32(len(encoded_value))
+        self.wbuf.extend(encoded_value)
+
```
wolffcm: Are these changes actually being used by sqlcoverage?

src/frontend/org/voltdb/PostGISBackend.java
```diff
@@ -0,0 +1,221 @@
...
+                "ST_DISTANCE(", ",FALSE", null, false, false, "columns");
+    }
+
+    /** Modify a query containing a CONTAINS(column1,column2) function, which
+     *  PostgreSQL/PostGIS does not support, and replace it with
+     *  ST_COVERS(column1,column2,FALSE), which is a (spherical Earth)
+     *  equivalent that PostGIS does support. */
+    static private String transformContainsQuery(String dml) {
+        return transformQuery(dml, containsQuery, "",
+                "ST_COVERS(", "", null, false, false, "columns");
```
wolffcm: Oh, `ST_CONTAINS` is not appropriate here?  I wonder why.

src/frontend/org/voltdb/PostGISBackend.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /** For a SQL DDL statement, replace (VoltDB) keywords not supported by
+     *  PostgreSQL/PostGIS with other, similar terms. */
+    static public String transformDDL(String ddl) {
+        // TODO: make this more robust, using a regex Pattern(s)??
+        String modified_ddl = PostgreSQLBackend.transformDDL(ddl)
+                .replace("GEOGRAPHY_POINT", "GEOGRAPHY(POINT,4326)")
```
wolffcm: 4326 is a spherical SRS?  If you have details it would be good to put them here as comments.

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -0,0 +1,69 @@
...
+-- An 8-point (approximately) star-shaped polygon, with an (approximately) star-shaped (symmetric, half-sized) hole:
+INSERT INTO _table VALUES (209, pointFromText('POINT(  0.0  0.0 )'),  pointFromText('POINT(-71.06 42.36)'), polygonFromText('POLYGON((-1.0 0.0, -0.09  -0.04, -0.7  -0.7 , -0.04 -0.09 , 0.0 -1.0, 0.09  -0.04, 0.7  -0.7 , 0.04 -0.09 , 1.02 0.0, 0.09  0.04, 0.7  0.7 , 0.04 0.09 , 0.0 1.0, -0.09  0.04, -0.7  0.7 , -0.04 0.09 , -1.0 0.0),(-0.5 0.0, -0.045 -0.02, -0.35 -0.35, -0.02 -0.045, 0.0 -0.5, 0.045 -0.02, 0.35 -0.35, 0.02 -0.045, 0.5 0.0, 0.045 0.02, 0.35 0.35, 0.02 0.045, 0.0 0.5, -0.045 0.02, -0.35 0.35, -0.02 0.045, -0.5 0.0))') )
+
+-- Test points and the asText, LONGITUDE & LATITUDE functions
+SELECT @star from _table G01
+SELECT ID,    asText(PT1),                   asText(PT2)                from _table G02
+SELECT ID, LONGITUDE(PT1), LATITUDE(PT1), LONGITUDE(PT2), LATITUDE(PT2) from _table G03
+SELECT ID, LONGITUDE(PT1), LATITUDE(PT1), LONGITUDE(PT2), LATITUDE(PT2) from _table G04 WHERE LONGITUDE(PT1) < 0 AND LATITUDE(PT1) > 0
+
+-- Test the DISTANCE function, with points & polygons
```
wolffcm: Functions like DISTANCE that accept two arguments might benefit from having two tables in the from clause, so you can compare each point in one table to each point in a second (possible the same) table.

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -0,0 +1,69 @@
...
+SELECT ID, DISTANCE(PT1,  PT2  ) from _table G11
+SELECT ID, DISTANCE(PT1,  PT2  ) from _table G12 WHERE DISTANCE(PT1,  PT2  ) > 200000
+SELECT ID, DISTANCE(PT1,  POLY1) from _table G13
+SELECT ID, DISTANCE(PT1,  POLY1) from _table G14 WHERE DISTANCE(PT1,  POLY1) > 200000
+SELECT ID, DISTANCE(POLY1,PT1  ) from _table G15
+SELECT ID, DISTANCE(POLY1,PT1  ) from _table G16 WHERE DISTANCE(POLY1,PT2  ) > 200000
+-- DISTANCE between two polygons is not yet supported
+SELECT ID, DISTANCE(POLY1,POLY1) from _table G17
+SELECT ID, DISTANCE(POLY1,POLY1) from _table G18 WHERE DISTANCE(POLY1,POLY1) < 200000
```
wolffcm: This would probably get an error in VoltDB but not PostGIS?  One way to work around this would be to do DISTANCE(CENTROID(poly1), poly2).

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -0,0 +1,69 @@
...
+SELECT ID, asText(PT1),    asText(POLY1) from _table G32 WHERE _maybe CONTAINS(POLY1,PT1)
+SELECT ID, LONGITUDE(PT2), LATITUDE(PT2) from _table G33 WHERE _maybe CONTAINS(POLY1,PT2)
+SELECT ID, asText(PT2),    asText(POLY1) from _table G34 WHERE _maybe CONTAINS(POLY1,PT2)
+-- Test UPDATE using CENTROID, and re-test CONTAINS afterward
+UPDATE _table G35 SET PT2 = CENTROID(POLY1)
+SELECT ID, LONGITUDE(PT2), LATITUDE(PT2) from _table G36 WHERE _maybe CONTAINS(POLY1,PT2)
+SELECT ID, asText(PT2),    asText(POLY1) from _table G37 WHERE _maybe CONTAINS(POLY1,PT2)
+
+-- These won't work until CONTAINS (& boolean return values) is supported in the initial SELECT clause
+SELECT ID, CONTAINS(POLY1,PT1), LONGITUDE(PT1), LATITUDE(PT1) from _table G38
```
wolffcm: Do we have a ticket for the CONTAINS on the select list issue?

wolffcm:
That's all my comments.  This looks like it took a lot of effort and I'm definitely impressed.

How close is this to passing?  

I have concerns about the changes to the Python client.  Can you explain why they were necessary?

Aside from the python client issue, I'm okay with merging this work.  Even if it's not passing, it won't be run in Jenkins until we alter some configuration in ant, right?

tests/scripts/examples/sql_coverage/geo-basic.sql
wolffcm: I don't see a ticket for this so I'm going to file one.

lib/python/voltdbclient.py
johnTcrawford: I have no idea.

lib/python/voltdbclient.py
johnTcrawford: No, I'm going to remove them, except for the first two lines:
VOLTTYPE_GEOGRAPHY_POINT = 26
VOLTTYPE_GEOGRAPHY = 27
which is all that is actually being used.

src/frontend/org/voltdb/PostGISBackend.java
johnTcrawford: ST_CONTAINS takes only geometry objects; ST_COVERS takes geography (or geometry) objects, which is what we want. See:
http://postgis.net/docs/ST_Contains.html
http://postgis.net/docs/ST_Covers.html

src/frontend/org/voltdb/PostGISBackend.java
johnTcrawford: 4326 is the standard, spheroidal SRIS used by PostGIS. I'll add a comment.

lib/python/voltdbclient.py
johnTcrawford: Ruth says that these two files are generally kept in sync; but that she would not bother to change the other file, for just these two lines (which is the only change to this file that I am keeping).

src/frontend/org/voltdb/PostGISBackend.java
johnTcrawford: BTW, 3857 is another SRIS (EPSG) that we could consider using, but I'm not sure whether it really matters, or by how much, or which is better.

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: No, because of the way that SQLCoverage and the various (HSQL, PostgreSQL, PostGIS) backends are set up, they both fail in the same way, in the (VoltDB) planner; of course, you are correct that PostGIS could handle this, if we let it get that far.

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: Not a bad idea. I'll add a line or two.

johnTcrawford:
Thanks. As noted above, I'm taking out almost all of the Python client work, which was not really used. There are still numerous failures, mostly in UPDATE statements, due to 2 issues (one involving comparison operators, the other involving setting points to polygons, or vice versa), but also some round-off failures (which may be improved by Huzefa's recent work).

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: Thanks.

wolffcm:
Looks good to me.

PR #3110, opened 2015-12-24
Eng 9637

manjujames:
Added stopImporter method in AbstractImporter that sets stopped flag and calls the importer specific stop method.


vtkstef:
Reviewed

PR #3114, opened 2015-12-28
Validate polygons.

BillWhite:
This commit adds two different sets of functionality, and completes the
work on two tickets.

We add a boolean SQL function isValid(), which takes a geography value
and tells if it is valid or not.  We add a varchar valued SQL function
polygon_invalid_reason which takes a geography value and tells why it is
invalid if it is invalid. If the geography is valid the function returns
the string "Valid polygon".

In the EE, polygons are always made up of sequences of sequences of
vertices. The inner sequences represent loops.  The sequence has an
implicit edge from the last vertex to the first vertex, forming a loop.
An implicit loop is either an exterior shell, or else interior,
contained holed.  Each polygon has one shell and zero or more holes.
The first implicit loop is the shell, and subsequent loops are holes.
In the EE, the implicit loop of a shell or a hole is always given in
counter-clockwise order.

The SQL function isValid rejects multi-polygons and polygons whose area
is more than half the earth. It also rejects polygons whose rings cross
or intersect in bad ways.

In Java, polygons may be constructed two ways, as OGC well known text,
or as Lists of Lists of GeographyPointValue objects.  The last vertex of
all loops must be equal to the first vertex, so that the loops are
closed.  Shells must be given in counter-clockwise order, so that the
interior of the polygon is on the left hand side of the border.  Holes
must be given in clockwise order, so that the interior of the polygon is
on the left hand side of the border, and the hole is on the right hand
side.

https://issues.voltdb.com/browse/ENG-9205
https://issues.voltdb.com/browse/ENG-9609


BillWhite:
Jenkins is mostly blue with this branch, with only TestFuzzArbiter and TestExportRejoin failing. These are unrelated to this change, so I'm merging this.

PR #3115, opened 2015-12-28
Use earth's mean radius for geo computation

hhakimi53:
Creating pull request again. 

Changes uses mean radius of 6371008.8 meters, when calculating great circle distance and area for the spherical earth. For comparison purpose, compared the distance and area provided by VoltDB to PostGIS. Below is summary of comparison.
- Distance between point-to-point location agree to 7 digit of precision. For shorter distance on scale of 1000 meters, the answers agree till milli-meters and for long distances like distance between Boston to San Jose, answers agree to precision level of meters. VoltDB answer: 4311575.515808559; PostGIS: 4311575.49646372.
- The difference in distances between spherical-to-spheriod (Spheriod distance calculated using PostGIS) distances is about 0.03% (https://en.wikipedia.org/wiki/Great-circle_distance#Radius_for_spherical_Earth)
- The spherical area of polygon differs by spheriod area for couple of polygons of shape similar to Colorado, Wyoming by around 0.08%.

### DIFF
--- a/src/ee/expressions/geofunctions.cpp
+++ b/src/ee/expressions/geofunctions.cpp
@@ -33,8 +33,8 @@ namespace voltdb {
 static const int POINT = FUNC_VOLT_POINTFROMTEXT;
 static const int POLY = FUNC_VOLT_POLYGONFROMTEXT;
 
-static const double EARTH_AREA_SQ_M = 510.072E12;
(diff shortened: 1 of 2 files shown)


wolffcm:
Looks good to me.

PR #3122, opened 2015-12-30
Eng 9522 geo polygon factory

BillWhite:
We create two functions to create polygons.  One creates regular polygons and one creates N-pointed stars.

We also add some functions to allow algebra on GeographyValuePoints.  We have point functions to add and subtract points, to scale points and to rotate a point around another point through a given angle.  This is useful in creating the factory functions.

We also add some functions to normalize longitude and latitude values which are out of the range [-180,180] and [-90,90] respectively.  Again, this is helpful when calculating the vertices of factory-made polygons.


src/frontend/org/voltdb/utils/PolygonFactory.java
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * A class to create polygons.
+ */
+public class PolygonFactory {
+    /**
+     * Create a regular convex polygon, with an optional hole.
+     * @param center The center of the polygon.
+     * @param firstVertex The coordinates of the first vertex.
+     * @param numVertices The number of vertices.
+     * @param sizeOfHole If this is positive, we also create a hole whose vertices are
```
wolffcm: Whose vertices are...?

src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -57,8 +59,9 @@
     static final double NULL_COORD = 360.0;
 
     public GeographyPointValue(double longitude, double latitude) {
-        m_latitude = latitude;
-        m_longitude = longitude;
+        // Add 0.0 to avoid -0.0.
+        m_latitude = latitude + 0.0;
```
wolffcm: This seems esoteric.  What's wrong with -0.0 here?

src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -105,16 +109,40 @@ public double getLongitude() {
...
+        // Choosing the default, which is 12, gives us about
+        // 1 um = 1/1000 mm precision.
+        double epsilon = Math.pow(10.0, -ndigits);
+        StringBuilder sb = new StringBuilder(ndigits + 4);
+        sb.append("##0.0");
+        while (--ndigits > 0) {
+            sb.append("#");
+        }
+        DecimalFormat df = new DecimalFormat(sb.toString());
+        // Explicitly test for differences less than 1.0e-12 and
```
wolffcm: Should we be rounding when we receive lng/lat values in the constructor (I'm thinking of the loop constructor)?

src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -193,4 +221,181 @@ public static void serializeNull(ByteBuffer buffer) {
...
+        double lngNorm = normalize(longitude, 360);
+        double latFinal = 0.0;
+        double lngFinal = 0.0;
+        assert(-180 <= latNorm && latNorm <= 180);
+        assert(-180 <= lngNorm && lngNorm <= 180);
+        // System.out.printf("        // longitude = %f, latitude = %f\n", longitude, latitude);
+        // System.out.printf("        // normLng   = %f, normLat  = %f\n", lngNorm, latNorm);
+        // Now, latOrig is the latitude in the range [-180,180].
+        // Let latWant be latitude in the range [-90, 90]. Then
+        // <ul>
```
wolffcm: I guess these HTML tags will make the web documentation look pretty?  I'm not sure I've seen it elsewhere in our code.

wolffcm: Wait, this isn't a javadoc comment, so I don't see how the HTML tags do anything here.

src/frontend/org/voltdb/types/GeographyPointValue.java
wolffcm: Oops, we're not in Geography here, we're in GeographyPoint.

src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -193,4 +221,181 @@ public static void serializeNull(ByteBuffer buffer) {
...
+     * map t = -500 = (-540 + 40) to 40 and t = 710 = (720 - 10) to -10.
+     *
+     * This can be used to normalize longitude, with range=360.  Latitude is more
+     * difficult, however.  See {@link normalizeLatLng} below.
+     *
+     * @param v
+     * @param range
+     * @return
+     */
+    protected static GeographyPointValue normalizeLngLat(double longitude, double latitude) {
```
wolffcm: Can this be private instead of protected?

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -138,17 +138,32 @@ public static GeographyValue fromText(String text) {
 
     /**
      * Gets the loops that make up the polygon, with the outer loop first.
-     * @return  The loops in the polygon as a list of a list of points
+     * Include the loop back to the first vertex.  Also, reverse the order
```
wolffcm: This comment will appear in the Javadoc for this class and I think it might confuse users.  All they really need to know is that what they get back will be consistent with WKT.

This comment would be appropriate inside the method body though.

wolffcm: This reminds me that we should go over Javadoc for user-facing classes.

src/frontend/org/voltdb/utils/PolygonFactory.java
```diff
@@ -0,0 +1,183 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2015 VoltDB Inc.
```
wolffcm: May as well make the copyright year 2016 here.

wolffcm:
What I've looked at so far looks very good.  I still need to review the tests.

src/frontend/org/voltdb/utils/PolygonFactory.java
BillWhite: Yes.  I've fixed this.

src/frontend/org/voltdb/types/GeographyPointValue.java
BillWhite: It throws off the conversion to string.  We get -0.0 rather than 0.0.
I suppose this could be done in the string conversion, however.

src/frontend/org/voltdb/types/GeographyPointValue.java
BillWhite: Ok, I'll change them from html.

src/frontend/org/voltdb/types/GeographyPointValue.java
BillWhite: It needs to be at least package, because I need it in the tests.  However, I would argue for it being public.  I know it's not public here, but it's really more friendly to do this calculation.  If users do some kind of geometric calculation they might very well produce non-normalized longitude/latitude pairs, and they would be forced to do this for themselves.

src/frontend/org/voltdb/types/GeographyValue.java
BillWhite: Ok, I suppose I could put it there.  I'm not sure anybody is going to be confused by this.

src/frontend/org/voltdb/types/GeographyPointValue.java
BillWhite: I'm not sure what digit to round to.  We need different sizes in different contexts.

src/frontend/org/voltdb/utils/PolygonFactory.java
BillWhite: Shouldn't all files changed in 2016 have copyright dates changed?  That's the way things were at The Mathworks.  I"m not sure I remember anybody who cared about it as much in other companies.

BillWhite: As it turns out, the licensecheck tool will fail with this copyright change.

src/frontend/org/voltdb/types/GeographyPointValue.java
wolffcm: Looking at your changes, I'm not sure why we need different sizes.  I see one place where you note in a comment that you're using 9 digits for no particular reason.  Making an additional toString method that accepts the number of digits to display seems like a needless complication to a public-facing API.  Am I missing something?

src/frontend/org/voltdb/types/GeographyPointValue.java
wolffcm: I see your point, this might be handy for some users.

src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -193,4 +221,178 @@ public static void serializeNull(ByteBuffer buffer) {
...
+     * Let v be in the interval [k*range/2, (k+1)*range/2]. Then return
+     * the v-(k*range/2)-range/2.  If v is an integer multiple of
+     * range/2, then either -range/2 or range/2 is possible.  In this
+     * case, preserve the sign of v.
+     *
+     * For example, when range = 360, map t in [-180, 180] to itself,
+     * map t = -500 = (-540 + 40) to 40 and t = 710 = (720 - 10) to -10.
+     *
+     * This can be used to normalize longitude, with range=360.  Latitude is more
+     * difficult, however.  See {@link normalizeLatLng} below.
```
wolffcm: This link seems obsolete?  There's no function with that name?

src/frontend/org/voltdb/utils/PolygonFactory.java
wolffcm: I guess do whatever you need to to get licensecheck to pass.

tests/frontend/org/voltdb/utils/TestPolygonFactory.java
```diff
@@ -0,0 +1,213 @@
...
+            assertEquals(2, loops.size());
+            shell = loops.get(0);
+            List<GeographyPointValue> hole = loops.get(1);
+            assertEquals(2*idx+1, shell.size());
+            assertEquals(2*idx+1, hole.size());
+        }
+    }
+
+    /**
+     * Create 60 regular convex polygons.  In returnValue.get(n).get(k) we put an
```
wolffcm: Minor: this seems like it will create 50 polygons, not 60.

tests/frontend/org/voltdb/utils/TestPolygonFactory.java
```diff
@@ -0,0 +1,213 @@
...
+                GeographyPointValue firstVertex = firstFirstVertex.add(x.mul(xmul*idx).add(y.mul(ymul*hidx)));
+                oneSize.add(PolygonFactory.CreateRegularConvex(center, firstVertex, idx + 3, hidx*0.2));
+            }
+            answer.add(oneSize);
+        }
+        return answer;
+    }
+
+    /**
+     * Create 250 = 10*5*6 star-like polygons.  In returnValue.get(n).get(s).get(k) we put an
```
wolffcm: This comment seems kind of obsolete, also 10*5*6 is 300, not 250.

wolffcm:
That's all my comments.  This all looks very nice---I'm glad you've taken the approach you did.

My only issue was with the toString method taking a precision argument.

src/frontend/org/voltdb/types/GeographyPointValue.java
BillWhite: This is probably a bug, and I'm ripping it out.  The problem was that in TestPolygonFactory.testRegularConvexPolygon, some coordinates had 9 digits of fractional precision in their string representation in the WKT I was testing for, but I got 12 digits from toString.  I replaced the 9 digits with 12 digit values and the test passes.

I think there is something here, however, but I don't know exactly what it is.  Say that a user creates coordinates which have an inherent accuracy, say from GPS which is accurate to a meter or a few meters.  The user then uses floating point values which are not rounded to the precision of the inherent accuracy, but, when rounded, are equal to the accurate value.  The WKT values for the accurate value and the represented value will not match.  We already have this when we convert from longitude/latitude to XYZPoint, where some numerical inaccuracy is introduced.  We ignore this by restricting to 12 fractional digits.  But this is just because we have done testing to show that we only introduce errors in the lower-order bits.  We don't know the user's accuracy.

I think this requires more thought.  Maybe we need to specify that users must round input to the accuracy they expect.  But then calculations done in stored procedures may cause precision errors in bits which will be lost in rounding, and these may show up in the WKT.

In any case, I'm ripping this out in GeographyValue and GeographyPointValue.

BillWhite:
I'm closing this pull request so that I can delete the branch and start over.

PR #3126, opened 2016-01-04
Eng 9516 a default query timeout for reads 

JadenW:
so that I don't think my new VoltDB cluster is hung

With default value of 10 Seconds.
Also set this default value to SQLCommand

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -1148,8 +1148,8 @@ public class ProcedureRunner {
                }
                else {
                    msg.append(" default query timeout period.");
-                   msg.append(" The default query timeout is currently " +  originalTimeout/1000.0 + " seconds and can be changed in the systemsettings section of the deployment file. ");
                }
+               msg.append(" The default query timeout is currently " +  originalTimeout/1000.0 + " seconds and can be changed in the systemsettings section of the deployment file. ");
            }
        }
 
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1028,7 +1028,7 @@ private static void printUsage(int exitCode)
         + "  and returns to the command prompt. Default: true.\n"
         + "\n"
         + "[--query-timeout=millisecond_number]\n"
-        + "  Read-only queries that take longer than this number of milliseconds will abort. Default: no timeout.\n"
+        + "  Read-only queries that take longer than this number of milliseconds will abort. Default: 10 Seconds.\n"
```
akhanzode: Use constant here as well in generating message.

akhanzode:
When query times out do we give in error message what user can do to change this?

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -1139,6 +1139,18 @@ else if (e.getClass() == SpecifiedException.class) {
        // ensure the message is returned if we're not going to hit the verbose condition below
        if (expected_failure || hideStackTrace) {
            msg.append("  ").append(e.getMessage());
+           if (e instanceof org.voltdb.exceptions.InterruptException) {
```
vtkstef: if you have to have the timeout value reflected in VoltDB.Configuration and get it with `VoltDB.instance().getConfig()` this bit i adding two map lookups for every transaction, which will simply not do.

akhanzode: Also check m_isReadOnly this message is only for RO queries.

PR #3127, opened 2016-01-04
Update geo-point test for decimal rounding

hhakimi53:
- Updated  GeographyPointValue test to take to precision of 12 decimal places when comparing two points.
- Minor changes to build script to ignore warnings for unused defines for gcc version 5. This started happening with version 5.2.1 abd results in build to fail due to unused defs flagged from s2-google apis.

Testing:
- Don't see asserts reported for points being different with changes locally that were on integration branch in memcheck environment.
- Jenkins run on the changes is under-way

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestGeograpyPointValue.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestGeograpyPointValue.java
@@ -38,6 +38,7 @@ public class TestGeograpyPointValue extends RegressionSuite {
         super(name);
     }
 
+    private static final double EPSILON = 1.0e-12;
     private static final GeographyPointValue BEDFORD_PT = new GeographyPointValue(-71.2767, 42.4906);
     private static final GeographyPointValue SANTA_CLARA_PT = new GeographyPointValue(-121.9692, 37.3544);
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/regressionsuites/TestGeograpyPointValue.java
```diff
@@ -375,4 +374,6 @@ public void testPointIn() throws Exception {
 
         return builder;
     }
+
+    private final double EPSILON = 1.0e-12;
```
wolffcm: Member variables typically go at the top of a class rather than the bottom, maybe put this up top?

wolffcm:
Looks good to me.

tests/frontend/org/voltdb/regressionsuites/TestGeograpyPointValue.java
hhakimi53: Sure, will move it to top.

In general what I have done, mostly in C++, public scope at top and private at bottom; within that functions first and member variables next :)

hhakimi53:
Thanks for reviewing the changes. Jenkins passed for the previously failing tests and there were couple of intermittent ones that failed - seems environmental related. I will make the changes in review feedback and push the changes to integration branch after a fresh jenkins run.

PR #3128, opened 2016-01-04
Fixed simple typo in RuntimeException message

basvanbeek:
Nothing Major... just found it when debugging wire protocol client


jpiekos:
Thanks Bas!

PR #3130, opened 2016-01-05
Size specifiers for GEOGRAPHY columns

wolffcm:
This was fairly straightforward.  There was only changes to the front end needed, and we can treat GEOGRAPHY just like VARBINARY.

I still need to push a commit to make the sizing worksheet in VMC work right.


src/hsqldb19b3/org/hsqldb_voltpatches/types/Type.java
```diff
@@ -945,7 +945,7 @@ public static Type  getType(int type, int collation, long precision,
                 return VOLT_GEOGRAPHY_POINT;
 
             case Types.VOLT_GEOGRAPHY :
-                return VOLT_GEOGRAPHY;
+                return new VoltGeographyType(precision);
```
hhakimi53: in getDefaultTypeWithSize() around line 670, do we need to include Geography type in the switch statement? Not sure, noticed it so thought will ask.

hhakimi53:
Overall changes looks good to me.

src/hsqldb19b3/org/hsqldb_voltpatches/types/Type.java
wolffcm: Hmm, I'm not clear on what that code is used for.  I think it's probably to support an HSQL feature that we don't care about.  None of my testing has touched getDefaultTypeWithSize().

wolffcm:
I just added another commit that updates the catalog sizing.  I verified by hand that the sizing worksheet treats GEOGRAPHY columns the same as VARBINARY.

I also added a test for alter table.

src/hsqldb19b3/org/hsqldb_voltpatches/types/Type.java
hhakimi53: oh ok. I was not also sure based on looking at the call heirarchy. I agree, if it's in code path for HSQL support only than it's not needed.

hhakimi53:
Changes looks fine to me.

wolffcm:
I think we only support increasing the size of columns when there are rows in the table.  I've modified the test so that we're resizing a populated table, which is the real use case we want to test.

hhakimi53:
Looks good.

wolffcm:
Squashed and merged on a separate branch.

PR #3131, opened 2016-01-05
Eng 9676

manjujames:
Use AtomicReference so that onChange and stop do not step on each other.

### DIFF
--- a/src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
+++ b/src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
@@ -53,7 +53,7 @@ public class ImporterLifeCycleManager implements ChannelChangeCallback
     private final AbstractImporterFactory m_factory;
     private ListeningExecutorService m_executorService;
     private ImmutableMap<URI, ImporterConfig> m_configs = ImmutableMap.of();
-    private AtomicReference<ImmutableMap<URI, AbstractImporter>> m_importers = new AtomicReference<>();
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
```diff
@@ -52,7 +53,7 @@
     private final AbstractImporterFactory m_factory;
     private ListeningExecutorService m_executorService;
     private ImmutableMap<URI, ImporterConfig> m_configs = ImmutableMap.of();
-    private ImmutableMap<URI, AbstractImporter> m_importers;
+    private AtomicReference<ImmutableMap<URI, AbstractImporter>> m_importers = new AtomicReference<>();
```
vtkstef: should you not start with an empty map?

src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
```diff
@@ -162,8 +168,12 @@ public final void onChange(ImporterChannelAssignment assignment)
             builder.put(added, importer);
         }
 
-        m_importers = builder.build();
-        startImporters(newImporters);
+        ImmutableMap<URI, AbstractImporter> newReference = builder.build();
```
vtkstef: on failure shouldn't you create a new builder? and loop again you map and toStop construction?

manjujames: The only way this would fail is if the importer has been stopped. So, there is no need to go ahead with the stop and start importers

manjujames:
squash merged.

PR #3132, opened 2016-01-05
Eng 9582 geo osm import example redo

prosegay:


examples/osm-import/client/osmimport/VoltDBOsmSink.java
```diff
@@ -0,0 +1,351 @@
...
+import org.voltdb.client.ClientStatusListenerExt;
+import org.voltdb.client.NoConnectionsException;
+import org.voltdb.client.ProcedureCallback;
+import org.voltdb.types.TimestampType;
+
+//import SyncBenchmark.StatusListener;
+
+public class VoltDBOsmSink extends VoltProcedure implements Sink, EntityProcessor  {
+
+  public static final String INS_NODE_PROC = "insertNodes";
```
paulmartel: Our [coding standards](https://github.com/VoltDB/voltdb/wiki/Code-conventions-&-licensing#code-conventions) say 4-space indents. 
(I personally never understood why anyone uses 2-space indents; 1-space indents are twice as space efficient and just as legible).

examples/osm-import/client/osmimport/VoltDBOsmSink.java
```diff
@@ -0,0 +1,351 @@
...
+   */
+  public void process(RelationContainer relationContainer) {
+
+    Relation relation;
+    int memberSequenceId;
+
+    relation = relationContainer.getEntity();
+
+    try {
+      client.callProcedure(new InsertCallback(), INS_RELATIONS_PROC, relation.getId(),relation.getVersion(),relation.getUser().getId(),relation.getTimestamp(),relation.getChangesetId());
```
paulmartel: Add spaces between args for readability and don't be afraid to wrap lines.

examples/osm-import/ddl.sql
```diff
@@ -0,0 +1,163 @@
...
+-- this isn't currently supported
+-- CREATE INDEX idx_nodes_geom ON nodes (geom);
+
+CREATE INDEX idx_way_tags_way_id ON way_tags (way_id);
+CREATE INDEX idx_way_nodes_node_id ON way_nodes (node_id);
+
+CREATE INDEX idx_relation_tags_relation_id ON relation_tags (relation_id);
+
+-- create the procedures
+CREATE PROCEDURE insertNodes PARTITION ON TABLE nodes COLUMN id PARAMETER 0 AS INSERT INTO nodes (id,version,user_id,tstamp,changeset_id,geom) values(?,?,?,?,?,?);
```
paulmartel: long lines -- I usually insert a newline and 4 spaces between `AS` and the sql statement.
Also, spaces after commas in lists aids readability -- POSSIBLE trivial exception: `values(?,?...)`

paulmartel: I'm not sure how many of these insert... procedures are, like `insertUsers`, completely redundant with the so-called "CRUD" default procedures that get generated automatically like `users.insert`. If you used the default procs by default and only defined and used custom insert procs where you could state a documented reason/difference, I would not have to think about this.

examples/osm-import/ddl.sql
```diff
@@ -0,0 +1,163 @@
...
+);
+
+PARTITION TABLE relation_tags ON COLUMN relation_id;
+
+-- Configure the schema version.
+-- INSERT INTO schema_info (version) VALUES (5);
+
+
+-- Add primary keys to tables.
+ALTER TABLE schema_info ADD CONSTRAINT pk_schema_info PRIMARY KEY (version);
```
paulmartel: Use the less verbose and more "apparent" syntax for defining primary keys (and other constraints) inside the class definitions.

examples/osm-import/ddl.sql
```diff
@@ -0,0 +1,163 @@
...
+CREATE TABLE nodes (
+    id bigint NOT NULL,
+    version int NOT NULL,
+    user_id int NOT NULL,
+    tstamp timestamp without time zone NOT NULL,
+    changeset_id bigint NOT NULL
+);
+-- Add a postgis point column holding the location of the node.
+-- SELECT AddGeometryColumn('nodes', 'geom', 4326, 'POINT', 2);
+ALTER TABLE nodes add COLUMN geom GEOGRAPHY_POINT;
```
paulmartel: Why is this not simply defined inline with the other columns?

examples/osm-import/ddl.sql
```diff
@@ -0,0 +1,163 @@
...
+CREATE TABLE ways (
+    id bigint NOT NULL,
+    version int NOT NULL,
+    user_id int NOT NULL,
+    tstamp timestamp without time zone NOT NULL,
+    changeset_id bigint NOT NULL
+);
+PARTITION TABLE ways ON COLUMN id;
+
+ALTER TABLE ways add COLUMN bbox GEOGRAPHY;
```
paulmartel: Same comment -- why not inlined?

examples/osm-import/client/osmimport/VoltDBOsmSink.java
```diff
@@ -0,0 +1,351 @@
...
+  // Reference to the database connection we will use
+  private String server;
+  private ClientStatsContext periodicStatsContext;
+  private ClientStatsContext fullStatsContext;
+
+  private Client client;
+
+  public VoltDBOsmSink(String server) {
+    this.server = server;
+    NodeLocationStoreType storeType = NodeLocationStoreType.TempFile;
```
paulmartel: Does it make sense to hide this storeType noise deeper inside WayPolygonGeometryBuilder's constructor and simplify its signature?

examples/osm-import/client/osmimport/VoltDBOsmSink.java
```diff
@@ -0,0 +1,351 @@
...
+import org.voltdb.client.ClientResponse;
+import org.voltdb.client.ClientStatsContext;
+import org.voltdb.client.ClientStatusListenerExt;
+import org.voltdb.client.NoConnectionsException;
+import org.voltdb.client.ProcedureCallback;
+import org.voltdb.types.TimestampType;
+
+//import SyncBenchmark.StatusListener;
+
+public class VoltDBOsmSink extends VoltProcedure implements Sink, EntityProcessor  {
```
paulmartel: It's a small but aggravating thing when naming is inconsistent in its use of upper/mixed case. Here "Osm", there "OSM". Consider standardizing on A) whatever the underlying library uses or B) upper case because Paul prefers it.

paulmartel: "extends VoltProcedure" seems like a mistake here (adds what besides overhead and head-scratching?).

paulmartel:
End of my review round.

examples/osm-import/ddl.sql
jwiebema1: I didn't know we did this automatically,  that would've saved me some significant time.

examples/osm-import/ddl.sql
jwiebema1: I will document this, but I wanted to keep the schema and it's generation as close as possible as the original OSM DDL that is used with PostgreSQL , so if their was a statement that I could keep exactly as they had written it, I kept it that way so it would be easy to diff.

examples/osm-import/ddl.sql
jwiebema1: for the reasons mentioned in the above comments, to keep the DDL from the original as close as possible.

examples/osm-import/client/osmimport/VoltDBOsmSink.java
jwiebema1: The StoreType is supposed to be user configurable from the command line and passed down, but since I don't use it that way in this example, I'll clean it up.

PR #3136, opened 2016-01-06
ENG-9312: Move VoltProcedure deprecated functions elsewhere

jhugg:
This is a bit hackish as we still use these internally and for test a bit. I think my solution is a pretty safe one though. I have some additional work to take this a bit further, but it won't make it for 6.0. Still, 6.0 is the best time to remove these procs.

Has a pro pull as well.


tests/testprocs/org/voltdb_testprocs/regressionsuites/saverestore/GetTxnId.java
```diff
@@ -35,7 +39,7 @@
         new SQLStmt("INSERT INTO PARTITION_TESTER VALUES (?, ?, ?, ?);");
 
     public VoltTable[] run(int partitioning) {
-        this.setAppStatusString(Long.toString(this.getVoltPrivateRealTransactionIdDontUseMe()));
+        this.setAppStatusString(Long.toString(DeprecatedProcedureAPIAccess.getVoltPrivateRealTransactionId(this)));
```
BillWhite: Should this.setAppStatusString be just setAppStatusString?

BillWhite:
Just one small comment.  You made these into static functions, right.  That seems safe.  You can't make them package or protected because they are used in inconvenient places.  So keeping them public seems ok.

tests/testprocs/org/voltdb_testprocs/regressionsuites/saverestore/GetTxnId.java
jhugg: Maybe? I saw a bunch of style changes I could make in code around where I made changes, but tried to change as little as possible for this pull.

BillWhite: Well, this is the kind of think Paul comments on, so I mentioned it.  But I'm indifferent, as I reckon the compiler will know what to do with it.

PR #3137, opened 2016-01-06
Eng 9634 Add some tests for polygons with multiple holes.  

BillWhite:
Verify that contains works in polygons with multiple holes, and that the
distance and area functions do sensible things with multiple holes.
Verify that the order of the loops is not changed by S2.

Note that I did some testing of areas of polygons with multiple holes.
This was not strictly in the ticket, and was marked as a follow-on ticket.
However, I misread the ticket in haste, and implemented it anyway.
It was pretty easy, given the remainder of the things I was doing.


tests/frontend/org/voltdb/regressionsuites/TestGeographyValueQueries.java
```diff
@@ -378,6 +387,190 @@ public void testArithmetic() throws Exception {
...
+        double dist = -1;
+        while (vt.advanceRow()) {
+            if (dist < 0) {
+                dist = vt.getDouble(1);
+            } else {
+                assertEquals("Distances are not equal", dist, vt.getDouble(1));
+            }
+        }
+    }
+
```
wolffcm: You could also check that the distance from center points of the cheesy polygon to the shell polygon is always zero?  But we might have other tests that hit that case.

tests/frontend/org/voltdb/utils/TestPolygonFactory.java
```diff
@@ -173,40 +184,98 @@ public static String formatWKT(String wkt) {
...
-        GeographyPointValue firstVertex = x.mul(0.1).add(y.mul(0.1));
-        List<List<GeographyValue>> polys = makeRegularConvexPolygons(center, firstVertex, 1.0, 1.0);
-        System.out.printf(":-------------------------------:\n");
-        System.out.printf(":------- Regular Convex --------:\n");
-        System.out.printf(":-------------------------------:\n");
-        for (int nsides = 0; nsides < polys.size(); nsides += 1) {
-            for (int holeSize = 0; holeSize < 5; holeSize += 1) {
-                System.out.printf("%s\n",
-                                  formatWKT(polys.get(nsides).get(holeSize).toString()));
+    public static void main(String args[]) {
```
wolffcm: I see that you've filled out more in this class.  Would it be valuable to apply some of our geospatial functions (distance, area, contains), to the polygons generated here?  I had thought that the ticket asked for this, but reading it now, it's unclear.

wolffcm:
That's all my comments.  In summary, I had a question about extending the testing of the geospatial functions to regular and star-shaped polygons---I think it would be a good idea, but is not strictly within the doneness criteria for the ticket.

tests/frontend/org/voltdb/utils/TestPolygonFactory.java
BillWhite: I added command line arguments while I was working on a factory function for cheesy polygons.  It turned out to be more complicated than was warranted for the ticket, and was taking a lot of time.  But I left the CLI processing relevant to the other polygons.

I thought about making tests that would use star-shaped polygons or regular convex polygons with holes and check for contains.  But it wasn't clear what would be tested by this.  I suppose we could ask if points are between the arms.  We could generate these points from the polygon generating parameters, I suppose.  Does this really test that much, though?  If it's broken, it's broken because S2 does the wrong thing.  I guess we want to know that.

wolffcm: Ok, sounds good.  I was thinking that we'd want some CONTAINS tests where the point is in the bounding box for the polygon, but not in the polygon itself--S2 does the bounding box check first to determine if the more expensive check is really necessary.

But you're right that we have other tests for this.  So I'm fine with just creating a second, post-6.0 ticket.

BillWhite:
Closing to squash this.

PR #3141, opened 2016-01-07
ENG-7963: Sizing worksheet overestimates the minimum size of VARCHAR(…

JadenW:
…>=64) columns

Manually test
Voter CONTESTANTS table estimation changes from:
Row Min = 100 Row Max = 292 
to 
Row Min = 32 Row Max = 276

if set CONTESTANT_NAME nullable
Row Min = 12  Row Max = 276

### DIFF
--- a/src/frontend/org/voltdb/utils/CatalogSizing.java
+++ b/src/frontend/org/voltdb/utils/CatalogSizing.java
@@ -211,7 +211,7 @@ public abstract class CatalogSizing {
         return bufferSize;
     }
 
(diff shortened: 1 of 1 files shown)


paulmartel:
I made one comment accidentally on the [commit](https://github.com/VoltDB/voltdb/commit/384631fb225bbdc4ea697fcbc1e1adebc3bde2b4) instead of on this pull request -- sorry.

Make that change -- passing false instead of isNullable when calculating max width in TWO places -- then this code will be ready to merge.

PR #3143, opened 2016-01-08
Update expression's indexing and MV gaurds for expressions

hhakimi53:
- Updated expression's indexing and MVs guards. Guards evaluate conditions list following. Expression that fail guard validation, are disallowed in indexing and in group-by clause of MV.
+ Expressions whose logical result is boolean (some of the expression whose return type is logically boolean, for example comparison expression, have the return value coded as bigint as boolean type is not fully supported). This includes all comparison expression (like =, <>, >, <), operator expression(NOT, isNull and Exists), all conjuction operator (like AND, OR) and functions whose return type is boolean (Contains(), isValid())
+ Expression whose (return) type is geo. This includes geography and geography_point type columns and Centroid() function expression.
- Fixed case where malformed MV ddl could cause out of bound exception
- Added junit test cases for the same.


src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -1185,15 +1184,16 @@ public static boolean hasInlineVarType(AbstractExpression expr) {
...
+     * group by and where clause expression.
+     * If false, put the tail of an error message in the string buffer. The
      * string buffer will be initialized with the name of the index.
      *
      * @param expr The expression to check
      * @param msg  The StringBuffer to pack with the error message tail.
      * @return true iff the expression can be part of an index.
      */
-    private boolean isIndexableExpression(StringBuffer msg) {
+    private boolean validateIndexingMVsGaurds(StringBuffer msg) {
```
wolffcm: I think this method could have a better name... "validateExprForMVsAndIndexes" maybe?  Note that "guard" is misspelled here too.

hhakimi53: ok, I will change the name to validateExprForMVsAndIndexes().

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1788,7 +1789,15 @@ private static void addIndexToCatalog(Database db,
                     AbstractExpression expr = dummy.parseExpressionTree(exprNode);
                     expr.resolveForTable(table);
                     expr.finalizeValueTypes();
-                    // We will check this for validity later.
+                    // string will will be populated expression details whose value
+                    // type are not indexable
+                    StringBuffer exprMsg = new StringBuffer();
+                    if (!expr.isValueTypeIndexable(exprMsg)) {
+                        // indexing on expression with boolean result is not supported.
+                        msg.append("can't include " + exprMsg + ".");
```
wolffcm: I think I'd prefer a friendlier message like "Cannot create index `<index name>` because it contains `<exprMsg>`, which is not supported."

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1803,7 +1812,7 @@ else if (subNode.name.equals("predicate")) {
         }
 
         // Check all the subexpressions we gathered up.
-        if (!AbstractExpression.areIndexableExpressions(checkExpressions, msg)) {
+        if (!AbstractExpression.validateIndexingMVsGaurds(checkExpressions, msg)) {
```
wolffcm: Is the call to `colType.isIndexable` at line 1836 now redundant as well?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2673,11 +2677,27 @@ private static void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt,
...
+            // for other gaurds on all of them together
             checkExpressions.add(outcol.expression);
         }
-        // Now, the display list must have a count(*).
-        AbstractExpression coli = stmt.m_displayColumns.get(i).expression;
-        if (coli.getExpressionType() != ExpressionType.AGGREGATE_COUNT_STAR) {
+
+        // check for count star in the display list
+        boolean countStarFound = false;
+        if (i < displayColCount) {
```
wolffcm: I'm trying to understand how this code changes behavior.  If `i` is not less than `displayColCount` then there are no aggregate expressions, including `count(*)`, right?

src/frontend/org/voltdb/expressions/ConjunctionExpression.java
```diff
@@ -66,4 +66,12 @@ public String explain(String impliedTableName) {
             m_right.explain(impliedTableName) + ")";
     }
 
+    @Override
+    public boolean isValueTypeIndexable(StringBuffer msg) {
```
wolffcm: I guess logical expressions don't produce a boolean value?  Is that why you can't just check for a boolean output type in AbstractExpression?

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -2844,63 +2858,107 @@ public void testCreateTableWithGeographyType() throws Exception {
 
     public void testGeographyNegative() throws Exception {
 
+        String ddl = "create table geogs ( geog geography not null );\n" +
```
wolffcm: Refactoring this test in this way will cause a lot of conflicts with my geography size specifier changes. I guess I'd better make sure I submit before you.  :)

wolffcm:
Looks good overall.  I think if you can improve the create index error message, it's good to go once it passes Jenkins.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: Yes, right. If 'i' is not less than displayColCount, the provided # col in display columns and group by columns (if any) are same and there after logic was trying to seek at an index which is greater thn array size by 1 and results in out of bound exception. Check helps in preventing that.

src/frontend/org/voltdb/expressions/ConjunctionExpression.java
hhakimi53: Yes, in here also the value type is set to big int and not a boolean. So had to work around it this way.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
hhakimi53: :-)

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: I will try for it.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: It's needed in here as for column types when operating at elements from volt xml element, they are not treated as expressions. So for the elements which are not expression, logic to evaluates if it's column is indexeable or not, is still needed. It would have been nice if we could have consolidated this evaluation at a single place.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: Done. Updated msg from "Index <index name> can't include function <expression msg>." to "Cannot create index <index name> because it contains <exprMsg >, which is not suppored."

hhakimi53:
Jenkins run is blue for all except for two tests which were known issues in master.

PR #3148, opened 2016-01-08
ENG-9696: Allow Best CommandLog Selection even when there is only a s…

wweiss-voltdb:
…ingle log segment


afrogers:
Merged externally.

PR #3152, opened 2016-01-12
Work around new runtime conversion restrictions

paulmartel:
Changing keyspace to VARBINARY makes it compatible with the query inputs.
Converting Scan's rangeMin parameter to a java String makes it compatible but may hurt the benchmark numbers. If that's the case, we may want to define and use a UTF8INTERPRET sql function to do the UTF8 VARBINARY to VARCHAR "conversion" more efficiently in C++ where it is a trivial shallow copy into an NValue with a different type field setting.

### DIFF
--- a/tests/test_apps/ycsb/src/com/procedures/Scan.java
+++ b/tests/test_apps/ycsb/src/com/procedures/Scan.java
@@ -22,11 +22,15 @@
  */
 package com.procedures;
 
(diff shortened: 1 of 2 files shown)


afrogers:
Since Scan is already more expensive than the others, my guess is that the impact of one extra Java string construction won't be felt too badly. This looks good to me.

PR #3153, opened 2016-01-12
Fail on ambiguous column references.

BillWhite:
If a column reference in a select list or a join or order by expression
is ambiguous it should be an error.  Such an ambiguity can arise when
two columns in two join tables have the same name.

https://issues.voltdb.com/browse/ENG-9367


tests/frontend/org/voltdb/planner/TestPlansJoin.java
```diff
@@ -122,15 +122,19 @@ public void testBasicThreeTableInnerJoin() {
         assertTrue(n.getChild(1) instanceof SeqScanPlanNode);
         assertEquals(1, pn.getOutputSchema().getColumns().size());
 
-        pn = compile("select C FROM R1 INNER JOIN R2 USING (C), R3 WHERE R1.A = R3.A");
+        pn = compile("select C FROM R1 INNER JOIN R2 USING (C), R3_NOC WHERE R1.A = R3_NOC.A");
```
wolffcm: Whoa. I didn't know about the USING clause before, but now I do.

tests/frontend/org/voltdb/planner/TestPlansJoin.java
```diff
@@ -122,15 +122,19 @@ public void testBasicThreeTableInnerJoin() {
...
         n = pn.getChild(0).getChild(0);
         assertTrue(n instanceof NestLoopPlanNode);
         assertTrue(n.getChild(0) instanceof NestLoopIndexPlanNode);
         assertTrue(n.getChild(1) instanceof SeqScanPlanNode);
         assertEquals(1, pn.getOutputSchema().getColumns().size());
-
-        failToCompile("select C, R3.C FROM R1 INNER JOIN R2 USING (C) INNER JOIN R3 ON R1.C = R3.C",
+        // Here, the USING(C) causes R1.C to be hidden from lookup.
```
wolffcm: Hmm, I don't understand why r1.c is hidden.  But maybe I don't understand USING.

wolffcm:
It would be good to touch base with John C. and see if sqlcoverage will pass with the offending patterns enabled (the ones that found this bug).

tests/frontend/org/voltdb/planner/TestPlansJoin.java
```diff
@@ -1210,6 +1214,29 @@ public void testMoreThan5TableJoins() {
...
+       // Ambiguous references in group by expressions.
+       failToCompile("select NOTC from R1, R3_NOC group by A;", "Column \"A\" is ambiguous");
+       failToCompile("select NOTC from R1, R3_NOC group by sqrt(A);", "Column \"A\" is ambiguous");
+       // Ambiguous references in subqueries.
+       failToCompile("select ALPHA from (select SQRT(A) as ALPHA from R1) as S1, (select SQRT(C) as ALPHA from R1) as S2;",
+                     "Column \"ALPHA\" is ambiguous");
+       failToCompile("select ALPHA from (select SQRT(A), SQRT(C) from R1, R3) as S1, (select SQRT(C) as ALPHA from R1) as S2;",
+                     "Column \"A\" is ambiguous");
+   }
+
```
wolffcm: I'm curious about subqueries---I think that the SQL standard says that column references in a nested query can resolve to a table in the nested query even if the outer query has a column of the same name.  In other words, I think this should be okay:
```sql
-- Both tables T and R have column called A
select * 
from T
where T in (select A from R); -- A is not ambiguous here
```
We should test for this and make sure we're not breaking any valid SQL queries (if I'm right about how resolution works for subqueries, which I'm not sure I am).

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,204 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
         if (tableName == null) {
             Expression e = rangeVar.getColumnExpression(columnName);
 
             if (e != null) {
-                opType   = e.opType;
-                nodes    = e.nodes;
-                dataType = e.dataType;
-
-                return true;
+                return new ColumnReferenceResolution(e);
```
wolffcm: Does this happen for a derived table (subquery in the FROM clause, with an expression on the SELECT list)?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,204 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+                ExpressionColumn ec = (ExpressionColumn)m_expr;
+                if (ec.columnName != null) {
+                    return ec.columnName.hashCode();
+                } else {
+                    return ec.hashCode();
+                }
+            } else if (m_rangeVariable != null && 0 <= m_colIndex) {
+                return 31*m_rangeVariable.hashCode() + 131*m_colIndex;
+            } else {
+                return 0;
```
wolffcm: Is this line reachable?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,204 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+            } else if (!m_expr.equals(other.m_expr))
+                return false;
+            if (m_rangeVariable == null) {
+                if (other.m_rangeVariable != null)
+                    return false;
+            } else if (!m_rangeVariable.equals(other.m_rangeVariable))
+                return false;
+            return true;
+        }
+        private ExpressionColumn getOuterType() {
```
wolffcm: This method name had me kind of confused.  Since this is not a static inner class, ExpressionColumn.this retrieves the instance of the outer class, I guess.  I can't think of a better name though.

wolffcm:
This looks promising.  I had a questions about resolution and subqueries, that I think we should understand before we can call this fixed.

tests/frontend/org/voltdb/planner/TestPlansJoin.java
BillWhite: The relevant part of the SQL 2003 standard is section 7.7, paragraph 7.  It's surprisingly lucid for the SQL standard.

tests/frontend/org/voltdb/planner/TestPlansJoin.java
BillWhite: I think these are different cases.  The first subquery test fails because the output columns are both named "ALPHA".  The SQL compiler doesn't know which ALPHA is wanted.  This is equivalent to having two tables both with ALPHA columns.  The second subquery test fails because the subquery S1 itsels is ambiguous, independent of where it's embedded.  I wanted to test that we dove into subqueries in deeply.

BillWhite:
Somehow the github branch became incompatible with my dev branch. Since the changes I want are all on my dev branch I am closing this pull request, deleting the github branch and pushing this again.

PR #3156, opened 2016-01-12
+ Guards to DR and Export for Geo types

hhakimi53:
Add guards to prevent DR/export against table containing geo types.

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -182,11 +182,13 @@ public class DDLCompiler {
     }
 
     private static boolean isVoltTypePresent(VoltXMLElement tableNode, VoltType lookupType, StringBuffer msg) {
-        assert tableNode.name.equals("table");
+        assert (tableNode.name.equals("table"));
 
         String valueTypeStr;
         VoltType type;
         String columnName;
+        boolean typeDetected = false;
+        String separatorStr;
 
         for (VoltXMLElement subNode : tableNode.children) {
             if (subNode.name.equals("columns")) {
@@ -196,15 +198,46 @@ public class DDLCompiler {
                         columnName = columnNode.attributes.get("name");
                         type = VoltType.typeFromString(valueTypeStr);
                         if (lookupType == type) {
-                            msg.append("column name: '" + columnName + "', type: '" + valueTypeStr + "'");
-                            return true;
+                            if (msg != null) {
+                                separatorStr = (typeDetected == true)? ", " : " ";
+                                msg.append(separatorStr + "column name: '" + columnName + "' type: '" + valueTypeStr + "'");
+                            }
+                            typeDetected = true;
                         }
                     }
                 }
             } // if (subNode.name.equals("columns"))
         }
 
-        return false;
+        return typeDetected;
+    }
+
+    private static boolean isVoltTypePresent(VoltXMLElement tableNode, VoltType lookupType) {
+        return isVoltTypePresent(tableNode, lookupType, null);
+    }
+
+    // Function to check if the table - TableXML, supplied as VoltXMLElement, does not contain geo types.
+    // If table does contain geo types, VoltCompilerException is generated with information about geo columns
+    // found in table wrapped in exception message
+    private void guardForGeoColumns(VoltXMLElement tableXML, String tableName, String tableVerb)
+            throws VoltCompilerException {
+        assert (tableXML.name.equals("table"));
+        assert(tableVerb != null);
+        assert(tableName != null);
+
+        boolean pointColDetected = false;
+        StringBuffer pointMsg = new StringBuffer();
+        pointColDetected = isVoltTypePresent(tableXML, VoltType.GEOGRAPHY_POINT, pointMsg);
+
+        boolean polygonColDetected = false;
+        StringBuffer polygonMsg = new StringBuffer();
+        polygonColDetected = isVoltTypePresent(tableXML, VoltType.GEOGRAPHY, polygonMsg);
+        //
+        if (pointColDetected || polygonColDetected) {
+            String separatorStr = (pointColDetected && polygonColDetected) ? "," : "";
+            throw m_compiler.new VoltCompilerException("Can't " + tableVerb + " table '" + tableName.toUpperCase() + "' containing geo type column(s) -" +
+                    pointMsg + separatorStr + polygonMsg + ".");
+        }
     }
 
     /**
@@ -891,12 +924,7 @@ public class DDLCompiler {
             VoltXMLElement tableXML = m_schema.findChild("table", tableName.toUpperCase());
 
             if (tableXML != null) {
-                StringBuffer msg = new StringBuffer();
-                msg.append("Can't EXPORT table '" + tableName.toUpperCase() + "' containing geo type column(s) - ");
-                if (isVoltTypePresent(tableXML, VoltType.GEOGRAPHY, msg) ||
-                    isVoltTypePresent(tableXML, VoltType.GEOGRAPHY_POINT, msg)) {
-                    throw m_compiler.new VoltCompilerException(String.format(msg + "."));
-                }
+                guardForGeoColumns(tableXML, tableName, EXPORT);
 
                 if (tableXML.attributes.containsKey("drTable") && tableXML.attributes.get("drTable").equals("ENABLE")) {
                     throw m_compiler.new VoltCompilerException(String.format(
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -181,6 +181,32 @@ public DDLCompiler(VoltCompiler compiler,
...
+                        valueTypeStr = columnNode.attributes.get("valuetype");
+                        columnName = columnNode.attributes.get("name");
+                        type = VoltType.typeFromString(valueTypeStr);
+                        if (lookupType == type) {
+                            msg.append("column name: '" + columnName + "', type: '" + valueTypeStr + "'");
+                            return true;
+                        }
+                    }
+                }
+            } // if (subNode.name.equals("columns"))
```
BillWhite: Is it possible to return false here?  If you find a "columns" element, won't that be the only one?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -181,6 +181,32 @@ public DDLCompiler(VoltCompiler compiler,
         m_classLoader = classLoader;
     }
 
+    private static boolean isVoltTypePresent(VoltXMLElement tableNode, VoltType lookupType, StringBuffer msg) {
+        assert tableNode.name.equals("table");
```
BillWhite: It seems a little weird to me to have a string buffer parameter in a predicate like this.  Will this make it hard to use this function in another context, one in which you don't need the stringbuffer?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -862,7 +889,15 @@ else if (matchStatus == ClassNameMatchStatus.NO_WILDCARD_MATCH) {
                     Constants.DEFAULT_EXPORT_CONNECTOR_NAME;
 
             VoltXMLElement tableXML = m_schema.findChild("table", tableName.toUpperCase());
+
             if (tableXML != null) {
+                StringBuffer msg = new StringBuffer();
+                msg.append("Can't EXPORT table '" + tableName.toUpperCase() + "' containing geo type column(s) - ");
+                if (isVoltTypePresent(tableXML, VoltType.GEOGRAPHY, msg) ||
```
BillWhite: What if the table has both points and polygons.  Do you want both of them in the error message?  What about the column names?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -862,7 +889,15 @@ else if (matchStatus == ClassNameMatchStatus.NO_WILDCARD_MATCH) {
...
+                    isVoltTypePresent(tableXML, VoltType.GEOGRAPHY_POINT, msg)) {
+                    throw m_compiler.new VoltCompilerException(String.format(msg + "."));
```
BillWhite: I don't think calling String.format is doing anything for you here.  If the msg has a percent sign in it, which seems unlikely, it will probably not do what you expect, and may cause an exception.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -894,8 +929,18 @@ else if (matchStatus == ClassNameMatchStatus.NO_WILDCARD_MATCH) {
...
+            //System.out.println("\n\n" + m_schema.toString());
+
             VoltXMLElement tableXML = m_schema.findChild("table", tableName.toUpperCase());
             if (tableXML != null) {
+                StringBuffer msg = new StringBuffer();
+
+                msg.append("Can't DR table '" + tableName.toUpperCase() + "' containing geo type column(s) - ");
+                if (isVoltTypePresent(tableXML, VoltType.GEOGRAPHY, msg) ||
+                    isVoltTypePresent(tableXML, VoltType.GEOGRAPHY_POINT, msg)) {
+                    throw m_compiler.new VoltCompilerException(String.format(msg + "."));
```
BillWhite: Same thing here with String.format.

BillWhite:
That's all I have.  It's all pretty good, except for some small structuring issues.

I would ditch the calls to String.format, at the very least.  The other issues are a matter of taste.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: The error message it will convey information that geo type is not supported, so that should be a good hint to user about what is wrong with the ddl. In case if multiple geo columns are present, it will flag the first (geography value first and than geography point next) column along with column name and column type (column name and column type info is  captured in string buffer passed on isVoltTypePresent()).

For example (we have similar junit test case for it):
for ddll:
create table geogs ( id integer primary key, point1 geography_point , region1 geography NOT NULL);
drr table geogs;

Error Message outputted
Can't DR table 'GEOGS' containing geo type column(s) - column name: 'REGION1', type: 'GEOGRAPHY'.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: Ok, will take format out. thnxs.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: We can return false from here but as the logic is only iterating on columns attribute, so felt better to have it at the end instead of returning.

src/frontend/org/voltdb/compiler/DDLCompiler.java
BillWhite: That's what I thought, but I'm wondering why you can't get an error message with point1 and region1 both listed.

Not that it matters all that much.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: Ok, I will update the logic so that if user wants to use the function without having to pass in stringbuffer, they can do it.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: Did it this way to be consistent with the approach we have in another place - if ddl has multiple errors, it flags the first one. I can update the logic to capture all geo columns if that seems more user friendly.

BillWhite: We only get one error message per compilation.  If there were N errors in the table creation statement that might mean N+1 compilations, one for each error and one for the success.  I have used compilers that were like that, and found it very annoying.

But you choose.

src/frontend/org/voltdb/compiler/DDLCompiler.java
BillWhite: I guess there will not be many top-level elements, so it's not that expensive.  I would exit early, though, and save a few times through the loop.

It probably doesn't matter that much.  Compiling SQL is not really a performance path.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: ok, I have started updating the logic so that reported message contains all geo columns in the ddl statement.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -181,6 +181,65 @@ public DDLCompiler(VoltCompiler compiler,
...
+                    }
+                }
+            } // if (subNode.name.equals("columns"))
+        }
+
+        return typeDetected;
+    }
+
+    private static boolean isVoltTypePresent(VoltXMLElement tableNode, VoltType lookupType) {
+        return isVoltTypePresent(tableNode, lookupType, null);
```
BillWhite: Is this ever called?  If not, maybe it's not doing anything for us, and we should delete it.

BillWhite:
That's all I have.  I have only one small comment, which you are welcome to ignore.

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: No, it's not called at present. I had put it in there so that if user wants can use this method with having to pass in Stringbuffer parameter based on earlier review feedback :)

BillWhite: I know that.  I think doing it this way, with an overload, confuses things.  I would go back to the original way, by deleting this overload.  I withdraw my original comment.

hhakimi53: ok, I will remove the overloaded function as it's not used and current implementation does not restrict user to pass in valid string buffer parameter.

PR #3157, opened 2016-01-13
Eng 9497 sqlcov geo func

johnTcrawford:
Tests of Geo accessor functions, added to SQLCoverage. Before merging, I need to investigate a few Jenkins failures, notably in the advanced-matview-join test suite of SQLCoverage, which I probably broke (& a couple JUnit failures that are probably just intermittent). But meanwhile, this can be reviewed.


src/frontend/org/voltdb/PostGISBackend.java
```diff
@@ -196,10 +268,17 @@ static public String transformDML(String dml) {
...
+                                        transformNumInteriorRingsQuery(
+                                            transformIsValidQuery(
+                                                transformIsInvalidReasonQuery(
+                                                    transformCastGeoAsVarcharQuery(
+                                                        PostgreSQLBackend.transformDML(dml)
+               ))   )   )   )   )   )   )   )   )   )
                 // TODO: make these more robust, using regex Patterns?
                 .replace("pointFromText('POINT",     "ST_GeographyFromText('POINT")
                 .replace("polygonFromText('POLYGON", "ST_GeographyFromText('POLYGON")
+                .replace("AsText",    "ST_AsText")
```
wolffcm: This seems a little fragile, since other variations in case will not be replaced with their PostGIS equivalents.  But for now perhaps it does not matter.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -300,7 +321,7 @@ static private String handleParens(String group, String prefix, String suffix) {
     @SuppressWarnings("unused")
     static protected String transformQuery(String query, Pattern queryPattern, String initText,
```
wolffcm: This method has 11 arguments, that last of which is a vararg?  The call sites for this method will tend to be difficult to read.  One could imagine (at some point, not now) employing a design pattern to make this easier to read and maintain, like maybe a `QueryTransformer` class that can be constructed using the builder pattern.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -342,7 +363,9 @@ static protected String transformQuery(String query, Pattern queryPattern, Strin
                 }
             }
             if (useWhole) {
-                if (intOnly && !isIntegerColumn(group)) {
+                if (columnType != null && (
+                           (columnType.equals(ColumnType.INTEGER) && !isIntegerColumn(group))
```
wolffcm: `columnType` has an enum type here, so you could compare with a simple `==` operator, right?  I would find it a little easier to read.

tests/scripts/examples/sql_coverage/advanced-template.sql
```diff
@@ -11,14 +13,10 @@
 -- {@dmltable = "_table"}
 -- {@fromtables = "_table"}
 -- {@optionalfn = "_numfun"}
+-- {@star = "*"}
```
wolffcm: My interpretation of the changes you are making to this file is that you need to redefine some of these symbols in the geo-specific queries so that they can actually run and/or have predictable results.  Do I have that right?

tests/scripts/examples/sql_coverage/configure-for-geo-point.sql
```diff
@@ -0,0 +1,46 @@
...
+{@idcol = "ID"}
+
+--TODO: this should be randomized:
+--{@insertvals = "_id, pointFromText('POINT(0 0)'), pointFromText('POINT(-71.0 42.0)'), null, polygonFromText('POLYGON((-1 -1, 1 -1, 1 1, -1 1, -1 -1))')"}
+{@insertvals = "_id, pointFromText('POINT(0 0)'), pointFromText('POINT(-71.0 42.0)'), null, polygonFromText('POLYGON((-1 -1, 1 -1, 1 1, -1 1, -1 -1))')"}
+--{@insertvals = "_id, pointFromText('POINT(_value[decimal:-1,1] _value[decimal:-1,1])'), null, pointFromText('POINT(_value[decimal:-1,1] _value[decimal:-1,1])'), polygonFromText('POLYGON((_value[decimal:-0.9,-0.3] _value[decimal:-0.9,-0.3], _value[decimal:0.3,0.9] _value[decimal:-0.9,-0.3], _value[decimal:0.3,0.9] _value[decimal:0.3,0.9], _value[decimal:-0.9,-0.3] _value[decimal:0.3,0.9], _value[decimal:-0.9,-0.3] _value[decimal:-0.9,-0.3]))')"}
+
+{@onefun = ""}  -- There are no handy unary point-to-point or polygon-to-polygon functions.
+{@optionalfn = "_geofun"}
+-- TODO: perhaps change this to use "AsText" (not LONGITUDE, LATITUDE, AREA), once fully supported (?);
```
wolffcm: `asText` should work now on master.

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -1,72 +1,222 @@
...
-SELECT ID, AREA(POLY1)                from _table G33 WHERE AREA(POLY1) > 2000000
-SELECT ID, AREA(POLY1), asText(POLY1) from _table G34
-SELECT ID, AREA(POLY1), asText(POLY1) from _table G35 WHERE AREA(POLY1) > 2000000
-SELECT ID, asText(CENTROID(POLY1))    from _table G36
-SELECT ID, LONGITUDE(PT1), LATITUDE(PT1), LONGITUDE(PT2), LATITUDE(PT2), LONGITUDE(CENTROID(POLY1)), LATITUDE(CENTROID(POLY1)) from _table G37
-
--- Test the CONTAINS function, with polygons & points (also using LONGITUDE, LATITUDE, asText)
+SELECT ID, DISTANCE(POLY1,POLY1) from _table G27
+SELECT ID, DISTANCE(POLY1,POLY1) from _table G28 WHERE DISTANCE(POLY1,POLY1) < 200000
+SELECT A.ID AID, B.ID BID, DISTANCE(A.POLY1,B.POLY1) G29DIST FROM _table A JOIN _table B ON A.ID <= B.ID
```
wolffcm: I would expect this to fail on VoltDB and produce a result in PostGIS?

tests/scripts/examples/sql_coverage/geo-advanced-point.sql
```diff
@@ -0,0 +1,3 @@
+-- Run the advanced-template against DDL with (Geographic) point type
+<configure-for-geo-point.sql>
+<advanced-template.sql>
```
wolffcm: I guess we'll be adding some geo-specific definitions and templates here later?

wolffcm:
This looks good to me.  Thanks for bashing on this.  I had some (future) suggestions and questions, but nothing major.  If we ran these suites in Jenkins, how close are we to passing?

src/frontend/org/voltdb/PostGISBackend.java
johnTcrawford: Agreed. As you may have noticed, there's a "TODO" comment suggesting I should fix this, but it's not my highest priority right now, so I think I'll leave it to a later task (e.g. ENG-9499º.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: Yeah, you raise a good point. I would never have done that with a public class, but I justified it initially as being OK in a private method (but now, protected - albeit with only one sub-class and unlikely to get more). Though, as you suggest, and like the one above, this is not my priority right now. Perhaps I'll write a clean-up ticket for these kinds of things, to do after 6.0 goes out.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: Yes. Done.

tests/scripts/examples/sql_coverage/advanced-template.sql
johnTcrawford: Yes, exactly. Many test suites "include", for example, basic-select.sql, which (now) contains a bunch of query patterns "SELECT @star ...", which will almost only be "SELECT * ..." but that won't work for Geo (for 2 reasons, one temporary, related to the Python client not working, and one not, related to PostGIS returning WTB by default, which is different from VoltDB). So, for the Geo test suites only, @star instead gets defined as a specific list of columns (which will mostly use AsText, once that works better).

tests/scripts/examples/sql_coverage/configure-for-geo-point.sql
johnTcrawford: Yes, though I still need to improve SQLCoverage's "normalizer", to recognize VARCHAR columns that start with "POINT (..." or "POLYGON (..." and then round-off the numbers therein to some appropriate number of significant digits; until then, LONGITUDE, LATITUDE & AREA will better compare with PostGIS results, since they already use the "fuzzy" normalizer that already applies to FLOAT columns.

tests/scripts/examples/sql_coverage/geo-advanced-point.sql
johnTcrawford: I think this will largely just work: apply the configure-for-geo-point.sql definitions, already used in geo-basic-point, to the advanced-template.sql, which already contains query patterns used by various other "advanced" test suites (e.g. advanced-ints, advanced-strings). There will probably be some issues (e.g. the "normalizer" one I mentioned above), which is the work of ENG-9499, but it won't be starting from scratch. Same goes for geo-advanced-polygon.sql, below.

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: No, they both fail in the (VoltDB) planner; but I added a comment, since I think you asked roughly the same question last time - so this obviously needs clarification.

wolffcm: Sorry for repeating myself.  I'm always forgetting that we use our planner even when we're evaluating with a different back end.

johnTcrawford: No problem! I should have added the clarifying comment the first time.

johnTcrawford:
Chris, in answer to your "how close are we to passing" question: when I run it on my machine, I get 30 failures, all in geo-basic (out of 304 queries there, 20408 total) - geo-basic-point & geo-basic-polygon completely pass (albeit with some invalid queries that could be improved, but a lot less - only around 3% - since I split these up, and thereby stopped trying to compare points with polygons). Of the 30, most are issues with AsText or CAST, involving format (extra ".0") and/or round-off (see above); there's also what looks like a PostGIS bug, affecting a couple of CENTROID and CONTAINS queries, for one very large polygon (1/8 of the Earth). But all of that is with the PRECISION parameter of the "fuzzy" normalizer set to 3 (which I have not checked in; normal is 12), which is a value I set earlier, when it was needed (without that, I would have many more failures); we'll see, after I merge to master, how many digits agree, since Huzefa's improvements (ENG-9471/9713).

PR #3162, opened 2016-01-13
Update  geo function name to be consistent with geo type names

hhakimi53:
With changes verified TestGeospatialFunction regression test passes and documentation are picked up in java docs ("ant -f mmt.xml dist.pro check"). There were no changes needed in geospatial demo apps. For sanity purpose verified, demo app runs fine with the changes.

To do: have to ping Andrew about the changes so that any documentation around it can be updated.

### DIFF
--- a/src/frontend/org/voltdb/VoltTableRow.java
+++ b/src/frontend/org/voltdb/VoltTableRow.java
@@ -297,7 +297,7 @@ public abstract class VoltTableRow {
             ret = getDecimalAsBigDecimal(columnIndex);
             break;
         case GEOGRAPHY_POINT:
-            ret = getPoint(columnIndex);
+            ret = getGeographyPointValue(columnIndex);
             break;
         case GEOGRAPHY:
             ret = getGeographyValue(columnIndex);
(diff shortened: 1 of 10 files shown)


src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -620,18 +620,45 @@ public final String getString(String columnName) {
         return getVarbinary(colIndex);
     }
 
-    public final GeographyPointValue getPoint(int columnIndex) {
+    /**
+     * Retrieve the GeographyPointValue value stored in the column specified by index.
+     * Looking at the return value is not a reliable way to check if the value is
+     * <tt>null</tt>. Use {@link #wasNull()} instead.
```
wolffcm: Actually in this case, a null pointer will be returned back reliably for null values.  But wasNull should work too.

wolffcm: I see that getString has a similar comment, but in that case also I believe we can check for a null pointer.  Hmm.

wolffcm:
Looks good.  I had some questions about the accuracy of comments about wasNull in the javadoc.  Maybe we can address them as part of the javadoc review.

PR #3164, opened 2016-01-14
Fail on ambiguous column references.

BillWhite:
If a column reference in a select list or a join or order by expression
is ambiguous it should be an error. Such an ambiguity can arise when
two columns in two join tables have the same name.

https://issues.voltdb.com/browse/ENG-9367


tests/frontend/org/voltdb/planner/TestPlansInExistsSubQueries.java
```diff
@@ -690,7 +690,7 @@ public void testExistsSimplification() {
         }
         {
             //EXISTS => FALSE, join predicate is retained
-            AbstractPlanNode pn = compile("select a from r1 join r2 on exists " +
+            AbstractPlanNode pn = compile("select r1.a from r1 join r2 on exists " +
```
wolffcm: I still don't see the example from the question I asked in my last review:
```sql
select * from T where A in (select A from R);
```
It seems like an important test case to me.  I grabbed your branch and it seems to produce the correct plan though.

BillWhite: Ok, I'll add this case.  I think it's a scoping issue, and not an ambiguous
column issue, though.

The A in the subquery is in the from clause of the subquery.  The A in the
containing query is from the from clause of the containing query.  They are
different scopes.

wolffcm: Fair enough.  Thanks for adding the test.  In other SQL engines I've worked on (okay, *the* other engine) column resolution was handled in a differently and cases like this were a frequent source of bugs.

wolffcm:
What changed between this and your last pull request?

BillWhite:
Mostly the changes were in ExpressionColumn.ColumnReferenceResolution.  I
changed the way the hashCode and equals functions work.

We currently say two potential column resolutions are equal if they have
the same name.  An expression's name is its alias or it's column name if it
has no alias.  A range variable's name is the range variable's alias, if
there is one, or the name of the range variable's table.  We only calculate
hash values of the names.  If we can't find a name, the hashcode is the
hashcode of the expression or range variable, and equals always returns
false.

Perhaps we should throw an exception of some sort if we can't find a name.
But that's not what I did.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,254 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+                    sb.append(String.format("Column \"%s\" is ambiguous.  It's in tables: ", columnName));
+                    String sep = "";
+                    for (ColumnReferenceResolution crr : foundInRangeVariables) {
+                        if (crr.isExpression()) {
+                            if (usingSb.length() == 0) {
+                                usingSb.append("USING(");
+                            } else {
+                                usingSb.append(", ");
+                            }
+                            usingSb.append(crr.toString());
```
paulmartel: This crr.isExpression() block could be made simpler and more consistent with the else block as follows.  Outside the loop: `String usingSep = "USING(";`
Replace the `if` block w/:

        usingSb.append(usingSep).append(crr.toString());
        usingSep = ", ";

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,254 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+            m_colIndex = -1;
+        }
+        public void finallyResolve() {
+            if (m_rangeVariable != null && m_colIndex != -1) {
+                setAttributesAsColumn(m_rangeVariable, m_colIndex);
+            } else if (m_expr != null) {
+                opType   = m_expr.opType;
+                nodes    = m_expr.nodes;
+                dataType = m_expr.dataType;
+            }
```
paulmartel: Add: else assert(false)?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,254 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+     * until we are more sure of the reference.
+     *
+     * We store these in a java.util.HashSet.  So we need to have our own
+     * notion of equality.
+     */
+    private class ColumnReferenceResolution {
+        RangeVariable m_rangeVariable;
+        int           m_colIndex;
+        Expression    m_expr;
+        public ColumnReferenceResolution(RangeVariable rangeVariable, int colIndex) {
```
paulmartel: can we assert rv not null and colIndex != -1?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,254 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+            m_rangeVariable = rangeVariable;
+            m_colIndex = colIndex;
+            m_expr = null;
+        }
+        public boolean isExpression() {
+            return m_expr != null;
+        }
+        public ColumnReferenceResolution(Expression expr) {
+            m_expr = expr;
```
paulmartel: assert expr not null?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -340,72 +357,254 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+     * is named by a range variable.  We keep the expression or the
+     * range variable/column index pair here.  We may have several resolutions
+     * if the column reference is ambiguous.  We can't actually commit to one
+     * until we have examined all of them.  So, we defer changing this object
+     * until we are more sure of the reference.
+     *
+     * We store these in a java.util.HashSet.  So we need to have our own
+     * notion of equality.
+     */
+    private class ColumnReferenceResolution {
```
paulmartel: Can this class be declared static?
Rather than branching in each method (and also sometimes in the caller), would it not be simpler to have an abstract base class (or interface) w/ two concrete implementations?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -332,7 +340,16 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
...
+                // Look in all the range variables.  We may
+                // find this column more than once, and that
+                // would be an error See ENG-9367.
+                //
+                // Note that we can't actually commit to a resolution
+                // until we have looked everywhere.  This means we need to
+                // store up potential resolutions until we have looked at all
+                // the range variables.  If we find just one, we finally
+                // resolve it. below.
+                java.util.Set<ColumnReferenceResolution> foundInRangeVariables = new java.util.HashSet<ColumnReferenceResolution>();
```
paulmartel: Use import (now that we "own" our hsql code) and the diamond operator for readability.

paulmartel:
End of review round.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: I refactored this in a different way, which does about the same thing, but which orders the USING list, if any, first, and then lists the tables.  This is kind of the order the SQL specification mandates, though that is for searching and this is just listing them.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: This can't be static, because finallyResolve needs to change the containing classes fields.  It started out as static, but it evolved.

Since it can't be static, it's a little hard to write a factory function.  In any case, the complexity is mostly in getting the name.  So, I collapsed the two functions getExprName and getRangeVariableName into getName, where all the iffing and elseing is done.

PR #3165, opened 2016-01-14
ENG-9620 memory monitoring default to 80%

dweiss88:


tests/frontend/org/voltdb/TestMemoryResourceMonitor.java
```diff
@@ -124,14 +124,13 @@ public void testGoodRssValues() throws Exception {
         }
     }
 
-    public void testNoRssLimit() throws Exception
+    public void testDefaultRssLimit() throws Exception
```
akhanzode: how about adding test where at 79% you are still running.

akhanzode:
If jenkins is happy push to master.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -912,6 +913,16 @@ public static void populateDefaultDeployment(DeploymentType deployment) {
             tt = new SystemSettingsType.Temptables();
             ss.setTemptables(tt);
         }
+        ResourceMonitorType rm = ss.getResourcemonitor();
```
vtkstef: should you not instantiate only if there is a legitimately defined resource limit? i.e. if I configure 0 for memory limit why should you instantiate it?

PR #3167, opened 2016-01-15
Unit tests for geo types and VoltTable

wolffcm:
Here's some unit tests for this class.  I found a couple of bugs: wasNull wasn't working right for polygons, and getRaw wasn't implemented for points.

I could have sworn that during a meeting we found a place where the geo types were missing from a `get` method.  I could not find it.  Maybe it was added sometime in the last few weeks?

### DIFF
--- a/src/frontend/org/voltdb/VoltTableRow.java
+++ b/src/frontend/org/voltdb/VoltTableRow.java
@@ -434,6 +434,7 @@ public abstract class VoltTableRow {
         case TIMESTAMP:
         case FLOAT:
         case DECIMAL:
+        case GEOGRAPHY_POINT:
             // all of these types are fixed length, so easy to get raw type
             retval = new byte[length];
             m_buffer.position(offset);
@@ -668,6 +669,7 @@ public abstract class VoltTableRow {
             return null;
         }
 
+        m_wasNull = false;
         offset += 4;
         GeographyValue gv = GeographyValue.unflattenFromBuffer(m_buffer, offset);
         return gv;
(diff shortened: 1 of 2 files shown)


hhakimi53:
Changes look good.

src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -434,6 +434,7 @@ final ByteBuffer getRawRow() {
         case TIMESTAMP:
         case FLOAT:
         case DECIMAL:
+        case GEOGRAPHY_POINT:
```
paulmartel: This has been "broken" forever, but
`length = type.getLengthInBytesForFixedTypesWithoutCheck();`
is called needlessly for variable-length types and should be moved under these (fixed-length-type) switch cases.

tests/frontend/org/voltdb/TestVoltTable.java
```diff
@@ -548,6 +555,74 @@ public void testFloats() {
...
+        assertTrue(vtCopy.advanceRow());
+
+        String wkt = GEOG_PT_VALUE.toString();
+        assertEquals(wkt, vt.getGeographyPointValue(0).toString());
+        assertEquals(wkt, vtCopy.getGeographyPointValue(0).toString());
+        assertEquals(wkt, vt.getGeographyPointValue("pt").toString());
+        assertEquals(wkt, vt.get(0, VoltType.GEOGRAPHY_POINT).toString());
+
+        byte[] raw = vt.getRaw(0);
+        assertEquals(GeographyPointValue.getLengthInBytes(), raw.length);
```
paulmartel: Can we validate that the raw forms of the point values are byte-wise equal between vt and vtCopy?

paulmartel:
end of my comments.

wolffcm:
Thanks, Paul.  I added a commit with your suggestions.

PR #3169, opened 2016-01-15
Vdm 11

sujmanshak:


lib/python/vdm/main.py
```diff
@@ -69,4 +70,6 @@
 except (IOError, OSError, ImportError), e:
     sys.stderr.write('Exception (%s): %s\n' % (e.__class__.__name__, str(e)))
     sys.exit(1)
-HTTPListener.main(None, HTTPListener, 8000)
+home = expanduser("~")
+config_path = home + '.vdm' if home.endswith('/') else home + '/' + '.vdm'
```
pdshaw: Keep it simple:  config_path = os.path.join(home, ".vdm")

lib/python/vdm/server/HTTPListener.py
```diff
@@ -438,8 +460,665 @@ def map_deployment_users(request, user):
...
+                handle_deployment_dict(deployment_elem, key, value, False)
+            elif type(value) is list:
+                handle_deployment_list(deployment_elem, key, value)
+            else:
+                if value is not None:
+                    deployment_elem.attrib[key] = str(value)
+        i += 1
+
+    try:
+        f = open(PATH + 'vdm.xml' if PATH.endswith('/') else PATH + '/' + 'vdm.xml','w')
```
pdshaw: Use os.path.join -- simpler and portable.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -937,7 +1649,94 @@ def delete(username, database_id):
...
+
+
+
+class VdmGetServerIP(MethodView):
+    """
+    Class to get the IP address of the server
+    """
+    @staticmethod
+    def get():
+        return jsonify({'ip_address': get_ip_address('eth0')})
```
pdshaw: Not safe to assume "eth0". Will be wrong a significant % of the time.

PR #3171, opened 2016-01-15
ENG-9317, always commit before switching DR streams.

luyangco:
Change-Id: I7b54b9cc25f56045863dcb113de10e323b7ecee5


afrogers:
Thought this is exactly what you'd done before. Looks like a merge error on my part?

PR #3173, opened 2016-01-15
ENG-9749, fix a DR issue that may send duplicate DR buffer after DR v…

luyangco:
…ersion negotiation.

Change-Id: Id3d3b015a87957f39522088e5ab7acc7cc3086ec

### DIFF
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -236,9 +236,10 @@ void ExecutorContext::setDrStream(AbstractDRTupleStream *drStream) {
     assert (m_drStream != NULL);
     assert (drStream != NULL);
     assert (m_drStream->m_committedSequenceNumber >= drStream->m_committedSequenceNumber);
+    m_drStream->periodicFlush(-1L, m_lastCommittedSpHandle);
     int64_t oldSeqNum = m_drStream->m_committedSequenceNumber;
-    m_drStream->periodicFlush(-1L, m_drStream->m_committedSpHandle);
     m_drStream = drStream;
+    m_drStream->m_committedSequenceNumber = oldSeqNum;
     m_drStream->setLastCommittedSequenceNumber(oldSeqNum);
 }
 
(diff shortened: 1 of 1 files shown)


src/ee/common/executorcontext.cpp
```diff
@@ -236,9 +236,13 @@ void ExecutorContext::setDrStream(AbstractDRTupleStream *drStream) {
...
     assert (drStream != NULL);
     assert (m_drStream->m_committedSequenceNumber >= drStream->m_committedSequenceNumber);
+    // Due to the sad fact that we have adopted separated way to track of the end of transaction,
+    // in endTransaction() and commit(), we have to call a extra flush to make sure all the numbers
+    // are synced before switching the stream. The same comment also applies to the function below.
+    m_drStream->periodicFlush(-1L, m_drStream->m_openSpHandle);
     int64_t oldSeqNum = m_drStream->m_committedSequenceNumber;
-    m_drStream->periodicFlush(-1L, m_drStream->m_committedSpHandle);
     m_drStream = drStream;
+    m_drStream->m_committedSequenceNumber = oldSeqNum;
```
afrogers: This should happen as a result of `commit()`?

src/ee/common/executorcontext.cpp
```diff
@@ -236,9 +236,13 @@ void ExecutorContext::setDrStream(AbstractDRTupleStream *drStream) {
     assert (m_drStream != NULL);
     assert (drStream != NULL);
     assert (m_drStream->m_committedSequenceNumber >= drStream->m_committedSequenceNumber);
+    // Due to the sad fact that we have adopted separated way to track of the end of transaction,
+    // in endTransaction() and commit(), we have to call a extra flush to make sure all the numbers
+    // are synced before switching the stream. The same comment also applies to the function below.
+    m_drStream->periodicFlush(-1L, m_drStream->m_openSpHandle);
```
afrogers: You are inside the `executorcontext`. Simply pass `m_lastCommittedSpHandle`, which is member state on the `executorcontext`. Looks less fishy...

src/ee/common/executorcontext.cpp
luyangco: Oh this line of code shouldn't be there, it's an accident, I'll remove it.

PR #3175, opened 2016-01-18
Fix the test for isInvalidReason.

BillWhite:
There was a typo for the test which caused it to always pass.  Then
error messages changed when we became more careful about ring
orientation.  We didn't see the error message changes because of the
typo.  I changed the definitions of the polygons in order to get the
error messages I want.

https://issues.voltdb.com/browse/ENG-9747


wolffcm:
Looks good to me.

PR #3176, opened 2016-01-18
fix license to copyright to 2016

jwiebema1:
changed the license to have a copyright of 2016.


akhanzode:
This is already on master.. You may want to pull and add commit hook so that license check failure does not get on master.

BillWhite:
Apparently my branch was not up to date with master on these files.  It seems that this failed before, and someone fixed it.  Maybe John Wiebe?

jwiebema1:
Anish fixed it just before the standup so both of our branches were just a checkin behind.

Thanks Anish,

PR #3178, opened 2016-01-18
ENG-9750 tweaked test json client invocations

vtkstef:


akhanzode:
Looks good to me if jenkins is blue merge.

PR #3181, opened 2016-01-18
Update GeographyPointValue equals method

wolffcm:
It wasn't smart about the poles or longitude [+-]180.


src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -144,12 +145,18 @@ public boolean equals(Object o) {
         }
 
         GeographyPointValue that = (GeographyPointValue)o;
+        if (this == that) {
+            return true;
+        }
+
+        GeographyPointValue normThis = normalizeLngLat(getLongitude(), getLatitude());
+        GeographyPointValue normThat = normalizeLngLat(that.getLongitude(), that.getLatitude());
```
BillWhite: Why use the normalized coordinates here?  Shouldn't the coordinates already be normalized?

wolffcm: Hmm, they will be normalized but not canonicalized.  I should have a function called canonicalizeLngLat that sets longitude to 0 at the poles, etc.  This need not be done in the normalize function, as it does in my commits.

I'll try again.

BillWhite: Normalizing converts (x, 90) to (0, 90) and (x, -90) to (0, -90), right?

BillWhite: I think what you have done is right.  The coordinates should be in range.  So all you will be doing with normalizeLngLat is fixing up lat=+/-90.  That seems right to me.

BillWhite:
What did you do about the normalization tests?  Maybe we should just do away with them.  I'm not sure they are giving up much.

Other than that, it seems as if you have gotten everything right, and the tests seem complete.

BillWhite:
The lng/lat normalization tests were auto-generated by the main in TestGeographyPointNormalization. It looks like you changed them by hand.  Perhaps the main routine there should do the right thing when Math.abs(lat) == 90.0.

src/frontend/org/voltdb/types/GeographyPointValue.java
wolffcm: We also force a -180 longitude coordinate to be +180.

I was thinking that we should make a distinction between *normalizing*, that is, forcing coordinates to be in range with modulo-like arithmetic, and *canonicalizing*, forcing different representations of the same point to be the same.

Maybe that's not needed though.  If you're okay with this, I won't sweat it.

wolffcm:
I didn't edit by hand, I re-auto-generated the TestGeographyPointNormalization tests, since they just use normalizeLngLat to generated expected values.

wolffcm:
Merged externally.

PR #3187, opened 2016-01-20
Clean up lookups for USING and Ambiguous columns.

BillWhite:
In ENG-9367 we change the lookup rules to reject queries which had
references to ambiguous columns.  This caused a number of queries in
plannerTester to fail, as they were ambiguous.

However, our implementation in ENG-9367 was flawed, in that we rejected
references like R1.C in "select R1.C from R1 join R2 using(C)".  HSQLDB
called these unreferenceable, and so we did too.  This is apparently
incorrect.  I checked with mysql and postgres.  Both consider "select C
from R1 join R2 using(C), R3" to be ambiguous when R1, R2 and R3 all
have columns named C.  They are happy with the same query and schemata
with "C" in the select list replaced by "R1.C" or "R2.C".  We have added
tests for all these cases.

The relevant section of the 2003 SQL standard seems to be section 7.7
paragraph (7).


wolffcm:
I'm looking at this now.

wolffcm:
Having read through the section of the standard that you cite, I'm convinced that the behavior we got on your first pass was valid.  It seems like the standard says that the corresponding columns are coalesced in the joined table, and are no longer accessible with a table qualifier elsewhere in the query.  But this is one of those cases where it's probably best to just follow what other vendors do.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
```diff
@@ -224,8 +224,19 @@ public OrderedHashSet getUniqueColumnNameSet() {
      * @return int index or -1 if not found
      */
     public int findColumn(String columnName) {
+        return findColumn(null, columnName);
+    }
 
+    public int findColumn(String tableName, String columnName) {
+        // The namedJoinColumnExpressions are ExpressionColumn objects
+        // for USING conditions.  If we are given a reference to something
+        // like "R1.C", then we never want to use these.  If we are given
```
wolffcm: I think I understand what the code is doing, but I'm finding this comment a little confusing.  If the arguments to this method are "T1" and "C", then we *do* want to resolve to the column T1.C, regardless of whether or not T1.C is one of a pair of corresponding columns in a USING clause.  That's what you're fixing here, right?  I'm having some difficulty getting there from the text of this comment.

wolffcm:
It looks like HSQL has a slightly separate code path for USING clauses that reference columns in derived tables, e.g.,
```SQL
SELECT C
FROM 
  (SELECT C*C FROM MYTBL) AS R1 
  JOIN R2 
  USING (C)
```
Do we have tests for this?

wolffcm:
What you have here looks good to me.  It would be worth getting a second opinion (like from Xin or Paul) since this is sensitive code and we are close to release.

BillWhite:
I don't have a jUnit test for this query, but I have verified manually that this seems to work as I expect.
I'll add some more tests.

tests/frontend/org/voltdb/planner/plannerTester.java
```diff
@@ -451,17 +451,22 @@ private static void configCompileSave(String config, boolean isSave) throws Exce
                 joinOrder = splitLine[1];
                 query = splitLine[2];
             }
-            List<AbstractPlanNode> pnList = s_singleton.compileWithJoinOrderToFragments(query, joinOrder);
-            AbstractPlanNode pn = pnList.get(0);
-            if (pnList.size() == 2) {// multi partition query plan
-                assert(pnList.get(1) instanceof SendPlanNode);
-                if ( ! pn.reattachFragment((SendPlanNode) pnList.get(1))) {
-                    System.err.println("Receive plan node not found in reattachFragment.");
+            // If one compilation, try subsequent ones.
```
paulmartel: This comment no verb. I think you mean "If one statement fails to compile,"

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
```diff
@@ -224,8 +224,19 @@ public OrderedHashSet getUniqueColumnNameSet() {
...
+        // a reference to something like "C", then we may want to use
+        // these, but they may be ambiguous with other range variables.
+        // We will find these other range variables in the other
+        // part of our search.
         if (namedJoinColumnExpressions != null
+                && tableName == null
```
paulmartel: I think that I approve of this change, based on the following understanding:
The old code seems to be refusing to resolve column names here if they match a USING column name. There is presumably code elsewhere that actively provides the resolution of a column name to a USING column, but that is NOT happening here.
The changed code only refuses to resolve column names here if they are NOT table-qualified.
If we continue to assume that there IS an alternative code path to resolve (only) unqualified names with USING columns, then it makes some sense that the code here would not want to block qualified names -- on which USING columns have no bearing.
In short, the old code seems to incorrectly allow USING columns to interfere with the resolution of columns qualified by table names (even/especially) when those table names are from a join OUTSIDE the "JOIN USING". The new code would only (correctly) interfere for unqualified names. 
It suggests that for "SELECT OTHER.C, C FROM LHS JOIN RHS USING (C) JOIN OTHER WHERE...;" "OTHER.C" will now find a correct resolution where it used to fail. The unqualified "C" would presumably resolve to the C in the USING clause (one of LHS.C or RHS.C) and ignore the hypothetical ambiguity with OTHER.C.  It's not trivial to explain WHY the "USING(C)" unambiguously trumps "OTHER.C" avoiding ambiguity -- but it seems like a plausible approach and a definite improvement over claiming that OTHER.C is unresolvable.

I could imagine a 4-way join of "(W JOIN X USING (C)) JOIN (Y JOIN Z USING (C)) where there would still not be a clear resolution for an unqualified C -- and AFAICT no way to qualify C without introducing aliases -- MAYBE this remaining ambiguous case is correctly flagged?

BillWhite: 1. There is a separate function, getExpressionColumn, which matches USING columns.  This is called elsewhere.
2. The query  "SELECT OTHER.C, C FROM LHS JOIN RHS USING (C) JOIN OTHER ON...;" fails. We think putting a WHERE clause there is a syntax error.  If we replace the last JOIN OTHER with ", OTHER" then the where is syntactically correct.  However, the compilation fails. The error message is that C is in USING(C) and OTHER.  There is (now) a test for this in TestPlansJoin.
3. The query 3. The query "SELECT OTHER.C, C FROM LHS JOIN RHS USING (C) JOIN OTHER USING(C);" compiles, though the semantics are different.

SO, my only question is, should query (2) fail to compile, or should it find C in the USING and ignore the C in OTHER?  I think failing to compile is not a terrible result, as the query "SELECT OTHER.C, LHS.C FROM LHS JOIN RHS USING(C), OTHER" gives the same result.

Changing schemas slightly, let R1 and R2 have the same schema, with one string column TM.

1. The query "select * from (R1 join R2 using(TM)) as S1, (R1 join R2 using(TM)) as S2;", where R1 and R2 have identical schemas, compiles and executes, and seems to get the right result.  This fails to compile in mysql, complaining about a syntax error at the "as S1" part.
1. The closely related query "select S1.* from (R1 join R2 using(TM)) as S1, (R1 join R2 using(TM)) as S2;"  crashes the EE.  This is also syntactically incorrect according to Mysql.
1. The

PR #3190, opened 2016-01-20
Edit isinvalidreason messages.

BillWhite:
Some of the messages for isinvalidreason were not well stated.  Now each
message is a single line which does not end with a newline or a period.

https://issues.voltdb.com/browse/ENG-9748


wolffcm:
Looks good.

PR #3191, opened 2016-01-20
Make an ambiguous query be valid in agg_ddl.sql.

BillWhite:
We recently changed the lookup rules for column references to throw an
error when a column reference could denote more than one column.  One of
the queries in this ddl file was rendered invalid by this change.  This
commit corrects the query.

https://issues.voltdb.com/browse/ENG-9776


wolffcm:
This seems like a reasonable change to me.  But maybe @jxchinaren or @paulmartel should have a look too to make sure.

PR #3192, opened 2016-01-20
Add null pointer check in debugVerifyCatalog

wolffcm:
I believe this will fix ENG-9769, ENG-9777, ENG-9778.


wolffcm:
Pro pull request is here:
https://github.com/VoltDB/pro/pull/1163

PR #3193, opened 2016-01-20
ENG-9764: add pro=True/False for --version CLI

akhanzode:


lib/python/voltcli/utility.py
```diff
@@ -858,6 +858,22 @@ def get_java_version():
         return ""
 
 #===============================================================================
+def is_pro_version(voltdb_jar):
+#===============================================================================
+    """
+    Assumes caller has already run "find_in_path(jar)" so we know it can be checked.
+    """
+    try:
+        proc = subprocess.Popen(['jar', 'tvf', voltdb_jar], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
vtkstef: wouldn't it be cheaper in resource usage and faster using a Zip library that are a plenty and standard in Python?

lib/python/voltcli/utility.py
```diff
@@ -858,6 +858,22 @@ def get_java_version():
...
+        out, err = proc.communicate()
+        if "org/voltdb/CommandLogImpl.class" in out:
```
vtkstef:
```python
return "org/voltdb/CommandLogImpl.class" in out
`

PR #3194, opened 2016-01-20
ENG-9788 more resilient topic metadata probes, and handle large topic offset values

vtkstef:


vtkstef:
Reviewed by @akhanzode

PR #3197, opened 2016-01-21
Eng 9500 sqlcov geo joins

johnTcrawford:
Added tests of joins on point or polygon columns. Also improved existing tests, and particularly improved the normalizer, to round-off string columns containing point or polygon values, as float columns, to a specified number of digits.


tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -11,29 +12,38 @@ INSERT INTO _table VALUES (103, pointFromText('POINT(  0.01    1.01 )'), pointFr
...
 
 -- Super-large polygons, dividing the Earth up into 8 octants:
 -- (Note: the longitude of the North and South Pole is arbitrary; however, it has been set here to the values
 --  -45, 45, -135, 135, in a (partially successful) attempt to avoid a PostGIS bug in computing the CENTROID)
--- First, the "front": the North-West, North-East, South-East & South-West octants that touch the Prime (Greenwich) Meridian
+-- First, the "front": the North-West, North-East, South-East & South-West octants that touch where the Prime (Greenwich) Meridian meets the Equator
 INSERT INTO _table VALUES (311, pointFromText('POINT(-71.065  42.355)'), pointFromText('POINT( 71.065 42.355)'), null, polygonFromText('POLYGON((   0 0, -45  90, -90   0,    0 0))') )
-INSERT INTO _table VALUES (312, pointFromText('POINT( 71.065  42.355)'), pointFromText('POINT(-71.065 42.355)'), null, polygonFromText('POLYGON((   0 0,  90   0,  45  90,    0 0))') )
+-- This one has been made slightly smaller, to avoid a PostGIS bug (it works fine either way, in VoltDB):
```
BillWhite: I wonder if we should be reporting these PostGIS bugs to PostGIS.  Maybe they are right and we are wrong.

BillWhite:
As a non-expert in sqlcoverage I say this looks alright to me.  I must admit that the only function I felt I really understood was round_to_digits, so my opinion is of limited value.

PR #3201, opened 2016-01-21
ENG-9243: Bail out of partial snapshot if list of specified tables co…

afrogers:
…ntains typos


nshi:
The change looks good to me. It'll be even better if you can add a test to `TestSnapshotInitiationInfo`.

PR #3203, opened 2016-01-22
Edit javadoc comments for new user-facing classes in 6.0

wolffcm:
Also change error messages to use "ring" instead of "loop" to be
consistent with docs.


src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -99,19 +109,30 @@ public static GeographyPointValue fromWKT(String param) {
         }
     }
 
+    /**
+     * Return the latitude of this point in degrees.
+     * @return The latitude of this point in degrees.
```
paulmartel: I usually leave off the general description when it is redundant with the @return like this, but I suppose tastes may vary.

paulmartel:
Looks good to me.

src/frontend/org/voltdb/types/GeographyPointValue.java
```diff
@@ -369,7 +410,7 @@ public GeographyPointValue mul(double alpha) {
     /**
      * Return a new point which is this point rotated by the angle phi around a given center point.
      *
-     * @param phi The angle to rotate.
+     * @param phi The angle to rotate in degrees.
```
BillWhite: Good catch.

BillWhite:
I like your changes.

hhakimi53:
Looks good.

src/frontend/org/voltdb/types/GeographyPointValue.java
wolffcm: That would be my preference too.  If leave off the general description, the generated javadoc looks funny, since it's missing the summary near the top of the page.  If I leave off the `@return` then javadoc produces warnings.  So I decided to go with both.

PR #3204, opened 2016-01-22
ENG-9770: Fix EE IPC interface error in exeucteTask().

nshi:
The Java side expects executeTask() to return something, so return 0 for
those tasks that don't have any return value to write. Otherwise, the
Java side will wait forever.


luyangco:
Looks good to me.

PR #3207, opened 2016-01-22
In EE, fix point comparison at poles and antimeridian

wolffcm:
We previously fixed a similar problem with our Java GeographyPointValue class.  This is the corresponding change for when we do point comparison in the EE.

Also make polygon comparison use the point comparison method for vertices.  Previously we were using S2's point compare, which puts latitude first.  This resulted in having to change some unit tests.

### DIFF
--- a/src/ee/common/GeographyPointValue.hpp
+++ b/src/ee/common/GeographyPointValue.hpp
@@ -80,6 +80,8 @@ public:
     // function should return a value that is the same as in Java
     // code: GeographyPointValue.EPSILON.
     static Coord epsilon() {
+        // Making this a static method rather than a static member
+        // variable for the same reason as nullCoord(), above.
         return 1e-12;
     }
 
@@ -229,7 +231,7 @@ private:
 
         // For longitudes within epsilon of the antimeridian
         // (on the east side), canonicalize to 180.0.
-        if (m_longitude < 0.0 && 180.0 + m_longitude < epsilon()) {
+        if (180.0 + m_longitude < epsilon()) {
             newLng = 180.0;
         }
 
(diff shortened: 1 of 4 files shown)


src/ee/common/GeographyPointValue.hpp
```diff
@@ -203,6 +214,28 @@ class GeographyPointValue {
...
+
+        if (90.0 - fabs(m_latitude) < epsilon()) {
+            // We are at one of the poles;
+            // longitude doesn't matter, so choose 0.
+            newLng = 0.0;
+        }
+
+        // For longitudes within epsilon of the antimeridian
+        // (on the east side), canonicalize to 180.0.
+        if (m_longitude < 0.0 && 180.0 + m_longitude < epsilon()) {
```
paulmartel: The m_longitude < 0.0 test is redundant -- just let the 180 + m_longitude < epsilon test fail naturally for non-negative values.
Also, shouldn't epsilon just be referenced as a final static member internally -- even if you want to provide an accessor function for external use?

paulmartel:
End of review.

src/ee/common/GeographyPointValue.hpp
wolffcm: Yes, you're right about the `<` test.

I had to make epsilon be a function because my compiler complains about constants being initialized inside a class definition without the `constexpr` keyword.  I guess it's a C++11 thing.  But `constexpr` is not yet supported on all the platforms we build on.  I could use a macro or something, but a function seemed a little less ugly.  I had to do the same thing for the nullCoord method.  I'll add a comment explaining this.

PR #3208, opened 2016-01-25
ENG-9750 clean up test code and continuation handling in the servelet 

vtkstef:


vtkstef:
Reviewed by @akhanzode

PR #3210, opened 2016-01-25
ENG-9809: fix null param values from VMC.

BillWhite:


wolffcm:
I guess both you and Anish have vetted this change, but it looks good to me too.

Have we done manual testing for passing null values to stored procedures for all of our data types?

BillWhite:
Not that I know of.  I'll do that now.

wolffcm:
Thanks, Bill.  And thanks, Anish.

BillWhite:
It is a bug that inserting a null into a timestamp column produces a
timestamp of 1970-01-01 00:00:00.000000?  This seems somehow wrong to me.

I created a table with one row for each type, and inserted rows with null
values in each of them.  The values in the timestamp column for each row of
nulls is Jan. 1, 1970.  Perhaps we are converting a 0 to a date string.

The same in sqlcmd produces a NULL in timestamp column.

This seems like a VMC bug.

wolffcm:
It seems like a bug to me, but a separate one from this one.  I guess it could be either an issue with how we display tables or how we pass null timestamps to stored procedures in VMC.  I wonder what happens if you insert a null timestamp from VMC and select it back from sqlcmd, and vice versa?

BillWhite:
This is ENG-9818.

BillWhite:
I verified manually that null can be passed to columns of all types in VMC with this fix.

The only problem is with printing NULL values for timestamps in VMC.  This is ENG-9818.

PR #3212, opened 2016-01-26
ENG-9759: Fix serialization in optimized DR UPDATE/DELETE path.

nshi:
The optimized UPDATE/DELETE path only serializes the indexed
columns. The offset in the null array should use the relative position
of the column in the interesting column vector, not the actual table
column index.


afrogers:
Reviewed. Looks good.

PR #3213, opened 2016-01-26
Added randomization of point and polygon values, and new

johnTcrawford:
geo-advanced-point and geo-advanced-polygon test suites that use them
(as do existing geo-basic, geo-basic-point, geo-basic-polygon). Also,
added two new polygon columns, POLY2 & POLY3. And, improved geo-basic,
adding additional points and polygons, and tweaking queries to avoid
test failures due to small round-off error and such.


tests/scripts/SQLGenerator.py
```diff
@@ -145,6 +145,155 @@ def generate_values(self, count):
                 yield decimal.Decimal(str(random.random() * 100.00))
 
 
+class PointValueGenerator:
+    """This generates (random) point (GEOGRAPHY_POINT) values.
+    """
+    DIGITS_BEYOND_DECIMAL_POINT = 2
```
wolffcm: Only two?  Maybe that's as precise as we can get to maintain agreement with PostGIS?

tests/scripts/SQLGenerator.py
```diff
@@ -145,6 +145,155 @@ def generate_values(self, count):
...
+
+    def __init__(self):
+        decimal.getcontext().prec = PointValueGenerator.DIGITS_BEYOND_DECIMAL_POINT
+        self.__nullpct = 0
```
wolffcm: No nulls by default?  Seems like you would want like 10% nulls or something.

tests/scripts/SQLGenerator.py
```diff
@@ -145,6 +145,155 @@ def generate_values(self, count):
...
+        else:
+            self.__latmax = self.__longmax
+
+    def set_nullpct(self, nullpct):
+        self.__nullpct = nullpct
+
+    def set_num_holes(self, num_holes):
+        self.__num_holes = num_holes
+
+    def generate_vertex(self, longmin, longmax, latmin, latmax):
```
wolffcm: Seems like you could use your construct a point value generator to do this, and avoid a little bit of code duplication?

wolffcm: That was just a suggestion, mostly to keep in mind for the future---I realize that these classes produce strings so that makes code reuse a little tricky between points and polygons.

You could imagine in the future having geography and geography_value classes in python that would help here.  That would be part of writing the python driver, I guess.

tests/scripts/SQLGenerator.py
```diff
@@ -145,6 +145,155 @@ def generate_values(self, count):
...
+    """
+    DIGITS_BEYOND_DECIMAL_POINT = 2
+
+    def __init__(self):
+        decimal.getcontext().prec = PolygonValueGenerator.DIGITS_BEYOND_DECIMAL_POINT
+        self.__nullpct = 0
+        # A negative value indicates a random number of holes (interior rings),
+        # with the number of holes ranging between 0 and the absolute value of
+        # the given number; so, in this case, -4 means between 0 and 4 holes.
+        self.__num_holes = -4
```
wolffcm: And a non-negative number indicates a non-random number of holes I guess?

tests/scripts/SQLGenerator.py
```diff
@@ -281,15 +430,19 @@ def __init__(self, token):
...
     TYPE_PATTERN_GROUP  =                                           "type" # optional type for columns, values
     #                   |       |             |                      |
-    __EXPR_TEMPLATE = r"%s" r"(\[\s*" r"(#(?P<label>\w+)\s*)?" r"(?P<type>\w+|[=<>!]{1,2})?\s*" \
-                      r"(:(?P<min>(-?\d*)),(?P<max>(-?\d*)))?\s*" r"(null(?P<nullpct>(\d*)))?" r"\])?"
-    #                         |                |                             |                   |
-    #                         |                |                             |       end of [] attribute section
-    NULL_PCT_PATTERN_GROUP  =                                               "nullpct" # optional null percentage
-    #                         |                |
-    MAX_VALUE_PATTERN_GROUP =                 "max" # optional max (only for numeric values)
+    MIN_VALUE_PATTERN_GROUP =                                                                          "min" # optional min (only for numeric values)
```
wolffcm: This code looks cryptic.  I am not so clear on how these generators work.  I think my ignorance of sqlcoverage is starting to show here.

The only thing I will say is that I become suspicious when I see references to subclass attributes in a base class, like `MAX_LAT_PATTERN` etc below.  But this seems to be an intentional expedient design choice consistent across all generators?

tests/scripts/examples/sql_coverage/configure-for-geo-polygon.sql
```diff
@@ -28,19 +28,13 @@
 {@fromtables = "_table"}
 {@idcol = "ID"}
 
---TODO: this should be randomized:
---{@insertvals = "_id, pointFromText('POINT(0 0)'), pointFromText('POINT(-71.0 42.0)'), null, polygonFromText('POLYGON((-1 -1, 1 -1, 1 1, -1 1, -1 -1))')"}
-{@insertvals = "_id, pointFromText('POINT(0 0)'), pointFromText('POINT(-71.0 42.0)'), null, polygonFromText('POLYGON((-1 -1, 1 -1, 1 1, -1 1, -1 -1))')"}
---{@insertvals = "_id, pointFromText('POINT(_value[decimal:-1,1] _value[decimal:-1,1])'), pointFromText('POINT(_value[decimal:-1,1] _value[decimal:-1,1])'), null, polygonFromText('POLYGON((_value[decimal:-0.9,-0.3] _value[decimal:-0.9,-0.3], _value[decimal:0.3,0.9] _value[decimal:-0.9,-0.3], _value[decimal:0.3,0.9] _value[decimal:0.3,0.9], _value[decimal:-0.9,-0.3] _value[decimal:0.3,0.9], _value[decimal:-0.9,-0.3] _value[decimal:-0.9,-0.3]))')"}
+{@insertvals = "_id, _value[point:10,12], null, null, _value[polygon], _value[polygon null25], _value[polygon null50], null"}
```
wolffcm: What does `_value[point:10,12]` mean here?  is the 10, 12 the min and max values generated for longitude, with the latitude unbounded?

tests/scripts/SQLGenerator.py
wolffcm: I see you've specified some generations of null values in your templates below, so please ignore this comment.

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -179,69 +221,65 @@ SELECT ID,      CAST(POLY1 as VARCHAR) POLY1VC                      from _table
...
--- Test joins against point & polygon columns
--- (Note: PostGIS considers two geographies to be equal if their 'bounding boxes' are the same, which
---  includes polygons with the same exterior ring but different interior rings, as well as points that
---  are very close together, but not quite identical (see: http://postgis.net/docs/ST_Geometry_EQ.html);
---  so such cases have been excluded here, since VoltDB treats them differently, by design. Similarly,
---  we avoid the use of <, <=, >, >= here, since PostGIS and VoltDB also treat those very differently,
---  by design.)
-SELECT A.ID AID, AsText(A.PT1  ) APT1  , B51.ID BID, AsText(B51.PT1  ) BPT1   FROM _table A JOIN _table B51 ON A.PT1   @cmp B51.PT1   AND BID NOT IN (107, 108)
-SELECT A.ID AID, AsText(A.POLY1) APOLY1, B52.ID BID, AsText(B52.POLY1) BPOLY1 FROM _table A JOIN _table B52 ON A.POLY1 @cmp B52.POLY1 AND NumInteriorRings(A.POLY1) = NumInteriorRings(B52.POLY1)
+SELECT ID, CONTAINS(POLY1,PT1), LONGITUDE(PT1), LATITUDE(PT1) from _table G58
```
wolffcm: This fails to compile, right?  Because of the boolean in the select list issue?

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -2,171 +2,213 @@
...
+{@ignoredistroundoff  = "A.ID <> 453"} -- ignore tiny DISTANCE round-off errors, between VoltDB and PostGIS
+{@ignoreisvaliddiffs  = "ID < 800"} -- ignore known differences with IsValid(polygon), between VoltDB and PostGIS
+{@ignorecentroiddiffs = "ID < 400 AND ID > 100"} -- ignore known differences with CENTROID(polygon), between VoltDB and PostGIS
+{@ignoreinvalidpolys  = "ID < 600"} -- ignore polygons that are invalid, according to VoltDB and/or PostGIS
+{@ignore2invalidpolys = "A.ID < 600 AND B.ID < 600"} -- ignore 2 invalid polygons, with a join
+{@ignore2diffnumholes = "NumInteriorRings(A.POLY1) = NumInteriorRings(B.POLY1)"} -- ignore 2 polygons with different numbers of holes, with a join
+
+
+-- Insert some random points & polygons
+INSERT INTO _table VALUES (_id, _value[point:-109.05,-102.05,37.0,41.0], _value[point:129,139,-26,-11 null25], _value[point null50], _value[polygon], _value[polygon:129,139,-26,-11 null50], null, null)
```
wolffcm: This one looks interesting---you're inserting some randomly generated polygons, including holes, within a bounding box, and some points within the same bounding box?  Do we have tests for `CONTAINS(poly2, pt2)`?

wolffcm:
That's all my comments.  I like what you've done.

I feel like sqlcoverage would begin to become hard to extend if we added more types (e.g., linestring or multipolygon).  I'm curious to know what you think about how easy/difficult it's been to do this work.  (It looks like it's been pretty tricky to me.)

tests/scripts/SQLGenerator.py
johnTcrawford: No, it's not that bad, and that value is specified elsewhere (normally 8 significant digits, but 4 for one of the test suites, mainly because of some  CENTROID queries). This is just the number of random digits beyond the decimal point - I just found it annoying to be looking at 12-digit random numbers. I'll add a clarifying comment.

johnTcrawford: Also, I think I'll remove this line:
decimal.getcontext().prec = PointValueGenerator.DIGITS_BEYOND_DECIMAL_POINT
which I stole from the DecimalValueGenerator class, but did not shorten the number of random digits for me, I still had to round off later.

tests/scripts/SQLGenerator.py
johnTcrawford: As you've noted, non-zero null percent values are specified in the actual tests. Here, I was just following the convention of other data types within SQLCoverage, where the default percent is always zero.

tests/scripts/SQLGenerator.py
johnTcrawford: Yes, I'll add a clarifying comment.

tests/scripts/SQLGenerator.py
johnTcrawford: Yeah, I did think of that, but in this case it just seemed slightly more complicated than it was worth.

tests/scripts/SQLGenerator.py
johnTcrawford: Yeah, it is very cryptic; I'm not sure I fully understand it myself - just well enough to add to it slightly. All I did was add:
(,(?P<latmin>(-?\d*\.?\d*)),(?P<latmax>(-?\d*\.?\d*)))?
and then I evened out the two lines a bit, so r"(:(?P<min>(-?\d*\.?\d*))," got moved up to the first line; and I added comments following the existing pattern.

tests/scripts/examples/sql_coverage/configure-for-geo-polygon.sql
johnTcrawford: It means that both longitude and latitude have a minimum of 10 and maximum of 12 - which is not something anyone would ever want in the real world, but is reasonable for testing, and I used it to test that my randomization was working correctly. Though actually, I meant to change it to something that would match the polygons better, so I'll do that. I'll make it "-109,-102,37,41", which is Colorado rounded off to the nearest integer - as you may recall, Colorado is the default limits for a random polygon.

johnTcrawford: (Not that it matters much here; the real testing of CONTAINS is done in the geo-basic test suite, not here.)

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: Yeah, good idea, I'll add that. (FYI, PT2 and POLY2 are roughly within Australia's Northern Territory, which is rectangular except for its northern coastline - I wanted to hit the Southern and Eastern hemispheres.) But, for the record, PT1 and POLY1 do the same thing within Colorado (since I've specified that for PT1, and POLY1 uses the default, which is Colorado.

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: Yes, but it fails on both sides, so the test passes. And when we do support this (which I look forward to), the test is already written. (There's also a similar test for IsValid.)

tests/scripts/SQLGenerator.py
johnTcrawford: My comment above is even more cryptic, due to GitHub/HTML swallowing the words 'latmin', 'latmax', and, in the one that moved up to the first line, 'min', because they had angle brackets around them.

johnTcrawford:
Thanks. Yes, it's been a huge effort - counting the SQLCoverage PostgreSQL preparatory work, it's been my main job since October, and parts of it have been quite tricky. However, having done it once, I think it will be a lot easier to add additional data types, partly because the groundwork has been laid, and partly because I'm now familiar with the relevant code (including the parts I did not write).

PR #3214, opened 2016-01-26
Minor fix to test for it's result set to be order deterministic

hhakimi53:
With changes, test passes in valgrind environment on local machine. 

I have started "test-nextrelease-memcheck-pro" Jenkins job on the current branch to verify changes fixes the issue in Jenkins environment.


wolffcm:
Looks good.  Thanks for jumping on this, Huzefa.

PR #3215, opened 2016-01-26
ENG-9552: Add primary key hash to end transaction and move buffer tra…

wweiss-voltdb:
…ce code to DRBufferParser


src/ee/storage/AbstractDRTupleStream.h
```diff
@@ -104,6 +106,7 @@ class AbstractDRTupleStream : public TupleStreamBase {
     size_t m_secondaryCapacity;
     int64_t m_rowTarget;
     bool m_opened;
+    int64_t m_txnPkHash;
```
luyangco: It's better to comment what does LONG_MIN and LONG_MAX mean here.

luyangco:
Review complete.

PR #3222, opened 2016-01-26
Add some code to print serialized polygons for the wire protocol doc.

BillWhite:
In the wire protocol doc we have examples of serialized data structures.
For GeographyValue, since this is difficult to calculate, I added some
code to the TestGeographyValue test to print an HTML table of the
serialized result.  This code is disabled by default, so it does not
have any real testing performance impact.

https://issues.voltdb.com/browse/ENG-9211

### DIFF
--- a/tests/frontend/org/voltdb/types/TestGeographyValue.java
+++ b/tests/frontend/org/voltdb/types/TestGeographyValue.java
@@ -23,10 +23,13 @@
 
 package org.voltdb.types;
 
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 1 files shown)


wolffcm:
This seems like an elegant way to generate documentation.  I like it.

Does this need to be in before 6.0 ships?  I'm not clear on how we generate the latest up-to-date wire protocol document on the website.

wolffcm:
I'm guessing that the right thing to do is to merge after we branch.

PR #3223, opened 2016-01-27
increase display column width for geography column

hhakimi53:
Increase print column width for geography column so that logic will display value of polygon aligned within  geography column for polygon up to:
- 5 vertices - 4 unique vertices + 1 repeating vertex
- Vertex's longitude can take upto 5 characters and latitude 4 characters.

Based on above, polygon described as:
"POLYGON ((179.1 89.9, -179.1 89.9, -179.1 -89.9, 179.1 -89.9, 179.1 89.9))"
can be displayed in aligned manner within geography column at the best.

If polygon goes beyond the above defined limit, the contents will push the following further out thereby resulting in unaligned display for following columns. This behavior is similar to what we see even within variable character length column where a string length of 30 or beyond will also push display for following outwards resulting in unaligned display w.r.t. column names.


wolffcm:
This looks fine.  I think you'll need to update sqlcmdtest as well---did you try running it?  Master being closed, you'll want to merge after we branch.

For real production polygons, I suspect that the WKT will be very long.  There doesn't seem to be much value in keeping columns aligned in that case.  But we can at least make VoltTable output look nice for the "kicking the tires" uses, and these changes accomplish that.

hhakimi53:
No, I have not changed the sqlcmdtest for this; I will update them. Thanks Chris for bringing that up.
Sure, after the changes for sqlcmd are done and once jenkins jobs are fine with the changes, I will merge the changes into master after we branch.

hhakimi53:
Jenkins is blue on the changes; there is an issue in PostGis sqlcoverage and is not related to this one. John C is aware and will look into the PostGis sqlcoverage issue.

PR #3225, opened 2016-01-28
ENG-9765 guard against stopping importers that have yet to start

vtkstef:


src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -109,20 +109,24 @@ protected ImportManager(int myHostId, HostMessenger messenger, ImporterStatsColl
...
-                .add("org.voltdb.client")
-                .add("org.slf4j")
-                .add("org.voltcore.utils")
                 .add("com.google_voltpatches.common.base")
                 .add("com.google_voltpatches.common.collect")
-                .add("com.google_voltpatches.common.net")
                 .add("com.google_voltpatches.common.io")
+                .add("com.google_voltpatches.common.net")
                 .add("com.google_voltpatches.common.util.concurrent")
+                .add("jsr166y")
```
akhanzode: Why this new addition now?

src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
```diff
@@ -228,9 +232,11 @@ public final void stop(ChannelDistributer distributer)
             success = m_importers.compareAndSet(oldReference, ImmutableMap.<URI, AbstractImporter> of());
         } while (!success);
 
+        if (!m_starting.get()) return;
```
manjujames: Shouldn't we do this m_starting.get() check right after we set m_stopping?

manjujames:
I think we should also change ImportManager to call importProcessor.shutdown only if start has been called. So, put that also within a "if (!m_serverStarted)" block like we do for calling readyForData.

src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
vtkstef: because we need to clear what was built in config which is called before readyForDate

vtkstef:
ImportManager shutdown still needs to unload the OSGi bundles

PR #3231, opened 2016-02-01
Update logic for geo type support to DR

hhakimi53:


BillWhite:
Looks good to me.  I would not have thought that GeographyValue did not have a definition for equals.

hhakimi53:
I also discovered that when tried adding test for it :-). 

Thanks for looking at the changes. Pushed a minor update where I have added a volt compiler test to check we can DR a table containing geo types. We have corresponding on pro side, so thought will add one on community as the guards have been removed.

I will merge changes to master once Jenkins is blue.

src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -255,6 +256,45 @@ public String toWKT() {
...
+            List<XYZPoint> expectedLoop = expectedLoopIt.next();
+
+            // check if number of the vertices in loops are equal
+            if (expectedLoop.size() != actualLoop.size()) {
+                return false;
+            };
+
+            Iterator<XYZPoint> expectedVertexIt = expectedLoop.iterator();
+            for (XYZPoint actualPt : actualLoop) {
+                XYZPoint expectedPt = expectedVertexIt.next();
```
wolffcm: What does `XYZPoint.equals` do?  When comparing polygon vertices, seems like we should do the same thing that we do for comparing instances of `GeographyPointValue`, including considering points equal if they are within some epsilon.  Perhaps we should convert instances of `XYXPoint` into `GeographyPointValue` and use that class's `equals` method?

hhakimi53: XYZPoint.equals() is currently comparing GeographyPointValue in three dimension representation. We can use GeographyPointValue.equals() to compare each of the vertices's of polygon and should be easy to do where we use getRings() to fetch loops of polygon which stores it's vertices's in GeographyPointValue.  I was thinking in term of the object serialization/deserialization should not effect precision of the points as no round-trip conversion is involved and chose xyzpoint.equals(). Will update logic to use GeographyPointValue.equals() with epsilon instead of xyzpoint.equals() for case for roundtrip conversion is involved and user can still the currently implementation of Geography.equals()

BillWhite: I wonder, in the EE we compare points in the dictionary, or lexicographical order.  But the loops there are always ordered CCW.  In Java we need to be sure we compare in the same order as the EE.  I think you are doing this right, as you seem to compare in the order of the rings, and the rings are in EE order.  But you might double check that, unless you already verified it.

This is the kind of mistake I would make.

hhakimi53: I have updated pro-tests to use polygon with 2 holes and they work fine. We should be good with it I think.

PR #3234, opened 2016-02-01
ENG-9624 add voltdb collect folder structure

dweiss88:


src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -480,7 +500,7 @@ private static boolean getUserResponse(String prompt) {
         }
     }
 
-    private static void cmd(TarGenerator tarGenerator, String[] command, String folderPathInTar, String resFilename)
+    private static void cmd(ZipOutputStream zipStream, String[] command, String folderPath, String resFilename)
```
akhanzode: Does TarMalformat exception get thrown now that you are using Zip?
If Tar is not in use anywhere delete the Tar code.
Make sure you handle exceptions from Zip as when file is open and in use we had issues writing/reading.
Make sure that you have a test where file is actively getting written (or deleted under you) but you can still collect.

akhanzode:
Update tests accordingly.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -246,8 +248,8 @@ private static void parseJSONObject(JSONObject jsonObject) {
         }
     }
 
-    private static List<String> listCollection(boolean skipHeapDump) {
-        List<String> collectionFilesList = new ArrayList<String>();
+    private static Set<String> listCollection(boolean skipHeapDump) {
```
manjujames: May be change this method name as well because it no longer returns a List.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -353,11 +358,15 @@ private static void generateCollection(List<String> paths, boolean copyToVEM) {
                 rootpath = System.getProperty("user.dir");
             }
 
-            String collectionFilePath = rootpath + File.separator + m_config.prefix + timestamp + ".tgz";
-            File collectionFile = new File(collectionFilePath);
-            TarGenerator tarGenerator = new TarGenerator(collectionFile, true, null);
-            String folderPath= m_config.prefix + timestamp + File.separator;
+            String folderBase = (m_config.prefix.isEmpty() ? "" : m_config.prefix + "_") +
```
manjujames: Do you need to add the isEmpty() check here? If it is empty string, you are using an empty string. So, you could just use it as is..

manjujames:
This doesn't verify the log file case that you discovered where 2 logs with the same name exist in 2 different location. Other than that, the changes look good to me.

PR #3237, opened 2016-02-02
Eng 9853 sqlcov postgis jenkins

johnTcrawford:
Fixes the problem where an UPDATE was failing on Ubuntu 12.04 (only). Also adds some useful debug print options, and a few queries that Chris suggested in a previous review, which I had thought were already on master.

### DIFF
--- a/src/frontend/org/voltdb/PostgreSQLBackend.java
+++ b/src/frontend/org/voltdb/PostgreSQLBackend.java
@@ -17,6 +17,8 @@
 
 package org.voltdb;
 
+import java.io.FileWriter;
+import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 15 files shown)


src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -158,6 +162,22 @@ public PostgreSQLBackend(String databaseName, String username, String password)
         super("PostgreSQL", "org.postgresql.Driver",
                 "jdbc:postgresql:" + databaseName,
                 username, password);
+        // If '-Dsqlcoverage.transform.sql.file=...' was specified on the
+        // command line, print debug info when transforming SQL statements
+        // into a format that PostgreSQL can understand
+        String transformSqlOutputFile = System.getProperty("sqlcoverage.transform.sql.file", null);
```
wolffcm: I did not know you could access the ant command line this way.  Neat.

wolffcm:
Looks good to me.

PR #3238, opened 2016-02-03
Add DWithin geo function

hhakimi53:
DWithin function finds whether two given geo objects are within specified distance of each other. Supported arguments:
- First two input arguments are geo objects which can a point and a polygon, a point and a point. Both arguments can't be polygons as distance within two polygons is not supported
- Third arguments is non-negative distance argument on which two geo objects are evaluated whether they are within that distance range or not.


tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -809,19 +809,170 @@ public void testPolygonAsText() throws Exception {
...
+                 {"Wyoming",    "North Point Not On Wyoming Border",    12768.354425089678},
+                 {"Wyoming",    "Fort Collins",                         48820.514427535185},
+                }, vt1, DISTANCE_EPSILON);
+
+        // above result must be same as result fetched using DISTANCE function
+        sql = "select borders.name, places.name, distance(borders.region, places.loc) as distance "
+                + "from borders, places where (distance(borders.region, places.loc) <= 50000) and borders.pk = 1 "
+                + "order by distance, places.pk;";
+        vt2 = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertEquals(vt2,  vt1);
```
wolffcm: We have a sensible equals operator defined for VoltTable?  I didn't know this.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -809,19 +809,170 @@ public void testPolygonAsText() throws Exception {
...
+
+        // point-to-point
+        // get data using DWithin function
+        sql = "select A.name, B.name, distance(A.loc, B.loc) as distance "
+                + "from places A, places B where DWithin(A.loc, B.loc, 100000) and A.pk <> B.pk "
+                + "order by distance, A.pk;";
+        vt1 = client.callProcedure("@AdHoc", sql).getResults()[0];
+        // get data using DISTANCE function
+        sql = "select A.name, B.name, distance(A.loc, B.loc) as distance "
```
wolffcm: Are both of these queries deterministic?  Seems like you'll want to order by B.pk too.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -809,19 +809,170 @@ public void testPolygonAsText() throws Exception {
...
+                + "from places A, places B where distance(A.loc, B.loc) <= 100000 and A.pk <> B.pk "
+                + "order by distance, A.pk;";
+        vt2 = client.callProcedure("@AdHoc", sql).getResults()[0];
+        assertEquals(vt2,  vt1);
+
+        // test results of within using contains function
+        sql = "select borders.name, places.name "
```
wolffcm: Good idea doing it this way.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
wolffcm: I've similar comments some of the other queries here.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -809,19 +809,170 @@ public void testPolygonAsText() throws Exception {
...
+        String sql;
+        // DWITHIN between polygon and polygon is not supported
+        ProcCallException exception = null;
+        try {
+            sql = "select A.name, B.name from borders A, borders B where DWithin(A.region, B.region, 100);";
+            vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+        }
+        catch (ProcCallException excp) {
+            exception = excp;
+            assertTrue(exception.getMessage().contains("incompatible data type in operation"));
```
wolffcm: There is a method in RegressionSuite called verifyStmtFails that could make this code more concise.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
wolffcm: More importantly, is `equals` sufficient for the testing you're doing here?

wolffcm:
The only thing I see missing as far as testing is passing a NULL as the distance argument.  Perhaps I missed it.  Besides that, this looks good to me.  Definitely demo-able at the end of iteration meeting as well.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: Yes, it can produce result set in different order specially in valgrind environment; thanks for catching that. I will update the query so that result is deterministic.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: Forgot about the verifyStmtFails(). it will helpful here as it will make logic more concise and cleaner.

hhakimi53:
NULL argument for distance is missing. I will add the test case for that. Thnxs for the feedback.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: The intent of the logic is to make sure the result from both of the queries match to verify the DWithin is working as expected as well cross check our other geo functions :). Voltable::equals() is deprecated and is intended for testing use only as per the comments. Based on the logic it seems it will assert if contents of two table are not same. I will look into if can use some other call instead. The downside to equals is it will not report which table row is different. Please let me know if you had other thoughts on it. Thnxs

wolffcm: If there is no better alternative in RegressionSuite somewhere, then what you have here seems okay---but if `equals` is deprecated it suggests that it may go away someday?

hhakimi53: There is assertTablesAreEqual() in RegressionSuite to check if the contents of two tables are equal.  So can use that one instead.

wolffcm:
That's all my comments on your latest commits.

wolffcm:
Looks good.  Thanks Huzefa.  If this is ready to merge, you can demo it tomorrow?

hhakimi53:
Thanks Chris for reviewing it again. Sure, I can try to demo it tomorrow. There are no merge conflicts with master so we should be able to push the changes to the master.

PR #3243, opened 2016-02-03
Make ambiguous columns in order by expressions be errors.

BillWhite:
with the expressions which the aliases name.  This code had the same
stopped when it found one definition.  This means that ambiguous

throws errors.  Tests were added for group by, order by and having.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
@@ -1309,9 +1309,10 @@ public class ExpressionColumn extends Expression {
                 // Column references which name columns in tables in the
                 // from clause are handled later an.
                 java.util.Set<SelectListAliasResolution> foundNames = new java.util.TreeSet<SelectListAliasResolution>();
-                Expression lastExpr = null;
+                Expression firstFoundExpr = null;
                 for (int i = 0; i < length; i++) {
-                     // Ferret out the table name, column name
+                    ExpressionColumn ecol = (columns[i] instanceof ExpressionColumn) ? (ExpressionColumn)columns[i] : null;
+                    // Ferret out the table name, column name
                     // and alias name from this select column.
                     // If the alias name is null, then use the column name.
                     // This may be null as well.
@@ -1349,7 +1350,7 @@ public class ExpressionColumn extends Expression {
                             }
                         }
                     }
-
+                }
                 // If we only got one answer, then we just return it.
                 // If we got more than one, then we print an ambiguous
                 // error message.  If we got no message, we let HSQL
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
@@ -394,17 +394,11 @@ public class ExpressionColumn extends Expression {
                     //       in name order.
                     if (usingResolutions.size() > 0) {
                         sb.append("USING(");
-                        for (ColumnReferenceResolution crr : usingResolutions) {
-                            sb.append(sep).append(crr.getName());
-                            sep = ", ";
-                        }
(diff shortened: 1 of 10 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,6 +1007,106 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+        public String getTableName() {
+            return m_tableName;
+        }
+
+        @Override
+        public String toString() {
+            if (m_tableName == null) {
+                return m_name + "(" + m_index + ")";
+            } else {
+                return getName();
```
hhakimi53: Question: If tablename is not null, in that case is m_index defined (from constructor definition it looks that it will be)? Not sure how this function is intended to be used but if this is for debugging/information thn might be useful to return m_index value along with it

hhakimi53:
Besides the question above, overall changes looks good to me. Though I do not know much about this code-path, it will be good to have it reviewed by Paul or Chris also.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: This is actually for error messages, not debugging.  The indices are sometimes useful, but not in the tableName != null case.

The T.C case occurs when one references a select list element T.C in a column reference.  For example:

   select T.C, T.C from T where T.C > 0;

In this case, T.C is repeated twice, but it's always the same column in the same table.  So, the indices, which is to say the order in the select list, would be irrelevant. That's why the compareTo method only cares about names if there is a table name.  If we see this SQL:

    select C, C from T where C > 0;

We know C is not an alias, and so we look up C first, and then we look up C's table name.  So we will be in the non-null table name case.

If there is not a table name, that is to say if tableName == null, then we are looking at an alias.  This would be the case for this SQL:
    select T.C as CC, T.D as CC from T where CC > 0;
In this case the aliases CC are equal and we care about the indices.  We want to give an error message which says "CC occurs in columns: CC(1), CC(2)".  The numbers in parentheses are the indices.

hhakimi53: Sounds good, thnxs Bill. In case if we need to extend the usage of the function in future, we can update the function. Just a thought: a short comment in there might be helpful for the future.

BillWhite: I decided I liked my own pull request comment enough to add it to the class comment.

paulmartel: As far as I can tell, the constructor that could set a non-null m_tableName is never called,
so all of the code dealing with the non-null table name case is suspect (dead) and could be
eliminated, greatly simplifying this class.

I have other comments I could send, but they are relatively minor in light of this one which seems to have overriding priority.

paulmartel: BTW, my understanding is that a reference to a select list item from an ORDER BY is NEVER table qualified in the ORDER BY but SHOULD successfully match the unique select list item that has that name as its alias. Also, my understanding is that each select list element that is a table-qualified column name with no alias should be processed as if it HAS an (implicit) alias which is equal to its unqualified column name.  This model allows the sqlcmd tests to be valid AS THEY WERE.

paulmartel: My understanding MAY introduce ambiguities that do not arise in other possible resolution models, such as for:

    SELECT t.a b, t.b FROM t ORDER BY b;
    SELECT t.a b, b FROM t ORDER BY b;
    SELECT a b, b FROM t ORDER BY b;

My model favors neither column in these cases and flags them as ambiguous.

PR #3244, opened 2016-02-04
Eng 8932 backmerged full join

paulmartel:


src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+
+#include <vector>
+#include <limits>
+
+namespace voltdb {
+
+class Table;
+
+template<typename Value>
+class TableView_iter;
```
paulmartel: The "View" part of the name is going to cause confusion. How about TableFilter? TableSieve?

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+        uint64_t blockIndex = findBlockIndex(tupleAddress);
+        return (tupleAddress - m_prevBlockAddress) / m_tupleLength + blockIndex;
+    }
+
+    /**
+     * Initialize tuple and advance last active tuple index
+     * This method should be called only once during the initialization.
+     * To update tuple value use setTupleBit method
+     */
+    void initTupleBit(const TableTuple& tuple, char marker)
```
paulmartel: I don't see a lot of benefit in having a customizable initial bit value.
I'd just hard-code it to 0 -- it doesn't really have a meaning at this level
the "caller" can decide whatever the initial default state of the element (tuple) is
and associate an otherwise meaningless 0 value with THAT state by definition.

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+        TableView_iter<Value>
+      , Value
+      , boost::forward_traversal_tag>
+{
+    public:
+    /**
+     * Default constructor
+     */
+    TableView_iter(char marker = 0)
+      : m_tableView(0), m_tupleIdx(), m_marker(marker), m_tupleIdxPtr(&m_tupleIdx)
```
paulmartel: The ability to initialize a Value* from the address of a uint64_t pretty much assumes that Value IS uint64_t. 
If we even need the templating here -- and I'm not convinced -- I think we could minimize complexity by having a complete working non-template implementation based on uint64_t and (maybe) having a thin layer of a template class that inherits from or delegates to this simpler core implementation.

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+        return *m_tupleIdxPtr;
+    }
+
+    void moveToNextTuple()
+    {
+        uint64_t lastActiveTupleIndex = m_tableView->getLastActiveTupleIndex();
+        do
+        {
+            ++m_tupleIdx;
+        } while(m_tupleIdx <= lastActiveTupleIndex && m_tableView->getTupleBit(m_tupleIdx) != m_marker);
```
paulmartel: Rather than having to specify a marker value in so many places and storing it I would take one of two approaches. For now, if we only use one marker value, I would just ise a #define for it here rather than m_marker (which I would remove).  If we DO need iteration over two different values, I would define most of the class as a generic base class and have a derived template parameterized by MARKER_VALUE that implemented increment/moveToNextTuple.

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+     */
+    explicit TableView_iter(const TableView* m_tableView, size_t tupleIdx, char marker = 0)
+        :  m_tableView(m_tableView), m_tupleIdx(tupleIdx), m_marker(marker), m_tupleIdxPtr(&m_tupleIdx)
+    {}
+
+    // Forward Iteration Support
+    void increment()
+    {
+        moveToNextTuple();
+    }
```
paulmartel: I don't see a need for both moveToNextTuple and increment. I'd rename moveToNextTuple to increment.

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+        do
+        {
+            ++m_tupleIdx;
+        } while(m_tupleIdx <= lastActiveTupleIndex && m_tableView->getTupleBit(m_tupleIdx) != m_marker);
+    }
+
+    const TableView* m_tableView;
+    uint64_t m_tupleIdx;
+    char m_marker;
+    Value* m_tupleIdxPtr;
```
paulmartel: I don't understand what m_tupleIdxPtr is providing over m_tupleIdx.

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+};
+
+inline
+TableView_iterator TableView::begin(char marker)
+{
+    return TableView_iterator(this, marker);
+}
```
paulmartel: I don't see the benefit of defining inline methods outside the class body, especially short ones like these -- just one more place to have to maintain names and signatures.

paulmartel: Nevermind. I see the forward definition problem here.

src/ee/storage/temptable.cpp
```diff
@@ -93,4 +93,15 @@ std::string TempTable::tableType() const { return "TempTable"; }
 
 voltdb::TableStats* TempTable::getTableStats() { return NULL; }
 
+std::vector<uint64_t> TempTable::getBlockAddresses() const
+{
+    std::vector<uint64_t> blockAddresses;
+    blockAddresses.reserve(m_data.size());
+    for(std::vector<TBPtr>::const_iterator iter = m_data.begin(); iter != m_data.end(); ++iter)
+    {
```
paulmartel: Minor inconsistency of style -- add space between "for" and "(" with ending "{" on the same line

src/ee/storage/tableview.h
alexeevm: Or const uint64_t. This is needed to be able to create a const iterator - its return type must be "const uint64_t". Otherwise you will have to define almost identical TableView_constiter where the only difference would be a signature of the dereference method:
const uint64_t& dereference() const  vs uint64_t& dereference() const

This is the only reason to have the template parameter Value

src/ee/storage/tableview.h
alexeevm: Paul, consider the current scenario. When a new TableView is created, all its bits are set to be INACTIVE(-1). Once it is initialized from a table, the active tuples bits are set to ACTIVE(0). The NLJ/NLIJ loop marks all tuples that have a match by setting the corresponding bit to MATCHED(1). Finally after the NLJ/NJIL is done, we need to find tuples that still have the value == ACTIVE(0) - non-matched.
If we want to hardcode the value the iterator looks for, then this value must be the value used for active tuple initialization. It looks a bit counterintuitive to me, do you agree? It will definitely work in current scenario, but what if you need to find all tuples that were marked for whatever reason?

src/ee/storage/tableview.h
alexeevm: Done

src/ee/storage/tableview.h
alexeevm: Nothing, except I can't make it to compile if I use m_tupleIdx in the dereference method
    Value& dereference() const
    {
        return m_tupleIdx;
    }
Produces the error even if m_tupleIdx type is declared as Value:  binding of reference to type 'unsigned long long' to a
      value of type 'const unsigned long long' drops qualifiers
        return m_tupleIdx;

and I can't get around it without this indirection. I will try to eliminate it

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -54,7 +72,10 @@ public static AbstractExpression combine(Collection<AbstractExpression> exps) {
         }
         Stack<AbstractExpression> stack = new Stack<AbstractExpression>();
         stack.addAll(exps);
+        return combine(stack);
+    }
 
+    private static AbstractExpression combine(Stack<AbstractExpression> stack) {
```
paulmartel: These were my changes that got dropped in by accident. Sorry.
I'll make a new branch without them.

paulmartel: I just realized that this really is your change.
Just coincidentally it happens that in order to merge this code, I had to set aside my own branch which has an almost identical change!
So, no need for a new branch.

src/ee/storage/tableview.h
```diff
@@ -0,0 +1,299 @@
...
+    bool equal(TableView_iter const& other) const
+    {
+        // Shouldn't compare iterators from different tables and looking for different values
+        assert(m_tableView == other.m_tableView && m_marker == other.m_marker);
+        return m_tupleIdx == other.m_tupleIdx;
+    }
+
+    Value& dereference() const
+    {
+        return *m_tupleIdxPtr;
```
paulmartel: It's not perfect, but I think this does the job:
        return (Value&) m_tupleIdx;

src/ee/storage/tableview.h
alexeevm: Maybe TableTupleFilter?

src/ee/storage/tableview.h
paulmartel: Got it.
Worth a more explicit comment than "Const and Non-Const versions" like:
Value is expected to be one of "uint64_t" or "const uint64_t".

src/ee/executors/nestloopexecutor.cpp
```diff
@@ -70,6 +71,62 @@
...
+
+bool NestLoopExecutor::outputTuple(TableTuple& join_tuple, ProgressMonitorProxy& pmp) {
+    if (m_aggExec != NULL) {
+        return m_aggExec->p_execute_tuple(join_tuple);
+    }
+    m_tmpOutputTable->insertTempTuple(join_tuple);
+    pmp.countdownProgress();
+    return false;
+}
+
```
paulmartel: I relly like this refactoring -- maybe it's time to take this one step further and inject a common AbstractJoinExecutor base class above NestLoopExecutor and NesLoopIndexExecutor to hold the common members and common code like outputTuple and common helper classes like PredicateLimitEvaluator.

src/ee/executors/nestloopexecutor.cpp
```diff
@@ -70,6 +71,62 @@
 using namespace std;
 using namespace voltdb;
 
+static const char UNMATCHED_TUPLE(0);
+static const char MATCHED_TUPLE(1);
+
+struct PredicateLimitEvaluator
```
paulmartel: Naming suggestion: PredicateLimitEvaluator --> CountingPostfilter

src/ee/executors/nestloopexecutor.cpp
```diff
@@ -156,12 +226,11 @@ bool NestLoopExecutor::p_execute(const NValueArray &params) {
...
     TableTuple outer_tuple(node->getInputTable(0)->schema());
     TableTuple inner_tuple(node->getInputTable(1)->schema());
-    const TableTuple& null_tuple = m_null_tuple.tuple();
+    const TableTuple& null_inner_tuple = m_null_inner_tuple.tuple();
 
     TableIterator iterator0 = outer_table->iteratorDeletingAsWeGo();
-    int tuple_ctr = 0;
-    int tuple_skipped = 0;
     ProgressMonitorProxy pmp(m_engine, this);
+    PredicateLimitEvaluator whereEvaluator(wherePredicate, limit, offset);
```
paulmartel: rename whereEvaluator --> postfilter.

src/ee/executors/nestloopexecutor.cpp
```diff
@@ -172,8 +241,7 @@ bool NestLoopExecutor::p_execute(const NValueArray &params) {
         join_tuple = m_tmpOutputTable->tempTuple();
     }
 
-    bool earlyReturned = false;
-    while ((limit == -1 || tuple_ctr < limit) && iterator0.next(outer_tuple)) {
+    while (iterator0.next(outer_tuple) && whereEvaluator.isUnderLimit()) {
```
paulmartel: This code used to check the limit BEFORE fetching another tuple (potentially for nothing) -- that seems better as it was.

src/ee/storage/tableview.h
alexeevm: Yes, it works - it casts away the const qualifier. The problem here is that the 'dereference' method is a const method (required by the Boost) and it makes the 'this' object it operates on and any reference to its members const as well (uint64_t m_tupleIdx becomes const uint64_t) and an attempt to bind it to the  return type uint64_t& results in error.

I googled for a better solution but in all examples that I was able to find, iterators kept pointers to data  and dereference method retuned *data which compiles.

src/ee/executors/nestloopexecutor.cpp
```diff
@@ -224,33 +284,43 @@ bool NestLoopExecutor::p_execute(const NValueArray &params) {
...
-                        earlyReturned = true;
-                    }
-                } else {
-                    m_tmpOutputTable->insertTempTuple(join_tuple);
-                    pmp.countdownProgress();
+            if (whereEvaluator.eval(outer_tuple, null_inner_tuple)) {
+                // Matched! Complete the joined tuple with the inner column values.
+                join_tuple.setNValues(outer_cols, null_inner_tuple, 0, inner_cols);
+                if (outputTuple(join_tuple, pmp)) {
+                    whereEvaluator.setAboveLimit();
```
paulmartel: "break;" here?

src/ee/executors/nestloopexecutor.cpp
```diff
@@ -224,33 +284,43 @@ bool NestLoopExecutor::p_execute(const NValueArray &params) {
...
+            // Restore the tuple value
+            uint64_t tupleAddr = innerTableView.getTupleAddress(*itr);
+            inner_tuple.move((char *)tupleAddr);
+            // Still needs to pass the filter
+            assert(inner_tuple.isActive());
+            if (whereEvaluator.eval(null_outer_tuple, inner_tuple)) {
+                // Passed! Complete the joined tuple with the inner column values.
+                join_tuple.setNValues(outer_cols, inner_tuple, 0, inner_cols);
+                if (outputTuple(join_tuple, pmp)) {
+                    whereEvaluator.setAboveLimit();
```
paulmartel: "break;" here?

src/ee/executors/nestloopindexexecutor.cpp
```diff
@@ -507,68 +570,56 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
...
-                                if (m_aggExec->p_execute_tuple(join_tuple)) {
-                                    // Get enough rows for LIMIT
-                                    earlyReturned = true;
-                                    break;
-                                }
-                            } else {
-                                m_tmpOutputTable->insertTempTuple(join_tuple);
-                                pmp.countdownProgress();
+                            if (outputTuple(join_tuple, pmp)) {
+                                whereEvaluator.setAboveLimit();
```
paulmartel: "break;"?

src/ee/executors/nestloopindexexecutor.cpp
```diff
@@ -507,68 +570,56 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
...
-                    // For the sake of consistency, we don't try to do
-                    // output expressions here with columns from both tables.
-                    join_tuple.setNValue(col_ctr,
-                                         m_outputExpressions[col_ctr]->eval(&outer_tuple, &null_tuple));
+            // Still needs to pass the filter
+            if (whereEvaluator.eval(outer_tuple, null_inner_tuple)) {
+                // Matched! Complete the joined tuple with the inner column values.
+                join_tuple.setNValues(num_of_outer_cols, null_inner_tuple, 0, num_of_inner_cols);
+                if (outputTuple(join_tuple, pmp)) {
+                    whereEvaluator.setAboveLimit();
```
paulmartel: "break;'?

src/ee/executors/nestloopindexexecutor.cpp
```diff
@@ -507,68 +570,56 @@ bool NestLoopIndexExecutor::p_execute(const NValueArray &params)
...
+            // Restore the tuple value
+            uint64_t tupleAddr = innerTableView.getTupleAddress(*itr);
+            inner_tuple.move((char *)tupleAddr);
+            // Still needs to pass the filter
+            assert(inner_tuple.isActive());
+            if (whereEvaluator.eval(null_outer_tuple, inner_tuple)) {
+                // Passed! Complete the joined tuple with the inner column values.
+                join_tuple.setNValues(num_of_outer_cols, inner_tuple, 0, num_of_inner_cols);
+                if (outputTuple(join_tuple, pmp)) {
+                    whereEvaluator.setAboveLimit();
```
paulmartel: "break;"?

src/ee/executors/nestloopindexexecutor.cpp
```diff
@@ -72,6 +73,63 @@
...
+};
+
+bool NestLoopIndexExecutor::outputTuple(TableTuple& join_tuple, ProgressMonitorProxy& pmp) {
+    if (m_aggExec != NULL) {
+        return m_aggExec->p_execute_tuple(join_tuple);
+    }
+    m_tmpOutputTable->insertTempTuple(join_tuple);
+    pmp.countdownProgress();
+    return false;
+}
```
paulmartel: The meaning of the return value from m_aggExec->p_execute_tuple was already obscure.
Wrapping it in outputTuple just makes it more so.
A comment near (before or after) each call to outputTuple is justified,
reminding that the return value of "true" signals a reached limit as enforced by an inline agg.

src/ee/storage/table.h
```diff
@@ -377,6 +378,9 @@ class Table {
                                      "May not use freeLastScanedBlock with streamed tables or persistent tables.");
     }
 
+    // Return tuple blocks addresses
+    virtual std::vector<uint64_t> getBlockAddresses() const = 0;
+
```
paulmartel: I'm growing tired of these abstract calls supposedly general to table classes but actually only common to temptable and persistenttable and not really applicable to streamedtable which gives a dummy response or a "not implemented" runtime exception. Maybe I'm looking for a common base class for just temp and persistent table -- something like readabletable? queriabletable? scannabletable? Though I suppose any such refactoring can wait.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2833,6 +2844,35 @@ private static void simplifyOuterJoinRecursively(BranchNode joinNode, List<Abstr
...
+            }
+            if (leftNode instanceof BranchNode) {
+                newExprs.addAll(exprs);
+                simplifyOuterJoinRecursively((BranchNode)leftNode, newExprs);
+            }
+        }
+    }
+
+    /**
+     * Verify if an expression from the input list is NULL-rejecting for any of the table from the list
```
paulmartel: table --> tables

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2833,6 +2844,35 @@ private static void simplifyOuterJoinRecursively(BranchNode joinNode, List<Abstr
...
+     * @param tableAliases list of tables
+     * @param exprs list of expressions
+     * @return TRUE if there is a NULL-rejection expression
```
paulmartel: rejection --> rejecting

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -133,41 +133,123 @@ private static void queueSubJoinOrders(List<ArrayList<JoinNode>> joinOrderList,
...
+    }
+
+    /**
+     * Helper method to generate join orders for an OUTER join tree.
+     * At the moment, permutations for LEFT Joins are not supported yet
+     *
+     * @param subTree join tree
+     * @return list of valid join orders
+     */
+    private static List<JoinNode> generateOuteroinOrdersForTree(JoinNode subTree) {
```
paulmartel: generateOuteroinOrdersForTree --> generateOuterJoinOrdersForTree

tests/frontend/org/voltdb/expressions/TestExpressionUtil.java
```diff
@@ -294,7 +294,7 @@ public void testCombine() {
         // sure that all of our sub-trees are contained within the new tree and that their
         // structure has not changed
         //
-        new TestExpressionTreeWalker() {
+        TestExpressionTreeWalker tertWalker = new TestExpressionTreeWalker() {
```
paulmartel: tertWalker --> treeWalker

tests/frontend/org/voltdb/planner/TestMultipleOuterJoinPlans.java
```diff
@@ -26,296 +26,482 @@
...
+        assertEquals(2, lpn.size());
+        n = lpn.get(1).getChild(0);
+        verifyJoinNode(n, PlanNodeType.NESTLOOPINDEX, JoinType.FULL, null, null, null, PlanNodeType.SEQSCAN, PlanNodeType.INDEXSCAN, "P1", "P2");
+        verifyIndexScanNode(n.getInlinePlanNode(PlanNodeType.INDEXSCAN), IndexLookupType.EQ, ExpressionType.COMPARE_GREATERTHAN);
+
+        // FULL join of two partition tables on non-partition column
+        failToCompile("select * FROM  P1 FULL JOIN P4 ON P1.C = P4.A ",
+                "Join of multiple partitioned tables has insufficient join criteria");
+}
+
```
paulmartel: Somewhere, one of these test cases should show that we support the explicit syntax: "full OUTER join" as synonymous with the terse "FULL JOIN".

src/ee/storage/temptable.cpp
alexeevm: corrected

src/frontend/org/voltdb/expressions/ExpressionUtil.java
alexeevm: Yes, I have introduced a new signature to be able to combine expressions from multiple collections without explicitly creating a temp collection

src/ee/executors/nestloopexecutor.cpp
alexeevm: corrected

src/ee/executors/nestloopexecutor.cpp
alexeevm: Technically it's not required because the outer while loop condition checks for the evaluator to be under the limit 
while (whereEvaluator.isUnderLimit() && iterator0.next(outer_tuple)) {

The only danger there if some logic could be introduced later to the OUTER WHILE LOOP right after the end of the LEFT OUTER JOIN block that is not supposed to be executed if the limit is reached. Is that what you are concerned about?

This addresses your 4 other similar comments. I think I have test cases to cover all these above the limit cases. I will double check.

src/ee/storage/persistenttable.cpp
```diff
@@ -1714,4 +1714,13 @@ void PersistentTable::computeSmallestUniqueIndex() {
     }
 }
 
+std::vector<uint64_t> PersistentTable::getBlockAddresses() const {
+    std::vector<uint64_t> blockAddresses;
+    blockAddresses.reserve(m_data.size());
+    for(TBMap::const_iterator i = m_data.begin(); i != m_data.end(); ++i) {
+            blockAddresses.push_back((uint64_t)i.value()->address());
```
paulmartel: This line would not compile (no member named value) until I changed it to:

            blockAddresses.push_back((uint64_t)(i->second->address()));

paulmartel:
The java code looks good. I've finished my review.
I tweaked it to compile (on a separate branch, not to confuse the code review) and pushed it to our regression test system. I'll let you know how that goes.
Let me know when you have a revised version and I'll re-submit it to the test suite.

src/ee/storage/tableview.h
alexeevm: initTupleBit (by the way, I renamed it to be initActiveTuple) is called from the public method
TableTupleFilter::init(Table *table, char initValue) where the second parameter is used to initialize all the active tuples. I agree, that the initial value can be hard-coded to 0 with the onus on the caller to mark tuples with some other value to be able to filter them later. The interface becomes

public void TableTupleFilter::init(Table *table);
private void TableTupleFilter::initActiveTuple(const TableTuple& tuple);

src/ee/storage/tableview.h
alexeevm: Updated

src/ee/executors/nestloopexecutor.cpp
alexeevm: I agree and will try to refactor. The other good candidate is the IndexScan/NestLoopIndexScan pair. Do you think to address them now as well?

src/ee/executors/nestloopexecutor.cpp
alexeevm: Renamed

src/ee/executors/nestloopexecutor.cpp
alexeevm: Renamed

src/ee/executors/nestloopindexexecutor.cpp
alexeevm: I maid the CountingPostfilter class to be a private inner class of the NestLoopExecutor (will be moved up to a new parent) and the post filter is a private member of the executor class now. The updated outputTuple now doesn't need to return anything:

void NestLoopExecutor::outputTuple(TableTuple& join_tuple, ProgressMonitorProxy& pmp) {
    if (m_aggExec != NULL && m_aggExec->p_execute_tuple(join_tuple)) {
        m_postfilter.setAboveLimit();
    }
    m_tmpOutputTable->insertTempTuple(join_tuple);
    pmp.countdownProgress();
}

src/frontend/org/voltdb/planner/PlanAssembler.java
alexeevm: corrected

src/frontend/org/voltdb/planner/PlanAssembler.java
alexeevm: corrected

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
alexeevm: corrected

tests/frontend/org/voltdb/expressions/TestExpressionUtil.java
alexeevm: corrected

tests/frontend/org/voltdb/planner/TestMultipleOuterJoinPlans.java
alexeevm: Yes, I will add the test

src/ee/storage/tableview.h
alexeevm: The updated signature for the iterator
template<int8_t MARKER, typename Value>
class TableTupleFilter_iter;
Using the C++11 syntax, the const and non-const iterators are defined as
// Non-const iterator
template<int8_t MARKER>
using TableTupleFilter_iterator = TableTupleFilter_iter<MARKER, uint64_t>;
// Const iterator
template<int8_t MARKER>
using TableTupleFilter_const_iterator = TableTupleFilter_iter<MARKER, uint64_t const>;

paulmartel:
FULL JOIN was merged from a rebased branch.

PR #3245, opened 2016-02-04
ENG-9858 Fix JDBC valid float range

dweiss88:


akhanzode:
Did u run the attached sample in ticket with this fix? also Math.abs(value) > 0 looks kinda odd check right?

manjujames:
Looks fine to me (as long as it was also tested with the sample as mentioned by Anish).

PR #3246, opened 2016-02-04
ENG-9884: Move Durable Snapshot state for new features to a single class

wweiss-voltdb:


luyangco:
Changes looks good to me, I'll land it once all jenkins tests pass.

PR #3247, opened 2016-02-04
ENG-9785: Voltdb crash if user name contains white space

JadenW:


akhanzode:
Add test for UpdateApplicationCatalog to ensure it does not crash.

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -393,15 +393,27 @@
...
   </xs:complexType>
 
+  <xs:simpleType name="userRoleNameString">
+	<xs:restriction base="xs:string">
+      <xs:pattern value="[A-Za-z0-9,_@\-\.\*]+"/>
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="passwordString">
+	<xs:restriction base="xs:string">
```
vtkstef: passwords may contain whitespaces does this pattern preclude this?

tests/frontend/org/voltdb/utils/TestCatalogUtil.java
```diff
@@ -364,6 +364,60 @@ public void testUserRoles() throws Exception {
...
+        final String badPassword = "<?xml version='1.0' encoding='UTF-8' standalone='no'?>" +
+                "<deployment>" +
+                "<security enabled=\"true\"/>" +
+                "<cluster hostcount='3' kfactor='1' sitesperhost='2'/>" +
+                "<paths><voltdbroot path=\"/tmp/" + System.getProperty("user.name") + "\" /></paths>" +
+                "<httpd port='0'>" +
+                "<jsonapi enabled='true'/>" +
+                "</httpd>" +
+                "<users> " +
+                "<user name=\"fancypants\" password=\"ad min\" roles=\"administrator\"/>" +
```
vtkstef: passwords can contain whitespaces this password should not fail

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -393,15 +393,27 @@
...
+          <xs:attribute name="password" type="passwordString" use="required"/>
           <xs:attribute name="plaintext" type="xs:boolean" default="true"/>
         </xs:complexType>
       </xs:element>
     </xs:sequence>
   </xs:complexType>
 
+  <xs:simpleType name="userRoleNameString">
+	<xs:restriction base="xs:string">
+      <xs:pattern value="[A-Za-z0-9,_@\-\.\*]+"/>
```
pdshaw: Who said a user name can't have embedded spaces? If defined by our interface designer and doc'd, cool. Otherwise, need to allow it and handle it carefully in code. Regardless, needs to be handled in code in case the password comes in through VMC (maybe) or other interfaces down the road.

Wrt the RE itself, does shorthand for [A-Za-z0-9_], "\w" work in this context?

pdshaw:
Fine though short-sighted if we disallow spaces in user names, but regardless, the coding error that led to the crash should be corrected so it's not a "land mine" waiting til a embedded space comes in from somewhere else.

tests/frontend/org/voltdb/utils/TestCatalogUtil.java
```diff
@@ -364,6 +364,60 @@ public void testUserRoles() throws Exception {
...
+            "<paths><voltdbroot path=\"/tmp/" + System.getProperty("user.name") + "\" /></paths>" +
+            "<httpd port='0'>" +
+            "<jsonapi enabled='true'/>" +
+            "</httpd>" +
+            "<users> " +
+            "<user name=\"fancy pants\" password=\"admin\" roles=\"administrator\"/>" +
+            "</users>" +
+            "</deployment>";
+
+        final String badPassword = "<?xml version='1.0' encoding='UTF-8' standalone='no'?>" +
```
akhanzode: this test should pass now right? password can contain spaces.

PR #3255, opened 2016-02-08
VDM-40: Send server start requests to the correct server.

manjujames:


lib/python/vdm/server/HTTPListener.py
```diff
@@ -1889,6 +1914,7 @@ def main(runner, amodule, config_dir, server):
     SERVER_VIEW = ServerAPI.as_view('server_api')
     DATABASE_VIEW = DatabaseAPI.as_view('database_api')
     START_DATABASE_SERVER_VIEW = StartServerAPI.as_view('start_server_api')
+    START_LOCAL_SERVER_VIEW = StartLocalServerAPI.as_view('start_local_server_api')
```
akhanzode: Don't you need similar for recover? How does start local know its "start" or "recover"

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1633,8 +1633,13 @@ def put(database_id):
             Status string indicating if the database start request was sent successfully
         """
 
-        database = voltdbserver.VoltDBServer(database_id)
-        return database.start_database()
+        try:
+            database = voltdbserver.VoltDatabase(database_id)
+            return database.start_database(True)
```
akhanzode: Whats the purpose of flag?

lib/python/vdm/server/voltdbserver.py
```diff
@@ -128,6 +128,24 @@ def start_database(self, recover=False):
         else:
             return create_response('Start request sent successfully to servers: ' + str(server_status), 200)
 
+    def start_server(server_id, recover=False):
```
akhanzode: Why not be explicit in action to use?

lib/python/vdm/server/voltdbserver.py
```diff
@@ -128,6 +128,24 @@ def start_database(self, recover=False):
...
+        """
+        Sends start request to the specified server
+        """
+        action = 'start'
+        if recover:
+            action = 'recover'
+
+        try:
+            server = find_server(server_id)
```
akhanzode: what if find_server fails?

akhanzode:
Rest looks alright.

lib/python/vdm/server/voltdbserver.py
manjujames: This is called with false for startup and true for recover. Defaulting in the method declaration is just a convenience..

lib/python/vdm/server/HTTPListener.py
manjujames: The flag says whether it is start in recover mode or not.

lib/python/vdm/server/HTTPListener.py
manjujames: There is no public endpoint to recover a *server*. When we get request to recover a database, we call the internal endpoint directly.

manjujames:
squash merged.

PR #3257, opened 2016-02-09
Eng 9829

hhakimi53:
Updated logic so that search key of length greater than the defined size of indexed variable-length column (varchar and varbinary) can be used in where clause. 

Logic flow updates is as: In index exceutor, when a search key, which is of greater length than it's variable length indexed column, is tried to be inserted into search tuple, a SQL exception with  state set to "TYPE_VAR_LENGTH_MISMATCH" is generated. Sensing generated SQL exception during insertion of search key into search tuple, index executor logic generates a new shrinked using original search key. For this logic computes the size to the shrink the search key based on defined length of variable length column taking into account if the length is defined in character (varchar) or bytes (varchar/varbinary). Along with that, the index lookup operator is updated as below:
- if lookup operator was "<" or "<=", it is updated to "<=" 
-  if lookup operator was ">=" or ">", it is updated to ">"
Logic than follows the usual code path to perform the indexed search using the updated search key and lookup operator.

For example, for schema below:
CREATE TABLE T ( C1 bigint NOT NULL, three_bytes varchar(3 BYTES));
CREATE INDEX IDX ON T (three_bytes); 

Query A: "Select three_bytes from T where three_bytes >= 'abbd';"
Query B: "Select three_bytes from T where three_bytes > 'abb';"
Result of Query A should provide  result that is same as result of Query B.

Similary for results for queries A and B should be same.
Query A: "Select three_bytes from T where three_bytes < 'abbd';"
Query B: "Select three_bytes from T where three_bytes <= 'abb';"


wolffcm:
I'm looking at this now.

wolffcm:
Do we have a similar problem with joins?  I'm wondering if we need similar logic in `NestLoopIndexExecutor`.
```SQL
create table t0 (long_str varchar(300));

create table t1 (short_str varchar(3));
create index t1idx on t1(short_str);

insert into t0 values ('foofoofoofoo');
insert into t1 values ('foo');

select ...
from t0, t1
where t0.short_str = t1.long_str;
```

hhakimi53:
I was thinking the same that we may have issue with joins but with current logic in NestLoopIndexExecutor, atleast joins on different variable length columns are working fine . 

For the we have a test case as below:
CREATE TABLE INDEXED_VL_TABLE1 (
  ID INTEGER DEFAULT 0 NOT NULL PRIMARY KEY,
  VC1 VARCHAR(3),
  VC2 VARCHAR(3 BYTES),
...
);

CREATE TABLE INDEXED_VL_TABLE2 (
  ID INTEGER DEFAULT 0 NOT NULL PRIMARY KEY,
  VC1 VARCHAR(6),
  VC2 VARCHAR(6 BYTES),
 ...
);
-- create index on varchar
CREATE INDEX VC1_INDEXED_VL1 ON INDEXED_VL_TABLE1 (VC1);
CREATE INDEX VC2_INDEXED_VL1 ON INDEXED_VL_TABLE1 (VC2);
...

CREATE INDEX VC1_INDEXED_VL2 ON INDEXED_VL_TABLE2 (VC1);
CREATE INDEX VC2_INDEXED_VL2 ON INDEXED_VL_TABLE2 (VC2);
...

Insert into INDEXED_VL_TABLE1 values(0, 'aaa', 'bbb' ...);
Insert into INDEXED_VL_TABLE2 values(0, 'aaaa', 'bbbb' ...);

Select * 
from INDEXED_VL_TABLE1 A, INDEXED_VL_TABLE2 B 
where B.VC1 < A.VC1 
order by ...;.

It works fine.

I just tried following query as below with the changes and that also works fine:
Select ...
from INDEXED_VL_TABLE1 A, INDEXED_VL_TABLE2 B 
where B.VC2 < A.VC2 
order by ...;

wolffcm:
If I understand the code correctly, the nest loop index join executor uses the same index API as index scan executor.  Any ideas why this bug occurs for index scans but not for joins?

src/ee/executors/indexscanexecutor.cpp
```diff
@@ -254,11 +260,70 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
...
                 }
+                if (e.getInternalFlags() & SQLException::TYPE_VAR_LENGTH_MISMATCH) {
+                    // form new search key to add it to search key table tuple from original candidate
+                    // search key. To form new key, use the search column's length in bytes from org table
+                    // by fetching column size. Shrink the candidate key to fit within the column size
+                    // note: for varchar the column size can be in bytes or character. So logic
+                    // takes into account whether variable length column size is in bytes or not to
+                    // calculate new length in bytes.
+
+                    const TupleSchema::ColumnInfo *colInfo = tableIndex->getKeySchema()->getColumnInfo(ctr);
```
wolffcm: One could imagine adding a method to NValue called `shrinkTo(ColumnInfo)` to shrink the key value.  Or something like that.  This logic seems a bit low-level to be in a plan node executor.

I guess the under- and overflow logic is in the same boat, so I realize that you are following the pattern of what was already here.  If you don't want to fix this now, feel free to write a ticket.

hhakimi53: Yes, the logic that is getting exposed is bit of low level for the executor and I had tried initially to refactor the logic but was losing the track and started refocusing on fixing the issue first. I will try if it refactoring is not that involved, I will do it as part of the same ticket else will write a different ticket for it.

src/ee/executors/indexscanexecutor.cpp
```diff
@@ -254,11 +260,70 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
...
+                    const char* candidateValueBuffPtr = ValuePeeker::peekObject_withoutNull(candidateValue, &candidateValueLength);
+                    // calculate the length for shrinked candidate key
+                    int32_t neededLength;
+                    if (inBytes) {
+                        // updated length of key will be equivalent to column length
+                        neededLength = colLength;
+                    }
+                    else {
+                        // column length is in terms of characters. Obtain the number of bytes needed for those many characters
+                        neededLength = static_cast<int32_t> (NValue::getIthCharPosition(candidateValueBuffPtr, candidateValueLength, colLength + 1) - candidateValueBuffPtr);
```
wolffcm: This call to `getIthCharPosition` could be costly for large strings.  I wonder would it be possible to avoid it somehow?  I guess there isn't since the column width is a character count, and characters may or may not be multi-byte.

wolffcm:
This looks good to me.

I wonder, did you consider just making length of varchar/varbinary components very large in the index key schema?  Probably there's something I don't understand here.

hhakimi53:
One of the things I had discovered that in NestLoopIndexExecutor, the variable length column in search tuple formed is based on the schema of the table which is of larger size, so the problem did not arise there where as in IndexScanExecutor that variable length column in search tuple was based of the table available and length of  variable column was not big enough to accomodate the search key. I didn't investigate NestLoopIndexExecutor for how it's picking up the schema of the larger further after this. 

When going through about the current approach with Paul, the approach for extending variable length columns for indexing tuples had came up and did not went that route for varbinary/varchar as we would have flexibility of inlining variable length columns thn in search tuples. There are still number of components in indexing executor that I don't understand fully and more of in learning phase :-). 

Thanks for reviewing the  changes. I have extended the test case for join and will push that in and see if shrink key logic in index executor can be refactored else will file a ticket for that.

hhakimi53:
Chris, we have issue in NestLoopIndexExecutor also, our intuition was right on that. For join operation, I swapped the tables in from clause and it produces the same issue as we were with IndexScanExecutor. I will update the NestLoopIndexExecutor logic also to handle this case. Thnxs

wolffcm:
Thanks for looking into it, Huzefa.  Glad to know we caught it.

src/ee/executors/indexscanexecutor.cpp
paulmartel: I don't think that using very long strings as index keys will be that common.

wolffcm:
I like the refactoring, and the added tests.  Looks good to me.  Thanks for making the additional changes.

hhakimi53:
Thanks Chris for reviewing the changes. Jenkins is blue; I will merge  the changes soon.

PR #3258, opened 2016-02-09
Eng 9828 Make Ambiguous Column References in Order By be errors.

BillWhite:
with the expressions which the aliases name. This code had the same
stopped when it found one definition. This means that ambiguous

throws errors. Tests were added for group by, order by and having.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1553,24 +1553,7 @@ NATIVE EE STUFF
     <antcall target="prep_libvoltdb"/>
 </target>
 
-<target name='ee_makefile.uptodate'>
-  <condition property="makefile.uptodate">
-    <and>
-    <!--set if makefile up to date and makefile non-zero length
-            Eliminates timeconsuming build.py in ee -->
-    <uptodate>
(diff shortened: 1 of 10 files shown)


build.xml
```diff
@@ -1553,24 +1553,7 @@ NATIVE EE STUFF
...
-      <mapper type="merge" to="${build.dir}/makefile"/>
-    </uptodate>
-    <length file="${build.dir}/makefile" when="greater" length="0" />
-    </and>
-  </condition>
-</target>
-
-<target name='ee_buildmakefile' depends="catalog, jnicompile, buildinfo, ee_makefile.uptodate"
-    unless="makefile.uptodate"
+<target name='ee_buildmakefile' depends="catalog, jnicompile, buildinfo"
```
BillWhite: This is actually a fix for ENG-9801, which I swept into this.

BillWhite:
I'm going to make a small change to the sqlcommand test.  I'm closing this.

PR #3259, opened 2016-02-09
Eng 9828 Make Ambiguous Column References in order by expressions be errors.

BillWhite:
with the expressions which the aliases name. This code had the same
stopped when it found one definition. This means that ambiguous

throws errors. Tests were added for group by, order by and having.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1553,24 +1553,7 @@ NATIVE EE STUFF
     <antcall target="prep_libvoltdb"/>
 </target>
 
-<target name='ee_makefile.uptodate'>
-  <condition property="makefile.uptodate">
-    <and>
-    <!--set if makefile up to date and makefile non-zero length
-            Eliminates timeconsuming build.py in ee -->
-    <uptodate>
-      <srcfiles dir='${src.ee.dir}' includes='**/*' />
-      <srcfiles dir='${vendor.cpp.dir}'  includes='**/*' />
-      <srcfiles dir='${src.ee.test.dir}'  includes='**/*' />
-      <mapper type="merge" to="${build.dir}/makefile"/>
-    </uptodate>
-    <length file="${build.dir}/makefile" when="greater" length="0" />
-    </and>
-  </condition>
-</target>
-
-<target name='ee_buildmakefile' depends="catalog, jnicompile, buildinfo, ee_makefile.uptodate"
-    unless="makefile.uptodate"
+<target name='ee_buildmakefile' depends="catalog, jnicompile, buildinfo"
     description="Just make the makefile for the C++ JNI library.">
     <exec dir='.' executable='python' failonerror='true'>
         <env key="EEONLYBUILDMAKEFILE" value="true"/>
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -59,6 +59,10 @@
 #include <vector>
 #include <jsoncpp/jsoncpp.h>
 
+#ifndef NDEBUG
+#include "debuglog.h"
+#endif /* !define(NDEBUG) */
+
 class CopyOnWriteTest_TestTableTupleFlags;
 
 namespace voltdb {
@@ -110,7 +114,12 @@ public:
      * backing store
      */
     inline void move(void *address) {
-        assert(m_schema);
+#ifndef  NDEBUG
+        if (m_schema == NULL && address != NULL) {
+            StackTrace::printStackTrace();
+        }
+#endif
+        assert(m_schema != NULL || address == NULL);
         m_data = reinterpret_cast<char*> (address);
     }
 
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
@@ -394,17 +394,11 @@ public class ExpressionColumn extends Expression {
                     //       in name order.
                     if (usingResolutions.size() > 0) {
                         sb.append("USING(");
-                        for (ColumnReferenceResolution crr : usingResolutions) {
-                            sb.append(sep).append(crr.getName());
-                            sep = ", ";
-                        }
+                        appendNameList(sb, usingResolutions, "");
                         sb.append(")");
                         sep = ", ";
                     }
-                    for (ColumnReferenceResolution crr : rangeVariableResolutions) {
-                        sb.append(sep).append(crr.getName());
-                        sep = ", ";
-                    }
+                    appendNameList(sb, rangeVariableResolutions, sep);
                     throw new HsqlException(sb.toString(), "", 0);
                 }
                 // If we get here we didn't find anything.  So, add this expression
@@ -419,6 +413,19 @@ public class ExpressionColumn extends Expression {
         return unresolvedSet;
     }
 
+    /*
+     * Append the names of all the elements in the set of resolutions to the
+     * string buffer.  This is only used for error messages.
+     */
+    private <T> void appendNameList(StringBuffer sb,
+                                    java.util.Set<T> resolutions,
+                                    String sep) {
+        for (T oneRes : resolutions) {
+            sb.append(sep).append(oneRes.toString());
+            sep = ", ";
+        }
+    }
+
     /**
      * Return a sort of closure which is useful for resolving a column reference.
      * The column reference is either an expression or a column in a table, which
@@ -428,7 +435,7 @@ public class ExpressionColumn extends Expression {
      * until we have examined all of them.  So, we defer changing this object
      * until we are more sure of the reference.
      *
-     * We store these in a java.util.HashSet.  So we need to have our own
+     * We store these in a java.util.TreeSet.  So we need to have our own
      * notion of equality.
      */
     private interface ColumnReferenceResolution {
@@ -484,6 +491,7 @@ public class ExpressionColumn extends Expression {
         }
 
     }
+
     /**
      * This class implements the interface for expression columns.
      * An expression column is created for a "USING(C)" join condition.
@@ -612,7 +620,7 @@ public class ExpressionColumn extends Expression {
 
         @Override
         public String getName() {
-            // We prefer to aliases.  If we can't find an
+            // We prefer to use aliases.  If we can't find an
             // alias, we use the table name.
             if (m_rangeVariable.tableAlias != null && m_rangeVariable.tableAlias.name != null) {
                 return m_rangeVariable.tableAlias.name;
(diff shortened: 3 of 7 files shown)


build.xml
```diff
@@ -1553,24 +1553,7 @@ NATIVE EE STUFF
...
-      <mapper type="merge" to="${build.dir}/makefile"/>
-    </uptodate>
-    <length file="${build.dir}/makefile" when="greater" length="0" />
-    </and>
-  </condition>
-</target>
-
-<target name='ee_buildmakefile' depends="catalog, jnicompile, buildinfo, ee_makefile.uptodate"
-    unless="makefile.uptodate"
+<target name='ee_buildmakefile' depends="catalog, jnicompile, buildinfo"
```
BillWhite: This is a change to address ENG-9801.

src/ee/common/tabletuple.h
```diff
@@ -59,6 +59,9 @@
 #include <vector>
 #include <jsoncpp/jsoncpp.h>
 
+#include <execinfo.h>
+#include <cxxabi.h>
+
```
BillWhite: Maybe these two includes should be surrounded by #ifndef NDEBUG.

tests/frontend/org/voltdb/regressionsuites/TestGroupByComplexSuite.java
```diff
@@ -116,6 +116,218 @@ private void strangeCasesAndOrderby() throws IOException, ProcCallException {
         }
     }
 
+
```
BillWhite: The test below causes an assertion failure in tabletuple.h.

tests/frontend/org/voltdb/regressionsuites/TestGroupByComplexSuite.java
```diff
@@ -222,8 +434,8 @@ private void complexAggsOrderbySuite() throws IOException, ProcCallException {
 
             // (3) Test Order by with FUNCTION expression, no group by column in display columns
             // Test Order by without tag
```
paulmartel: Why does the comment here say "without tag"?

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -489,7 +489,7 @@ public void testJoinsOfInExists() throws Exception
                     "where T1.id = T2.id " +
                     "and exists " +
                     "    (select 1 from R1 where R1.dept * 2 = T2.dept) " +
-                    "order by id;";
+                    "order by t1.id;";
```
paulmartel: I still posit that the code as it stood should (also) be treated as unambiguous because id matches T1.id's (implicit) alias as the only column in the select list having that alias. This assumes the simple conventions that:
    -- an implicit tag/alias has equal weight to an explicit tag/alias in the select list
    -- the select list takes precedence over the from-clause-based column list when resolving (unqualified) tags/aliases in the order by clause

Agreed?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+     *
+     *   select lr.b b, rr.b b from r1 lr, r2 rr order by b;
+     *   -- b is ambiguous.  There are two aliases named b.
+     *   select lr.b b, b from r1 lr join r1 rr using (b) order by b;
+     *   -- Mysql and postgresql differ in this case.  It may
+     *   -- be because of different rules for using.
+     *   -- This is ambiguous with postgresql, but not
+     *   -- with mysql.  A close reading of the standard might
+     *   -- suggest that the first reference to lr.b is illegal,
+     *   -- since the using removes a from the table expression
```
paulmartel: ? Complete the comment:  "since the using removes a (???) from the table expression columns"

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1010,36 +1296,74 @@ Expression replaceAliasInOrderBy(Expression[] columns, int length) {
...
 
             case OpTypes.COALESCE :
             case OpTypes.COLUMN : {
+                // Look through all the columns in columns.  These are
+                // the select list, and they may have aliases equal to
+                // the column which we are trying to resolve.  In that
+                // case we really want to use the expression in the
+                // select list.  Note that we only look for aliases here.
+                // Column references which name columns in tables in the
+                // from clause are handled later an.
```
paulmartel: Correct comment: an --> on.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1010,36 +1296,74 @@ Expression replaceAliasInOrderBy(Expression[] columns, int length) {
...
+                                if (firstFoundExpr == null) {
+                                    firstFoundExpr = columns[i];
+                                }
+                            }
                         }
                     }
                 }
+                // If we only got one answer, then we just return it.
+                // If we got more than one, then we print an ambiguous
+                // error message.  If we got no message, we let HSQL
```
paulmartel: got no message --> got no answer

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+            // use the column name as an alias.  If
+            // m_columnName is null this is ok, and can
+            // happen in statements like: SELECT T.X * R.Y AS PROD FROM T, R
+            // where we don't have any canonical table name or column name.
+            if (m_alias == null) {
+                m_alias = getColumnName();
+            }
+        }
+
+        public final String getTableAlias() {
```
paulmartel: I don't see the rationale for public (not that it matters for nested classes) accessors for use solely within the nested class. Direct access to the member would suffice. Better to distinguish read access from write access (only in the constructor) by declaring members final.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+            return m_tableAlias;
+        }
+
+        public final String getAlias() {
```
paulmartel: AFAIK "final" declarations on methods is in most cases just noise -- modern java compilers are able to infer the non-existence of overrides and optimize accordingly. It's the MEMBERS that could usefully be declared final for the sake of comprehension of the flow.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+         */
+        @Override
+        public int compareTo(SelectListAliasResolution o) {
+            ExpressionColumn ecol = getExpressionColumn();
+            ExpressionColumn oecol = o.getExpressionColumn();
+            if (ecol == null || oecol == null) {
+                // We only want to be equal if the other's index is
+                // equal to ours.  This will always be false, I think.
+                return getIndex() - o.getIndex();
+            } else {
```
paulmartel: ALL of the "else"s in this routine follow return statements and just add needless nesting vs. a battery of top-level "return or fall through" tests.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+                // They are both column references.  We want to return
+                // 0 if they refer to the same table expression column.
+                // This is to say, if they have the same column name and
+                // the same table alias or else their column name is in
+                // the using list for their range variables.  It has to
+                // be in the using list for both range variables.  In
```
paulmartel: "Both range variables" seems inaccurate. In general, there may be fewer or more than 2 range variables in question. Maybe you mean "ALL range variables that define the name"?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+                return (ExpressionColumn)m_expression;
+            } else {
+                return null;
+            }
+        }
+
+        public String getName() {
+            if (getTableAlias() == null) {
+                return getAlias();
+            } else if (getTableAlias().length() == 0) {
```
paulmartel: Is a table alias of length 0 really a thing requiring special casing to drop the apparently unprefixed '.' ?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+                    return nc;
+                } else {
+                    RangeVariable rv = ecol.getRangeVariable();
+                    RangeVariable orv = oecol.getRangeVariable();
+                    if (rv == null) {
+                        if (orv == null) {
+                            return getIndex() - o.getIndex();
+                        } else {
+                            // Find out if this column name is in the
+                            // using list of o.  If it is, these both
```
paulmartel: "using list of orv" would be clearer?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: Sorry.  I think I was thinking column a, not article a.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1010,36 +1296,74 @@ Expression replaceAliasInOrderBy(Expression[] columns, int length) {
...
+                    SimpleName selectAliasName = columns[i].alias;
+                    String     selectAlias     = selectAliasName == null ? null : selectAliasName.name;
+                    if (selectAlias == null) {
+                        selectAlias = selectColumnName;
+                    }
+                    // For VoltDB, schema will always be null.
+                    if (schema == null) {
+                        // If this reference has no table name, then
+                        // just compare the aliases.
+                        if (tableName == null) {
```
paulmartel: These two tests on object invariants are being applied rather late in the process. Better a "if schema != null || tableName != null, return this" before the switch statement.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+                        }
+                    } else {
+                        ExpressionColumn myEC = rv.getColumnExpression(getColumnName());
+                        ExpressionColumn oEC = orv.getColumnExpression(getColumnName());
+
+                        if (myEC != null && oEC != null && myEC == oEC) {
+                            // They are the same in a using list.
+                            return 0;
+                        } else {
+                            return getTableAlias().compareTo(o.getTableAlias());
```
paulmartel: Couldn't getTableAlias() return null for an NPE here?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+                        if (rv.getColumnExpression(getAlias()) != null) {
+                            return 0;
+                        } else {
+                            return getIndex() - o.getIndex();
+                        }
+                    } else {
+                        ExpressionColumn myEC = rv.getColumnExpression(getColumnName());
+                        ExpressionColumn oEC = orv.getColumnExpression(getColumnName());
+
+                        if (myEC != null && oEC != null && myEC == oEC) {
```
paulmartel: Is there a actually a valid case where this condition holds yet "getTableAlias().compareTo(o.getTableAlias()) != 0"?
Simply testing the table aliases would be simpler if we could get away with it.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+                    RangeVariable orv = oecol.getRangeVariable();
+                    if (rv == null) {
+                        if (orv == null) {
+                            return getIndex() - o.getIndex();
+                        } else {
+                            // Find out if this column name is in the
+                            // using list of o.  If it is, these both
+                            // denote the same column.  We want to use the
+                            // alias here because we always have an alias
+                            // and we may not have a column name.
```
paulmartel: In what case would we not have a column name and yet the alias would be apropos for lookup in the rangevariable. Column name seems more correct if we have one. If we don't have one, only an alias, I don't see how that alias could have any reasonable connection to the names from the rv.
Is there really an edge case where we have an rv but no column name?
In this edge case, is it not just a coincidence (mishandled here) that the alias would match a name in a using list, so it should NOT compare equal?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1000,9 +1008,287 @@ void collectRangeVariables(RangeVariable[] rangeVariables, Set set) {
...
+         *     table aliases and column names, and we look up the column
+         *     name in the using list.  These will be in the range variable.
+         */
+        @Override
+        public int compareTo(SelectListAliasResolution o) {
+            ExpressionColumn ecol = getExpressionColumn();
+            ExpressionColumn oecol = o.getExpressionColumn();
+            if (ecol == null || oecol == null) {
+                // We only want to be equal if the other's index is
+                // equal to ours.  This will always be false, I think.
```
paulmartel: "false, I think." ? How about "definitely non-zero".
 If we're concerned about an entry being compared to itself, put the this == o test at the top, return 0, and be done with it.  
Similarly, the "Mostly these will not be equal." comment above seems needlessly hedgy.
The indexes for the entries in a tree are being initialized with a counter, so lets be bold and claim that two different SelectListAliasResolution instances will have different indexes and that the indexes will provide a fallback collation for instances whose differences are too complex to base a strict collation upon.
I think that this means that you could have two identical complex select list items with an intervening complex select list item that differed in its detail from the other two and that the result could be an invalid collation that could pretty much invalidate your tree set.

BillWhite: I don't understand the phrase "invalidate your tree set".  I only care about the order when printing error messages.  I care about equality when deciding if a match between a column reference in an order by expression and a select list element is the same as a match between the same column reference and another select list element.  If they are different, we just add both to the set.  If they are the same we only add one.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
BillWhite: This is right.  The SQL 2003 standard references for this are 

- 7.12.Syntax Rules.17 says that in "SELECT T.ID ..." the implicit column name of the derived column "T.ID" is "ID".
- 6.6.Syntax Rules.8.a.i says given an occurrence of "ID" in an order by clause which is in a <cursor specification> which has a <derived column> whose explicit or implicit column name is "ID", the "ID" reference refers to the implicit or explicit column in the select list.   Since the order by expression is part of the <cursor specification> and not a <query expression> it's different from the <group by> clauses.

So, we are agreed.  This seems quite counterintuitive to me.  It seems very likely to cause errors to have non-aliased derived columns magically acquiring aliases.  But that seems to be what the standard wants.

tests/frontend/org/voltdb/regressionsuites/TestGroupByComplexSuite.java
BillWhite: I don't know.  Xin added it in commit 6ba31e5, and I didn't change it.  It used two derived columns, both with aliases "tag".  So the order by was ambiguous in the new *regime*.  I'll change it to "with alias" and "without any alias".  I think that's what's going on here anyway.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: But putting the else shows that they are really logically mutually exclusive cases.  In a
long sequence of equally-indented code with multiple exit points, it's easy to miss the
returns, causing one to not see the structure.

I think it's much clearer this way.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: Both this and o are SelectListAliasResolutions for ExpressionColumns here.  An ExpressionColumn has a range variable.  So, there are two range variables in question
here, and "both" seems accurate.  There may be more range variables in the entire
select statement, but we are looking at only two.

They could both be equal, I suppose, so that we were looking at only one.  But that's
a quibble, I think.  In general there will be two, though there may be only one.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: Well, it looks odd to say 'Name "A" is ambiguous.  It appears in: .A(0) and .A(1).".
It's easy to eliminate the oddity, so I did.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: Ok.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: That's the way it was before.  But I thought it was clearer to break it up into the HSQL artifact and the test which we really care about.  The comments are more clear at least, since they are attached to the test they explain.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: I don't think so.  In the caller, if the select table alias is null we use the select table name for the alias, just as we do with columns.  Since this is an expression column, which is to say, a column reference, it will always have a table name at least.

But I'll put an assertion here just in case.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: No, these are doing different things.  rv.getColumnExpression(name) returns the column expression of a range variable if the name is in the using list of the range variable.  If the name
is the column name of the range variable and the range variable is a column reference, this
function returns null.  It's really misnamed.  So, testing if these are equal tests that they are the same using variable.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: Ok.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: I thought final was an indication to the reader that this function would never be overridden, and that the code they were reading was the actual implementation.  I never thought of it as a note to the compiler.  But I
eliminated the accessor functions anyway.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
BillWhite: You're right, this should be m_columnName.   In fact, we should never be doing anything with aliases.  We only want to return 0 if this and o refer to the same column in the table expression's result set.  The alias is irrelevant.

PR #3260, opened 2016-02-10
This is for Bill W's review of Xin's IW project only.  Don't review it, and don't merge it.

BillWhite:
expression

temporary check in for the RankScanPlanNode work and shift working from
home

Add rank scan plan node and executor

various fix to test equal look up of rank

more fixes for the rank range look up

various fixes here and there and added more tests covering unique index
rank scan

fix the include path for rankscanexecutor

Adding various tests

Add the rank scan replacement guards for non-unique index

turn on all the windowing test cases

temporary checkin for partial index where clause check

various fix to have partial index matching logic for where clause, add
tests also

add percent_rank parser support and hack to implement it in where clause

### DIFF
--- a/build.py
+++ b/build.py
@@ -248,6 +248,7 @@ CTX.INPUT['executors'] = """
  nestloopindexexecutor.cpp
  orderbyexecutor.cpp
  projectionexecutor.cpp
+ rankscanexecutor.cpp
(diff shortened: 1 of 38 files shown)


BillWhite:
Wrong source.

PR #3261, opened 2016-02-10
Review of Xin's IW work with the rank windowing function.  This is only for Bill W's review.  Don't review this unless you are Bill W (or Bill Rogers).  Don't merge this.

BillWhite:
expression

temporary check in for the RankScanPlanNode work and shift working from
home

Add rank scan plan node and executor

various fix to test equal look up of rank

more fixes for the rank range look up

various fixes here and there and added more tests covering unique index
rank scan

fix the include path for rankscanexecutor

Adding various tests

Add the rank scan replacement guards for non-unique index

turn on all the windowing test cases

temporary checkin for partial index where clause check

various fix to have partial index matching logic for where clause, add
tests also

add percent_rank parser support and hack to implement it in where clause

### DIFF
--- a/src/ee/common/NValue.hpp
+++ b/src/ee/common/NValue.hpp
@@ -254,6 +254,9 @@ class NValue {
     /* Create an NValue with the null representation for valueType */
     static NValue getNullValue(ValueType);
 
+    /* Create a max NValue with the input valueType */
+    static NValue getMaxValue(ValueType);
+
     /* Create an NValue promoted/demoted to type */
     NValue castAs(ValueType type) const;
 
@@ -3317,6 +3320,21 @@ inline NValue NValue::getNullValue(ValueType type) {
     NValue retval(type);
     retval.setNull();
     return retval;
+}
+
+inline NValue NValue::getMaxValue(ValueType type) {
+    switch(type) {
+    case VALUE_TYPE_TINYINT:
+        return NValue::getTinyIntValue(INT8_MAX);
+    case VALUE_TYPE_SMALLINT:
+        return NValue::getSmallIntValue(INT16_MAX);
+    case VALUE_TYPE_INTEGER:
+        return NValue::getIntegerValue(INT32_MAX);
+    case VALUE_TYPE_BIGINT:
+        return NValue::getBigIntValue(INT64_MAX);
+    default:
+        return NValue::getNullValue(type);
+    }
 }
 
 inline void NValue::hashCombine(std::size_t &seed) const {
--- a/src/ee/executors/rankscanexecutor.cpp
+++ b/src/ee/executors/rankscanexecutor.cpp
@@ -1,0 +1,233 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2015 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "rankscanexecutor.h"
+
+#include "common/debuglog.h"
+#include "common/common.h"
+#include "common/tabletuple.h"
+#include "common/ValueFactory.hpp"
+#include "expressions/abstractexpression.h"
+#include "expressions/expressionutil.h"
+#include "expressions/rankexpression.h"
+#include "execution/ProgressMonitorProxy.h"
+#include "executors/aggregateexecutor.h"
+#include "indexes/tableindex.h"
+#include "storage/table.h"
+#include "storage/tableiterator.h"
+#include "storage/temptable.h"
+#include "storage/persistenttable.h"
+#include "plannodes/aggregatenode.h"
+#include "plannodes/rankscannode.h"
+#include "plannodes/projectionnode.h"
+#include "plannodes/limitnode.h"
+
+using namespace voltdb;
+
+bool RankScanExecutor::p_init(AbstractPlanNode *abstractNode,
+                                TempTableLimits* limits)
+{
+    VOLT_DEBUG("init IndexCount Executor");
+
+    m_node = dynamic_cast<RankScanPlanNode*>(abstractNode);
+    assert(m_node);
+    assert(m_node->getTargetTable());
+
+    // Create output table based on output schema from the plan
+    setTempOutputTable(limits);
+
+    // Miscellanous Information
+    m_lookupType = m_node->getLookupType();
+    m_endType = m_node->getEndType();
+
+    m_predicate = m_node->getPredicate();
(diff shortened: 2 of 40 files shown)


src/ee/common/NValue.hpp
```diff
@@ -3319,6 +3322,21 @@ inline NValue NValue::getNullValue(ValueType type) {
     return retval;
 }
 
+inline NValue NValue::getMaxValue(ValueType type) {
```
BillWhite: This is used when creating indices.  See TableIndex::getIndexedTableTuple in tableindex.h.

src/ee/executors/rankscanexecutor.cpp
```diff
@@ -0,0 +1,233 @@
...
+}
+
+bool RankScanExecutor::p_execute(const NValueArray &params)
+{
+    // update local target table with its most recent reference
+    RankExpression * rankExpr = m_node->getRankExpression();
+    TableIndex * tableIndex = rankExpr->refreshGetTableIndex();
+
+    if (rankExpr->getPartitonbySize() > 0) {
+        throwDynamicSQLException("rank partition by clause is not supported");
```
BillWhite: getPartitonBySize is probably misspelt.  I suspect "getPartitionBySize" is more like it.

BillWhite:
Most of these changes seem straightforward.  They are mostly just adding things which
are required by the addition of a new function or new syntax.

The only really interesting files are these

1. rankscanexecutor.{h,cpp}
1. rankexpression.{h, cpp}
1. rankpercentageexpression.[h, cpp}
1. rankscannode.{h, cpp}
1. RankExpression.java
1. RankPercentageExpression.java
1. org/voltdb/planner/microoptimizations/ReplaceWithRankPercentage.java
1. org/voltdb/plannodes/RankScanPlanNode.java
1. org/hsqldb_voltpatches/ExpressionRank.java

src/ee/expressions/rankexpression.cpp
```diff
@@ -0,0 +1,141 @@
...
+        delete []m_parititonbySearchKeyBackingStore;
+        delete []m_parititonbyMaxSearchKeyBackingStore;
+        delete []m_orderbySearchKeyBackingStore;
+    }
+
+    TableIndex * RankExpression::refreshGetTableIndex() {
+        VoltDBEngine* engine = ExecutorContext::getEngine();
+        m_tcd = engine->getTableDelegate(m_target_table_name);
+        Table* targetTable = m_tcd->getTable();
+        m_tableIndex = targetTable->index(m_target_index_name);
```
BillWhite: This is the line which will fail if there is no index by the given name.

src/ee/expressions/rankexpression.cpp
```diff
@@ -0,0 +1,141 @@
...
+            if (m_partitionbySize == 0) {
+                rkStart = 1;
+            } else {
+                assert(m_tableIndex->hasKey(&partitionbySearchKey) == false);
+                rkStart = m_tableIndex->getCounterGET(&partitionbySearchKey, false, partitionbyCursor);
+            }
+        }
+//        std::cout << "rkEnd: " << rkEnd << ", rkStart: " << rkStart << std::endl;
+        rkRes = rkEnd - rkStart + 1;
+
```
BillWhite: It might be difficult to implement DENSE_RANK with this algorithm.  Perhaps I'm missing something obvious.

Maybe if we had the kind of index which had a tree not of the actual values of the rows, but of the unique
values of the rows.  Maybe if we had an index on the keyset of an index.  That would work.

It's still unclear what to do if there is no index to be had.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -635,4 +651,54 @@ public static AbstractExpression evaluateExpression(AbstractExpression expr) {
...
+        }
+
+        ae.setLeft(replaceCVEasRankPercentageExpression(ae.getLeft()));
+        ae.setRight(replaceCVEasRankPercentageExpression(ae.getRight()));
+
+        if (ae instanceof ComparisonExpression == false) {
+            return ae;
+        }
+
+        ComparisonExpression ce = (ComparisonExpression) ae;
```
BillWhite: It looks to me as if this code is looking for expressions of the form "percent_rank() over (...) < N", where
N is a numeric constant, and complaining if N is not a valid probability.  This seems very odd to me.
Can't one write "percent_rank() over (...) < 2.0"?  This would always be true, but so would "percent_rank() over (...) <= 1.0", and we allow that just fine.

src/frontend/org/voltdb/plannodes/RankScanPlanNode.java
```diff
@@ -0,0 +1,329 @@
...
+                } else if (ce.getRight() instanceof RankExpression) {
+                    m_rankExpression = (RankExpression) ce.getRight();
+                    m_rankStartExpression = ce.getLeft();
+                } else {
+                    continue;
+                }
+
+                // temporary feature guards
+                // The partition by is not supported right now
+                assert(m_rankExpression != null);
```
BillWhite: This comment may not be correct.  It may be that partition by is at least partially supported.

src/frontend/org/voltdb/plannodes/RankScanPlanNode.java
```diff
@@ -0,0 +1,329 @@
...
+            expressionList.remove(rankIdx);
+        }
+
+        AbstractExpression newPredicate = ExpressionUtil.combine(expressionList);
+        this.setPredicate(newPredicate);
+
+        return true;
+    }
+
+    private int findRankExpressionComparison(List<AbstractExpression> expressList, boolean forStart) {
```
BillWhite: I don't understand why this function is here.  Why do we need to know if a rank scan is associated with a comparison?

src/frontend/org/voltdb/expressions/RankExpression.java
```diff
@@ -0,0 +1,406 @@
...
+
+    public Index getIndex() {
+        return m_index;
+    }
+
+    public boolean isUsingPartialIndex() {
+        return ! m_index.getPredicatejson().isEmpty();
+    }
+
+    public boolean hasPartitionTableIssue(Database db, Table targetTable, List<AbstractExpression> partitionbyExprs) {
```
BillWhite: I think this function wants to calculate if the table's partition column is in the rank expression's partition expressions.  If it is, then all values in a particular rank table partition will be in the same table partition, and can be calculated in an SP query.

src/frontend/org/voltdb/expressions/RankExpression.java
```diff
@@ -0,0 +1,406 @@
...
+                }
+            }
+            // find index candidate
+            Integer left = indexedColRefs.size() - exprs.size();
+            assert(left >= 0);
+            m_indexColumnsCovered.add(left);
+            m_indexCandidates.add(index);
+        } else {
+            // TODO(xin): add support of expression index for rank
+        }
```
BillWhite: TODO here.

BillWhite:
As I said in the sprint retrospective earlier today, there are several problems here.  Xin has also said similar things, so it's no surprise.

1. It's really easy to get into a situation where a stored procedure can't be planned.  For example, create a stored procedure with a rank() expression, then drop the index.  The stored procedure has to be replanned, like all stored procedures which depend on indices, but it can now not be planned.  When sqlcmd executes this, at least, this is not transactional.  The index is dropped, the stored procedure is deleted.
1. It seems very hard to implement DENSE_RANK and the other cumulative distribution functions which depend on values more than rows.  The counting index on rows does not generalize to index keys.
1. It seems hard implement the partition by clause.  getting the index coverage and table partitioning right may be fiddly.  For example, a query such as "select rank() over (partition by a group by b) from R1" plans, but "select rank() over (partition by b group by a)" works just fine in one example I created.  The method to check whether the counting index covers the partition and group by columns depends on the order of the index columns and the actual expression.   This is a bug, of course.
1. There's also some odd code looking for rank functions being compared to constant values which seems unnecessary.

PR #3262, opened 2016-02-10
ENG-9697, generate DR test buffer for given partition.

luyangco:
Change-Id: I4183a3f917dcc9665417fc755fd39c974c0336b3

### DIFF
--- a/src/ee/storage/CompatibleDRTupleStream.cpp
+++ b/src/ee/storage/CompatibleDRTupleStream.cpp
@@ -456,7 +456,7 @@ bool CompatibleDRTupleStream::checkOpenTransaction(StreamBlock* sb, size_t minLe
     return false;
 }
 
-int32_t CompatibleDRTupleStream::getTestDRBuffer(int32_t primaryKeyNValue, int32_t partitionId, char *outBytes) {
+int32_t CompatibleDRTupleStream::getTestDRBuffer(int32_t partitionKeyNValue, int32_t partitionId, char *outBytes) {
     CompatibleDRTupleStream stream;
     stream.configure(partitionId);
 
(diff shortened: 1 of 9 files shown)


src/ee/storage/CompatibleDRTupleStream.cpp
```diff
@@ -455,9 +456,9 @@ bool CompatibleDRTupleStream::checkOpenTransaction(StreamBlock* sb, size_t minLe
     return false;
 }
 
-int32_t CompatibleDRTupleStream::getTestDRBuffer(char *outBytes) {
+int32_t CompatibleDRTupleStream::getTestDRBuffer(int32_t primaryKeyNValue, int32_t partitionId, char *outBytes) {
```
nshi: Do you mean `partitionKeyNValue` instead of `primaryKeyNValue`?

luyangco: Yes, `partitionKeyNValue` will be a more appropriate name.

src/frontend/org/voltdb/LegacyHashinator.java
```diff
@@ -146,4 +146,9 @@ public int pHashToPartition(VoltType type, Object obj) {
         }
         return set;
     }
+
+    @Override
+    public int getPartitionFromHashedToken(int hashedToken) {
+        return java.lang.Math.abs(hashedToken % catalogPartitionCount);
```
nshi: You can throw unsupported operation exception here, because there is no hashed token for the legacy hashinator. It should never be called.

luyangco: I can't do this, I used legacy hashinator in one of the unit tests to create different partitioning for the txnHash.

PR #3263, opened 2016-02-10
ENG-9830: Fix DR buffer rolling.

nshi:
If a transaction does not fit in the current DR buffer, we create a new
one and move the whole transaction over to that buffer. If the
transaction is larger than the largest single buffer we can allocate, it
fails. At least, this is how it's supposed to work.

However, instead of looking at the space needed, it was looking at the
size of the current buffer to determine how big the next buffer should
be. If the current buffer is already a large buffer, it will refuse to
roll to a new buffer even though the currently open transaction can fit
the new buffer perfectly. This caused DR to fail prematurely.


luyangco:
Looks good to me.

PR #3267, opened 2016-02-11
Eng 9506 vmc geb g

johnTcrawford:
Added GEOGRAPHY_POINT and GEOGRAPHY columns to VMC tests; and included tests of Stored Procedures passing null parameters (see ENG-9809, VMC-267). Also, added a new mechanism for specifying variables used repeatedly within the (SQL Query) tests, and used such variables to greatly simplify many of the tests, making them much more readable.

### DIFF
--- a/tests/geb/vmc/src/tests/SqlQueriesTest.groovy
+++ b/tests/geb/vmc/src/tests/SqlQueriesTest.groovy
@@ -27,6 +27,7 @@ import geb.*
 import groovy.json.*
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
 import spock.lang.*
 import vmcTest.pages.*
 
@@ -86,6 +87,7 @@ class SqlQueriesTest extends SqlQueriesTestBase {
     static Boolean runningVoter = null;
     static Boolean runningGenqa = null;
     static Boolean runningVmcTestSever = null;
+    static Map<String,Object> sqlQueryVariables = [:]
 
     @Shared String tablesFileName = TABLES_FILE
     @Shared String viewsFileName  = VIEWS_FILE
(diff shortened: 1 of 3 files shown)


wolffcm:
Looks good to me.

tests/geb/vmc/src/tests/SqlQueriesTest.groovy
```diff
@@ -654,6 +665,71 @@ class SqlQueriesTest extends SqlQueriesTestBase {
...
+        if (!(text instanceof String)) {
+            return text
+        }
+        String result = text
+        for (Matcher variables = result =~ /(__\w+)/; variables.find(); variables = result =~ /(__\w+)/ ) {
+            String variable = variables.group(1)
+            // Special case, where the entire text consists of one variable
+            if (variable.equals(text)) {
+                return sqlQueryVariables.get(variable)
+            }
```
BillWhite: There's something odd about this case.  If the text is entirely a variable, say text = "__var", and __var has
variables in it, say __var = "a __planet." and __planet is "pluto".  Then don't we just return "a __planet."  But
if text = "There is __var".  We would return "There is a pluto."  Isn't this an asymetry between the special _entirely a variable_ case an other cases.

And if __var is "a __var", and text is "There is a __var." it seems likely  to loop.  Maybe there should be a limit on the number of times this loop is executed.

BillWhite:
That's all I have.  Most of it I don't understand, though it seems plausible.

tests/geb/vmc/src/tests/SqlQueriesTest.groovy
johnTcrawford: OK, I've now fixed the infinite loop problem, by adding a maxCount (set to 100), after which it gives up, and prints a WARNING. As for the "special case", I have not really changed it, but I've made it clearer (both with an additional input arg and better comments) when it applies, which is for a "result" rather than a "query" - and a variable in a result may represent a Map, rather than a String - so the idea is that you can return that entire Map this way.

BillWhite:
Seems simple and elegant to me.

PR #3268, opened 2016-02-11
Vdm 20

deerwalk:


lib/python/vdm/server/HTTPListener.py
```diff
@@ -959,6 +1009,162 @@ def get_deployment_from_xml(deployment_xml, is_list):
     return deployments
 
 
+def get_deployment_for_upload(deployment_xml, is_list):
+    new_deployment = {}
+    deployments = []
+    for field in deployment_xml:
```
manjujames: Please don't duplicate this code. This big 'if' with deployment fields is already in 2 places. You should break this out into a separate method, call it from other places and thus have it only in one place.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1020,11 +1226,17 @@ def get_deployment_properties(export_xml, is_list):
...
-                new_export[field] = export[field]
+                if field == 'plaintext':
+                    new_export[field] = parse_bool_string(export[field])
+                else:
+                    new_export[field] = export[field]
             exports.append(new_export)
     else:
         for field in export_xml:
-            new_export[field] = export_xml[field]
+            if field == 'plaintext':
```
manjujames: Anytime there is a need to copy-paste code, think about whether it would be better to break it out into a separate method.  In this case, this whole for block can be in a separate method and thus avoid code duplication.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1207,32 +1448,40 @@ def delete(server_id):
...
+                abort(404)
+            # remove the server from given database member list
+            current_database = [database for database in Global.DATABASES if database['id'] == database_id]
+            current_database[0]['members'].remove(server_id)
+            # Check if server is referenced by database
+            for database in Global.DATABASES:
+                if database["id"] == database_id:
+                    continue
+                if server_id in database["members"]:
+                    return jsonify({'success': "Server deleted from given member list only. "
```
manjujames: A server_id must only be in one database. So, this code can be removed and the restriction enforced when adding a server to a database.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1242,44 +1491,52 @@ def put(server_id):
...
-            return jsonify(success=False, errors=inputs.errors)
-        current_server = [server for server in Global.SERVERS if server['id'] == server_id]
-        if len(current_server) == 0:
-            abort(404)
+        database = [database for database in Global.DATABASES if database['id'] == database_id]
+        if len(database) == 0:
+            return make_response(jsonify( { 'statusstring': 'No database found for id: %u' % database_id } ), 404)
+        else:
+            members = database[0]['members']
+        if server_id in members:
```
manjujames: This code to look up a server_id in a database is there in get, delete and put. May be that can be separated out into its own method and reused from all three places.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1653,17 +1959,14 @@ def put(database_id):
...
-            # because voltadmin shutdown gives 'Connection broken' output
-            return make_response(jsonify({'statusstring': 'Shutdown request sent successfully'}), 200)
-        except Exception, err:
-            print traceback.format_exc()
-            return make_response(jsonify({'statusstring': str(err)}),
-                                 500)
+        # try:
+        #     test = Get_Voltdb_Process().processId
+        #     # os.kill(Get_Voltdb_Process().processId, signal.SIGTERM)
+        #     return make_response(jsonify({'statusstring': 'success'}), 200)
```
manjujames: Why was this changed to a kill? We should use voltadmin shutdown (what was there before) to shutdown all the servers in database.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1748,8 +2051,8 @@ def put(database_id, server_id):
 
         try:
             server = voltdbserver.VoltDatabase(database_id)
-            response = server.stop_server(server_id)
-            return make_response(jsonify({'statusstring': response}), 200)
+            response = server.kill_server(server_id)
```
manjujames: Same here. We should not randomly kill server. voltadmin stop should be used here, which will check if the server can actually be shutdown without breaking the # of hosts requirement and such.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1825,11 +2130,90 @@ class DatabaseDeploymentAPI(MethodView):
...
+                                return jsonify({'error': deployment_data['error']})
+                        else:
+                            deployment_json = deployment_data[0]
+                        req = DictClass()
+                        req.json = {}
+                        req.json = deployment_json
+                        inputs = JsonInputs(req)
+                        if not inputs.validate():
+                            return jsonify(success=False, errors=inputs.errors)
+                        if 'systemsettings' in req.json and 'resourcemonitor' in req.json['systemsettings']:
```
manjujames: Are these validations things you can do while mapping to xml?

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1842,6 +2226,82 @@ def get():
...
+
+class StatusDatabaseServerAPI(MethodView):
+    """Class to return status of servers"""
+
+    @staticmethod
+    def get(database_id, server_id):
+
+        server = [server for server in Global.SERVERS if server['id'] == server_id]
+        try:
+            client = voltdbclient.FastSerializer(str(server[0]['hostname']), 21212)
```
manjujames: You should use port configured in server instead of 21212

lib/python/vdm/server/Log.py
```diff
@@ -0,0 +1,39 @@
...
+        return myLine
+    except:
+        return "error"
+    pass
+
+
+def get_most_recent_log():
+    logdir = HTTPListener.Global.PATH
+
+    logfiles = sorted([ f for f in os.listdir(logdir) if f.startswith('voltserver.output')])
```
manjujames: This is only the console output collected at startup, not the server log. Is this giving you the error details you need?

lib/python/vdm/server/voltdbserver.py
```diff
@@ -58,18 +61,36 @@ def ignore_signals():
...
+
 def get_voltdb_dir():
     """
     Utility method to get voltdb bin directory
     """
     return os.path.realpath(os.path.join(HTTPListener.Global.MODULE_PATH, '../../../..', 'bin'))
 
+
+def check_snapshot_folder():
+    outfilename = os.path.join(HTTPListener.Global.PATH, "voltdbroot/snapshots")
```
manjujames: snapshots path shouldn't be hardcoded. This should be picked up from deployment.

PR #3270, opened 2016-02-11
ENG-8932 Code review comment implementation 

alexeevm:


src/ee/executors/indexscanexecutor.cpp
```diff
@@ -410,43 +412,16 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
...
-                        break;
-                    }
-                } else {
-                    //
-                    // Straight Insert
-                    //
-                    m_outputTable->insertTupleNonVirtual(tuple);
-                }
+            else {
+                outputTuple(postfilter, temp_tuple);
```
paulmartel: This else clause (for m_projector.numSteps() == 0) used to process "tuple" in place of the projected temp_tuple of the else clause -- is temp_tuple even initialized here?

paulmartel:
I just had the one comment. The rest looks really good.
I'm a bit surprised/disappointed that junits were not badly broken by that one issue.
I don't think that I've ever understood it 100%. I know that this kind of temp_tuple vs. tuple confusion used to cause a serious column indexing bug, but that was before the recent changes by Xin and Chris.

src/ee/executors/indexscanexecutor.cpp
alexeevm: Yes, this must be 
outputTuple(postfilter, tuple);

corrected. Cut and paste error. The temp_tuple is initialized there in a sense that is has a valid schema and not-null data storage  but that about it.
I can see two reasons why it wasn't caught by the junit suite -
 An inline projection is there - select * from T.PKEY where PKEY > 0 - IndexScan has it
 Some test cases only assert the number of output rows without comparing the actual results. In this case, the EE won't crash because the temp tuple is 'initialized' but would produce invalid results

I will try to create a test case with IndexScan without the inline projection

alexeevm: Paul, I am having a hard time finding an example that would exercise the ELSE part of the IF statement. For that to happen, the IndexScan must not have an inline projection and/or an aggregate nodes. If there is no aggregation, the projection node is inlined by the PlanAsembler.  If there is an aggregation, it gets inlined during the micro optimization unless there is an inline limit node (InlineAggregation - line 87). To have an inline limit node, it must sit directly on top of the IndexScan or a projection node followed by the scan (PushDownLimits). But the aggregation node is below the limit node thus preventing the latter to be inline, if I read the code right.
Could you think of any SQL where an IndexScan won't have inline projection and aggregation nodes?

PR #3271, opened 2016-02-12
ENG-9830: Make max PBD segment size public.

nshi:


src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -32,7 +32,8 @@
     static final int COUNT_OFFSET = 0;
     static final int SIZE_OFFSET = 4;
 
-    static final int CHUNK_SIZE = (1024 * 1024) * 64;
+    // Has to be able to hold at least one object (compressed or not)
+    public static final int CHUNK_SIZE = (1024 * 1024) * 64;
```
akhanzode: If making public why not introduce a hidden property?

nshi: I'm not sure what you mean. Making it public just gives tests and other part of the system read access to this constant. Making it a hidden Java property makes it a user-settable nob, it doesn't solve the visibility problem.

PR #3272, opened 2016-02-12
Eng 9626 runtime exception

paulmartel:


src/frontend/org/voltdb/plannodes/AbstractScanPlanNode.java
```diff
@@ -176,29 +176,22 @@ public AbstractExpression getPredicate() {
...
-        if (predicate != null)
-        {
-            // PlanNodes all need private deep copies of expressions
-            // so that the resolveColumnIndexes results
-            // don't get bashed by other nodes or subsequent planner runs
-            m_predicate = (AbstractExpression) predicate.clone();
-        } else {
-            m_predicate = null;
-        }
+    public void setPredicate(List<AbstractExpression> exps) {
```
wolffcm: This parameter name doesn't match the updated Javadoc comment...

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -44,6 +44,26 @@ public static void finalizeValueTypes(AbstractExpression exp)
         exp.finalizeValueTypes();
     }
 
+    public static AbstractExpression cloneAndCombine(List<AbstractExpression> exps) {
```
wolffcm: For this method and `combine`, the name seems a little too general.  I would call them `combinePredicates` and `cloneAndCombinePredicates`, since it's producing an AND chain of boolean expressions.  It was like that when you got here, though, I don't feel strongly that you should change it.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -575,18 +577,33 @@ private AbstractPlanNode getSelectSubPlanForJoin(BranchNode joinNode,
...
-                // InnerPlan is an IndexScan. In this case the inner and inner-outer
-                // non-index join expressions (if any) are in the otherExpr. The former should stay as
-                // an IndexScanPlan predicate and the latter stay at the NLJ node as a join predicate
+            // Copy innerAccessPath.joinExprs to leave it unchanged,
+            // avoiding accumulation of redundant expressions when
+            // joinClauses gets built up for various alternative plans.
+            ArrayList<AbstractExpression> joinClauses = new ArrayList<>(innerAccessPath.joinExprs);
+            if ((innerPlan instanceof IndexScanPlanNode) ||
+                (innerPlan instanceof NestLoopIndexPlanNode
+                    && innerPlan.getChild(0) instanceof MaterializedScanPlanNode)) {
```
wolffcm: Ok, so I think that when you have an index optimized IN predicate, that's expressed in the plan as a nest loop index node where the inner scan is really a "materialized scan", meaning it just produces each value in the IN list as a separate row.  Do I have that right?

This is on the edge of my current understanding of this code.  Can you summarize for me how plans produced by this change in particular are different from before?

wolffcm:
Looks good to me.  I am unclear on what the SelectSubplanAssembler change does, if you could clarify, that would be helpful.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
paulmartel: @wolffcm  Yes. You have the model correct. The recent change is two-fold. 

I added an ArrayList copy because we were accidentally using the list of predicates from the parsed statement as our "mutable working copy" in each attempt to generate a plan.

The more complicated part of the change has to do with a difference in where join criteria are applied for NLIJs vs NLJs. The planner initially pushes both outer-inner join criteria AND inner-table-specific criteria to the inner index scan, hoping to maximize the number of filters that can be index-optimized for an NLIJ. If it ends up falling back to an NLJ with a non-inlined inner indexscan, that means that the indexscan will get executed once independently of the outer scan -- so it must search the indexscan node for any post-processing filters that reference other tables and migrate them back up to the NLJ to be executed as post-filters there -- once per joined row in the inner nested loop. 
In contrast, an NLIJ makes one pass at the index scan PER outer row, so it does not matter if the post-filter join criteria is applied by the indexscan before it returns to the NLIJ or by the NLIJ after it returns from the indexscan.
The edge case I fixed is for the IN LIST case when the expected child indexscan of the NLJ got replaced with a child NLIJ of a materialized scan and an index scan. This case needs to be recognized by this code that "pulls back up" the post-filter for an NLJ. In particular, it needs to tunnel through this intervening NLIJ to find its "grandchild" inner index scan node and pull back its outer-inner join criteria post-predicates just as when the index scan is its actual inner child.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
paulmartel: Good point. I think this naming had a younger me confused for a time until I just got used to it. No point in putting another generation of engineers through that.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
paulmartel: @wolffcm Of course, the problem is wording a more concise description of all that in a comment that is clear without being too long for the next generation to bother with reading.

paulmartel: @wolffcm Actually, the short answer to your question is that in an edge case involving index-optimized IN LISTs index scans that were on the inner side of an NLJ could contain post-filters that referenced columns from an outer table. These have no resolution in this context when the index scan is run stand-alone prior to the NLJ executor. The new planner code finds these predicates in all such cases and moves them into the NLJ post-filters where they resolve as usual.

wolffcm: Ok, I think I've got it.  Thanks for the explanation.  We weren't pulling up predicates into a resolvable place in the query tree for some IN list cases.  I guess without this fix, we would see errors where there's invalid column indexes in the EE, or (even worse) wrong answers?

PR #3280, opened 2016-02-15
ENG-9299 ddl compiler and planner support geography columns indexed f…

paulmartel:
…or CONTAINS

### DIFF
--- a/src/frontend/org/voltdb/planner/SubPlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/SubPlanAssembler.java
@@ -26,6 +26,7 @@ import java.util.Set;
 
 import org.json_voltpatches.JSONException;
 import org.voltdb.VoltType;
+import org.voltdb.catalog.Column;
 import org.voltdb.catalog.ColumnRef;
 import org.voltdb.catalog.Database;
 import org.voltdb.catalog.Index;
@@ -154,7 +155,7 @@ public abstract class SubPlanAssembler {
                     // For optimization purposes, keep track of the covering (query) expressions that exactly match the
                     // covered index sub-expression. They can be eliminated from the post-filter expressions.
                     List<AbstractExpression> exactMatchCoveringExprs = new ArrayList<AbstractExpression>();
-                    if (isPartialIndexPredicateIsCovered(tableScan, allExprs, path.index, exactMatchCoveringExprs)) {
+                    if (isPartialIndexPredicateCovered(tableScan, allExprs, path.index, exactMatchCoveringExprs)) {
                         filterPostPredicateForPartialIndex(path, exactMatchCoveringExprs);
                     } else {
                         path = null;
@@ -164,7 +165,7 @@ public abstract class SubPlanAssembler {
                 // Partial index can be used solely to eliminate a post-filter
                 // even when the indexed columns are irrelevant
                 List<AbstractExpression> exactMatchCoveringExprs = new ArrayList<AbstractExpression>();
-                if (isPartialIndexPredicateIsCovered(tableScan, allExprs, index, exactMatchCoveringExprs)) {
+                if (isPartialIndexPredicateCovered(tableScan, allExprs, index, exactMatchCoveringExprs)) {
                     path = getRelevantNaivePath(allJoinExprs, filterExprs);
                     filterPostPredicateForPartialIndex(path, exactMatchCoveringExprs);
                     path.index = index;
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -767,6 +787,99 @@ else if (retval.lookupType == IndexLookupType.GT) {
...
+        // These are identified by their unique function type signature
+        // -- safe for now, until we happen to add an
+        // unrelated function with the same signature.
+        // The alternative would be to import the specific function id for CONTAINS
+        // from FunctionForVoltDB to match on that, and then probably expand that to
+        // include APPROX_CONTAINS if/when we decide to explicitly support APPROX_CONTAINS
+        // as a filter that COMPLETELY eliminates post-filtering on an indexed geography
+        // column OR to export an additional optional "is geo indexable" attribute to the
+        // VoltXML that sets up FunctionExpressions. Maybe we can revisit this if/when
+        // we change FunctionForVoltDB to support APPROX_CONTAINS.
```
wolffcm: Seems like we would want to explicitly check for the right function id in FunctionForVoltDB.  If we add a function that has the same signature, it won't be obvious that we need to update this code, and the planner will begin behaving strangely.  Seems better to add the explicit check now?

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -767,6 +787,99 @@ else if (retval.lookupType == IndexLookupType.GT) {
...
+            }
+            if (coveringColId != geoTve.getColumnIndex()) {
+                continue;
+            }
+
+            AbstractExpression searchKeyArg = args.get(1);
+            assert searchKeyArg.getValueType() == VoltType.GEOGRAPHY_POINT;
+            // Search key operand must not be from the same table,
+            // e.g. contains(t.a, t.b) is not indexable.
+            if (isOperandDependentOnTable(searchKeyArg, tableScan)) {
```
wolffcm: This behaves right for self joins, which should be indexable?  Seems like it will, would be good to test though.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -767,6 +787,99 @@ else if (retval.lookupType == IndexLookupType.GT) {
...
+                continue;
+            }
+
+            retval.lookupType = IndexLookupType.GEO_CONTAINS;
+            filtersToCover.remove(searchKeyArg);
+            retval.indexExprs.add(searchKeyArg);
+            retval.otherExprs.addAll(filtersToCover);
+            // It's unlikely but possible that the query has more than one
+            // CONTAINS filter that uses the same geography column, e.g.
+            //  "WHERE CONTAINS(place, point) AND CONTAINS(place, ?)"
```
wolffcm: If one of the filters is a join predicate, and the other is not, seems like we would want to always prefer the join predicate?

src/frontend/org/voltdb/planner/SubPlanAssembler.java
wolffcm: I've noticed that you test this, very nice.

wolffcm:
This is looking pretty good, but it seems like more is needed.

We should have guards so that users don't use more than one component, or try to use the UNIQUE or ASSUMEUNIQUE keywords.  Also, partial indexes: seems like they might work out of the box...

Also, it would be good to have some planner tests that make sure that if there are two indexable filters, we choose the geo index if it applies, unless there is an equality filter on a primary key.

In my EE changes, the EE is looking for the index type COVERING_CELL_INDEX (as opposed to BALANCED_TREE_INDEX or HASH_INDEX), so the geo index should get created with this index type.  I mentioned this in the ticket.  This could also be done as part of the integration step.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
paulmartel: @wolffcm Why the preference? I could argue the counter-case. If two filters are -- for all we know -- equally effective -- why not scan the index for the one with the constant argument so you can do that once and be done? I think I've had this argument about other kinds of joins. I think that "your side" lost in those cases. So now I think you've got consistency with that precedent weighing against you.

paulmartel:
Will add guards, test guards, fix cost model, test cost model, and change index type.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
wolffcm: I see your point.  I'm happy to follow precedent here.  I was getting confused with the other execution engine I worked on years ago.

PR #3282, opened 2016-02-16
ENG-9830: Fix DR buffer limit error handling.

nshi:
Move the DR stream append around in the UPDATE/DELETE paths to make sure
it won't cause the operation half completed without proper rollback. The
UPDATE/DELETE code paths assume that certain part of the operation will
always succeed. If it fails in the middle of those critical sections,
things won't rollback.

### DIFF
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -440,6 +440,23 @@ void PersistentTable::insertTupleCommon(TableTuple &source, TableTuple &target, 
 
     }
 
+    // Write to DR stream before everything else to ensure nothing gets left in
+    // the index if the append fails.
+    ExecutorContext *ec = ExecutorContext::getExecutorContext();
+    if (hasDRTimestampColumn()) {
+        setDRTimestampForTuple(ec, target);
+    }
+
+    DRTupleStream *drStream = getDRTupleStream(ec);
+    size_t drMark = 0;
+    if (drStream && !m_isMaterialized && m_drEnabled && shouldDRStream) {
+        const int64_t lastCommittedSpHandle = ec->lastCommittedSpHandle();
+        const int64_t currentTxnId = ec->currentTxnId();
+        const int64_t currentSpHandle = ec->currentSpHandle();
+        const int64_t currentUniqueId = ec->currentUniqueId();
+        drMark = drStream->appendTuple(lastCommittedSpHandle, m_signature, currentTxnId, currentSpHandle, currentUniqueId, target, DR_RECORD_INSERT);
+    }
+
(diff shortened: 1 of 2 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -440,6 +440,19 @@ void PersistentTable::insertTupleCommon(TableTuple &source, TableTuple &target,
...
+    // the index if the append fails.
+    ExecutorContext *ec = ExecutorContext::getExecutorContext();
+    DRTupleStream *drStream = getDRTupleStream(ec);
+    size_t drMark = 0;
+    if (drStream && !m_isMaterialized && m_drEnabled && shouldDRStream) {
+        const int64_t lastCommittedSpHandle = ec->lastCommittedSpHandle();
+        const int64_t currentTxnId = ec->currentTxnId();
+        const int64_t currentSpHandle = ec->currentSpHandle();
+        const int64_t currentUniqueId = ec->currentUniqueId();
+        drMark = drStream->appendTuple(lastCommittedSpHandle, m_signature, currentTxnId, currentSpHandle, currentUniqueId, target, DR_RECORD_INSERT);
```
luyangco: You need also move `setDRTimestampForTuple()` prior to this block, otherwise the tuple will miss hiddenColumn value.

src/ee/storage/persistenttable.cpp
```diff
@@ -575,6 +578,21 @@ bool PersistentTable::updateTupleWithSpecificIndexes(TableTuple &targetTupleToUp
...
+    DRTupleStream *drStream = getDRTupleStream(ec);
+    size_t drMark = 0;
+    if (drStream && !m_isMaterialized && m_drEnabled) {
+        const int64_t lastCommittedSpHandle = ec->lastCommittedSpHandle();
+        const int64_t currentTxnId = ec->currentTxnId();
+        const int64_t currentSpHandle = ec->currentSpHandle();
+        const int64_t currentUniqueId = ec->currentUniqueId();
+        std::pair<const TableIndex*, uint32_t> uniqueIndex = getSmallestUniqueIndex();
+        drMark = drStream->appendTuple(lastCommittedSpHandle, m_signature, currentTxnId, currentSpHandle, currentUniqueId, targetTupleToUpdate, DR_RECORD_DELETE, uniqueIndex.first, uniqueIndex.second);
+        drStream->appendTuple(lastCommittedSpHandle, m_signature, currentTxnId, currentSpHandle, currentUniqueId, sourceTupleWithNewValues, DR_RECORD_INSERT);
```
luyangco: My above comment also applies to update path

luyangco:
This is the only comment I have, other parts look fine. Review Complete.

PR #3286, opened 2016-02-17
Add regression-test to test end-to-end functionality for Contains() with geospatial index

hhakimi53:
Adding test to regression suite to verify the Contains() operation on table which is indexed on geo columns produces correct answer by verify it against the result of contains on data-set using non-indexed table. 

In this review request create index on geo columns in commented out and can be enabled once the geo indexes are available.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
@@ -49,9 +49,8 @@ public class TestGeospatialIndexes extends RegressionSuite{
     private class PolygonPoints {
         private GeographyValue      m_polygon;
         private GeographyPointValue m_center;
-        private GeographyPointValue m_pointInside;                          // point inside polygon which is not the center
-        private GeographyPointValue m_pointNearCellOutsidePolygon;            // point which is the near cell boundary covering regular polygon and is outside the polygon
-        private GeographyPointValue m_pointOutsidePolygonCell;              // point which is outside the polygon's surrounding cell
+        // stores that are inside and outside polygons
+        private List<GeographyPointValue> m_pointsForPolygons = new ArrayList<GeographyPointValue>();
         private final double        m_incrementFactor = 0.01;               // used to calculate point inside polygon with or without hole
 
         PolygonPoints(GeographyPointValue center,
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,358 @@
...
+        super(name);
+    }
+
+    private class PolygonPoints {
+        private GeographyValue      m_polygon;
+        private GeographyPointValue m_center;
+        private GeographyPointValue m_pointInside;                          // point inside polygon which is not the center
+        private GeographyPointValue m_polygonCellVertex;                    // point which is the vertex of surrounding cell regular convex polygon but is outside the polygon
+        private GeographyPointValue m_pointOutSidePolygonCell;              // point which is outside the polygon's surrounding cell
+        private Random              m_random = new Random(999);             // used for randomly generating phi for first vertex of the polygon w.r.t center's latitude line
```
wolffcm: Because you are seeding this Random instance is a member variable, and you are seeding it with the same number every time, it seems like you will get the same stream of random numbers each time?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,358 @@
...
+
+    public TestGeospatialIndexes(String name) {
+        super(name);
+    }
+
+    private class PolygonPoints {
+        private GeographyValue      m_polygon;
+        private GeographyPointValue m_center;
+        private GeographyPointValue m_pointInside;                          // point inside polygon which is not the center
+        private GeographyPointValue m_polygonCellVertex;                    // point which is the vertex of surrounding cell regular convex polygon but is outside the polygon
```
wolffcm: Just a comment about the name of this variable: if this point is outside the polygon, then it's not a vertex, it's just a point.  At least that's how I think about points vs. vertexes.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
hhakimi53: Yes, I will update it. Thanks

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
hhakimi53: Sure, will rename the variable.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,411 @@
...
+                + "\n"
+                + "CREATE TABLE INDEXED_PLACES (\n"
+                + "  id         INTEGER NOT NULL,\n"
+                + "  loc        GEOGRAPHY_POINT\n"
+                + ");\n"
+                + "CREATE TABLE INDEXED_BORDERS (\n"
+                + "  id         INTEGER NOT NULL,\n"
+                + "  region     GEOGRAPHY\n"
+                + ");\n"
+                //+ "CREATE INDEX INDEX_LOC_INDEXED_PLACES ON INDEXED_PLACES(loc)\n;"           // Enable this once Geo Indexes are implemented
```
wolffcm: There's no plans at the moment to allow indexing on GEOGRAPHY_POINT columns.  It will strictly be about polygons for now, so it might be a while before you can uncomment this line.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,411 @@
...
+        String prefixMsg;
+        final double EPSILON = 1.0e-12;
+
+        sql = "Select * from PLACES order by id, loc;";
+        resultsUsingGeoIndex = client.callProcedure("@AdHoc", sql).getResults()[0];
+
+        sql = "Select * from INDEXED_PLACES order by id, loc;";
+        resultsFromNonGeoIndex = client.callProcedure("@AdHoc", sql).getResults()[0];
+        prefixMsg = "Assertion failed comparing contents of places and indexed_places to be same: ";
+        assertTablesAreEqual(prefixMsg, resultsUsingGeoIndex, resultsFromNonGeoIndex);
```
wolffcm: One could imagine these tables being different in memcheck mode... should probably add an ORDER BY here.

wolffcm: Oops, you have an order by, never mind.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,411 @@
...
+        sql = "Select * from BORDERS order by id, region;";
+        resultsUsingGeoIndex = client.callProcedure("@AdHoc", sql).getResults()[0];
+
+        sql = "Select * from INDEXED_BORDERS order by id, region;";
+        resultsFromNonGeoIndex = client.callProcedure("@AdHoc", sql).getResults()[0];
+        prefixMsg = "Assertion failed comparing contents of places and indexed_borders to be same: ";
+        assertTablesAreEqual(prefixMsg, resultsUsingGeoIndex, resultsFromNonGeoIndex);
+
+        // test contains with fixed data
+        sql = "Select A.id, A.region, B.loc from INDEXED_BORDERS A, INDEXED_PLACES B "
```
wolffcm: I don't think that this query will use a geo index, since you are joining on a primary key, which is unique.  If the planner is working correctly, then it will not choose the geo index in this case.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
hhakimi53: Oh yes. I can remove the line for indexing GEOGRAPHY_POINT and whenever we support for it, it can be added to appropriate place.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
wolffcm: Oops, you are not joining on a primary key.  I was looking at the wrong table definition.

It would still be good to have some queries where the CONTAINS is the only predicate.

hhakimi53: id is not defined as primary key intentionally. The only index that will be defined will be on Geography column. So for that case, planner should opt for geo index, I think.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,411 @@
...
+            assertTablesAreEqual(prefixMsg, resultsFromNonGeoIndex, resultsUsingGeoIndex);
+        }
+
+        // Test parameterized Contains() - test with point argument to Contains() being parameterized
+        // To verify this, point which is inside polygon is fetched from the cached generated data and is supplied to SP.
+        // Output result of the query should have only one matching polygon that contains the supplied. This polygon is
+        // same as geography value in corresponding entry of polygon-point data
+
+        // Create procedure on Contains functions
+        sql = "Create procedure PointInPolygon as select A.region from INDEXED_BORDERS A "
```
wolffcm: Again, here it would be good to have a variant of this query with just the CONTAINS predicate.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
hhakimi53: Sure, I will add a query which has only Contains predicate.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -0,0 +1,411 @@
...
+        final int longitude = m_random.nextInt(10) - 178;   // generate longitude between -178 to -168
+        final int latitude = m_random.nextInt(10) + 72;     // generate latitude between 82 to 72
+        GeographyPointValue center = new GeographyPointValue(longitude, latitude);
+        // offset to use for generating new center of polygon. this is not randomized, it will generate polygons in same order
+        // almost around regular steps running from north-west to south-east
+        final GeographyPointValue centerShiftOffset = new GeographyPointValue(2.1 * polygonRadius, -1 * polygonRadius);
+
+        // triangles without holes
+        //System.out.println("Generate triangles without hole, center: " + center.toString());
+        center = generatePolygonPointData(center, centerShiftOffset, polygonRadius, 3, 0, numberOfPolygonPoints);
```
wolffcm: So this generates a bunch of polygons all in sort of a diagonal path?  Just trying to visualize the set of polygons being created.

hhakimi53: Yes, this will generate bunch of polygons almost around the diagonal line running through north-west to south-east.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
wolffcm: Would also be good to pick some points at random within the bounding box of the generated polygons, and then see which polygons they are in.  Since the polygons are generated along a diagonal path, and not a grid, there will be fewer hits though.

wolffcm:
That's all my comments.  Would be nice to see the polygons generated in a grid pattern instead of diagonally if it's not too much trouble.

hhakimi53:
Changing logic to generate polygon along horizontal or vertical grid should not be difficult. I had taken approach to fit more polygon along diagonal line was done to accommodate more non-overlapping I will look into it and may be have logic using overlapping polygon. Thanks Chris, for reviewing the changes and the feedback.

hhakimi53:
The test logic for data generated has been updated so that it generates polygons that are overlapping and are placed on 3*3 horizontal grid. The point generation logic has been tweaked so that it generates around 12-18 points such that:
- there are n points that are inside polygon's (this number is equal to number of vertex the polygon has)
- 8 random points (4 in valgrind env) that are inside polygon's outer shell and along the line of axis of polygon's vertex
- 4 points that are outside polygon and near cell covering polygon
- another one that is outside the bounding box of polygon.
Query predicates also has been updated so that it uses only Contains and Not Contains predicate.

wolffcm:
Looks good to me.  Go ahead and merge this when Jenkins is blue.

hhakimi53:
Thanks Chris for reviewing the changes. Jenkins is blue for the newly added tests. There were some tests that failed and they are not related to the test changes here. I have restarted Jenkins job and will merge the changes once Jenkins is done.

PR #3287, opened 2016-02-17
Vdm 20

deerwalk:


lib/python/vdm/runner.py
```diff
@@ -0,0 +1,124 @@
...
+# heavy lifting happens in runner.main(). The script name determines the verbs
+# that are loaded from <name>.d subdirectories. It loads the version number
+# from version.txt in the script's parent directory. It can be copied to other
+# names, and also to other locations if the path-building is adjusted. The
+# description should also be changed if re-used for another base command.
+import sys
+import os
+from optparse import OptionParser
+from os.path import expanduser
+
```
manjujames: I am ignoring VDM-33 related changes for now because that is not part of VDM-20 task. However, I will say rename this file to vdmrunner.py or something similar because we have runner.py which is a common script that we use to run many of our commands.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -567,13 +577,13 @@ def make_configuration_file():
...
         i += 1
-    return tostring(main_header,encoding='UTF-8')
+    return tostring(main_header, encoding='UTF-8')
 
 
 def sync_configuration():
     headers = {'content-type': 'application/json'}
-    url = 'http://'+__IP__+':'+str(__PORT__)+'/api/1.0/vdm/configuration/'
-    response = requests.post(url,headers = headers)
+    url = 'http://' + __IP__ + ':' + str(__PORT__) + '/api/1.0/vdm/configuration/'
```
manjujames: Might be easier to read if you used ('http://%s:%u/api/1.0/vdm/configuration') % \
                                                                     (__IP__, __PORT__)

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1653,17 +1958,42 @@ def put(database_id):
         Returns:
             Status string indicating if the stop request was sent successfully
         """
+        is_force = request.args.get('force')
 
-        try:
+        if is_force == "false":
```
manjujames: Can you compare case-insensitive?

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1207,32 +1468,40 @@ def delete(server_id):
...
+            database_id = request.json['dbId']
+            # delete a single server
+            server = [server for server in Global.SERVERS if server['id'] == server_id]
+            if len(server) == 0:
+                abort(404)
+            # remove the server from given database member list
+            current_database = [database for database in Global.DATABASES if database['id'] == database_id]
+            current_database[0]['members'].remove(server_id)
+            # Check if server is referenced by database
+            for database in Global.DATABASES:
```
manjujames: A server can be in only one database. So, this code is not necessary.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1653,17 +1958,42 @@ def put(database_id):
         Returns:
             Status string indicating if the stop request was sent successfully
         """
+        is_force = request.args.get('force')
 
-        try:
+        if is_force == "false":
+            try:
+                server = voltdbserver.VoltDatabase(database_id)
```
manjujames: For shutting down a database, we can probably try a 'voltadmin shutdown' always first. If there is a connection failure and force=true, then send stop server request to the individual servers with force=true

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1653,17 +1958,42 @@ def put(database_id):
...
-            # because voltadmin shutdown gives 'Connection broken' output
-            return make_response(jsonify({'statusstring': 'Shutdown request sent successfully'}), 200)
-        except Exception, err:
-            print traceback.format_exc()
-            return make_response(jsonify({'statusstring': str(err)}),
-                                 500)
+            return server.kill_database(database_id)
+
+
+    # @staticmethod
```
manjujames: We should cleanup this commented out code.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1748,8 +2078,8 @@ def put(database_id, server_id):
 
         try:
             server = voltdbserver.VoltDatabase(database_id)
-            response = server.stop_server(server_id)
-            return make_response(jsonify({'statusstring': response}), 200)
+            response = server.kill_server(server_id)
```
manjujames: Why is this not using 'force' flag?

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1811,10 +2143,10 @@ def post():
         for member in result['vdm']['members']:
             try:
                 headers = {'content-type': 'application/json'}
-                url = 'http://'+member['hostname']+':'+str(__PORT__)+'/api/1.0/vdm/sync_configuration/'
+                url = 'http://' + member['hostname'] + ':' + str(__PORT__) + '/api/1.0/vdm/sync_configuration/'
```
manjujames: Use string formatting to make this more readable.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1842,6 +2253,82 @@ def get():
...
+            return jsonify({'status':'errorNoMembers'})
+        for server_id in database[0]['members']:
+
+            server = [server for server in Global.SERVERS if server['id'] == server_id]
+            url = ('http://%s:%u/api/1.0/databases/%u/servers/%u/status/') % \
+                  (server[0]['hostname'], __PORT__, database_id, server[0]['id'])
+            try:
+                response = requests.get(url)
+            except:
+                return jsonify({'status':'error', 'hostname':server[0]['hostname']})
```
manjujames: log the exception details and send the exception error back as 'status'. Otherwise, we have no information about what went wrong.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1842,6 +2253,82 @@ def get():
...
+
+            if response.json()['status'] == "stalled":
+                if not has_stalled:
```
manjujames: You can avoid the 'if not has_stalled' check here and avoid clutter.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1842,6 +2253,82 @@ def get():
...
+                    has_stalled = True
+            elif response.json()['status'] == "running":
+                if not has_run:
```
manjujames: Same here.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1842,6 +2253,82 @@ def get():
...
+                    has_run = True
+            elif response.json()['status'] == "stopped":
+                if not has_stopped:
```
manjujames: same here.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1842,6 +2253,82 @@ def get():
...
+                    has_stopped = True
+            serverDetails.append({server[0]['hostname']: response.json()})
+
+        if has_stalled:
```
manjujames: I don't think you need this complicated logic here. It should simply be:
If any server is running, the database is running
else if any server is stalled, the database is stalled/starting
else the database is stopped

This also means that as soon as you find a server that is running, you can break out of the previous loop.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1927,11 +2401,12 @@ def main(runner, amodule, config_dir, server):
...
     DATABASE_DEPLOYMENT_VIEW = DatabaseDeploymentAPI.as_view('database_deployment_api')
+    STATUS_DATABASE_VIEW = StatusDatabaseAPI.as_view('status_database_api')
+    STATUS_DATABASE_SERVER_VIEW = StatusDatabaseServerAPI.as_view('status_database_server_view')
     VDM_VIEW = VdmAPI.as_view('vdm_api')
-    APP.add_url_rule('/api/1.0/servers/', defaults={'server_id': None},
-                     view_func=SERVER_VIEW, methods=['GET'])
-    APP.add_url_rule('/api/1.0/servers/<int:database_id>', view_func=SERVER_VIEW, methods=['POST'])
-    APP.add_url_rule('/api/1.0/servers/<int:server_id>', view_func=SERVER_VIEW,
+
+    APP.add_url_rule('/api/1.0/databases/<int:database_id>/servers/', view_func=SERVER_VIEW, methods=['POST'])
```
manjujames: Why is this a POST and not a GET?

lib/python/vdm/server/Log.py
```diff
@@ -0,0 +1,39 @@
...
+
+def get_error_log():
+
+    outfilename = os.path.join(HTTPListener.Global.PATH, get_most_recent_log())
+    try:
+        rfile = open(outfilename, 'r')
+        read_it = rfile.read()
+        myLine = ""
+        for line in read_it.splitlines():
+            if "ERROR:" in line:
```
manjujames: This will only get the oldest error. We should get all errors.

lib/python/vdm/server/Log.py
```diff
@@ -0,0 +1,39 @@
...
+                myLine = line
+                break
+        if myLine == "":
```
manjujames: If there are no errors, return something to the effect 'No errors were found in the log'

lib/python/vdm/server/voltdbserver.py
```diff
@@ -58,18 +61,36 @@ def ignore_signals():
...
+
 def get_voltdb_dir():
     """
     Utility method to get voltdb bin directory
     """
     return os.path.realpath(os.path.join(HTTPListener.Global.MODULE_PATH, '../../../..', 'bin'))
 
+
+def check_snapshot_folder():
+    outfilename = os.path.join(HTTPListener.Global.PATH, "voltdbroot/snapshots")
```
manjujames: You should pick up voltdbroot from the deployment xml.

lib/python/vdm/server/voltdbserver.py
```diff
@@ -124,7 +145,10 @@ def start_database(self, recover=False):
                 server_status[curr['hostname']] = str(err)
 
         if failed:
-            return create_response('There were errors starting servers: ' + str(server_status), 500)
+            url = ('http://%s:%u/api/1.0/databases/%u/status/') % \
```
manjujames: I don't think we need to do status request at this point. We can simply say there were some errors while sending start requests to server and the UI/user can ask for status as needed.

lib/python/vdm/server/voltdbserver.py
```diff
@@ -170,7 +194,8 @@ def check_and_start_local_server(self, recover=False):
         if (retcode == 0):
             return create_response('Success', 200)
         else:
-            return create_response('Error starting server', 500)
+            return create_response(Log.get_error_log(), 500)
```
manjujames: Here again we shouldn't return all the error logs. We can simply say there were errors and users can ask for error details as needed. This is a start database request anyway, which means none of the servers may be on this machine. So this error log request is not going to get anything pertinent.

lib/python/vdm/server/voltdbserver.py
```diff
@@ -347,3 +396,56 @@ def stop_server(self, server_id):
...
+
+        # Now stop each server
+        failed = False
+        server_status = {}
+        action = "stop"
+        for server_id in members:
+            server = [server for server in HTTPListener.Global.SERVERS if server['id'] == server_id]
+            curr = server[0]
+            try:
+                url = ('http://%s:%u/api/1.0/databases/%u/servers/%u/%s') % \
```
manjujames: Action cannot be anything other than 'stop' here, correct? If that is the case, there is no need to parameterize it here.

lib/python/vdm/server/HTTPListener.py
sujmanshak: This will be handled in the client side as discussed earlier

lib/python/vdm/server/voltdbserver.py
sujmanshak: Sending status request in this case has made us easy to expose individual error logs each cluster has got. We might have to rethink the logic of how to display the error log for each cluster if we remove this code

lib/python/vdm/server/Log.py
sujmanshak: This function has been removed since it is not used

lib/python/vdm/server/Log.py
sujmanshak: This function has been removed since it is not used

lib/python/vdm/server/HTTPListener.py
sujmanshak: This complicated logic is for getting the overall status of the database.

If all the servers are stalled, then the database is stalled.
if some cluster is running but some is stopped, the the database is stalled.
if no cluster is stalled and no cluster is stopped and all cluster is running, then the database is running.
if any cluster is stopped but not stalled and running, then the database is stopped

lib/python/vdm/server/HTTPListener.py
```diff
@@ -1020,11 +1225,17 @@ def get_deployment_properties(export_xml, is_list):
         for export in export_xml:
             new_export = {}
             for field in export:
-                new_export[field] = export[field]
+                if field == 'plaintext':
```
akhanzode: what is plaintext in export?

lib/python/vdm/server/voltdbserver.py
```diff
@@ -58,18 +61,44 @@ def ignore_signals():
...
 
+
 def get_voltdb_dir():
     """
     Utility method to get voltdb bin directory
     """
     return os.path.realpath(os.path.join(HTTPListener.Global.MODULE_PATH, '../../../..', 'bin'))
 
+
+def check_snapshot_folder(database_id):
```
akhanzode: Its not just the snapshot folder but command log as well. Also this should be something guarded by is_pro

akhanzode:
squash merged.

PR #3288, opened 2016-02-17
ENG-9605 support for update on export view

dweiss88:


tests/testprocs/org/voltdb_testprocs/regressionsuites/fixedsql/sql-update-ddl.sql
```diff
@@ -146,7 +146,7 @@ CREATE TABLE NO_JSON (
 );
 CREATE INDEX test_field ON NO_JSON (var2, field(var3,'color'));
 
-CREATE VIEW P1_VIEW AS
+CREATE VIEW P1_VIEW (ID, ID_COUNT, NUM_SUM) AS
```
manjujames: Why did you have to update this create view statement?

dweiss88: This was to name the fields so I can add a test that verifies that normal views can't be updated (TestSqlUpdateSuite).

PR #3289, opened 2016-02-17
ENG-9874 crash volt when export_overflow cant be written to

dweiss88:


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -494,6 +494,9 @@ public void discard() {
                                 }
                             }, uso, false));
                 } catch (IOException e) {
+                    if (e.getMessage().contains("No space left on device")) {
+                        VoltDB.crashLocalVoltDB("Unable to write to export overflow, no space left on device.");
+                    }
```
rmorgenstein: This seems overly specific. There are many reasons why it could not be written, such as a disk errors or permission errors.  Can you make it  more generic?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -494,10 +494,7 @@ public void discard() {
                                 }
                             }, uso, false));
                 } catch (IOException e) {
-                    exportLog.error(e);
-                    if (!deleted.get()) {
-                        cont.discard();
-                    }
+                    VoltDB.crashLocalVoltDB("Unable to write to export overflow.", true, e);
```
akhanzode: If the offer which throws the exception, deleted will never be set so crash right after eportLog.error()
Also see the sync path where IOException can occur as well. You should crash there too.
Have you tested this manually?

dweiss88: It's been tested manually using the steps provided in the ticket, but the other cases in which IOException can be thrown here haven't been tested.

akhanzode:
squash merged.

PR #3291, opened 2016-02-17
Eng 9845 postgres backend

johnTcrawford:
Added a QueryTransformer class, and updated the transformQuery method to make use of it, adding QueryTransformer object definitions (and a few new query Patterns), and eliminating a bunch of specialized methods that transformQuery used to call.

### DIFF
--- a/src/frontend/org/voltdb/NonVoltDBBackend.java
+++ b/src/frontend/org/voltdb/NonVoltDBBackend.java
@@ -58,7 +58,7 @@ public abstract class NonVoltDBBackend {
     protected static NonVoltDBBackend m_backend = null;
     protected String m_database_type = null;
     protected Connection dbconn;
-    protected static FileWriter TRANSFORMED_SQL_FILE_WRITER;
+    protected static FileWriter transformedSqlFileWriter;
     protected static final boolean DEBUG = false;
 
     /** Constructor specifying the databaseType (e.g. HSQL or PostgreSQL),
@@ -83,14 +83,14 @@ public abstract class NonVoltDBBackend {
         // If '-Dsqlcoverage.transform.sql.file=...' was specified on the
         // command line, print info when transforming SQL statements into
         // a format that the backend database can understand
-        String transformSqlOutputFile = System.getProperty("sqlcoverage.transform.sql.file", null);
-        if (transformSqlOutputFile == null) {
-            TRANSFORMED_SQL_FILE_WRITER = null;
+        String transformedSqlOutputFileName = System.getProperty("sqlcoverage.transform.sql.file", null);
+        if (transformedSqlOutputFileName == null) {
+            transformedSqlFileWriter = null;
         } else {
             try {
-                TRANSFORMED_SQL_FILE_WRITER = new FileWriter(transformSqlOutputFile, true);
+                transformedSqlFileWriter = new FileWriter(transformedSqlOutputFileName, true);
             } catch (IOException e) {
-                TRANSFORMED_SQL_FILE_WRITER = null;
+                transformedSqlFileWriter = null;
                 System.out.println("Caught IOException:\n    " + e
                         + "\nTransformed SQL output will not be printed.");
             }
@@ -460,10 +460,10 @@ public abstract class NonVoltDBBackend {
      *  the original and modified SQL are not the same, i.e., only if some
      *  transformation has indeed taken place. */
     static protected void printTransformedSql(String originalSql, String modifiedSql) {
-        if (TRANSFORMED_SQL_FILE_WRITER != null && !originalSql.equals(modifiedSql)) {
+        if (transformedSqlFileWriter != null && !originalSql.equals(modifiedSql)) {
             try {
-                TRANSFORMED_SQL_FILE_WRITER.write("original SQL: " + originalSql + "\n");
-                TRANSFORMED_SQL_FILE_WRITER.write("modified SQL: " + modifiedSql + "\n");
+                transformedSqlFileWriter.write("original SQL: " + originalSql + "\n");
+                transformedSqlFileWriter.write("modified SQL: " + modifiedSql + "\n");
             } catch (IOException e) {
                 System.out.println("Caught IOException:\n    " + e);
                 System.out.println("original SQL: " + originalSql);
--- a/src/frontend/org/voltdb/NonVoltDBBackend.java
+++ b/src/frontend/org/voltdb/NonVoltDBBackend.java
@@ -61,8 +61,7 @@ public abstract class NonVoltDBBackend {
     protected Connection dbconn;
 
     protected static final boolean DEBUG = false;
-    protected static boolean PRINT_TRANSFORMED_SQL = false;
-    protected static FileWriter TRANSFORMED_SQL_FILE_WRITER = null;
+    protected static final FileWriter TRANSFORMED_SQL_FILE_WRITER;
 
     /** Constructor specifying the databaseType (e.g. HSQL or PostgreSQL),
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -54,6 +60,10 @@
     protected String m_database_type = null;
     protected Connection dbconn;
 
+    protected static final boolean DEBUG = false;
+    protected static boolean PRINT_TRANSFORMED_SQL = false;
```
paulmartel: Eliminate the boolean member as redundant with the FileWriter being null.
Add a boolean accessor method returning (TRANSFORMED_SQL_FILE_WRITER != null), instead, if you like.
ALL_CAPS is usually reserved for final members -- declare TRANSFORMED_SQL_FILE_WRITER final.

src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -79,6 +106,378 @@ protected NonVoltDBBackend(Connection dbconn) {
...
+     * A QueryTransformer object is used to specify (using the builder pattern)
+     * which options should be used, when transforming SQL statements (DDL or
+     * DML) fitting certain patterns, as is often needed in order for the
+     * results of a backend database (e.g. PostgreSQL) to match those of VoltDB.
+     * This collection of options (the QueryTransformer object) is then passed
+     * to the <i>transformQuery</i> method (see below).
+     */
+    protected static class QueryTransformer {
+        // Required parameter
+        private Pattern queryPattern;
```
paulmartel: Lacking the conventional m_ member name prefix on all these.

src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -79,6 +106,378 @@ protected NonVoltDBBackend(Connection dbconn) {
...
+            System.out.println("In NonVoltDBBackend.isGeoColumn, with columnName: '" + columnName + "'");
+        }
+        if (columnName == null) {
+            return false;
+        }
+        String columnNameUpper = columnName.trim().toUpperCase();
+        return columnNameUpper.startsWith("PT") || columnNameUpper.startsWith("POLY");
+    }
+
+    /** Returns the number of occurrence of the specified character in the specified String. */
```
paulmartel: occurrence --> occurrences

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: Done.

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: Done.

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: Done.

src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -79,6 +106,378 @@ protected NonVoltDBBackend(Connection dbconn) {
...
+     *  inserted before the extra close-parens, instead of at the very end. */
+    static private String handleParens(String group, String prefix, String suffix) {
+        int numOpenParens  = numOccurencesOfCharIn(group, '(');
+        int numCloseParens = numOccurencesOfCharIn(group, ')');
+        if (numOpenParens >= numCloseParens || suffix.isEmpty()) {
+            return (prefix + group + suffix);
+        } else {  // numOpenParens < numCloseParens
+            int index;
+            if (numOpenParens == 0) {
+                index = indexOfNthOccurrenceOfCharIn(group, ')', 1) - 1;
```
paulmartel: subtracting 1 here in the "if" so we can unconditionally add 1 later?
Seems better to JUST add 1 in the else where it's needed.
And yet something smells suspicious about numOpenParens == 0 requiring special handling. Let me ponder on it.

src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -79,6 +104,373 @@ protected NonVoltDBBackend(Connection dbconn) {
...
+            result = transformQuery(result, qt);
+        }
+        return result;
+    }
+
+    /** Optionally (only if the <i>print</i> argument is true and
+     *  TRANSFORMED_SQL_FILE_WRITER is non-null), prints the original and
+     *  modified SQL statements, to a "Transformed SQL" output file, for the
+     *  current SQLCoverage test suite (if any). */
+    static protected void printTransformedSql(String originalSql, String modifiedSql, boolean print) {
```
paulmartel: The calculation and effect of the boolean print argument area bit strange.
Is is really serving any purpose other than to avoid doing anything for disabled or no-op transformations?
Some of the callers also test for a null input string, but I am doubtful that passing null into the statement transformer will or even SHOULD cause a quiet return, so I'd recommend asserting or otherwise ensuring that the input is not null (starting) somewhere well upstream of the calls to this function. With respect to no-op transformations being a universal turn-off, that could be factored down into this function as a comparison of its inputs without the need for an extra boolean argument. In general it seems needless to duplicate logic and silly to make a caller smart enough to tell the function not to do anything but NOT smart enough to simply skip the function call.
ddl was not null. ly trasnformer that you'd ever return from a transformation is a little strange

src/frontend/org/voltdb/NonVoltDBBackend.java
paulmartel: Here's what I've been able to ponder:
    If there is no open paren in group, then (we know that there is 1 or more close parens). The first close paren is extra and so we locate it and the -1 ensures that we insert our suffix just before it. This works as the method comment describes.
    If there is 1 open paren in group, then (we know that there are 2 or more close parens). The second close paren is extra so we locate the 1st close paren and (without the -1) insert our suffix just after it. This only strictly matches what the method comment describes if there is NOTHING between the last balanced close paren and the first unbalanced close paren.
   For numOpenParens > 1, we similarly base the placement on the last balanced paren, NOT the first unbalanced paren, so same discrepency.
Is it safe to assume that NOTHING will ever be in between these close parentheses for our use cases, OR is the behavior correct but the method description incorrect for some edge cases OR is there a bug? 
End of pondering -- end of review round -- end of another day of beauty.

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: Actually, I could not make TRANSFORMED_SQL_FILE_WRITER static and final and set it in the constructor; so I removed final.

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: There is no bug (this is not new code, it was just moved from PostgreSQLBackend.java), but you are correct that it is confusing. I've fixed that by, first, changing the code to use "+1" only in the case where it's needed, and, second, changing the comments to accurately describe what happens, namely that the suffix is inserted just after the last "matching" close-parens (rather than just before the first "extra" close-parens); and I also added a comment about the 0 open-parens special case.

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: Yes, good point(s). I've removed the "print" argument, and included the logic in the printTransformedSql method, to only print if the original and modified SQL are different. I have removed any null-check: the (original) SQL should never be null, and if it were, you would have problems long before you got here.

PR #3299, opened 2016-02-18
ENG-9875 Make VoltTableRow column methods public.

paulmartel:
These methods are already available on the VoltTable that provides the VoltTableRow.
Making them public also on VoltTableRow makes it more convenient to define stored procedure and client helper functions that accept a VoltTableRow (but not also its VoltTable) as input.


hhakimi53:
Looks good.

Thought: making getColumnIndex(String) might also be helpful.

PR #3302, opened 2016-02-18
ENG-9958. Update findbugs.

rmorgenstein:


rmorgenstein:
Too many files.  This cannot possibly require all this just to run in Jenkins.  I need to prune down to the necessary stuff.

PR #3303, opened 2016-02-19
Add a geospatial index type to the EE

wolffcm:
This just adds the implementation of the index itself and an EE unit test.


src/ee/indexes/CoveringCellIndex.h
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * Given a point, the first map facilitates scanning the index for
+ * potentially containing polygons.  The point is converted to a cell
+ * ID, and this cell (and all its containing cells) are located in the
+ * cell map.
+ *
+ * The second map, the tuple map, facilitates deletion of polygons
+ * from the index, so we do not need to recompute cell coverings when
+ * polygons are deleted.  (Computation of a cell covering is
+ * expensive.)
```
paulmartel: Aren't these comments confusing the first map with the second map?

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+    //    avg area of cells in level 11:        20.271 km^2.
+    //    avg area of cells in level 12:         5.068 km^2.
+    //    avg area of cells in level 13:         1.267 km^2.
+    //    avg area of cells in level 14:         0.317 km^2.
+    //    avg area of cells in level 15:         0.079 km^2.
+    //    avg area of cells in level 16:         0.020 km^2.
+    //    avg area of cells in level 17:         0.005 km^2.
+    //    avg area of cells in level 18:         0.001 km^2.
+    //    avg area of cells in level 19:         0.000 km^2.
+
```
BillWhite: This is somewhat trivial, but maybe this list would be more useful if the
units at finer levels were expressed in finer units.

src/ee/indexes/CoveringCellIndex.h
wolffcm: Yes, they are.  I'll fix it.

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+    TupleValueType cells;
+    for (int i = 0; i < MAX_CELL_COUNT; ++i) {
+        if (i < covering.size()) {
+            cells[i] = covering[i].id();
+        }
+        else {
+            cells[i] = S2CellId::Sentinel().id();
+        }
+    }
+
```
BillWhite: Is the data in the tuple map always MAX_CELL_COUNT long?  Is this a problem?  What if there is only
one cell in the polygon?  I guess it's only 8, so maybe it's not so important.

BillWhite:
That's all I've got.

src/ee/indexes/tableindexfactory.cpp
```diff
@@ -76,7 +78,7 @@ class TableIndexPicker
             if (m_type != BALANCED_TREE_INDEX) {
                 return new CompactingHashMultiMapIndex<TKeyType >(m_keySchema, m_scheme);
             } else if (m_scheme.countable) {
-                return new CompactingTreeMultiMapIndex<PointerKeyValuePair<TKeyType>, true>(m_keySchema, m_scheme);
+                    return new CompactingTreeMultiMapIndex<PointerKeyValuePair<TKeyType>, true>(m_keySchema, m_scheme);
```
paulmartel: bad re-indent

src/ee/storage/persistenttable.h
```diff
@@ -197,6 +199,7 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
     friend class ::CopyOnWriteTest;
     friend class ::CompactionTest_BasicCompaction;
     friend class ::CompactionTest_CompactionWithCopyOnWrite;
+    friend class CoveringCellIndexTest_TableCompaction;
```
paulmartel: I'm guessing that this new friend access (and possibly some or all of the existing friends) are really about a small number of compaction-related calls. Would it make sense to have one friend class -- something like PersistentTableAccessForTesting  -- that is a test utility class that defines static pass-through methods for these calls, so we don't have to tweak the product source each time we add a test?

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+#include "harness.h"
+#include "polygons.hpp"
+
+using std::unique_ptr;
+using std::chrono::microseconds;
+
+namespace voltdb {
+
+class CoveringCellIndexTest : public Test {
+public:
```
paulmartel: why public:?

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+    static bool isMemcheck() {
+#ifdef MEMCHECK
+        return true;
+#else
+        return false;
+#endif
+    }
```
paulmartel: You really must dislike ifdef syntax to have gone this far.
Strikes me as over-the-top, but I guess it doesn't hurt much.
If the choice is between seeing this get redefined in several places over time as needed or it having a home in a test harness header or even a product code header, I'd be in favor of finding it a home.

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+        for (int i = 2; i < 2 + numExtraCols; ++i) {
+            builder.setColumnAtIndex(i, VALUE_TYPE_VARBINARY, UNINLINEABLE_OBJECT_LENGTH - 1);
+        }
+        return builder.build();
+    }
+
+    // Create a tuple schema where the first two columns are
+    //   INTEGER
+    //   GEOGRAPHY(32767)
+    static TupleSchema* createTupleSchema() {
```
paulmartel: createTupleSchema seems like overkill for a local function vs. 0 default argument to createTupleSchemaWithExtraCols.

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+                                false, // countable
+                                "",    // expression as text
+                                "",    // predicate as text
+                                schema);
+        return TableIndexFactory::getInstance(scheme);
+    }
+
+    static CoveringCellIndex* createGeospatialIndex(const TupleSchema* schema) {
+        std::vector<int32_t> columnIndices;
+        columnIndices.push_back(1);
```
paulmartel: magic number 1 is just a little obscure.

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+        TableIndex* pkIndex = createPrimaryKeyIndex(table->schema());
+        table->addIndex(pkIndex);
+        table->setPrimaryKeyIndex(pkIndex);
+
+        return table;
+    }
+
+    static TableIndex* createPrimaryKeyIndex(const TupleSchema* schema) {
+        std::vector<int32_t> columnIndices;
+        columnIndices.push_back(0);
```
paulmartel: magic number 0 is just a little obscure.

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+    // In memcheck mode, only loads 50 rows.
+    void loadTable(PersistentTable* table) {
+        int rowLimit = -1;
+        if (isMemcheck()) {
+            // memcheck slows things down a lot...
+            rowLimit = 50;
+        }
+
+        std::cout << "\n            Loading polygons...\n";
+        std::istringstream infile(POLYGONS); // defined in polygons.hpp
```
paulmartel: The name "infile" threw me off. It's not a file. Maybe use instream.

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    // Create a tuple schema where the first two columns are
+    //   INTEGER
+    //   GEOGRAPHY(32767)
+    static TupleSchema* createTupleSchema() {
+        return createTupleSchemaWithExtraCols(0);
+    }
+
+    // Create a table with the schema described above.  Also add two
```
paulmartel: "with the schema described above" --> "the schema provided"?
OTOH, if the intent is always to be used with createTupleSchemaWithExtraCols,
why not just take that burden off the hands of the caller? Replace the schema with an int extra column count
defaulting to 0 and bypass the trivial createTupleSchema().

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+        tableTuple.tuple().setNValue(1, NValue::getNullValue(VALUE_TYPE_GEOGRAPHY));
+
+        auto start = std::chrono::high_resolution_clock::now();
+        std::chrono::microseconds usSpentDeleting = std::chrono::duration_cast<microseconds>(start - start);
+
+        while (numDeleted < numTuplesToDelete) {
+            int idOfTupleToDelete = std::rand() % totalTuples;
+            tableTuple.tuple().setNValue(0, ValueFactory::getIntegerValue(idOfTupleToDelete));
+            TableTuple tupleToDelete = table->lookupTupleByValues(tableTuple.tuple());
+            if (! tupleToDelete.isNullTuple()) {
```
paulmartel: This do-over algorithm for handling duplicate random numbers is OK where totalTuples is significantly greater than numTuplesToDelete but seems like beating your head against a wall for numTuplesToDelete == (totalTuples *0.99) where already-deleted tuples are going to eventually dominate.
Is this determination of tupleToDelete really that fast that these retries don't take forever?

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+        }
+
+        auto avgTotalUsSpentScanning = usSpentScanning.count() / numScans;
+        auto avgUsSpentContainsing = usSpentContainsing.count() / numScans;
+        auto avgUsSpentScanning = avgTotalUsSpentScanning - avgUsSpentContainsing;
+        std::cout << "              Average duration of each index lookup total: " << avgTotalUsSpentScanning << " us\n";
+        std::cout << "                Average duration spent on CONTAINS: " << avgUsSpentContainsing << " us\n";
+        std::cout << "                Average duration spent on B-tree traversal: " << avgUsSpentScanning << " us\n";
+
+        double pctFalsePositives = (double(numContainingCells - numContainingPolygons) / numContainingCells) * 100.0;
```
paulmartel: Just curious: what's the rationale for sometimes using auto and sometimes specific types like int and double?

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+        IndexCursor cursor(ccIndex->getTupleSchema());
+
+        bool b = ccIndex->moveToCoveringCell(&searchKey, cursor);
+        if (expectedTuples.size() > 0) {
+            EXPECT_TRUE(b);
+        }
+        else {
+            EXPECT_FALSE(b);
+            return;
+        }
```
paulmartel: The control flow seems a little awkward here vs.

        bool b = ccIndex->moveToCoveringCell(&searchKey, cursor);
        if (expectedTuples.empty()) {
            EXPECT_FALSE(b);
            return;
        }
        EXPECT_TRUE(b);

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+        EXPECT_EQ(expectedTuples.size(), foundTuples.size());
+
+        BOOST_FOREACH(int32_t expectedPk, expectedTuples) {
+            EXPECT_TRUE(foundTuples.find(expectedPk) != foundTuples.end());
+        }
+    }
+
+    static NValue wktToNval(const std::string& wkt) {
```
paulmartel: I'm not seeing the upside of having this generic function over polygon-specific and point-specific variants.
It just adds complexity to the known point cases to try them as polygons, throw and catch.

src/ee/indexes/CoveringCellIndex.cpp
wolffcm: Yes, you're right.  I cut and pasted this from our IW branch last September.  I'll regenerate it with meaningful output for all the levels.

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+    oldTuple.tuple().setNValue(1, geog1);
+
+    // old and new tuple now contain same instance of geography, no update required.
+    EXPECT_FALSE(ccIndex->checkForIndexChange(&oldTuple.tuple(), &newTuple.tuple()));
+}
+
+// This ought to go into harness.h, but I experienced compile failures
+// (on OS X) when I tried to put it there.  Something about a .h file
+// containing C++ code.
```
paulmartel: Wow. What?
I thought that header name sensitivity between ".hpp" and ".h" was a thing of the past.
I wonder if we renamed harness.h to harness.hpp if everything would work -- or I suppose we could
ADD harness.hpp, have it include the .h and C++ definitions and migrate tests as they needed the C++.

paulmartel: I don't think we have any .c files including harness.h, where this kind of error might be acceptable?

tests/ee/indexes/CoveringCellIndexTest.cpp
```diff
@@ -0,0 +1,658 @@
...
+                                                    topless,        // topend
+                                                    testPool.get(), // tempStringPool
+                                                    NULL,           // params
+                                                    NULL,           // engine
+                                                    "",             // hostname
+                                                    0,              // hostId
+                                                    NULL,           // drTupleStream
+                                                    NULL,           // drReplicatedStream
+                                                    0));            // drClusterId
+
```
paulmartel: We usually leave main as "return TestSuite::globalInstance()->runAll();" and put custom setup code like this into the members and constructor of the Test subclass so it gets called magically by TestSuite::globalInstance()->runAll();
I don't see much advantage/disadvantage to following or flouting this convention. Do you?

src/ee/indexes/CoveringCellIndex.cpp
wolffcm: Yes, the data in the tuple map is always eight 64-bit unsigned ints long.

In the unit test, we used around 7 cells per polygon, which is not a huge amount of wasted space.  My impression is the covering algorithm will use as many cells as it can to approximate the polygon.  For polygons that closely match the shape of a cell, we might have 7 wasted elements here, but that seems pretty rare.

Solving the problem of having a value type that is variable length seems like it would open a can of worms, so I opted for this simple solution.

src/ee/storage/persistenttable.h
wolffcm: Yes, that might make sense.  There is already the PersistentTableSurgeon but I couldn't figure out how to get an instance of it after staring at the code for a while.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: You're right, the methods could be protected.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: It's true I dislike preprocessor syntax for things like this.  I can move it to harness.h.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: Ok, I'll change it.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: I'll add a named constant.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: Good point.  It started off as a file, but there's set up code that sets the working directory to `/`, so a file wasn't practical.  I'll rename the variable.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: I only want to add the extra columns when testing table compaction.  It probably doesn't matter though.  Having the extra columns does not seem to affect run time that much.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: It's doesn't take forever.  The test completes in a few seconds.  There's no reason to delete random tuples though.  I can change it so it deletes a range of records based on primary key value.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: I'm not so familiar with the `std::` timer methods and their return types.  When I wasn't sure, I used `auto` because I was lazy.

I'm not yet sure what the right guiding principles are for the use of `auto`.  Herb Sutter seems to like `auto` a lot because you can do stuff like change the return types of functions without changing any call sites.

There seems to be some value in knowing the types of variables, but maybe having better IDEs that can do type inference reduces the need to put the types explicitly in the program.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: I'll change it.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: Ok, I'll change it.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: I'm not sure exactly what was going on.  I went down the road of renaming harness.h, and it turned out to be more work than I wanted to bite off at that moment.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: Either way is fine with me.  I'm happy to follow convention.  Creating an executor context seems to be common enough that it could even go into the Test superclass.

tests/ee/indexes/CoveringCellIndexTest.cpp
paulmartel: why protected?

tests/ee/indexes/CoveringCellIndexTest.cpp
paulmartel: I wasn't suggesting always having extra columns -- I was suggesting having a specifiable number of 0 or more columns (that you could even default to 0), but more to the point moving the redundant schema building call into the table building code.

tests/ee/indexes/CoveringCellIndexTest.cpp
paulmartel: There MAY be a reason for -- or more importantly an effect of deleting random tuples instead of contiguous ranges as the former seems more likely to leave large numbers of pages with small numbers of surviving tuples. The latter would be more likely to clear entire pages, which would cause them to be immediately freed without much call for compaction.

tests/ee/indexes/CoveringCellIndexTest.cpp
paulmartel: OTOH - there's something to be said for keeping VoltDB-isms out of the generic Test framework, so maybe we should factor out an intermediate TestVoltDB base class that layers in the most common VoltDB-isms.

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+}
+
+
+static CoveringCellIndex::CellKeyType setKeyFromCellId(uint64_t cellId, const TableTuple* tuple) {
+    CoveringCellIndex::CellKeyType key;
+    int keyOffset = 0;
+    int intraKeyOffset = static_cast<int>(sizeof(uint64_t) - 1);
+    key.insertKeyValue<uint64_t>(keyOffset, intraKeyOffset, cellId);
+    if (tuple != NULL)
+        key.setValue(tuple->address());
```
paulmartel: Coding standard wants braces around single-line blocks.

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+    coverer.set_max_cells(CoveringCellIndex::MAX_CELL_COUNT);
+    coverer.set_level_mod(CELL_LEVEL_MOD);
+    coverer.GetCovering(poly, coveringCells);
+}
+
+
+static CoveringCellIndex::CellKeyType setKeyFromCellId(uint64_t cellId, const TableTuple* tuple) {
+    CoveringCellIndex::CellKeyType key;
+    int keyOffset = 0;
+    int intraKeyOffset = static_cast<int>(sizeof(uint64_t) - 1);
```
paulmartel: Clean up the code by defining these two constants as constants.

src/ee/common/ValuePeeker.hpp
```diff
@@ -98,6 +98,14 @@ class ValuePeeker {
         return value.getDecimal();
     }
 
+    static const GeographyValue peekGeographyValue(const NValue& value) {
+        return value.getGeographyValue();
+    }
```
paulmartel: Seems like these functions and their callers are doing a lot of needless copy construction
because they don't use "const ... &" declarations.

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+
+static void* extractTupleAddress(const CoveringCellIndex::TupleKeyType &key) {
+    int keyOffset = 0;
+    int intraKeyOffset = static_cast<int>(sizeof(void*) - 1);
+    return reinterpret_cast<void*>(key.extractKeyValue<uint64_t>(keyOffset, intraKeyOffset));
+}
+
+
+static CoveringCellIndex::CellMapIterator& castToIter(IndexCursor& cursor) {
```
paulmartel: I found these "castTo..." names to be confusing -- their only secondarily casting. I'd prefer "get...".

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+    S2CellId cell = S2CellId(extractCellId(mapIter.key()));
+
+    mapIter.moveNext();
+    if (mapIter.equals(mapEndIter)) {
+        // No more matches at the current level, but check the lower levels
+        // (that is, the cells that contain the one we just checked).
+        int nextLevel = cell.level() - CELL_LEVEL_MOD;
+
+        while (nextLevel >= MIN_CELL_LEVEL) {
```
paulmartel: Should this be a a 'for loop' as elsewhere?

src/ee/indexes/CoveringCellIndex.cpp
```diff
@@ -0,0 +1,466 @@
...
+        }
+
+        // No more matches.
+        cursor.m_match.move(NULL);
+        return retval;
+    }
+
+    // There is another match at the current level
+    cursor.m_match.move(const_cast<void*>(mapIter.value()));
+    return retval;
```
paulmartel: The above could be a little simpler:

        S2CellId cell = S2CellId(extractCellId(mapIter.key()));
        int nextLevel = cell.level();
    
        mapIter.moveNext();
        while (mapIter.equals(mapEndIter)) {
            // No more matches at the current level, but check the lower levels
            // (that is, the cells that contain the one we just checked).
            nextLevel -= CELL_LEVEL_MOD;
    
            if (nextLevel < MIN_CELL_LEVEL) {
                // No more matches.
                cursor.m_match.move(NULL);
                return retval;
            }
            cell = cell.parent(nextLevel);
            CellMapRange iterPair = m_cellEntries.equalRange(setKeyFromCellId(cell.id()));
    
            mapIter = iterPair.first;
            mapEndIter = iterPair.second;
        }
    
        // There is another match at the current level
        cursor.m_match.move(const_cast<void*>(mapIter.value()));
        return retval;

paulmartel:
end of round.

src/ee/storage/persistenttable.h
wolffcm: This wasn't immediately straightforward so I'll file a separate ticket for this.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: Because each test created with the `TEST_F` macro actually creates a subclass of  `CoveringCellIndexTest` whose code is in a member method called `run`.  So methods I want to call in the parent class can't be made private.

tests/ee/indexes/CoveringCellIndexTest.cpp
wolffcm: Yes, that occurred to me.  I started thinking of how to delete tuples in a way that doesn't do contiguous ranges (by reversing the bits of the PK of the row to be deleted, for example) but really, deleting 99% of rows happens pretty much instantaneously.  The run time of the test is taken up by inserting rows more than anything else.

I'll add a comment explaining this.

src/ee/common/ValuePeeker.hpp
wolffcm: You must mean for the return value.  Returning a reference (even a const one) isn't appropriate here since the `GeographyValue` returned from NValue.hpp will be a temporary object, and will be out of scope in the caller of this method.  GeographyValue is a little different from the other data types because it's not a simple cast of the data bytes in the NValue.

This isn't necessarily bad for performance though since the return value optimization is likely to kick in.

src/ee/indexes/CoveringCellIndex.cpp
wolffcm: These can't be const because the callee is expecting a non-const ref.  Their typical use is with multi-component indexes where the callee updates them with values to use on the next call.

src/ee/indexes/CoveringCellIndex.cpp
wolffcm: I agree, that reads much nicer.

paulmartel:
End of review round

wolffcm:
Merged in a separate branch.

PR #3306, opened 2016-02-19
VDM-46: use all network options. Shutdown work also needs this..

akhanzode:


lib/python/vdm/server/voltdbserver.py
```diff
@@ -237,7 +240,14 @@ def Get_Voltdb_Process(self):
...
         return VoltdbProcess
     
-    def start_local_server(self, recover=False):
+    def start_local_server(self, sid, recover=False):
+        """
+        Find server from id if its missing bail out.
+        """
+        server = [server for server in HTTPListener.Global.SERVERS if server['id'] == sid]
+        if not server:
+            return 0
```
manjujames: 0 indicates success. This should return non-zero so that we will return an error status to the user.

lib/python/vdm/server/voltdbserver.py
```diff
@@ -237,7 +240,14 @@ def Get_Voltdb_Process(self):
 
         return VoltdbProcess
     
-    def start_local_server(self, recover=False):
+    def start_local_server(self, sid, recover=False):
+        """
```
manjujames: This makes it look like this is the comment for the method. The comment for the method with """ got moved down. We should move that up and this can just be a # comment.

lib/python/vdm/server/voltdbserver.py
```diff
@@ -252,7 +262,9 @@ def start_local_server(self, recover=False):
         if recover:
             verb = 'recover'
         voltdb_cmd = [ 'nohup', os.path.join(voltdb_dir, 'voltdb'), verb, '-d', filename, '-H', primary ]
-    
+        self.build_network_options(server, voltdb_cmd)
```
manjujames: We already check that server has something in it, so we could just pass server[0] here to build_network_options.

lib/python/vdm/server/voltdbserver.py
```diff
@@ -252,7 +262,9 @@ def start_local_server(self, recover=False):
...
+        print voltdb_cmd
```
manjujames: Was this print stmt left here by accident or do we actually want to print it?

lib/python/vdm/server/voltdbserver.py
```diff
@@ -275,6 +287,52 @@ def start_local_server(self, recover=False):
         else:
             return 1
 
+    # Build network options for command line.
+    def build_network_options(self, server, voltdb_cmd):
+        if len(server) == 0:
+            return
+        sconfig = server[0]
+        ii = sconfig['internal-interface']
```
manjujames: This will look much better if we write a utility method that takes in the configuration name and option name and appends to voltdb_cmd. For example, something like:
add_voltdb_option(config_name, option_name, voltdb_cmd):
...

manjujames:
Looks good to me.

PR #3307, opened 2016-02-20
Eng 8932 full join

alexeevm:
Paul, 

I refactored the Aggregate and MergeReceive executors to use CountingPostfilter, plus couple of bug fixes for nestloopindex executor.
I think there was a bug in the AggregateExecutorBase::insertOutputTuple - the offset was evaluated only if limit is greater or equal to zero. If there were no limit set (-1), the offset was ignored. The new logic lives now inside the AggCountingPostfilter::eval and verifies the offset first prior to checking the limit (if tuple passes a predicate). I added a few test cases to verify the aggregated SQL with OFFSET only 

Mike

### DIFF
--- a/src/ee/executors/executorutil.cpp
+++ b/src/ee/executors/executorutil.cpp
@@ -47,6 +47,7 @@
 
 #include "common/tabletuple.h"
 #include "expressions/abstractexpression.h"
+#include "storage/temptable.h"
 
 namespace voltdb {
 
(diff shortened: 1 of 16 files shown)


src/ee/executors/executorutil.cpp
```diff
@@ -72,4 +73,53 @@ bool CountingPostfilter::eval(const TableTuple* outer_tuple, const TableTuple* i
...
+
+// Returns true if predicate evaluates to true and LIMIT/OFFSET conditions are satisfied.
+bool AggCountingPostfilter::eval(const TableTuple* tuple) {
+    assert(m_table);
+    if (m_postfilter == NULL || m_postfilter->eval(tuple, NULL).isTrue()) {
+        // Verify the OFFSET first
+        if (m_tuple_skipped < m_offset) {
+                m_tuple_skipped++;
+                return false;
+            }
```
paulmartel: 3 lines oddly indented.

src/ee/executors/executorutil.cpp
```diff
@@ -72,4 +73,53 @@ bool CountingPostfilter::eval(const TableTuple* outer_tuple, const TableTuple* i
...
+                }
+                return false;
+            }
+        }
+        // LIMIT/OFFSET are satisfied
+        return true;
+    }
+    // Predicate is not NULL and was evaluated to FALSE
+    return false;
+}
```
paulmartel: I'm undecided whether there is enough code duplication between AggCountingPostfilter and CountingPostfilter to justify inheritence.

paulmartel:
End of review round. I'm pushing this into branch testing now.
I'll let you know how that works out.

alexeevm:
I submitted a new pull request where both post filters are merged into a single class
https://github.com/VoltDB/voltdb/pull/3330

PR #3308, opened 2016-02-22
Vdm 31

deerwalk:


lib/python/vdm/server/Validation.py
```diff
@@ -309,7 +309,7 @@ class ServerInputs(Inputs):
                     "properties": {
                         "prefix": {
                             "id": "prefix",
-                            "type": "string"
+                            "type": "boolean"
```
akhanzode: This is undoing my changes prefix needs to be string.

manjujames:
This looks fine to me once you take care of Anish's comment.

akhanzode:
why change mode of volt deploy it should be 755

PR #3309, opened 2016-02-22
Vdm 51

deerwalk:


lib/python/vdm/server/HTTPListener.py
```diff
@@ -2215,38 +2215,41 @@ def get(database_id):
...
-            status.append({'status': 'stalled'})
-        elif has_run == True and has_stopped:
-            status.append({'status': 'stalled'})
-        elif not has_stalled and not has_stopped and has_run:
-            status.append({'status': 'running'})
-        elif has_stopped and not has_stalled and not has_run:
-            status.append({'status': 'stopped'})
-
-        isFreshStart = voltdbserver.check_snapshot_folder(database_id)
+        if len(database) != 0:
```
manjujames: More readable if you check the negative case first and exit out. Something like:

if not database:
    make_response(...., 404)
Now the rest of the code.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -2254,27 +2257,31 @@ class StatusDatabaseServerAPI(MethodView):
...
-                client = voltdbclient.FastSerializer(str(server[0]['hostname']), 21212)
-                proc = voltdbclient.VoltProcedure(client, "@Ping")
-                response = proc.call()
-                return jsonify({'status': "running"})
-            except:
-                voltProcess = voltdbserver.VoltDatabase(database_id)
-
-                error = ''
+        database = [database for database in Global.DATABASES if database['id'] == database_id]
+        if len(database) !=0:
```
manjujames: Same thing here. Check 'if not database' first and exit out early. Less indentation that way and easier to read.

PR #3318, opened 2016-02-23
Eng 9067 new

akhanzode:


doc/tutorials/auction/deployment.xml
```diff
@@ -4,13 +4,14 @@
     <httpd enabled="true">
         <jsonapi enabled="true" />
     </httpd>
+<!-- If you are running community edition disable export section -->
     <export>
-        <configuration stream="oldfile" enabled="true" type="file">
+        <configuration target="oldfile" enabled="true" type="file">
```
vtkstef: add skip internals to true

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -760,8 +762,45 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp)
 
             // set export info on the new table
             if (tcd->exportEnabled()) {
-                table->setSignatureAndGeneration(catalogTable->signature(), timestamp);
-                m_exportingTables[catalogTable->signature()] = table;
+                tcd->getTable()->setSignatureAndGeneration(catalogTable->signature(), timestamp);
+                m_exportingTables[catalogTable->signature()] = tcd->getTable();
+                StreamedTable *streamedtable = dynamic_cast<StreamedTable *>(tcd->getTable());
+                //Deal with views if we have
```
vtkstef: encapsulate this section as one routine so if wee need to modify this section we won't need to do it into two different places

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -795,6 +834,42 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp)
...
+
+                const catalog::CatalogMap<catalog::MaterializedViewInfo> & views = catalogTable->views();
+
+                streamedtable->segregateMaterializedViews(views.begin(), views.end(),
+                                                            survivingInfos, survivingViews,
+                                                            obsoleteViews);
+
+                for (int ii = 0; ii < survivingInfos.size(); ++ii) {
+                    catalog::MaterializedViewInfo * currInfo = survivingInfos[ii];
+                    PersistentTable* oldTargetTable = survivingViews[ii]->targetTable();
```
vtkstef: see comment above

src/ee/storage/ExportMaterializedViewMetadata.cpp
```diff
@@ -0,0 +1,821 @@
...
+
+void ExportMaterializedViewMetadata::setFallbackExecutorVectors(const catalog::CatalogMap<catalog::Statement> &fallbackQueryStmts) {
+    m_fallbackExecutorVectors.clear();
+    m_usePlanForAgg.clear();
+    VoltDBEngine* engine = ExecutorContext::getEngine();
+    int idx = 0;
+    BOOST_FOREACH (LabeledStatement labeledStatement, fallbackQueryStmts) {
+        catalog::Statement *stmt = labeledStatement.second;
+        const string& b64plan = stmt->fragments().begin()->second->plannodetree();
+        const string jsonPlan = engine->getTopend()->decodeBase64AndDecompress(b64plan);
```
vtkstef: can't believe we call java to do this. can we have a native implementation for c++ ?

src/ee/storage/ExportMaterializedViewMetadata.cpp
```diff
@@ -0,0 +1,821 @@
...
+#include "boost/foreach.hpp"
+#include "boost/shared_array.hpp"
+
+ENABLE_BOOST_FOREACH_ON_CONST_MAP(Statement);
+typedef std::pair<std::string, catalog::Statement*> LabeledStatement;
+
+using namespace std;
+namespace voltdb {
+
+ExportMaterializedViewMetadata::ExportMaterializedViewMetadata(StreamedTable *srcTable,
```
vtkstef: refactor to create a common parent class for both export and non export materialized views metadata

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -510,6 +511,56 @@ void migrateViews(const catalog::CatalogMap<catalog::MaterializedViewInfo> & vie
     }
 }
 
+void migrateExportViews(const catalog::CatalogMap<catalog::MaterializedViewInfo> & views,
+                  StreamedTable *existingTable, StreamedTable *newTable,
```
vtkstef: if the materialized views implementations shared a common parent class you could omit this code duplication

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1440,6 +1453,29 @@ private static void setExportInfo(Catalog catalog, ExportType exportType) {
...
+        // If both are specified, they must be equal
+        if (targetName!=null && streamName!=null && !targetName.equals(streamName)) {
+            throw new RuntimeException("Only one of 'target' or 'stream' attribute must be specified");
+        }
+        // handle old deployment files, which will only have 'stream' specified
+        if (targetName==null) {
+            targetName = streamName;
+        }
+        if (targetName == null) {
+            throw new RuntimeException("Target must be specified along with type in export configuration.");
```
vtkstef: is there a better more descriptive exception class for this?

src/ee/storage/TableCatalogDelegate.cpp
akhanzode: Yeah but Streamed table and persistent table are so different...we may need to add a new UserDefinedTable before we can use common mat view code.

doc/tutorials/auction/deployment.xml
akhanzode: sure will do

src/ee/executors/insertexecutor.cpp
```diff
@@ -230,7 +234,7 @@ bool InsertExecutor::p_execute(const NValueArray &params) {
 
             // if it doesn't map to this site
             if (!isLocal) {
-                if (!m_multiPartition) {
+                if (!m_multiPartition && !m_isStreamed) {
```
nshi: In what case would you insert a tuple that doesn't belong to this partition in a single partition procedure?

src/ee/executors/insertexecutor.cpp
```diff
@@ -248,11 +252,15 @@ bool InsertExecutor::p_execute(const NValueArray &params) {
         // with partitioned tables, we need to perform the insert on
         // every partition.
         if (m_isStreamed && m_multiPartition && !m_sourceIsPartitioned) {
-            bool isLocal = m_engine->isLocalSite(ValueFactory::getBigIntValue(0));
+            bool isLocal;
+            if (m_partitionColumn != -1) {
+                isLocal = m_engine->isLocalSite(templateTuple.getNValue(m_partitionColumn));
```
nshi: Now that the `m_partitionColumn` is set correctly for export tables, this case is handled in the if block above. You don't need to check it again here.

src/ee/executors/updateexecutor.h
```diff
@@ -89,6 +89,7 @@ class UpdateExecutor : public AbstractExecutor
     TableTuple m_inputTuple;
     int m_partitionColumn;
     bool m_partitionColumnIsString;
+    bool m_isMaterialized;
```
nshi: I don't see any user of this member variable.

akhanzode: I will remove it. leftover from original hack.

src/ee/executors/insertexecutor.cpp
akhanzode: Yeah thats the redundency I was talking I will remove.

src/ee/executors/insertexecutor.cpp
akhanzode: if SP is partitioned and its insert into partitioned export table.

nshi: In that case, the tuple would belong to this partition, right?

src/ee/executors/insertexecutor.cpp
```diff
@@ -230,29 +236,42 @@ bool InsertExecutor::p_execute(const NValueArray &params) {
...
+                                templateTuple,
+                                "Mispartitioned tuple in single-partition insert statement.");
+                    } else {
+                        if (m_hasStreamView) {
+                            StreamedTable *stable = dynamic_cast<StreamedTable*>(targetTable);
+                            throw ConstraintFailureException(stable,
+                                    templateTuple,
+                                    "Mispartitioned tuple in single-partition insert statement.");
+                        } else {
+                            //Old style export table let is slide.
```
nshi: The comment here needs to be more specific. It'll be without context several months later. The next person who look at this may not know what "old style" is.

src/ee/storage/ConstraintFailureException.cpp
```diff
@@ -45,29 +46,58 @@ ConstraintFailureException::ConstraintFailureException(
...
+    m_type(CONSTRAINT_TYPE_PARTITIONING)
+{
+    assert(table);
+    assert(!tuple.isNullTuple());
+}
+
+ConstraintFailureException::ConstraintFailureException(
+        StreamedTable *table,
+        TableTuple tuple,
+        string message) :
```
nshi: I don't see the point of adding another flavor of constructor for streamed table here. All the methods the `ConstraintFailureException` cares about are actually methods of the base `Table` class. In fact, you can just change the existing constructor that takes a `PersistentTable` to take a `Table` instead.

src/ee/storage/ConstraintFailureException.cpp
```diff
@@ -45,29 +46,58 @@ ConstraintFailureException::ConstraintFailureException(
...
     SQLException(
             SQLException::integrity_constraint_violation,
             message,
             VOLT_EE_EXCEPTION_TYPE_CONSTRAINT_VIOLATION),
-    m_table(table),
+    m_Streamtable(table),
```
nshi: By the way, I don't think C++ initializes class members with NULL by default. Without initializing `m_table` here, you may get garbage data later when you use `m_table` for NULL check.

src/ee/storage/ConstraintFailureException.h
```diff
@@ -62,6 +72,7 @@ class ConstraintFailureException: public SQLException {
     void p_serialize(ReferenceSerializeOutput *output) const;
 
     PersistentTable *m_table;
+    StreamedTable *m_Streamtable;
```
nshi: Per the comment in `ConstraintFailureException.cpp`, I don't think this is necessary.

akhanzode: Sure I will refactor.

PR #3324, opened 2016-02-24
Add gaurds in EE to geo-functions where caller can accidentally pass in string instead of geo-types

hhakimi53:
Theoretically user can pass in arguments of different types to voltQueueSql() than the one which has been planned for using the corresponding statement. This can result in incompatible type flowing to EE and if the implicit conversion is not defined it can result in undefined behavior in release build (for debug builds, logic today asserts) . For this ticket adding guards in EE for geo types, so that EE does not do process the request where the supplied argument is not as expected geo-type.

### DIFF
--- a/src/ee/common/NValue.hpp
+++ b/src/ee/common/NValue.hpp
@@ -955,8 +955,16 @@ private:
         return *reinterpret_cast<bool*>(m_data);
     }
 
+    static inline void generateExceptionIfUnexpectedType(ValueType expected, ValueType actual) {
+        if (actual != expected) {
+            throwCastSQLException(actual, expected);
+        }
+    }
+
     GeographyPointValue& getGeographyPointValue() {
-        assert(getValueType() == VALUE_TYPE_POINT);
+        const ValueType valueType = getValueType();
+        assert(valueType == VALUE_TYPE_POINT);
+        generateExceptionIfUnexpectedType(VALUE_TYPE_POINT, valueType);
         return *reinterpret_cast<GeographyPointValue*>(m_data);
     }
 
@@ -965,12 +973,16 @@ private:
         BOOST_STATIC_ASSERT_MSG(sizeof(GeographyPointValue) <= sizeof(m_data),
                                 "Size of Point is too large for NValue m_data");
 
-        assert(getValueType() == VALUE_TYPE_POINT);
+        const ValueType valueType = getValueType();
+        assert(valueType == VALUE_TYPE_POINT);
+        generateExceptionIfUnexpectedType(VALUE_TYPE_POINT, valueType);
         return *reinterpret_cast<const GeographyPointValue*>(m_data);
     }
 
     const GeographyValue getGeographyValue() const {
-        assert(getValueType() == VALUE_TYPE_GEOGRAPHY);
+        const ValueType valueType = getValueType();
+        assert(valueType == VALUE_TYPE_GEOGRAPHY);
+        generateExceptionIfUnexpectedType(VALUE_TYPE_GEOGRAPHY, valueType);
 
         if (isNull()) {
             return GeographyValue();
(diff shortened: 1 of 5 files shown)


src/ee/common/NValue.hpp
```diff
@@ -965,12 +973,16 @@ class NValue {
         BOOST_STATIC_ASSERT_MSG(sizeof(GeographyPointValue) <= sizeof(m_data),
                                 "Size of Point is too large for NValue m_data");
 
-        assert(getValueType() == VALUE_TYPE_POINT);
+        const ValueType valueType = getValueType();
```
wolffcm: The other "get" methods all assume that the value type of `this` is the same as what's being requested.  These methods are used a lot, so doing type-checking here is probably not so good for performance.  Surely the codepath which generated the undefined behavior did not use the "get" methods?  I think you can leave these methods as-is.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctionsExtended.java
```diff
@@ -0,0 +1,159 @@
...
+import org.voltdb.client.Client;
+import org.voltdb.client.NoConnectionsException;
+import org.voltdb.client.ProcCallException;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.regressionsuites.TestGeospatialFunctions.Border;
+import org.voltdb.types.GeographyPointValue;
+import org.voltdb_testprocs.regressionsuites.failureprocs.GeoPointProcsWithIncompatibleParameter;
+import org.voltdb_testprocs.regressionsuites.failureprocs.GeographyProcsWithIncompatibleParameter;
+
+// the negative test case that triggers assert in valgrind and debug environment are placed
```
wolffcm: The name of this class seems a little unintuitive?  I thought that the TestGeospatialFunctions test was specifically for the geo functions, like CONTAINS, DISTANCE, etc.  This test is more like TestGeospatialNegative?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctionsExtended.java
```diff
@@ -0,0 +1,159 @@
...
+// in this file
+
+public class TestGeospatialFunctionsExtended extends RegressionSuite {
+
+    public TestGeospatialFunctionsExtended(String name) {
+        super(name);
+    }
+
+    static private void setupGeoSchema(VoltProjectBuilder project) throws IOException {
```
wolffcm: Is a lot of this code copied and pasted from elsewhere?  I'd be okay with you just putting the two new tests into TestGeospatialFunctions I guess.  Maybe that class could be renamed as TestGeospatialQueries?

src/ee/common/NValue.hpp
hhakimi53: All VoltDB geo functions operating on geo types are calling the getgeoxxx() and are fetching geo values through them even when the type was not geo. So that's it seemed optimal to put the check in. The undefined behavior happens after this call has been executed as the later call are operating with assumption they are operating on Geo values. If we run the logic in debug mode, assertion will get triggered in the getter function sometimes as the VoltDB functions which called getGeographyxxx(), did not perform type-check (debug environment only) explicitly in there before calling this function. If we don't want it to be here, an alternate can be to have the above check in the calling functions themselves and add those checks individually to all the functions though with that approach we will have to make new function that gets added will have to the same logic. This  specific  getter "const GeographyPointValue& getGeographyPointValue() const" may not be getting called all the time and only in selected case. So can remove it from and apply the check outside to the callers of it.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctionsExtended.java
hhakimi53: The code that is duplicate with TestGeoSpatialFunctions is the populating point table.  
To start with these two new tests where in TestGeoSpatialFunctions but as these two tests will result into assertion when ran with debug version of EE , so I had moved these two tests into new file so that it does not effect TestGeoSpatialFunctions in development when using debug version of EE. That's why had named it TestGeoSpatialFunctionsExtended.

hhakimi53:
closing to discard the pull request

PR #3327, opened 2016-02-24
ENG-8545: Only crash the local server if remote action is also not recover

manjujames:
Tested manually.
Running system tests now on the branch to see if rejoin/recover cluster related tests continue to work.

### DIFF
--- a/src/frontend/org/voltdb/RestoreAgent.java
+++ b/src/frontend/org/voltdb/RestoreAgent.java
@@ -1165,7 +1165,7 @@ SnapshotCompletionInterest, Promotable
                 if (remoteRecover != recover) {
                     String msg = "Database actions are not consistent. Remote node action is not 'recover'. " +
                                  "Please enter the same database action on the command-line.";
-                    VoltDB.crashLocalVoltDB(msg, true, null);
+                    VoltDB.crashLocalVoltDB(msg, false, null);
                 }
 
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -1163,9 +1163,9 @@ private Long deserializeRestoreInformation(List<String> children,
                 }
                 int remoteRecover = json.getInt("action");
                 if (remoteRecover != recover) {
-                    String msg = "Database actions are not consistent, please enter " +
-                        "the same database action on the command-line.";
-                    throw new RuntimeException(msg);
+                    String msg = "Database actions are not consistent. Remote node action is not 'recover'. " +
+                                 "Please enter the same database action on the command-line.";
+                    VoltDB.crashLocalVoltDB(msg, true, null);
```
nshi: The stack trace would not be of much use here since the error message is pretty clear about what happened.

akhanzode: Test add to make sure it does not fall in this trap as in case of add current cluster is either with "recover" or "create" and new one will be "add"

manjujames:
Not logging the stack trace anymore.
Tested all add cases that we talked about. I will check jenkins and merge, if it is OK with you guys.

nshi:
Looks good to me.

manjujames:
squash merged.

PR #3328, opened 2016-02-24
ENG-9871: File export doesn't overflow if .CSV file can't be written

JadenW:
the only change to CSVWriter is just a resetWriter method. But I have to change the file format from dos to unix in order to get rid of ^M for newlines.


third_party/java/src/au/com/bytecode/opencsv_voltpatches/CSVWriter.java
```diff
@@ -1,334 +1,338 @@
...
+     */
+    public boolean checkError() {
+        return pw.checkError();
+    }
+
+    public void setResultService(ResultSetHelper resultService) {
+        this.resultService = resultService;
+    }
+
+    public void resetWriter() {
```
akhanzode: Add comment that this is our extension to third party api. Why does diff shows whole file has changed as opposed to just the added method?

JadenW: I have transformed the file from dos format to unix.
Otherwise the newline is ^M which couldn't pass the licence check.

PR #3330, opened 2016-02-25
Merged CountinPostfilter and AggCountingPostfilter into a single class

alexeevm:
Paul,
Single AggCountingPostfilter can do the job. I eliminated the CountingPostfilter class and renamed the AggCountingPostfilter to be just CountingPostfilter. It has a minor overhead (the NULL check for the parent post filter) in case of non-aggregated executors (the parent will be NULL there), but I don't think it has any performance implications.
Mike


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2828,7 +2827,9 @@ private static void simplifyOuterJoinRecursively(BranchNode joinNode, List<Abstr
             exprs.add(rightNode.getWhereExpression());
         }
 
-        if (joinNode.getJoinType() == JoinType.INNER || joinNode.getJoinType() == JoinType.FULL) {
+        // In case of outer join, ON expressions can only be NULL rejecting for the inner node.
+        // The outer node does not introduce NULLs
+        if (joinNode.getJoinType() == JoinType.INNER) {
```
paulmartel: This code could be made more readable by replacing the if/else chain with a switch statement.
Also, instead of repeating the nested "if ... simplify..." for each case, just set two variables, leftExprs and rightExprs
to one of exprs or newExprs and break the "if ... simplify..." out to a common code path that uses leftExprs and rightExprs.

tests/frontend/org/voltdb/regressionsuites/TestJoinsSuite.java
```diff
@@ -1353,7 +1357,74 @@ private void subtestNonEqualityFullJoin(Client client)
                 {1,MINVAL},
                 {10, 5}
         });
-}
+    }
+
+    private void subtestMultiplFullJoins(Client client)
```
paulmartel: typo Multipl ==> Multiple

src/ee/executors/executorutil.h
```diff
@@ -60,60 +61,31 @@ struct CountingPostfilter {
...
-        return m_under_limit;
+        m_under_limit = false;
     }
 
-    // Returns true if predicate evaluates to true and LIMIT/OFFSET conditions are satisfied.
-    bool eval(const TableTuple* tuple);
-
-    private:
-    const TempTable* m_table;
+    const Table *m_table;
```
paulmartel: I don't see why this needs to be Table instead of TempTable.

src/ee/executors/executorutil.h
```diff
@@ -60,60 +61,31 @@ struct CountingPostfilter {
...
     const AbstractExpression *m_postfilter;
     CountingPostfilter* m_parentPostfilter;
+
```
paulmartel: On revisiting this, it is a little confusing to have two members with postfilter in their names when only one is of the CountingPostFilter type.

src/ee/executors/executorutil.h
```diff
@@ -60,60 +61,31 @@ struct CountingPostfilter {
     static const int NO_OFFSET = -1;
     static const int LIMIT = NO_LIMIT + 1;
 
-    CountingPostfilter(const AbstractExpression * wherePredicate, int limit, int offset);
+    CountingPostfilter();
```
paulmartel: Add a comment: A CountingPostfilter is not fully initialized by its default constructor. It should be re-initialized before use via assignment from a properly initialized CountingPostFilter.

src/frontend/org/voltdb/planner/PlanAssembler.java
alexeevm: Indeed. Refactored

tests/frontend/org/voltdb/regressionsuites/TestJoinsSuite.java
alexeevm: Corrected

src/ee/executors/executorutil.h
alexeevm: I renamed the m_postfilter to be m_postPredicate

src/ee/executors/executorutil.h
alexeevm: Added

src/ee/executors/executorutil.h
alexeevm: Not sure, what was the reason why I changed that. The output table is always a TempTable in this context.

alexeevm:
Paul,

I just committed your code review suggestions.

Mike

PR #3331, opened 2016-02-25
Add geospatial index support to nest loop index join and index scan

wolffcm:
Also tweaked some tests and added a geospatial index to fullDDL.sql.


hhakimi53:
Looks good to me.  As part of the changes, we can enable "CREATE INDEX INDEX_REGION ON INDEXED_BORDERS(Region)" in  tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java L#148. With this, end-to-end test geo index will also be functional with geo indexes.

src/ee/executors/indexscanexecutor.cpp
```diff
@@ -145,6 +145,32 @@ bool IndexScanExecutor::p_init(AbstractPlanNode *abstractNode,
     return true;
 }
 
+// There is an identical function in nestloopindexexecutor.cpp.  These
+// two functions should be merged into a common place.  But it's
+// unclear where that would be?
```
paulmartel: I think this is part of a broader question of how to refactor NLIJ and indexcan to reduce the current duplication.
We need a home for the common code declarations and definitions. Maybe indexscanexecutor could hold them on behalf of both callers, or they could be added to executorutils which Mike A is now repurposing for general executor support functions, or we could create a new module.

src/ee/executors/indexscanexecutor.cpp
```diff
@@ -145,6 +145,32 @@ bool IndexScanExecutor::p_init(AbstractPlanNode *abstractNode,
...
+                                   int activeNumOfSearchKeys) {
+    if (lookupType == INDEX_LOOKUP_TYPE_EQ
+        || lookupType == INDEX_LOOKUP_TYPE_GEO_CONTAINS) {
+        *tuple = index->nextValueAtKey(*cursor);
+        if (! tuple->isNullTuple()) {
+            return true;
+        }
+    }
+
+    if ((lookupType != INDEX_LOOKUP_TYPE_EQ
```
paulmartel: Suddenly this is looking messed up. It looks like we are supporting a case where the first block sets *tuple to a null tuple and we respond by trying to set it again with a different index call.
That doesn't sound right. I think the "saving grace" here may be that activeNumOfSearchKeys is never 0 when 
(lookupType == INDEX_LOOKUP_TYPE_EQ  || lookupType == INDEX_LOOKUP_TYPE_GEO_CONTAINS),
so the second block should just be an "else" block.
That would seem to make sense -- If activeNumOfSearchKeys == 0, what could we have been using as a "key"?

WRT wider-ranging performance concerns you raised in SU:
As the tuple initialization step becomes more complex, an idea worth considering is making IndexScanExecutor a template class parameterized by something like index type and possibly a flag indicating full unique key equality.
The class would mostly be generic -- maybe it would have a generic non-template base class.
But p_exec would be based on explicitly instantiated specialized implementations of a getAnotherTuple template function. This moves the choice of index operation outside the tight loop and into an executor construction (instantiation) code path when loading the plan. NLIJ could use the same approach -- and possibly reuse the specialized methods.

I think that Xin did not like the complexity of the template solution, but I think that the complexity is already upon us and are driving something like the template solution. 

The way I'd proceed is to get this implementation (optionally with my first concern about the else block addressed) merged ASAP into master and pay attention to any benchmark regressions. Then, either to address performance regressions or as a later performance improvement ticket, pick an initial case to optimize via its own template instantiation and have all other cases use a generic instantiation based on the current "known working" generic approach. If this meets with success, then proceed to write specializations for the remaining cases.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialIndexes.java
```diff
@@ -145,7 +146,7 @@ static private void setupGeoSchema(VoltProjectBuilder project) throws IOExceptio
                 + "  id         INTEGER NOT NULL,\n"
                 + "  region     GEOGRAPHY\n"
                 + ");\n"
-                //+ "CREATE INDEX INDEX_REGION ON INDEXED_BORDERS(Region)\n;"    // Enable this once Geo Indexes are integrated
+                + "CREATE INDEX INDEX_REGION ON INDEXED_BORDERS(Region)\n;"    // Enable this once Geo Indexes are integrated
```
paulmartel: Remove the comment. Today is the day.

paulmartel:
end of round.

src/ee/executors/indexscanexecutor.cpp
wolffcm: I agree with your assessment of the else block.  I'm reluctant to change such a fundamental piece of code soon before a release.  But it also seems like the simplification might make the loop more efficient, so I'm torn.

I've pushed a separate branch with the logic simplification.  If it doesn't seem to have broken anything I may merge it.

Thanks Paul and Huzefa for the quick reviews.

src/ee/executors/indexscanexecutor.cpp
wolffcm: I like your first suggestion.  I'll merge them into indexscanexecutor as a public static method.

wolffcm:
Squashed and merged on separate branch.

PR #3332, opened 2016-02-25
Eng 9844 sqlcov less invalid

johnTcrawford:
Reduced the number of invalid queries; added tests of DWithin; and various other minor improvements

### DIFF
--- a/tests/scripts/examples/sql_coverage/geo-basic.sql
+++ b/tests/scripts/examples/sql_coverage/geo-basic.sql
@@ -2,16 +2,50 @@
 {@cmp = "_eqne"} -- geo types don't do <, >, <=, >= (at least, not in a way that agrees with PostGIS)
 {@star = "ID, LONGITUDE(PT1), LATITUDE(PT1), LONGITUDE(PT2), LATITUDE(PT2), AREA(POLY1)"}
 
--- Define clauses used to ignore points or polygons that demonstrate known differences
--- between VoltDB and PostGIS, which would otherwise cause tests to fail:
-{@ignorepointeqdiffs  = "(ID < 350 OR ID > 360)"} -- ignore known differences with point equality, between VoltDB and PostGIS
-{@ignore2pointeqdiffs = "((A.ID < 308 OR A.ID > 400) AND (B.ID < 308 OR B.ID > 400))"} -- ignore point equality diffs, with a join
+-- Define clauses used to ignore points or polygons that demonstrate known differences between
+-- VoltDB and PostGIS, which would otherwise cause tests to fail. The ID numbers correspond
+-- specifically to values in the hard-coded test data defined below, that were shown to uncover
+-- incompatibilities with PostGIS. These incompatibilities are currently deemed acceptable and
+-- mostly uninteresting.
+-- First, PostGIS, oddly, considers different, but equivalent, representations of the North or
+-- South pole (e.g. 'POINT(0 90)' and 'POINT(90 90)' to be both equal and unequal (see Comments
+-- in ENG-9804; and also ENG-9991, which is a minor VoltDB bug); adding these clauses ignores
+-- certain points at the North and South poles:
+{@ignorepointeqdiffs  = "(ID < 350 OR ID > 360)"} -- ignore known differences with point equality/inequality, between VoltDB and PostGIS
+-- Same as above, for two tables with a join (and for a few more points); plus, PostGIS considers
+-- the points defined below with ID 307 and 308, which are extremely close to each other, to be
+-- equal, but VoltDB does not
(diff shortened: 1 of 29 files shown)


tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -4,108 +4,125 @@
 
 -- Define clauses used to ignore points or polygons that demonstrate known differences
 -- between VoltDB and PostGIS, which would otherwise cause tests to fail:
-{@ignorepointeqdiffs  = "(ID < 250 OR ID > 300)"} -- ignore known differences with point equality, between VoltDB and PostGIS
-{@ignore2pointeqdiffs = "((A.ID < 208 OR A.ID > 300) AND (B.ID < 208 OR B.ID > 300))"} -- ignore point equality diffs, with a join
-{@ignoredistroundoff  = "A.ID <> 453"} -- ignore tiny DISTANCE round-off errors, between VoltDB and PostGIS
+{@ignorepointeqdiffs  = "(ID < 350 OR ID > 360)"} -- ignore known differences with point equality, between VoltDB and PostGIS
```
paulmartel: These magic numbers merit stronger comment support.
I'm guessing that it should be something like "These numbers correspond specifically to values in the hard-coded test data that were shown to uncover incompatibilities with PostGIS. These incompatibilities are currently deemed acceptable and uninteresting". Are there jira ticket(s) -- even a "low priority so don't plan to fix for a long long time" ticket that can be referenced here describing the known issue(s)?

tests/scripts/examples/sql_coverage/geo-basic.sql
```diff
@@ -250,41 +288,44 @@ SELECT A.ID AID, B.ID BID, DISTANCE(A.POLY1,B.POLY1) G29DIST FROM _table A JOIN
...
+SELECT ID, LONGITUDE(PT2), LATITUDE(PT2) from _table G93 WHERE _maybe CONTAINS(POLY1,PT2) AND @ignorecentroiddiffs
+SELECT ID, AsText(PT2),    AsText(POLY1) from _table G94 WHERE _maybe CONTAINS(POLY1,PT2) AND @ignorecentroiddiffs
+SELECT ID, LONGITUDE(PT2), LATITUDE(PT2) from _table G95 WHERE _maybe CONTAINS(POLY2,PT2) AND @ignorecentroiddiffs
+SELECT ID, AsText(PT2),    AsText(POLY2) from _table G96 WHERE _maybe CONTAINS(POLY2,PT2) AND @ignorecentroiddiffs
 
 -- These won't work until CONTAINS (& boolean return values) is supported in the initial SELECT clause
-SELECT ID, CONTAINS(POLY1,PT1), LONGITUDE(PT1), LATITUDE(PT1) from _table G67
-SELECT ID, CONTAINS(POLY1,PT1), AsText(PT1),    AsText(POLY1) from _table G68
+SELECT ID, CONTAINS(POLY1,PT1), LONGITUDE(PT1), LATITUDE(PT1) from _table G97
+SELECT ID, CONTAINS(POLY1,PT1), AsText(PT1),    AsText(POLY1) from _table G98
```
paulmartel: Are we also testing the positive/workaround syntax like : SELECT ID, CASE CONTAINS... THEN 'TRUE' ELSE 'FALSE' END ...
I understand that this question is not specific to the current change set, but I just happened to notice these lines.

paulmartel:
Looks very good, ok to merge.

tests/scripts/examples/sql_coverage/grammar.sql
```diff
@@ -34,11 +34,13 @@
 {_point2numfun |= "LATITUDE"}
 {_point2numfun |= "LONGITUDE"}
 {_poly2numfun |= "AREA"}
+{_poly2numfun |= "NumPoints"}
+{_poly2numfun |= "NumInteriorRings"}
 {_2geo2numfun |= "DISTANCE"}
```
hhakimi53: Do we need DWithin function defined in here too?

hhakimi53:
I don't understand a lot of context in here and as Paul has already reviewed it, I am also good with the changes.

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: I will add more extensive comments here.

tests/scripts/examples/sql_coverage/geo-basic.sql
johnTcrawford: Good idea: I will add a test like that, and also similar ones for IsValid and DWithin.

tests/scripts/examples/sql_coverage/grammar.sql
johnTcrawford: Good idea, I'll add that; it will need to be in a new category (something like _2geonum2boolfun), which I'll also need to add to configure-for-geo-point.sql and configure-for-geo-polygon.sql.

PR #3333, opened 2016-02-25
Vdm 45 and 60

manjujames:


lib/python/vdm/server/HTTPListener.py
```diff
@@ -2274,7 +2270,20 @@ def get(database_id, server_id):
             else:
 
                 try:
-                    client = voltdbclient.FastSerializer(str(server[0]['hostname']), 21212)
+                    if not server[0]['client-listener']:
```
akhanzode: should check for length as well right?

manjujames: not actually checks that.

manjujames:
squash merged.

lib/python/vdm/server/HTTPListener.py
```diff
@@ -2274,7 +2270,20 @@ def get(database_id, server_id):
...
+                        client_port = 21212
+                        client_host = str(server[0]['hostname'])
+                    else:
+                        client_listener = server[0]['client-listener']
+                        if ":" in client_listener:
```
pdshaw: urlparse python module will do this. not a big deal, just fyi

PR #3334, opened 2016-02-25
ENG-9625 add additional stats to voltdb collect

dweiss88:


manjujames:
Are there existing automated tests for this? If so, we should add to it for these new outputs.
If not, these need to be tested on mac and linux to test the behavior on both platforms.

dweiss88:
According to Peter Z it's fine if the commands fail (some of these commands require the installation of packages). We just want the results if the commands succeed with the idea of the more information the better. Failing commands wont cause collect to fail, the failure message will be written to the corresponding file.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -305,8 +307,33 @@ private static void parseJSONObject(JSONObject jsonObject) {
                 }
             }
```
pzhao-voltdb: copy messages from /var/log
ls /var/log/messages*

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -423,11 +450,91 @@ private static void generateCollection(Set<String> paths, boolean copyToVEM) {
...
+
+            String[] sysctlCmd = {"bash", "-c", "sysctl -a"};
+            cmd(zipStream, sysctlCmd, folderPath + "system_logs" + File.separator, "sysctldata");
+
+            String[] ulimitCmd = {"bash", "-c", "ulimit -a"};
+            cmd(zipStream, ulimitCmd, folderPath + "system_logs" + File.separator, "ulimitdata");
+
+            String[] virtwhatCmd = {"bash", "-c", "sudo virt-what"};
+            cmd(zipStream, virtwhatCmd, folderPath + "system_logs" + File.separator, "virtwhatdata");
+
```
pzhao-voltdb: get messages* data from /var/log
#list of all files
ls /var/log/messages*

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -305,8 +307,33 @@ private static void parseJSONObject(JSONObject jsonObject) {
                 }
             }
 
-            collectionFilesList.add("sardata (result of executing \"sar -A\" if sar enabled)");
+            collectionFilesList.add("sardata (result of executing \"LC_ALL=C; sar -A\" if sar enabled)");
```
akhanzode: Since this is in java and many CLIs have diff options maybe creating a mapping file would help rather than hardcode these commands. this will make tool flexible for diff os support.

tests/frontend/org/voltdb/utils/TestCollector.java
```diff
@@ -110,7 +110,8 @@ private ZipFile collect(String voltDbRootPath, boolean skipHeapDump, int days) t
                                     "--copyToVEM=true",
                                     "--calledFromVEM=true",  // calledFromVem (set to true so that resulting collection can be easily located)
                                     "--fileInfoOnly=false",  // fileInfoOnly
-                                    "--days="+String.valueOf(days)
+                                    "--days="+String.valueOf(days),
+                                    "--libPathForTest="+getWorkingDir(voltDbRootPath)+"/lib"
                                     });
 
         rootDir = CoreUtils.getHostnameOrAddress() + "_voltlogs_";
```
manjujames: Will you be able to verify here that files for the entries in the properties file did get created?

tests/frontend/org/voltdb/utils/TestCollector.java
```diff
@@ -283,15 +287,25 @@ public void testBasicFilesAndCrash() throws Exception {
...
+        else
+            systemStatsIS = new FileInputStream(getWorkingDir(voltDbRootPath)+"/lib/linuxstats.properties");
+        assertNotNull(systemStatsIS);
+        Properties systemStats = new Properties();
+        systemStats.load(systemStatsIS);
+        for (String fileName : systemStats.stringPropertyNames()) {
+            ZipEntry statdata = collectionZip.getEntry(subFolderPath + "system_logs" + File.separator + fileName);
+            assertNotNull(statdata);
+        }
+
```
manjujames: Thanks. This looks fine.

PR #3337, opened 2016-02-26
Add geospatial support to nest loop index join and index scan

wolffcm:


wolffcm:
Reviewed on separate branch.

PR #3342, opened 2016-02-29
ENG-9949: DRBufferSplitter for strict SP case

slowsoul:
1. Adjusted the getTestDRBuffer() for single hash case to generate txns with different hashes so that splitter can be tested
2. Addressed remaining review comments from last pull request.


nshi:
Reviewed

PR #3343, opened 2016-02-29
Update geospatial example to use index

wolffcm:
I added a CREATE INDEX to ddl.sql so that the CONTAINS query can run much faster.

I generated a set of 100 polygons for use in the demo, and replaced the dynamically generated bounding boxes with a class that can pick one of these 100 polygons at random for use in the demo.

I also updated the README to point out the benefit of using a geospatial index.


hhakimi53:
Changes looks good to me.

BillWhite:
This seems straightforward to me.

PR #3344, opened 2016-03-01
Update geospatial example to use index

wolffcm:


wolffcm:
Reviewed on merge to master.

PR #3346, opened 2016-03-01
ENG-9990 Repeating GROUP BY partition column in SELECT

paulmartel:


src/frontend/org/voltdb/plannodes/AbstractReceivePlanNode.java
```diff
@@ -91,7 +91,7 @@ protected void resolveColumnIndexes(NodeSchema outputSchema)
             int index = tve.resolveColumnIndexesUsingSchema(input_schema);
             tve.setColumnIndex(index);
         }
-        outputSchema.sortByTveIndex();
+        // output schema for ReceivePlanNode should never be re-sorted
```
wolffcm: I'd like to understand what broke here.  If the corresponding SendPlanNode had its output schema sorted, then would it be okay to do it here, too?

wolffcm:
End of review round, just had a question.

src/frontend/org/voltdb/plannodes/AbstractReceivePlanNode.java
wolffcm: Also, I don't see why this bug wasn't found much earlier.  Does it have to do with the fact that group by keys are almost always first on the select list?

paulmartel: @wolffcm We had a bug for some time where the SendPlanNode was getting its outputschema re-sorted, but since it was often being re-sorted based on the same criteria that drove how the outputschema was laid out in the first place, it only mattered in certain edge cases. Since that got fixed, the ReceivePlanNode parents of SendPlanNodes were mostly unaffected by their continued (pointless) re-sorting of their outputschema. They were laid out as pass-through outputschemas, consisting of TVEs with ascending column indexes and references to their underlying expressions in the SendPlanNode's outputschema. The re-sorting of the RecievePlanNode's outputschema usually validated that the original referent expressions were still at their original column indexes in the SendPlanNode, and left everything as is. The edge case where this did not work was when the SendPlanNode contained two identical output columns. This case breaks down under the usually reliable process of putting all the SendPlanNode outputschema expressions into a hashmap mapped to their "new" column indexes and looking up each ReceivePlanNode expression in that hashmap to find it. The hash table entry for the duplicate expression overwrites the entry for its first occurrence. The ReceivePlanNode outputschema entry gets a new (greater) column index, and gets shifted in the ReceivePlanNode outputschema to be listed beside its duplicate.
The parent coordinator SendNode remains stable rather than adopt or adapt to this changed ordering, and so there is a mismatch. The generateOutputSchema/resolveIndexes logic has been a long-standing over-complicated nesting ground for bugs.

PR #3352, opened 2016-03-02
Eng 9849 handle initialization

wweiss-voltdb:


src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -404,11 +405,37 @@ public void forceAllDRNodeBuffersToDisk(final boolean nofsync)
...
+            }
+            else {
+                DRConsumerDrIdTracker targetTracker = clusterSources.get(producerPartitionId);
+                if (targetTracker == null) {
+                    if (lastReceivedDRId == -1L) {
+                        return true;
+                    }
+                }
+                else {
+                    if (targetTracker.getDrIdRanges().lastEntry().getValue() == lastReceivedDRId) {
```
nshi: I wonder if it's possible that all DR ID ranges in the tracker are truncated, so there is only a truncation point in there with no entry.

nshi:
Reviewed

PR #3356, opened 2016-03-03
Eng 9220 Add geo support for the C++ client.

BillWhite:
The C++ client needs some messages in binary files.  This commit adds the Java file which generates these messages to the VoltDB test source tree.  It also adds a small application in tests/scripts/generate_cpp_test_data which allows generation of the test data through a single test script.  Typically this data goes into $CPP_CLIENT_SRC/test_src/test_data, where $CPP_CLIENT_SRC is the path to the C++ client source folder.

This file and the tests in the C++ client source folder test_src should be synchronized.  If either changes the other should change in tandem.


.gitignore
```diff
@@ -1,6 +1,7 @@
 /.classpath
 /.depcache
 /.project
+/.cproject
```
wolffcm: What generates cproject files?  Just curious.

BillWhite: Eclipse generates them when you create a project.  The existing .cproject file doesn't work for me, and I had to regenerate it.  It seems difficult to be saving them for this little bit of code.

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
```diff
@@ -0,0 +1,630 @@
...
+        oclientThread.start();
+        sc = ssc.accept();
+        sc.socket().setTcpNoDelay(true);
+        ByteBuffer authReqSHA1 = ByteBuffer.allocate(DEFAULT_BUFFER_SIZE);
+        sc.configureBlocking(true);
+        readMessage(authReqSHA1, sc);
+        writeDataFile(clientDataDir, "authentication_request.msg", authReqSHA1);
+        writeServerAuthenticationResponse(sc, true);
+        //
+        // Read the some call procedure messages.
```
wolffcm: Nit: "read the some..."?

tests/scripts/generate_cpp_test_data/deployment.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<deployment>
```
wolffcm: Is the deployment file necessary?  These seem like all default options?

tests/scripts/generate_cpp_test_data/run.sh
```diff
@@ -0,0 +1,261 @@
...
+    server "$DEPLOYMENT_FILE" "$LICENSE_FILE" \
+        && init_database "$SCHEMA_FILE" \
+        && client \
+        && noserver \
+        && cleanup
+}
+
+while [ -z "$DONE" ]; do
+    case "$1" in
+        --#)
```
wolffcm: What does this option do?  `#` is a shell comment, so you would have to escape it on the command line.  I guess this is just a way to provide an option to debug the script in a way that won't get accidentally triggered?

tests/scripts/generate_cpp_test_data/run.sh
```diff
@@ -0,0 +1,261 @@
...
+
+function cleanup() {
+    if [ "$CLEANUP" == YES ] ; then
+        echo "Cleaning up."
+        rm -rf "$OBJ"
+        rm -rf "${START_DIR}/log"
+    else
+        echo "The server's data and log files are in \"$OBJ\" and perhaps \"${START_DIR}/log\"."
+    fi
+}
```
wolffcm: You could shut down the server here?  Or maybe you don't want to for some reason?

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
```diff
@@ -0,0 +1,630 @@
...
+                    String strings[] = new String[] { "oh", "noes" };
+                    byte bytes[] = new byte[] { 22, 33, 44 };
+                    short shorts[] = new short[] { 22, 33, 44 };
+                    int ints[] = new int[] { 22, 33, 44 };
+                    long longs[] = new long[] { 22, 33, 44 };
+                    double doubles[] = new double[] { 3, 3.1, 3.14, 3.1459 };
+                    TimestampType timestamps[] = new TimestampType[] { new TimestampType(33), new TimestampType(44) };
+                    BigDecimal bds[] = new BigDecimal[] { new BigDecimal( "3" ), new BigDecimal( "3.14" ), new BigDecimal( "3.1459" ) };
+                    try {
+                        newClient.callProcedure("foo", strings, bytes, shorts, ints, longs, doubles, timestamps, bds, null, "ohnoes!", (byte)22, (short)22, 22, (long)22, 3.1459, new TimestampType(33), new BigDecimal("3.1459"));
```
wolffcm: You're passing all arrays to the stored procedure?  Is this to test array parameters like we support for the `IN` operator?

Should we be testing the scalar versions of these parameters as well?  Also, would there be value in testing null values for each type?

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
BillWhite: Ah, yes.

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
```diff
@@ -0,0 +1,630 @@
...
+            }
+        }
+
+        oclient.close();
+        ssc.close();
+        oclientThread.join();
+
+        // Now, connect to a real server.  We are going to pretend to be a
+        // client and write the messages we just read from the client, as we pretended to be
+        // a server.  We will then capture the responses in files.
```
wolffcm: Your capturing the server responses for use in the C++ client unit test, so that you can fake out a server there, right?

tests/scripts/generate_cpp_test_data/run.sh
BillWhite: I always do that it always just adds -x so that I can debug the script.  It's not in the help, as it's only useful for debugging the shell script.

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
```diff
@@ -0,0 +1,630 @@
...
+import org.voltdb.client.ClientAuthScheme;
+import org.voltdb.client.ClientConfig;
+import org.voltdb.client.ClientFactory;
+import org.voltdb.client.ClientStatusListenerExt;
+import org.voltdb.client.NullCallback;
+import org.voltdb.messaging.FastSerializer;
+import org.voltdb.types.GeographyPointValue;
+import org.voltdb.types.GeographyValue;
+import org.voltdb.types.TimestampType;
+
```
wolffcm: Here's my high-level understanding of what this code does, just from having read it.

To capture what the client messages should look like, you instantiate a Client that connects to a fake server, and you capture the messages.

To capture what server messages should look like, this class acts like a "man-in-the-middle" and passes messages from a client along to a real server and then captures the response.

Do I have this right?

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
BillWhite: Yes.  The test first captures the client requests, both authentication and invocation.  It then connects to a real server to which it sends the captured messages.  It captures the responses.  Everything gets written to files so that the C++ client's mock VoltDB server can replay the responses.

wolffcm:
This looks good to me.  Obviously we sorely needed something like this to stave off bitrot.  Thanks for taking this on.

It seems like there would be value in running this in Jenkins somehow, perhaps comparing the generated messages files to a set of checked-in gold files.  If the protocol changes, a failing test would tell us to regenerate the gold files and re-run the C++ client unit tests.

tests/frontend/org/voltdb/GenerateCPPTestFiles.java
BillWhite: This is slightly different.  We read the client request, but we don't pass this to the server.  We just save the response.

That's the way it was, and I didn't extend it at all.

tests/scripts/generate_cpp_test_data/deployment.xml
BillWhite: No, it's not necessary.  I started out needing it, because I thought I had to add users to use SHA256 authentication.  But it turns out I didn't.  If we ran a server with authentication and really tested out the messages for authentication we would need a deployment file and also a special sql file.  But for now the deployment file is extra.   I don't think it really hurts anything to have it, though.

tests/scripts/generate_cpp_test_data/run.sh
BillWhite: The server gets shut down in the shell function noserver, defined above.

BillWhite:
I think the client test should be in Jenkins as well.  I don't necessarily think Jenkins should do
anything but get the client repository, build the release (default) build and then run the unit test.
This should be pretty fast.  If we did this for every Jenkins job I don't think it would increase the
testing time very much and it would keep things from degrading.

Recreating the gold files should probably be done by a developer when the protocol changes
somehow.  The Jenkins job should just pass or fail and have done with it.

BillWhite:
Closing to squash this.

PR #3359, opened 2016-03-03
ENG-9560: AAUIW a higher limit on maximum AdHoc query length

JadenW:
Seems only need to change serialize/deserialize part of AdHocPlannedStatement.
No need to change ParameterSet, not affect wire protocol.


akhanzode:
If jenkins is happy merge it.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -507,11 +502,19 @@ void VoltDBEngine::resetExecutionMetadata() {
 
     // set this back to -1 for error handling
     m_currentInputDepId = -1;
+
     if (m_currExecutorVec == NULL) {
         // This is usually the result of some planner error producing an
         // invalid plan that can not be converted into plan nodes / executors.
         return;
     }
+
```
akhanzode: Why this change is needed?

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -79,6 +79,7 @@ public synchronized static AdHocCompilerCache getCacheForCatalogHash(byte[] cata
 
     // cache sizes determined at construction time
     final int MAX_LITERAL_ENTRIES;
+    final long MAX_LITERAL_MEM;
```
akhanzode: Can you make this into a hidden tunable via system property. Do you need the number of entries now that its bounded by memory?

src/ee/execution/VoltDBEngine.cpp
JadenW: m_executorContext could be invalid if planner hit some error. 
So It should first check m_currExecutorVec == NULL before check m_executorContext

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
JadenW: I now made this cache bound by both memory and number of entries. 
We still don't want too many cache entries for small size queries, do we? 
Shoud I drop the entries limit?

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -147,9 +138,73 @@ public boolean removeEldestEntry(Map.Entry<String, List<BoundPlan> > eldest) {
...
+    // define a LinkedHashMap based LRU cache bounds by both entry number and entry value on-heap size
+    // without changing Map.Entry, only works for value of type AdHocPlannedStatement
+    // only extend put, remove,clear and removeEldestEntry methods to account weight
+    public class MemoryBoundedLinkedHashMap<K, AdHocPlannedStatement> extends LinkedHashMap<K, AdHocPlannedStatement> implements Map<K, AdHocPlannedStatement>{
+       private static final long serialVersionUID = 2988383448026641836L;
+        private final int maxEntries;
+        private final long maxMemory; // in bytes
+        private long currentMemory;   // in bytes
+
+        public long getCurrentMemory() {
```
akhanzode: Do you need this method? seems like unused.

src/frontend/org/voltdb/ParameterSet.java
```diff
@@ -255,6 +255,9 @@ public static ParameterSet fromByteBuffer(ByteBuffer buffer) throws IOException
         int startPos = buffer.position();
 
         short count = buffer.getShort();
+        if (count < 0) {
+            throw new RuntimeException("ParameterSet exceeds the length limitation " + Short.MAX_VALUE);
```
akhanzode: The error message is wrong it should be invalid parameter length. Also add count in message.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -147,9 +138,73 @@ public boolean removeEldestEntry(Map.Entry<String, List<BoundPlan> > eldest) {
...
                 return false;
             }
+
         };
     }
 
+    // define a LinkedHashMap based LRU cache bounds by both entry number and entry value on-heap size
+    // without changing Map.Entry, only works for value of type AdHocPlannedStatement
+    // only extend put, remove,clear and removeEldestEntry methods to account weight
+    public class MemoryBoundedLinkedHashMap<K, AdHocPlannedStatement> extends LinkedHashMap<K, AdHocPlannedStatement> implements Map<K, AdHocPlannedStatement>{
```
akhanzode: Can you move this class in its own file so if anyone wants to use it they can. Also you can write unit test for just this memory bounded LHM.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -147,9 +138,73 @@ public boolean removeEldestEntry(Map.Entry<String, List<BoundPlan> > eldest) {
...
+            // default max value size of 32MB
+            this(1000, 32 * 1024 * 1024);
+        }
+
+        public MemoryBoundedLinkedHashMap(final int maxEntries) {
+            this(maxEntries, 32 * 1024 * 1024);
+        }
+
+        public MemoryBoundedLinkedHashMap(final int maxEntries, final long maxMemory) {
+            super(maxEntries * 2, .75f, true);
```
akhanzode: comment here that true is for access order.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
akhanzode: No thats good. Remember the cache is still not thread safe.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
JadenW: This class is current limited for general use. For accounting memory, it now use gerSerizedSize, so it can now only works for the value of type AdHocPlannedStatement.
And it now only supported put,remove and clear methods. Other operations liked putall and iterator cannot be easily extended, (and it's not needed if used here). Should I try to extend these methods?

akhanzode: No lets stick to inner class.

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
JadenW: MAX_LITERAL_MEM changed to use system property

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
JadenW: removed

src/frontend/org/voltdb/ParameterSet.java
JadenW: fixed

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
JadenW: fixed.

src/frontend/org/voltdb/ParameterSet.java
```diff
@@ -255,6 +255,9 @@ public static ParameterSet fromByteBuffer(ByteBuffer buffer) throws IOException
         int startPos = buffer.position();
 
         short count = buffer.getShort();
+        if (count < 0) {
+            throw new RuntimeException("Invalid parameter length " + count + " for ParameterSet." );
```
vtkstef: use IllegalArgumentException

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -108,31 +110,18 @@ private AdHocCompilerCache() {
...
-            @Override
-            public boolean removeEldestEntry(Map.Entry<String, AdHocPlannedStatement> eldest) {
-                if (size() > MAX_LITERAL_ENTRIES) {
-                    ++m_literalEvictions;
-                    return true;
-                }
-                return false;
-            }
-        };
+        m_literalCache = new MemoryBoundedLinkedHashMap<String, AdHocPlannedStatement>(MAX_LITERAL_ENTRIES, MAX_LITERAL_MEM);
```
vtkstef: Don't template the MemoryBoundedLinkedHashMap call it AdHocStatementCache which extends LinkedHashMap<String, AdHocPlannedStatement>

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -147,9 +136,70 @@ public boolean removeEldestEntry(Map.Entry<String, List<BoundPlan> > eldest) {
...
+                ++m_literalEvictions;
+                this.currentMemory -= ((org.voltdb.compiler.AdHocPlannedStatement) eldest.getValue()).getSerializedSize();
+                return true;
+            }
+            return false;
+        }
+
+        @Override
+        public AdHocPlannedStatement put(K key, AdHocPlannedStatement value) {
+            this.currentMemory += ((org.voltdb.compiler.AdHocPlannedStatement) value).getSerializedSize();
```
vtkstef: why do you need to cast it?

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
```diff
@@ -147,9 +136,70 @@ public boolean removeEldestEntry(Map.Entry<String, List<BoundPlan> > eldest) {
...
+            this.maxMemory = maxMemory;
+            this.currentMemory = 0;
+        }
+
+        // This method is called just after a new entry has been added
+        @Override
+        public boolean removeEldestEntry(final Map.Entry<K, AdHocPlannedStatement> eldest) {
+            if ((size() > maxEntries) || (this.currentMemory > this.maxMemory))  {
+                ++m_literalEvictions;
+                this.currentMemory -= ((org.voltdb.compiler.AdHocPlannedStatement) eldest.getValue()).getSerializedSize();
```
vtkstef: why do you need to cast here?

src/frontend/org/voltdb/ParameterSet.java
JadenW: fixed

src/frontend/org/voltdb/compiler/AdHocCompilerCache.java
JadenW: removed the unnecessary cast due to previous template

PR #3360, opened 2016-03-03
ENG-9869 fix gaps in column reference resolution

paulmartel:


paulmartel:
Simplified the ORDER BY alias resolution process.
Simplified the general column resolution process.
Added validation of ORDER BY columns against GROUP BY columns when they are not in the SELECT list

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -320,378 +306,191 @@ RangeVariable getRangeVariable() {
...
+        // resolve it below.
+        Map<String, ExpressionColumn> usingResolutions = new TreeMap<>();
+        Map<String, RangeVariableColumnReferenceResolution>
+            rangeVariableResolutions = new TreeMap<>();
+        for (int i = 0; i < rangeCount; i++) {
+            RangeVariable rangeVar = rangeVarArray[i];
+            if (rangeVar == null) {
+                continue;
+            }
+            resolveColumnReference(rangeVar, usingResolutions, rangeVariableResolutions);
```
BillWhite: I like adding this function.  It makes the loop easier to see.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -320,378 +306,191 @@ RangeVariable getRangeVariable() {
...
+            sep = ", ";
         }
     }
 
-    private class RangeVariableColumnReferenceResolution implements ColumnReferenceResolution {
-        RangeVariable m_rangeVariable;
-        int           m_colIndex;
-        private final static String m_unknownTableName = "UnknownTable";
-        public RangeVariableColumnReferenceResolution(RangeVariable rangeVariable, int colIndex) {
+    private static class RangeVariableColumnReferenceResolution {
```
BillWhite: What is this class doing for you now?  It used to hold finallyResolve.  But since you've move that out into an overloaded function the m_colIndex and m_opType member variables are not used at all, and there is only a range variable here.  Can't you just keep the range variable?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -320,378 +306,191 @@ RangeVariable getRangeVariable() {
...
+            rangeVariableResolutions = new TreeMap<>();
+        for (int i = 0; i < rangeCount; i++) {
+            RangeVariable rangeVar = rangeVarArray[i];
+            if (rangeVar == null) {
+                continue;
+            }
+            resolveColumnReference(rangeVar, usingResolutions, rangeVariableResolutions);
+        }
+        if (usingResolutions.isEmpty()) {
+            if (rangeVariableResolutions.isEmpty()) {
```
BillWhite: I make ColumnReferenceResolution to avoid this complicated construction.
  if (a.isEmpty()) {
    if (b.isEmpty()) {
     do both are empty;
    } else {
      do a empty and b is not empty
    }
  } else if (b.isEmpty()) {
    do a is not empty and b is empty;
  } else {
    do neither a nor b are empty
  }

I generally dislike this particular construction.  I find it leads to errors.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1358,6 +1021,16 @@ Expression replaceAliasInOrderBy(Expression[] columns, int length) {
         return this;
     }
 
+    private static void addColumnIfUnique(List<ExpressionColumn> foundColumns,
+            ExpressionColumn ecol) {
```
BillWhite: This and isSameColumn seem much simpler.  They are only called once, so they are logically unnecessary, but it seems easier to read.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
paulmartel: The m_colIndex and m_opType are used in finallyResolve along with the range variable, and that is why a class was needed.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -926,6 +926,15 @@ private void resolveGroups() {
                 }
             }
 
+            for (int i = indexStartOrderBy; i < indexStartAggregates; i++) {
+                if (!exprColumns[i].isComposedOf(
```
BillWhite: Is this loop the ultimate fix to the problem?  I just want to understand.  I think it's right, and it was an oversight.

BillWhite:
That's all I have for this.  I have verified that the original problem is solved.

If HSQL had a real symbol table with real scopes this would never have happened.  You'd have one scope for regions where the variables need to be functionally dependent on the group by columns and another for regions where the variables need to be functionally dependent on table expression columns.  The scopes would know when they nest and when then don't, so there would just be "lookup this name in the current scope".

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
paulmartel: Yes. This turned out to be sufficient cause of the specific error. The other changes stem from problems I ran into experimenting with different variants of the test query that were casing other strange behaviors. I should have retained the reproducers for them as regression cases, but it has been a rough week and they got lost to one too many context switch.

paulmartel:
The critical changes were manually rebased.
Some code cleanup was deferred indefinitely.

PR #3365, opened 2016-03-04
ENG-9949: Fix EE DR tuple stream hash flag.

nshi:
No need to lookup the hash for replicated table changes because it
doesn't have a partitioning column. The decision to lookup the hash or
not should depend on the table type, not the partition ID of the
stream.


nshi:
Merged externally

PR #3366, opened 2016-03-04
ENG-9930 create only allowed when managed paths are empty

dweiss88:


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -411,12 +413,20 @@ public void startUp(boolean clearLocalDataDirectories) {
...
+        templateCmdLine.force(force);
+    }
+
     public void setDeploymentAndVoltDBRoot(String pathToDeployment, String pathToVoltDBRoot) {
         templateCmdLine.pathToDeployment(pathToDeployment);
         m_voltdbroot = pathToVoltDBRoot;
         m_compiled = true;
     }
 
+    public void setFilePrefix(File filePrefix) {
```
manjujames: This is a new method, but I don't see it being called..

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -478,7 +494,54 @@ public void initialize(VoltDB.Configuration config) {
...
+                PathsType paths = m_catalogContext.getDeployment().getPaths();
+                String voltDbRoot;
+                if (paths == null || paths.getVoltdbroot() == null || paths.getVoltdbroot().getPath() == null)
+                    voltDbRoot = "voltdbroot";
+                else
+                    voltDbRoot = paths.getVoltdbroot().getPath();
+
+                File managedPath;
+                if (paths != null && paths.getSnapshots() != null && paths.getSnapshots().getPath() != null &&
+                        (new File(paths.getSnapshots().getPath()).isAbsolute()))
```
manjujames: This won't handle the case when snapshots is specified, but is a relative path.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -478,7 +494,54 @@ public void initialize(VoltDB.Configuration config) {
...
+            if (!config.m_deploymentDefault) {
+                PathsType paths = m_catalogContext.getDeployment().getPaths();
+                String voltDbRoot;
+                if (paths == null || paths.getVoltdbroot() == null || paths.getVoltdbroot().getPath() == null)
+                    voltDbRoot = "voltdbroot";
+                else
+                    voltDbRoot = paths.getVoltdbroot().getPath();
+
+                File managedPath;
+                if (paths != null && paths.getSnapshots() != null && paths.getSnapshots().getPath() != null &&
```
manjujames: The check (paths.getSnapshots().getPath()!=null) is not required because the generated classes always return the default value specified i XSD.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -478,7 +494,54 @@ public void initialize(VoltDB.Configuration config) {
             validateStartAction();
 
             Map<Integer, String> hostGroups = null;
+
             final int numberOfNodes = readDeploymentAndCreateStarterCatalogContext();
+
+            if (!config.m_deploymentDefault) {
+                PathsType paths = m_catalogContext.getDeployment().getPaths();
+                String voltDbRoot;
+                if (paths == null || paths.getVoltdbroot() == null || paths.getVoltdbroot().getPath() == null)
```
manjujames: May be we can check if paths is null in the beginning itself and do the same thing that is done for when no deployment path is specified. That will simplify the if checks below a lot.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -367,6 +377,12 @@ public void initialize(VoltDB.Configuration config) {
             // If there's no deployment provide a default and put it under voltdbroot.
             if (config.m_pathToDeployment == null) {
                 try {
+                    managedPathEmptyCheck(config, new VoltFile("voltdbroot", "snapshots"));
```
manjujames: We can reuse the defaults defined in the XSD. For example, (new PathsType.SnapShots()).getPath() will give you the default snapshots path.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -350,6 +351,79 @@ public String getLicenseInformation() {
...
+                defaultPathsEmptyCheck(config);
+            else {
+                String voltDbRoot;
+                if (paths.getVoltdbroot() == null)
+                    voltDbRoot = (new PathsType.Voltdbroot()).getPath();
+                else
+                    voltDbRoot = paths.getVoltdbroot().getPath();
+
+                VoltFile managedPath;
+                if (paths.getSnapshots() == null)
```
akhanzode: why so many if doing same stuff put them in one function.

akhanzode: You should build paths list which you want in one place default or not and have a function that validates.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -350,6 +351,30 @@ public String getLicenseInformation() {
...
+    private void managedPathEmptyCheck(VoltDB.Configuration config, String voltDbRoot, String path) {
+        VoltFile managedPath;
+        if (new File(path).isAbsolute())
+            managedPath = new VoltFile(path);
+        else
+            managedPath = new VoltFile(voltDbRoot, path);
+        if (managedPath.exists() && managedPath.list().length > 0 && !config.m_forceStartVoltdb &&
+                config.m_isEnterprise && config.m_startAction == StartAction.CREATE) {
+            VoltDB.crashLocalVoltDB("Files from a previous database session exist in " + managedPath.getAbsolutePath() +
+                                    ". Use the recover command to restore the previous database or use create --force " +
```
akhanzode: we should not put --force option in error message otherwise everyone will use it.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -478,7 +503,10 @@ public void initialize(VoltDB.Configuration config) {
             validateStartAction();
 
             Map<Integer, String> hostGroups = null;
+
             final int numberOfNodes = readDeploymentAndCreateStarterCatalogContext();
+            managedPathsEmptyCheck(config);
```
akhanzode: shouldn't this be done only on create looks like this will be done on rejoin too??

PR #3371, opened 2016-03-07
Make run.sh more consistent, cleaner and slightly shorter across examples

jhugg:
Add "voltenv" script to hide boilerplate for examples.
Unify run.sh across examples.
Other small cleanup.

I know many people don't care for run.sh. I don't really either? I've tried some alternatives personally, and haven't found something I like better. I encourage others with less tarnished minds to try as well.

### DIFF
--- a/bin/voltenv
+++ b/bin/voltenv
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+
+#set -o nounset #exit if an unset variable is used
+set -o errexit #exit on any single command fail
+
+# next few lines to get the current script path from stackoverflow
+# http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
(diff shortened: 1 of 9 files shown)


akhanzode:
sweet....merge it.

nshi:
The majority of the Shell scripts under `bin/` share the same code that `voltenv` has now. They should also leverage `voltenv`.

I also wonder if it would simpler to have `voltdb` provide these information, e.g. `voltdb --print-classpath`, instead of having a standalone script to just get the environment.

benjaminballard:
Looks great.  Definitely an improvement!

benjaminballard:
One suggestion, along the lines of how I envision minimalist example app scripting (https://github.com/benjaminballard/app-template) is to separate the compiling of the procedure jar file into a separate script.  Eliminate server() and instruct the user to use the CLI to start the database.  This would remove something like 5% of run.sh.  The other 95% is all about running the client, perhaps that should be the one purpose for the script.

jhugg:
@nshi Good suggestion about cleaning up bin. It's not really my focus at the moment, but I would welcome that change. "voltdb --print-classpath" is interesting. Wonder what @ajgent thinks?

@benjaminballard We already tell people to do as much via the command line as possible. We also ship compiled jars with official kits so you don't have to compile anything to run the examples. We've stopped short of removing the targets because I feel they have some value as examples.

PR #3373, opened 2016-03-08
Eng 9695 validating polygon

BillWhite:
Define polygonfromvalidtext, a validating polygonfromtext.

The function polygonfromvalidtext constructs a polygon and the validates
it using the standard isvalid tests.

https://issues.voltdb.com/browse/ENG-9695

### DIFF
--- a/src/ee/expressions/geofunctions.cpp
+++ b/src/ee/expressions/geofunctions.cpp
@@ -38,6 +38,7 @@ static const double RADIUS_SQ_M = SPHERICAL_EARTH_MEAN_RADIUS_M * SPHERICAL_EART
 
 typedef boost::tokenizer<boost::char_separator<char> > Tokenizer;
 
+static bool isMultiPolygon(const Polygon &poly, std::stringstream *msg);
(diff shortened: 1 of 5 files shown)


src/ee/expressions/geofunctions.cpp
```diff
@@ -347,11 +340,44 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POLYGONFROMTEXT>() const
...
+        }
+    }
     SimpleOutputSerializer output(storage, length);
     poly.saveToBuffer(output);
     return nval;
 }
 
+template<> NValue NValue::callUnary<FUNC_VOLT_POLYGONFROMTEXT>() const
+{
+    if (isNull()) {
```
BillWhite: Note:  NValue::isNull is an NValue member function.  Even though it's replicated here and in FUNC_VOLT_POLYGONFROMVALIDTEXT, and both functions call polygonFromText, this null check can't be moved to polygonFromText.

paulmartel:
Looks good.

PR #3375, opened 2016-03-09
ENG-9869 column resolution fixes -- minimal approach

paulmartel:


paulmartel:
Everything here has been reviewed before in some form -- except that (for) now there's less of it.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -352,9 +352,9 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
                 // the range variables.  If we find just one, we finally
                 // resolve it. below.
                 java.util.Set<ColumnReferenceResolution> usingResolutions
-                    = new java.util.TreeSet<ColumnReferenceResolution>(m_comparator);
+                    = new java.util.TreeSet<>(m_comparator);
```
BillWhite: I'm curious why you don't add the generic type here?  I know it's going to be erased at run time, and I
know the variable usingResolutions really carries the compile-time type.  But is there reason other than
that it's not really needed?  Documentation?  Is it a bug to use it?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -366,20 +366,19 @@ public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
 
                     ColumnReferenceResolution resolution = resolveColumnReference(rangeVar);
                     if (resolution != null) {
-                        if (resolution.isExpression()) {
+                        if (resolution instanceof ExpressionColumnReferenceResolution) {
```
BillWhite: Is instanceof faster than calling a member function in Java?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -547,79 +499,28 @@ public String getName() {
             return m_unknownColumnName;
         }
 
-        @Override
```
BillWhite: These, equals and hashCode, are artifacts of when I was maintaining a hashset.  When I changed it to a treeset, so that error messages are ordered by name, they became unneeded. I never went back and deleted them.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -686,12 +569,10 @@ public ColumnReferenceResolution resolveColumnReference(RangeVariable rangeVar)
         }
 
         int colIndex = rangeVar.findColumn(tableName, columnName);
-
-        if (colIndex != -1) {
-            return makeRangeVariableResolution(rangeVar, colIndex);
+        if (colIndex == -1) {
+            return null;
```
BillWhite: Is != slower than ==?  Why did you make this change?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -1086,24 +965,18 @@ public final ExpressionColumn getExpressionColumn() {
...
-                return m_columnName;
-            } else {
-                return m_tableAlias + "." + m_columnName;
-            }
-        }
-
         @Override
         public String toString() {
-            if (m_alias != null && !m_alias.equals(m_columnName)) {
+            if (!m_alias.equals(m_columnName)) {
```
BillWhite: What if m_alias == null?

BillWhite:
Mostly questions.  In substance all seems right.
That's all I have for this.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
paulmartel: No "other" reason.
"not needed" does the trick for me.
Generally, less code is better than more code.
Specifically, if the type ever needs to be changed, it can be changed in one place.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
paulmartel: I don't see why performance would be a critical actor here.
I'm opting for code size, maintainability, and readability.
Everyone knows what instanceof means -- what isExpression means requires research.
Generally, less code (by three member function definitions so far -- + 1 per any future leaf classes) is better than more code. Yes?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
paulmartel: I thought this syntax was stylistically more consistent with the surrounding code that runs through a series of qualifications that return null so that the non-null result is produced on the "main code path" when the case has survived the gantlet.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
paulmartel: Exactly. It's just more code. Now one less thing to suspect/debug.

PR #3376, opened 2016-03-09
Logic update to voltqueuesql to disallow operation where implicit type conversion is not possible

hhakimi53:
Update logic in voltqueuesql to perform check if the implicit type conversion for the passed-in parameter is feasible. If not feasible, the operation in voltqueuesql fails with error message indicating which argument, its type. Conversion matrix is based on the broadest scenarios where implicit conversion can  happen in EE, so that we disallow only the conversion which are not feasible in EE to avoid regression on working case. Disallowing such operation will help in some of the cases where type conversion is not happening in EE, like passing in string for geo type column, and preventing potential undefined behavior.


src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -837,49 +838,91 @@ public DependencyPair executeSysProcPlanFragment(
         return sysproc.executePlanFragment(dependencies, fragmentId, params, m_systemProcedureContext);
     }
 
-    private final ParameterSet getCleanParams(SQLStmt stmt, Object... inArgs) {
+    private final void typeConversionFeasible(SQLStmt stmt, Class <?> argClass, int argInd, VoltType expectedType) {
```
wolffcm: I would change the name of this method to `throwIfInfeasableTypeConversion` or something like that to better reflect what it does.  The way it's named now I would expect it to return a Boolean.

hhakimi53: sure, will update the name of the function.

tests/testprocs/org/voltdb_testprocs/regressionsuites/failureprocs/ProcToTestTypeConversion.java
```diff
@@ -0,0 +1,232 @@
...
+                                                      + " T.dummy6 = ? OR "
+                                                      + " T.dummy7 = ? OR "
+                                                      + " T.dummy8 = ? OR "
+                                                      + " T.dummy9 = ? "
+                                                      + " order by dummy1;");
+
+    public final static SQLStmt insert = new SQLStmt ("Insert into T "
+                                            + "values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);");
+
+    public final static byte TestHCTypeConvAllowed = 0;
```
wolffcm: What does HC stand for here?

src/frontend/org/voltdb/utils/VoltTypeUtil.java
```diff
@@ -213,6 +213,82 @@ else if ((left.isNumber() && !right.isNumber()) ||
...
 
+
+    // broadest type conversion which can happen in EE
+    // from is the type of parameter that is passed in
+    // to is the type of parameter that it can be converted-to
+    //                                   MATRIX
+    //  from     |             to
+    //--------------------------------------------------------------------------------------------------------
+    //           |Tiny  Small  Int   BigInt  Float  Decimal  Timestamp  Varchar  VarBinary GeoPoint  Geography
+    // Tiny      | y     y      y      y       y      y         y        y          x        x          x
```
wolffcm: Ok, so this is mostly a rectangle on the upper left, except you can convert from VARCHAR to most types, and all types can be converted to themselves?

using `.` instead of `x` might make the table a little easier to read.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -990,6 +993,29 @@ public void testPolygonPointDWithinNegative() throws Exception {
         verifyStmtFails(client, sql, expectedMsg);
     }
 
+    public void testPolygonIncompatibleTypes() throws NoConnectionsException, IOException, ProcCallException {
```
wolffcm: Does the coverage provided by this test overlap with the tests you added in TestTypeConversionsSuite?  If so, consider dropping this test.  Given equal amounts of test coverage, fewer tests is better, since it means less maintenance in the future.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -371,7 +372,7 @@ else if (m_language == Language.GROOVY) {
             else {
                 assert(m_catProc.getStatements().size() == 1);
                 try {
-                    m_cachedSingleStmt.params = getCleanParams(m_cachedSingleStmt.stmt, paramList);
+                    m_cachedSingleStmt.params = getCleanParams(m_cachedSingleStmt.stmt, false, paramList);
```
wolffcm: Curious why you don't need to do type checking here?  I guess I don't understand this code very well.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -684,7 +685,7 @@ public void voltQueueSQL(final String sql, Object... args) {
                     throw new VoltAbortException(msg);
                 }
             }
-            queuedSQL.params = getCleanParams(queuedSQL.stmt, argumentParams);
+            queuedSQL.params = getCleanParams(queuedSQL.stmt, false, argumentParams);
```
wolffcm: Is this the "experimental" version of voltQueueSQL, that takes a String instead of a SQLStmt?  I guess in that case we have no way of doing type-checking since we don't yet know the parameter types.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
hhakimi53: True. Sure, I update it. thnxs

wolffcm:
End of review round.

src/frontend/org/voltdb/ProcedureRunner.java
hhakimi53: I went for type checking inside getCleanParams as logic iterates through the parameter list and all the information needed to perform is available in there to save some computation cycles. If we want, we can have type checking be done separately before getCleanParams() though it will have to iterate over the parameter list again. I am fine doing it either way.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
hhakimi53: Yes, there is an overlap with the tests in here. Only gain is this are much more granular tests and had came up with them when had with initial goal for this ticket which was to apply specific to Geo. If there is complete over-lap, there is, I will drop these tests.

src/frontend/org/voltdb/ProcedureRunner.java
hhakimi53: The intent here is to limit the type checking logic specifically to voltqueuesql interface only and this path is exercised by other system procedures, so did not intentionally enable type compatibility checking here when fetching parameter set for this case. Type compatibility checking currently is happening in voltqueuesql path only.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -840,21 +840,37 @@ public DependencyPair executeSysProcPlanFragment(
...
         }
         else {
+            if (argClass.isArray() && (argClass.getComponentType().isArray() ||     // supplied argument can be array of varbinary
+                                       argClass != byte[].class)) {                 // supplied argument is not varbinary but array of some other type
+                assert(expectedType == VoltType.VARBINARY);
+                // For in list arguments, passed in argument can be an array of
+                // varbinary. It would be nice if the information about expected type
+                // being array of types or not was available at this level and would haved
+                // helped in making logic simple and more concise. In absence of that
+                // will have to weaken the checks so that we don
```
wolffcm: So that we don... seems like you meant to finish this comment.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
```diff
@@ -213,6 +213,68 @@ else if ((left.isNumber() && !right.isNumber()) ||
...
+    // Float     | y     y      y      y       y      y         y        .          .        .          .
+    // Double    | y     y      y      y       y      y         y        .          .        .          .
+    // Decimal   | y     y      y      y       y      y         y        .          .        .          .
+    // Timestamp | y     y      y      y       y      y         y        .          .        .          .
+    // Varchar   | .     .      .      .       .      .         .        y          .        .          .
+    // VarBinary | .     .      .      .       .      .         .        .          y        .          .
+    // GeoPoint  | .     .      .      .       .      .         .        .          .        y          .
+    // Geography | .     .      .      .       .      .         .        .          .        .          y
+
+    private static boolean implicitTypeConvFeasible4Comparison(VoltType from, VoltType to) {
```
wolffcm: Do we use the numeral "4" as shorthand for "for" in other places in the code?  If not, I think it would be good to spell out the whole word.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
```diff
@@ -213,6 +213,68 @@ else if ((left.isNumber() && !right.isNumber()) ||
                                     left + "' and '" + right + "' types");
     }
 
+    // type conversion which can happen in EE that triggers comparison.
+    // this case is for arguments that supplied as array of arguments which
+    // implies that it uses In list operator and which will atmost trigger
+    // conversion between data-types involved for comparison
```
wolffcm: Could I trouble you to add more punctuation to this comment?  Capital letters at the beginning of sentences, periods at the end, and quotes around variable names would be much appreciated.  I'm having trouble parsing this text, especially with variables named "from" and "to."

src/frontend/org/voltdb/ProcedureRunner.java
hhakimi53: Oh yes, the last sentence is would be read as "so that we don't introduce regression." I will fix that.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
hhakimi53: Sure, will update it.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
wolffcm: If this function is only called for feasibility where the `from` type is known to be an array, please make the method name reflect this.  Maybe `arrayTypeIsImplicityConvertible`?

hhakimi53: I had used shorthand to shorten function name and I don't know if it's used at other place or not. I will spell out the whole word.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
```diff
@@ -213,6 +213,68 @@ else if ((left.isNumber() && !right.isNumber()) ||
...
+    // Int       | y     y      y      y       y      y         y        .          .        .          .
+    // BigInt    | y     y      y      y       y      y         y        .          .        .          .
+    // Float     | y     y      y      y       y      y         y        .          .        .          .
+    // Double    | y     y      y      y       y      y         y        .          .        .          .
+    // Decimal   | y     y      y      y       y      y         y        .          .        .          .
+    // Timestamp | y     y      y      y       y      y         y        .          .        .          .
+    // Varchar   | .     .      .      .       .      .         .        y          .        .          .
+    // VarBinary | .     .      .      .       .      .         .        .          y        .          .
+    // GeoPoint  | .     .      .      .       .      .         .        .          .        y          .
+    // Geography | .     .      .      .       .      .         .        .          .        .          y
```
wolffcm: So this table is a little different from the one below?  Why is that?

hhakimi53: For param which is for In list arguments, the type conversion matrix is locked down to comparison matrix in EE and that's why the function name has comparison in it. The below matrix is for params which are non-array elements and can use conversion matrix that is based on castas(). Comparison matrix is more stricter and does not allow numeric to string conversion, so have used that information in here to identify more conversion that are not possible based on that.

tests/frontend/org/voltdb/regressionsuites/TestTypeConversionSuite.java
```diff
@@ -144,10 +180,44 @@ public void testTypeConversion() throws IOException, ProcCallException {
...
             rowId++;
         }
+
+        // test cases where supplied is an array of arguments. This internally uses select statements
+        // with IN predicate to test broadest conversion allowed in comparison of in list arguments
+        rowId = 0;
+        for(boolean[] from: m_typeConversionMatrixInList) {
+            int colInd = 0;
+            typeToTest = m_tableColTypeVal[rowId];
+            for (boolean to: from) {
```
wolffcm: I find variable names that are prepositions difficult to read.  I would suggest appending a noun to help make this more readable, `toType`, `fromType`.

hhakimi53: Sure, will update the variable names. thnxs

src/frontend/org/voltdb/utils/VoltTypeUtil.java
wolffcm: Ok, got it.  In the comment here can you indicate the source file in the EE where you're getting this table from?

hhakimi53: Sure, will do that. It's based on NValue::compare_withoutNull() in NValue.hpp

tests/frontend/org/voltdb/regressionsuites/TestTypeConversionSuite.java
```diff
@@ -68,6 +68,25 @@
             {false, false, false, false, false, false, false, false, false, false, true},   // POLYGON
         };
 
+    // row index provides the type converting "from" (user supplied data of certain type)
+    // column index provides the type converting "to" (the actual column type in table)
+    // to see type for column (to) or row (from), map it's index m_tableColTypeVal to get it's type
+    // this is based on allowed type conversion for comparison which in list param arguments uses
+    boolean [][] m_typeConversionMatrixInList =
```
wolffcm: This type conversion matrix is never modified?  Please make it static and final.

hhakimi53: ok, sure.

wolffcm:
End of review round.

hhakimi53:
Thanks Chris for the review feedback. I have pushed the changes based on review feedback and jenkins on the changes so far is looking fine.

wolffcm:
Looks good.  Thanks for making the changes.  The logic here seems really more complicated that one would expect because we don't know if the formal parameters are arrays or not here.  Hopefully as part of one of the other tickets we can fix this soon and simplify the logic.

Go ahead and merge when Jenkins looks good.

PR #3378, opened 2016-03-10
sqlQueries.txt: fixed test to expect a correct NULL timestamp value (not

johnTcrawford:
'1970-01-01 00:00:00.000000'), now that VMC-267 bug is fixed.


paulmartel:
Looks good.

PR #3382, opened 2016-03-10
Fix for edge case for point at poles

hhakimi53:
Fix edge case in canonicalize/normalize logic to not evaluate/update point for antimeridian condition when the point is at pole.

### DIFF
--- a/src/ee/common/GeographyPointValue.hpp
+++ b/src/ee/common/GeographyPointValue.hpp
@@ -228,10 +228,9 @@ private:
             // longitude doesn't matter, so choose 0.
             newLng = 0.0;
         }
-
-        // If point is not at the poles, evaluate longitudes within epsilon
-        // of the antimeridian (on the east side), canonicalize to 180.0.
         else if (180.0 + m_longitude < epsilon()) {
+            // If point is not at the poles, evaluate longitudes within epsilon
+            // of the antimeridian (on the east side), canonicalize to 180.0.
             newLng = 180.0;
         }
 
--- a/src/ee/common/GeographyPointValue.hpp
+++ b/src/ee/common/GeographyPointValue.hpp
@@ -229,9 +229,9 @@ private:
             newLng = 0.0;
         }
 
-        // For longitudes within epsilon of the antimeridian
-        // (on the east side), canonicalize to 180.0.
-        if (180.0 + m_longitude < epsilon()) {
+        // If point is not at the poles, evaluate longitudes within epsilon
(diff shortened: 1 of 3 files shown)


wolffcm:
There is equivalent code in the GeographyPoint equals method.  Could you check to see if we have the same issue there too?  And if so, could you apply the same fix, and add a test?

src/ee/common/GeographyPointValue.hpp
```diff
@@ -229,9 +229,9 @@ class GeographyPointValue {
             newLng = 0.0;
         }
 
-        // For longitudes within epsilon of the antimeridian
-        // (on the east side), canonicalize to 180.0.
-        if (180.0 + m_longitude < epsilon()) {
+        // If point is not at the poles, evaluate longitudes within epsilon
```
wolffcm: I just have a style comment, that I find it a little confusing to have a newline and a couple line of comments between an `if` block and its following `else` block.  This is a matter of personal style though.

wolffcm:
That's all my comments.

src/ee/common/GeographyPointValue.hpp
hhakimi53: I will move the comment inside the else condition block.

wolffcm:
Sorry I meant to say that the Java class GeographyPoint might have the same problem, in case my earlier comment wasn't clear.

hhakimi53:
GeographyPointValue.equals() is good. I have added test cases to verify the same.

src/ee/common/GeographyPointValue.hpp
```diff
@@ -229,9 +229,9 @@ class GeographyPointValue {
...
+        // of the antimeridian (on the east side), canonicalize to 180.0.
```
BillWhite: I guess this works.

BillWhite:
This seems ok to me.  I would have imagined that you would be changing the equals method, and the constructor doesn't matter that much. But maybe I don't understand it.

That's my two shekels.

hhakimi53:
Bill, sorry, I did not get the point about the changes needed to equals method. The issue seems to be   canonicalize() that is used by compareWith() to evaluate if two geo points are equal. In canonicalize() we were losing the information whether point was at pole or not for the case the point was at antimeridian longitude. May be I am missing the point you are trying to make, please let me know. Also GeographyPoint (frontend/java) equals method works when it is asked to evaluate if two points are equal or not. So did not see need for changes in equals().

hhakimi53:
Talked to Bill and clarified about it. Will merge the changes to master.

PR #3387, opened 2016-03-10
ENG-9900 remove export enabled from catalog

dweiss88:


akhanzode:
if you have tested it and jenkins is happy merge.

PR #3388, opened 2016-03-11
ENG-9902 report bad export config on catalog update

dweiss88:


akhanzode:
Can you add tests where you do update with bad deployment?

PR #3389, opened 2016-03-11
Don't include column names when skip meta data is set

hhakimi53:
Fixed issue for sqlcmd's fixed format output, to honor skip metadata flagso that column names and the  extract line break if skip meta-data flag was set. Skip meta-data flag works as expected for tabulated and csv ouput format, so no changes needed in there.

### DIFF
--- a/tests/frontend/org/voltdb/TestVoltTable.java
+++ b/tests/frontend/org/voltdb/TestVoltTable.java
@@ -971,6 +971,17 @@ public class TestVoltTable extends TestCase {
         assertEquals(rowcounter, content.length);
     }
 
+    private void assertExpectedOutput(String expected, String actual) {
+        if (!actual.equals(expected))
+        {
+            System.out.println("Received formatted output:");
+            System.out.println(actual);
+            System.out.println("Expected output:");
+            System.out.println(expected);
+        }
+        assertEquals(expected, actual);
+    }
+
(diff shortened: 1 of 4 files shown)


tests/frontend/org/voltdb/TestVoltTable.java
```diff
@@ -1105,15 +1118,32 @@ public void testFormattedString() throws JSONException, IOException {
...
+" 12345678901 POLYGON ((179.12 89.9, -179.12 89.9, -179.1 -89.9, 179.1 -89.9, 0.0 0.0, 1.123 1.11, 179.12 89.9)) POINT (0.0 0.0)       1970-01-01 00:00:00.000099 \n";
 
+        if (!formattedOutput.equals(expectedOutput))
+        {
+            System.out.println("Received formatted output: (length: " + formattedOutput.length() +")");
+            System.out.println(formattedOutput);
+            System.out.println("Expected output: (length: " + expectedOutput.length() + ")");
+            System.out.println(expectedOutput);
+        }
+        assertTrue(formattedOutput.equals(expectedOutput));
```
paulmartel: This block of code should be factored out as a function like "void assertExpectedOutput(String expectedOutput, String formattedOutput)"

paulmartel:
Looks correct.

PR #3392, opened 2016-03-11
ENG-10010 don't replay deployment file changes during recover

vtkstef:


src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1044,6 +1055,7 @@ public void run() {
                         log.error("Received error response for updating catalog " + r.getStatusString());
                         return;
                     }
+                    m_catalogContext.set(VoltDB.instance().getCatalogContext());
```
nshi: Is there any race condition for updating the reference when the response comes back? What if something uses the reference after the catalog update txn but before the response gets back?

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -154,7 +154,14 @@ else if (work.invocationName.startsWith("@AdHoc")) {
             }
             newCatalogBytes = loadResults.getFirst().getFullJarBytes();
             retval.catalogBytes = newCatalogBytes;
-            retval.catalogHash = loadResults.getFirst().getSha1Hash();
+            retval.isForReplay = work.replayHashOverride != null;
```
nshi: Just call `work.isForReplay()`.

src/frontend/org/voltdb/InvocationDispatcher.java
vtkstef: good point. I left this bit of logic unchanged form how it was implemented before. Consider also that this bit of code gets executed only and if only it is the first snapshot restore on an empty schema

PR #3398, opened 2016-03-15
ENG-9948 add insert CRUD proc to export tables

dweiss88:


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1115,8 +1116,10 @@ private static void loadStoredProcedures(Map<String,Map<Integer, List<String>>>
             proc_param_counts.put(this_proc, curr_val);
         }
         params.resetRowPosition();
+        List<String> userProcs = new ArrayList<String>();
```
manjujames: Use Set here instead of List.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1132,6 +1135,10 @@ private static void loadStoredProcedures(Map<String,Map<Integer, List<String>>>
             argLists.put(param_count, this_params);
             procedures.put(proc_name, argLists);
         }
+        for (String proc_name : new ArrayList<String>(procedures.keySet())) {
```
manjujames: I don't understand this. Can we talk when you come in?

PR #3399, opened 2016-03-16
ENG-9547: DR consumer tracker fix.

nshi:
- Use Guava RangeSet instead of implementing from scratch.
- Allow overlaps when merging two trackers.
- Move JSON serialization/deserialization to the tracker class.
- Fix some logic around last ack point and the calculation of the safe
  point to use for acking. The last ack point could be way before the
  start of the first entry. It's incorrect to use the end of the first
  entry to ack. The range between the last ack point and the first entry
  should be counted as a gap. The last ack point should not be tracked
  separately. It acts the same way as a range in the range set. I added
  it to the first entry of the range set, which makes tracking a lot
  easier.

### DIFF
--- a/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
+++ b/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
@@ -18,41 +18,98 @@
 package org.voltdb;
 
 import java.nio.ByteBuffer;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.TreeMap;
+import java.util.Iterator;
 
+import com.google_voltpatches.common.collect.BoundType;
+import com.google_voltpatches.common.collect.DiscreteDomain;
+import com.google_voltpatches.common.collect.Range;
+import com.google_voltpatches.common.collect.RangeSet;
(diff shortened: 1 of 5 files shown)


slowsoul:
End of review.

src/frontend/org/voltdb/DRConsumerDrIdTracker.java
```diff
@@ -86,150 +141,82 @@ public void serialize(byte[] flattened) {
...
+        if (newTruncationPoint < getFirstDrId()) return;
+        final Iterator<Range<Long>> iter = m_map.asRanges().iterator();
+        while (iter.hasNext()) {
+            final Range<Long> next = iter.next();
+            if (end(next) < newTruncationPoint) {
+                iter.remove();
+            } else if (next.contains(newTruncationPoint)) {
+                iter.remove();
+                m_map.add(range(newTruncationPoint, end(next)));
+                break;
```
wweiss-voltdb: this can be a return

nshi: True. I'll make the change.

PR #3400, opened 2016-03-16
Eng 9322

jpiekos:
CLI changes to read password directly from user (the console), vs supply it in plain text on cmd line.
Changes for voltadmin, all the bulk loaders, and sqlcmd.

### DIFF
--- a/src/frontend/org/voltdb/CLIConfig.java
+++ b/src/frontend/org/voltdb/CLIConfig.java
@@ -243,7 +243,17 @@ public abstract class CLIConfig {
                 if (console == null) {
                     throw new IOException("Unable to read password from console.");
                 }
-                return new String(console.readPassword(prompt));
+                char[] val = console.readPassword("%s", prompt);
+                if (val != null)
+                {
+                    return new String(val);
+                }
+                else
+                {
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/CLIConfig.java
```diff
@@ -228,6 +231,24 @@ public void parse(String cmdName, String[] args) {
...
+        // then read password from the console
+        if ((null!=user) && (user.trim().length()>0))
+        {
+            if ((pwd==null) || (pwd.trim().length() == 0))
+            {
+                Console console = System.console();
+                if (console == null) {
+                    throw new IOException("Unable to read password from console.");
+                }
+                return new String(console.readPassword(prompt));
```
nshi: `readPassword()` takes a format string and arguments for the format. If the given prompt contains things like `%s`, it will throw `IllegalFormatException` here because it expects a matching argument. To be safe, you can use `readPassword("%s", prompt)`.

`readPassword()` can also return null if EOF is read no input. In that case, you can prompt for input again until an password is read. A while loop would do.

nshi:
Squash merged externally.

PR #3402, opened 2016-03-17
ENG-10071 - Remove more dead code, post VEM.

jpiekos:
Pull out underlying SSH lib and associated code, as nothing uses it now that :boom: VEM :boom: is removed.

**ENG-10071**


akhanzode:
Collector and TestCollector can not be removed as is. voltdb collect still uses them. remove only the uploadToServer part.

PR #3410, opened 2016-03-21
a command line tool for reserving equipment via jenkins rest api

jwiebema1:


tools/jenkins/resv
```diff
@@ -0,0 +1,336 @@
+#!/usr/bin/env python
+
```
prosegay: add license header & make sure this passes license check b4 you check it in

PR #3411, opened 2016-03-22
ENG-10049: Change clock skew threshold.

nshi:
- Configurable max clock skew via `MAX_CLOCKSKEW` system property.
- Change default max clock skew to 200ms.
- Change warning threshold to 100ms.


luyangco:
Reviewed. Looks fine to me.

PR #3413, opened 2016-03-23
Eng 10072 upsert with defaults

paulmartel:
Fixed three problems: 
The reported problem was that UPSERT statements that did not provide values for some columns were resetting the missing columns back to their default values for the UPDATE case rather than leaving them alone.
Added testing showed two other problems:
1. The planner was forcing UPSERTs (just like INSERTs) to mention all NOT NULL columns that had NO/NULL default values. For UPSERTs, this needs to be let through and conditionally caught in a runtime check.
It is a completely irrelevant condition for UPSERT's UPDATE subcase as hinted by the first fix described above. For UPSERT's INSERT subcase, we will throw a null constraint violation from the EE. That is, we won't bother to distinguish this edge case of defaulting the NOT NULL column to null from the case of explicitly setting it to a NULL value.
2. The planner was NOT forcing UPSERTs to provide values for all primary key components. I may be wrong here, but using the default value, commonly null, for the primary key that the UPSERT will use for matching seems like it is most likely just going to be covering for a pilot error and not what they intended. Is this too presumptious?

### DIFF
--- a/src/ee/executors/insertexecutor.cpp
+++ b/src/ee/executors/insertexecutor.cpp
@@ -123,6 +123,9 @@ bool InsertExecutor::p_init(AbstractPlanNode* abstractNode,
     TableTuple tuple = m_templateTuple.tuple();
 
     std::set<int> fieldsExplicitlySet(m_node->getFieldMap().begin(), m_node->getFieldMap().end());
+    // These default values are used for an INSERT including the INSERT sub-case of an UPSERT.
+    // The defaults are purposely ignored in favor of existing column values
+    // for the UPDATE subcase of an UPSERT.
     m_node->initTupleWithDefaultValues(m_engine,
                                        &m_memoryPool,
                                        fieldsExplicitlySet,
@@ -133,7 +136,7 @@ bool InsertExecutor::p_init(AbstractPlanNode* abstractNode,
     return true;
 }
 
-bool InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable** ptrToTable) {
+void InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable** ptrToTable) {
     PersistentTable* table = *ptrToTable;
     int tupleLimit = table->tupleLimit();
     int numTuples = table->visibleTupleCount();
@@ -155,8 +158,6 @@ bool InsertExecutor::executePurgeFragmentIfNeeded(PersistentTable** ptrToTable) 
         // the correct instance of PersistentTable.
         *ptrToTable = static_cast<PersistentTable*>(m_node->getTargetTable());
     }
-
-    return true;
 }
 
 bool InsertExecutor::p_execute(const NValueArray &params) {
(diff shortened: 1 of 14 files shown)


tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -444,44 +453,43 @@ static private void validateTableOfLongs(String messagePrefix,
...
-        }
+        validateTableOfLongs("", vt, expected);
     }
 
     static protected void validateRowOfLongs(String messagePrefix, VoltTable vt, long [] expected) {
         int len = expected.length;
         assertTrue(vt.advanceRow());
         for (int i=0; i < len; i++) {
-            String message = messagePrefix + "at column " + i + ", ";
+            String message = messagePrefix + "at column " + (i+1) + ", ";
```
wolffcm: Maybe because I have never programmed in Fortran I prefer to see array indexes formatted starting from 0.  But I understand that this is a matter of preference.

wolffcm:
I agree with you about needing to supply values for all the primary key components for UPSERT.  Seems that the whole point of the statement is update or create values for some particular PK, so the PK values should always be completely specified.

I am less sure about allowing values to remain unspecified for columns with a NOT NULL constraint.  In this case, the insert subcase of UPSERT will always fail, and this is something that can be detected at plan time.  Wouldn't the user just rather execute an UPDATE statement for this case?  I guess the primary difference is that an UPDATE statement would just return a 0 (no rows modified) instead of throwing an null constraint exception.  I don't have terribly strong feeling either way.

src/ee/executors/insertexecutor.cpp
```diff
@@ -227,120 +257,80 @@ bool InsertExecutor::p_execute(const NValueArray &params) {
...
+                // the templateTuple.
                 upsertTuple.move(templateTuple.address());
-                TableTuple &tempTuple = persistentTable->getTempTupleInlined(upsertTuple);
-
-                if (!persistentTable->updateTupleWithSpecificIndexes(existsTuple, tempTuple,
-                        persistentTable->allIndexes())) {
-                    VOLT_INFO("Failed to update existsTuple from table '%s'",
-                            persistentTable->name().c_str());
-                    return false;
+                TableTuple &tempTuple = persistentTable->copyIntoTempTuple(existsTuple);
```
wolffcm: I recall that this logic was pretty sensitive and attempts on my part to simplify it seemed to lead to performance regressions in TPCC.  That said it looks a lot better so I think we should just keep an eye on the perf graphs after it merges.

wolffcm:
These changes look good to me.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
paulmartel: SQL is a 1-based language, and there is a specific precedent for 1-based column numbering, namely "order by 1, 2"

PR #3418, opened 2016-03-23
Test the bulk loader with geo types.

BillWhite:
The kafka, rabbitmq, csv and jdbc loaders all use the Bulk Loader to do
the actual insertions.  I added some geo values to the tests.

https://issues.voltdb.com/browse/ENG-9922


BillWhite:
I'm trying something different.  I don't this is the right test.

PR #3422, opened 2016-03-24
PostgreSQLBackend.java: VARBINARY column types are now transformed to

johnTcrawford:
PostgreSQL's BYTEA type, which seems to be the closest analogue (was BIN
VARYING); and added a transformation for binary constants, e.g. x'12AF'
gets transformed to E'\\x12AF', which is in PostgreSQL's preferred
format for BYTEA data.
sql_coverage_test.py: removed the kludge that ignored the known failures
(up to 1520 of them) in the index-varbinary test suite, when running
against PostgreSQL.
StandardNormalizer.py: changed the output format of VARBINARY column
data, to roughly match the same format that is used upon input (in an
INSERT statement), rather than Python's crazy default format for binary
data.

### DIFF
--- a/src/frontend/org/voltdb/PostgreSQLBackend.java
+++ b/src/frontend/org/voltdb/PostgreSQLBackend.java
@@ -55,6 +55,7 @@ public class PostgreSQLBackend extends NonVoltDBBackend {
         m_PostgreSQLTypeNames.put("int8", "BIGINT");
         m_PostgreSQLTypeNames.put("float8", "FLOAT");
         m_PostgreSQLTypeNames.put("numeric", "DECIMAL");
+        m_PostgreSQLTypeNames.put("bytea", "VARBINARY");
         m_PostgreSQLTypeNames.put("varbit", "VARBINARY");
         m_PostgreSQLTypeNames.put("char", "CHARACTER");
         m_PostgreSQLTypeNames.put("text", "VARCHAR");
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -144,6 +145,18 @@
...
+    private static final Pattern varbinaryConstant = Pattern.compile(
+            "x'(?<bytes>[0-9A-Fa-f]+)'",
+            Pattern.CASE_INSENSITIVE);
+    // Modifies a query containing a VARBINARY constant, e.g. x'12AF', which
+    // PostgreSQL does not support in that format, and replaces it with a
+    // VARBINARY constant in the format it does support, e.g. E'\\x12AF'
+    // (with lots of extra backslashes, for escaping at various levels)
+    private static final QueryTransformer varbinaryConstantTransformer
+            = new QueryTransformer(varbinaryConstant)
+            .prefix("E'\\\\\\\\x").suffix("'").groups("bytes");
```
wolffcm: Wow that is a lot of backslashes.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
```diff
@@ -134,10 +127,17 @@ def normalize_value(v, vtype, num_digits):
...
+        #    print "DEBUG: normalizing:", v
+        #    print "DEBUG: modified to:", r
         return r
+    elif vtype == FastSerializer.VOLTTYPE_VARBINARY and v:
+        # output VARBINARY columns in a format similar to that in which they
+        # are inserted into VoltDB, e.g. x'abcd1234', and not in Python's crazy
+        # default format, e.g. array('c', '\xab\xcd\x124'); and no, that is not
+        # a typo: Python (without calling hexlify) actually changes '34' to '4'
+        # because 34 is the (hex) ASCII value of the character '4'!
+        return "x'" + hexlify(v)
```
wolffcm: You don't need to close this single-quote here?

wolffcm:
Looks good to me.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: Yup! But that's what is needed to get, e.g., E'\\x12AF', as the comment says.

tests/scripts/examples/sql_coverage/StandardNormalizer.py
johnTcrawford: This is a cosmetic question, since we are returning a string here. Ultimately, as a string, the result will be displayed (if at all, e.g., if this query is a mismatch/failure, or if the verbose option is turned on), on one of the HTML results pages, in single quotes. I originally did close the single-quote, but then the displayed result had the format 'x'abcd1234'', ending with two single-quotes. In my judgement, 'x'abcd1234', with just one single-quote at the end, looks better. Ideally, I'd love to get just x'abcd1234' without the leading single-quote either, but, since I like normal, non-binary strings being enclosed in single quotes, this is not achievable without some special case code for strings starting with x', or something like that, which just does not seem worth it.

PR #3424, opened 2016-03-24
load_remove_classes.outbaseline: removed the (last) line that reads:

johnTcrawford:
InsertEmployee            	varchar, varchar, varchar
since ENG-8532 has been fixed, this stored procedure is no longer listed
by SHOW PROCEDURES, *after* it has been dropped (though it is still
listed earlier); and added additional lines to match the ones added to
load_remove_classes.in.
load_remove_classes.in: added some additional:
    drop table FOO if exists;
statements (where FOO is t2, addable, dropable, prefixes, or raw); now
that ENG-8532 is fixed, such lines could actually help this test past,
by removing extraneous procedures from the ones listed by SHOW
PROCEDURES (whereas before, they had no real effect).

### DIFF
--- a/tests/sqlcmd/baselines/ddl/load_remove_classes.outbaseline
+++ b/tests/sqlcmd/baselines/ddl/load_remove_classes.outbaseline
@@ -8,6 +8,21 @@ Command succeeded.
 drop table t if exists;
 Command succeeded.
 
+drop table t2 if exists;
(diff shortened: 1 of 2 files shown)


wolffcm:
Looks good to me.

hhakimi53:
looks good.

PR #3427, opened 2016-03-28
ENG-9547: Keep the size of the Site tracker as small as possible to a…

wweiss-voltdb:
…void very large snapshot digest files. We keep the tracker small by assigning the last acked to the tracker just before we send the tracker to the site.

### DIFF
--- a/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
+++ b/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
@@ -338,7 +338,8 @@ public class DRConsumerDrIdTracker implements Serializable {
         }
         else {
             sb.append("span [").append(DRLogSegmentId.getSequenceNumberFromDRId(getFirstDrId())).append("-");
-            sb.append(DRLogSegmentId.getSequenceNumberFromDRId(getLastDrId())).append("]");
+            sb.append(DRLogSegmentId.getSequenceNumberFromDRId(getLastDrId()));
+            sb.append(", size=").append(size()).append("]");
         }
         return sb.toString();
     }
--- a/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
+++ b/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
@@ -84,11 +84,22 @@ public class DRConsumerDrIdTracker implements Serializable {
         }
     }
 
-    public DRConsumerDrIdTracker(long initialAckPoint, long spUniqueId, long mpUniqueId) {
+
+    private DRConsumerDrIdTracker(long spUniqueId, long mpUniqueId) {
         m_map = TreeRangeSet.create();
-        m_map.add(range(initialAckPoint, initialAckPoint));
         m_lastSpUniqueId = spUniqueId;
         m_lastMpUniqueId = mpUniqueId;
+    }
+
+    public static DRConsumerDrIdTracker createBufferTracker(long spUniqueId, long mpUniqueId) {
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/DRConsumerDrIdTracker.java
```diff
@@ -84,13 +84,24 @@ private static long end(Range<Long> range) {
...
+
+    private DRConsumerDrIdTracker(long spUniqueId, long mpUniqueId) {
         m_map = TreeRangeSet.create();
-        m_map.add(range(initialAckPoint, initialAckPoint));
         m_lastSpUniqueId = spUniqueId;
         m_lastMpUniqueId = mpUniqueId;
     }
 
+    public static DRConsumerDrIdTracker createBufferTracker(long spUniqueId, long mpUniqueId) {
+        DRConsumerDrIdTracker newTracker = new DRConsumerDrIdTracker(spUniqueId, mpUniqueId);
```
nshi: There are some method in this class that assume the underlying `m_map` is never empty. This may break the assumption.

nshi: I know that this allows for setting the truncation point after building the tracker. The other approach is to create the tracker with -1 as the initial ack point, then call `truncate()` with the actual truncation point later.

src/frontend/org/voltdb/DRConsumerDrIdTracker.java
```diff
@@ -306,6 +326,23 @@ public JSONObject toJSON() throws JSONException
...
+            return "Empty Map";
+        }
+        StringBuilder sb = new StringBuilder();
+        sb.append("lastSpUniqueId ").append(UniqueIdGenerator.toShortString(m_lastSpUniqueId)).append(" ");
+        sb.append("lastMpUniqueId ").append(UniqueIdGenerator.toShortString(m_lastMpUniqueId)).append(" ");
+        if (m_map.isEmpty()) {
+            sb.append("[empty map]");
+        }
+        else {
+            sb.append("span [").append(DRLogSegmentId.getSequenceNumberFromDRId(getFirstDrId())).append("-");
```
nshi: I think it'll also be helpful to print the number of entries in the map.

src/frontend/org/voltdb/SnapshotSiteProcessor.java
```diff
@@ -215,7 +215,7 @@ public static void populateSequenceNumbersForExecutionSite(SystemProcedureExecut
             }
 
             long[] ackOffSetAndSequenceNumber =
-                context.getSiteProcedureConnection().getUSOForExportTable(t.getSignature());
+                    context.getSiteProcedureConnection().getUSOForExportTable(t.getSignature());
```
nshi: Are these just indentation changes?

src/frontend/org/voltdb/SnapshotSiteProcessor.java
```diff
@@ -792,7 +792,11 @@ private static void logSnapshotCompleteToZK(
                     jsonObj.put("isTruncation", false);
                 }
                 extraSnapshotData.mergeToZooKeeper(jsonObj, SNAP_LOG);
-                zk.setData(snapshotPath, jsonObj.toString(4).getBytes("UTF-8"), stat.getVersion());
```
nshi: It's weird that we pretty print the JSON that gets stored in ZK. Dropping those spaces can save a fair amount.

nshi:
Reviewed.

PR #3429, opened 2016-03-28
Update statistic table/tuple column's for memory size to bigint instead of integer

hhakimi53:
Update column type's for table and memory components of statistics system procedure to BigInt instead of Integer so that the reported memory storage size for does not get reported in negative for values > ~2TB. The updated columns are for reporting non-inlined memory, total block storage for the table (used + free) and storage used by total tuples.

### DIFF
--- a/src/ee/storage/TableStats.cpp
+++ b/src/ee/storage/TableStats.cpp
@@ -51,9 +51,9 @@ void TableStats::populateTableStatsSchema(
     types.push_back(VALUE_TYPE_VARCHAR); columnLengths.push_back(4096); allowNull.push_back(false);inBytes.push_back(false);
     types.push_back(VALUE_TYPE_VARCHAR); columnLengths.push_back(4096); allowNull.push_back(false);inBytes.push_back(false);
     types.push_back(VALUE_TYPE_BIGINT);  columnLengths.push_back(NValue::getTupleStorageSize(VALUE_TYPE_BIGINT));  allowNull.push_back(false);inBytes.push_back(false);
-    types.push_back(VALUE_TYPE_INTEGER); columnLengths.push_back(NValue::getTupleStorageSize(VALUE_TYPE_INTEGER)); allowNull.push_back(false);inBytes.push_back(false);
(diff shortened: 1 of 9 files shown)


BillWhite:
Seems right to me.  I'm not an expert, though.

src/frontend/org/voltdb/MemoryStats.java
```diff
@@ -74,10 +74,10 @@ protected void populateColumnSchema(ArrayList<ColumnInfo> columns) {
         columns.add(new VoltTable.ColumnInfo("RSS", VoltType.INTEGER));
```
wolffcm: Would it make sense to change all of the INTEGER columns to be BIGINT?

wolffcm: Especially index memory seems to be in the same boat as the other fields.

wolffcm:
VMC makes use of these stats, right?  Do the GEB tests exercise this?  If not, maybe a ticket would be appropriate.  Also, did you test VMC by hand?  Would be good to test that output in sqlcmd looks right as well.

Do the tests you modified lock down any actual values?

wolffcm:
That's all my comments.

src/frontend/org/voltdb/MemoryStats.java
hhakimi53: I also had thought about it but looking at the numbers in ticket that was reported from customer end, the numbers were not high enough but than we didn't had .  I agree, this also appears to be in same boat as the other field and it's better to change the type, I will do that.

hhakimi53:
I had ran the statistics procedure by manually in VMC and sqlmd, they looked to fine to me. I don't know if VMC internally uses these fields for it's stats or not and whether GEB tests exercise this or not. I will try to find out about it and if needed, will file a ticket for it. That's a good point, thanks for pointing that out; I had not thought about it. 
Also the junit test that were modified don't locked the values, they are testing with empty tables as far as remember.  I had manually verified the values getting reported were the same before and after the test.

src/ee/storage/TableStats.cpp
```diff
@@ -42,6 +42,7 @@ vector<string> TableStats::generateTableStatsColumnNames() {
     return columnNames;
 }
 
+// make sure to update schema in IndexStats.java when updating the index stats schema.
```
wolffcm: Do you mean TableStats.java here?

wolffcm:
Looks good to me.

PR #3430, opened 2016-03-28
ENG-9985: Ignore table row limit when applying DR data.

nshi:
- Ignore table row limit when applying DR data.
- Allow different table row limits/delete policies on master and replica clusters.

### DIFF
--- a/src/catgen/in/javasrc/DRCatalogDiffEngine.java
+++ b/src/catgen/in/javasrc/DRCatalogDiffEngine.java
@@ -31,11 +31,10 @@ import org.voltdb.utils.Encoder;
 
 /**
  * Specialized CatalogDiffEngine that checks the following conditions:
- * 1. The localCatalog contains all DR tables contained in the remote catalog
- * 2. All shared DR tables contain the same columns in the same order
- * 3. All shared DR tables have the same partition column
- * 4. All shared DR tables have the same row limits/delete policies
- * 5. All shared DR tables have the same unique indexes/primary keys
+ * - The localCatalog contains all DR tables contained in the remote catalog
+ * - All shared DR tables contain the same columns in the same order
+ * - All shared DR tables have the same partition column
+ * - All shared DR tables have the same unique indexes/primary keys
  */
 public class DRCatalogDiffEngine extends CatalogDiffEngine {
     public DRCatalogDiffEngine(Catalog localCatalog, Catalog remoteCatalog) {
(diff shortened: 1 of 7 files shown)


src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -34,7 +34,6 @@
  * 1. The localCatalog contains all DR tables contained in the remote catalog
  * 2. All shared DR tables contain the same columns in the same order
  * 3. All shared DR tables have the same partition column
- * 4. All shared DR tables have the same row limits/delete policies
  * 5. All shared DR tables have the same unique indexes/primary keys
```
luyangco: The comment needs to be reindexed to 4.

nshi: I just pushed the change.

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -107,14 +113,9 @@ protected String checkModifyWhitelist(final CatalogType suspect, final CatalogTy
...
-            if ("cost".equals(field) ||
-                "seqscancount".equals(field) ||
-                "explainplan".equals(field) ||
-                "indexesused".equals(field) ||
-                "cachekeyprefix".equals(field)) {
-                return null;
-            }
-        } else if (isUniqueIndex(suspect) == isUniqueIndex(prevType) && !isUniqueIndexColumn(suspect)) {
+            return null;
+        } else if ((suspect instanceof Index && isUniqueIndex(suspect) == isUniqueIndex(prevType)) ||
```
luyangco: Should here to be `&&`?

nshi: No. The original conditions were very misleading, and it turned out to be wrong. I fixed it by splitting it into two conditions.

What it's trying to check are
* change of the uniqueness of an index, or
* change of columns inside an index

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -1939,6 +1939,23 @@ TEST_F(DRBinaryLogTest, DeleteOverBufferLimit) {
     ASSERT_TRUE(false);
 }
 
+TEST_F(DRBinaryLogTest, IgnoreTableRowLimit) {
+    m_tableReplica->setTupleLimit(100);
+
+    const int total = 101;
```
luyangco: It's better to use a greater number to make the test more obvious, e.g. 200, so the reader doesn't have to calculate how many rows you insert into the table.

nshi: Will do.

luyangco:
Review complete

PR #3433, opened 2016-03-29
ENG-10135: Fix XDCR timestmap mismatch conflict handling.

nshi:
When a timestamp mismatch conflict occurs, the resolver receives
incorrect timestamps for comparison. The timestamp for the remote
operation is Volt-epoch-based, but the local existing tuple's timestamp
is Unix-epoch-based. This resulted in always rejecting remote changes.

Now all timestamps passed to the resolver are Unix-epoch-based.


luyangco:
This patch looks fine to me.

PR #3435, opened 2016-03-30
Eng 10127 Remove catalog & project.xm & @procinfo from the test suite

pdshaw:
(Without creating discontinuity in the performance charts!)


tests/test_apps/voltkvqa/src/voltkvqa/procedures_withexport/PutMp.java
```diff
@@ -36,12 +36,6 @@
 import org.voltdb.VoltType;
 import org.voltdb.DeprecatedProcedureAPIAccess;
 
-//@ProcInfo
-//(
-//  partitionInfo   = "store.key:0"
-//, singlePartition = true
-//)
-
```
akhanzode: Isnt PutMP supposed to be MP?

pdshaw: It was commented; just removed the dead code.
However maybe it was a mistake to comment it out in the first place? We can go over this.

tests/test_apps/voltkvqa/run.sh
```diff
@@ -163,18 +169,18 @@ function http-benchmark() {
         --maxvaluesize=1024 \
         --usecompression=false \
         --warmup=15 \
-        --threads=250
+        --threads=35
 }
```
prosegay: double check that you really intended to change these values in this change set...

pdshaw: Yes, dropped intentionally. Was hitting connection exhaustion that left the HTTP interface essentially dead.

Will cross check to make sure apprunner setup reflects a similar reduction.

PR #3437, opened 2016-03-31
Handle incompatible param types to timestamp sql functions in EE

hhakimi53:
Updated logic to generate exception if the caller supplies value of different type for timestamp function where timestamp value is expected and if conversion of value of given type to timestamp is not applied.

### DIFF
--- a/src/ee/expressions/datefunctions.h
+++ b/src/ee/expressions/datefunctions.h
@@ -133,6 +133,11 @@ template<> inline NValue NValue::callUnary<FUNC_EXTRACT_YEAR>() const {
     if (isNull()) {
         return *this;
     }
+
+    if(getValueType() != VALUE_TYPE_TIMESTAMP) {
+        throwCastSQLException(getValueType(), VALUE_TYPE_TIMESTAMP);
+    }
+
     int64_t epoch_micros = getTimestamp();
     boost::gregorian::date as_date;
     micros_to_date(epoch_micros, as_date);
@@ -144,6 +149,11 @@ template<> inline NValue NValue::callUnary<FUNC_EXTRACT_MONTH>() const {
     if (isNull()) {
         return *this;
     }
+
+    if(getValueType() != VALUE_TYPE_TIMESTAMP) {
+        throwCastSQLException(getValueType(), VALUE_TYPE_TIMESTAMP);
+    }
+
     int64_t epoch_micros = getTimestamp();
     boost::gregorian::date as_date;
     micros_to_date(epoch_micros, as_date);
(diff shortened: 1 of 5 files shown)


paulmartel:
There are two possible implementations described in the ticket.
One is described as "what we should do".
The other is described as an "undesirable" alternative.
This changeset seems to exactly match the "undesirable" option.
That's not good.
Also, we put one space between the "if" and the "(" as in `if (condition) {` NOT `if(condition) {`.

hhakimi53:
The description in ticket is out of date at present, I will update it.

The desirable behavior we were expecting earlier was to perform implicit type conversion at VoltQueqeSql level. The implicit conversion could have followed either type conversion model defined in castas() or compare() in NValue.hpp. The end result of these model are not the same for all cases, specially in castas() conversion the supplied value of type varchar can be converted to numeric though for type conversion following compare(), this conversion is blocked and is an error condition. At voltqueueSql, the information available does not provide us the context needed to make decision which conversion we should follow. So, it's not feasible today to do those conversion at voltQueueSql level.

The alternative approach was to error out or perform conversion in EE but was undesirable as it was earlier deemed to be not good performance wise. After looking into implementation of other function expression in EE, it was discovered we are already performing implicit conversion or type checking at runtime such as in numeric, string and bitwise function expression. The only place where conversion or type checking was missing is in timestamp functions. So it was decided in meeting given that we don't have full context available for conversion at voltQueueSql level and as we already performing type checking/conversion in EE, let's take same approach for timestamp function for the cases where conversion is not happening. 

I will update ticket with above information as ticket is not reflecting the expected behavior from this ticket.

I will also update the changes formatting for "if condition" to be inline with current formatting standards. Thanks.

paulmartel:
This changeset introduces type checks rather than type conversions.
Has the team agreed on this strict typing approach -- that there will be NO implicit conversion from java String or and numeric types passed to voltQueueSql in these cases?
That is, all voltQueueSql parameters passed in these cases will be required to be explicit instances of a VoltDB-defined or java standard Date/Time class?
This is as opposed to a less strict approach that could AT LEAST allow a properly formatted string.

An alternative is to implement this strict type check in the EE functions BUT ALSO to have the planner do a compensating expression tree rewrite whenever these specific functions are applied directly to parameters, so that, for example, `EXTRACT(YEAR FROM ?)` produces the exact same code as EXTRACT(YEAR FROM (CAST (? AS TIMESTAMP))). I
Your current approach would force the user to inject this cast syntax themselves. I'm not saying we SHOULD do this, and in fact I think that I am slightly opposed to doing this, but I AM saying that we should make sure that we have team consensus on what our goal is.

hhakimi53:
Chris and Bill has agreed with approach of strict type checks for these cases - error out for such cases and the current approach got formulated during discussion. The other alternative approach was also discussed if we should perform implicit conversion instead of error out directly. First one was in EE directly as we do for of the some cases but we didn't go with that one as it can be potentially costly performance wise and more of it's a silent one. Another alternative which was suggested, same as what you mentioned above - injection of explicit cast in planner for timestamp functions and we had decided not to take that route for 6.2 and for the currently discovered cases generate an error.

paulmartel:
Thanks for clarifying.

wolffcm:
Looks good to me.

PR #3439, opened 2016-03-31
DR mixed size tests cleanup

slowsoul:


src/frontend/org/voltdb/DRConsumerDrIdTracker.java
```diff
@@ -353,8 +353,8 @@ public String toString() {
         sb.append("lastSpUniqueId ").append(UniqueIdGenerator.toShortString(m_lastSpUniqueId)).append(" ");
         sb.append("lastMpUniqueId ").append(UniqueIdGenerator.toShortString(m_lastMpUniqueId)).append(" ");
         for (Range<Long> entry : m_map.asRanges()) {
-            sb.append("[").append(DRLogSegmentId.getSequenceNumberFromDRId(start(entry))).append(", ")
-              .append(DRLogSegmentId.getSequenceNumberFromDRId(end(entry))).append("] ");
+            sb.append("[").append(start(entry)).append("(").append(DRLogSegmentId.getSequenceNumberFromDRId(start(entry))).append("), ")
```
nshi: Why print the number twice?

slowsoul: It only shows the extracted sequence number range before, sometimes the original dr Id is needed (for clusterID information etc), so here I print like this: "drId(corresponding seq num)"

nshi: This makes the output really long. I suggest creating a static method in DRLogSegmentId to print a DR ID in the form of clusterID:seqNum and use that instead.

PR #3440, opened 2016-03-31
ENG-9977, bug fix, DR consumer doesn't send sync snapshot request whe…

luyangco:
…n DR is disabled at startup and enabled later.

if the database starts without DR enabled, the DR consumer won't be notified during the leader election.
After DR is enabled, DR consumer doesn't have the knowledge of who the locally led partitions are. So
all the dispatchers think they are not the leaders and wait for other to send the sync snapshot request.

Change-Id: I2e777a5ad6cb6d001e03e9b8392220d0c9b3f13b

### DIFF
--- a/src/frontend/org/voltdb/iv2/BaseInitiator.java
+++ b/src/frontend/org/voltdb/iv2/BaseInitiator.java
@@ -27,7 +27,6 @@ import org.voltdb.BackendTarget;
 import org.voltdb.CatalogContext;
 import org.voltdb.CatalogSpecificPlanner;
 import org.voltdb.CommandLog;
-import org.voltdb.ConsumerDRGateway;
 import org.voltdb.LoadedProcedureSet;
 import org.voltdb.MemoryStats;
 import org.voltdb.PartitionDRGateway;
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/iv2/BaseInitiator.java
```diff
@@ -239,6 +240,11 @@ public void setDurableUniqueIdListener(DurableUniqueIdListener listener)
     public void setConsumerDRGateway(ConsumerDRGateway gateway) {
         assert m_consumerDRGateway instanceof ConsumerDRGateway.DummyConsumerDRGateway;
         m_consumerDRGateway = gateway;
+        if (m_term != null && m_consumerDRGateway instanceof ClientInterfaceRepairCallback) {
+            // We're the leader, and this consumer gateway is late to the party
+            ClientInterfaceRepairCallback callback = (ClientInterfaceRepairCallback)gateway;
+            callback.repairCompleted(m_partitionId, m_initiatorMailbox.getHSId());
```
nshi: I don't think the initiator has to care about the consumer DR gateway anymore. We can remove it from the initiator completely and call this callback directly in `RealVoltDB.catalogUpdate()`.

voltdb/log4j.xml
```diff
@@ -115,7 +115,7 @@
     </logger>
 
     <!-- Always let the bulk loader and its ilk to write to the console. -->
-    <logger name="LOADER">
```
nshi: ???

nshi:
Review finished.

voltdb/log4j.xml
luyangco: An accidental change, fixed.

PR #3442, opened 2016-03-31
Throw an exception for internal tables with duplicate column names

wolffcm:
This change set addresses incorrect behavior that can occur when doing a "select *" from a subquery that contains multiple columns with the same name.  This can happen in a couple of ways:
```SQL
SELECT * FROM (SELECT * FROM T, T) AS subq;
```
Where if persistent table T has columns (A, B) the output of the suquery wll be (subq.A, subq.B, subq.A, subq.B).  This can also happen for this case:
```SQL
SELECT * FROM (SELECT t1.a AS z, t2.a AS z FROM T AS t1, T AS t2) AS subq;
```
VoltDB's name-based column resolution pass does not deal well with these queries and was resulting in undefined behavior.

This change set only detects queries that would have caused incorrect behavior and throws an exception.  A more thorough solution should be implemented soon to make these queries execute in accordance with the SQL standard.

### DIFF
--- a/src/frontend/org/voltdb/plannodes/NodeSchema.java
+++ b/src/frontend/org/voltdb/plannodes/NodeSchema.java
@@ -20,9 +20,12 @@ package org.voltdb.plannodes;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 import org.voltdb.expressions.TupleValueExpression;
+import org.voltdb.planner.PlanningErrorException;
 
 /**
  * This class encapsulates the representation and common operations for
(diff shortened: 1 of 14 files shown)


src/frontend/org/voltdb/plannodes/SchemaColumn.java
```diff
@@ -80,50 +86,81 @@ protected SchemaColumn clone()
...
 
-        if (tableAlias != null) {
-            if (tableAlias.equals(m_tableAlias)) {
-                sameTable = true;
-            }
-        } else if (m_tableName.equals(tableName)) {
-            sameTable = true;
+    private int nullSafeStringCompareTo(String str1, String str2) {
+        if (str1 == null ^ str2 == null) {
+            return str1 == null ? -1 : 1;
```
BillWhite: This is clever.  I didn't know Java let one use xor this way.

src/frontend/org/voltdb/plannodes/NodeSchema.java
```diff
@@ -78,13 +108,56 @@ public int size()
     public SchemaColumn find(String tableName, String tableAlias, String columnName, String columnAlias)
     {
         SchemaColumn col = new SchemaColumn(tableName, tableAlias, columnName, columnAlias);
-        Integer index = m_columnsMapHelper.get(col);
-        if (index != null) {
-            return m_columns.get(index.intValue());
+        int index = findIndexOfColumn(col);
```
BillWhite: There's something vaguely weird about this.  You create a SchemaColumn so that you can create a SchemaColumnFloor so that you can search.  That's two object creations just to do a single search, and then you throw away the two objects.  But I can't see any other way to do it.

In C++ you would create a local object and skip the malloc, but this is Java.

Maybe it's just me.

BillWhite:
That's all of my comments.  I don't really have anything substantive to say.  This looks good.

hhakimi53:
I am not much familiar with this code path. Went through the changes and overall changes look fine to me also.

src/frontend/org/voltdb/plannodes/NodeSchema.java
wolffcm: Creating short-lived objects is pretty normal in Java.  You can only do dynamic allocation of user-defined types, even when you want lifetime limited to the current stack frame.  Presumably the runtime has a way of making this really fast, since it's so common.  I agree that coming from C++ it seems strange.

PR #3450, opened 2016-04-04
ENG-10109 take advantage of Jettys continuation listener

vtkstef:


akhanzode:
if you want to add a log when we timeout that will also validate the cause. rest looks good.

PR #3453, opened 2016-04-05
ENG-10119-backport rate limit time moving backward error message

dweiss88:


manjujames:
It looks alright to me, other than the fact that it is a bit different from master change.

PR #3460, opened 2016-04-07
ENG-9908 enable HTTPS on web/rest interface

manjujames:


src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -963,6 +977,67 @@ public HTTPAdminListener(
...
     }
 
+    private String getKeyTrustStoreAttribute(String sysPropName, KeyOrTrustStoreType store, String valueType, boolean throwForNull) {
+        String sysProp = System.getProperty(sysPropName);
+        if (StringUtils.isNotBlank(sysProp)) {
+            return sysProp.trim();
+        } else {
+            String value = null;
+            if (store!=null) {
+                value = valueType.equals("path") ? store.getPath() : store.getPassword();
```
vtkstef: `"path".equals(valueType)`

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -963,6 +977,67 @@ public HTTPAdminListener(
...
+            }
+            if (StringUtils.isBlank(value) && throwForNull) {
+                    throw new RuntimeException(
+                        "To enable HTTPS, keystore must be configured with password in deployment file or using system property");
```
vtkstef: IllegalArgumentException

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -963,6 +977,67 @@ public HTTPAdminListener(
...
+            } else {
+                return (value!=null) ? value.trim() : value;
```
vtkstef: are you trimming passwords? if so don't

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -963,6 +977,67 @@ public HTTPAdminListener(
...
+        if (value!=null) {
+            sslContextFactory.setTrustStorePath(value);
+        }
+        value = getKeyTrustStoreAttribute("javax.net.ssl.trustStorePassword", httpsType.getTrustStore(), "password", false);
+        if (value!=null) {
+            sslContextFactory.setTrustStorePassword(value);
+        }
+        /* More configurable things that we are not using for now.
+        sslContextFactory.setKeyManagerPassword("password");
+        sslContextFactory.setExcludeCipherSuites("SSL_RSA_WITH_DES_CBC_SHA",
```
vtkstef: I know these are weak ciphers so we should exclude them

src/frontend/org/voltdb/utils/HTTPAdminListener.java
manjujames: Good catch. I don't think there is a need to trim paths either, so I'll just get rid of the trim.

akhanzode:
Does jetty automatically handle KEYSTORE_PASSWD_OBFUSCATED ? How does user specify this?

manjujames:
Yes, jetty handles it automatically. In my test with obfuscated password, I document how to generate the obfuscated password.  Users will need to generate the obfuscated password and can use that instead of the clear text password.

akhanzode:
Let Andrew know to document it.

PR #3472, opened 2016-04-12
ENG-10190 support voltdb initialize

vtkstef:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -341,6 +346,14 @@ public void startSampler() {
 
     private LatencyHistogramStats m_latencyHistogramStats;
 
+    private File getConfigLogDirectory() {
```
akhanzode: you may want this not private so Mock can do this.

PR #3473, opened 2016-04-13
ENG-10208 --force now deletes the contents of export overflow directory

manjujames:


src/frontend/org/voltdb/utils/VoltFile.java
```diff
@@ -189,6 +189,22 @@ public static void resetSubrootForThisProcess() throws IOException {
...
+        }
+
+        if (deleteRoot) {
+            recursivelyDelete(file);
+        } else {
+            if (file.isDirectory()) {
+                for (File f : file.listFiles()) {
+                    recursivelyDelete(f);
+                }
+            }
```
akhanzode: Shouldnt else delete the file?

src/frontend/org/voltdb/export/ExportManager.java
```diff
@@ -388,6 +392,21 @@ private ExportManager(
         exportLog.info(String.format("Export is enabled and can overflow to %s.", cluster.getExportoverflow()));
     }
 
+    private void clearOverflowPath(CatalogContext catContext) throws ExportManager.SetupException {
```
akhanzode: better name? clearOverFlowData() also log that you are doing this so we know when we look into log its was done.

src/frontend/org/voltdb/utils/VoltFile.java
manjujames: No, I wrote this method for the cases when DONT want to delete the root directory, but just its contents.

manjujames:
Squash merged.

PR #3475, opened 2016-04-13
ENG-10249, DR consumer initialization should be called after recover …

luyangco:
…completed,

disallow @UAC to initiate DR consumer during cl replay.

Change-Id: I59326dcc00ba65f49a5cab638b9ecdf8290b239f


nshi:
Review finished.

PR #3478, opened 2016-04-13
ENG-10024 fix export view recover test fix in valgrind.

JadenW:
Changes in voltdbipc:
enable getUSOForExportTable;
tweak buffer size calculation in tableStreamSerializeMore;

### DIFF
--- a/src/frontend/org/voltdb/jni/ExecutionEngineIPC.java
+++ b/src/frontend/org/voltdb/jni/ExecutionEngineIPC.java
@@ -463,15 +463,8 @@ public class ExecutionEngineIPC extends ExecutionEngine {
             if (dirty)
                 m_dirty = true;
 
-            if (resultTablesLength <= 0) {
-                System.out.println("resultTablesLength: " + resultTablesLength
-                        + " tables size " + tables.length);
+            if (resultTablesLength <= 0)
                 return;
-            } else {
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/jni/ExecutionEngineIPC.java
```diff
@@ -1352,10 +1368,11 @@ public void exportAction(boolean syncAction,
             retval[1] = results.getLong();
 
         } catch (final IOException e) {
+            e.printStackTrace();
```
vtkstef: why a printStackTrace() here?

JadenW: Sorry. this is for debug only. I will remove it.

tests/frontend/org/voltdb/regressionsuites/EEProcess.java
```diff
@@ -72,6 +72,7 @@ public int port() {
         args.add("--leak-check=full");
         args.add("--show-reachable=yes");
         args.add("--num-callers=32");
+        args.add("--track-origins=yes");
```
akhanzode: Do you want to set this option? It makes it slowwww

JadenW: This should also for debug only. I will remove it.

PR #3482, opened 2016-04-14
ENG-10222: Allow all supported online schema change on DR tables

slowsoul:
Allow users to locally add, drop and modify columns in DR tables


nshi:
Review finished.

PR #3483, opened 2016-04-14
ENG-10074 error logging for importers

zeemanhe:
Log warn message in callbacks rather than InternalClientResponseAdapter

### DIFF
--- a/src/frontend/org/voltdb/importer/AbstractImporter.java
+++ b/src/frontend/org/voltdb/importer/AbstractImporter.java
@@ -18,11 +18,14 @@
 package org.voltdb.importer;
 
 import java.net.URI;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.voltcore.logging.Level;
 import org.voltcore.logging.VoltLogger;
+import org.voltcore.utils.EstTime;
+import org.voltcore.utils.RateLimitedLogger;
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/importer/AbstractImporter.java
```diff
@@ -54,6 +59,27 @@
     private ImporterServerAdapter m_importServerAdapter;
     private volatile boolean m_stopping;
     private AtomicInteger m_backPressureCount = new AtomicInteger(0);
+    private final AtomicLong m_failures = new AtomicLong(0);
+
+    public class ImporterCallback implements ProcedureCallback{
```
vtkstef: Nope. This would be a wrapper of a wrapper. The InternalClientResponseAdapter.InternalCallback already wraps a callback. You should modify that one and use `InternalAdapterTaskAttributes` m_proxy field to see if it is an importer callback. Create a `boolean isImporterCallback()` method in `InternalAdapterTaskAttributes` where it would test whether or not the `m_proxy` field is null

```java
public boolean isImporterCallback() {
    return m_proxy != null;
}
```
use that method to see whether or not you need to log non successful client response

akhanzode: This sounds bit hackish can we put onus on concrete importer to log rather than here?

zeemanhe: updated as recommended. It makes sense to check for m_proxy as an importer since it should be null for http clients.

PR #3484, opened 2016-04-14
Fix lookups in group by expression lists.

BillWhite:
This commit fixes lookup problems when an expression in a group by list
is exactly equal to an alias in the select list.  The group by
expression must be a single identifier, not a more complex expression.
The identifier is replaced with the expression from the select list.

There were two problems here.  One is that a select statement with no
aggregate function call in the select list did not get any alias
expansion at all.  This is not the behavior MySQL and Postresql show.
Also, we did not check for ambiguity.  If a select list has more than
one expression with an alias, say "select a as ID, b as ID ...", then
grouping by ID would be ambiguous.  We always chose the first one.

Finally, if an alias as a group by expression names an expression which
contains an aggregate function it should be rejected.  We were doing
this in the lookup algorithm itself.  But this means that the error
message said that the alias in the group by was not defined, when the
problem was with aggregates.  So, with this commit we just do the
subtitution without checking.  Later on when the group by expressions
are processed the compiler will notice aggregate functions and reject
the statement.  But then it will be a better error message.


src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -327,26 +323,31 @@ void resolveColumnReferencesInGroupBy() {
...
-
-                    newUnresolvedExpressions.remove(element);
-                    break;
+                    if (matchcount == 1) {
+                        newUnresolvedExpressions.remove(element);
+                    }
                 }
             }
+            if (matchcount > 1) {
+                throw new HsqlException(String.format("Group by expression \"%s\" is ambiguous", alias), "", 0);
```
paulmartel: Why `String.format("Group by expression \"%s\" is ambiguous", alias)`
not simply, (readably?) `"Group by expression \"" + alias "\" is ambiguous"` ?
I see this idiom quite a bit, but never know why.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -919,8 +920,11 @@ private void resolveGroups() {
                     Expression.subqueryExpressionSet);
 
                 if (!tempSet.isEmpty()) {
-                    throw Error.error(ErrorCode.X_42572,
-                                      ((Expression) tempSet.get(0)).getSQL());
+                    String sql = ((Expression) tempSet.get(0)).getSQL();
+                    if (sql.endsWith("(")) {
+                        sql += ")";
+                    }
```
paulmartel: This is a bit obscure to go without a comment re: when/why `sql.endsWith("(")`

paulmartel:
End of review round

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
BillWhite: I find the string concatenation more difficult to read.  I have been using printf for 40 years or more, and it seems natural to me.  That's the only reason.

You left off a plus sign.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
BillWhite: Ok.

PR #3485, opened 2016-04-14
Fix an error message for unplannable queries.

BillWhite:
Some queries are not plannable because their joins cannot be computed
with our two fragment query model.  The error message for one such query
was misleading.  This commit fixes this.

Jira: ENG-10097.

### DIFF
--- a/src/frontend/org/voltdb/planner/StatementPartitioning.java
+++ b/src/frontend/org/voltdb/planner/StatementPartitioning.java
@@ -141,6 +141,10 @@ public class StatementPartitioning implements Cloneable{
     private String m_fullColumnName;
 
     private boolean m_joinValid = true;
+    
+    // If m_joinValid is set to false, we also set
+    // this string to a hint telling why it is false.
+    private String m_recentInvalidReason = null;
 
     /** Most of the time DML on a replicated table for a plan that is executed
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/planner/StatementPartitioning.java
```diff
@@ -494,6 +501,16 @@ public boolean isJoinValid() {
...
+    public String getJoinInvalidReason() {
+        return m_recentInvalidReason;
+    }
+
+    public void setJoinValid(boolean isValid, String why) {
+        assert((isValid && why == null) || (!isValid && why != null));
+        m_joinValid = isValid;
+        m_recentInvalidReason = why;
+    }
+
```
paulmartel: Refactor into super simple `setJoinValid()` and `setJoinInvalid(String why)`You always know (hard-code) at compile time when you are calling this to set valid or set iinvalid.
Unifying/weakening these into 1 function with a weaker runtime assert does not seem helpful.

paulmartel:
End of review round. Messages seem clear to me.

src/frontend/org/voltdb/planner/StatementPartitioning.java
BillWhite: Done.

PR #3488, opened 2016-04-18
Made CountingPostfilter::eval  method inline

alexeevm:


wolffcm:
This looks pretty straightforward to me.  If there's no objections I'd like to merge this so we can see how it performs.

PR #3489, opened 2016-04-18
Eng 10073 det

hhakimi53:
Logic updates to evaluate determinism based on equivalence filters and unique constraints defined on the table for multi-fragment plan. If the query has parameterized / constant equivalence filters and the equivalence filter covers the unique filters defined on them such that it results in at most one output row, the query result is order deterministic irrespective if it's SP or MP. For eg  "Select * from T where a = ?;" where a is unique constrained defined on the table T will result in one output row. Today though similar query gets flagged as non-deterministic for multi-fragment plan. Logic updates with this pull request addresses this issue by adding additional logic to order-determinism evaluation where it evaluates collector fragment to see if the collector fragment will result in at most one output row. Currently logic to evaluate if the plan results in at most one output row has been implemented for Index scan plan only nodes to address the issue. In future we can add logic to other plan node logic to determine if they will result in at most one output row only where ever feasible.

### DIFF
--- a/src/frontend/org/voltdb/planner/CompiledPlan.java
+++ b/src/frontend/org/voltdb/planner/CompiledPlan.java
@@ -161,7 +161,19 @@ public class CompiledPlan {
         if (m_statementIsOrderDeterministic) {
             return true;
         }
-        return rootPlanGraph.isOrderDeterministic();
+
+        if (rootPlanGraph.isOrderDeterministic()) {
+            return true;
+        }
+
+        // check if plans contains second fragment.
+        if (subPlanGraph != null) {
+            // If so evaluate case where the collector fragment will
+            // return at most one row to the co-ordinator fragment
+            return subPlanGraph.producesOneOutputRowOnly();
+        }
+        return false;
+
     }
 
     public boolean isContentDeterministic() {
--- a/src/frontend/org/voltdb/planner/PlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/PlanAssembler.java
@@ -497,8 +497,9 @@ public class PlanAssembler {
             rawplan = getNextPlan();
 
             // stop this while loop when no more plans are generated
-            if (rawplan == null)
+            if (rawplan == null) {
                 break;
+            }
             // Update the best cost plan so far
             m_planSelector.considerCandidatePlan(rawplan, parsedStmt);
         }
--- a/src/frontend/org/voltdb/planner/SubPlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/SubPlanAssembler.java
@@ -1452,11 +1452,11 @@ public abstract class SubPlanAssembler {
      * assess the data from the table according to the path.
      *
      * @param table The table to get data from.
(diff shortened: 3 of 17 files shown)


BillWhite:
I don't understand why we need to go to the plans for this.  You say a query returns only one row if it's on a table which has a unique constraint and all the keys in the constraint are somehow constrained.  We can already check for that in the AbstractParsedStatement, and I think more clearly than this.  I'm not sure adding another layer of determinism testing this way is helping.

wolffcm:
@BillWhite I encouraged Huzefa to query the plan nodes directly for cases like this.  In general, I'd like to see more analysis done on the plan nodes themselves, because it means they can be used more flexibly.

But you make a good point.  There's already lots of determinism checking code on in AbstractParsedStmt and ParsedSelectStmt.  Maybe it makes sense to put it there instead.

hhakimi53:
Bill, yes the information is available at AbstractParsedStatement about unique constraints defined on table  and also the expression on which equivalence filter is defined. Using this information logic to evaluate if the result will contain at most one row at AbstractParsedStatement level  can be reconstituted and ParsedSelectStatement can consume that information. I had started looking at AbstractParsedStatement when I had started with the issue though the logic to derive if the output row will be atmost one at AbstractParsedStatement gets little bit complicated specially for case where unique constraint is formed of composite keys and we have to verify for their presence in equivalence expression. 
Also there already is logic in the Plan nodes that evaluates if plan will result in order determinism for case where statement determinism is not guaranteed which compiled plan uses to check for determinism. Currently, for single fragment plan, for tables which have unique constraints defined and is used in search key, IndexScanPlanNode helps in determining determinism based on that  (as the statement-determinism based on AbstractParsedStatement  today is evaluated to be not guaranteed).  IndexScanPlanNode also has logic about unique constraints and search expression, which is used to estimate tuple count from the current plan and to estimate cost. This, I had discovered after Chris had suggested to explore if we can drive the determinism information at plan node instead of adding logic AbstractParsedStatement. So the current logic uses most of the pieces available in IndexScanPlanNode to determine if it will result in at most one output row. Though yes, determinism check has been expanded to sub plan graph for two fragment plan. If this approach does not seem appropriate, I can start looking into how to figure order determine based on unique constraint equivalence at AbstractParsedStatement. Thanks.

paulmartel:
I largely agree with Bill here. 

There IS an edge case involving plans with (potentially) multi-row results where the statement does not guarantee order determinism and yet order determinism CAN BE guaranteed by the chosen plan (possibly at worst after a trivial micro-optimization for this very purpose). In spite of possible precedent, I don't see that this justifies plan-based determinism checks that fall outside this edge case.

In particular, the case of a statement with one output row seems to fall well outside this set of edge cases.

If the statement-based determinism checks happen to uncover/produce artifacts (like constraints) that are also useful to the planning process, then by all means cache them on the statement for use in the planner phase. That should be enough to eliminate truly redundant logic between the statement processing executed once and the plan processing potentially executed in multiple permutated iterations.

hhakimi53:
Closing pull request. Will create another with pk equivalence logic defined at parsed statement level.

PR #3490, opened 2016-04-18
ENG-10230: Protect MP sysprocs from stale, unmatched fragment respons…

wweiss-voltdb:
…e messages during restart

### DIFF
--- a/src/frontend/org/voltdb/iv2/FragmentTask.java
+++ b/src/frontend/org/voltdb/iv2/FragmentTask.java
@@ -213,8 +213,7 @@ public class FragmentTask extends TransactionTask
 
         if (m_fragmentMsg.isEmptyForRestart()) {
             int outputDepId = m_fragmentMsg.getOutputDepId(0);
-            currentFragResponse.addDependency(outputDepId,
-                    new VoltTable(new ColumnInfo[] {new ColumnInfo("UNUSED", VoltType.INTEGER)}, 1));
+            currentFragResponse.addDependency(outputDepId, m_dummyResultTable);
             return currentFragResponse;
         }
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/iv2/FragmentTask.java
```diff
@@ -282,15 +282,30 @@ public FragmentResponseMessage processFragmentTask(SiteProcedureConnection siteC
             } catch (final EEException e) {
                 hostLog.l7dlog( Level.TRACE, LogKeys.host_ExecutionSite_ExceptionExecutingPF.name(), new Object[] { Encoder.hexEncode(planHash) }, e);
                 currentFragResponse.setStatus(FragmentResponseMessage.UNEXPECTED_ERROR, e);
+                if (currentFragResponse.getTableCount() == 0) {
+                    // Make sure the response has at least 1 result with a valid DependencyId
+                    currentFragResponse.addDependency(outputDepId,
+                            new VoltTable(new ColumnInfo[] {new ColumnInfo("UNUSED", VoltType.INTEGER)}, 1));
```
nshi: Maybe you can cache this dummy table as a constant so that you don't have to create a new one every time.

wweiss-voltdb: Yes. I suppose we could allocate a static. Good Idea.

nshi: I think I made a bad suggestion. I take that back. It's a bad idea to create a constant table because `VoltTable` is not thread-safe. Using that same constant in different threads will result in weird errors.

PR #3492, opened 2016-04-18
update partition guard message

hhakimi53:
Partitioning of table on varbinary type is permitted though the guard message does not indicate it (verfied murmur hash in EE on varbinary type is supported). A minor change updating partition-by guard message.

.................................
5> create table t (digest varbinary(128) NOT Null, pt geography_point NOT Null);
Command succeeded.
6> partition table T on column pt;
[null]: In database, Partition column 'T.pt' is not a valid type. Partition columns must be an integer, varchar or varbinary type.
7> partition table T on column digest;
Command succeeded.
.............................


paulmartel:
Looks good.

PR #3495, opened 2016-04-19
ENG-10230: Protect MP sysprocs from stale, unmatched fragment respons…

wweiss-voltdb:
…e messages during restart


nshi:
Review finished.

PR #3501, opened 2016-04-21
ENG-10287, fix a bug that causing dr reset crashes EE.

luyangco:
Change-Id: I99ba2ea48e9cfc2ae696f8df1695bff952e874be


nshi:
Review finished.

PR #3502, opened 2016-04-21
Eng 10073 - determinism updates for PK equivalence

hhakimi53:
Logic updates to evaluate the SQL statement determinism based on value equivalence filters in the statement and unique indexes defined on the table. If the statement has parameterized/constant equivalence filters and the equivalence filter covers the unique indexes defined on the table such that it results in at most one output row, the query result is order deterministic irrespective if it result in single or two fragment plan. 

For eg "Select * from T where a = ?;" 
where a is unique constrained defined on the table T will result in atmost one output row. 

Today though similar statement gets flagged as non-deterministic for two fragment plan. Logic updates with this pull request addresses this issue by extending determinism logic in abstract parsed statement to also check if the statement will produce at most one output row based on value equivalence filters in the statement and unique indexes defined on the table.

### DIFF
--- a/src/frontend/org/voltdb/planner/AbstractParsedStmt.java
+++ b/src/frontend/org/voltdb/planner/AbstractParsedStmt.java
@@ -1838,7 +1838,7 @@ public abstract class AbstractParsedStmt {
         HashMap<AbstractExpression, Set<AbstractExpression>> valueEquivalence = analyzeValueEquivalence();
 
         // If no value equivalence filter defined in SQL statement, there's no use to continue
-        if (valueEquivalence == null || valueEquivalence.isEmpty()) {
+        if (valueEquivalence.isEmpty()) {
             return false;
         }
 
@@ -1863,21 +1863,24 @@ public abstract class AbstractParsedStmt {
         // Iterate over the unique indexes defined on the table to check if the unique
         // index defined on table appears in tve equivalence expression gathered above.
         for (Index index : indexes) {
-            Set<AbstractExpression> indexExpressions = new HashSet<AbstractExpression>();
-            // Fetch pure column indices which are unique
-            if (index != null && index.getUnique() && index.getExpressionsjson().isEmpty()) {
-                CatalogMap<ColumnRef> indexColRefs = index.getColumns();
-                for (ColumnRef indexColRef:indexColRefs) {
-                    Column col = indexColRef.getColumn();
-                    TupleValueExpression tve = new TupleValueExpression(scan.getTableName(),
-                                                                        scan.getTableAlias(),
-                                                                        col.getName(),
-                                                                        col.getName(),
-                                                                        col.getIndex());
-                    indexExpressions.add(tve);
-                }
+            // Perform lookup only on pure column indices which are unique
+            if (!index.getUnique() || !index.getExpressionsjson().isEmpty()) {
+                continue;
             }
-            if (!indexExpressions.isEmpty() && parameterizedConstantKeys.containsAll(indexExpressions)) {
+
+            Set<AbstractExpression> indexExpressions = new HashSet<AbstractExpression>();
+            CatalogMap<ColumnRef> indexColRefs = index.getColumns();
+            for (ColumnRef indexColRef:indexColRefs) {
+                Column col = indexColRef.getColumn();
+                TupleValueExpression tve = new TupleValueExpression(scan.getTableName(),
+                                                                    scan.getTableAlias(),
+                                                                    col.getName(),
+                                                                    col.getName(),
+                                                                    col.getIndex());
+                indexExpressions.add(tve);
+            }
+
+            if (parameterizedConstantKeys.containsAll(indexExpressions)) {
                 return true;
             }
         }
--- a/src/frontend/org/voltdb/planner/AbstractParsedStmt.java
+++ b/src/frontend/org/voltdb/planner/AbstractParsedStmt.java
@@ -1808,11 +1808,13 @@ public abstract class AbstractParsedStmt {
         return m_contentDeterminismMessage != null;
     }
 
-    // Functional evaluates whether the statement will result in at most
-    // one output tuple. This is implemented for single table by checking
-    // value equivalence of predicates in where clause and using information
-    // if all defined unique indexes are in value equivalence set
-    protected boolean producesOneRowOuputUsingValueEquivalence () {
+    // Function evaluates whether the statement results in at most
+    // one output row. This is implemented for single table by checking
+    // value equivalence of predicates in where clause and checking
+    // if all defined unique indexes are in value equivalence set.
+    // Returns true if the statement results is at most one output
+    // row else false
+    protected boolean producesOneRowOutput () {
         if (m_tableAliasMap.size() != 1) {
             return false;
         }
@@ -1835,13 +1837,12 @@ public abstract class AbstractParsedStmt {
         // Collect value equivalence expression for the SQL statement
         HashMap<AbstractExpression, Set<AbstractExpression>> valueEquivalence = analyzeValueEquivalence();
 
-        // If no value equivalence filter defined in SQL statement,
-        // there's no use to continue
+        // If no value equivalence filter defined in SQL statement, there's no use to continue
         if (valueEquivalence == null || valueEquivalence.isEmpty()) {
             return false;
         }
 
-        // Collect all tve keys from value equivalence exp which have equivalence
+        // Collect all tve expressions from value equivalence set which have equivalence
         // defined to parameterized or constant value expression.
         // Eg: T.A = ? or T.A = 1
         Set <AbstractExpression> parameterizedConstantKeys = new HashSet<AbstractExpression>();
@@ -1859,8 +1860,8 @@ public abstract class AbstractParsedStmt {
             }
         }
 
-        // Iterate over the unique indexes defined on the table to check in the
-        // unique index defined on table appears in list of equivalence expr above
+        // Iterate over the unique indexes defined on the table to check if the unique
+        // index defined on table appears in tve equivalence expression gathered above.
         for (Index index : indexes) {
             Set<AbstractExpression> indexExpressions = new HashSet<AbstractExpression>();
             // Fetch pure column indices which are unique
@@ -1876,11 +1877,9 @@ public abstract class AbstractParsedStmt {
                     indexExpressions.add(tve);
                 }
             }
-
             if (!indexExpressions.isEmpty() && parameterizedConstantKeys.containsAll(indexExpressions)) {
                 return true;
             }
-
         }
         return false;
     }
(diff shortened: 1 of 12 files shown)


tests/frontend/org/voltdb/compiler/TestProcCompiler.java
```diff
@@ -26,10 +26,10 @@
 import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
 
-import junit.framework.TestCase;
-
 import org.voltdb.VoltDB.Configuration;
 
+import junit.framework.TestCase;
+
```
paulmartel: Minor: this import re-org seems like a regression against the alphabetical algorithm used by eclipse?

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -612,10 +612,87 @@ public void testFloatingAggs() throws Exception {
...
-            DeterminismMode detMode)
-    {
-        CompiledPlan cp = compileAdHocPlan(sql, detMode);
+            boolean forcePartitioning, DeterminismMode detMode) {
+
+        boolean inferPartitioning = true; // by default, use the in
+        boolean forceSP = false;
+        if (forcePartitioning) {
+            inferPartitioning = false;
+        }
```
paulmartel: Setting a local value once is clearer than setting to default and then resetting.
Suggest: `boolean inferPartitioning = ! forcePartitioning;`
or maybe better to simplify even further: change the sense of the input argument in the caller(s) and just have them pass inferPartitioning in instead of forcePartitioning.

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -612,10 +612,87 @@ public void testFloatingAggs() throws Exception {
...
+        // different column than in where clause predicate.
+        sql = "select b from punique where a = ?";
+        assertMPPlanDeterminismCore(sql, ORDERED, CONSISTENT, true, DeterminismMode.FASTER);
+        assertMPPlanDeterminismCore(sql, ORDERED, CONSISTENT, true, DeterminismMode.SAFER);
+
+        // where clause using equivalence filter on table using all columns
+        // forming composite primary
+        // key as this select statement can be in insert clause too, test it for
+        // safer determinism also
+        sql = "select z from ppkcombo where a = 1 AND b = 2 AND c =3;";
```
paulmartel: readability: add space before3

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -612,10 +612,87 @@ public void testFloatingAggs() throws Exception {
...
+        assertMPPlanDeterminismCore(sql, ORDERED, CONSISTENT, true, DeterminismMode.SAFER);
+
+        // using parameters
+        sql = "select z from ppkcombo where a = ? AND b = ? AND c = ?;";
+        assertMPPlanDeterminismCore(sql, ORDERED, CONSISTENT, true, DeterminismMode.FASTER);
+        assertMPPlanDeterminismCore(sql, ORDERED, CONSISTENT, true, DeterminismMode.SAFER);
+
+        // predicate containing value-equivalence through transitive behavior that form unique
+        // index on table
+        sql = "select c, z from ppkcombo where a = ? AND a = b AND c = B;";
```
paulmartel: readability, lowercasE 'b' for consistency.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1807,4 +1807,80 @@ protected void updateContentDeterminismMessage(String msg) {
...
+        if (indexes == null || indexes.size() == 0) {
+            // no indexes defined on the table
+            return false;
+        }
+
+        // Collect value equivalence expression for the SQL statement
+        HashMap<AbstractExpression, Set<AbstractExpression>> valueEquivalence = analyzeValueEquivalence();
+
+        // If no value equivalence filter defined in SQL statement, there's no use to continue
+        if (valueEquivalence == null || valueEquivalence.isEmpty()) {
```
paulmartel: Simplify: function never returns null, no need to check for that.

tests/frontend/org/voltdb/compiler/TestProcCompiler.java
hhakimi53: Import re-org was performed by eclipse. May be my eclipse environment settings may be different. I will check on that.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1807,4 +1807,80 @@ protected void updateContentDeterminismMessage(String msg) {
...
+                }
+            }
+        }
+
+        // Iterate over the unique indexes defined on the table to check if the unique
+        // index defined on table appears in tve equivalence expression gathered above.
+        for (Index index : indexes) {
+            Set<AbstractExpression> indexExpressions = new HashSet<AbstractExpression>();
+            // Fetch pure column indices which are unique
+            if (index != null && index.getUnique() && index.getExpressionsjson().isEmpty()) {
```
paulmartel: no need to check index != null. The catalog does not contain null entries.
Also, avoid extra code block nesting w/

    if (!index.getUnique()) {
        continue;
    }
    if (!index.getExpressionsjson().isEmpty()) {
        continue;
    }
Also move indexExpressions declaration/initialization to below these checks -- when it is actually needed.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1807,4 +1807,80 @@ protected void updateContentDeterminismMessage(String msg) {
...
+                    Column col = indexColRef.getColumn();
+                    TupleValueExpression tve = new TupleValueExpression(scan.getTableName(),
+                                                                        scan.getTableAlias(),
+                                                                        col.getName(),
+                                                                        col.getName(),
+                                                                        col.getIndex());
+                    indexExpressions.add(tve);
+                }
+            }
+            if (!indexExpressions.isEmpty() && parameterizedConstantKeys.containsAll(indexExpressions)) {
```
paulmartel: Simplify: only do this check when we have found a suitable index -- and in that case, we know that indexExpressions.isEmpty() is false -- the index will have at least one tve column, so you don't have to check that.

paulmartel:
End of review round. This change looks correct for the single-table-query case, with some good test coverage and code cleanup.  Should we consider handling the similarly constrained multi-table case in this ticket -- if not, we should make sure that there is an open ticket for that and it would be nice to add one or more multi-table test cases that show the symptoms of the remaining bug we hope some day to fix -- these could validate that the issue still remains for some multi-table cases and can cross-reference the ticket in a comment like `// Change this test to expect determinism when ENG-XXXXX is fixed.`

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
hhakimi53: I will update it.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
hhakimi53: Yes, that will actually optimize how logic will be looping. I will update it, thnxs.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
hhakimi53: sure.

tests/frontend/org/voltdb/planner/TestDeterminism.java
hhakimi53: missed this one, I will update it. thnxs

tests/frontend/org/voltdb/planner/TestDeterminism.java
hhakimi53: sure, will change it 'b', that way it will be consistent.

tests/frontend/org/voltdb/planner/TestDeterminism.java
hhakimi53: thnxs for suggestion, I will look into it.

hhakimi53:
Thanks Paul for reviewing the changes and the feedback. For multi-table case, I think we should create another ticket and handle that issue in there. I can create the ticket and cross reference the tests with ticket number.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1807,4 +1807,83 @@ protected void updateContentDeterminismMessage(String msg) {
...
     }
+
+    // Function evaluates whether the statement results in at most
+    // one output row. This is implemented for single table by checking
+    // value equivalence of predicates in where clause and checking
+    // if all defined unique indexes are in value equivalence set.
+    // Returns true if the statement results is at most one output
+    // row else false
+    protected boolean producesOneRowOutput () {
+        if (m_tableAliasMap.size() != 1) {
```
BillWhite: I still don't understand why this and addHonoraryOrderByExpressions are not unified.  It seems like a lot of this code came from there, and this is computing part of the same thing that function is computing.

BillWhite:
That's all.  Just the one question.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
hhakimi53: I had looked into addHonoraryOrderByExpressions() call also. addHonoraryOrderByExpressions is dealing with orderby expressions and  the processed display columns in select statement, with end result an expanded order-by expression list. Also this call is nested two-three level deep in call hierarchy and is consumed by functionality processing order-by expressions. The current producesOneRowOutput() is specific to equivalence filter on indexes defined, with end result an answer whether it will produce one row or not. It's not operating on any display column lists and even if I had went the way to overload addHonoraryOrderByExpressions() to include index equivalence in order-by list, we would have to adjust order-by logic of how it processes value equivalence expressions (currently in addHonoraryXXX()  order-by filter is based on display column list), this would have made logic in addHonoraryOrderByExpressions complicated. So I did not  overload order-by expression list addHonoraryOrderByExpressions() and kept the functionality separate it's operating on index value-equivalence specifically. Also there is already a call for checking the same condition, so logically it seemed reasonable to hook in functionality there.

wolffcm:
This looks good to me.  

Do we need a separate ticket for refactoring some of this?

hhakimi53:
Thanks Chris reviewing the changes. 

Sorry, I didn't understand question about separate ticket for refactoring some of this? Can you please elaborate, thanks.

wolffcm:
I just wonder if there are places where similar code that does the same thing could be merged.  I could be wrong.

hhakimi53:
Oh ok. If it's with respect to addHonoraryXXX(), in addHonoraryXXX() the logic on equivalence set and constraints defined on table operates in context of display columns in select statement to expand order-by expression. For producesOneRowOutput(), logic iterates over equivalence in context of indexes defined on table and is independent of display columns in select list and order-by list. I had talked to Bill also about it. So given how the flow is structured there is not much we can abstract between these functions as they are operating on different contexts.

PR #3506, opened 2016-04-21
Execute an EE fragment in a unit test.

BillWhite:
This adds an example of a unit test function which executes a fragment
of a plan.  There was an example of this already, but it was not
completely fleshed out.  We simply create an object of a subclass of
ExecutionEngine, another of a subclass of Topend, and join them
together.  With some careful buffer management we can execute a fragment
and arrange for the output table to be accessible.  We can then validate
the output for correctness.

ENG-10022


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -335,7 +335,33 @@ void VoltDBEngine::serializeTable(int32_t tableId, SerializeOutput& out) const
 // ------------------------------------------------------------------
 // EXECUTION FUNCTIONS
 // ------------------------------------------------------------------
-
+/**
```
wolffcm: Thank you for adding this comment!

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+        /*
+         * Initialize the engine.  We create our own
+         * topend, to make sure we can supply fragments
+         * by id, and then make sure we know where the
+         * shared buffers are.  Note that calling setBuffers
+         * sets the shared buffer pointers, and calling
+         * resetReusedResultOutputBuffer causes the engine to
+         * use them.
+         */
+        m_topend = new EngineTestTopend();
```
wolffcm: You could use boost::scoped_ptr for both of these, and avoid having to do anything in the destructor.

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
         CatalogId database_id;
         CatalogId site_id;
-        VoltDBEngine *engine;
+        VoltDBEngine *m_engine;
         string catalog_string;
         catalog::Catalog *catalog; //This is not the real catalog that the VoltDBEngine uses. It is a duplicate made locally to get GUIDs
         catalog::Cluster *cluster;
         catalog::Database *database;
         catalog::Constraint *constraint;
+        EngineTestTopend *m_topend;
```
wolffcm: Some of the members you've added have the `m_` prefix and some do not.  I'm okay with not adding the prefix to all the existing members, but for new members, maybe use it consistently?

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+// 2. When the test runs there will be a folder called debugoutput
+//    wherever the test is run.  In the Eclipse method above this will
+//    be the project folder, which is the root of the source folder.
+//    There you will find a file with name something like
+//           debugoutput/statement_plans/ENG-10022-stmt-0_json.txt.
+//    The contents of this file should be this string.
+// 3. Some emacs cleanup will be required to paste this into
+//    the source code as a string.  Escape double quotes and add
+//    initial double quotes, '\n' lines and terminal double quotes
+//    as usual.
```
wolffcm: Can you include the text of the SQL statement used to create this plan in this comment?

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+    // Execute the plan.  You'd think this would be more
+    // impressive.
+    //
+    m_engine->executePlanFragments(1, &fragmentId, NULL, emptyParams, 1000, 1000, 1000, 1000, 1);
+
+    // Fetch the results.  We have forced them to be written
+    // to our own buffer in the local engine.  But we don't
+    // know how much of the buffer is actually used.  So we
+    // need to query the engine.
+    size_t result_size = m_engine->getResultsSize();
```
wolffcm: You use underbars in some places and camel case in others.  Maybe standardize on camel case in C++?

We don't have much by way of coding standards, but there is this:

https://github.com/VoltDB/voltdb/wiki/Code-conventions-&-licensing

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+    size_t row_count = result.readInt();
+    int64_t long_value;
+    int32_t int_value;
+    int16_t short_value;
+    int8_t  byte_value;
+    double   double_value;
+    std::string string_value;
+    ByteArray varbinary_value;
+    uint64_t column_values[100];
+    for (int ridx = 0; ridx < row_count; ridx += 1) {
```
wolffcm: This seems like an awful lot of code.  Is there any way you can get the `voltdb::Table*` that is the output table of the `SendPlanNode`?

If you have the `Table*` you've got access to a high-level data structure that would do all of this for you.

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+    ReferenceSerializeInputBE result(m_resultBuffer.get(), result_size);
+    if (debug_print) {
+        printf("Result Table:\n");
+    }
+    size_t msg_len = result.readInt();
+    uint8_t status = result.readByte();
+    if (debug_print) {
+        printf("  message length = %lu, status = %hhu\n", msg_len, status);
+    }
+    size_t intercostal_clavicle = result.readInt();
```
wolffcm: Inter-whatnow?  What is this really?

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+        /*
+         * This is true because of collusion between the query
+         * and the test.  The query selects two values, both
+         * integral, and the second is twice the first.
+         */
+        ASSERT_TRUE(2*column_values[0] == column_values[1]);
+    }
 }
 
+#if       0
```
wolffcm: These stubs have been around since the repo was initialized.  Please just go ahead and delete them.

wolffcm:
Looks pretty good.  I feel like the validation of the output could be a whole lot simpler.  Maybe I'm missing something though.  Can you clarify?

tests/ee/execution/engine_test.cpp
BillWhite: It's the inter cluster latency time.  Since we don't have a cluster, it's irrelevant to us.

Apparently nobody watches screwball comedy anymore.

tests/ee/execution/engine_test.cpp
BillWhite: Well, if the one I used had not been there I would not have known where to start.  But I don't think these are contributing anything anymore, so I will delete them.

BillWhite:
I agree.  I couldn't see how to get the Table* of the output table.  But then I see you can get them from the send node, so I guess that will be easier than parsing out the serialized data.

tests/ee/execution/engine_test.cpp
BillWhite: True.

tests/ee/execution/engine_test.cpp
BillWhite: It's in the test function.  I didn't think it was all that helpful to replicate it here, since
it might very well become stale.  If the test function changes somehow the one in the
comment may not change along with it.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -335,7 +335,33 @@ void VoltDBEngine::serializeTable(int32_t tableId, SerializeOutput& out) const
...
+ * Execute the given fragments serially and in order.
+ * Return 0 if there are no failures and 1 if there are
+ * any failures.  Note that this is the meat of the JNI
+ * call org.voltdb.jni.ExecutionEngine.nativeExecutePlanFragments.
+ *
+ * @param numFragments          The number of fragments to execute.
+ *                              This is directly from the JNI call.
+ * @param planfragmentIds       The array of fragment ids.  This is
+ *                              This is indirectly from the JNI call,
```
hhakimi53: * @param planfragmentIds       The array of fragment ids.  This is
*                              This is indirectly from the JNI call,

I think you meant:
* @param planfragmentIds       The array of fragment ids.  This is
+ *                              from the JNI call,

src/ee/execution/VoltDBEngine.cpp
hhakimi53: +1

tests/ee/execution/engine_test.cpp
wolffcm: It would be a big help if we could give a schema and SQL statement to a standalone program and produce the JSON plan for the statement.

One thing that occurs to me: I think the class VoltCompiler was at one time intended to be invoked as a standalone program.  I believe that it still has a "main" method.  Whether or not that still works, or would be useful for generating plans without a full blown VoltDB instance is unclear to me.

This is good for now, I'm just thinking aloud.

tests/ee/execution/engine_test.cpp
```diff
@@ -92,209 +92,719 @@ ValueType COLUMN_TYPES[NUM_OF_COLUMNS]  = { VALUE_TYPE_INTEGER,
...
+         * topend, to make sure we can supply fragments
+         * by id, and then make sure we know where the
+         * shared buffers are.  Note that calling setBuffers
+         * sets the shared buffer pointers, and calling
+         * resetReusedResultOutputBuffer causes the engine to
+         * use them.
+         */
+        m_topend = new EngineTestTopend();
+        m_engine = new VoltDBEngine(m_topend);
+        m_parameterBuffer.reset(new char [4 * 1024]);
```
hhakimi53: Question: Are the 4KB and so on for the different buffers upper-bound limits specific to this test or to the engine?

hhakimi53:
Overall changes look fine.

src/ee/execution/VoltDBEngine.cpp
BillWhite: No, this is what I meant.  The JNI call passes a Java object, which the JNI interface must translate to be a C++ array.  So, this is indirectly from the JNI call, but not directly.

tests/ee/execution/engine_test.cpp
BillWhite: Most of the existing member data were camel case, two that I added were one word long, but the three multi-word variables I added were camel case.  Rather than change all the test to be camel case I change the three whose case is relevant to be snake case.

wolffcm:
Looks good to me.  Thanks for making the changes.

BillWhite:
Squash merged on another branch.

PR #3507, opened 2016-04-21
ENG-10219: Remove DR update/delete serialization optimization.

nshi:


src/ee/storage/BinaryLogSink.cpp
```diff
@@ -360,73 +328,11 @@ int64_t BinaryLogSink::apply(const char *taskParams, boost::unordered_map<int64_
...
-            IndexCursor indexCursor(index->getTupleSchema());
-            index->moveToKey(&tempTuple, indexCursor);
-            TableTuple deleteTuple = index->nextValueAtKey(indexCursor);
-            if (deleteTuple.isNullTuple()) {
-                throwSerializableEEException("Unable to find tuple for deletion: binary log type (%d), DR ID (%jd), unique ID (%jd), tuple %s\n",
-                                             type, (intmax_t)sequenceNumber, (intmax_t)uniqueId, tempTuple.debug(table->name()).c_str());
-            }
-
-            table->deleteTuple(deleteTuple, true);
+            throwSerializableEEException("Delete by index is not supported for DR");
```
wweiss-voltdb: So removing this from the Sink implies that there is no dr based upgrade path from a version that generates index optimization records.

PR #3516, opened 2016-04-22
ENG-10203: Don't start processing events for persisted generations un…

manjujames:
…til they become the

currently processed generation. Queue up events until the generation is activated.

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -950,6 +950,13 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
         if (m_executor==null) {
             synchronized (m_executorLock) {
                 if (m_executor==null) {
+                    // Bound the queue. It shouldn't get to this high value.
+                    // Log an error so that we know if it does get to the high value.
+                    if (m_queuedActions.size() > 50) {
+                        exportLog.error("Queue for export source for generation " + m_generation +
+                                " is going beyond 50. Not queueing anymore events");
+                        return Futures.immediateFuture(null);
+                    }
                     m_queuedActions.add(runnable);
                     return Futures.immediateFuture(null);
                 }
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -105,8 +105,7 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
     private final Semaphore m_allowAcceptingMastership = new Semaphore(0);
     private volatile boolean m_closed = false;
     private volatile boolean m_mastershipAccepted = false;
-    private volatile boolean m_isActive;
-    private AtomicReference<ListeningExecutorService> m_executor = new AtomicReference<>();
+    private volatile ListeningExecutorService m_executor;
     private Integer m_executorLock = new Integer(0);
     private LinkedTransferQueue<RunnableWithES> m_queuedActions = new LinkedTransferQueue<>();
 
@@ -302,7 +301,6 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
     }
 
     public void activate() {
-        m_isActive = true;
         setupExecutor();
     }
 
@@ -502,7 +500,7 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
                             new BBContainer(buffer) {
                                 @Override
                                 public void discard() {
-                                    final ByteBuffer buf = checkDoubleFree();
+                                    checkDoubleFree();
                                     cont.discard();
                                     deleted.set(true);
                                 }
@@ -544,10 +542,13 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
             Throwables.propagate(e);
         }
         ListeningExecutorService es = getExecutorService();
+        assert(es!=null); // tests will fail because we run with assertion enabled
         if (es==null) {
-            //TODO: Is this the right thing to do?
-            throw new RuntimeException("This generation must be active by the time pushExportBuffer is called");
+            exportLog.error(
+                "Push export buffer called before the generation is active. This push will be ignored");
+            return;
         }
+
         if (es.isShutdown()) {
            m_bufferPushPermits.release();
            return;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -295,7 +299,11 @@ public void run() {
         // compute the number of bytes necessary to hold one bit per
         // schema column
         m_nullArrayLength = ((m_columnTypes.size() + 7) & -8) >> 3;
-        m_es = CoreUtils.getListeningExecutorService("ExportDataSource gen " + m_generation + " table " + m_tableName + " partition " + m_partitionId, 1);
+    }
+
+    public void activate() {
+        m_isActive = true;
```
akhanzode: Shouldnt this be on executor lock?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -927,7 +945,63 @@ public void setRunEveryWhere(boolean runEveryWhere) {
         m_runEveryWhere = runEveryWhere;
     }
 
+    private ListenableFuture<?> stashOrSubmitTask(RunnableWithES runnable, boolean callExecute) {
+        if (!m_isActive) {
+            synchronized (m_executorLock) {
```
akhanzode: Then you will need to check only once if active or not.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -927,7 +945,63 @@ public void setRunEveryWhere(boolean runEveryWhere) {
...
+                if (m_queuedActions.size()>0) {
+                    for (RunnableWithES queuedR : m_queuedActions) {
+                        queuedR.setExecutorService(es);
+                        es.submit(queuedR);
+                    }
+                    m_queuedActions.clear();
+                }
+            }
+            if (!m_executor.compareAndSet(null, es)) {
+                es.shutdown();
```
akhanzode: Shouldnt compare and set be done after shutdown? your !es.isShutdown() will end up in race.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -927,7 +945,63 @@ public void setRunEveryWhere(boolean runEveryWhere) {
...
+    }
+
+    private abstract class RunnableWithES implements Runnable {
+        private ListeningExecutorService m_executorService;
+
+        public void setExecutorService(ListeningExecutorService executorService) {
+            m_executorService = executorService;
+        }
+
+        public ListeningExecutorService getLocalExecutorService() {
```
akhanzode: why getLocalExecutorService() its always local word local means something is remote also.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -927,7 +945,63 @@ public void setRunEveryWhere(boolean runEveryWhere) {
...
+            }
+        }
+    }
+
     public ListeningExecutorService getExecutorService() {
-        return m_es;
+        return m_executor.get();
+    }
+
+    private abstract class RunnableWithES implements Runnable {
```
akhanzode: Isnt this more like QueuedRunnable?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -927,7 +945,63 @@ public void setRunEveryWhere(boolean runEveryWhere) {
...
+    public void setupExecutor() {
+        if (m_executor.get()==null) {
+            ListeningExecutorService es = CoreUtils.getListeningExecutorService(
+                            "ExportDataSource gen " + m_generation
+                            + " table " + m_tableName + " partition " + m_partitionId, 1);
+            synchronized(m_executorLock) {
+                if (m_queuedActions.size()>0) {
+                    for (RunnableWithES queuedR : m_queuedActions) {
+                        queuedR.setExecutorService(es);
+                        es.submit(queuedR);
```
akhanzode: es could be shutdown unless you protect close() and closeAndDelete() with lock as well.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -893,20 +909,22 @@ public synchronized void acceptMastership() {
...
         }
+        exportLog.info("Accepting mastership for export generation " + getGeneration() + " Table " + getTableName() + " partition " + getPartitionId());
         m_mastershipAccepted = true;
-        m_es.execute(new Runnable() {
+        RunnableWithES runnable = new RunnableWithES() {
             @Override
             public void run() {
                 try {
-                    if (!m_es.isShutdown() || !m_closed) {
+                    if (!getLocalExecutorService().isShutdown() || !m_closed) {
```
akhanzode: possible NPE?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -615,62 +636,61 @@ public void run() {
...
+        RunnableWithES runnable = new RunnableWithES() {
             @Override
             public void run() {
                 try {
                     m_committedBuffers.close();
                 } catch (IOException e) {
                     exportLog.error(e);
                 } finally {
-                    m_es.shutdown();
+                    getLocalExecutorService().shutdown();
```
akhanzode: possible NPE is DS was never activated but closed()

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -615,62 +636,61 @@ public void run() {
     }
 
     public ListenableFuture<?> sync(final boolean nofsync) {
-        try {
-            return m_es.submit(new SyncRunnable(nofsync));
-        } catch (RejectedExecutionException e) {
-            exportLog.error("Error scheduling export buffer sync", e);
-        }
-        return null;
+        RunnableWithES runnable = new RunnableWithES() {
```
akhanzode: only active DS should do sync rest PBD is not being read or written so can be ignored as they are safe.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -529,16 +543,21 @@ public void pushExportBuffer(
         } catch (InterruptedException e) {
             Throwables.propagate(e);
         }
-        if (m_es.isShutdown()) {
+        ListeningExecutorService es = getExecutorService();
+        if (es==null) {
+            //TODO: Is this the right thing to do?
+            throw new RuntimeException("This generation must be active by the time pushExportBuffer is called");
```
akhanzode: I would say log and return.

src/frontend/org/voltdb/export/ExportDataSource.java
manjujames: This will never get executed unless the data source is activated.

src/frontend/org/voltdb/export/ExportDataSource.java
manjujames: No, executor service must always be set up. If not, there is something wrong in code.

src/frontend/org/voltdb/export/ExportDataSource.java
manjujames: I am purposely avoiding synchronized on every call, which is why this is being tested twice.

src/frontend/org/voltdb/export/ExportDataSource.java
manjujames: No, this shutdown is for executor service that was created, but we never used.

src/frontend/org/voltdb/export/ExportDataSource.java
manjujames: The only reason I need this is because we need access to the executor service inside the run. However, executor service is not known at the time we define run method.

src/frontend/org/voltdb/export/ExportDataSource.java
manjujames: There is a getExecutorService at the outer class level. I named this different so that there won't be any confusion. I could change it to something else...

src/frontend/org/voltdb/export/ExportDataSource.java
akhanzode: That does not make it full proof though.

src/frontend/org/voltdb/export/ExportDataSource.java
akhanzode: Can getExecutorService() be called before es shutdown in setupExecutor? if yes there are many NPE situations.

manjujames: Let's talk in person tomorrow

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -104,6 +105,10 @@
     private final Semaphore m_allowAcceptingMastership = new Semaphore(0);
     private volatile boolean m_closed = false;
     private volatile boolean m_mastershipAccepted = false;
+    private volatile boolean m_isActive;
```
vtkstef: you may want to consider AtomicBoolean

src/frontend/org/voltdb/export/ExportDataSource.java
vtkstef: with AtomicBoolean you only set it once compareAndSet(false,true)

akhanzode:
Did you measure the Queue length? What was your thought around bounded singleton queued tasks?

manjujames:
squash merged.

PR #3526, opened 2016-04-26
Eng 10073 vmc upsert test

johnTcrawford:
Fixed the UPSERT test in sqlQueries.txt that has been failing since ENG-10073 was fixed; and added a few similar tests to confirm which cases do and still do not work (and should not). Also fixed unrelated problems with the SqlQueriesTest.checkUserStoredProcs test failing, due to new stored procedures showing up in the list (which I've now added, see ENG-10137), and due to new stream T25N (I fixed FullDdlSqlTest.groovy to now drop any streams that are added via a CREATE STREAM statement).


paulmartel:
Looks correct.

PR #3528, opened 2016-04-27
Refactor VoltType in advance of ENG-9873 type checking improvements.

paulmartel:
VoltType-related refactoring to:
   migrate some static methods to more relevant classes.
   use the "smart enum" feature of java 7 to special-case behavior of specific VoltTypes and reduce reliance on switch statements (should make adding new types easier) -- the new model follows a kind of 80/20 rule -- VoltType provides a simple "usual" implementation of each function which only a FEW specific VoltTypes will need to override to implement special case behavior. This is more efficient and maintainable than lots of switch statements.
    better distinguish fixed and variable length types
    avoid boilerplate especially in initialization
    improve javadoc coverage

### DIFF
--- a/src/frontend/org/voltdb/TheHashinator.java
+++ b/src/frontend/org/voltdb/TheHashinator.java
@@ -20,7 +20,6 @@ package org.voltdb;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
@@ -32,7 +31,6 @@ import org.voltcore.utils.Pair;
 import org.voltdb.dtxn.UndoAction;
 import org.voltdb.sysprocs.saverestore.HashinatorSnapshotData;
 
-import com.google_voltpatches.common.base.Charsets;
 import com.google_voltpatches.common.base.Supplier;
 import com.google_voltpatches.common.base.Suppliers;
 import com.google_voltpatches.common.base.Throwables;
@@ -167,7 +165,10 @@ public abstract class TheHashinator {
     abstract public Pair<Integer, Integer> pPredecessor(int partition, int token);
     abstract public Map<Integer, Integer> pGetRanges(int partition);
     public abstract HashinatorType getConfigurationType();
-    abstract public int pHashToPartition(VoltType type, Object obj);
+    // This method is only called within this class,
+    // {@see getHashedPartitionForParameter}
+    // but it must be declared protected to allow the required overriding.
+    abstract protected int pHashToPartition(VoltType type, Object obj);
     abstract protected Set<Integer> pGetPartitions();
     abstract public int getPartitionFromHashedToken(int hashedToken);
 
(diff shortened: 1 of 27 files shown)


src/frontend/org/voltdb/TheHashinator.java
```diff
@@ -352,24 +276,25 @@ public int getHashedPartitionForParameter(VoltType partitionParamType, Object pa
         // requiring the loader to know precise the schema.
```
BillWhite: This is not your change, but should this be "requiring the loader to know the schema *precisely*?
I think you need an adverb to modify "to know".

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -232,7 +232,8 @@ public ColumnInfo(String name, VoltType type) {
         public ColumnInfo clone() {
             try {
                 return (ColumnInfo) super.clone();
-            } catch (CloneNotSupportedException e) {
+            }
+            catch (CloneNotSupportedException e) {
                 assert(false);
```
BillWhite: Perhaps we should make a code formatting rule and make Eclipse and Emacs configurations to enforce them.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -269,11 +270,21 @@ public boolean equals(Object obj) {
...
+            }
+            if (defaultValue != other.defaultValue) {
+                return false;
+            }
+            if (size != other.size) {
+                return false;
+            }
+            if (type != other.type) {
+                return false;
+            }
```
BillWhite: Is this easier to read if it's a big "or" of conditions?  I think so, but opinions may vary.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -945,7 +949,8 @@ private void addColumnValue(Object value, VoltType columnType, int col) {
                     if (value instanceof java.util.Date ||
                             value instanceof TimestampType) {
                          micros = ParameterSet.timestampToMicroseconds(value);
-                    } else {
+                    }
```
BillWhite: This is just code formatting, but it seems weird to me.  Why is there a { at the end of the else line, but not before the else line?  I don't think I've ever seen that rule.

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -1653,7 +1690,9 @@ public static VoltTable fromJSONObject(JSONObject json) throws JSONException, IO
      */
     public boolean hasSameContents(VoltTable other) {
         assert(verifyTableInvariants());
-        if (this == other) return true;
+        if (this == other) {
+            return true;
```
BillWhite: Any why is this not
  if (this == other)
  {
    return true;
  }

BillWhite:
This seems alright to me.  The reindentation choices seem odd, but not terribly odd.

It might be nice to have a theory of operations for VoltTypes.  By this I mean some description of what is the 80 and what is the 20.  Maybe it's there and I just missed it.

hhakimi53:
In VoltType.java, should the VOLT_TYPE_MAX_ENUM be defined as 27? Currently it's defined as 30.

Overall changes look fine to me.

src/frontend/org/voltdb/VoltTable.java
paulmartel: In actual practice, we normally do not put `{` on its own line -- especially for the common if/else/while/switch statements. There tends to be more variation for the opening `{` of a class or method. I am somewhat proactively trying to move the code in that direction, in spite of there being no official coding standard (yet) that says this is preferred.

In contrast, '}' on its own line is a whole different question. I have been on the fence about it, but Chris finds that `}` on its own line before an `else` makes for a more readable alignment of the `if` and its `else`, so I have been recently going along with that and extending it to `catch` and `finally` as well. At some point, I may propose these as enhancements to our coding standards -- possibly even if I can only get them adopted by the SQL team.

PR #3532, opened 2016-04-28
ENG-10064: Kafka importer UAC system test scenario fails with short row count

JadenW:
1. tweak to backpressure In InternalConnectionHandler to be less aggressive;
2. In kafka Importer,  wait for voltdb if observe large gap;
3. refine the KafkaImportBenchmark


vtkstef:
Reviewed

PR #3535, opened 2016-04-28
ENG-10229: Break replication instead of crash on unexpected exceptions from the callbacks

slowsoul:
- Upon unexpected exceptions from the StateMachineInstance derived classes' callbacks, remove current instance from the community, initialize the instance with a different member id and triggers predefined reset actions to be executed.
- For DR Producer state machine, the reset action is to break replication so that the conversation on the victim producer node will be reset.
- If the replication mode is active-passive and sync snapshot provided by the victim producer node is still in progress, consumer cluster nodes will go down. In other cases, consumer cluster nodes will remain in RECEIVE state while keep failing to subscribe to the victim producer node because of conversation mismatch.
- In either case, subsystems other than DR in the producer cluster will remain functional.

Tests:
- Added tests in TestStateMachine that inject exception to different callbacks and make sure the state machine instance is gracefully reset.

### DIFF
--- a/src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
+++ b/src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -93,7 +94,14 @@ public class SynchronizedStatesManager {
     private final String m_ssmRootNode;
     private final String m_stateMachineRoot;
     private final String m_stateMachineMemberPath;
-    private final String m_memberId;
+    private String m_memberId;
+    private final String m_canonical_memberId;
+    private int m_resetCounter;
+    private int m_resetLimit;
+    private final int m_resetAllowance; // number of resets allowed within a reset clear threshold duration
+    private long m_lastResetTimeInMillis = -1L;
+
+    private static final long RESET_CLEAR_THRESHOLD = TimeUnit.DAYS.toMillis(1);
 
     private enum REQUEST_TYPE {
         INITIALIZING,
@@ -175,15 +183,16 @@ public class SynchronizedStatesManager {
      */
 
     public abstract class StateMachineInstance {
-        protected final String m_stateMachineId;
+        protected String m_stateMachineId;
+        private final String m_stateMachineName;
         private Set<String> m_knownMembers;
         private boolean m_membershipChangePending = false;
         private final String m_statePath;
         private final String m_lockPath;
         private final String m_barrierResultsPath;
-        private final String m_myResultPath;
+        private String m_myResultPath;
         private final String m_barrierParticipantsPath;
-        private final String m_myPartiticpantsPath;
+        private String m_myParticipantPath;
         private boolean m_stateChangeInitiator = false;
         private String m_ourDistributedLockName = null;
         private String m_lockWaitingOn = null;
@@ -196,6 +205,18 @@ public class SynchronizedStatesManager {
         private int m_currentParticipants = 0;
         private Set<String> m_memberResults = null;
         private int m_lastProposalVersion = 0;
+
+        private boolean m_initializationCompleted = false;
+
+        private boolean isInitializationCompleted() {
+            lockLocalState();
+            unlockLocalState();
+            return m_initializationCompleted;
+        }
+
+        public int getResetCounter() {
+            return m_resetCounter;
+        }
 
         private final Lock m_mutex = new ReentrantLock();
         private int m_mutexLockedCnt = 0;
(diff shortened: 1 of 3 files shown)


nshi:
|`src/frontend/org/voltcore/zk/SynchronizedStatesManager.java:350`|
|---|
Why not use `ZKUtil.joinZKPath()` to assemble the path here?

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
```diff
@@ -383,7 +404,12 @@ private void initializeStateMachine(Set<String> knownMembers) throws KeeperExcep
                 cancelDistributedLock();
                 checkForBarrierParticipantsChange();
                 // Notify the derived object that we have a stable state
-                setInitialState(readOnlyResult);
+                try {
+                    setInitialState(readOnlyResult);
+                } catch (Exception e) {
+                    org.voltdb.VoltDB.crashLocalVoltDB(
```
nshi: Why crash?

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
```diff
@@ -776,7 +853,12 @@ private void processResultQuorum(Set<String> memberList) {
                     m_log.info(m_stateMachineId + ": Initialized (concensus) with State " +
                             stateToString(m_synchronizedState.asReadOnlyBuffer()));
                     unlockLocalState();
-                    setInitialState(readOnlyResult);
+                    try {
+                        setInitialState(readOnlyResult);
+                    } catch (Exception e) {
+                        org.voltdb.VoltDB.crashLocalVoltDB(
```
nshi: Same here.

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
```diff
@@ -976,14 +1082,27 @@ private void initializeFromActiveCommunity() {
...
-                        "Unexepected failure in StateMachine.", true, e);
+                        "Unexpected failure in StateMachine.", true, e);
             }
             if (readOnlyResult != null) {
                 // Notify the derived object that we have a stable state
-                setInitialState(readOnlyResult);
+                try {
+                    setInitialState(readOnlyResult);
+                } catch (Exception e) {
+                    org.voltdb.VoltDB.crashLocalVoltDB(
```
nshi: And here.

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
```diff
@@ -1548,27 +1698,23 @@ public void ShutdownSynchronizedStatesManager() throws InterruptedException {
...
-                m_done.set(true);
             } catch (KeeperException.ConnectionLossException e) {
                 // lost the full connection. some test cases do this...
                 // means shutdown without the elector being
                 // shutdown; ignore.
-                m_done.set(true);
             } catch (InterruptedException e) {
-                m_done.set(true);
             } catch (Exception e) {
                 org.voltdb.VoltDB.crashLocalVoltDB(
```
nshi: Can the callback `stateMachine.disableMembership()` throw?

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
slowsoul: The reason was that setInitialState() will be called every time the instance is reinitialized, causing it to be removed and reinitialized repeatedly. But as we have a reset limit default to 5 now, when the limit is reached, no reinitialization will be attempted again and this node will disappear from other nodes forever. So yes, we should treat it like other callbacks. I will fix these 3 places.

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
slowsoul: KeeperException is catched inside, InterruptedException is catched outside, so the only other exception I can think of is that m_zk is NULL in this function.

slowsoul: m_zk is actually the reference to the ZK instance in HostMessenger in the VoltDB instance, so if it is NULL somehow, crashLocalVoltDB() seens to be the right choice.

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
```diff
@@ -535,10 +607,26 @@ private void checkForBarrierParticipantsChange() {
...
+                                            m_log.debug("Error in StateMachineInstance callbacks.", e);
+                                        }
+                                        m_deregistered = true;
+                                        m_shared_es.submit(new CallbackExceptionHandler(this));
+                                    }
                                 }
                                 else {
-                                    taskRequested(proposedState);
+                                    try {
+                                        taskRequested(proposedState);
```
wweiss-voltdb: Perhaps you should wrap this in a catch block. I am not sure whether you want to fail outright at this point or ignore the exception and continue as if nothing has happened. In either case I don't think you want the executor to catch this exception.

src/frontend/org/voltcore/zk/SynchronizedStatesManager.java
```diff
@@ -1676,6 +1847,37 @@ private synchronized void registerStateMachine(StateMachineInstance machine) thr
         }
     }
 
+    private class CallbackExceptionHandler implements Runnable {
+        StateMachineInstance m_directVictim;
```
slowsoul: Make it final

tests/frontend/org/voltcore/zk/TestStateMachine.java
```diff
@@ -521,7 +697,41 @@ protected String taskToString(ByteBuffer task)
...
+            correlatedTask = true;
+
+            notifiedOfReset = true;
+
+            return msm_states[0]; // 100 as reset state
+        }
+    }
+
+    class BrokenByteStateMachine extends ByteStateMachine {
+
```
wweiss-voltdb: It would be less errorprone to declare the brokenCallbackName variable inside the BrokenStateMachine class

PR #3536, opened 2016-04-28
ENG-10289: temporarily disable testAdHocLengthLimit for memcheck build

JadenW:


tests/frontend/org/voltdb/TestAdHocQueries.java
```diff
@@ -470,12 +470,15 @@ public void testAdHocLengthLimit() throws Exception {
         System.out.println("Starting testAdHocLengthLimit");
         TestEnv env = new TestEnv(m_catalogJar, m_pathToDeployment, 2, 2, 1);
 
+        env.setUp();
+        // by pass valgrind due to ENG-7843
+        if (env.isValgrind() || env.isMemcheckDefined()) {
```
manjujames: Was isValgrind check not working? Why wasn't it working?

JadenW: The Jerkins jobs will use a memcheck build. 
However, LocalCluster will always use Native_EE_JNI backend even with memchcheckDefined if there are more one node in the cluster. 
Line 266 in LocalCluster.java      
 if (isMemcheckDefined() && (target == BackendTarget.NATIVE_EE_JNI) && m_hostCount == 1)  {
           m_target = BackendTarget.NATIVE_EE_VALGRIND_IPC;
....

So valgrind check will not working. The test will then try to run a JNI backend which isn't build in jerkins. I think that's the reason this test fails.

manjujames: I see. Thanks.

PR #3542, opened 2016-05-02
Eng 8932 to review full joins

paulmartel:


src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -1298,51 +1282,77 @@ private static VoltXMLElement voltGetXMLSpecification(QuerySpecification select,
...
+            }
+        }
+    }
+
+    static private void collapseCoalesceExpression(VoltXMLElement element, String table, String tableAlias) {
+        List<VoltXMLElement> exprCols = element.extractSubElements("columnref", null, null);
+
+        // Iterate over the columns looking for the first columnref expression that matches input table
+        // Once found, replace the input element 'in-place' with the matching columnref
+        for(VoltXMLElement columnref : exprCols) {
```
paulmartel: space after `for`

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -1298,51 +1282,77 @@ private static VoltXMLElement voltGetXMLSpecification(QuerySpecification select,
...
+            if ((tableAlias != null && columnref.hasValue("tablealias", tableAlias)) ||
+                    (table != null && columnref.hasValue("table", table))) {
```
paulmartel: hasValue aleady does the right thing for null values.
So, you have the option of making this code simpler or changing the hasValue guard to an assert(value != null).

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1989,6 +1993,91 @@ private void resolveTableColumns(Table table) {
...
+            uniqueColumnrefs.add(columnref);
+        }
+        // Delete original children
+        exp.children.clear();
+
+        // There should be at least 2 columnref expressions
+        assert(uniqueColumnrefs.size() > 1);
+        VoltXMLElement lastAlternativeExpr = null;
+        VoltXMLElement resultColaesceExpr = null;
+        while(!uniqueColumnrefs.isEmpty()) {
```
paulmartel: space after `while`

paulmartel: I would find this a little easier to follow as:

    while (true) {
        VoltXMLElement next = uniqueColumnrefs.pop();
        if (uniqueColumnrefs.isEmpty()) {
            // Last columnref. Simply plug it in to the last Then Expression
            assert(lastAlternativeExpr != null);
            // Add next as the first child
            lastAlternativeExpr.children.add(0, next);
            break;
        }
        VoltXMLElement next = uniqueColumnrefs.pop();
        if (!uniqueColumnrefs.isEmpty()) {
        ...

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
alexeevm: corrected

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
alexeevm: Simplified. I think it would be more practical to allow null arguments to the hasValue method.
if (columnref.hasValue("tablealias", tableAlias) || columnref.hasValue("table", table)) {
             ....
}

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
alexeevm: I think you meant
while (true) {
    VoltXMLElement next = uniqueColumnrefs.pop();
    if (uniqueColumnrefs.isEmpty()) {
        // Last columnref. Simply plug it in to the last Then Expression
        assert(lastAlternativeExpr != null);
        // Add next as the first child
        lastAlternativeExpr.children.add(0, next);
        break;
    }
   ...

I removed the second pop call and the following IF statement from the WHILE loop

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
alexeevm: I also added a BREAK statement inside the IF - once a match is found no need to iterate over the rest of the columnref expressions

PR #3543, opened 2016-05-02
ENG-9873 IN LIST parameter type checks to prevent undefined EE behaviors

paulmartel:
moved the checks into ProcedureRunner, the only caller.

Removed "Java" from some java-unrelated SQLStmt names.
Removed the "FromTest" recently appended to the name of a VoltType method with potentially wider (end user) uses.

Expanded the statement parameter type check junit test matrix.


wolffcm:
I'm looking at this now.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -970,22 +995,43 @@ public void initSQLStmt(SQLStmt stmt, Statement catStmt) {
...
     }
 
 
     protected void reflect() {
+        Map<String, SQLStmt> stmtMap;
+
         // fill in the sql for single statement procs
-        if (m_catProc.getHasjava() == false) {
+        if (m_catProc.getHasjava()) {
+            // this is where, in the case of java procedures, m_method is set
```
hhakimi53: The comment may be out of date as procedure does not have member called m_method. I think we should update it to: "this is where, in the case of java procedures, m_procMethod is set"

src/frontend/org/voltdb/VoltType.java
```diff
@@ -33,10 +33,27 @@
 
 
 /**
- * Represents a type in a VoltDB Stored Procedure or {@link VoltTable VoltTable}.
+ * Represents a type for a {@link VoltTable VoltTable} column or a SQLStmt
```
wolffcm: This is a class that is included in the VoltDB Java client.  We should perhaps restrict the content of Javadoc comments to what a user might really need to know to write their applicaiton?  Implementation details could be in a non-Javadoc comment.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -838,50 +834,78 @@ public DependencyPair executeSysProcPlanFragment(
...
+                }
+            }
+            // Allow initialization of string or integer or even float
+            // (microseconds) yet not decimal (?) from timestamp,
+            // for backward compatibility.
+            // TODO: Should we deprecate this soon as something
+            // easy enough to work around and too easy to abuse?
+            else if (argType.isBackendIntegerType()) {
+                if ((expectedType.isNumber() &&
+                        argType != VoltType.DECIMAL) ||
```
hhakimi53: If  argType is BackendIntegerType(), than argType will not be VoltType.Decimal, so " argType != VoltType.DECIMAL" check is not necessary.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -838,50 +834,78 @@ public DependencyPair executeSysProcPlanFragment(
...
-    private final void throwIfInfeasibleTypeConversion(SQLStmt stmt, Class <?> argClass, int argInd, VoltType expectedType) {
-        VoltType argType = VoltType.INVALID;
-        boolean isArray = false;
-
-        // Argument to param for IN list will come in as array of arguments. The varbinary can also appear
-        // as array of tinyint/byte. Check if the expected type is var-binary and filter it based on that
-        if(argClass.isArray() && expectedType != VoltType.VARBINARY) {
-            argType = VoltType.typeFromClass(argClass.getComponentType());
-            isArray = true;
+    private final void throwIfInfeasibleTypeConversion(SQLStmt stmt,
```
hhakimi53: :+1:  with VoltType and current refactor, logic is much cleaner thn before.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -895,7 +919,8 @@ private final ParameterSet getCleanParams(SQLStmt stmt, boolean verifyTypeConv,
             // handle non-null values
             if (inArgs[ii] != null) {
                 args[ii] = inArgs[ii];
-                if (verifyTypeConv) {
+                assert(type != VoltType.INVALID);
+                if (verifyTypeConv && type != VoltType.INVALID) {
```
hhakimi53: Thought: In release code, if for some reason the expected type is evaluated to be invalid (not sure which scenario it will happen), for that case should the logic let the type checking still proceed so that it can be flagged?

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -838,50 +834,78 @@ public DependencyPair executeSysProcPlanFragment(
...
+        if (argClass.isArray()) {
+            // The statement parameter model doesn't currently support a
+            // general concept of an array-typed parameter. Instead, it
+            // defines certain special VoltTypes that are implicitly
+            // convertible from specific array-typed arguments.
+            // These are provided in the sqlType arg.
+            // For example,
+            // VARBINARY parameters accept byte[] arguments,
+            // INLIST_OF_STRING parameters accept String[] arguments,
+            // INLIST_OF_LONG parameters accept integer-typed array
```
hhakimi53: Minor: should it be INLIST_OF_BIGINT?

hhakimi53:
Overall the changes look fine.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -1008,45 +1054,35 @@ protected void reflect() {
...
+                    }
+                    else if (type == VoltType.SMALLINT) {
                         type = VoltType.BIGINT;
-                    } else if (type == VoltType.TINYINT) {
+                    }
+                    else if (type == VoltType.TINYINT) {
                         type = VoltType.BIGINT;
-                    } else if (type == VoltType.NUMERIC) {
+                    }
+                    else if (type == VoltType.NUMERIC) {
```
wolffcm: It was like this before you were here, but arbitrarily converting from numeric to float here seems wrong.  One type is exact and the other is approximate.  Could be a bug where we lose precision here?  (If so, very small compared to the ones we've been fixing lately.)

tests/frontend/org/voltdb/regressionsuites/TestTypeConversionSuite.java
```diff
@@ -47,46 +47,111 @@
...
+            "byte\\[\\]",
+            "GeographyPointValue",
+            "GeographyValue",
+    };
+    private static final VoltType[] m_tableColInListTypeVal = {
+            VoltType.INLIST_OF_BIGINT,
+            VoltType.INLIST_OF_BIGINT,
+            VoltType.INLIST_OF_BIGINT,
+            VoltType.INLIST_OF_BIGINT,
+            null, // IN LIST of FLOAT not supported
```
wolffcm: I can understand why FLOAT would not be supported, since equality comparison with floats can be dubious.  IN lists of DECIMAL seem reasonable though.

Has this limitation always existed?

wolffcm:
That's all my comments.

PR #3545, opened 2016-05-03
Eng 10148 pause incompatible dr stream

luyangco:
The best order to review those changes is

Producer Side:
Site.java::updateCatalog() -> generatedDRCatalogChangeEvent() -> VoltDBEngine.cpp::VoltDBEngine::executeTask()->DRTupleStream.cpp::generateDRCatalogChangeEvent()->PartitionDRGatewayImpl.java:: onBinaryDR()->DRProducerConnectionInterface.java::prepareForDispatch()

Consumer Side:
AbstractDRPartitionReceiver.java::processBuffer()->preprocessBuffer() in both syncSnapshot and normal buffer receiver -> handleInbandDREvent()

Also, this patch includes removal of an unused parameter in generate DR log path, removal the old out-of-band CATALOG_UPDATE event.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -1783,8 +1783,7 @@ void VoltDBEngine::updateHashinator(HashinatorType type, const char *config, int
     }
 }
 
-void VoltDBEngine::dispatchValidatePartitioningTask(const char *taskParams) {
-    ReferenceSerializeInputBE taskInfo(taskParams, std::numeric_limits<std::size_t>::max());
+void VoltDBEngine::dispatchValidatePartitioningTask(ReferenceSerializeInputBE &taskInfo) {
     std::vector<CatalogId> tableIds;
     const int32_t numTables = taskInfo.readInt();
     for (int ii = 0; ii < numTables; ii++) {
@@ -1792,7 +1791,7 @@ void VoltDBEngine::dispatchValidatePartitioningTask(const char *taskParams) {
     }
 
     HashinatorType type = static_cast<HashinatorType>(taskInfo.readInt());
-    const char *config = taskParams + (sizeof(int32_t) * 2) +  (sizeof(int64_t) * tableIds.size());
+    const char *config = taskInfo.getRawPointer();
     TheHashinator* hashinator;
     switch(type) {
         case HASHINATOR_LEGACY:
@@ -1867,16 +1866,15 @@ int64_t VoltDBEngine::applyBinaryLog(int64_t txnId,
     return rowCount;
 }
 
-void VoltDBEngine::executeTask(TaskType taskType, const char* taskParams) {
+void VoltDBEngine::executeTask(TaskType taskType, ReferenceSerializeInputBE &taskInfo) {
     switch (taskType) {
     case TASK_TYPE_VALIDATE_PARTITIONING:
-        dispatchValidatePartitioningTask(taskParams);
+        dispatchValidatePartitioningTask(taskInfo);
         break;
     case TASK_TYPE_GET_DR_TUPLESTREAM_STATE:
         collectDRTupleStreamStateInfo();
         break;
     case TASK_TYPE_SET_DR_SEQUENCE_NUMBERS: {
-        ReferenceSerializeInputBE taskInfo(taskParams, std::numeric_limits<std::size_t>::max());
         int64_t partitionSequenceNumber = taskInfo.readLong();
         int64_t mpSequenceNumber = taskInfo.readLong();
         if (partitionSequenceNumber >= 0) {
@@ -1889,7 +1887,6 @@ void VoltDBEngine::executeTask(TaskType taskType, const char* taskParams) {
         break;
     }
     case TASK_TYPE_SET_DR_PROTOCOL_VERSION: {
-        ReferenceSerializeInputBE taskInfo(taskParams, std::numeric_limits<std::size_t>::max());
         uint32_t drVersion = taskInfo.readInt();
         if (drVersion != DRTupleStream::PROTOCOL_VERSION) {
             m_executorContext->setDrStream(m_compatibleDRStream);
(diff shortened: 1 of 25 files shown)


src/ee/voltdbipc.cpp
```diff
@@ -1523,8 +1523,9 @@ void VoltDBIPC::executeTask(struct ipc_command *cmd) {
     try {
         execute_task *task = (execute_task*)cmd;
         voltdb::TaskType taskId = static_cast<voltdb::TaskType>(ntohll(task->taskId));
+        ReferenceSerializeInputBE input(m_engine->getParameterBuffer(), m_engine->getParameterBufferCapacity());
```
slowsoul: For voltdbipc, getParameterBuffer() seems to return NULL, getParameterBufferCapacity() seems to return 0, as set in LINE 615. Maybe we should revert to the original code here.

src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -1537,4 +1552,18 @@ public void setDRProtocolVersion(int drVersion) {
         m_ee.executeTask(TaskType.SET_DR_PROTOCOL_VERSION, paramBuffer);
         hostLog.info("DR protocol version has been set to " + drVersion);
     }
+
+    /**
+     * Generate a DR catalog change event in DR stream when schema change made on DR tables
+     */
+    public void generatedDRCatalogChangeEvent(long uniqueId,
```
slowsoul: Small typo here? an extra 'd' after 'generate'

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1906,6 +1903,20 @@ void VoltDBEngine::executeTask(TaskType taskType, const char* taskParams) {
         m_resultOutput.writeInt(0);
         break;
     }
+    case TASK_TYPE_GENERATE_DR_CATALOG_CHANGE_EVENT: {
+        int64_t uniqueId = taskInfo.readLong();
```
wweiss-voltdb: With three changes this could be a generic event:
1. treat the catalogcommands as opaque data
2. distinguish mp and sp uniqueIds
3. make sure the previous buffer is flushed

src/ee/storage/DRTupleStream.cpp
```diff
@@ -493,6 +490,22 @@ bool DRTupleStream::checkOpenTransaction(StreamBlock* sb, size_t minLength, size
     return false;
 }
 
+// Assume the stream has already been drained, the current block is empty
+void DRTupleStream::generateDRCatalogChangeEvent(int64_t lastCommittedSpHandle, int64_t spHandle,
+        int64_t uniqueId, ByteArray catalogCommands) {
+    ExportSerializeOutput io(m_currBlock->mutableDataPtr(), m_currBlock->remaining());
+    io.writeBinaryString(catalogCommands.data(), catalogCommands.length());
+    m_currBlock->consumed(io.position());
+    m_uso += io.position();
```
nshi: I wonder if there is any benefit of making it look like a normal transaction by writing the BEGIN TXN/END TXN pair. In case we have to pass the record to the site for serialization, the EE on the consumer can easily ignore it.

src/ee/storage/DRTupleStream.cpp
```diff
@@ -493,6 +490,22 @@ bool DRTupleStream::checkOpenTransaction(StreamBlock* sb, size_t minLength, size
     return false;
 }
 
+// Assume the stream has already been drained, the current block is empty
```
nshi: You can call `extendBufferChain()` to make sure that the current block is flushed if there's data in it.

src/ee/storage/TupleStreamBase.cpp
```diff
@@ -189,6 +189,15 @@ void TupleStreamBase::commit(int64_t lastCommittedSpHandle, int64_t currentSpHan
     }
 }
 
+void TupleStreamBase::advanceSequenceNumber(int64_t uniqueId) {
+    m_openSequenceNumber++;
+    m_openUniqueId = uniqueId;
+    m_committedSequenceNumber = m_openSequenceNumber;
```
nshi: Don't you want to assign this prior to advancing the `m_openSequenceNumber`?

src/ee/storage/TupleStreamBase.cpp
```diff
@@ -189,6 +189,15 @@ void TupleStreamBase::commit(int64_t lastCommittedSpHandle, int64_t currentSpHan
     }
 }
 
+void TupleStreamBase::advanceSequenceNumber(int64_t uniqueId) {
+    m_openSequenceNumber++;
+    m_openUniqueId = uniqueId;
```
nshi: What about `m_openSpHandle`?

src/ee/storage/TupleStreamBase.cpp
```diff
@@ -189,6 +189,15 @@ void TupleStreamBase::commit(int64_t lastCommittedSpHandle, int64_t currentSpHan
     }
 }
 
+void TupleStreamBase::advanceSequenceNumber(int64_t uniqueId) {
```
nshi: You can also refactor `commit()` to call this. It's the same code.

src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -1403,6 +1406,13 @@ public boolean updateCatalog(String diffCmds, CatalogContext context, CatalogSpe
             return true;
         }
 
+        boolean DRCatalogChange = false;
+        CatalogMap<Table> tables = m_context.catalog.getClusters().get("cluster").getDatabases().get("database").getTables();
+        for (Table t : tables) {
+            if (t.getIsdred()) {
+                DRCatalogChange |= diffCmds.contains("tables#" + t.getTypeName());
```
nshi: As soon as this returns true, you can break out of the loop.

nshi:
Review finished.

src/ee/voltdbipc.cpp
luyangco: Replace this line with `ReferenceSerializeInputBE input(task->task, MAX_MSG_SZ);`

src/frontend/org/voltdb/iv2/Site.java
luyangco: Fixed.

src/ee/storage/DRTupleStream.cpp
luyangco: I already create a special case for event buffer, catalog update buffer will be treat as a MP multi-hash txn but with empty binary log, when the site detects the txn has empty binary log, it knows to merge the tracker but ignore the binary log.

src/frontend/org/voltdb/iv2/Site.java
luyangco: Addressed.

src/ee/execution/VoltDBEngine.cpp
luyangco: Addressed

src/ee/storage/TupleStreamBase.cpp
luyangco: Addressed.

src/ee/storage/AbstractDRTupleStream.h
```diff
@@ -61,7 +61,6 @@ class AbstractDRTupleStream : public TupleStreamBase {
     virtual size_t appendTuple(int64_t lastCommittedSpHandle,
                        char *tableHandle,
                        int partitionColumn,
-                       int64_t txnId,
                        int64_t spHandle,
```
nshi: Is the order of the first two parameters wrong? They seem to be passed in reverse order.

PR #3546, opened 2016-05-03
ENG-10264 support auto sensing startup mode

vtkstef:


lib/python/voltcli/verbs.py
```diff
@@ -536,8 +532,6 @@ def go(self, verb, runner):
             final_args.extend(['publicinterface', runner.opts.publicinterface])
         if runner.opts.voltdbroot:
             final_args.extend(['voltdbroot', runner.opts.voltdbroot])
-        if runner.opts.clustername:
-            final_args.extend(['clustername', runner.opts.clustername])
         if self.subcommand in ('create', 'initialize'):
```
akhanzode: Do u want initialize as verb? I like init easy to type and looks similar to postgresql

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -798,6 +866,15 @@ public void initialize(VoltDB.Configuration config) {
                 VoltDB.crashLocalVoltDB("Elastic join is prohibited on a replica cluster.", false, null);
             }
 
+            // we can leave here if we start without command logs
+            if (!m_catalogContext.getDeployment().getCommandlog().isEnabled() && m_config.m_startAction == StartAction.INITIALIZE) {
```
akhanzode: So community will fall in here?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1546,6 +1643,13 @@ int readDeploymentAndCreateStarterCatalogContext() {
                 if (deployment.getCluster().getHostcount() < forcedHostCount) {
                     deployment.getCluster().setHostcount(forcedHostCount);
                 }
+                // force http port enabled
+                if (deployment.getHttpd() == null) {
+                    deployment.setHttpd(new HttpdType());
```
akhanzode: How does listener gets setup? why do u need to enable it?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1546,6 +1643,13 @@ int readDeploymentAndCreateStarterCatalogContext() {
...
+                }
+                if (!Boolean.TRUE.equals(deployment.getHttpd().isEnabled())) {
+                    deployment.getHttpd().setEnabled(true);
```
akhanzode: should be false right?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1754,39 +1858,87 @@ else if (sysType.getQuery().getTimeout() == 0) {
...
+            this.deploymentBytes = deploymentBytes;
+            this.deployment = deployment;
+        }
+    }
+
+    Optional<byte []> modifyIfNecessaryDeploymentHostCount(DeploymentType deployment, int hostCount) {
+        Optional<byte[]> deploymentBytes = Optional.empty();
+        if ((deployment.getCluster().getKfactor()+1) > hostCount) {
+            VoltDB.crashLocalVoltDB(
+                    "Number cluster members " + hostCount + " must greater than ksafety "
```
akhanzode: Number "of"

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1754,39 +1858,87 @@ else if (sysType.getQuery().getTimeout() == 0) {
...
+        final byte [] deploymentBytes;
+        final DeploymentType deployment;
+
+        ReadDeploymentResults(byte [] deploymentBytes, DeploymentType deployment) {
+            this.deploymentBytes = deploymentBytes;
+            this.deployment = deployment;
+        }
+    }
+
+    Optional<byte []> modifyIfNecessaryDeploymentHostCount(DeploymentType deployment, int hostCount) {
```
akhanzode: What is Optional about it?

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: What happens in community?

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -64,6 +74,13 @@
     public static final int SITES_TO_HOST_DIVISOR = 100;
     public static final int MAX_SITES_PER_HOST = 128;
 
+    // Staged filenames for advanced deployments
+    public static final String STAGED_CLUSTER_NAME = "_CLUSTERNAME";
+    public static final String STAGED_DEPLOYMENT = "_DEPLOYMENT";
+    public static final String STAGED_MESH = "_MESH";
+    public static final String DEFAULT_CLUSTER_NAME = "aquila"; // Italian for eagle
```
akhanzode: Why do you need a default? If needed why cant it be something user will never type?

src/frontend/org/voltdb/deploy/CommandLineMeshProvider.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.UUID;
+
+import org.voltdb.utils.Digester;
+import org.voltdb.utils.MiscUtils;
+
+import com.google_voltpatches.common.base.Splitter;
+import com.google_voltpatches.common.collect.ImmutableSortedSet;
+import com.google_voltpatches.common.net.HostAndPort;
+
+public class CommandLineMeshProvider {
```
akhanzode: Should we define an interface so we know how other provider will look like.

src/frontend/org/voltdb/deploy/ConfigProbeTracker.java
```diff
@@ -0,0 +1,138 @@
...
+
+    protected final ConfigProbeResponse initialProbeResponse;
+    protected final AtomicLong meshTimeout;
+    protected final ConcurrentMap<UUID,Long> receivedProbes =new ConcurrentHashMap<>();
+    protected final SettableFuture<Map<UUID,Long>> done = SettableFuture.create();
+    protected final SettableFuture<LeaderAllClear> allClear = SettableFuture.create();
+    protected final int nodeCount;
+
+    public ConfigProbeTracker(
+            int nodeCoount,
```
akhanzode: nodecount

src/frontend/org/voltdb/deploy/ConfigProber.java
```diff
@@ -0,0 +1,685 @@
...
+        int unmeshed = 0, operational = 0;
+        for (StatusTracker st: m_statuses.values()) {
+            if (st.getNodeState().unmeshed()) ++unmeshed;
+            if (st.getNodeState().operational()) ++operational;
+        }
+        // if (kfactor + 1) nodes are unmeshed it means that every member in the cluster
+        // is starting. This requires more stringent checks
+        //    1) this node has to recieve probes from all the members
+        //    2) the mesh configuration must match in all
+        if (unmeshed >= m_safety) {
```
akhanzode: Shouldnt this involve operational? What is unmeshed means?

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -308,14 +338,61 @@ public String getUser() {
...
+
+        @Override
+        public void handle(String target, Request baseRequest,
+                HttpServletRequest request, HttpServletResponse response)
+                        throws IOException, ServletException {
+            response.setContentType(jsonContentType);
+            response.setStatus(HttpServletResponse.SC_OK);
+            KSafetyResponse ksr = m_notYetAvailable;
+
+            if (m_statusProvider.getNodeState().operational()) {
```
akhanzode: There is a flag in admin listener which can be refactored to use this.

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -308,14 +338,61 @@ public String getUser() {
...
+            }
+            m_ksafetyWriter.writeValue(response.getWriter(), ksr);
+            baseRequest.setHandled(true);
+        }
+    }
+
     // /management/status handler
     class ManagementStatusHandler extends AbstractHandler {
-        private final ObjectMapper m_mapper;
+        private final ObjectWriter m_respw =
```
akhanzode: There is only one handler so why need a holder?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -511,14 +536,58 @@ public void initialize(VoltDB.Configuration config) {
                 }
             }
             if (config.m_startAction != StartAction.INITIALIZE) {
-                DeploymentType deployment = readPrimedDeployment(config.m_pathToDeployment);
+                ReadDeploymentResults readDepl = readPrimedDeployment(config);
```
akhanzode: How do we let people prime existing cluster so next time they can use start?

src/frontend/org/voltdb/RealVoltDB.java
vtkstef: tricky Boolean.TRUE.equals(null) == false Boolean.FALSE.equals(null) == false that is why I use !Boolean.TRUE.equals()

src/frontend/org/voltdb/RealVoltDB.java
vtkstef: in community you can only do create no point in allowing them to init and start

src/frontend/org/voltdb/RealVoltDB.java
vtkstef: you initialize so you can start. I am forcing it here. I could crash instead if you prefer it.

src/frontend/org/voltdb/RealVoltDB.java
vtkstef: The optional is populated only if we need to change the host count

src/frontend/org/voltdb/VoltDB.java
vtkstef: a more common name you mean? or do you want to make the cluster name option mandatory?

src/frontend/org/voltdb/deploy/CommandLineMeshProvider.java
vtkstef: I wanted to see a pattern first before we define the interface. I agree it will be an interface. let's see how it looks like after we do the etcd one

src/frontend/org/voltdb/deploy/ConfigProber.java
vtkstef: From NodeState
```
static final EnumSet<NodeState> unmeshed = EnumSet.of(INITIALIZING,WAITINGFORLEADER);
```

src/frontend/org/voltdb/utils/HTTPAdminListener.java
vtkstef: for object mapper? all our JSON deployment ops use it. let's leverage only one. see comment where I set up the holder

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: Or can we flip to no deployment automation mode?

vtkstef:
--squash merged

PR #3547, opened 2016-05-04
Planner support for PartitionByPlanNode and Windowed Rank

BillWhite:
This commit adds planner support for PartitionByPlanNode and for the
windowed rank function.  This is the slow path implementation, not the
fast path implementation of Xin's Innovation Week project.

We create a PartitionByPlanNode preceded by an OrderByPlanNode.  The
windowed rank expression is kept in the display list.  The
OrderByPlanNode sorts on the partition by columns, and then by the order
by columns in the windowed rank expression.

There are tests for the unsupported case of having more than one rank
and for having a rank with a group by.  There are tests for the
supported case of an ordinary windowed rank function and for having a
rank computed over a subquery.

This is bluish on Jenkins.  The only failures are the usual TestFuzzMeshArbiter,
and two others which fail on master.


src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,465 @@
...
+import org.voltdb.types.IndexType;
+import org.voltdb.types.SortDirectionType;
+import org.voltdb.utils.CatalogUtil;
+
+public class WindowedExpression extends AbstractExpression {
+    public enum Members {
+        PARTITION_BY_EXPRESSIONS,
+        ORDER_BY_EXPRESSIONS,
+        ORDER_BY_DIRECTIONS,
+        EXP,
```
wolffcm: These two members `EXP` and `DIR`, seem kind of generic, maybe give them more descriptive names?  As it is I'm not sure what they're for.

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,465 @@
...
+import org.voltdb.catalog.Index;
+import org.voltdb.catalog.Table;
+import org.voltdb.planner.PlanningErrorException;
+import org.voltdb.planner.SubPlanAssembler;
+import org.voltdb.planner.parseinfo.StmtTableScan;
+import org.voltdb.types.ExpressionType;
+import org.voltdb.types.IndexType;
+import org.voltdb.types.SortDirectionType;
+import org.voltdb.utils.CatalogUtil;
+
```
wolffcm: We would benefit from a class comment here.

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,465 @@
...
+        tve.setTypeSizeBytes(partitionCol.getType(), partitionCol.getSize(), partitionCol.getInbytes());
+
+        boolean pkCovered = false;
+        for (AbstractExpression ex: partitionbyExprs) {
+            if (ex.equals(tve)) {
+                pkCovered = true;
+                break;
+            }
+        }
+        // If PARTITION BY not covering PK, we can not use that index to calculate ranking
```
wolffcm: I usually use PK to abbreviate "primary key" which is different from the table's "partition key," so this usage of PK could be confusing.

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,465 @@
...
+
+        /*
+         * Load the array of partition expressions.  The AbstractExpression class
+         * has a useful routine to do just this for us.
+         */
+        AbstractExpression.loadFromJSONArrayChild(m_partitionByExpressions, jobj, Members.PARTITION_BY_EXPRESSIONS.name(), null);
+
+        /*
+         * Unfortunately we cannot use AbstractExpression.loadFromJSONArrayChild here,
+         * as we need to get a sort expression and a sort order for each column.
```
wolffcm: This seems like a good candidate for refactoring?  We must do this elsewhere as well.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -497,7 +502,54 @@ private SelectSubqueryExpression parseSubqueryExpression(VoltXMLElement exprNode
...
+                }
+
+            } else if (ele.name.equals("orderbyList")) {
+                for(int i = 0; i < ele.children.size(); i++) {
+                    VoltXMLElement childNode = ele.children.get(i);
+                    boolean isDecending = Boolean.valueOf(childNode.attributes.get("decending"));
+                    if (i == 0) {
+                        areAllDecending = isDecending;
+                    } else if (areAllDecending != isDecending) {
+                        throw new PlanningErrorException("invalid RANK order by expression without consistent "
```
wolffcm: Why do all the OB expressions need to be ascending or descending?

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,465 @@
...
+    public SortDirectionType getOrderByDirectionOfExpression(AbstractExpression partitionByExpr) {
+        for (int idx = 0; idx < m_orderByExpressions.size(); idx += 1) {
+            if (m_orderByExpressions.get(idx).equals(partitionByExpr)) {
+                return m_orderByDirections.get(idx);
+            }
+        }
+        return DEFAULT_ORDER_BY_DIRECTION;
+    }
+}
+
```
wolffcm: This class is a little different from other expression classes in that it has ownership of other expressions that are not children.  Some methods in the planner try to accumulate all expressions of a particular type by recursing over an expression tree through child nodes.  I'm thinking of methods like AbstractExpression.findAllSubexpressionsOfClass, but I think there are others like it.

When we update TVEs during the "column index resolution" phase of planning, we'll want to make sure not to miss the order by and partition by expressions in a WindowedExpression.  Let me know if you have questions about this.

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,465 @@
...
+        }
+
+        int minValue = Collections.min(m_indexColumnsCovered);
+        int minIndex = m_indexColumnsCovered.indexOf(minValue);
+        Index index = m_indexCandidates.get(minIndex);
+        refreshWithBestIndex(index, minValue);
+        return index;
+    }
+
+    private void isExpressionListSubsetOfIndex(List<AbstractExpression> exprs, Index index) {
```
wolffcm: I generally expect method names that start with `is` to return a Boolean.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,93 @@ public PlanNodeType getPlanNodeType() {
...
     @Override
     protected String explainPlanForNode(String indent) {
         String optionalTableName = "*NO MATCH -- USE ALL TABLE NAMES*";
-        StringBuilder sb = new StringBuilder(" PARTIION PLAN: " + super.explainPlanForNode(indent));
-        sb.append(indent).append("SORT BY: \n");
-        for (int idx = 0; idx < m_sortExpressions.size(); idx += 1) {
-            AbstractExpression ae = m_sortExpressions.get(idx);
-            SortDirectionType dir = m_sortDirections.get(idx);
+        StringBuilder sb = new StringBuilder(" PARTITION BY PLAN: " + super.explainPlanForNode(indent) + "\n");
+        sb.append(indent).append(" SORT BY: \n");
```
wolffcm: The PARTITION BY clause comes first in the SQL syntax, so it might make sense to put it first here?

src/frontend/org/voltdb/planner/microoptimizations/InlineAggregation.java
```diff
@@ -62,8 +63,10 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
 
     AbstractPlanNode inlineAggregationApply(AbstractPlanNode plan) {
         // check for an aggregation of the right form
-        if ((plan instanceof AggregatePlanNode) == false)
+        if ((plan instanceof AggregatePlanNode) == false
+                || (plan instanceof PartitionByPlanNode)) {
             return plan;
+        }
```
wolffcm: Question unrelated to this code: do we have a microoptimization that can eliminate an order by node if we can convert a seq scan to an index scan using an appropriate index?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionRank.java
```diff
@@ -0,0 +1,140 @@
+/* Copyright (c) 2001-2009, The HSQL Development Group
```
wolffcm: This needs the HSQL license header?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionRank.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Override
+    public String getSQL() {
+
+        StringBuffer sb = new StringBuffer();
+
+        sb.append(Tokens.T_RANK).append("()").append(Tokens.T_ORDER + ' ' + Tokens.T_BY).append(' ');
+
+        // Todo
+        return sb.toString();
```
wolffcm: Do we actually need this for anything?  For scalar functions, I think we use the getSQL method for indexes that use expressions.  That probably doesn't apply here.

tests/bills-allconsole.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
wolffcm: Did you mean to check this in?

tests/frontend/org/voltdb/planner/TestWindowingFunctions.java
```diff
@@ -0,0 +1,111 @@
...
+        assertEquals("ARANK", column.getColumnAlias());
+        AbstractPlanNode OBNode = abstractPBPlanNode.getChild(0);
+        assertTrue(OBNode instanceof OrderByPlanNode);
+        AbstractPlanNode abstractSScanNode = OBNode.getChild(0);
+        assertTrue(abstractSScanNode instanceof NestLoopPlanNode);
+    }
+
+    public void testRankFailures() {
+        failToCompile("SELECT RANK() OVER (PARTITION BY A ORDER BY B ) FROM AAA GROUP BY A;",
+                      "windowed operations are not allowed in the same query as an ORDER BY.");
```
wolffcm: It should read "as a GROUP BY," yes?

wolffcm: And I would think that using a windowed expression and a trailing ORDER BY that applies to the output of the SELECT statement *is* supported?

tests/frontend/org/voltdb/planner/TestWindowingFunctions.java
```diff
@@ -0,0 +1,111 @@
...
+        NodeSchema  schema = pbPlanNode.getOutputSchema();
+        assertEquals(2, schema.getColumns().size());
+        SchemaColumn column = schema.getColumns().get(1);
+        assertTrue(column.getExpression() instanceof WindowedExpression);
+        assertEquals("ARANK", column.getColumnAlias());
+        AbstractPlanNode OBNode = abstractPBPlanNode.getChild(0);
+        assertTrue(OBNode instanceof OrderByPlanNode);
+        AbstractPlanNode SScanNode = OBNode.getChild(0);
+        assertTrue(SScanNode instanceof SeqScanPlanNode);
+    }
```
wolffcm: There's a bunch of planner code that mentions indexes with respect to windowed expressions.  Is any of that actually live code?

wolffcm:
That's all my comments this round.  Overall, I like the approach you've taken.  I think there are some items that need to be addressed, especially with respect to resolving column indexes.

tests/bills-allconsole.xml
BillWhite: No, this was a mistake.

tests/frontend/org/voltdb/planner/TestWindowingFunctions.java
BillWhite: Yes, this is a mistake in the error message.  The test clearly has a windowed function call in the select list and a group by, not an order by.

tests/frontend/org/voltdb/planner/TestWindowingFunctions.java
BillWhite: No.  I copied Xin's IW work, took some stuff out, and left some other stuff to remind us of how we will want to take the fast path.  Perhaps I should just remove all of the unused stuff.  I guess the problem is that we don't know what's used and what's unused, and we might go to some length to preserve an invariant which is not important.  So, it may be better to remove all but minimum.

wolffcm: The Pragmatic Programmer book talks about prototypes vs. tracer code.  I wonder if that applies here?

http://blog.codinghorror.com/the-prototype-pitfall/

There's definitely some weirdness here because Xin implemented the fast path first, but we can't ship until we get the slow path working.

src/frontend/org/voltdb/expressions/WindowedExpression.java
BillWhite: All this indexing stuff is gone.  I'm going with a very bare bones implementation.  We'll expand it when we get to the fast path.

src/frontend/org/voltdb/expressions/WindowedExpression.java
BillWhite: Done.  We do the same thing in OrderByPlanNode.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: This is an artifact of the IW implementation.  It was corrected elsewhere, but not here.

Good catch.

src/frontend/org/voltdb/expressions/WindowedExpression.java
BillWhite: This is gone anyway.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
BillWhite: Ok.

src/frontend/org/voltdb/planner/microoptimizations/InlineAggregation.java
BillWhite: Not that I know of.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionRank.java
BillWhite: It looks like it has one to me.  I'm not sure I added it.  I suspect licensescheck.py thinks
everything in the directory needs an HSQL license header.

tests/frontend/org/voltdb/planner/TestWindowingFunctions.java
BillWhite: Interesting.  I'm pulling out everything from the IW work which we don't really need.

src/frontend/org/voltdb/expressions/WindowedExpression.java
BillWhite: Done.

src/frontend/org/voltdb/expressions/WindowedExpression.java
BillWhite: Good catch.  I missed this.  I made WindowExpression versions of the four operations "findAllSubexpressionsOfType", "findAllSubexpressionsOfClass", "hasSubexpressionsOfType" and "hasSubexpressionOfClass".  They call the super version and then search the partitionby and orderby lists.

In the column index resolution phase in the planner?  I thought we only resolve columns indices in planning nodes, and not in expressions.  In any case, I'm wondering if I've done the right thing in the planner itself.  I made the output columns of the PartitionByPlanNode be the display columns.  I'm wondering if the output of the PBPN should be the output of the OrderByPlanNode which precedes the PBPN plus a column for the rank expression, and that the PBPN should have a ProjectPlanNode as a parent to compute the final output.  The ProjectionPlanNode could eventually be made inline, since the PBPN is a subclass of AggregatePlanNode.  Then the PBPN doesn't have to know anything about display lists, and resolving column indices is very simple.  The output indices are either exactly the input indices or 1 plus the input indices plus the rank function's output, depending on whether the rank function is first or last.  I think first is better, bacause it makes generateOutputSchema simpler.  The PBPN constructor can create the rank column in the output schema, and then let generateOutputSchema copy from the input to the output.

BillWhite:
This is somewhat changed.  I'm going to create a fresh pull request.

BillWhite:
Apparently one cannot recreate a pull request.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -586,6 +587,44 @@ public void toJSONString(JSONStringer stringer) throws JSONException {
         }
     }
 
+    /**
+     * We need some enumerals which are common to PartitionByPlanNode and OrderByPlanNode
+     * and maybe others.  These are use as keys to create JSON.
```
paulmartel: typo "use" --> "used"

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -0,0 +1,279 @@
...
+        return false;
+    }
+
+    /**
+     *
+     * Given an expression, E, in the partition by list, if E is in the
+     * order by list, then return the order direction of E in that list.
+     * Otherwise, return the default value.
+     *
+     * There's got to be a better way to do this.
```
paulmartel: There IS a better way to do this. Use an old-school counted "for" loop in the caller instead of a "foreach" loop and use get(int i) to index through the parallel lists in parallel.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1477,12 +1492,16 @@ private boolean isValidJoinOrder(List<String> tableAliases)
...
+        return m_hasWindowingExpression;
+    }
+
     public boolean hasAggregateExpression() {
         return m_hasAggregateExpression;
     }
 
     public boolean hasAggregateOrGroupby() {
-        return m_hasAggregateExpression || isGrouped();
+        return hasAggregateExpression() || isGrouped();
```
paulmartel: Adds cost and ambiguity for no practical benefit.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -1123,8 +1134,7 @@ else if (m_parsedSelect.hasDistinctWithGroupBy() &&
      * @return true if a project node is required
      */
     private boolean needProjectionNode (AbstractPlanNode root) {
-        if ( root instanceof AggregatePlanNode ||
-             root.getPlanNodeType() == PlanNodeType.PROJECTION) {
+        if ( false == root.planNodeClassNeedsProjectionNode()) {
```
paulmartel: `false == ` is a peculiar way to express `!`.
BTW, `peculiar` is a polite way to say `wrong`.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2101,6 +2114,57 @@ private boolean switchToIndexScanForGroupBy(AbstractPlanNode candidate, IndexGro
...
+            }
+        }
+        // If we haven't set this it's probably bad news.
+        assert(windowedSchemaColumn != null);
+        // This will set the output schema to contain the
+        // windowed schema column only.  In generateOutputSchema
+        // we will add the input columns.
+        PartitionByPlanNode pnode = new PartitionByPlanNode(windowedSchemaColumn);
+        OrderByPlanNode onode = new OrderByPlanNode();
+        pnode.setWindowedColumn(windowedSchemaColumn);
```
paulmartel: Calling a setter function to set a field to the value that it was already set to in the constructor?

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2101,6 +2114,57 @@ private boolean switchToIndexScanForGroupBy(AbstractPlanNode candidate, IndexGro
...
+            AbstractExpression colExpr = colInfo.expression;
+            if (colExpr instanceof WindowedExpression) {
+                windowedSchemaColumn = new SchemaColumn(colInfo.tableName,
+                                                        colInfo.tableAlias,
+                                                        colInfo.columnName,
+                                                        colInfo.alias,
+                                                        colExpr);
+                break;
+            }
+        }
```
paulmartel: You went through the bother of defining `ParsedSelectStmt.getWindowedColinfo()`. Use it here.

src/frontend/org/voltdb/planner/PlanAssembler.java
paulmartel: BTW, this seems to be the only use pattern for this setter, so it can be dropped.

src/frontend/org/voltdb/planner/QueryPlanner.java
```diff
@@ -336,6 +336,48 @@ public boolean wasBadPameterized() {
...
+     *   of the notion of scanning a table or an index.
+     * </li>
+     * <li>
+     *   Create a PlanAssembler, and ask it for the best cost plan.  This uses the
+     *   side data created by the parser in the previous step.
+     * </li>
+     * <li>
+     *   If the plan is read only, slap a SendPlanNode on the front.  Presumably
+     *   an insert, delete or upsert will have added the SendPlanNode and perhaps
+     *   a ReceivePlanNode in the plan node tree already.
```
paulmartel: The generation of the ReceivePlanNode is not really relevant here.
This makes it sound like some feature specific to write statements when it is more a matter of single-fragment vs. multi-fragment plans. It has nothing to do with the top-most SendPlanNode, so I wouldn't mention it.

src/frontend/org/voltdb/planner/QueryPlanner.java
```diff
@@ -336,6 +336,48 @@ public boolean wasBadPameterized() {
...
+     *   Compute the output schema.  This computes the output schema for each
+     *   node recursively, using a node specific method.
+     * </li>
+     * <li>
+     *   Resolve the column indices.  This makes sure that the indices of all
+     *   TVEs in the output columns refer to the right input columns.
+     * </li>
+     * <li>
+     *   Do some final cleaning up and verifying of the plan.  For example,
+     *   We renumber the nodes staring at 1.
```
paulmartel: It's rude to stare.

src/frontend/org/voltdb/planner/microoptimizations/InlineAggregation.java
```diff
@@ -33,7 +34,7 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
     {
         assert(planNode != null);
 
-        // breath first:
+        // breadth first:
         //     find AggregatePlanNode with exactly one child
         //     where that child is an AbstractScanPlanNode.
         //     Inline any qualifying AggregatePlanNode to the its AbstractScanPlanNode.
```
paulmartel: While we're fixing typos, how about "to the its"?

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,153 @@ public PlanNodeType getPlanNodeType() {
...
+    }
+
+    public PartitionByPlanNode() {
+        this(null);
+    }
 
-        for (AbstractExpression expr : m_sortExpressions) {
-            expr.validate();
+    public PartitionByPlanNode(SchemaColumn winAggregateColumn) {
+        m_outputSchema = new NodeSchema();
```
paulmartel: I don't see the advantage to starting the setup of m_outputSchema here vs. doing it all in one place in generateOutputSchema. Save the trouble of revalidating this work there.

paulmartel: Also, there's a question of the utility of the empty outputschema allocated for the no-argument-constructor code path.
I THINK it will always get written over.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,153 @@ public PlanNodeType getPlanNodeType() {
...
 
     public int numberSortExpressions() {
-        return m_sortExpressions.size();
+        WindowedExpression we = getWindowedExpression();
+        return we.getOrderbySize();
+    }
+
+    public void addWindowedColumn(SchemaColumn col) {
+        m_windowedSchemaColumn = col;
+    }
```
paulmartel: Dead code.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,153 @@ public PlanNodeType getPlanNodeType() {
...
+    @Override
+    /**
+     * This is an AggregatePlanNode.  Normally these don't need projection
+     * nodes, but this one does.
+     */
+    public boolean planNodeClassNeedsProjectionNode() {
+        return true;
+    }
+
+    SchemaColumn       m_windowedSchemaColumn = null;
```
paulmartel: Wouldn't it be simpler to reference a WindowedExpression directly instead of a SchemaColumn?

src/frontend/org/voltdb/types/ExpressionType.java
```diff
@@ -145,7 +158,13 @@
     // -----------------------------
     ROW_SUBQUERY                 (RowSubqueryExpression.class, 400, "<row subquery>"),
     SELECT_SUBQUERY              (SelectSubqueryExpression.class, 401, "<select subquery>"),
-;
+
+    // ----------------------------
+    // WINDOWING
+    // ----------------------------
+    WINDOWING_RANK                (WindowedExpression.class, 500, "RANK")
```
paulmartel: Unused? redundant with AGGREGATE_WINDOWED_RANK?

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -1552,7 +1554,49 @@ private Expression readAggregateExpression(int tokenT) {
...
+                    continue;
+                }
+                break;
+            }
+        }
+
+        if (token.tokenType != Tokens.ORDER) {
+            throw unexpectedToken();
+        }
+        assert(token.tokenType == Tokens.ORDER);
```
paulmartel: No point in asserting that throw really works.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
paulmartel: Make private.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -37,9 +32,7 @@
  */
 public class PartitionByPlanNode extends HashAggregatePlanNode {
```
paulmartel: It's not clear to me that this class is inheriting anything useful from HashAggregatePlanNode.
I don't see anything useful in the (tiny) definition of HashAggregatePlanNode itself, so maybe direct inheritance from AggregatePlanNode would be clearer, but then I'm not seeing what's actually useful in the (large) definition of AggregatePlanNode, and it seems that most of the AggregatePlanNode members are unused/wasted/degenerate for a PartitionByPlanNode. Inheriting directly from AbstractPlanNode means defining a few abstract functions, but I'm not seeing how AggregatePlanNode's implementation are applicable here -- their only virtue appears to be that they exist and pass tests.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -511,7 +516,51 @@ private SelectSubqueryExpression parseSubqueryExpression(VoltXMLElement exprNode
...
+        List<AbstractExpression> partitionbyExprs = new ArrayList<AbstractExpression>();
+        List<AbstractExpression> orderbyExprs = new ArrayList<AbstractExpression>();
+        List<SortDirectionType>  orderbyDirs  = new ArrayList<SortDirectionType>();
+        boolean areAllDecending = false;
+
+        boolean isPercentRank = Boolean.valueOf(exprNode.attributes.get("isPercentRank"));
+
+        for (VoltXMLElement ele : exprNode.children) {
+            if (ele.name.equals("partitionbyList")) {
+                for(int i = 0; i < ele.children.size(); i++) {
```
paulmartel: missing conventional space between `for` and `(`.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -511,7 +516,51 @@ private SelectSubqueryExpression parseSubqueryExpression(VoltXMLElement exprNode
...
+                    VoltXMLElement childNode = ele.children.get(i);
+                    AbstractExpression expr = parseExpressionNode(childNode);
+                    partitionbyExprs.add(expr);
+                }
+
+            } else if (ele.name.equals("orderbyList")) {
+                for(int i = 0; i < ele.children.size(); i++) {
```
paulmartel: `for (`

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1477,12 +1492,16 @@ private boolean isValidJoinOrder(List<String> tableAliases)
         return true;
     }
 
+    public boolean hasWindowedExpression() {
```
paulmartel: "WindowED" VS. "WindowING" . I have no preference except to pick one and use it consistently.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -511,7 +516,51 @@ private SelectSubqueryExpression parseSubqueryExpression(VoltXMLElement exprNode
...
+    private AbstractExpression parseRankValueExpression(VoltXMLElement exprNode) {
+        // Parse individual rank expressions
+        List<AbstractExpression> partitionbyExprs = new ArrayList<AbstractExpression>();
+        List<AbstractExpression> orderbyExprs = new ArrayList<AbstractExpression>();
+        List<SortDirectionType>  orderbyDirs  = new ArrayList<SortDirectionType>();
+        boolean areAllDecending = false;
+
+        boolean isPercentRank = Boolean.valueOf(exprNode.attributes.get("isPercentRank"));
+
+        for (VoltXMLElement ele : exprNode.children) {
```
paulmartel: I was surprised and a little disturbed to see the partition by and order by lists as attributes of the rank expression, but I can't really explain why or what the alternative was that I was expecting. Maybe what I'm really hung up on is where the RANK fits in this structure. So, maybe I'd be more comfortable with a generic WindowedExpression parent object that references a WindowedAggFunction (like the RANK) and the partition by list and the order by list. That might extend more easily to a future of multiple aggs (becoming a list attribute of the WindowedExpression) all having the same partition-by-order-by context. Part of my problem is that it seems like the partition-by-list and order-by-list are likely to get picked off (picked apart even) and moved to different attributes of different plan nodes and it doesn't seem that useful to (also?) serialize/deswrialize them into the EE's plan fragment as attributes of the windowed agg function C++ object.

Thoughts?

Alternatively, I suppose nothing keeps us from doing a "shallow copy" of your WindowedExpression into the PartitionByPlanNode, leaving off one or both lists -- on the assumption that their plan contributions are captured elsewhere, perhaps as separately serializable attribute(s) of the PartitionByPlanNode or OrderByPlanNode.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,153 @@ public PlanNodeType getPlanNodeType() {
...
+        assert(abstractSE instanceof WindowedExpression);
+        return (WindowedExpression)abstractSE;
+    }
+
+    public AbstractExpression getPartitionByExpression(int idx) {
+        WindowedExpression we = getWindowedExpression();
+        return we.getPartitionByExpressions().get(idx);
+    }
+
+    public int numberPartitionByExpressions() {
```
paulmartel: "number" used as a verb dost make thy names hard to read -- prithee how about something more conventional like a "Size" suffix instead.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,153 @@ public PlanNodeType getPlanNodeType() {
...
+        return getWindowedExpression().getPartitionbySize();
```
paulmartel: standardize on the uppercase B.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -48,110 +41,153 @@ public PlanNodeType getPlanNodeType() {
...
+        WindowedExpression we = getWindowedExpression();
+        return we.getOrderbySize();
+    }
+
+    public void addWindowedColumn(SchemaColumn col) {
+        m_windowedSchemaColumn = col;
+    }
+
+    public SchemaColumn getWindowedColumns() {
+        return m_windowedSchemaColumn;
```
paulmartel: Dead code. I actually find it easier to understand a changeset when it does not contain future support.
Also, I find it harder to understand a follow-on changeset when it introduces whatever support functions it needs
rather than relying on pre-existing dormant code which now appears nowhere in the changeset.

Even this review has had that flavor for me as a rework of the false start that already exists in master, but I guess that's just how it went down.

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
```diff
@@ -0,0 +1,442 @@
...
+
+        // no rank scan support yet
+        vt = client.callProcedure("@Explain", "select a from tu where percent_rank() over (partition by b order by a) > 0.1 order by abs(a);").getResults()[0];
+        assertTrue(vt.toString().contains("SEQUENTIAL SCAN"));
+
+        vt = client.callProcedure("@AdHoc", "select a from tu where percent_rank() over (partition by b order by a) > 0.1 order by abs(a);").getResults()[0];
+        validateTableOfScalarLongs(vt, new long[]{10, 20, 30, 40, 50});
+
+        vt = client.callProcedure("@Explain", "select a from tu where percent_rank() over (partition by b order by a) > 0.5 order by abs(a);").getResults()[0];
+        System.err.println(vt);
```
paulmartel: disable distracting verbose output on non-errors

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
```diff
@@ -0,0 +1,442 @@
...
+        //        success = config.compile(project);
+        //        assertTrue(success);
+        //        builder.addServerConfig(config);
+
+        // no clustering tests for functions
+
+        return builder;
+    }
+}
+
```
paulmartel: I don't have complete confidence in the column resolution code. Add tests with joins in the FROM clause (not in a subquery) and with more complex expressions (not just columns) in the select list. 

Also I don't see where you have disabled the index support for order by (the final non-over ORDER BY), so it seems like that would give the same wrong answers as the "microoptimization" approach -- when there is an index that wants to eliminate the final OrderByPlanNode but it shouldn't because a conflicting "over" ORDER BY is intervening. Test for that case.

paulmartel:
End of review round.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
paulmartel: FWIW, I was able to verify that AggregatePlanNode in place of HashAggregatePlanNode passes all the new test suites. Trying AbstractPlanNode would take a little more effort -- possibly even some thinking.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: I'm not sure what you are getting at.  Is this an idea for extending the current implementation to one which has more than one windowed aggregate display list column?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: Is this really more expensive?  I would think the compiler would inline this, since the variable is private.

src/frontend/org/voltdb/planner/PlanAssembler.java
BillWhite: Hmm.  The net seems to be split on this.  Some people think "if (false == ...)" is easier to understand than "if (!...)".  The idea is that it's easier to miss the single character "!" than to miss the phrase "false ==".

I'm not strongly attached to either.

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
BillWhite: This is all copied directly from Xin's IW branch.  I only change the "public void test" to "public void notest".  So it's not surprising that it has some peculiarities.

BillWhite:
Squash merged to master.

PR #3548, opened 2016-05-04
ENG-10160: include batch commands in the ddl displayed in VMC

miketex:
Changes for ENG-10160


src/frontend/org/voltdb/utils/CatalogSchemaTools.java
```diff
@@ -76,12 +76,17 @@
      * SQL DDL needs to be generated, so instead, we opt to build the CREATE TABLE DDL
      * separately as we go here, and then fill it in to the StringBuilder being used
      * to construct the full canonical DDL at the appropriate time.
-     * @param Table - object to be analyzed
-     * @param String - the Query if this Table is a View
-     * @param Boolean - true if this Table is an Export Table
+     * @param sb
```
akhanzode: Add description of param in doc.

akhanzode:
Can you put toSchema() to return single String that has commented batch command and let ReportMaker replace commented to uncommented batch start and stop.

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
```diff
@@ -492,76 +500,114 @@ else if (proc.getLanguage().equals("JAVA")) {
 
         // The SQL statement variant may have terminated the CREATE PROCEDURE statement.
         if (!sb.toString().endsWith(";")) {
-            sb.append(";");
+            append(sb, ";");
```
vtkstef: which sb at which index is going to end with `;`

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
```diff
@@ -492,76 +500,114 @@ else if (proc.getLanguage().equals("JAVA")) {
...
-        for (String importLine : importLines) {
-            sb.append(importLine);
+        for (StringBuilder sb : sbs) {
+            for (String importLine : importLines) {
+                sb.append(importLine);
+            }
+        }
+    }
+
+    private static void append(StringBuilder[] sbs, String line) {
```
vtkstef: how big is this array? if its size is known or predictable could we turn them into ImmutableLists? arrays can be easily modified

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
```diff
@@ -492,76 +500,114 @@ else if (proc.getLanguage().equals("JAVA")) {
...
      * Convert a catalog into a string containing all DDL statements.
-     * @param String[] classNames
+     * @param importLines
      * @return String of DDL statements.
      */
-    public static String toSchema(Catalog catalog, String[] importLines)
+    public static String[] toSchema(Catalog catalog, String[] importLines)
     {
-        StringBuilder sb = new StringBuilder();
+        StringBuilder[] schemas = new StringBuilder[2];
```
vtkstef: I would rather you use ImmutableList of StringBuilder or better yet a dedicated two member class where its members describe what you are actually holding and that would encapsulate the append operation as you intend to implement it

akhanzode:
looks good. dont merge until master is branched.

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
miketex: Stefano - I changed this to ImmutableSet, its a Set in VoltDDLElementTracker.

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
miketex: This is now just one StringBuilder with later changes.  I could still wrap this StringBuilder with a Schema object if that seems desirable...

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
miketex: There's only one StringBuilder now.  I didn't change this code, my understanding is that the 'CREATE PROCEDURE' statement created above might or might not be terminated with a ';' depending on which part of the conditional is executed.

PR #3552, opened 2016-05-05
Eng 9656 ban boolean display columns

paulmartel:
Other notable details:
 Clariified actual minimal API of PlanNodeList by replacing largely unused "PlanNodeList is a PlanNodeTree" relationship with a "PlanNodeList has a PlanNodeTree" and some trivial delegation.

Removed pointless serialization of PlanNodeTree parameters -- they are not even deserialized by the EE.
Renamed some planner/*.sql ddl files to follow a naming convention.


src/frontend/org/voltdb/expressions/ComparisonExpression.java
```diff
@@ -40,12 +40,12 @@
 
     public ComparisonExpression(ExpressionType type) {
         super(type);
-        setValueType(VoltType.BIGINT);
+        setValueType(VoltType.BOOLEAN);
```
wolffcm: Glad to see this more sensible return type now.  I guess the sky did not fall when you changed it?

src/frontend/org/voltdb/expressions/HashRangeExpression.java
```diff
@@ -46,7 +46,7 @@
     public HashRangeExpression() {
         super(ExpressionType.HASH_RANGE);
         //See the comment in ConjunctionExpression
-        setValueType(VoltType.BIGINT);
+        setValueType(VoltType.BOOLEAN);
```
wolffcm: I'm puzzling over what this expression actually does.  Could you add a class comment for it?

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -2543,10 +2543,11 @@ public void testDDLCompilerMatView()
                 "create view my_view as select id, count(*), approx_count_distinct(num) from t group by id;";
         checkDDLErrorMessage(ddl, errorMsg);
 
-        // comparison expression not supported in group by clause
-        errorMsg = "Materialized view \"MY_VIEW\" with comparison expression '=' in GROUP BY clause not supported.";
+        // comparison expression not supported in group by clause -- actually gets caught because it's not allowed
+        // in the select list either.
```
wolffcm: Aren't we allowed to have expressions in the select list that are (um) functionally-dependent on GB exprs?  So I could `SELECT gbk1 + gbk2, sum(f) FROM ... GROUP BY gbk1, gbk2`, for example.  For something like that maybe you could group by a Boolean expression but have a CASE expression that uses the GB keys in the select list.  Would that work?

wolffcm: My comment doesn't really apply to this test case, but I'm curious about how general SELECT statements behave.

tests/frontend/org/voltdb/planner/TestPlansInExistsSubQueries.java
```diff
@@ -778,9 +778,10 @@ public void testHSQLFailed() {
                     "subquery with WHERE expression with aggregates on parent columns are not supported");
         }
         {
-            failToCompile("select * from r1 join r2 on exists " +
-                    " (select a from r2 where a > 1) ",
-                    "Join with filters that do not depend on joined tables is not supported in VoltDB");
+            // This seems to have suddenly started working, not at all sure why . Can we trust it?
```
wolffcm: You're changes don't seem like they would affect this, so this seems worth investigating.

wolffcm: The more I think about this, the more puzzled I am.  I'm not clear why we couldn't allow this join on replicated tables in the first place.  If this statement is plannable now, there should be a regressionsuite test locking down a right answer.

src/frontend/org/voltdb/plannodes/AbstractReceivePlanNode.java
```diff
@@ -71,8 +71,8 @@ public boolean isOrderDeterministic() {
     public String nondeterminismDetail() { return m_nondeterminismDetail; }
 
     @Override
-    public boolean reattachFragment( SendPlanNode child  ) {
-        this.addAndLinkChild(child);
+    public boolean reattachFragment(AbstractPlanNode child) {
```
wolffcm: Maybe have an assertion that `child` is a `SendPlanNode` here?  Previously we could have expected a class cast exception if anyone passed something other than a send node.  Having a receive node with a child that is not a send could be confusing to debug.

wolffcm:
I'm done reviewing.  I'd like to feel more confident about the planner test that you disabled if this work is going into 6.3.

wolffcm:
The mysterious planner test that started passing could be a result of changing the return type of some expressions from BIGINT to BOOLEAN?

tests/frontend/org/voltdb/planner/TestPlansInExistsSubQueries.java
paulmartel: OK will add test.

src/frontend/org/voltdb/plannodes/AbstractReceivePlanNode.java
paulmartel: Sure OK, BUT...
since the usage of this function is to reconstitute a two-fragment plan from its fragments for the purposes of "explain", it's kind of a strange almost accidental place to detect such a serious protocol breach.
Also, I have in my back pocket an innovation to eliminate the non-optional yet void-of-useful-information SendPlanNode/Executor classes, migrating their minimal p_execute functionality into either a VoltDBEngine or ExecutorContext method OR MAYBE into a new kind of table (sibling of temptable, streamedtable) that immediately serializes rather than materializes its inserted rows.

tests/frontend/org/voltdb/planner/TestPlansInExistsSubQueries.java
wolffcm: The error message that we're producing here suggests there was an explicit guard against certain join predicates.  Why was it there, and can it go away?

src/frontend/org/voltdb/plannodes/AbstractReceivePlanNode.java
wolffcm: Ok, fair enough.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
paulmartel: I'll make sure there's a test of GB booleans outside matview -- arguably if we guard against it BUT we think we could support it without too much trouble, we can make that another ticket.
The next incremental hurtle may be to allow boolean columns in strictly INTERNAL temp tables that don't need to be serialized into VoltTables -- or maybe we're better off NOT splitting such hairs and instead taking on full support for BOOLEAN columns in VoltTable and probably persistenttable table too -- that is, user schema, not just result sets.

PR #3554, opened 2016-05-06
ENG-10326 tweaks to API and diagnostics of catalog change and stmt pa…

paulmartel:
…ram issues.


BillWhite:
I'm not sure I understand this, but nothing jumps out at me that would cause problems.

There is a lot of sensible refactoring, but the real change seems to be in adding the arrays
m_javaTypeTriedByInsertProc and m_javaTypeTriedByInListProc.  Is that right, or have I
misunderstood?

src/frontend/org/voltdb/planner/PlanSelector.java
```diff
@@ -231,6 +231,7 @@ private void outputPlan(CompiledPlan plan, AbstractPlanNode planGraph, String fi
 
             // get the explained plan for the node
             plan.explainedPlan = planGraph.toExplainPlanString();
+            plan.explainedPlan += plan.expectedParamsExplained();
```
wolffcm: Do you think that the param types should be appended to the explained plan moving forward?  It seems like it would be helpful for us and for customers.

wolffcm:
Bill, the two new arrays are just test changes.

The behavior-affecting changes I see here is that we catch all throwables in the AsyncCompilerAgentHelper, and that we now list parameter types in the explained plan.  Do I have this right?

The changes look good to go to me.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -124,10 +124,17 @@ else if (work.invocationName.startsWith("@AdHoc")) {
                         "original catalog: " + ioe.getMessage();
                     return retval;
                 }
+                catch (Throwable t) {
```
hhakimi53: Thought: will it better to catch specific type of error or exception here. Thinking if server takes fatal error in this path, server may still be kept alive. Also should the error message be logged at error level?

hhakimi53:
Overall changes looks fine.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
paulmartel: Not expecting a fatal error to arise from applying DDL -- I very much doubt that we do or that we should have code on this compiler thread that crashes the server. I believe that this recovery code may keep the compiler service thread from stopping due to a bug, which is probably a good thing. I'm OK with making this error level.

src/frontend/org/voltdb/planner/PlanSelector.java
paulmartel: I'm not sure the parameter type list is normally that useful -- it would probably be more useful to annotate in place the parameters actually mentioned in the explain output something like "filter by intcol < ?1:BIGINT"

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
hhakimi53: Ok, I am fine with that.

hhakimi53:
Looks good, thanks.

paulmartel:
I don't think this was ever merged to the official 6.2.x branch -- it was sufficient to have it on the branch, but 6.3 and later already have equivalent changes.

PR #3555, opened 2016-05-06
Work around some ExpressionLike optimisations.

BillWhite:
ExpressionLike is optimized at one point to change some cases to be
ordinary boolean expressions.  However, the logic for resolving
expression types was not updated to look for these boolean expressions,
and still expected operands to be string expressions.  The resulting
confusion caused unexpected HSQL parsing errors, and therefore planning
errors.

We currently do the computation in a slightly different order.  When the
optimization is done in HSQL, we set the types of the operands, and then
no further resolution is required.

https://issues.voltdb.com/browse/ENG-10016

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
@@ -72,7 +72,6 @@ public final class ExpressionLike extends ExpressionLogical {
 
         this.nodes      = other.nodes;
         this.likeObject = other.likeObject;
-        this.noOptimisation = true;
     }
 
     @Override
@@ -234,7 +233,7 @@ public final class ExpressionLike extends ExpressionLogical {
             /*
              * Can guarantee this won't work with an escape in the EE
              */
-            if (nodes.length > 2) {
+            if (hasEscape()) {
                 throw new RuntimeException("Like with an escape is not supported in parameterized queries");
             }
             return;
@@ -303,7 +302,7 @@ public final class ExpressionLike extends ExpressionLogical {
                 /*
                  * Escape is not supported in the EE yet
                  */
-                if (nodes.length > 2) {
+                if (hasEscape()) {
                     throw new RuntimeException("Like with an escape is not supported unless it is prefix like");
                 }
                 return;
@@ -338,7 +337,7 @@ public final class ExpressionLike extends ExpressionLogical {
                 /*
                  * Escape is not supported in the EE yet
                  */
-                if (nodes.length > 2) {
+                if (hasEscape()) {
                     throw new RuntimeException("Like with an escape is not supported unless it is prefix like");
                 }
                 nodes        = new Expression[BINARY];
@@ -363,7 +362,7 @@ public final class ExpressionLike extends ExpressionLogical {
                 /*
                  * Escape is not supported in the EE yet
                  */
-                if (nodes.length > 2) {
+                if (hasEscape()) {
                     throw new RuntimeException("Like with an escape is not supported unless it is prefix like");
                 }
             }
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
@@ -95,7 +95,7 @@ public final class ExpressionLike extends ExpressionLogical {
     }
 
     private boolean hasEscape() {
-        return (TERNARY <= nodes.length);
+        return (TERNARY <= nodes.length) && (nodes[ESCAPE] != null);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


BillWhite:
This is mostly blue in Jenkins.  On the mac the pcre library is not building, but this change has nothing to do with that, and I suspect the failure to be environmental.  I'm rerunning the test.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
```diff
@@ -100,7 +108,7 @@ public Object getValue(Session session) {
         Object leftValue   = nodes[LEFT].getValue(session);
         Object rightValue  = nodes[RIGHT].getValue(session);
 // A VoltDB extension to disable LIKE pattern escape characters
-        Object escapeValue = (nodes.length < TERNARY || nodes[ESCAPE] == null) ? null
+        Object escapeValue = (!hasEscape() || nodes[ESCAPE] == null) ? null
```
wolffcm: Can't you push the null check for `nodes[ESCAPE]` into your new `hasEscape()` method?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
```diff
@@ -345,6 +358,7 @@ public void resolveTypes(Session session, Expression parent) {
                                                      leftOld, rightBound);
                 opType     = OpTypes.AND;
```
wolffcm: We change the `opType` of `this` to be `AND` here, even though the object's type remains `ExpressionLike`?  Wow, that's really awful.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
```diff
@@ -251,23 +264,23 @@ public void resolveTypes(Session session, Expression parent) {
...
 // A VoltDB extension to disable LIKE pattern escape characters
-        likeObject.setPattern(session, pattern, escape, (nodes.length > 2));
+        likeObject.setPattern(session, pattern, escape, hasEscape());
 /* disable 1 line ...
         likeObject.setPattern(session, pattern, escape, nodes[ESCAPE] != null);
 ... disabled 1 line */
 // End of VoltDB extension
 
-        if (noOptimisation) {
+        if (noOptimisation || simplified) {
```
wolffcm: It doesn't seem possible for a LIKE expression to be simplified before calling `resolveTypes`, so when would `simplified` be true here?  Do we call `resolveTypes` more than once?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
```diff
@@ -72,12 +72,15 @@ private ExpressionLike(ExpressionLike other) {
 
         this.nodes      = other.nodes;
         this.likeObject = other.likeObject;
+        this.noOptimisation = true;
```
wolffcm: Why set this to true here, instead of copying the value from other?

wolffcm:
I'm wondering if the `getSQL` method will do the right thing in the case of simplified LIKE expressions.  We use `getSQL` in DDL statements that include expressions, such as in expression indexes, or partial indexes, or in the DELETE statement of row limit constraint.  Do you think there are existing bugs here?

wolffcm:
That's all my comments.

BillWhite:
I wondered about getSQL.  I think it will not return the like expression, but it will make it into the string comparison we have turned it into.  That's not necessarily wrong, though it seems peculiar.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
BillWhite: Done.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
BillWhite: If resolveTypes returned an AbstractExpression we could replace it.  But this is not in general possible.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
BillWhite: Hmm.  That seems right.  I don't remember why I thought this was necessary.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
BillWhite: Mostly because I always want noOptimization to be true.  We do the same optimization in the planner, according to Paul, so it really doesn't need to be done here.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
```diff
@@ -204,22 +208,29 @@ public void resolveTypes(Session session, Expression parent) {
                    && (nodes[ESCAPE] == null
                        || nodes[ESCAPE].dataType.isBinaryType())) {
             likeObject.isBinary = true;
-        } else {
+        } else if (false == (nodes[LEFT].dataType.isBooleanType()
+                              && nodes[RIGHT].dataType.isBooleanType())
+                              && dataType.isBooleanType()) {
+            // If both argument nodes are boolean we have resolved
+            // this before.  So, this is ok.  Otherwise, this is not
+            // properly typed.
```
paulmartel: Is this boolean typing really a thing or should it have been backed out with other changes?

paulmartel:
End of review round.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
BillWhite: This is actually the fix to the ticket.  At this point we will be resolving this for the second time.  I'm not sure why.  The first time we change the ExpressionLike node to be an AND expression node with two substring comparisons as parameters.  This test catches that case.

BillWhite:
This is all Blue on Jenkins, except one test which runs manually on my Linux test machine.

I'm going to squash merge this separately.

BillWhite:
Squash merged.

PR #3556, opened 2016-05-06
ENG-10134: Add a base class in community for the DRProducer to handle…

wweiss-voltdb:
… Statistics calls when the DRProducer is not running.


slowsoul:
Review finished.

PR #3559, opened 2016-05-09
Fix issue with materialized views and alter table

wolffcm:
This was a bug in which we saw assertion failures when instantiating an index scan executor in the EE, because the required index seemed to be missing.

This issue happened because a materialized view had a "fallback executor vector" which was being instantiated from the catalog JSON while a schema change was in progress (add or drop column).  At this time, the TableCatalogDelegate pointed to the stale, existing table which had already had its indexes removed (and recreated on the new table).

The fix was to set the table referred to by the TableCatalogDelegate to the new table before migrating the views.

The first commit has the actual behavior-changing modifications that I made, and in the second commit I just moved some methods around for the sake of consistency with other functions in the file.


paulmartel:
Looks good.

PR #3561, opened 2016-05-10
ENG-10135: DR small table truncation

nshi:
* An optimization that uses tuple-by-tuple deletion on tables with low number of rows instead table swapping for truncation caused DR to generate a large binary log, sometimes over the 45MB limit. I'm putting a guard around that to prevent it from generating DR delete records in the binary log when it should actually be a truncation record. Regardless of the optimization, it will always generate a single truncation record in the binary log.

* Provide some context when the DR buffer size limit is reached.

### DIFF
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -78,6 +78,7 @@
 #include "storage/PersistentTableUndoDeleteAction.h"
 #include "storage/PersistentTableUndoTruncateTableAction.h"
 #include "storage/PersistentTableUndoUpdateAction.h"
+#include "storage/DRTupleStreamUndoAction.h"
 #include "storage/ConstraintFailureException.h"
 #include "storage/TupleStreamException.h"
 #include "storage/CopyOnWriteContext.h"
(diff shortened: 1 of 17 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -321,10 +322,29 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+        }
+    }
+
     // If the table has only one tuple-storage block, it may be better to truncate
     // table by iteratively deleting table rows. Evalute if this is the case
     // based on the block and tuple block load factor
     if (m_data.size() == 1) {
+        // Temporarily disable DR binary logging so that we don't turn the
+        // truncate into a bunch of delete records.
+        DRTupleStreamDisableGuard drGuard(ec, false);
```
nshi: Yes, that's correct. I thought about putting it in `deleteAllTuples()`, but that feels too generic if it's not called from here. But maybe that's what we want, always record a truncation record if `deleteAllTuple()` is called?

wweiss-voltdb:
Review Complete

PR #3562, opened 2016-05-10
fix the classpath template

zeemanhe:


zeemanhe:
add libs to the classpath template

PR #3563, opened 2016-05-10
Handle stack overflow error condition for queries due to deep expression tree

hhakimi53:
If the query's where clause is composed of 240+ predicate expressions, when parsing these expressions via mutual recursion, thread stack can grow to it's maximum capacity, resulting in stack overflow error. The optimal solution to avoid this is to convert the mutual recursion into tail-recursion or iterative approach. Though the logic change needed are at multiple places like generating unique id for the expression during construction of volt XML elements for plan generation and converting the expression tree of the plan to json-string for hash computation. So at present handle stack overflow error condition which arises in such scenario.


src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariableResolver.java
```diff
@@ -246,6 +248,21 @@ void assignToLists(Expression e, HsqlArrayList[] expressionLists,
 
     void expandConditions() {
         expandConditions(whereExpressions, false);
+        for (HsqlArrayList whereExpression : whereExpressions) {
+            if (whereExpression.size() > whereExpressionElementsUpperBound) {
```
BillWhite: Hasn't HSQL already parsed the expressions here?  Doesn't this mean the parser will have already crashed?

BillWhite:
Do we have a test for stored procedures?

This all looks ok to me, but I think I'd like to know if stored procedures had the same problems.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariableResolver.java
hhakimi53: This guard will get triggered when HSQL is parsing the sql to compile sql to "Statement" during parsing which is used for constructing Volt XML element tree. In parsing logic, stack overflow error happens when fetching Volt XML element from compiled Statement. 

Below is stack trace in call heirarchy where stack over flow gets triggered:

`java.lang.StackOverflowError
    at org.hsqldb_voltpatches.Expression$1.traverse(Expression.java:1949)
    at org.hsqldb_voltpatches.Expression.getUniqueId(Expression.java:1954)
    at org.hsqldb_voltpatches.Expression$1.traverse(Expression.java:1951)
...
...
    at org.hsqldb_voltpatches.Expression$1.traverse(Expression.java:1951)
    at org.hsqldb_voltpatches.Expression.getUniqueId(Expression.java:1954)
    at org.hsqldb_voltpatches.Expression.voltGetXML(Expression.java:1637)
...
    at org.hsqldb_voltpatches.StatementDMQL.voltGetXMLExpression(StatementDMQL.java:920)
    at org.hsqldb_voltpatches.StatementQuery.voltGetStatementXML(StatementQuery.java:131)
    at org.hsqldb_voltpatches.HSQLInterface.getXMLCompiledStatement(HSQLInterface.java:332)
    at org.voltdb.planner.QueryPlanner.parse(QueryPlanner.java:170)
    at org.voltdb.compiler.StatementCompiler.compileStatementAndUpdateCatalog(StatementCompiler.java:189)
    at org.voltdb.compiler.StatementCompiler.compileFromSqlTextAndUpdateCatalog(StatementCompiler.java:307)
    at org.voltdb.compiler.ProcedureCompiler.compileSingleStmtProcedure(ProcedureCompiler.java:705)
....
`

hhakimi53:
Stored procedure also has same issue. Positive and negative test for stored procedure has been added to sqlcmd tests. I will add unit test also for the same. Thnxs

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariableResolver.java
```diff
@@ -74,6 +74,8 @@
     MultiValueHashMap map               = new MultiValueHashMap();
     int               inExpressionCount = 0;
     boolean           hasOuterJoin      = false;
+    // max number of predicates in where expression
+    final int         whereExpressionElementsUpperBound = 232;
```
wolffcm: I find it helps readability to define constants in all caps, like `WHERE_EXPRESSION_ELEMENTS_UPPER_BOUND`.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariableResolver.java
wolffcm: Does this guard only apply to WHERE conditions?  It's the most obvious place for deep expression trees, but not the only one.  It seems like we need to limit expression tree depth anywhere that expressions can occur.  For example select list (consider a CASE WHEN expression with a really large conditionals), ON clause, HAVING, and so on.

wolffcm:
That's all my comments.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariableResolver.java
hhakimi53: Yes, currently the guard has been established for Where conditions only. I see, so you are thinking of system is still potentially exposed for the cases where expression count can grow similar to where clause. I hadn't though of those expressions though I had looked into establishing the recursion depth during mutual recursive calls but due to the nature of mutual recursion, couldn't do it there - else we could have applied it generically. The places may be can expand it to are probably having, order by and join condition expression (may be StatementDMQL or StatementDML). For those can evaluate the tree depth (currently logic is max childs in the tree) and guard against if tree depth on those expression grows beyond certain threshold. Thoughts?

wolffcm: I think a generic check would be best if at all possible.  There are lots of places where expressions can appear, and it would be easy to miss some.

You could add a new field to Expression that is the number of levels in the tree, that is computed whenever a building an expression.  Any time an expression is constructed, the new number of levels is the max(number_of_levels_in_nth_child_tree) + 1.  Not sure if this is feasible.

hhakimi53: Thanks Chris for the feedback. I will look into if we can have a generic check.

wolffcm:
Seems like the files sqlcmd1.* could get more descriptive names, maybe "stackoverflow.*" or something?

That's my only comment.  Looks good otherwise.

hhakimi53:
Thanks Chris for reviewing the changes. Sure, I will update the file names.

PR #3569, opened 2016-05-10
Eng 8872 block changes

hhakimi53:
- Update persistent table logic to allocate tuple storage block to table on table creation. Don't release the only remaining tuple storage block back to system during truncate or delete operation. This will prevent persistent table to request fresh tuple storage block on next row insertion into empty table. 
- Updated snapshot save logic so that when system is performing snapshot save, it takes into account the persistent table can contain empty tuple storage

### DIFF
--- a/src/ee/storage/CopyOnWriteIterator.cpp
+++ b/src/ee/storage/CopyOnWriteIterator.cpp
@@ -69,6 +69,10 @@ bool CopyOnWriteIterator::needToDirtyTuple(char *tupleAddress) {
         return false;
     }
 
+    // code will be removed. Not for check-in: for debugging purpose only
+    if (m_currentBlock == NULL) {
+        StackTrace::printStackTrace();
+    }
     assert(m_currentBlock != NULL);
     /**
      * Now check where this is relative to the COWIterator.
--- a/src/ee/storage/CopyOnWriteIterator.cpp
+++ b/src/ee/storage/CopyOnWriteIterator.cpp
@@ -57,10 +57,6 @@ CopyOnWriteIterator::CopyOnWriteIterator(
  * it skiped a dirty tuple and didn't end up with the right found tuple count upon reaching the end.
  */
 bool CopyOnWriteIterator::needToDirtyTuple(char *tupleAddress) {
-    if (m_currentBlock == NULL) {
-        return false;
-    }
-
     /**
      * Find out which block the address is contained in. Lower bound returns the first entry
      * in the index >= the address. Unless the address happens to be equal then the block
--- a/src/ee/storage/CopyOnWriteIterator.cpp
+++ b/src/ee/storage/CopyOnWriteIterator.cpp
@@ -28,6 +28,7 @@ CopyOnWriteIterator::CopyOnWriteIterator(
         m_location(NULL),
         m_blockOffset(0),
         m_currentBlock(NULL),
+        m_emptyTable(false),
         m_skippedDirtyRows(0),
         m_skippedInactiveRows(0) {
 
@@ -36,6 +37,7 @@ CopyOnWriteIterator::CopyOnWriteIterator(
         // has empty tuple storage block associated with it. So no need
         // to set it up for snapshot
         m_blockIterator = m_end;
+        m_emptyTable = true;
         return;
     }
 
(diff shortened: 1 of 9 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -340,17 +340,17 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
-        const double tableLFCutoffForTrunc = 0.105666;
+        const double tableWithNoViewLFCutoffForTrunc = 0.105666;
         //cut-off for table with views
         const double tableWithViewsLFCutoffForTrunc = 0.015416;
 
         const double blockLoadFactor = m_data.begin().data()->loadFactor();
-        if ((blockLoadFactor <= tableLFCutoffForTrunc) ||
-            (m_views.size() > 0 && blockLoadFactor <= tableWithViewsLFCutoffForTrunc)) {
+        if ((m_views.empty() && (blockLoadFactor <= tableWithNoViewLFCutoffForTrunc)) ||    // table with no view
+            (!m_views.empty() && (blockLoadFactor <= tableWithViewsLFCutoffForTrunc))) {    // table with views
```
paulmartel: There's no need to test m_views.empty() twice. If the second attempt has not been short-circuited out, then we know that m_views.empty() implies blockLoadFactor > tableWithNoViewLFCutoffForTrunc > tableWithViewsLFCutoffForTrunc, so all that we need to test after the || is (blockLoadFactor <= tableWithViewsLFCutoffForTrunc)

src/ee/storage/persistenttable.cpp
```diff
@@ -340,17 +340,17 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
             return deleteAllTuples(true);
         }
     }
-    //For MAT view dont optimize needs more work.
+    // For MAT view don't optimize, needs more work.
```
paulmartel: There should be more detail in this comment. Is there a bug number for the issue with truncate and streamed table views?

src/ee/storage/persistenttable.cpp
hhakimi53: Yes, I agree, the condition "!m_views.empty()" is redundant and the reason I had put in there for explicit readability. I can remove it.

src/ee/storage/persistenttable.cpp
hhakimi53: In here update was the spelling and spacing in the comment. I think there was a ticket for which the above workaround was implemented for streamed table for truncate scenario.

paulmartel:
End of review round.

src/ee/storage/CopyOnWriteIterator.cpp
```diff
@@ -48,6 +57,10 @@ CopyOnWriteIterator::CopyOnWriteIterator(
  * it skiped a dirty tuple and didn't end up with the right found tuple count upon reaching the end.
  */
 bool CopyOnWriteIterator::needToDirtyTuple(char *tupleAddress) {
+    if (m_currentBlock == NULL) {
```
nshi: Could this ever happen?

hhakimi53: When we were working on tuple block changes, there were couple of snapshot tests that were failing. The reason was the first row insert for the table between window snapshot getting activated and getting started (during this scenario m_currentBlock is marked null) and what I recall is we don't want to dirty any new tuple insert as table was empty. So I had added this check-in there else it was crashing the system.

src/ee/storage/persistenttable.h
```diff
@@ -246,7 +246,7 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
     // ------------------------------------------------------------------
     // GENERIC TABLE OPERATIONS
     // ------------------------------------------------------------------
-    virtual void deleteAllTuples(bool freeAllocatedStrings, bool fallible = true);
+    virtual void deleteAllTuples(bool, bool fallible = true);
```
nshi: Why not just remove the first argument completely?

nshi:
If a table is truncated without using the optimized truncation path, does the new table have a preallocated block?

nshi:
Review finished.

src/ee/storage/persistenttable.h
paulmartel: @nshi To remove the argument, you need to remove it from all implementations of the virtual function, which means making sure that it is never used in other implementations and/or always given the same value by the callers -- both of these things are ALMOST true and SHOULD BE true in the sense that the only contrary cases come from awkward stats code that could be made simpler/smarter -- but I think that's an innovation for another day.

nshi: @paulmartel That makes sense. It would be easier if they are of different types that the compiler can check for you. I've seen EE code being called with completely wrong arguments but the compiler happily casted integer type to boolean.

hhakimi53:
Yes, even if the table is truncated without using the optimized truncation, the new table will have pre-allocated tuple storage block associated with it. As part of the change-list, table factory generates new persistent table with an empty tuple storage block associated with persistent table.

nshi:
Review finished.

hhakimi53:
Closing the branch. Will merge the changes through a new branch.

PR #3571, opened 2016-05-11
ENG-10303 add db version info to log

zeemanhe:


src/frontend/org/voltdb/utils/PlatformProperties.java
```diff
@@ -237,7 +238,7 @@ public String toLogLines() {
 
         sb.append(String.format("OS PROFILE:       %s %s %s %s\n",
                                 osName, osVersion, osArch, locale));
-
+        sb.append(String.format("DB VERSION:       %s\n", VoltDB.instance().getVersionString()));
```
vtkstef: Where are is PlatformProperties instantiated? can we pass the version string to its constructor instead of referencing VoltDB.instance() from such an elementary class implementation?

zeemanhe: The constructor is  protected, a singleton, may be instantiated in various places.  Would it be better to pass in as arg param in PlatformProperties.toLogLines()?

zeemanhe: update: pass db version to toLogLines().

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1802,7 +1802,7 @@ void logDebuggingInfo(int adminPort, int httpPort, String httpPortExtraLogMessag
 
         // print out a bunch of useful system info
         PlatformProperties pp = PlatformProperties.getPlatformProperties();
-        String[] lines = pp.toLogLines().split("\n");
+        String[] lines = pp.toLogLines(VoltDB.instance().getVersionString()).split("\n");
```
vtkstef: RealVoltDB is VoltDB.instance() so getVersionString() alone should suffice

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -771,7 +771,7 @@ public static void crashLocalVoltDB(String errMsg, boolean stackTrace, Throwable
                     writer.println();
                     writer.println("Platform Properties:");
                     PlatformProperties pp = PlatformProperties.getPlatformProperties();
-                    String[] lines = pp.toLogLines().split("\n");
+                    String[] lines = pp.toLogLines(VoltDB.instance().getVersionString()).split("\n");
```
vtkstef: you are in VoltDB instance() alone will suffice

PR #3574, opened 2016-05-11
ENG-10335: Backport ENG-10194 to V6.2.x

JadenW:
* ENG-10194: disable @StopNode while rejoining node still in process

* address the review

* safe error handling for check rejoin nodes existence


vtkstef:
Reviewed

PR #3580, opened 2016-05-12
ENG-10393-fix null pointer

zeemanhe:


src/frontend/org/voltdb/importclient/kinesis/KinesisStreamImporter.java
```diff
@@ -308,19 +307,22 @@ public void clientCallback(ClientResponse response) throws Exception {
...
             }
             lagLen = leeway;
+            checkpoints = new BigInteger[(int)m_config.getMaxReadBatchSize()];
         }
 
-        synchronized void allocate(int capacity) {
+        synchronized void resetTo() {
 
-            chceckpoints = (BigInteger[]) Array.newInstance(BigInteger.class, capacity);
+            for(int i = 0; i < checkpoints.length; i++){
```
akhanzode: You can use Arrays.fill()

src/frontend/org/voltdb/importclient/kinesis/KinesisStreamImporter.java
```diff
@@ -166,7 +165,7 @@ public void processRecords(ProcessRecordsInput records) {
             }
 
             BigInteger seq = BigInteger.ZERO;
-            m_gapTracker.allocate(records.getRecords().size());
+            m_gapTracker.resetTo();
```
akhanzode: Why do you need to reset? Can you put comment also why gap tracker needs to be zeroed?

zeemanhe: The offsets are used for reverse lookup for real offsets in the checkpoints array. so for every batch, checkpints need to be reset.

PR #3581, opened 2016-05-13
ENG-10137: added tests of new Streams tab, 

johnTcrawford:
with support in the relevant page class (in SqlQueryPage.groovy, SqlQueriesTest.groovy, new
expectedStreams.txt); fixed tests of the Stored Procedures tab, by
moving the new-ish Default Stored Procs for Export tables, to where they
now correctly are (in the list of Default Stored Procs), from where they
used to incorrectly be (in the list of User Stored Procs) - in
expectedDefaultStoredProcs.txt, expectedUserStoredProcs.txt.


tests/geb/vmc/src/tests/SqlQueriesTest.groovy
```diff
@@ -214,6 +222,19 @@ class SqlQueriesTest extends SqlQueriesTestBase {
     }
 
     /**
+     * Returns the list of stream names, as displayed on the page, but saving
+     * the list for later, so you don't need to get it over and over again.
+     * @param sqp - the SqlQueryPage from which to get the list of stream names.
+     * @return the list of stream names, as displayed on the page.
+     */
+    static List<String> getStreams(SqlQueryPage sqp) {
```
wolffcm: I don't see this actually being called from anywhere?

I know little of GEB testing though.  I also don't see the code where we're actually comparing expected and actual values... what am I missing?

wolffcm:
I just have the one comment.

tests/geb/vmc/src/tests/SqlQueriesTest.groovy
johnTcrawford: It's called by the checkStreams() (test) method. Easy to miss - I thought for a second it really was not called anywhere, but it is, and that test does pass.

PR #3588, opened 2016-05-16
ENG-10259-licensecheck-add-go-support

miketex:
Initial pass at adding support for go to licensecheck.py

### DIFF
--- a/tools/approved_licenses/gpl3_voltdb_go.txt
+++ b/tools/approved_licenses/gpl3_voltdb_go.txt
@@ -1,0 +1,15 @@
+// This file is part of VoltDB.
+// Copyright (C) 2008-2016 VoltDB Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


tools/approved_licenses/gpl3_voltdb_go.txt
```diff
@@ -0,0 +1,15 @@
+// This file is part of VoltDB.
```
jhugg: Can we use the gp3_voltdb.txt license in golang? Why do we need to use this custom one?

tools/licensescheck.py
```diff
@@ -56,6 +58,18 @@ def verifyLicense(f, content, approvedLicensesJavaC, approvedLicensesPython):
         if licenseStartsHere(content, approvedLicensesPython):
             return 0
         print "ERROR: \"%s\" does not start with an approved license." % f
+    elif f.endswith('.go'):
```
jhugg: Should the go code be different than c++ or java code? Is there a problem using "/*" in go?

jhugg:
My general comment is why does Golang have to be different from Java or C++?

tools/licensescheck.py
miketex: No, go also supports block comments, meaning "/*".  I followed the example of the go src code, which uses line comments.  But looking at https://golang.org/doc/effective_go.html#commentary, it doesn't seem that there's really a preference for line comments over block comments, so I'll make that change.

miketex: On second thought, I have some doubts about how this might really work, given that the go client is in it's own repo...  I imagined that licensecheck.py could be symlinked from the voltdb repository.  But it will still then miss the license files themselves - they would need to be symlinked as well.  An alternative is to make a smaller licensecheck.py in the go client repository and copy the one applicable license file as well...

PR #3589, opened 2016-05-17
Eng 10348

benjaminballard:
This is a test app for ENG-10348, to test changes made in ENG-8872 and to provide a baseline for any future changes that may affect inserting the first row into an empty table, and deleting the last row from a table.


johnTcrawford:
Looks good.

PR #3590, opened 2016-05-17
ENG-10431 Enable VoltDB build on Ubuntu 16.04 in backward compatible way

pdshaw:
Small set of changes so VoltDB community and/or pro will build and run on the newest Ubuntu, 16.04, while continuing to build and run on older CentOS/RHEL 6 and Ubuntu 12.04 without further source changes or preprocessor conditionals.

It's not necessarily a requirement to avoid preprocessor fencing but it keeps things simpler, though at the expense of a broad namespace "using".

### DIFF
--- a/third_party/cpp/google-s2-geometry/geometry/include/s2geo/util/math/mathlimits.h
+++ b/third_party/cpp/google-s2-geometry/geometry/include/s2geo/util/math/mathlimits.h
@@ -190,13 +190,21 @@ DECL_UNSIGNED_INT_LIMITS(unsigned long long int)
   static bool IsPosInf(const Type x) { return _fpclass(x) == _FPCLASS_PINF; } \
   static bool IsNegInf(const Type x) { return _fpclass(x) == _FPCLASS_NINF; }
 #else
-  using namespace std;
+#if __cplusplus >= 201103L
+#define DECL_FP_LIMIT_FUNCS \
+  static bool IsFinite(const Type x) { return !std::isinf(x)  &&  !std::isnan(x); } \
+  static bool IsNaN(const Type x) { return std::isnan(x); } \
+  static bool IsInf(const Type x) { return std::isinf(x); } \
+  static bool IsPosInf(const Type x) { return std::isinf(x)  &&  x > 0; } \
+  static bool IsNegInf(const Type x) { return std::isinf(x)  &&  x < 0; }
+#else
 #define DECL_FP_LIMIT_FUNCS \
   static bool IsFinite(const Type x) { return !isinf(x)  &&  !isnan(x); } \
   static bool IsNaN(const Type x) { return isnan(x); } \
   static bool IsInf(const Type x) { return isinf(x); } \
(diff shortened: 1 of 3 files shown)


third_party/cpp/google-s2-geometry/geometry/include/s2geo/util/math/mathlimits.h
```diff
@@ -190,6 +190,7 @@ DECL_UNSIGNED_INT_LIMITS(unsigned long long int)
   static bool IsPosInf(const Type x) { return _fpclass(x) == _FPCLASS_PINF; } \
   static bool IsNegInf(const Type x) { return _fpclass(x) == _FPCLASS_NINF; }
 #else
+  using namespace std;
```
wolffcm: I guess I'd prefer not to import all of the `std` namespace into this file.  If you just preface calls to `isinf` and `isnan` on the following lines with `std::`, will it compile?  Alternatively you could add the same `using` lines that you added to exactfloat.cc.

That's all my comments.

pdshaw: Tried those approaches in a concerted effort to avoid pulling in the whole namespace.

The "using std::is{inf|nan}" builds fine with 16.04's default c++ environment but the CentOS world choked.

Maybe fencing IS the best approach since I too realize the "using namespace..." approach is not ideal.

wolffcm: What do you mean by fencing?

It looks like std::isnan and std::isinf were added in C++11, so naturally older compilers will not know them.  But I guess older versions of g++ do have `::isnan` and `::isinf`.

There must be a preprocessor symbol that will let you detect the compiler version?

http://stackoverflow.com/questions/259248/how-do-i-test-at-compile-time-the-current-version-of-gcc

I'm not sure in which version isnan and isinf moved to the std namespace though.

paulmartel: I'm also having trouble with terminology: "fencing"? "choked"?
I'm having trouble understanding how the using::is* trick can work fine for exactfloat.cc but not work in the .h -- unless the .h is getting included in something being built with strange compiler settings?
Peter, can you share here or via e-mail what the choking build output looks like?
WRT what version of the compiler to test for, we don't have to be all that exact -- we could probably get by with a crude test for "any version equal or prior to the one shipped with CentOS 6" and rely on other current and future builds being well past that.

pdshaw: Fencing: Wrap changes needed for newer compiler (or reverse) in #if -- #endif so we can keep namespace/scope related changes minimal.

Choked: Build failed with errors related to scope operator changes to "isinf" and "isnan"

I'll put together a email with specific errors seen in CentOS 6 or Ubuntu 16.04 builds with their respective compilers.

PR #3592, opened 2016-05-17
streamline streams especially with views now refactored in place

paulmartel:
Still to do: reorganize the refactored mat view code into separate compilation units -- or at least group by class and functional area.


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -787,20 +789,20 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp)
             m_delegatesByName[table->name()] = tcd;
 
             // set export info on the new table
-            if (tcd->exportEnabled()) {
-                tcd->getTable()->setSignatureAndGeneration(catalogTable->signature(), timestamp);
-                m_exportingTables[catalogTable->signature()] = tcd->getTable();
-                StreamedTable *streamedtable = dynamic_cast<StreamedTable *>(tcd->getTable());
-                //Deal with views if we have
+            StreamedTable *streamedtable = tcd->getStreamedTable();
```
paulmartel: Testing `if (streamedtable)` is preferable to `tcd->exportEnabled()` because the latter factors in the state of export streams in the current deployment, which does not really matter here.

paulmartel:
To be merged on better-named branch.

PR #3596, opened 2016-05-18
ENG-10164: duplicate column bug

wolffcm:
This set of changes follows through on implementing a "differentiator" field for various data structures in the planner that represent table columns.  They're really only used as a "tie breaker" for the case when there is a `SELECT *`, and the thing we're selecting from is a subquery.  (Note that HSQL treats parenthesized joins in the FROM clause as if they were subqueries.)

### DIFF
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -765,7 +765,7 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
         // materialized views (which use the parsed select statement but
         // don't go through the planner pass that does more involved
         // column index resolution).
-        col.index = m_displayColumns.size();
+        col.index = index;
 
         insertAggExpressionsToAggResultColumns(m_aggregationList, col);
         if (m_aggregationList.size() >= 1) {
@@ -988,9 +988,7 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
             m_distinctGroupByColumns.add(pcol);
 
             // ParsedColInfo, TVE, SchemaColumn, NodeSchema ??? Could it be more complicated ???
-            SchemaColumn schema_col = new SchemaColumn(
-                    col.tableName, col.tableAlias, col.columnName, col.alias, tve, col.differentiator);
-            m_distinctProjectSchema.addColumn(schema_col);
+            m_distinctProjectSchema.addColumn(col.asSchemaColumn());
         }
 
         return groupbyElement;
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -466,7 +466,7 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
             if (hasComplexAgg()) {
                 expr = col.expression.replaceWithTVE(aggTableIndexMap, indexToColumnMap);
             }
-            SchemaColumn schema_col = new SchemaColumn(col.tableName, col.tableAlias, col.columnName, col.alias, expr);
+            SchemaColumn schema_col = new SchemaColumn(col.tableName, col.tableAlias, col.columnName, col.alias, expr, col.differentiator);
             m_projectSchema.addColumn(schema_col);
         }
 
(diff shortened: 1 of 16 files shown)


src/frontend/org/voltdb/planner/parseinfo/StmtSubqueryScan.java
```diff
@@ -70,9 +71,9 @@ public StmtSubqueryScan(AbstractParsedStmt subqueryStmt, String tableAlias, int
         int i = 0;
         for (ParsedColInfo col: ((ParsedSelectStmt)subqueryStmt).displayColumns()) {
             String colAlias = col.alias == null? col.columnName : col.alias;
-            SchemaColumn scol = new SchemaColumn(col.tableName, col.tableAlias, col.columnName, col.alias, col.expression);
+            SchemaColumn scol = new SchemaColumn(col.tableName, col.tableAlias, col.columnName, col.alias, col.expression, col.differentiator);
```
paulmartel: Simplifying this line to "SchemaColumn scol = col.asSchemaColumn();" may insulate against future maintenance edits like this.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -466,7 +466,7 @@ private void placeTVEsinColumns () {
             if (hasComplexAgg()) {
                 expr = col.expression.replaceWithTVE(aggTableIndexMap, indexToColumnMap);
             }
-            SchemaColumn schema_col = new SchemaColumn(col.tableName, col.tableAlias, col.columnName, col.alias, expr);
+            SchemaColumn schema_col = new SchemaColumn(col.tableName, col.tableAlias, col.columnName, col.alias, expr, col.differentiator);
```
paulmartel: This is ALMOST `SchemaColumn schema_col = col.asSchemaColumn();` except that `expr` is not `col.expression` but rather its re-write, `col.expression.replaceWithTVE(...)`. I am ASSUMING that things would break if we tried to simplify it by setting col.expression to its rewrite, but I don't KNOW that for a fact. So then, maybe this line should stay as you have it UNLESS we add a method/overload like `ParsedColInfo.asSchemaColumn(AbstractExpression` overrideExpr);` but probably not worth it?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -983,7 +989,7 @@ private VoltXMLElement processDistinct(VoltXMLElement displayElement,
 
             // ParsedColInfo, TVE, SchemaColumn, NodeSchema ??? Could it be more complicated ???
             SchemaColumn schema_col = new SchemaColumn(
-                    col.tableName, col.tableAlias, col.columnName, col.alias, tve);
+                    col.tableName, col.tableAlias, col.columnName, col.alias, tve, col.differentiator);
```
paulmartel: This ALMOST looks like another caller for the hypothetical `ParsedColInfo.asSchemaColumn(AbstractExpression overrideExpr);` but I think that it may actually be replaceable by `SchemaColumn schema_col = pcol.asSchemaColumn()` IF we can safely first assign `pcol.differentiator = col.differentiator;`

Browsing the code to look for everywhere we create ParsedColInfos, there are code paths where we let the differentiator default to -1, and this is probably fine, but it's hard to tell whether these values "benefit" from having the -1 value or if they are just "don't care" cases.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -689,12 +689,14 @@ static void parseLimitAndOffset(VoltXMLElement limitNode, VoltXMLElement offsetN
...
+        int index = 0;
         for (VoltXMLElement child : columnsNode.children) {
-            parseDisplayColumn(child, isDistributed);
+            parseDisplayColumn(index, child, isDistributed);
+            ++index;
         }
     }
 
-    private void parseDisplayColumn(VoltXMLElement child, boolean isDistributed) {
+    private void parseDisplayColumn(int index, VoltXMLElement child, boolean isDistributed) {
```
paulmartel: This routine had some pent up demand for the `index` parameter that you added, satisfied by using a "clever" proxy calculation. Change the hacky `col.index = m_displayColumns.size();` below to use index `col.index = index;`.

paulmartel:
End of review round.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
wolffcm: We really only care in the places where `*` can appear with the meaning "all the columns".  That's only the SELECT list.  I'll make the change you suggest, and if it has issues, then existing tests should catch problems.

PR #3597, opened 2016-05-19
Change the misleading error message.

yzhang1991:
VoltDB does not support WHERE clauses containing only constants
=====>
VoltDB does not support constant boolean values

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
@@ -1691,7 +1691,7 @@ public class Expression {
                 // Going forward, it seems to make more sense to leverage the surviving VoltDB code path by hard-wiring here:
                 // valueType="BIGINT", value="1"/"0".
                 throw new org.hsqldb_voltpatches.HSQLInterface.HSQLParseException(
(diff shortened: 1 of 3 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1691,7 +1691,7 @@ VoltXMLElement voltGetXML(Session session, List<Expression> displayCols,
                 // Going forward, it seems to make more sense to leverage the surviving VoltDB code path by hard-wiring here:
                 // valueType="BIGINT", value="1"/"0".
                 throw new org.hsqldb_voltpatches.HSQLInterface.HSQLParseException(
-                        "VoltDB does not support WHERE clauses containing only constants");
+                        "VoltDB does not support constant boolean values");
```
wolffcm: I would end the message with "constant Boolean values, like TRUE or FALSE" just so the message is clearer to the user.

Also, (this is minor) the word "Boolean" should start with a capital letter since it's named after George Boole.

Looks good otherwise.

Does this pass Jenkins?

yzhang1991:
There are two tests failing, both related to the error message, I have updated them and let's wait for Jenkins to finish.

PR #3601, opened 2016-05-20
Fix build error on Ubuntu 16.04

wolffcm:
Seems like on newer platforms isinf and isnan have moved to
namespace std.


wolffcm:
Reviewed on a separate branch.

PR #3603, opened 2016-05-20
ENG-10419 streamline streams especially with views now refactored in …

paulmartel:
…place


tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -132,25 +133,23 @@ class MockVoltDBEngine : public VoltDBEngine {
...
     }
-    ~MockVoltDBEngine() {
-        delete m_conflictExportTable;
-    }
+    ~MockVoltDBEngine() { }
 
-    bool getIsActiveActiveDREnabled() const { return m_isActiveActiveEnabled; }
-    void setIsActiveActiveDREnabled(bool enabled) { m_isActiveActiveEnabled = enabled; }
-    Table* getPartitionedDRConflictTable() const{ return m_conflictExportTable; }
+    Table* conflictTable() const { return m_conflictExportTable.get(); }
```
yzhang1991: Would it be better we use more consistent names for the member and the getter function?
Do you think DRConflictTable is more precise here? I think exportTable actually refers to something else?

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -506,7 +510,15 @@ class DRBinaryLogTest : public Test {
         return temp_tuple;
     }
 
-    void createUniqueIndex(Table* table, int indexColumn, bool isPrimaryKey = false) {
+    void createUniqueIndexes() {
+        createUniqueIndexes(m_table);
+        createUniqueIndexes(m_tableReplica);
+    }
```
yzhang1991: Will it be a better practise to have spaces between functions even for the ones that are closely related?

tests/ee/indexes/index_test.cpp
```diff
@@ -798,8 +798,7 @@ TEST_F(IndexTest, TupleKeyUnique) {
     setWideIndexToRow(searchkey, 2);  // DELETE row 2.
     EXPECT_TRUE(index->moveToKey(&searchkey, indexCursor));
     tuple = index->nextValueAtKey(indexCursor);
-    bool deleted = table->deleteTuple(tuple, true);
-    EXPECT_TRUE(deleted);
+    table->deleteTuple(tuple, true);
```
yzhang1991: I am not sure if we have a good reason not to test the result of the `deleteTuple()` function.
I thought we were going to have something like:
```cpp
EXPECT_TRUE(table->deleteTuple(tuple, true));
```

src/ee/storage/table.h
```diff
@@ -119,22 +119,13 @@ class Table {
     // ACCESS METHODS
```
yzhang1991: I cannot add a comment outside your changed region, but can you correct a typo in line 99 of this file?
`Table lifespan can be managed bya reference count`, `bya` => 'by a'

src/ee/storage/table.cpp
```diff
@@ -427,21 +420,14 @@ bool Table::serializeTupleTo(SerializeOutput &serialize_io, voltdb::TableTuple *
     return true;
 }
 
-bool Table::equals(voltdb::Table *other) {
+bool Table::equals(voltdb::Table *other)
+{
```
yzhang1991: Move this brace to the line above.

src/ee/storage/temptable.h
```diff
@@ -211,7 +211,6 @@ inline void TempTable::deleteAllTuplesNonVirtual(bool freeAllocatedStrings) {
     }
```
yzhang1991: I think we should delete line 106-108 in this file:
```cpp
// Deleting a tuple from temp table is not supported. use deleteAllTuples instead
// TODO: change meaningless bool return type to void (starting in class Table) and migrate callers.
virtual bool deleteTuple(TableTuple &tuple, bool);
```
Because we already deleted them in table.h and temptable does not need it

src/ee/storage/persistenttable.cpp
```diff
@@ -1126,7 +1121,7 @@ bool PersistentTable::checkNulls(TableTuple &tuple) const {
 /*
```
yzhang1991: I fell like we can directly replace the `pkeyIndex` pointer with `m_pkeyIndex` in the `PersistentTable::lookupTuple()` function.
*Line number: 1008-1052*

src/ee/storage/table.h
yzhang1991: I cannot add a comment outside a diff region, but can you correct a typo in line 99 of this file?
`Table lifespan can be managed bya reference count`, `bya` => `by a`

src/ee/storage/persistenttable.h
```diff
@@ -278,18 +281,31 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
...
+    int uniqueIndexCount() const {
+        return static_cast<int>(m_uniqueIndexes.size());
     }
 
+    // returned via shallow vector copy -- seems good enough.
+    const std::vector<TableIndex*>& allIndexes() const { return m_indexes; }
+
+    TableIndex *index(std::string name);
+
+    TableIndex *primaryKeyIndex() { return m_pkeyIndex; }
```
yzhang1991: I want to suggest adding a `const` keyword to both `index()` and `primaryKeyIndex()`:
```cpp
TableIndex *index(std::string name) const;
```
and
```cpp
TableIndex *primaryKeyIndex() const { return m_pkeyIndex; }
```
Make sure to change `persistenttable.cpp`, line 1726.

src/ee/storage/persistenttable.cpp
```diff
@@ -1759,4 +1703,110 @@ void PersistentTable::computeSmallestUniqueIndex() {
...
+    BOOST_FOREACH(TableIndex *i2, indexes) {
+        if (i2 == index) {
+            return true;
+        }
+    }
+    return false;
+}
+#endif
+
+TableIndex* PersistentTable::index(std::string name)
```
yzhang1991: I have suggested adding a `const` keyword here.
```cpp
TableIndex* PersistentTable::index(std::string name) const {
```
Put the brace on the same line of function header.

src/ee/storage/persistenttable.cpp
```diff
@@ -1759,4 +1703,110 @@ void PersistentTable::computeSmallestUniqueIndex() {
...
 
+std::vector<uint64_t> PersistentTable::getBlockAddresses() const {
+    std::vector<uint64_t> blockAddresses;
+    blockAddresses.reserve(m_data.size());
+    for(TBMap::const_iterator i = m_data.begin(); i != m_data.end(); ++i) {
+            blockAddresses.push_back((uint64_t)i->second->address());
+    }
+    return blockAddresses;
+}
+#ifdef DEBUG
```
yzhang1991: Add a blank line here please.

src/ee/storage/persistenttable.cpp
```diff
@@ -1759,4 +1703,110 @@ void PersistentTable::computeSmallestUniqueIndex() {
...
 }
 
+std::vector<uint64_t> PersistentTable::getBlockAddresses() const {
+    std::vector<uint64_t> blockAddresses;
+    blockAddresses.reserve(m_data.size());
+    for(TBMap::const_iterator i = m_data.begin(); i != m_data.end(); ++i) {
+            blockAddresses.push_back((uint64_t)i->second->address());
+    }
+    return blockAddresses;
+}
```
yzhang1991: Add a blank line here please.

src/ee/storage/persistenttable.cpp
```diff
@@ -1759,4 +1703,110 @@ void PersistentTable::computeSmallestUniqueIndex() {
...
+#ifdef DEBUG
+static bool isExistingTableIndex(std::vector<TableIndex*> &indexes, TableIndex* index)
+{
+    BOOST_FOREACH(TableIndex *i2, indexes) {
```
yzhang1991: What does `i2` mean here? Is there a better name?

src/ee/storage/persistenttable.h
```diff
@@ -255,8 +259,7 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
     // The initial use case is a live catalog update that changes table schema and migrates tuples
     // and/or adds a materialized view.
     // Constraint checks are bypassed and the change does not make use of "undo" support.
-    // TODO: change meaningless bool return type to void (starting in class Table) and migrate callers.
-    virtual bool deleteTuple(TableTuple &tuple, bool fallible=true);
+    void deleteTuple(TableTuple &tuple, bool fallible=true);
     // TODO: change meaningless bool return type to void (starting in class Table) and migrate callers.
     virtual bool insertTuple(TableTuple &tuple);
```
yzhang1991: Can we change the return type for `insertTuple()` now as well?
* `persistenttable.h`, line 264 (*here*)
* `persistenttable.cpp`, line 473
* `streamedtable.h`, line 74
* `streamedtable.cpp`, line 131
* `table.h`, line 130
* `temptable.h`, line 111
* `temptable.cpp`, line 71

src/ee/storage/persistenttable.h
```diff
@@ -512,8 +519,10 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
 
     std::pair<const TableIndex*, uint32_t> getUniqueIndexForDR();
 
-  private:
+protected:
+    std::vector<uint64_t> getBlockAddresses() const;
```
yzhang1991: I don't see where this function is used?

src/ee/storage/persistenttable.cpp
```diff
@@ -291,9 +295,9 @@ void PersistentTable::truncateTableForUndo(VoltDBEngine * engine, TableCatalogDe
         this->unsetPreTruncateTable();
     }
 
-    std::vector<MaterializedViewMetadata *> views = originalTable->views();
+    std::vector<MaterializedViewWriteTrigger*> views = originalTable->views();
```
yzhang1991: Stopped using `MatViewType` in this file?

yzhang1991:
End of review round.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1175,32 +1171,60 @@ void VoltDBEngine::rebuildTableCollections()
...
     BOOST_FOREACH (LabeledTCD cd, m_catalogDelegates) {
         TableCatalogDelegate *tcd = cd.second;
-        if (tcd) {
-            Table* localTable = tcd->getTable();
-            catalog::Table *catTable = m_database->tables().get(localTable->name());
-            m_tables[catTable->relativeIndex()] = localTable;
-            m_tablesByName[tcd->getTable()->name()] = localTable;
-            if (!tcd->exportEnabled() && !tcd->materialized()) {
-                m_tablesBySignatureHash[*reinterpret_cast<const int64_t*>(tcd->signatureHash())] = tcd->getPersistentTable();
+        assert(tcd);
```
wolffcm: So in a debug build, we'll fail an assert if the TCD is null, but then you follow up with a check.  Do you think that it's possible that the TCD might be null?

wolffcm:
I just had a quick look.  Looks good to me, and much improved.

src/ee/storage/persistenttable.h
paulmartel: This would lead to executor code changes that are minor but would likely cause conflicts with the upcoming full outer join merge, so I'd like to defer it for now.

PR #3604, opened 2016-05-20
Eng 10399 more info in conflict log

luyangco:


src/ee/common/tabletuple.h
```diff
@@ -386,6 +386,20 @@ class TableTuple {
         return std::string(retval, 0, retval.length() - 1);
     }
 
+    std::string toJsonArray(const std::vector<std::string>& columnNames) const {
+        int totalColumns = sizeInValues();
+        Json::Value array(Json::arrayValue);
+
+        array.resize(totalColumns);
+        for (int i = 0; i < totalColumns; i++) {
+            array[i] = columnNames[i] + ":" + getNValue(i).toString();
```
nshi: Is this still valid JSON, by creating an array of key:value pairs?

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -156,13 +162,22 @@ void setConflictOutcome(boost::shared_ptr<TempTable> metadataTable, bool acceptR
```
nshi: |`line 231`|
|---|
I think you can get this directly by calling UniqueId::timestampSinceUnixEpoch().

nshi:
|`src/ee/storage/BinaryLogSink.cpp:351`|
|---|
This should also be protected by the if.

src/ee/common/tabletuple.h
luyangco: I think this is legit, it is still an JSON array with colon inside the value of each element, JSON spec doesn't require to escape any colon in value section.

luyangco: I mean, this is still an array of strings, not an array of key:value pairs

nshi: I'm thinking that maybe a JSON dictionary makes more sense here, e.g. "{c1: val1, c2: val2}".

PR #3605, opened 2016-05-20
ENG-10383: Support JSON Hashinator config

JadenW:
JSON Format for the hashinator config
{"count": 16384,
 "map": {"token1": partID1,
                     "token2": partID2,
                     ...}
 }

Compressing using GZIP

use "@Statsistics TOPO 1" for retrieving new format

checkForTopologyChanges in ClientIntface still using old binary format

### DIFF
--- a/src/frontend/org/voltdb/ElasticHashinator.java
+++ b/src/frontend/org/voltdb/ElasticHashinator.java
@@ -35,7 +35,7 @@ import java.util.zip.GZIPOutputStream;
 
 import org.apache.cassandra_voltpatches.MurmurHash3;
 import org.json_voltpatches.JSONException;
-import org.json_voltpatches.JSONObject;
+import org.json_voltpatches.JSONStringer;
 import org.voltcore.utils.Bits;
 import org.voltcore.utils.Pair;
 import org.voltdb.utils.CompressionService;
@@ -218,11 +218,13 @@ public class ElasticHashinator extends TheHashinator {
      */
 
     private String toJSONString() {
-        JSONObject js = new JSONObject();
+        JSONStringer js = new JSONStringer();
         try {
+            js.object();
             for (Map.Entry<Integer, Integer> entry : m_tokensMap.get().entrySet()) {
-                js.put(entry.getKey().toString(),entry.getValue());
+                js.key(entry.getKey().toString()).value(entry.getValue());
             }
+            js.endObject();
         } catch (JSONException e) {
             throw new RuntimeException("Failed to serialize Hashinator Configuration to JSON.", e);
         }
@@ -250,25 +252,26 @@ public class ElasticHashinator extends TheHashinator {
 
     public static byte[] compressJSONString(String data) throws IOException {
         ByteArrayOutputStream bos = new ByteArrayOutputStream(data.length());
-        GZIPOutputStream gzip = new GZIPOutputStream(bos);
-        gzip.write(data.getBytes(StandardCharsets.UTF_8));
-        gzip.close();
-        byte[] compressed = bos.toByteArray();
-        bos.close();
-        return compressed;
+        try (GZIPOutputStream gzip = new GZIPOutputStream(bos)) {
+            gzip.write(data.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
+            throw e;
+        }
+        return bos.toByteArray();
     }
 
     public static String decompressJSONString(byte[] compressed) throws IOException {
-        ByteArrayInputStream bis = new ByteArrayInputStream(compressed);
-        GZIPInputStream gis = new GZIPInputStream(bis);
         ByteArrayOutputStream result = new ByteArrayOutputStream();
-        byte[] buffer = new byte[1024];
-        int length;
-        while ((length = gis.read(buffer)) != -1) {
-            result.write(buffer, 0, length);
+        try (ByteArrayInputStream bis = new ByteArrayInputStream(compressed);
+                GZIPInputStream gis = new GZIPInputStream(bis)) {
+            byte[] buffer = new byte[1024];
+            int length;
+            while ((length = gis.read(buffer)) != -1) {
+                result.write(buffer, 0, length);
+            }
+        } catch (IOException e) {
+            throw e;
         }
-        gis.close();
-        bis.close();
         return result.toString("UTF-8");
     }
 
--- a/src/frontend/org/voltdb/ElasticHashinator.java
+++ b/src/frontend/org/voltdb/ElasticHashinator.java
@@ -65,8 +65,8 @@ public class ElasticHashinator extends TheHashinator {
         Integer.parseInt(System.getProperty("ELASTIC_TOTAL_TOKENS", "16384"));
 
     // JSON KEYS FOR SERIALIZATION
-    public static final String JSON_TOKENCOUNT_KEY = "tokencount";
-    public static final String JSON_TOKENPARTITION_KEY = "tokenpartition";
+    public static final String JSON_TOKENCOUNT_KEY = "count";
+    public static final String JSON_TOKENPARTITION_KEY = "map";
 
     /**
      * Tokens on the ring. A value hashes to a token if the token is the first value <=
--- a/src/frontend/org/voltdb/ElasticHashinator.java
+++ b/src/frontend/org/voltdb/ElasticHashinator.java
@@ -17,7 +17,11 @@
 package org.voltdb;
 
 
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.lang.Thread.State;
 import java.nio.ByteBuffer;
 import java.util.HashSet;
@@ -27,11 +31,17 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
 
 import org.apache.cassandra_voltpatches.MurmurHash3;
+import org.json_voltpatches.JSONException;
+import org.json_voltpatches.JSONStringer;
 import org.voltcore.utils.Bits;
 import org.voltcore.utils.Pair;
 import org.voltdb.utils.CompressionService;
+
+import sun.misc.Cleaner;
 
 import com.google_voltpatches.common.base.Preconditions;
 import com.google_voltpatches.common.base.Supplier;
@@ -41,8 +51,6 @@ import com.google_voltpatches.common.collect.ImmutableSortedSet;
 import com.google_voltpatches.common.collect.Maps;
 import com.google_voltpatches.common.collect.SortedMapDifference;
 import com.google_voltpatches.common.collect.UnmodifiableIterator;
-
-import sun.misc.Cleaner;
 
 /**
  * A hashinator that uses Murmur3_x64_128 to hash values and a consistent hash ring
@@ -55,6 +63,10 @@ public class ElasticHashinator extends TheHashinator {
 
     public static int DEFAULT_TOTAL_TOKENS =
         Integer.parseInt(System.getProperty("ELASTIC_TOTAL_TOKENS", "16384"));
+
+    // JSON KEYS FOR SERIALIZATION
+    public static final String JSON_TOKENCOUNT_KEY = "tokencount";
+    public static final String JSON_TOKENPARTITION_KEY = "tokenpartition";
 
     /**
      * Tokens on the ring. A value hashes to a token if the token is the first value <=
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -56,6 +64,10 @@
     public static int DEFAULT_TOTAL_TOKENS =
         Integer.parseInt(System.getProperty("ELASTIC_TOTAL_TOKENS", "16384"));
 
+    // JSON KEYS FOR SERIALIZATION
+    public static final String JSON_TOKENCOUNT_KEY = "tokencount";
+    public static final String JSON_TOKENPARTITION_KEY = "tokenpartition";
+
```
zeemanhe: Short key names could save a few bytes.

src/frontend/org/voltdb/StatsAgent.java
```diff
@@ -312,10 +312,14 @@ private void collectTopoStats(PendingOpsRequest psr)
             VoltTable vt =
                     new VoltTable(
                             new VoltTable.ColumnInfo("HASHTYPE", VoltType.STRING),
-                            new VoltTable.ColumnInfo("HASHCONFIG", VoltType.VARBINARY));
+                            new VoltTable.ColumnInfo("HASHCONFIG", VoltType.VARBINARY),
+                            new VoltTable.ColumnInfo("HASHCONFIGJSON", VoltType.STRING),
```
akhanzode: Why keep both? Just use compressed.

src/frontend/org/voltdb/ElasticHashinator.java
JadenW: Is "count" and  "map" simply enough and not so confusing ?

src/frontend/org/voltdb/StatsAgent.java
JadenW: removed

src/frontend/org/voltdb/ElasticHashinator.java
zeemanhe: or "count", "tokens"?

tests/frontend/org/voltdb/regressionsuites/statistics/TestStatisticsSuite.java
```diff
@@ -362,9 +362,10 @@ public void testTopoStatistics() throws Exception {
         expectedSchema1[2] = new ColumnInfo("Leader", VoltType.STRING);
         VoltTable expectedTable1 = new VoltTable(expectedSchema1);
 
-        ColumnInfo[] expectedSchema2 = new ColumnInfo[2];
+        ColumnInfo[] expectedSchema2 = new ColumnInfo[3];
         expectedSchema2[0] = new ColumnInfo("HASHTYPE", VoltType.STRING);
         expectedSchema2[1] = new ColumnInfo("HASHCONFIG", VoltType.VARBINARY);
+        expectedSchema2[2] = new ColumnInfo("HASHCONFIGJSONCompressed", VoltType.VARBINARY);
```
akhanzode: This test will depend on what is the delta flag right? So the test should have 2 diff tests with diff flags.

src/frontend/org/voltdb/StatsAgent.java
```diff
@@ -315,7 +315,16 @@ private void collectTopoStats(PendingOpsRequest psr)
                             new VoltTable.ColumnInfo("HASHCONFIG", VoltType.VARBINARY));
             tables[1] = vt;
             HashinatorConfig hashConfig = TheHashinator.getCurrentConfig();
-            vt.addRow(hashConfig.type.toString(), hashConfig.configBytes);
+            // hacky way to support two format of hashconfig
+            // if interval == true (delta-flag == 1), sent compressed json
+            // otherwise sent original binary format
+            boolean interval = psr.getInterval();
+            if (!interval) {
```
akhanzode: Call this config type.

src/frontend/org/voltdb/OpsAgent.java
```diff
@@ -105,6 +106,14 @@ public PendingOpsRequest(
             this.clientData = clientData;
             this.request = request;
         }
+
+        public boolean getInterval() {
```
akhanzode: You can call this method getConfigType() and document its reading interval.

tests/frontend/org/voltdb/regressionsuites/statistics/TestStatisticsSuite.java
JadenW: Sorry, forget to change this back. 
The delta flag will not affect the result table schema. 
They will all return same one row table of two column.

tests/frontend/org/voltdb/regressionsuites/statistics/TestStatisticsSuite.java
```diff
@@ -392,6 +392,30 @@ public void testTopoStatistics() throws Exception {
...
+        // Make sure we can find the MPI, at least
+        found = false;
+        while (topo.advanceRow()) {
+            if ((int)topo.getLong("Partition") == MpInitiator.MP_INIT_PID) {
+                found = true;
+            }
+        }
+        assertTrue(found);
+        // and only one row in the second table
+        assertEquals(1, results[1].getRowCount());
```
akhanzode: Now that HASHCONFIG is available in JSON can you verify content assuming lets say 1 site?

akhanzode: Never mind you have it in other place.

nshi:
JSON map is self-describing, you don't need to put the count there.

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -94,6 +106,19 @@ public Long get() {
         }
     });
 
+    private final Supplier<String> m_configJSON = Suppliers.memoize(new Supplier<String>() {
```
nshi: Who uses the uncompressed version? I don't see any callers other than test. If nobody calls this, you don't need to cache it.

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +224,71 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+        JSONStringer js = new JSONStringer();
+        try {
+            js.object();
+            js.key(JSON_TOKENCOUNT_KEY).value(m_tokenCount);
+            js.key(JSON_TOKENPARTITION_KEY).object();
+            for (Map.Entry<Integer, Integer> entry : m_tokensMap.get().entrySet()) {
+                js.key(entry.getKey().toString()).value(entry.getValue());
+            }
+            js.endObject().endObject();
+        }catch (JSONException e) {
```
nshi: Nitpick: missing a space here before `catch`.

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +224,71 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+            throw new RuntimeException("Failed to serialized Hashinator Configuration to JSON.", e);
```
nshi: "Failed to serialize..."

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +224,71 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+    private byte[] toJSONStringCompressed() {
+        String str = m_configJSON.get();
+        if (str == null || str.length() == 0) {
+            return new byte[0] ;
+        }
+        byte[] outStr;
+        try {
+            outStr = compressJSONString(str);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to serialized Hashinator Configuration to Compressed JSON .", e);
```
nshi: "Failed to serialize..."

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +224,71 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+        }
+        return outStr;
+    }
+
+    public static byte[] compressJSONString(String data) throws IOException {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream(data.length());
+        GZIPOutputStream gzip = new GZIPOutputStream(bos);
+        gzip.write(data.getBytes());
```
nshi: Use UTF-8.

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +224,71 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+        return compressed;
+    }
+
+    public static String decompressJSONString(byte[] compressed) throws IOException {
+        ByteArrayInputStream bis = new ByteArrayInputStream(compressed);
+        GZIPInputStream gis = new GZIPInputStream(bis);
+        BufferedReader br = new BufferedReader(new InputStreamReader(gis, "UTF-8"));
+        StringBuilder sb = new StringBuilder();
+        String line;
+        while((line = br.readLine()) != null) {
```
nshi: Might be faster to use a StringWriter which doesn't parse the content.

src/frontend/org/voltdb/OpsAgent.java
```diff
@@ -105,6 +106,17 @@ public PendingOpsRequest(
             this.clientData = clientData;
             this.request = request;
         }
+
+        // hacky way to support two format of hashconfig using interval value
+        // return true if interval == true (delta-flag == 1), indicate sent compressed json
+        // otherwise return false, indicate sent original binary format
+        public boolean getConfigType() {
```
nshi: I suggest renaming this method. `getConfigType()` sounds like it should return a type.

nshi: Since this is only used by the `topo` selector, why not let the topo collector make the decision. This code is certainly not applicable to other selectors.

JadenW:
Thanks. Then can furthure simply the config to just a "map".

src/frontend/org/voltdb/ElasticHashinator.java
JadenW: compressed version use it.

nshi: But if the compressed config is cached, it's only gonna compress once. There is no point in caching for a one time use.

src/frontend/org/voltdb/ElasticHashinator.java
JadenW: Seem the fastest way to convert an inputstream to a string is using ByteArrayOutputStream
http://stackoverflow.com/questions/309424/read-convert-an-inputstream-to-a-string

src/frontend/org/voltdb/ElasticHashinator.java
JadenW: I see. Fixed

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +213,66 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
     }
 
     /**
+     * Serializes the configuration into JSON, also updates the currently cached m_configJSON.
+     * @return The JSONString of the current configuration.
+     */
+
+    private String toJSONString() {
+        JSONObject js = new JSONObject();
```
vtkstef: if you use JSONStringer it will keep the order. JsonObject is backed by a Hashtable which does not preserve insert order upon entry set iterations

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +213,66 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+    }
+
+    /**
+     * Serializes the configuration into JSON, also updates the currently cached m_configJSON.
+     * @return The JSONString of the current configuration.
+     */
+
+    private byte[] toJSONStringCompressed() {
+        String str = toJSONString();
+        if (str == null || str.length() == 0) {
```
vtkstef: str == null || str.trim().isEmpty()

src/frontend/org/voltdb/ElasticHashinator.java
```diff
@@ -199,6 +213,66 @@ private ElasticHashinator(SortedMap<Integer, Integer> tokens) {
...
+        try {
+            outStr = compressJSONString(str);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to serialize Hashinator Configuration to Compressed JSON .", e);
+        }
+        return outStr;
+    }
+
+    public static byte[] compressJSONString(String data) throws IOException {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream(data.length());
```
vtkstef:
```java
public static byte[] compressJSONString(String data) throws IOException {
    ByteArrayOutputStream bos = new ByteArrayOutputStream(data.length());
    try (GZIPOutputStream gzip = new GZIPOutputStream(bos)) {
        gzip.write(data.getBytes(StandardCharsets.UTF_8));
    } catch (IOException e) { throw e; }
    return bos.toByteArray();
}
```

JadenW: Try with resources?
Thanks. this is much cleaner

src/frontend/org/voltdb/StatsAgent.java
```diff
@@ -220,7 +220,11 @@ protected void collectStatsImpl(Connection c, long clientHandle, OpsSelector sel
                     clientHandle,
                     System.currentTimeMillis(),
                     obj);
-            collectTopoStats(psr);
+            // hacky way to support two format of hashconfig using interval value
```
vtkstef: I prefer we either support a new TOPOJSON stats selector or PROMOTE topo as system procedure which can be easily coded into InvocationDispatcher (we may need to adjust sqlcmd filters)

PR #3606, opened 2016-05-21
ENG-10404 DO NOT MERGE add initialize start action and voltdb init command

vtkstef:
this is the integration branch for the new CLI. This pull request is intended just for review


vtkstef:
this is an integration branch

PR #3607, opened 2016-05-23
ENG-10436 snapshot restore constraint violations

zeemanhe:
1) prevent db from being crashed by constraint violations. 
(2) log constraint violations with rate limit logger
(3) Restore process continues to end even constraint violations are noted.


src/frontend/org/voltdb/sysprocs/saverestore/PartitionedTableSaveFileState.java
```diff
@@ -108,18 +108,18 @@ int getTotalPartitions()
 
     @Override
     public SynthesizedPlanFragment[]
-    generateRestorePlan(Table catalogTable, SiteTracker st)
+    generateRestorePlan(Table catalogTable, SiteTracker st, boolean isRestore)
```
vtkstef: when can you generate a restore plan when it is not a `isRestore`?

src/frontend/org/voltdb/sysprocs/saverestore/SnapshotUtil.java
```diff
@@ -1534,6 +1534,17 @@ public static HashinatorSnapshotData retrieveHashinatorConfig(
...
+            try{
+                JSONObject jsObj = new JSONObject((String)params[0]);
+                String path = jsObj.getString(JSON_PATH);
+                String dupPath = jsObj.getString(JSON_DUPLICATES_PATH);
+                if(path != null && dupPath == null){
+                    jsObj.put(JSON_DUPLICATES_PATH, path)  ;
+                }
+                task.setParams( jsObj.toString() );
+            } catch (JSONException e){
+                //move on
```
vtkstef: Uh? why move on? if you want the equivalent of if json.contains then get get then use jsObj.optString with a default

src/frontend/org/voltdb/sysprocs/saverestore/PartitionedTableSaveFileState.java
zeemanhe: Thought recover goes through this. I may be wrong. Let me double check.

vtkstef: a recover always restores a truncation snapshot

src/frontend/org/voltdb/sysprocs/saverestore/SnapshotUtil.java
zeemanhe: The input string should be always valid.  If this happens, then we got problems. Theoretically this exception should not happen.

vtkstef: so we ignore it when it does happen?

vtkstef: use jsObj.optString() instead of getString() because getString throws when it does not find it

src/frontend/org/voltdb/sysprocs/saverestore/PartitionedTableSaveFileState.java
zeemanhe: When  db is brought up with recover on my dev environment, it goes through generateRestorePlan(). Does the recover work as restore?

src/frontend/org/voltdb/sysprocs/saverestore/SnapshotUtil.java
```diff
@@ -1566,6 +1577,7 @@ public static ClientResponseImpl transformRestoreParamsToJSON(StoredProcedureInv
             try {
                 jsObj.put(SnapshotUtil.JSON_PATH, params[0]);
                 jsObj.put(SnapshotUtil.JSON_NONCE, params[1]);
+                jsObj.put(SnapshotUtil.JSON_DUPLICATES_PATH, params[0]);
```
vtkstef: are we doing this also for truncation snapshot restores? For those we cannot enable duplicates

zeemanhe: have commandlog enabled. run recover, step through CommandLogReinitiatorImpl & SnapshotRestore.  It does not hit the above code.

vtkstef:
Reviewed

PR #3610, opened 2016-05-24
ENG-10429 fix

yzhang1991:
Change in Expression.java, manually convert simple column back to column.
Added test case for ENG-10429.


yzhang1991:
**Fix note:**
This is the minimum fix to get this case working in both VoltDB and HSQL backend. 
In Expression.java, line 1621: convert the `SIMPLE_COLUMN` into a `COLUMN`
The two lines: 

    opType = OpTypes.COLUMN; 
    exprOp = OpTypes.COLUMN; 

were in the `if (displayCols == null)` branch. 
The ticket was to solve the exception caused by executing
`select sum(a)/b as val from t group by b;`
The error message was `Unsupported expression node 'simplecolumn'`
This was because VoltDB tried to find simplecolumn replacement in the `displayCols`,
but `b` was not in it. 
For those columns, simply revert their `OpTypes` flags from `SIMPLE_COLUMN` back to `COLUMN`. 
The detailed fields in the expression itself is enough for VoltDB to reconstruct the columnref node.

BillWhite:
So, what is a node with a SIMPLE_COLUMN opType?  The expression, as I recall is something like "SUM(T.C)/T2.C2".  Which is the SIMPLE_COLUMN?   Is this a kind of common subexpression elimination?  We put the value of the aggregate in a column someplace, just like we would put the common subexpression in a temporary or a register, and we replace it with a reference to the temporary column.  This works because the index of the SIMPLE_COLUMN is the same as the index of the temporary column in which the value will be computed.

Does this seem right?

This looks ok to me.

PR #3611, opened 2016-05-24
Eng 10330 upsert sqlcov

johnTcrawford:
Adds SQLCoverage tests of UPSERT statements


src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -76,6 +83,9 @@ public NonVoltDBBackend(String databaseType, String driverClassName,
             dbconn = DriverManager.getConnection(connectionURL, username, password);
             dbconn.setAutoCommit(true);
             dbconn.setTransactionIsolation(Connection.TRANSACTION_SERIALIZABLE);
+            DatabaseMetaData meta = dbconn.getMetaData();
+            System.out.println("Using database: " + meta.getDatabaseProductName()+" "+meta.getDatabaseProductVersion());
```
wolffcm: Did you mean to leave these println statements in here, or are they just for debugging?

tests/scripts/SQLCoverageReport.py
```diff
@@ -300,7 +304,7 @@ def generate_html_reports(suite, seed, statements_path, cmpdb_path, jni_path,
     cmpdb_file = open(cmpdb_path, "rb")
     jni_file = open(jni_path, "rb")
     if modified_sql_path:
-        modified_sql_file = open(modified_sql_path, "rb")
+        modified_sql_file = codecs.open(modified_sql_path, encoding='utf-8')
```
wolffcm: Curious why you needed to make this change?

tests/scripts/examples/sql_coverage/geo-config.py
```diff
@@ -49,4 +51,17 @@
                              "ddl": "geo-DDL.sql",
                              "template": "geo-advanced-polygon.sql",
                              "normalizer": "geo-normalizer.py"},
+    # UPSERT tests
```
wolffcm: I guess this will get moved to its own config at some point?  Doesn't seem to have much to do with geo.

wolffcm:
Looks good to me.  This code is getting pretty complicated.  I think I say that every time I review changes to sqlcoverage.  Do you think at some point refactoring will be necessary to continue adding new tests?

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: I meant to leave them in: I think it's useful to know (e.g., when running on some Jenkins machine) exactly which database and JDBC version are being used. (Good question, though: I do tend to use a lot of debug print, and some other time it could easily get left behind.)

tests/scripts/examples/sql_coverage/geo-config.py
johnTcrawford: Either that, or this config will get renamed to something like postgres-only or postgis-only, for stuff that cannot be run against HSQL, only PostgreSQL (with or without PostGIS).

tests/scripts/SQLCoverageReport.py
johnTcrawford: Because it allows the "modified_sql[cgi.escape(orig_sql).encode..." stuff below to work with non-ascii characters (i.e., with non-ascii VARCHAR values); and the combination allows the printing of the modified (for PostgreSQL) version of the SQL statement to be printed in the browser. This fixed an existing problem where some modified SQL statements did not show up in the browser, which was not directly related to the UPSERT task at hand, but it made it much harder to diagnose the problems.

johnTcrawford:
Thanks. I think the refactoring I did a little while back, at your suggestion, of the transformQuery method, adding the QueryTransformer (builder) class went a long way to making this maintainable. The current additions are very complex because PostgreSQL's equivalent of UPSERT is so complex and so different from VoltDB's UPSERT, but I think they are limited in scope, and should not limit adding new tests, with the possible exception of more UPSERT tests.

wolffcm:
Ah yes, I remember that you've already done some refactoring here.  The QueryTransformer stuff is nice. I guess I'm wondering if at some point regular expressions will cause trouble (as they have in the past with other parts of the product).  So far, they seem to be holding up.

johnTcrawford:
Of course I cannot say that we will never have problems, but we do have the advantage here that the regular expressions only have to capture the specific query templates written in SQLCoverage. Although there are probably hundreds of those, they are a finite set, as opposed to needing to parse every conceivable SQL statement that a user might come up with.

PR #3614, opened 2016-05-26
Two small patches related to partition detection

nshi:
* Don't trigger PPD if feature is disabled.
* Don't trigger if node hasn't joined the mesh.

### DIFF
--- a/src/frontend/org/voltcore/messaging/HostMessenger.java
+++ b/src/frontend/org/voltcore/messaging/HostMessenger.java
@@ -362,23 +362,21 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
         boolean partitionDetected = makePPDDecision(previousHosts, currentHosts);
 
         if (partitionDetected) {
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -362,38 +362,38 @@ private void doPartitionDetectionActivities(Set<Integer> failedHostIds)
...
             synchronized(HostMessenger.this) {
 
                 // Decide if the failures given could put the cluster in a split-brain
                 // Then decide if we should shut down to ensure that at a MAXIMUM, only
                 // one viable cluster is running.
                 // This feature is called "Partition Detection" in the docs.
-                doPartitionDetectionActivities(failedHostIds);
+                if (m_partitionDetectionEnabled.get()) {
+                    doPartitionDetectionActivities(failedKnownHostIds);
+                }
```
jhugg: This is actually quite intentional. If someone turned off PD in deployment and hit a partition, we want to log!
If PD got turned off for shutdown, that's actually a new state.

PR #3619, opened 2016-05-26
ENG-10421: allow multiple hosts trying to rejoin at the same time

nshi:
On startup, each node sends a request to rejoin the mesh to the live node. The live node will check if the request can be accepted at the moment. For example, we only allow one node to rejoin at a time. If more than one node tries to rejoin, the first rejoning node will atomically check and set the ZK blocker. Subsequent node will fail to do so and be rejected to join the mesh. They will retry after a while until they can succeed.

### DIFF
--- a/src/frontend/org/voltcore/messaging/HostMessenger.java
+++ b/src/frontend/org/voltcore/messaging/HostMessenger.java
@@ -59,7 +59,6 @@ import org.voltcore.utils.ShutdownHooks;
 import org.voltcore.zk.CoreZK;
 import org.voltcore.zk.ZKUtil;
 import org.voltdb.VoltDB;
-import org.voltdb.utils.MiscUtils;
 
 import com.google_voltpatches.common.base.Charsets;
 import com.google_voltpatches.common.base.Preconditions;
@@ -95,6 +94,17 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
          * @return true if the new host can join the mesh, false otherwise.
          */
         boolean shouldAccept(int hostId, String request, StringBuilder errMsg);
+    }
+
+    /**
+     * Callback for watching for host failures.
+     */
+    public interface HostWatcher {
+        /**
+         * Called when host failures are detected.
+         * @param failedHosts    List of failed hosts, including hosts currently unknown to this host.
+         */
+        void hostsFailed(Set<Integer> failedHosts);
     }
 
     /**
@@ -136,10 +146,6 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
             this(null, 3021);
             zkInterface = "127.0.0.1:" + ports.next();
             internalPort = ports.next();
-        }
-
-        public int getZKPort() {
-            return MiscUtils.getPortFromHostnameColonPort(zkInterface, VoltDB.DEFAULT_ZK_PORT);
         }
 
         private void initNetworkThreads() {
@@ -238,6 +244,7 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
     private boolean m_partitionDetected = false;
 
     private final MembershipAcceptor m_membershipAcceptor;
+    private final HostWatcher m_hostWatcher;
 
     private final Object m_mapLock = new Object();
 
@@ -268,17 +275,18 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
     }
 
     /**
-     *
      * @param network
      * @param coordinatorIp
      * @param expectedHosts
      * @param catalogCRC
      * @param hostLog
      * @param membershipAcceptor
+     * @param m_hostWatcher
      */
-    public HostMessenger(Config config, MembershipAcceptor membershipAcceptor) {
+    public HostMessenger(Config config, MembershipAcceptor membershipAcceptor, HostWatcher hostWatcher) {
         m_config = config;
         m_membershipAcceptor = membershipAcceptor;
+        m_hostWatcher = hostWatcher;
         m_network = new VoltNetworkPool(m_config.networkThreads, 0, m_config.coreBindIds, "Server");
         m_joiner = new SocketJoiner(
                 m_config.coordinatorIp,
@@ -421,6 +429,10 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
                         // reportForeignHostFailed should print on the console once
                         m_networkLog.info(String.format("Host %d failed (DisconnectFailedHostsCallback)", hostId));
                     }
+                }
+
+                if (m_hostWatcher != null) {
+                    m_hostWatcher.hostsFailed(failedHostIds);
                 }
             }
         }
--- a/src/frontend/org/voltcore/messaging/HostMessenger.java
+++ b/src/frontend/org/voltcore/messaging/HostMessenger.java
@@ -486,7 +486,12 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
     /**
      * Start the host messenger and connect to the leader, or become the leader
      * if necessary.
-     * @param request    The requested action to send to other nodes when joining the mesh.
+     *
+     * @param request The requested action to send to other nodes when joining
+     * the mesh. This is opaque to the HostMessenger, it can be any
+     * string. HostMessenger will encode this in the request to join mesh to the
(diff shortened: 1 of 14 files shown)


src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -554,6 +575,12 @@ private void connectToPrimary() {
             // read the json response sent by HostMessenger with HostID
             JSONObject jsonObj = readJSONObjFromWire(socket, remoteAddress);
 
+            // check if the membership request is accepted
+            if (!jsonObj.getBoolean("accepted")) {
```
jhugg: What happens if accepted is not a thing? i.e. if you try to connect to an old version? Does it give a sane error?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -503,6 +503,12 @@ public void initialize(VoltDB.Configuration config) {
 
             buildClusterMesh(isRejoin || m_joining);
 
+            // Create rejoin guard immediately after joining the mesh to prevent simultaneous rejoins
+            if (isRejoin) {
+                final String node = VoltZK.rejoinNodesBlockerHost+m_messenger.getHostId();
+                VoltZK.createRejoinNodeIndicator(m_messenger.getZK(),node);
```
jhugg: Why can't this code be in the acceptor?

jhugg:
Does the pro branch have a test?

nshi:
I'm still working on tests, they will be in the community repo, mostly just HostMessenger tests with ZK.

src/frontend/org/voltcore/messaging/SocketJoiner.java
nshi: I can make this optional and always accept if it's not there.

src/frontend/org/voltdb/RealVoltDB.java
nshi: The acceptor runs before the foreign host is set up, I can try to do it there see if the new node still gets the correct ZK info. If it works, that makes it atomic and closes the window.

jhugg: Yeah, give it a shot. Crossing fingers.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -402,6 +425,10 @@ public void disconnect(Set<Integer> failedHostIds) {
                         m_networkLog.info(String.format("Host %d failed (DisconnectFailedHostsCallback)", hostId));
                     }
                 }
+
+                if (m_hostWatcher != null) {
```
jhugg: Could probably use a comment here

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -466,7 +494,7 @@ public void start() throws Exception {
          * If start returns true then this node is the leader, it bound to the coordinator address
          * It needs to bootstrap its agreement site so that other nodes can join
          */
-        if(m_joiner.start(zkInitBarrier)) {
+        if(m_joiner.start(zkInitBarrier, request)) {
```
jhugg: Can you add a comment explaining the params.

src/frontend/org/voltcore/messaging/HostMessenger.java
nshi: RealVoltDB always adds a callback so that it's not null. There are a lot of tests that create HostMessenger, so I put a guard here in case they are not interested in the callback.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -651,7 +674,7 @@ private void removeForeignHost(int hostId) {
      * is done via ZK
      */
     @Override
-    public void requestJoin(SocketChannel socket, InetSocketAddress listeningAddress) throws Exception {
+    public void requestJoin(SocketChannel socket, InetSocketAddress listeningAddress, String request) throws Exception {
```
jhugg: Can you add a comment here explaining what "request" could be?

nshi:
I added unit tests and resolved merge conflicts. Let me know if there's further comment. I'll squash them into a single commit if all things look good.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1038,6 +1038,49 @@ public void initialize(VoltDB.Configuration config) {
...
+    public boolean shouldAccept(int hostId, String request, StringBuilder errMsg)
+    {
+        Preconditions.checkNotNull(m_messenger);
+
+        StartAction action = null;
+        try {
+            action = StartAction.valueOf(request);
+        } catch (IllegalArgumentException e) {}
+
+        if (action != null && action.doesRejoin()) {
```
jhugg: Maybe a comment here saying this whole shouldAccept thing is currently used to protect against rejoins when we shouldn't (multiple simultaneous, lack of consensus, etc...). We return true in non-rejoin cases for now.

src/frontend/org/voltdb/VoltZK.java
```diff
@@ -286,33 +286,62 @@ public static boolean removeCatalogUpdateBlocker(ZooKeeper zk, String node, Volt
...
+            Stat stat = new Stat();
+            final int rejoiningHost = ByteBuffer.wrap(zk.getData(rejoinNodeBlocker, false, stat)).getInt();
+            if (hostId == rejoiningHost) {
+                zk.delete(rejoinNodeBlocker, stat.getVersion());
+                return true;
+            }
         } catch (KeeperException e) {
-            if (e.code() != KeeperException.Code.NONODE) {
+            if (e.code() != KeeperException.Code.NONODE &&
+                e.code() != KeeperException.Code.BADVERSION) {
```
jhugg: Why is BADVERSION ok? I'm new to this.

tests/frontend/org/voltcore/messaging/TestHostMessenger.java
```diff
@@ -222,4 +226,36 @@ public void testPartitionDetection5050KillNonBlessed() throws Exception
         // this should not trip partition detection
         assertFalse(HostMessenger.makePPDDecision(previous, current));
     }
+
+    @Test
+    public void testMembershipAcceptor() throws Exception
```
jhugg: Have we tried it with 3 host messengers?

src/frontend/org/voltdb/RealVoltDB.java
nshi: Sure.

src/frontend/org/voltdb/VoltZK.java
nshi: `BADVERSION` means some other node has deleted this ZK node and created a new rejoin blocker. In that case, it should be ok to move forward because the purpose of this method is to remove the given hostId's blocker, which is gone now.

tests/frontend/org/voltcore/messaging/TestHostMessenger.java
nshi: I tried it manually. I'll add a host messenger to this test.

nshi:
All of John's comments are addressed now.

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -572,6 +595,12 @@ private void connectToPrimary() {
             // read the json response sent by HostMessenger with HostID
             JSONObject jsonObj = readJSONObjFromWire(socket, remoteAddress);
 
+            // check if the membership request is accepted
+            if (jsonObj.has("accepted") && !jsonObj.getBoolean("accepted")) {
```
vtkstef: optBoolean("accepted",true)

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1038,6 +1038,53 @@ public void initialize(VoltDB.Configuration config) {
...
+        }
+
+        return true;
+    }
+
+    @Override
+    public void hostsFailed(Set<Integer> failedHosts)
+    {
+        scheduleWork(new Runnable() {
+            @Override
```
vtkstef: getSES(true).submit( new Runnable()

akhanzode:
Will this impact add with ksafe where we expect 2 or more joins right?

nshi:
@akhanzode Not sure I understand your question. If it's about elastic join, no, it won't have any impact on that.

PR #3621, opened 2016-05-27
Tuple block changes for empty persistent table

hhakimi53:
- Updated persistent table logic to allocate tuple storage block to table on table creation. Don't release the only remaining tuple storage block back to system during truncate or delete operation. This will prevent persistent table to request fresh tuple storage block on next row insertion into empty table.
- Updated snapshot save logic so that when system is performing snapshot save, it takes into account the persistent table can contain empty tuple storage


hhakimi53:
Changes on this branch have been reviewed on ENG-8872-block-changes. Merging changes to master.

PR #3625, opened 2016-05-31
ENG-10198: Generate DR sequence number for DR table changes and DR events all the time

slowsoul:
**Problem:**
1. Previously, DR sequence number is only generated for DR table changes when "listen" is set to true in the deployment file. This causes problem in recovery if the command log records corresponding to these unnumbered write transactions are not truncated before the cluster goes down. In this case, they will be newly numbered with the same DR sequence number as transactions after them, which may have already been received and recorded in the IdempotencyFilter in consumer cluster. Thus, when the command log replayed transactions hit the consumer cluster, the cluster will find two inconsistent transactions with the same DR sequence number and throw DR exception.
2. Because of the change, we cannot skip a write transaction immediately if "listen" is set to false since we need to increment the DR sequence number and update some states. However, it is the desired behavior if the DR table changes originated from remote cluster so that we don't do a circular replication.

**Solution:**
1. Move DR sequence number states from TupleStreamBase to AbstractDRTupleStream.
2. Get rid of calls to commit() in DRTupleStream related code, move transaction state update code to transactionChecks() and endTransaction(). Same for CompatibleDRTupleStream.
3. Adjust generateDREvent() to fit into the new design, also fix a bug of not updating the last committed uniqueIds for DR events.
4. AbstractDRTupleStream now overrides and has its own implementation of periodicFlush(), its flow is similar to commit() but greatly simplified. It basically pushes newly committed data to top end periodically.
5. A separate flag "m_guarded" is added to specifically guard the circular replication, while the existing "m_enabled" indicates if "listen" is set to true or not.

### DIFF
--- a/src/ee/storage/AbstractDRTupleStream.cpp
+++ b/src/ee/storage/AbstractDRTupleStream.cpp
@@ -24,16 +24,19 @@ using namespace voltdb;
 AbstractDRTupleStream::AbstractDRTupleStream(int partitionId, int defaultBufferSize)
         : TupleStreamBase(defaultBufferSize, MAGIC_DR_TRANSACTION_PADDING),
           m_enabled(true),
+          m_guarded(false),
+          m_openSequenceNumber(-1),
+          m_committedSequenceNumber(-1),
           m_partitionId(partitionId),
           m_secondaryCapacity(SECONDARY_BUFFER_SIZE),
           m_rowTarget(-1),
           m_opened(false),
-          m_txnRowCount(0) {
-
-}
+          m_txnRowCount(0)
+{}
 
 // for test purpose
-void AbstractDRTupleStream::setSecondaryCapacity(size_t capacity) {
+void AbstractDRTupleStream::setSecondaryCapacity(size_t capacity)
+{
     assert (capacity > 0);
     if (m_uso != 0 || m_openSpHandle != 0 ||
         m_openTransactionUso != 0 || m_committedSpHandle != 0)
@@ -44,7 +47,8 @@ void AbstractDRTupleStream::setSecondaryCapacity(size_t capacity) {
     m_secondaryCapacity = capacity;
 }
 
-void AbstractDRTupleStream::pushExportBuffer(StreamBlock *block, bool sync, bool endOfStream) {
+void AbstractDRTupleStream::pushExportBuffer(StreamBlock *block, bool sync, bool endOfStream)
+{
     if (sync) return;
     int64_t rowTarget = ExecutorContext::getExecutorContext()->getTopend()->pushDRBuffer(m_partitionId, block);
     if (rowTarget >= 0) {
@@ -54,8 +58,13 @@ void AbstractDRTupleStream::pushExportBuffer(StreamBlock *block, bool sync, bool
 
 // Set m_opened = false first otherwise checkOpenTransaction() may
 // consider the transaction being rolled back as open.
-void AbstractDRTupleStream::rollbackTo(size_t mark, size_t drRowCost) {
+void AbstractDRTupleStream::rollbackTo(size_t mark, size_t drRowCost)
+{
     if (mark == INVALID_DR_MARK) {
+        m_openSpHandle = m_committedSpHandle;
+        m_openUniqueId = m_committedUniqueId;
+        m_openSequenceNumber = m_committedSequenceNumber;
+        m_opened = false;
         return;
     }
     if (drRowCost <= m_txnRowCount) {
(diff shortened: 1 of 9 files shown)


src/ee/storage/DRTupleStream.cpp
```diff
@@ -321,9 +330,17 @@ void DRTupleStream::transactionChecks(int64_t lastCommittedSpHandle, int64_t spH
...
+        if (m_enabled) {
+            beginTransaction(m_openSequenceNumber, spHandle, uniqueId);
+        }
+        else {
+            m_openSpHandle = spHandle;
+            m_openUniqueId = uniqueId;
+            m_opened = true;
+        }
     }
     assert(m_opened);
```
manjujames: This assert will fail if !m_enabled?

manjujames: Never mind, saw the setting to true line.

src/ee/storage/CompatibleDRTupleStream.cpp
```diff
@@ -232,9 +238,17 @@ void CompatibleDRTupleStream::transactionChecks(int64_t lastCommittedSpHandle, i
                 );
     }
 
-    commit(lastCommittedSpHandle, spHandle, uniqueId, false, false);
     if (!m_opened) {
```
manjujames: transactionChecks and beginTransaction implementation seems really similar in DRTupleStream and CompatibleDRTupleStream. Is it possible to put the common code into AbstractDRTupleStream?

slowsoul: Yes, that's certainly possible, transactionChecks seems to be exactly the same, beginTransaction and endTransaction only differs in the binary log format.

luyangco: I agree, to make `DRTupleStream` and `CompatibleDRTupleStream` tight, we can put those common functions in `AbstractDRTupleStream`. The downside is once we need to change the implementation in `AbstractDRTupleStream`, make sure don't break compatible stream.

luyangco: `checkOpenTransaction` can be moved into abstract class too.

src/ee/storage/DRTupleStream.cpp
```diff
@@ -532,16 +580,51 @@ bool DRTupleStream::checkOpenTransaction(StreamBlock* sb, size_t minLength, size
 
 void DRTupleStream::generateDREvent(DREventType type, int64_t lastCommittedSpHandle, int64_t spHandle,
         int64_t uniqueId, ByteArray payloads) {
+    assert(!m_opened);
+
     switch (type) {
     case CATALOG_UPDATE: {
+        if (!m_enabled) {
+            if (UniqueId::isMpUniqueId(uniqueId)) {
```
luyangco: This block appears multiple times in `DRTupleStream`, wrap into a function?

luyangco: At line `691`, `getTestDRBuffer()` still uses `commit()`, should be replaced by `endTranscation()`. Same applies to `CompatibleDRTupleStream`

src/ee/storage/AbstractDRTupleStream.cpp
```diff
@@ -67,13 +74,66 @@ void AbstractDRTupleStream::rollbackTo(size_t mark, size_t drRowCost) {
     }
     if (mark == m_committedUso) {
         assert(m_txnRowCount == 0);
+        m_openSequenceNumber = m_committedSequenceNumber;
```
luyangco: Why `m_openSpHandle ` and `m_openUniqueId` not reset here?

src/ee/storage/AbstractDRTupleStream.cpp
```diff
@@ -67,13 +74,66 @@ void AbstractDRTupleStream::rollbackTo(size_t mark, size_t drRowCost) {
...
+{
+    // negative timeInMillis instructs a mandatory flush
+    if (timeInMillis < 0 || (m_flushInterval > 0 && timeInMillis - m_lastFlush > m_flushInterval)) {
+        int64_t currentSpHandle = std::max(m_openSpHandle, lastCommittedSpHandle);
+        if (timeInMillis > 0) {
+            m_lastFlush = timeInMillis;
+        }
+
+        if (currentSpHandle < m_openSpHandle)
+        {
```
luyangco: Keep code style consistent here, put curly bracket in the same line.

luyangco:
Review complete.

src/ee/storage/AbstractDRTupleStream.cpp
slowsoul: They are reset in the parent class implementation: TupleStreamBase::rollbackTo()

src/ee/storage/AbstractDRTupleStream.cpp
slowsoul: Fixed for DRTupleStream related class files

src/ee/storage/DRTupleStream.cpp
slowsoul: Addressed. refactored into openTransactionCommon() and commitTransactionCommon() in AbstractDRTupleStream

src/ee/storage/AbstractDRTupleStream.cpp
slowsoul: Made the code style consistent in DRTupleStream related classes.

luyangco:
Looks good to me. I'll squash merge the patch once it passes Jenkins test.

PR #3626, opened 2016-05-31
resolveColumnIndex Experiment 1

paulmartel:
Removed code repeated for no apparent reason from 40 lines prior. Jenkins passes fine without it.


yzhang1991:
Looks good to me.
The repeated code blocks seems to be independent of the code block in between them.

End of the review round.

BillWhite:
Looks good to me.

End of review round.

PR #3627, opened 2016-05-31
Remove underused expression tree search code.

paulmartel:
Eliminate code by standardizing on one way to search for subexpressions -- by java class rather than by ExpressionType tag.
These two search algorithms are theoretically not exactly equivalent in all cases, but they ARE equivalent in the ONE and ONLY oddball case we were using ExpressionType, namely for ExpressionType.VALUE_TUPLE. 

(this instanceof TupleValueExpression) === (m_expressionType == ExpressionType.VALUE_TUPLE)

### DIFF
--- a/src/frontend/org/voltdb/expressions/AbstractExpression.java
+++ b/src/frontend/org/voltdb/expressions/AbstractExpression.java
@@ -841,60 +841,7 @@ public abstract class AbstractExpression implements JSONString, Cloneable {
     }
 
     public ArrayList<AbstractExpression> findBaseTVEs() {
-        return findAllSubexpressionsOfType(ExpressionType.VALUE_TUPLE);
-    }
-
-    /**
-     * @param type expression type to search for
-     * @return a list of contained expressions that are of the desired type
-     */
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -841,60 +841,7 @@ public AbstractExpression replaceAVG () {
...
-                if (argument.hasAnySubexpressionOfType(type)) {
-                    return true;
-                }
-            }
-        }
-        return false;
+        return findAllSubexpressionsOfClass(TupleValueExpression.class);
     }
 
     /**
```
yzhang1991: Could you change the comment here?

    @param type expression type to search for

should be changed to an explanation for `aeClass`

yzhang1991:
The change looks good to me, except for a change in the comment to explain the function parameters.
End of the review round.

BillWhite:
I always wondered why we had two such functions.  This is a good change.

src/frontend/org/voltdb/expressions/AbstractExpression.java
paulmartel: Will change comment in upcoming round of commits.

PR #3629, opened 2016-05-31
NonVoltDBBackend.java: tweaked NonVoltDBBackend.transformQuery, adding a

johnTcrawford:
lastGroup variable that is never null, as opposed to the 'group' variable which sometimes is - thus avoiding the NPE we were seeing in certain cases.


paulmartel:
Looks good.

BillWhite:
Looks right to me.

PR #3632, opened 2016-06-01
Don't forget conjuncts in where clauses.

BillWhite:
In ENG-10478 we discovered that we sometimes could forget conjuncts in
where clauses.  For example, in the select statement:

select * from SOMEVIEW where THERENT < 42 AND THERENT < 26000;

where SOMEVIEW is a view, we create an index on the view and use that to
help us evaluate the where expression.  Since both conjuncts are tests
on the same column of SOMEVIEW, we can use one index to answer both
predicates.  But unfortunately we lost the left hand conjunct, and only
used the right hand one.

Plus some other refactoring and unit test creation.

Jenkins seems to be blue on this.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
@@ -492,17 +492,21 @@ final class RangeVariable {
 
             case OpTypes.GREATER :
             case OpTypes.GREATER_EQUAL :
-                indexCondition = e;
+            	indexCondition = makeConjunction(indexCondition, e);
                 break;
 
             case OpTypes.SMALLER :
             case OpTypes.SMALLER_EQUAL :
-                indexEndCondition = e;
+            	indexEndCondition = makeConjunction(indexEndCondition, e);
                 break;
 
             default :
                 Error.runtimeError(ErrorCode.U_S0500, "Expression");
         }
+    }
+
+    private Expression makeConjunction(Expression existingExpr, Expression newExpr) {
+    	return (existingExpr == null) ? newExpr : new ExpressionLogical(OpTypes.AND, existingExpr, newExpr);
     }
 
     /**
@@ -1257,19 +1261,11 @@ final class RangeVariable {
         if (isJoinIndex == true) {
             joinCond = indexCondition;
             if (indexEndCondition != null) {
-                if (joinCond != null) {
-                    joinCond = new ExpressionLogical(OpTypes.AND, joinCond, indexEndCondition);
-                } else {
-                    joinCond = indexEndCondition;
-                }
+            	joinCond = makeConjunction(joinCond, indexEndCondition);
             }
             // then go to the nonIndexJoinCondition
             if (nonIndexJoinCondition != null) {
-                if (joinCond != null) {
-                    joinCond = new ExpressionLogical(OpTypes.AND, joinCond, nonIndexJoinCondition);
-                } else {
-                    joinCond = nonIndexJoinCondition;
-                }
+            	joinCond = makeConjunction(joinCond, nonIndexJoinCondition);
             }
             // then go to the nonIndexWhereCondition
             whereCond = nonIndexWhereCondition;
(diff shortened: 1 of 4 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
```diff
@@ -492,19 +492,23 @@ void addIndexCondition(Expression e, Index index, boolean isJoin) {
...
-                indexEndCondition = e;
+            	indexEndCondition = makeConjunction(indexEndCondition, e);
                 break;
 
             default :
                 Error.runtimeError(ErrorCode.U_S0500, "Expression");
         }
     }
 
+    private Expression makeConjunction(Expression existingExpr, Expression newExpr) {
```
wolffcm: Could this be a static method?

tests/testprocs/org/voltdb_testprocs/regressionsuites/indexes/indexes-ddl.sql
```diff
@@ -126,3 +126,13 @@ CREATE TABLE varbinaryTableTree (
 );
 CREATE INDEX varbinaryTableTree_INDEX_varb2 ON varbinaryTableTree(varb2);
 CREATE INDEX varbinaryTableTree_INDEX_varb512 ON varbinaryTableTree(varb512);
+
+-- ENG-10478
+-- Sometimes we mangle boolean expressions of the form "col < 42 and col < 26661"
+-- when "col" is the same table in both conjuncts.  See TestIndexesSuite.testBooleanExpressions.
+DROP VIEW V_BTEST_R2_ABS IF EXISTS;
+DROP TABLE BTEST_R2 IF EXISTS;
+CREATE TABLE BTEST_R2 (ID INTEGER NOT NULL, WAGE SMALLINT, DEPT SMALLINT, AGE SMALLINT, RENT SMALLINT, PRIMARY KEY (ID));
```
wolffcm: Having one column definition per line is much easier to read.

tests/testprocs/org/voltdb_testprocs/regressionsuites/indexes/indexes-ddl.sql
```diff
@@ -126,3 +126,13 @@ CREATE TABLE varbinaryTableTree (
...
+CREATE VIEW V_BTEST_R2_ABS (V_G1, V_G2, V_CNT, V_sum_age, V_sum_rent) AS SELECT ABS(wage), dept, count(*), sum(age), sum(rent)  FROM BTEST_R2 GROUP BY ABS(wage), dept;
```
wolffcm: Similarly this view is difficult to parse.

wolffcm:
Don't forget to run plannertester, which might be affected by this change.

wolffcm:
The change to HSQL looks fine.  I think a planner tester might be more appropriate to lock down the correct behavior for this ticket rather than a regressionsuite test.  But I don't feel terribly strongly about it.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
BillWhite: Sure.  Why not?  But it's private.  Are static methods measurably faster?

tests/testprocs/org/voltdb_testprocs/regressionsuites/indexes/indexes-ddl.sql
BillWhite: Ok.

BillWhite:
I ran the plannertester with no failures.  I didn't want to let just a plannertester test be the test, since I want to make sure that the answers are really what we want.  But maybe a plannertester test is in order as well.

src/hsqldb19b3/org/hsqldb_voltpatches/RangeVariable.java
wolffcm: As far as I know they're not measurable faster, but it is an indication that the method doesn't change any state of "this" except what's explicitly passed in through input parameters.  This can be useful information to have when browsing the code.

BillWhite:
I'm going to squash merge this on the branch ENG-10478-squash.

wolffcm:
Sorry I made a type in that comment.  I meant to remind you to run "ant plannertester" before merging, and to suggest that a planner JUnit test might be good to have.  But what you have here is fine.

PR #3634, opened 2016-06-02
ENG-10422 Add range checks to TRUNCATE to avoid boost exceptions

wolffcm:
Add EE unit tests for timestamp functions.

The functional changes here are just to avoid the boost exception produced with using the boost functions with years greater than 9999, and to avoid overflow of the int64_t storage for timestamps in the TO_TIMESTAMP method.

In particular, I didn't add code that would throw exceptions if out-of-range timestamps would be produced as output.  I wanted to avoid breaking existing apps as much as possible.  This issue can be addressed in the future ticket that will more properly guard against out-of-range timestamps.


BillWhite:
Looks right to me.

PR #3636, opened 2016-06-02
Eng 10461 improve heap logging

aquafemi:
Heap usage logging even more useful for understanding


third_party/java/src/org/apache/cassandra_voltpatches/GCInspector.java
```diff
@@ -130,11 +130,20 @@ private void logGCResults()
...
-                double usage = (double) memoryUsed / memoryMax;
-
                 if (memoryUsed > .5 * memoryMax)
                 {
-                    logger.warn("Heap is " + usage + " full out of " + memoryMax + ".");
+                    double usage = (double) memoryUsed / memoryMax;
+
+                    String usageStr = String.format("%.2f", usage * 100);
+                    String memoryMaxStr = String.format("%.2f", (double) memoryMax / 1000000);
+
```
vtkstef: should be 1024 * 1024 = 1048576 = 2^20

vtkstef: in java you may have numeric with underscores so setting the number to 1_048_576 would make it more legible

vtkstef:
Reviewed

PR #3637, opened 2016-06-02
Don't forget conjuncts in where clauses.

BillWhite:
In ENG-10478 we discovered that we sometimes could forget conjuncts in
where clauses. For example, in the select statement:

select * from SOMEVIEW where THERENT < 42 AND THERENT < 26000;

where SOMEVIEW is a view, we create an index on the view and use that to
help us evaluate the where expression. Since both conjuncts are tests
on the same column of SOMEVIEW, we can use one index to answer both
predicates. But unfortunately we lost the left hand conjunct, and only
used the right hand one.

Plus some other refactoring and unit test creation.

https://issues.voltdb.com/browse/ENG-10478

### DIFF
--- a/src/ee/storage/table.h
+++ b/src/ee/storage/table.h
@@ -154,7 +154,7 @@ class Table {
 
     /**
      * Includes tuples that are pending any kind of delete.
-     * Used by iterators to determine how many tupels to expect while scanning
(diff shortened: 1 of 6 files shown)


BillWhite:
Closing to redo this squash.

PR #3638, opened 2016-06-02
Patch stackoverflowerror

paulmartel:


wolffcm:
Are you attempting to fix the issue with this test that's occurring in  memcheck mode?  I was just now looking at this ticket:

https://issues.voltdb.com/browse/ENG-10475

I tracked down the cause to HSQL catching the exception in HSQLInterface.getXMLCompiledStatement, where we catch all unexpected exceptions and re-throw them as an HSQLParseException.  Not sure if that's what you're fixing here.

paulmartel:
Squashed, extended, and merged.

PR #3640, opened 2016-06-03
ENG-8154 precursor work to enabling all subqueries on replicated tables

paulmartel:
Mostly invisible changes for correctness and performance -- sometimes having an actual detectable behavior effect only later when partitioned table queries have replicated table subqueries:
Apply recursive expression searches more consistently and simplify their use and implementation
-- critical to later enabling edge cases of replicated table subqueries with partitioned tables in the parent query. This was especially true when it came to applying algorithms consistently between the structures representing a "normal case" parsed select stmt and the "avg push down special cases".
This should have been done from the start.

Integrate the new differentiator as a first class member of TVE rather than an apparent afterthought.

Normalize use of whitespace and braces.

Unrelated bugs found and fixed:
Fix some algorithmic glitches causing typos when explaining agg nodes.
Give a more accurate negative index diagnostic in VoltTableRow.

### DIFF
--- a/src/frontend/org/voltdb/expressions/AbstractExpression.java
+++ b/src/frontend/org/voltdb/expressions/AbstractExpression.java
@@ -840,15 +840,11 @@ public abstract class AbstractExpression implements JSONString, Cloneable {
         return this;
     }
 
-    public ArrayList<AbstractExpression> findBaseTVEs() {
-        return findAllSubexpressionsOfClass(TupleValueExpression.class);
-    }
-
     /**
-     * @param type expression type to search for
-     * @return a list of contained expressions that are of the desired type
+     * @param aeClass AbstractExpression-based class of instances to search for.
+     * @return a list of contained expressions that are instances of the desired class
      */
-    public ArrayList<AbstractExpression> findAllSubexpressionsOfClass(Class< ? extends AbstractExpression> aeClass) {
+    public List<AbstractExpression> findAllSubexpressionsOfClass(Class< ? extends AbstractExpression> aeClass) {
         ArrayList<AbstractExpression> collected = new ArrayList<AbstractExpression>();
         findAllSubexpressionsOfClass_recurse(aeClass, collected);
         return collected;
@@ -856,8 +852,13 @@ public abstract class AbstractExpression implements JSONString, Cloneable {
 
     public void findAllSubexpressionsOfClass_recurse(Class< ? extends AbstractExpression> aeClass,
             ArrayList<AbstractExpression> collected) {
-        if (aeClass.isInstance(this))
+        if (aeClass.isInstance(this)) {
             collected.add(this);
+            // Don't return early, because in a few rare cases,
+            // like when searching for function expressions,
+            // an instance CAN be a parent expression of another instance.
+            // It's probably not worth optimizing for the special cases.
+        }
 
         if (m_left != null) {
             m_left.findAllSubexpressionsOfClass_recurse(aeClass, collected);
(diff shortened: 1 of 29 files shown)


src/frontend/org/voltdb/expressions/ScalarValueExpression.java
```diff
@@ -37,7 +37,9 @@ public void finalizeValueTypes() {
     @Override
     public boolean equals(Object obj) {
         assert(m_left != null);
-        if (obj instanceof ScalarValueExpression == false) return false;
+        if (obj instanceof ScalarValueExpression == false) {
+            return false;
+        }
```
BillWhite: Why is "(obj instanceof ScalarValueExpression == false)" ok but "(false == obj instanceof ScalarValueExpression)" wrong?  Why isn't "(!obj instanceof ScalarValueExpression)" called for here?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1950,30 +1948,77 @@ else if (display_col.columnName != null && ! display_col.columnName.equals(""))
...
+                continue;
+            }
+            Collection<AbstractExpression> found = groupByExpr.findAllSubexpressionsOfClass(aeClass);
+            if (found.isEmpty()) {
+                continue;
+            }
+            exprs.addAll(found);
+        }
+    }
+
```
BillWhite: These two functions, or at least the node schema one, could be useful in the plan nodes.  I do something very similar when generating output schemas, and when validating column index resolutions.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1950,30 +1948,77 @@ else if (display_col.columnName != null && ! display_col.columnName.equals(""))
...
-                exprs.addAll(col.expression.findAllSubexpressionsOfClass(aeClass));
+        if (m_avgPushdownHaving != null &&
+                m_avgPushdownHaving != m_having) {
+            Collection<AbstractExpression> found =
+                    m_avgPushdownHaving.findAllSubexpressionsOfClass(aeClass);
+            if (! found.isEmpty()) {
+                exprs.addAll(found);
             }
         }
+        if (m_avgPushdownGroupByColumns != null && m_avgPushdownGroupByColumns != m_groupByColumns) {
```
BillWhite: Could we have a hint about what special case this is optimizing?  It's not clear to me what's going on with these PushDown columns lists, except that we are swapping columns out and then swapping them back in.  This has a bad code smell to me, but since I don't understand the context I can't tell if it's a good idea or not.

src/frontend/org/voltdb/plannodes/AbstractJoinPlanNode.java
```diff
@@ -194,8 +197,7 @@ public void resolveColumnIndexes()
         IndexScanPlanNode index_scan =
             (IndexScanPlanNode) getInlinePlanNode(PlanNodeType.INDEXSCAN);
         assert(m_children.size() == 2 && index_scan == null);
-        for (AbstractPlanNode child : m_children)
-        {
+        for (AbstractPlanNode child : m_children) {
```
BillWhite: If we put the curly brace on the end of the line here, why not on the end of an else line?

BillWhite:
That's all I've got.  Overall it all looks good to me.  The only thing I'd like to see is an explanation of the PushDownColumns optimization.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -840,24 +840,25 @@ public AbstractExpression replaceAVG () {
...
-    }
-
     /**
-     * @param type expression type to search for
-     * @return a list of contained expressions that are of the desired type
+     * @param aeClass AbstractExpression-based class of instances to search for.
+     * @return a list of contained expressions that are instances of the desired class
      */
-    public ArrayList<AbstractExpression> findAllSubexpressionsOfClass(Class< ? extends AbstractExpression> aeClass) {
+    public List<AbstractExpression> findAllSubexpressionsOfClass(Class< ? extends AbstractExpression> aeClass) {
```
wolffcm: I wonder if there is a way to make this return a list of instances of the subclass you are looking for to avoid having to cast in the callers of this method?  I'm not so familiar with Java generics so I'm not sure.

src/frontend/org/voltdb/expressions/ScalarValueExpression.java
paulmartel: Unfortunately, ! has precedence over instanceof so the best you can do is ` ! (obj instanceof ...)`.
The place where I called it wrong didn't have this problem.
And I actually agree that ! is easy to miss and I usually try to whitespace it to death to make up for that but I don't love that either. If `false == ...` is really a thing (specifically, a commonly used JAVA idiom) unbeknownst to me, I'm more than willing to be schooled, here. I'm having trouble finding anyone arguing this point on the net -- usual problem with googling punctuation. So, got links?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
paulmartel: Read on... There ARE PlanNode equivalents and this changeset makes better use of them -- there may still be room for improvement -- like packaging up stuff that still gets repeated as boilerplate?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
paulmartel: I'll try to dredge something up and augment the comments. Once I remember and can explain what this is doing, we can discuss whether there's a less smelly way of doing it -- none came immediately to mind when Xin first invented it.

src/frontend/org/voltdb/expressions/AbstractExpression.java
paulmartel: Seems like this should not be that hard -- I gave up on a brief attempt at this very problem when I was first learning generics, but maybe it's about due for a second try. Maybe I've incidentally made it easier by tweaking the API.

wolffcm:
I've no more comments.  Looks good to me.  Like Bill I'm curious about the average pushdown optimization.

paulmartel:
I will clarify the comments re: average pushdown after the high-priority merge into 6.4.

PR #3643, opened 2016-06-04
ENG-10475 improve testability of stressor queries that lead to stack …

paulmartel:
…overflow

Includes Chris' change to satisfy the junit and other changes to satisfy sqlcmdtest on multiple platforms.


wolffcm:
Looks good to me.

PR #3645, opened 2016-06-06
ENG-8154 allow any subquery on replicated data.

paulmartel:


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -459,15 +459,11 @@ CompiledPlan getBestCostPlan(AbstractParsedStmt parsedStmt) {
...
-                // no partition tables in parent query
-                for (Table tb: parsedStmt.m_tableList) {
-                    if (! tb.getIsreplicated()) {
-                        m_recentErrorMsg = IN_EXISTS_SCALAR_ERROR_MESSAGE;
-                        return null;
-                    }
-                }
-
-                // no partition tables in subqueries
+                // Don't allow partitioned tables in subqueries.
```
wolffcm: So the change here is that previously there could only be replicated tables in the outer query, and that restriction is being lifted.  In the subqueries, there can still only be replicated tables.  I just want to make sure I have that right.

wolffcm:
This looks good to me.  Basically all the testing that we have for in/exists/scalar subqueries from before is all now being applied to the new cases that we support?  If so, then that's good.

Is there a ticket to turn start testing these kinds of queries in sqlcoverage?

PR #3646, opened 2016-06-06
ENG-10416: While shutting down discard BBContainer to avoid memcheck …

akhanzode:
…fail.


src/frontend/org/voltdb/sysprocs/saverestore/TableSaveFile.java
```diff
@@ -579,6 +579,7 @@ private void readChunksV2() {
                 try {
                     m_chunkReads.acquire();
                 } catch (InterruptedException e) {
+                    fileInputBufferC.discard();
```
vtkstef: maybe finally would be more appropriate?

akhanzode: if acquire is successful we dont want to discard as BBContainer bb is in use.

vtkstef:
Reviewed

PR #3656, opened 2016-06-08
ENG-10527 joint table view guards

yzhang1991:
This pull request adjusts the guards for view defined on joint tables. The guards are against non-inner joins and self-joins. Restrictions on partitioned table join should be enforced when we plan the query, I have already proposed in the standup meeting about this.


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -2923,6 +2927,45 @@ private static AbstractExpression buildPartialIndexPredicate(
...
+    private static void checkViewJoinTree(JoinNode joinTree, VoltCompiler compiler) throws VoltCompilerException {
+        if (joinTree instanceof BranchNode) {
+            BranchNode joinBranch = (BranchNode)joinTree;
+            if (joinBranch.getJoinType() != JoinType.INNER) {
+                throw compiler.new VoltCompilerException("Materialized view only supports INNER JOIN.");
+            }
+            checkViewJoinTree(joinBranch.getLeftNode(), compiler);
+            checkViewJoinTree(joinBranch.getRightNode(), compiler);
+        }
+    }
```
paulmartel: This should be a JoinNode method overridden by BranchNode.
Would it be simpler to have a boolean function return so we don't need to pass the compiler into the recursive call?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -3045,6 +3084,11 @@ private static void checkViewMeetsSpec(String viewName, ParsedSelectStmt stmt,
             throw compiler.new VoltCompilerException(msg.toString());
         }
 
+        checkViewSources(stmt.m_tableList, compiler);
+        if (stmt.m_tableList.size() != 1) {
+            throw compiler.new VoltCompilerException("Materialized views on joint tables are not allowed yet.");
```
paulmartel: You don't have to change the error message for this case from what it was before (at least for now).

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -2481,7 +2511,7 @@ public void testDDLCompilerMatView()
                 "create table t2(id integer not null, num integer, wage integer);\n" +
                 "create view my_view1 (id, num, total) " +
                 "as select t1.id, t2.num, count(*) from t1 join t2 on t1.id = t2.id group by t1.id, t2.num; \n";
-        checkDDLErrorMessage(ddl, "Materialized view \"MY_VIEW1\" has 2 sources. Only one source table is allowed.");
+        checkDDLErrorMessage(ddl, "Materialized views on joint tables are not allowed yet.");
```
paulmartel: This is an inferior error message to the one that was here if it does not echo back the view name.

paulmartel:
End of round.

src/frontend/org/voltdb/planner/parseinfo/BranchNode.java
```diff
@@ -431,4 +431,15 @@ protected boolean replaceChild(JoinNode node) {
...
     }
+
+    /**
+     * Returns if all the join operations within this join tree are inner joins.
+     * @return true or false.
+     */
+    @Override
+    public boolean allInnerJoins() {
+        return m_joinType == JoinType.INNER &&
+               (m_leftNode == null ? true : m_leftNode.allInnerJoins()) &&
```
paulmartel: `X ? true : Y` simplifies to `X || Y`.

PR #3657, opened 2016-06-08
ENG-10557: If not active generation get in overflow.

akhanzode:


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -560,7 +573,7 @@ public void pushExportBuffer(
                 public void run() {
                     try {
                         if (!es.isShutdown()) {
-                            pushExportBufferImpl(uso, buffer, sync, endOfStream);
+                            pushExportBufferImpl(uso, buffer, sync, endOfStream, true);
```
vtkstef: at your own discretion of course, but it'd be nice if we add an inline comment to 'true'

```java
pushExportBufferImpl(uso, buffer, sync, endOfStream, true /* poll */);
```

vtkstef:
Reviewed

src/frontend/org/voltdb/export/ExportDataSource.java
akhanzode: Adding comment....

PR #3658, opened 2016-06-08
ENG-9797 Lift prohibition of TRUE/FALSE literals in VoltDB SQL

yzhang1991:


tests/frontend/org/voltdb/TestAdHocQueries.java
```diff
@@ -860,10 +860,9 @@ public void testXopenSubSelectQueries() throws Exception {
                     "                           AND PNUM > 'P2';";
             try {
                 env.m_client.callProcedure("@AdHoc", adHocQuery);
-                fail("did not fail on static clause");
             }
             catch (ProcCallException pcex) {
-                assertTrue(pcex.getMessage().indexOf("does not support constant Boolean values, like TRUE or FALSE") > 0);
+                fail("failed on static clause");
```
BillWhite: Is this error message a little cryptic?

BillWhite:
This all looks good to me.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLogical.java
```diff
@@ -618,11 +618,15 @@ private void resolveTypesForLogicalOp() {
         }
 
         if (nodes[LEFT].opType == OpTypes.ROW
-                || nodes[RIGHT].opType == OpTypes.ROW
-                || Type.SQL_BOOLEAN != nodes[LEFT].dataType
-                || Type.SQL_BOOLEAN != nodes[RIGHT].dataType) {
+                || nodes[RIGHT].opType == OpTypes.ROW) {
```
wolffcm: I'm not clear on why you needed to make this change... can you explain?

yzhang1991: I actually sent the team an email about this, basically it is just a misleading error message.
Consider the following case:
```SQL
CREATE TABLE a (aint INTEGER NOT NULL);
CREATE TABLE b (aint INTEGER NOT NULL);
SELECT * FROM a JOIN b ON a.aint=b.aint AND 1;
```
We will get the following error message before this commit:

> Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: Error compiling query: org.voltdb.planner.PlanningErrorException: Error in "SELECT * FROM a JOIN b ON a.aint=b.aint AND 1" row column count mismatch

But the problem here was having a non-boolean value in the `ON` clause.
So here I changed the error code from `row column count mismatch` to `data type of expression is not boolean`, whose error code is `X_42568`.

tests/frontend/org/voltdb/TestAdHocQueries.java
yzhang1991: This is a test case that will no longer fail after the fix for the boolean restriction, maybe I should just delete this case?

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLogical.java
wolffcm: That makes sense, thanks for the explanation.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -702,60 +708,73 @@ public void resolveExpressionTypes(Session session) {
...
-                        String tableAlias = (nextColumnEx.rangeVariable.tableAlias != null) ?
-                                nextColumnEx.rangeVariable.tableAlias.name : null;
-                        boolean resolved = false;
-                        for (RangeVariable rv : rangeVariables) {
-                            if (rv.rangeTable.tableName.name.equals(tableName)) {
-                                if (rv.tableAlias == null && tableAlias == null) {
-                                    resolved = true;
-                                } else if (rv.tableAlias != null && tableAlias != null) {
-                                    resolved = tableAlias.equals(rv.tableAlias.name);
+            if (queryCondition.opType == OpTypes.VALUE) {
```
wolffcm: You could have also put this transformation into `ParsedSelectStmt` on the VoltDB side, but this looks okay too.

tests/frontend/org/voltdb/TestAdHocQueries.java
wolffcm: I don't understand why this was supposed to be failing before... or why it started succeeding because of your change?

tests/frontend/org/voltdb/planner/TestPlansJoin.java
```diff
@@ -94,6 +94,10 @@ public void testBasicInnerJoin() {
                       "user lacks privilege or object not found: X");
         failToCompile("select R2.C FROM R1 JOIN R2 ON R1.X = R2.X",
                       "user lacks privilege or object not found: R1.X");
+        failToCompile("select * FROM R1 JOIN R2 ON R1.C = R2.C AND 1",
+                          "data type of expression is not boolean");
+        failToCompile("select * FROM R1 JOIN R2 ON R1.C = R2.C AND MOD(3,1)=1",
+                          "Join with filters that do not depend on joined tables is not supported in VoltDB");
```
wolffcm: This is surprising and I am curious about it but it should not prevent you from merging.  Do we need another ticket for removing another guard somewhere?

tests/frontend/org/voltdb/regressionsuites/TestBooleanLiteralsSuite.java
```diff
@@ -0,0 +1,128 @@
...
+        String sqlBody;
+        Client client = getClient();
+        for (int i=1; i<=4; i++) {
+            populateTable(client, i);
+        }
+
+        long[] resultsForWhere = {3, 0, 3, 0, 0, 3};
+        long[] resultsForOn = {9, 0, 9, 0, 0, 9};
+        long[] resultsForCaseWhen = {1, -1, 1, -1, -1, 1};
+        String whereQuery = "SELECT AINT FROM T%d WHERE %%s";
```
wolffcm: There should be a `HAVING` query in here too.  Also, would it make sense to test a Boolean expression with a parameter, like `? > 100`?

tests/frontend/org/voltdb/regressionsuites/TestBooleanLiteralsSuite.java
```diff
@@ -0,0 +1,128 @@
...
+
+    static String tableDDL = "CREATE TABLE T%d (" +
+                             "AINT INTEGER DEFAULT 0 NOT NULL," +
+                             "PRIMARY KEY (AINT) );";
+    static String partitionDDL = "PARTITION TABLE T%d ON COLUMN AINT;";
+
+    static private void addTableToProject(VoltProjectBuilder project, int tableId, boolean partitioned) {
+        try {
+            project.addLiteralSchema(String.format(tableDDL, tableId));
+            project.addStmtProcedure("Insert" + tableId, "INSERT INTO T" + tableId + " VALUES (?)");
```
wolffcm: You will automatically get a CRUD proc called T1.Insert, etc when you create the table, so I think these procedures are unnecessary.

tests/frontend/org/voltdb/regressionsuites/TestBooleanLiteralsSuite.java
```diff
@@ -0,0 +1,128 @@
...
+        assertEquals(conditions.length, expectedResults.length);
+        VoltTable vt;
+        for (int i=0; i<conditions.length; i++) {
+            String sql = String.format(sqlBody, conditions[i]);
+            vt = client.callProcedure("@AdHoc", sql).getResults()[0];
+            vt.advanceRow();
+            if (examineCount)
+                assertEquals(expectedResults[i], vt.getRowCount());
+            else
+                assertEquals(expectedResults[i], vt.getLong(0));
```
wolffcm: So for the join queries, you just make sure that the number of rows match?  You could use `RegressionSuite.assertContentOfTable` to validate all your queries here.

wolffcm:
Did you look at ExpressionUtil.evaluateExpression and see if any changes are needed?

That's all my comments.

tests/frontend/org/voltdb/TestAdHocQueries.java
yzhang1991: This test got this error before:
>VoltDB does not support constant Boolean values, like TRUE or FALSE

This error does not exist any more.

tests/frontend/org/voltdb/planner/TestPlansJoin.java
yzhang1991: I assume you mean the `R1.C = R2.C AND MOD(3,1)=1` test.
In my opinion, I think this test case should not produce an error message like that and we need to fix this in another ticket.
If we all agree I can open a ticket for this.

tests/frontend/org/voltdb/regressionsuites/TestBooleanLiteralsSuite.java
yzhang1991: My answer for testing `HAVING` is no. There is a more strict guard on the `HAVING` clause, it will guard aginst any non-aggregate columns.
I will check what you said about parameter.

yzhang1991:
Yes, I checked `evaluateExpression()`, I believe there is nothing we should change there.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
yzhang1991: I actually looked into this possibility, but I found the logic there is much more scattered for different parts of the query. Making changes there looks more error-prone.

yzhang1991:
**A note for myself**
I will need to add a comment before merging this branch, in `TestBooleanLiteralsSuite` line 74 about why I am not testing partitioned table joining another partitioned table there.
This case was tested in TestPlansJoin for proper error message because a query joining two partitioned tables on a constant boolean value cannot be planned.
>This query is not plannable.  The planner cannot guarantee that all rows would be in a single partition.

tests/frontend/org/voltdb/planner/TestPlansJoin.java
wolffcm: Yes, I agree, please go ahead and file a ticket.

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -1192,7 +1192,10 @@ private static VoltXMLElement voltGetXMLSpecification(QuerySpecification select,
             if (isGroupByColumn(select, i)) {
                 groupByCols.add(expr);
             } else if (expr.opType == OpTypes.ORDER_BY) {
-                orderByCols.add(expr);
+                if (select.sortAndSlice.hasOrder())
```
wolffcm: It should also be checking that the sortAndSlice has a limit of 0, right?  If for some other reason we added a sortAndSlice, this test might drop the OrderBy erroneously?

tests/frontend/org/voltdb/regressionsuites/TestBooleanLiteralsSuite.java
wolffcm: Sorry I didn't see your reply last week.  It's a surprise to me that there are guards against non-aggregate columns in HAVING.  This seems like non-standard behavior, and there's legitimate reasons one would want this.  Where is the guard on non-aggregates in HAVING?

wolffcm:
Other than my comment on SortAndSlice, this looks good to go to me.

tests/frontend/org/voltdb/regressionsuites/TestBooleanLiteralsSuite.java
yzhang1991: `StatementDMQL.java`, line 1198:
```cpp
if( !(expr instanceof ExpressionLogical && expr.isAggregate) ) {
    throw new org.hsqldb_voltpatches.HSQLInterface.HSQLParseException(
                    "VoltDB does not support HAVING clause without aggregation. " +
                    "Consider using WHERE clause if possible");
}
```

It seems you are right, in PostgreSQL, `having true` is a valid HAVING clause:
```
yzhang=# select count(1) from a;
 count 
-------
     0
(1 row)

yzhang=# select count(1) from a having true;
 count 
-------
     0
(1 row)
```
Maybe I should file a ticket for this?

wolffcm: Yes, please do.  There are some conditions involving non-aggregates that can't be expressed in a WHERE clause, e.g.,
```SQL
SELECT state, SUM(revenue) AS sumrev
FROM taxes
GROUP BY state
HAVING state IN ('NH', 'MA') OR sumrev > 1000000;
```

PR #3659, opened 2016-06-08
Eng 10201 jsonnan

rharding6373:


src/frontend/org/voltdb/VoltTableRow.java
```diff
@@ -897,6 +897,9 @@ void putJSONRep(int columnIndex, JSONStringer js) throws JSONException {
             if (wasNull()) {
                 js.value(null);
             }
+            else if (dvalue != dvalue) {
```
wolffcm: It would be good to use Double.isNan() here for better readability.

Also, what about Inf and -Inf?  The ticket doesn't mention them, but we should be able to handle them as well.

wolffcm:
That's all my comments.  Did you test manually in VMC?

PR #3660, opened 2016-06-09
Fix logic for row length check

wolffcm:
Add unit tests for same (there were none).


paulmartel:
Looks good.

wolffcm:
This is blue in Jenkins so I am merging it.

PR #3662, opened 2016-06-09
Eng 10549 kafka system tests short row counts issue

JadenW:
This should solve following two issues:
1. ImportManager might gets new catalog with new procedure before invocation dispatcher.
    callProcedure will fail with losing kafka records

2. When restarting kafkaStreamimport, channel add assignment may be ignored if one node is still in stopping status.

### DIFF
--- a/src/frontend/org/voltdb/importer/ChannelChangeCallback.java
+++ b/src/frontend/org/voltdb/importer/ChannelChangeCallback.java
@@ -19,6 +19,9 @@ package org.voltdb.importer;
 
 
 public interface ChannelChangeCallback {
+    default public boolean canAcceptAssignments() {
+        return true;
+    }
     void onChange(ImporterChannelAssignment assignment);
     void onClusterStateChange(VersionedOperationMode mode);
 }
--- a/src/frontend/org/voltdb/importer/ChannelChangeCallback.java
+++ b/src/frontend/org/voltdb/importer/ChannelChangeCallback.java
@@ -19,9 +19,6 @@ package org.voltdb.importer;
 
 
 public interface ChannelChangeCallback {
-    default public boolean canAcceptAssignments() {
-        return true;
-    }
     void onChange(ImporterChannelAssignment assignment);
     void onClusterStateChange(VersionedOperationMode mode);
 }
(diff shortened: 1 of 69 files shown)


src/frontend/org/voltdb/importer/ChannelChangeCallback.java
```diff
@@ -19,6 +19,9 @@
 
 
 public interface ChannelChangeCallback {
+    default public boolean canAcceptAssignments() {
```
akhanzode: what is default keyword here?

src/frontend/org/voltdb/importer/ChannelDistributer.java
```diff
@@ -430,13 +430,17 @@ public void registerCallback(String importer, ChannelChangeCallback callback) {
...
+                    if (dispatch.canAcceptAssignments()) {
+                        m_buses.submit(new DistributerRunnable() {
+                            @Override
+                            public void susceptibleRun() throws Exception {
+                                dispatch.onChange(assignment);
+                            }
+                        });
+                        itr.remove();
+                    } else {
+                        LOG.debug("importer " + dispatch + " is stopping, not sending the assigment " + assignment);
```
akhanzode: should message say "channel assignments will be sent when importer is started"

src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
```diff
@@ -141,6 +141,12 @@ private void startImporters(Collection<AbstractImporter> importers)
         }
     }
 
+    @Override
+    public boolean canAcceptAssignments()
+    {
+        return !m_stopping;
```
akhanzode: will there be a race between set and get here being on different threads?

akhanzode:
Can we add a junit for channels getting redelivered?

src/frontend/org/voltdb/importer/ChannelChangeCallback.java
JadenW: Java 8 start support define default method for interface. https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html

This method is only override of ImporterLifeCycleManger

akhanzode: oh ok nice.

vtkstef:
--squash merged

PR #3663, opened 2016-06-09
Added duration of forced compaction to log message

rharding6373:
Verified accuracy of log statements manually.


BillWhite:
Seems right to me.

PR #3665, opened 2016-06-09
ENG-10543, fix the bug that may bypass the snapshot completion check …

luyangco:
…if there are replication tables only,

This will lead to the problem that in recover the restore agent choose the incomplete c/l snapshot to restore.

Change-Id: I8f3690f1f1f9898c467cdf619f7fbafb08a35d0a


luyangco:
r. = stefano

PR #3667, opened 2016-06-10
Reorganize/rename EE MatView modules.

paulmartel:
This is a branch-to-branch pull request to avoid re-review of the previous refactoring.
The effective changes here are to where pieces of code are located, rather than how they work, with a few compromises --
I changed some functions to non-inline to make them callable from multiple modules without having to move all the ugly details and dependencies into a header. There's a possible workaround for this if we see measurable performance regression.


src/ee/storage/MaterializedViewWriteTrigger.cpp
```diff
@@ -14,89 +14,26 @@
...
 #include "storage/persistenttable.h"
-#include "storage/streamedtable.h"
-#include "execution/VoltDBEngine.h"
 #include "execution/ExecutorVector.h"
 #include "plannodes/indexscannode.h"
 #include "executors/abstractexecutor.h"
 
-#include <cassert>
-#include <cstdio>
+//#include <cassert>
```
yzhang1991: Remove them?

build.py
```diff
@@ -352,7 +352,8 @@
  ElasticIndexReadContext.cpp
  ElasticScanner.cpp
  ExportTupleStream.cpp
- MaterializedViewMetadata.cpp
+ MaterializedViewInsertTrigger.cpp
+ MaterializedViewWriteTrigger.cpp
  persistenttable.cpp
```
yzhang1991: I feel these two names are confusing to me, I always have to read the source code to remember what they are actually doing.

src/ee/storage/MaterializedViewInsertTrigger.cpp
```diff
@@ -0,0 +1,398 @@
...
+        delete m_groupByExprs[ii];
+    }
+    for (int ii = 0; ii < m_aggExprs.size(); ++ii) {
+        delete m_aggExprs[ii];
+    }
+    m_target->decrementRefcount();
+}
+
+NValue MaterializedViewInsertTrigger::getAggInputFromSrcTuple(int aggIndex,
+        const TableTuple& tuple) {
```
yzhang1991: It may be better if you could align this parameter with the one in the line above it.
I thought it was with the function body at first glance.
Same for line 103.

src/ee/storage/MaterializedViewInsertTrigger.cpp
```diff
@@ -0,0 +1,398 @@
...
+MaterializedViewInsertTrigger::MaterializedViewInsertTrigger(PersistentTable *destTable,
+                                                             catalog::MaterializedViewInfo *mvInfo)
+    : m_target(destTable)
+    , m_index(destTable->primaryKeyIndex())
+    , m_filterPredicate(parsePredicate(mvInfo))
+    , m_groupByColumnCount(parseGroupBy(mvInfo)) // also loads m_groupByExprs/Columns as needed
+    , m_searchKeyValue(m_groupByColumnCount)
+    , m_aggColumnCount(parseAggregation(mvInfo))
+{
+    VOLT_TRACE("construct materializedViewMetadataBase...");
```
yzhang1991: construct materializedViewMetadataBase...
Should be updated now?

yzhang1991:
Looks like just moving stuff around, but I am not sure where you changed functions to non-inline.

PR #3670, opened 2016-06-13
Cache an empty block in indices.

BillWhite:
Some rows add nodes to indices on the table of the row.  When the last
row of a table is deleted, the nodes for that row are deleted.  If we
delete a row and then add a row repeatedly we will thrash the allocator,
which is expensive.

This commit caches the last allocated block of nodes in the index'
allocator, and reuses it when another node is added to the index.  This
costs some memory for empty tables, but is much faster.

https://issues.voltdb.com/browse/ENG-8872


src/ee/structures/ContiguousAllocator.h
```diff
@@ -34,19 +34,37 @@ class ContiguousAllocator {
         Buffer *prev;
         char data[0];
     };
-
+    /**
```
wolffcm: I appreciate you adding comments to this header.  There are several different terms here--element, block, chunk, allocation, sub-allocation, buffer, node.  I think the proliferation of terms here make this class seem more complicated than it is.

src/ee/structures/ContiguousAllocator.h
```diff
@@ -34,19 +34,37 @@ class ContiguousAllocator {
...
+     * This is the number of elements in use.  The total number of
+     * allocated blocks is this number plus the number of free blocks.
+     */
```
paulmartel: This second part of this comment is not correct. m_count does not count blocks.

wolffcm:
That's all my comments.  I think the new comments in ContiguousAllocator.h are somewhat confusing, but it looks good otherwise.

src/ee/structures/ContiguousAllocator.h
```diff
@@ -34,19 +34,37 @@ class ContiguousAllocator {
...
     int64_t m_count;
+    /**
+     * This is the size of a node from this allocator.
```
paulmartel: I find the term "node" in these comments confusing. These are referred to as "allocations" elsewhere. "Node" seems to imply some kind of connections which are not part of the abstraction at this level.  The allocations only (optionally) become "nodes" when/if the caller casts them to connected data structures.

src/ee/structures/ContiguousAllocator.h
```diff
@@ -34,19 +34,37 @@ class ContiguousAllocator {
...
+     */
     int32_t m_allocSize;
+    /** This is the size of a suballocation for this allocator. */
```
paulmartel: This is the number off allocations that can fit in a block. Emphasizing it as a "size" makes it sound like a number of bytes rather than number of fixed-size allocations. The term "suballocation" used only here seems like an ambiguous synonym for block. It does not fit well with using "allocation" elsewhere to mean a retail allocation vs. a wholesale allocation (block).

src/ee/structures/ContiguousAllocator.h
```diff
@@ -34,19 +34,37 @@ class ContiguousAllocator {
...
+    /** This is the last deleted block if it needs to be cached. */
+    Buffer *m_cachedBuffer;
 
 public:
     /**
      * @param allocSize is the size in bytes of individual allocations.
      * @param chunkSize is the number of allocations per buffer (not bytes).
      */
-    ContiguousAllocator(int32_t allocSize, int32_t chunkSize);
+    ContiguousAllocator(int32_t allocSize, int32_t chunkSize, bool cacheLast = false);
```
paulmartel: What's the justification for the added complexity of making this optional?

src/ee/structures/CompactingHashTable.h
```diff
@@ -177,6 +178,10 @@ namespace voltdb {
 
         /** verification for debugging and testing */
         bool verify();
+        /** Do we have a cached last buffer?  This is used in testing. */
+        bool hasCachedLastBuffer() const { return (m_allocator.hasCachedLastBuffer()); }
+        /** Are we caching the last buffer?  This is used in testing. */
+        bool isCachingLastBuffer() const { return (m_allocator.isCachingLastBuffer()); }
```
paulmartel: isCachingLastBuffer sounds too much like hasCachedLastBuffer (as in is "currently" caching). Would cachesLastBuffer be better?

paulmartel: Would the purpose of these members be clearer is they were named ...LastFreedBuffer? "LastBuffer" by itself could be subject to misinterpretation.

tests/test_apps/insertdelete/ddl.sql
```diff
@@ -0,0 +1,207 @@
...
+DROP TABLE tmp_s4 IF EXISTS;
+DROP TABLE tmp_s5 IF EXISTS;
+DROP TABLE tmp_s6 IF EXISTS;
+DROP TABLE tmp_s7 IF EXISTS;
+DROP TABLE tmp_s8 IF EXISTS;
+DROP TABLE tmp_s9 IF EXISTS;
+
+END_OF_BATCH
+
+-- Update classes from jar to that server will know about classes but not procedures yet.
```
paulmartel: "to that" --> "so that" ?

paulmartel:
End of review round.

wolffcm:
Thanks Bill, the new comments are much clearer.

BillWhite:
Squashed on ENG-8872-squashed.

PR #3672, opened 2016-06-13
Change the algorithm for generated column names to avoid collisions.

BillWhite:
If a user's column name in a subquery happened to be cne of the ones we use for generated
column names we can get wrong answers.  We change the algorithm for generating column
names so that if a nominally named generated name clashes with a user defined name we
add an initial underbar (_).  So, C1 becomes _C1, but only if the user has defined a column
alias C1.

https://issues.voltdb.com/browse/ENG-10497

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionOp.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionOp.java
@@ -130,7 +130,7 @@ public class ExpressionOp extends Expression {
     }
 
     @Override
-	public String getSQL() {
+        public String getSQL() {
 
         StringBuffer sb    = new StringBuffer(64);
         String       left  = getContextSQL(nodes.length > 0 ? nodes[LEFT]
@@ -200,7 +200,7 @@ public class ExpressionOp extends Expression {
     }
 
     @Override
-	protected String describe(Session session, int blanks) {
+        protected String describe(Session session, int blanks) {
 
         StringBuffer sb = new StringBuffer(64);
 
@@ -254,7 +254,7 @@ public class ExpressionOp extends Expression {
     }
 
     @Override
-	public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
+        public HsqlList resolveColumnReferences(RangeVariable[] rangeVarArray,
             int rangeCount, HsqlList unresolvedSet, boolean acceptsSequences) {
 
         if (opType == OpTypes.VALUE) {
@@ -281,7 +281,7 @@ public class ExpressionOp extends Expression {
     }
 
     @Override
-	public void resolveTypes(Session session, Expression parent) {
+        public void resolveTypes(Session session, Expression parent) {
 
         for (int i = 0; i < nodes.length; i++) {
             if (nodes[i] != null) {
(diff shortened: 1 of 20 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionOp.java
```diff
@@ -129,7 +129,8 @@
         this.alias = e.alias;
     }
 
-    public String getSQL() {
+    @Override
+        public String getSQL() {
```
wolffcm: This line should be indented to the same depth as the previous line.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionOp.java
```diff
@@ -198,7 +199,8 @@ public String getSQL() {
         return sb.toString();
     }
 
-    protected String describe(Session session, int blanks) {
+    @Override
+        protected String describe(Session session, int blanks) {
```
wolffcm: Same comment for this line.

wolffcm: Etc.

wolffcm:
Looks good to me.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
```diff
@@ -2120,7 +2120,9 @@ public String describe(Session session, int blanks) {
         sb.append(name).append("(");
 
         for (int i = 0; i < nodes.length; i++) {
-            sb.append("[").append(nodes[i].describe(session)).append("]");
+        	if (nodes[i] != null) {
+        		sb.append("[").append(nodes[i].describe(session)).append("]");
+        	}
```
paulmartel: Non-standard 8-column indents.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
```diff
@@ -1431,7 +1431,12 @@ void setReferenceableColumns() {
         accessibleColumns = new boolean[indexLimitVisible];
 
         IntValueHashMap aliases = new IntValueHashMap();
-
+        // Bundle up all the user defined aliases here.
+        // We can't import java.util.Set because there is a Set
+        // already imported into this class from Hsql itself.
+        java.util.Set<String> userAliases = new java.util.HashSet<String>();
+        // Bundle up all the generated aliases here.
+        java.util.Map<String, Integer> genAliases = new java.util.HashMap<String, Integer>();
```
paulmartel: We've lifted the moratorium on changing imports in hsql code -- they'd help readability/conciseness here, as would use of `<>` for the RHS expressions.

paulmartel:
End of review round.

src/hsqldb19b3/org/hsqldb_voltpatches/QuerySpecification.java
BillWhite: HSQL has its own version of Set which conflicts with java.util.Set.  I wanted to use the latter here, but the former is already in use.  So, I need to fully qualify the name.

PR #3675, opened 2016-06-13
ENG-10486:

wweiss-voltdb:
When a cluster is partitioned due to network failure, we would like to detect that the cluster is not KSafe (all partitions are present) before performing any partition repair process. This patch addresses this issue.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -1071,17 +1071,20 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
     @Override
     public void hostsFailed(Set<Integer> failedHosts)
     {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1077,6 +1077,14 @@ public void hostsFailed(Set<Integer> failedHosts)
                 @Override
                 public void run()
                 {
+                    // First check to make sure that the cluster still is viable before
+                    // before allowing the fault log to be updated by the notifications
+                    // generated below.
+                    Set<Integer> hostsOnRing = new HashSet<Integer>();
+                    if (!m_leaderAppointer.isClusterKSafe(hostsOnRing)) {
```
nshi: I wonder if this will create a race between this runs and the new leaders being promoted. Both of these operations run on different threads and there is no guarantee that this thread won't be stalled for a while.

PR #3676, opened 2016-06-13
ENG-10350 DB not responsive after truncating large table

wolffcm:
This commit adds a "deferred release mode" to our pools for relocatable objects (non-inlined table data).

When deleting large numbers of objects one-by-one, the ContiguousAllocator class can be very inefficient because it's constantly moving the "last" item allocated to fill a hole made by deleting an allocation.

For the most part, we only create objects during transactions, and only delete them when the transaction commits or is rolled back.  This change bundles all the allocations that need to be freed when transactions end so as to avoid excessive memmove operations.

In the reproducible that Ben provided for this ticket, I saw on my machine that truncating a very large table (around 1 million rows) caused the DB to be non-responsive for 100s.  With this fix in place, it only takes 5s.


tests/ee/storage/persistenttable_test.cpp
```diff
@@ -166,6 +174,119 @@ class PersistentTableTest : public Test {
         return payload;
     }
 
+    static const std::string& catalogPayloadForTableWithManyObjects() {
```
paulmartel: I'm guessing that this test payload was generated by some process. It would be useful to give the exact recipe in a comment here. In the case that the format gets changed, we would be better off regenerating an authentic/accurate new version of this input rather than trying to hand-edit it into compliance.

src/ee/execution/VoltDBEngine.h
```diff
@@ -298,6 +298,7 @@ class __attribute__((visibility("default"))) VoltDBEngine {
 
         void releaseUndoToken(int64_t undoToken)
         {
+            ThreadLocalPool::enableDeferredReleaseMode();
```
yzhang1991: Hi Chris, I think I might need your help explain a little bit here to having a better understanding.
Can we keep this deferred release mode always on by default and only call something like releaseNow() to do the actual release work, instead of turning it on and off like this?
Are there other cases I am not aware of, where we do not need this deferred release mode at all so we have to flip the switch here?

tests/ee/storage/persistenttable_test.cpp
wolffcm: Yes, I agree.  All I did was set a breakpoint in Eclipse as we were processing the DDL and copied the string into the unit test.  I'll update the test with a comment.

It would be great if it was easier to get the payload from the DDL, perhaps with a standalone tool.  I will file a ticket for that.

PR #3679, opened 2016-06-14
run_voltdb_server.sh: fixed to expect the new "Unable to connect to

johnTcrawford:
VoltDB cluster" message (rather than the old "Connection refused"); and
cleaned up to give better feedback, the next time there is a similar
problem, with waiting for the VoltDB server to complete initialization -
e.g., an unrecognized message is assumed to be a problem, not assumed to
be a successful run of ddl.sql in sqlcmd.


BillWhite:
Looks plausible to me.

PR #3680, opened 2016-06-14
Eng 10444

manjujames:
Added warning in DrConsumerDrIdTracker class about it not being thread-safe. 
Separate commit in pro repo removing unsafe getSafePointDrId call from subscribe thread.

### DIFF
--- a/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
+++ b/src/frontend/org/voltdb/DRConsumerDrIdTracker.java
@@ -27,6 +27,7 @@ import java.util.Iterator;
 import org.json_voltpatches.JSONArray;
 import org.json_voltpatches.JSONException;
 import org.json_voltpatches.JSONObject;
+import org.voltcore.logging.VoltLogger;
 import org.voltdb.iv2.UniqueIdGenerator;
 
 import com.google_voltpatches.common.collect.BoundType;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/DRConsumerDrIdTracker.java
```diff
@@ -279,8 +283,11 @@ public boolean contains(long startDrId, long endDrId) {
      * @return The current safe-to-ack DrId
      */
     public long getSafePointDrId() {
-        assert (!m_map.isEmpty());
-        return end(m_map.asRanges().iterator().next());
+        // Acquire lock to avoid seeing bad data while another thread is in truncate()
+        synchronized (m_map) {
```
luyangco: Looks good to me, but worth to protect `getFirstDrId()` and `getLastDrId()` too, they'll throw `NoSuchElementException` also if `m_map` is empty.

manjujames: Those are called even from toShortString() though.

src/frontend/org/voltdb/DRConsumerDrIdTracker.java
```diff
@@ -234,19 +234,23 @@ public void append(long startDrId, long endDrId, long spUniqueId, long mpUniqueI
...
-                iter.remove();
-            } else if (next.contains(newTruncationPoint)) {
-                iter.remove();
-                m_map.add(range(newTruncationPoint, end(next)));
-                return;
-            } else {
-                break;
+        // Get a lock so that getSafePointDrId will not see empty map
+        // if another thread calls it while this is in the middle of truncation
+        synchronized(m_map) {
```
luyangco: Stylish nitpick, put space between `synchronized` and left parenthese

src/frontend/org/voltdb/DRConsumerDrIdTracker.java
luyangco: Right now only debug output would call `toShortString()`, but still needs to be protected just in case we call it from another thread.

luyangco:
Looks good to me.

PR #3683, opened 2016-06-14
Add guards for windowed functions.

BillWhite:
We have added initial support for windowed aggregate functions.  The
planner can plan for them, but the EE cannot execute the resulting
plans.  This commit adds guards in the planner to keep these plans from
being generated.  It's possible to disable the guards in the server by
setting the system property org.voltdb.handlewindowedfunctions to true.

https://issues.voltdb.com/browse/ENG-10602


wolffcm:
Looks good.

PR #3691, opened 2016-06-16
ENG-10529-join-view-DDLCompiler

yzhang1991:


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1343,6 +1346,22 @@ template<class TABLE> static void initMaterializedViews(catalog::Table *srcCatal
         }
 
     }
+
+    BOOST_FOREACH (LabeledViewHandler labeledViewHandler, srcCatalogTable->mvHandler()) {
```
yzhang1991: This part is just for printing the query plan for checking. I will remove it after the work is done.

src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -375,7 +375,8 @@ protected String checkAddDropWhitelist(final CatalogType suspect, final ChangeTy
             // So, in short, all of these constraints will pass or fail tests of other catalog differences
             // Even if they did show up as Constraints in the catalog (for no apparent functional reason),
             // flagging their changes here would be redundant.
-            suspect instanceof Constraint)
+            suspect instanceof Constraint ||
+            suspect instanceof ViewTrigger)
```
yzhang1991: What if a table is changed to `export only` at runtime, the createQuery cannot be planned anymore.

    (ERROR: Illegal to read a stream.)

I should ban view on joint tables containing export tables before I could think of a good solution for this.

PR #3693, opened 2016-06-16
Add time stamp validity functions

BillWhite:
This commit adds the timestamp validity functions MIN_VALID_TIMESTAMP, MAX_VALID_TIMESTAMP and IS_VALID_TIMESTAMP.

Along with this we added a valgrind suppression file which suppresses false positive memory leak warnings.  We also added a boolean comparison in the EE.  This is necessary to test the results of boolean valued functions.  Finally, we added a testNullary() test operation in the EE tests.

https://issues.voltdb.com/browse/ENG-10559

### DIFF
--- a/src/frontend/org/voltdb/expressions/FunctionExpression.java
+++ b/src/frontend/org/voltdb/expressions/FunctionExpression.java
@@ -359,6 +359,10 @@ public class FunctionExpression extends AbstractExpression {
                 connector = ", ";
             }
             result += ")";
+        } else {
+        	// The two functions MIN_VALID_TIMESTAMP and MAX_VALID_TIMESTAMP
+        	// are nullary.  Others may be in the future.
+        	result += "()";
         }
         return result;
     }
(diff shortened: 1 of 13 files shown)


src/frontend/org/voltdb/expressions/FunctionExpression.java
```diff
@@ -359,6 +359,10 @@ public String explain(String impliedTableName) {
                 connector = ", ";
             }
             result += ")";
+        } else {
+            // The two functions MIN_VALID_TIMESTAMP and MAX_VALID_TIMESTAMP
+            // are nullary.  Others may be in the future.
+            result += "()";
```
paulmartel: I was under the impression that SQL expects NOT to see parens for a nullary function call.

paulmartel: Nevermind? Apparently PI also needs the parens, so there's precedent -- and yet the unary function NOW only works without parens -- so there's precedent for inconsistency. There should be a ticket for making unary function call syntax consistent.

paulmartel:
Good to go.

src/frontend/org/voltdb/expressions/FunctionExpression.java
BillWhite: I didn't add the parentheses and I got an error.  When I add the parentheses I get no errors.

BillWhite:
I'm closing this pull request, and squashing this.

PR #3698, opened 2016-06-17
Bypass unused hsql LobManager

paulmartel:


BillWhite:
Thanks.   This was often annoying if one damages HSQL enough to make parsing the LOB schema break.  It's not clear how to debug it.  I'm glad to see that this is unnecessary and that it can be eliminated.

PR #3701, opened 2016-06-17
VDM-127 VDM stop node REST API

zeemanhe:
(1)flip force stop server flag
(2)fix VoltDatabase.kill_server()


akhanzode:
This looks good. check tests for VDM and if looking good squash merge.

PR #3708, opened 2016-06-20
Limit 0 should really return 0 rows.

BillWhite:
The algorithm in the order by executor for limit nodes was incorrect.
This commit makes limit 0 do almost no work.  It was reading the input
table and sorting it.  The sort was a special case which only does a
partial sort, but copying the input table into a std::vector is
unnecessary and can't be zero cost.

https://issues.voltdb.com/browse/ENG-10559

### DIFF
--- a/src/ee/executors/orderbyexecutor.cpp
+++ b/src/ee/executors/orderbyexecutor.cpp
@@ -131,49 +131,56 @@ OrderByExecutor::p_execute(const NValueArray &params)
     VOLT_TRACE("Input Table:\n '%s'", input_table->debug().c_str());
     TableIterator iterator = input_table->iterator();
     TableTuple tuple(input_table->schema());
-    vector<TableTuple> xs;
-    ProgressMonitorProxy pmp(m_engine, this);
-    while (iterator.next(tuple))
-    {
-        pmp.countdownProgress();
-        assert(tuple.isActive());
-        xs.push_back(tuple);
-    }
-    VOLT_TRACE("\n***** Input Table PreSort:\n '%s'",
-               input_table->debug().c_str());
+
+    // If limit == 0 we have no work here.  There's no need to sort anything,
+    // or to fetch the vector of tuples from the input.  If limit < 0 we
+    // need to do the loop below, though.  The only case where we can skip
+    // is if limit == 0.
+    if (limit != 0) {
+        vector<TableTuple> xs;
+        ProgressMonitorProxy pmp(m_engine, this);
+        while (iterator.next(tuple))
+        {
+            pmp.countdownProgress();
+            assert(tuple.isActive());
+            xs.push_back(tuple);
+        }
+        VOLT_TRACE("\n***** Input Table PreSort:\n '%s'",
+                   input_table->debug().c_str());
 
 
-    if (limit >= 0 && xs.begin() + limit + offset < xs.end()) {
-        // partial sort
-        partial_sort(xs.begin(), xs.begin() + limit + offset, xs.end(),
-                AbstractExecutor::TupleComparer(node->getSortExpressions(), node->getSortDirections()));
-    } else {
-        // full sort
-        sort(xs.begin(), xs.end(),
-                AbstractExecutor::TupleComparer(node->getSortExpressions(), node->getSortDirections()));
-    }
-
-    int tuple_ctr = 0;
-    int tuple_skipped = 0;
-    for (vector<TableTuple>::iterator it = xs.begin(); it != xs.end(); it++)
-    {
-        //
-        // Check if has gone past the offset
-        //
-        if (tuple_skipped < offset) {
-            tuple_skipped++;
-            continue;
+        if (limit >= 0 && xs.begin() + limit + offset < xs.end()) {
+            // partial sort
+            partial_sort(xs.begin(), xs.begin() + limit + offset, xs.end(),
+                    AbstractExecutor::TupleComparer(node->getSortExpressions(), node->getSortDirections()));
+        } else {
+            // full sort
+            sort(xs.begin(), xs.end(),
+                    AbstractExecutor::TupleComparer(node->getSortExpressions(), node->getSortDirections()));
         }
 
-        VOLT_TRACE("\n***** Input Table PostSort:\n '%s'",
-                   input_table->debug().c_str());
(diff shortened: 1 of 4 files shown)


src/ee/executors/orderbyexecutor.cpp
```diff
@@ -131,49 +131,56 @@ OrderByExecutor::p_execute(const NValueArray &params)
...
-        xs.push_back(tuple);
-    }
-    VOLT_TRACE("\n***** Input Table PreSort:\n '%s'",
-               input_table->debug().c_str());
 
+    // If limit == 0 we have no work here.  There's no need to sort anything,
+    // or to fetch the vector of tuples from the input.  If limit < 0 we
+    // need to do the loop below, though.  The only case where we can skip
+    // is if limit == 0.
+    if (limit != 0) {
```
yzhang1991: Could you change this to `limit <= 0`?

src/ee/executors/orderbyexecutor.cpp
```diff
@@ -131,49 +131,56 @@ OrderByExecutor::p_execute(const NValueArray &params)
...
-    int tuple_skipped = 0;
-    for (vector<TableTuple>::iterator it = xs.begin(); it != xs.end(); it++)
-    {
-        //
-        // Check if has gone past the offset
-        //
-        if (tuple_skipped < offset) {
-            tuple_skipped++;
-            continue;
+        if (limit >= 0 && xs.begin() + limit + offset < xs.end()) {
```
yzhang1991: Now, remove `limit >= 0` here because this code is only reachable if `limit > 0`.

src/ee/executors/orderbyexecutor.cpp
```diff
@@ -131,49 +131,56 @@ OrderByExecutor::p_execute(const NValueArray &params)
...
-        //
-        if (limit >= 0 && ++tuple_ctr >= limit) {
-            break;
+        int tuple_ctr = 0;
+        int tuple_skipped = 0;
+        // If (limit < 0), so we don't have a limit at all, then just compare
+        // the iterator with the end.  Otherwise check that the tuple_counter is
+        // not over the limit.
+        for (vector<TableTuple>::iterator it = xs.begin();
+             ((limit < 0) ? true : (tuple_ctr < limit)) && it != xs.end();
```
yzhang1991: Similarly, could you replace `(limit < 0) ? true : (tuple_ctr < limit))` with just `tuple_ctr < limit`?

tests/frontend/org/voltdb/regressionsuites/TestLimitOffsetSuite.java
```diff
@@ -351,6 +351,68 @@ public void testSubqueryLimit() throws Exception {
...
+        cr = client.callProcedure("PLAINJANE.insert", 200, 201, 202);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure("PLAINJANE.insert", 300, 301, 302);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure("PLAINJANE.insert", 400, 401, 402);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure("PLAINJANE.insert", 500, 501, 502);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+
+        for (int idx = 0; idx < 5; idx += 1) {
```
yzhang1991: I want to suggest three changes for this test:
* Might need to check the table content to make sure the order is correct, just checking rowCount is not enough.
  You can refer to `testBooleanLiteralsInQueries()` in `TestBooleanLiteralsSuite.java` for this.
* Might need to check more combinations:
  * Is the ORDER BY in ASC | DESC order?
  * Does the query have an OFFSET clause or not?

yzhang1991:
End of review round.

src/ee/executors/orderbyexecutor.cpp
```diff
@@ -131,49 +131,56 @@ OrderByExecutor::p_execute(const NValueArray &params)
...
+            if (tuple_skipped < offset) {
+                tuple_skipped++;
+                continue;
+            }
+
+            VOLT_TRACE("\n***** Input Table PostSort:\n '%s'",
+                       input_table->debug().c_str());
+            output_table->insertTupleNonVirtual(*it);
+            pmp.countdownProgress();
+            tuple_ctr += 1;
```
yzhang1991: ++tuple_ctr;

BillWhite: I always use tuple_ctr += 1.  I think it seems more natural, but I could be wrong.

I don't like little notations, like !someBoolean or someInt++ or ++someInt.  I'd rather see something grander.

tests/frontend/org/voltdb/regressionsuites/TestLimitOffsetSuite.java
BillWhite: We check that order by does the right thing in other tests.  This test is just to see that we are getting the right number of rows.

src/ee/executors/orderbyexecutor.cpp
BillWhite: No, because limit == -1 means there is no limit.  So if I change this to limit > 0 I will not catch that case.

src/ee/executors/orderbyexecutor.cpp
BillWhite: Again, this is not redundant.  If limit == -1, so that there is no limit, we don't want to take this branch at all.  I could test "limit > 0" here, but that's hardly clearer or faster.

src/ee/executors/orderbyexecutor.cpp
BillWhite: No, because if limit < 0 we don't care about tuple_ctr at all.  If we use "tuple_ctr < limit" and limit is -1 we will never get true, and no rows will be output.

yzhang1991: Sorry I didn't know the -1 case, then this could be `limit < 0 || tuple_ctr < limit`.
But I don't know, maybe you prefer to stay this way to be more natural.

BillWhite:
I'm going to squash this on a separate branch.

PR #3709, opened 2016-06-20
ENG-10618: Fix mesh arbiter fault resolution algorithm.

nshi:
On single link failures, a host A may claim another host B is down
because host C witnessed the link failure between B and C. Same goes for
C. In this case, A has the freedom to go with either B or C. Both B and
C should wait until A makes the decision before they continue to make
theirs. Otherwise B and C may both think they still have A in their
subclusters and continue on because they are the majority.


src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -347,6 +357,13 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
         sfmb.safeTxnIds(getSafeTxnIdsForSites(hsIds));
 
         SiteFailureMessage sfm = sfmb.build();
+
+        // Send the decision to each host
+        for (long failedSite : sfm.getFailedSites()) {
+            m_recoveryLog.info("Agreement, Telling " + CoreUtils.hsIdToString(failedSite) + " it is excluded");
```
vtkstef: send it only to destinations that you have not witnessed their failure

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -460,6 +478,31 @@ private boolean discoverGlobalFaultData_rcv(Set<Long> hsIds) {
 
                 updateFailedSitesLedger(hsIds, sfm);
 
+                if (sfm.getFailedSites().contains(m_hsId)) {
```
vtkstef: are you checking that it is a decision

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -460,6 +478,31 @@ private boolean discoverGlobalFaultData_rcv(Set<Long> hsIds) {
...
+                    // Wait for the final decision from this host
+                    if (m_sitesPendingDecisions.add(m.m_sourceHSId)) {
+                        m_recoveryLog.info("Agreement, Will wait for final decision from " + CoreUtils.hsIdToString(m.m_sourceHSId) +
+                                           " because it claimed the local host gone");
+                    }
+                }
+                if (!sfm.m_decision.isEmpty()) {
```
vtkstef: I would not track what could become decisions. I would only track suicides that are decided

nshi: It's essentially only tracking suicides that are decided. If it's not a suicide, the following if condition will be false.

src/frontend/org/voltcore/agreement/MeshArbiter.java
nshi: I don't think it matters whether it's a decision or not here. If it's not a decision, we definitely want to track that so that we can wait for the final decision from that host. If it's a decision, it will be canceled immediately in the block below, essentially a no-op.

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +358,91 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+            if (msg == null) {
+                // Send a heartbeat to keep the dead host timeout active.
+                m_meshAide.sendHeartbeats(m_seeker.getSurvivors());
+                if (System.currentTimeMillis() - start > 20000) {
+                    m_recoveryLog.error("Agreement, still waiting for " +
+                    CoreUtils.hsIdCollectionToString(Sets.difference(expectedSurvivors, m_decidedSurvivors.keySet())) +
+                    " expected " + CoreUtils.hsIdCollectionToString(expectedSurvivors) + " decided " +
+                    CoreUtils.hsIdCollectionToString(m_decidedSurvivors.keySet()));
+                    start = System.currentTimeMillis();
+                }
```
nshi: I'm not sure if this block is worth keeping. I'll remove it if not.

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +358,91 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+                continue;
+            }
+
+            if (m_hsId != msg.m_sourceHSId && !expectedSurvivors.contains(msg.m_sourceHSId)) {
+                m_recoveryLog.info("Received message from failed site " + CoreUtils.hsIdToString(msg.m_sourceHSId) + " " + msg);
```
nshi: I'll demote this to debug.

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +358,91 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+                // Ignore messages from failed sites
+                continue;
+            }
+
+            m_recoveryLog.info("Agreement, received message " + msg);
```
nshi: This debug message will be removed.

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +358,91 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+            if (expectedSurvivors.contains(remoteDecision.m_sourceHSId)) {
+                if (remoteDecision.m_decision.contains(m_hsId)) {
+                    m_decidedSurvivors.clear();
+                    m_recoveryLog.info("Agreement, Received inconsistent decision from " +
+                                       CoreUtils.hsIdToString(remoteDecision.m_sourceHSId) + ", " + remoteDecision);
+                    final FaultMessage localFault = new FaultMessage(m_hsId, remoteDecision.m_sourceHSId);
+                    localFault.m_sourceHSId = m_hsId;
+                    m_mailbox.deliverFront(localFault);
+                    return false;
+                } else if (!sfm.m_survivors.equals(remoteDecision.m_survivors)) {
```
vtkstef: just to be thorough is this equals implementation doing the right thing?

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +358,91 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+
+        long start = System.currentTimeMillis();
+        do {
+            final VoltMessage msg = m_mailbox.recvBlocking(receiveSubjects, 5);
+            if (msg == null) {
+                // Send a heartbeat to keep the dead host timeout active.
+                m_meshAide.sendHeartbeats(m_seeker.getSurvivors());
+                if (System.currentTimeMillis() - start > 20000) {
+                    m_recoveryLog.error("Agreement, still waiting for " +
+                    CoreUtils.hsIdCollectionToString(Sets.difference(expectedSurvivors, m_decidedSurvivors.keySet())) +
```
vtkstef: warning if you see a lot of this in your tests

src/frontend/org/voltcore/agreement/MeshArbiter.java
vtkstef: anything that can help diagnostically

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +358,91 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+                    }
+                } else {
+                    m_mailbox.deliverFront(fm);
+                    return false;
+                }
+            } else if (msg.getSubject() == Subject.FAILURE.getId()) {
+                final FaultMessage fm = (FaultMessage) msg;
+                if (!fm.decided) {
+                    // In case of concurrent fault, handle it
+                    m_mailbox.deliverFront(msg);
```
vtkstef: if the source id is yourself should you redeliver it?

src/frontend/org/voltcore/agreement/MeshArbiter.java
nshi: Yes, I checked the java doc. It does what you expect it to do, which is to compare each element in both sets.

src/frontend/org/voltcore/agreement/MeshArbiter.java
nshi: Exactly, I don't expect to see any in test. I used this for debugging.

src/frontend/org/voltcore/agreement/MeshArbiter.java
nshi: Good point. When will I deliver a decision FaultMessage to myself?

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +361,105 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+                    m_recoveryLog.error("Agreement, Still waiting for decisions from " +
+                                        CoreUtils.hsIdCollectionToString(Sets.difference(expectedSurvivors, m_decidedSurvivors.keySet())) +
+                                        " after " + TimeUnit.MILLISECONDS.toSeconds(duration) + " seconds");
+                    start = System.currentTimeMillis();
+                }
+                continue;
+            }
+
+            if (m_hsId != msg.m_sourceHSId && !expectedSurvivors.contains(msg.m_sourceHSId)) {
+                // Ignore messages from failed sites
```
vtkstef: do you contain the this guy in your survivor set too?

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -351,11 +361,105 @@ protected void notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
...
+            m_recoveryLog.warn("Agreement, " + m_localHistoricDecisions.size() +
+                               " local decisions have been made without converging");
+        }
+        for (SiteFailureMessage lhd : m_localHistoricDecisions) {
+            if (lhd.m_survivors.equals(sfm.m_survivors)) {
+                m_recoveryLog.info("Agreement, detected decision loop. Exiting");
+                return true;
+            }
+        }
+        m_localHistoricDecisions.add(sfm);
```
vtkstef: is this cleared somewhere too?

src/frontend/org/voltcore/agreement/MeshArbiter.java
nshi: No. If it gets here, it means the message is from a host not in the survivor set. We drop these messages on the floor on master, which is what I'm doing here now.

src/frontend/org/voltcore/agreement/MeshArbiter.java
nshi: It's cleared here https://github.com/VoltDB/voltdb/pull/3709/files#diff-84d2878443675f50bec9431f2d105054R462

PR #3712, opened 2016-06-21
ENG-10354, fix NPE during elastic join. (#3702)

luyangco:
NPE is caused by using default CRUD procedure while elastic joining, CI fails to
lookup default procedure before it thinks this is a sysproc.


luyangco:
r. = nshi

PR #3713, opened 2016-06-21
ENG-10641: Populate and use fragment parameters as before when replay…

slowsoul:
…ing ApplyBinaryLogMP on cluster recover, use only invocation parameters for normal execution


luyangco:
Review complete.

PR #3714, opened 2016-06-21
Streamline voltGetXML recursion, add rank sublist column resolution

paulmartel:


BillWhite:
Seems alright to me.  Your changes won't interact with my current changes, mostly.  The TestWindowedExpression.java will change a bit, but not all that much, now that a PartitionByPlanNode is a kind of AggregatePlanNode.

PR #3717, opened 2016-06-22
Fixed a bug in SnapshotSaveAPI.java

yt14:
- Update startSnapshotting() to show the table name when doing 
voltadmin save --tables=replicated-table

When snapshotting replicated table only, there are chance that earlyResultTable != null while result table is not empty. It's because in this case the site who creates snapshot targets is not the same one who do the actual job.

Table name can be shown now.

### DIFF
--- a/src/frontend/org/voltdb/SnapshotSaveAPI.java
+++ b/src/frontend/org/voltdb/SnapshotSaveAPI.java
@@ -307,8 +307,8 @@ public class SnapshotSaveAPI
             earlyResultTable = result;
         }
 
-        // If earlyResultTable is set, return here
-        if (earlyResultTable != null) {
+        // If earlyResultTable is set and result table is empty, return here
+        if (earlyResultTable != null && result.getRowCount() == 0) {
             if (runPostTasks) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/SnapshotSaveAPI.java
```diff
@@ -307,8 +307,8 @@ public void run()
             earlyResultTable = result;
         }
 
-        // If earlyResultTable is set, return here
-        if (earlyResultTable != null) {
+        // If earlyResultTable is set and result table is empty, return here
+        if (earlyResultTable != null && result.getRowCount() == 0) {
```
nshi: Looking more at how the earlyResultTable is used, I'm slightly concerned that this check may hide some errors we want to return in the earlyResultTable and miss performing the post snapshot tasks in this if block.

The issue may be that we are not setting the result in `m_createResult` to earlyResultTable here https://github.com/VoltDB/voltdb/pull/3717/files#diff-495fb907cdc93322c12e76194931bbddR250.

PR #3719, opened 2016-06-22
Long running reads [DO NOT MERGE]

rharding6373:


src/ee/common/executorcontext.cpp
```diff
@@ -83,7 +83,8 @@ ExecutorContext::ExecutorContext(int64_t siteId,
                 CatalogId hostId,
                 AbstractDRTupleStream *drStream,
                 AbstractDRTupleStream *drReplicatedStream,
-                CatalogId drClusterId) :
+                CatalogId drClusterId,
+                std::string pathName) :
```
BillWhite: Could pathName have a better name, here and in voltdbjni.cpp and other places it shows up?

Maybe a comment telling what it's there for would be sufficient.

src/ee/common/executorcontext.hpp
```diff
@@ -150,6 +151,33 @@ class ExecutorContext {
         return m_topEnd;
     }
 
+    std::string getPathName() {
+        return m_pathName;
+    }
+
+    std::string nextOutFileName() {
+        // 255 is typical system limit on file name size
+        char outFileChars[255];
```
BillWhite: It is possible for m_pathName to be too long and overflow outFileChars?  I know you won't get a crash, but you don't seem to have a check for it.

I don't think there's any reason to skimp on stack space.  I would use 1024 here, or just use a std::stringstream.  I'm not sure this is a performance path operation.

src/ee/common/NValue.hpp
```diff
@@ -3129,6 +3132,77 @@ inline void NValue::serializeTo(SerializeOutput &output) const {
                               "with ValueType '%s' that is not handled", getValueTypeString().c_str());
 }
 
+/**
+ * Serialize this NValue to the provided SerializeOutputFile
+ */
+inline void NValue::serializeToFile(SerializeOutputFile &output) const {
```
BillWhite: Bummer.  If serializeTo was somehow a template, with the buffer management functions relegated to some utility class, then maybe there wouldn't be so much code duplication here.  But this is probably good enough.

src/ee/common/serializeio.h
```diff
@@ -594,7 +594,102 @@ class CopySerializeOutput : public SerializeOutput {
 };
 
 
+/** Class for writing to disk. */
+class SerializeOutputFile {
+public:
```
BillWhite: Too bad SerializeOutput isn't a template, with writePrimitive in the template class.

src/ee/storage/table.cpp
```diff
@@ -280,6 +280,56 @@ size_t Table::getColumnHeaderSizeToSerialize(bool includeTotalSize) const {
     return bytes;
 }
 
+bool Table::serializeColumnHeaderToFile(SerializeOutputFile &serialize_iof) {
+
```
BillWhite: Doesn't this and serializeColumnHeader seem like they do almost the same thing?  Could they be template instances?

Maybe not.

BillWhite:
Seems ok to me.  It would be good to generalize SerializeTo and SerializeToFile as templated functions somehow, but that might be more work than is warranted here.

src/ee/common/NValue.hpp
paulmartel: But there IS a trivial way to eliminate the code duplication HERE, though not (as easily) in serializeio.h
And that's to parameterize just the NValue method:

    template<class SO> void serializeTo(SerializeOutput &output) const;

    template<class SO> inline void NValue::serializeTo(SO& output) const {

src/ee/common/executorcontext.cpp
```diff
@@ -94,12 +95,17 @@ ExecutorContext::ExecutorContext(int64_t siteId,
...
+    m_outFileName(""),
+    m_outFileCount(0),
     m_lastCommittedSpHandle(0),
     m_siteId(siteId),
     m_partitionId(partitionId),
     m_hostname(hostname),
     m_hostId(hostId),
     m_drClusterId(drClusterId)
+
+
```
paulmartel: avoid excessive whitespace

src/ee/common/tabletuple.h
```diff
@@ -415,6 +425,7 @@ class TableTuple {
     void deserializeFrom(voltdb::SerializeInputBE &tupleIn, Pool *stringPool);
     void deserializeFromDR(voltdb::SerializeInputLE &tupleIn, Pool *stringPool);
     void serializeTo(voltdb::SerializeOutput &output, bool includeHiddenColumns = false);
+    void serializeToFile(voltdb::SerializeOutputFile &output, bool includeHiddenColumns = false);
```
paulmartel: This also could be templatized?

src/ee/executors/sendexecutor.h
```diff
@@ -62,6 +62,7 @@ class SendExecutor : public AbstractExecutor
         : AbstractExecutor(engine, abstractNode)
     {
         m_engine = engine;
+        m_highVolume = false;
```
paulmartel: You wouldn't know it from all our bad examples, but initializer lists should generally be used rather than in-constructor-body assignments. Bonus: allows declaration of const data members. Feel free to evolve the existing (nearby) bad example code for consistency with this better approach.

src/ee/plannodes/sendnode.cpp
```diff
@@ -60,6 +60,12 @@ std::string SendPlanNode::debugInfo(const std::string &spacer) const
     return (buffer.str());
 }
 
-void SendPlanNode::loadFromJSONObject(PlannerDomValue obj) { }
+void SendPlanNode::loadFromJSONObject(PlannerDomValue obj) {
+    if(obj.hasKey("HIGH_VOLUME")) {
+        m_highVolume = obj.valueForKey("HIGH_VOLUME").asBool();
+    } else {
```
paulmartel: Space after `if` is conventional for VoltDB. 
`else` ( and `catch` and `finally`) starting on its own line is something I'm trying to encourage as a new convention at least within the CORE Team.

src/ee/storage/table.cpp
paulmartel: Why not?

src/ee/storage/table.cpp
```diff
@@ -346,6 +396,36 @@ bool Table::serializeColumnHeaderTo(SerializeOutput &serialize_io) {
 
 }
 
+
+bool Table::serializeToFile(SerializeOutputFile &serialize_iof) {
```
paulmartel: Yes another job for templates.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1477,6 +1516,100 @@ private final void createAdHocTransaction(final AdHocPlannedStmtBatch plannedStm
...
+        assert(isSinglePartition);
+        assert(plannedStmtBatch.isReadOnly());
+        task.procName = "@ReadOnlySlow";
+
+        int type = VoltType.NULL.getValue();
+        // replicated table read is single-part without a partitioning param
+        // I copied this from below, but I'm not convinced that the above statement is correct
+        // or that the null behavior here either (a) ever actually happens or (b) has the
+        // desired intent.
+        Object partitionParam = plannedStmtBatch.partitionParam();
```
paulmartel: I'm suddenly confused WRT whether we support a (rigged) partitioning parameter even when we plan the ad hoc query without parameterizing it. We must? It seems there's no other way this could work on partitioned table data.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1477,6 +1516,100 @@ private final void createAdHocTransaction(final AdHocPlannedStmtBatch plannedStm
...
+        partition = TheHashinator.getPartitionForParameter(type, partitionParam);
+        // Send the partitioning parameter and its type along so that the site can check if
+        // it's mis-partitioned. Type is needed to re-hashinate for command log re-init.
+        task.setParams(param, (byte)type, buf.array());
+
+        task.clientHandle = plannedStmtBatch.clientHandle;
+
+        ClientResponseImpl error = null;
+        if (VoltDB.instance().getMode() == OperationMode.PAUSED &&
+                !plannedStmtBatch.isReadOnly() && !plannedStmtBatch.adminConnection) {
```
paulmartel: There's no need for such a casual test of readonly for LRR. If the statement (batch) is not read-only, we're in deep trouble here.

src/frontend/org/voltdb/planner/CorePlan.java
```diff
@@ -54,6 +54,9 @@
     /** Does the statement write? */
     public final boolean readOnly;
 
+    /** Is the statement high volume? */
+    public final boolean highVolume;
```
paulmartel: Can this be private?

src/frontend/org/voltdb/compiler/AdHocPlannedStmtBatch.java
```diff
@@ -101,6 +104,15 @@ public AdHocPlannedStmtBatch(
             }
         }
         this.readOnly = allReadOnly;
+
+        boolean highVolume = false;
```
paulmartel: To me, defining a local with the same name as a member is asking for trouble. 
To a lesser degree, so is not using the m_ prefix convention -- some of us appear to enjoy life on the edge.

src/frontend/org/voltdb/planner/QueryPlanner.java
```diff
@@ -63,6 +63,7 @@
     DeterminismMode m_detMode;
     PlanSelector m_planSelector;
     boolean m_isUpsert;
+    boolean m_isHighVolume;
```
paulmartel: can this be private?

paulmartel: can this be final?

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -59,7 +59,7 @@
 
     // Keep this flag turned off in production or when testing user-accessible EXPLAIN output or when
     // using EXPLAIN output to validate plans.
-    protected static boolean m_verboseExplainForDebugging = false; // CODE REVIEWER! KEEP false on master!
+    protected static boolean m_verboseExplainForDebugging = true; // CODE REVIEWER! KEEP false on master!
```
paulmartel: The comment says "keep false on master". If you were planning on keeping it true on your feature branch I guess that's OK. It MIGHT interfere with junits that validate explain plan output -- unless they explicitly turn it off?

src/frontend/org/voltdb/plannodes/SendPlanNode.java
```diff
@@ -29,6 +29,13 @@
 
 public class SendPlanNode extends AbstractPlanNode {
 
+    public enum Members {
+        HIGH_VOLUME;
+    }
```
paulmartel: This overengineered Members convention is actually on its way out. You're better off defining

    private final static String HIGH_VOLUME_MEMBER_NAME = "HIGH_VOLUME";

and using that in place of `Members.HIGH_VOLUME.name()`

src/frontend/org/voltdb/plannodes/SendPlanNode.java
```diff
@@ -75,21 +89,32 @@ public void computeCostEstimates(long childOutputTupleCountEstimate, Cluster clu
...
-            return "SEND PARTITION RESULTS TO COORDINATOR";
+        if (m_highVolume) {
+            assert(m_parents.size() == 0);
+            return "WRITE RESULTS TO DISK";
+        } else {
+            if (m_parents.size() == 0)
+                return "RETURN RESULTS TO STORED PROCEDURE";
+            else
+                return "SEND PARTITION RESULTS TO COORDINATOR";
+        }
```
paulmartel: Please no `else` after a `return`.

src/frontend/org/voltdb/sysprocs/ReadOnlySlow.java
```diff
@@ -0,0 +1,228 @@
...
+        // the final param is the byte array we need
+        byte[] serializedBatchData = (byte[]) params.getParam(params.size() - 1);
+
+        Pair<Object[], AdHocPlannedStatement[]> data = decodeSerializedBatchData(serializedBatchData);
+        Object[] userparams = data.getFirst();
+        AdHocPlannedStatement[] statements = data.getSecond();
+
+        StringBuilder sb = new StringBuilder();
+        if (statements.length == 0) {
+            sb.append("ONLYSLOW INVOCATION HAS NO SQL");
```
paulmartel: That's supposed to be READONLYSLOW?

src/frontend/org/voltdb/sysprocs/ReadOnlySlow.java
```diff
@@ -0,0 +1,228 @@
...
+import org.voltdb.VoltTable;
+import org.voltdb.common.Constants;
+import org.voltdb.compiler.AdHocPlannedStatement;
+import org.voltdb.compiler.AdHocPlannedStmtBatch;
+import org.voltdb.planner.ActivePlanRepository;
+
+import com.google_voltpatches.common.base.Charsets;
+
+
+public class ReadOnlySlow extends VoltSystemProcedure {
```
paulmartel: I don't see any critical differences between this class and its sibling class AdHocBase -- suggest inheriting from AdHocBase.  The last-minute check for non-read-only statements seems like it should be triggering earlier (like in the planner).

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1711,6 +1723,29 @@ public static File getVoltDbRoot(PathsType paths) {
         return voltDbRoot;
     }
 
+    public static File getHighVolumeOutput(PathsType.Highvolumeoutput paths, File voltDbRoot) {
+        File highVolumeOutputPath;
+        highVolumeOutputPath = new File(paths.getPath());;
+        if (!highVolumeOutputPath.isAbsolute())
+        {
```
paulmartel: Put open brace on same line.

paulmartel: To clarify, this is not an officially adopted VoltDB coding standard, it's another one of these I'm trying to foist on the CORE team. I've long promoted it for control statements. I've historically been more on the fence about the open brace for a class or method, but I've been lately shifting towards a policy of "total consistency".

tests/ee/storage/serialize_test.cpp
```diff
@@ -157,6 +158,68 @@ TEST_F(TableSerializeTest, RoundTrip) {
     delete deserialized;
 }
 
+TEST_F(TableSerializeTest, FileRoundTrip) {
+    // Serialize the table
+	ChTempDir tempdir;
```
paulmartel: Eliminate the tab character? I'm surprised. Don't these tests need to pass licensecheck?

tests/ee/storage/serialize_test.cpp
```diff
@@ -157,6 +158,68 @@ TEST_F(TableSerializeTest, RoundTrip) {
     delete deserialized;
 }
 
+TEST_F(TableSerializeTest, FileRoundTrip) {
```
paulmartel: These cloned tests could use some code reduction by refactoring (possibly templatizing) common code out of the tests that they were cloned from.

tests/frontend/org/voltdb/LRRHelper.java
```diff
@@ -0,0 +1,57 @@
...
+import org.voltdb.VoltTable;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+
+public class LRRHelper {
+    public static VoltTable getTableFromFileTable (VoltTable fileTable) {
```
paulmartel: Please, no space after method name in definitions/declarations.

tests/frontend/org/voltdb/ReadOnlySlowQueryTester.java
```diff
@@ -0,0 +1,167 @@
...
+
+import org.voltcore.utils.PortGenerator;
+import org.voltdb.VoltDB.Configuration;
+import org.voltdb.client.NoConnectionsException;
+import org.voltdb.client.ProcCallException;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.utils.MiscUtils;
+
+// Copied from AdHocQueryTester
+public abstract class ReadOnlySlowQueryTester extends TestCase {
```
paulmartel: Even though you may need to redefine runAllAdHocSPtests to cull the test queries, inheriting from AdHocQueryTester seems preferable to all the duplication here.

tests/frontend/org/voltdb/TestClientInterface.java
```diff
@@ -358,6 +358,18 @@ public void testAdHoc() throws IOException {
     }
 
     @Test
+    public void testReadOnlySlow() throws IOException {
```
paulmartel: Comparing this to testAdHoc, I see that you replicated the MP invocation test but not the SP invocation test.
That seems backwards to me. Why shouldn't SP work? Why should MP work?

tests/frontend/org/voltdb/TestReadOnlySlowQueries.java
```diff
@@ -0,0 +1,397 @@
...
+                    fail("Failed to shut down cluster");
+                }
+            }
+            m_cluster = null;
+
+            // no clue how helpful this is
+            System.gc();
+        }
+
+        boolean isValgrind() {
```
paulmartel: Are isValgrind and isMemcheckDefined are dead code in this class?

tests/frontend/org/voltdb/TestReadOnlySlowQueries.java
```diff
@@ -0,0 +1,397 @@
...
+import org.voltdb.client.ProcCallException;
+import org.voltdb.compiler.AsyncCompilerAgent;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.regressionsuites.LocalCluster;
+import org.voltdb.types.TimestampType;
+import org.voltdb.utils.MiscUtils;
+import org.voltdb.utils.VoltFile;
+
+// XXX Once ReadOnlySlow support MP queries, can extend from AdHocQueryTester again
+public class TestReadOnlySlowQueries extends ReadOnlySlowQueryTester {
```
paulmartel: IS the code duplication here with TestAdHocQueries justified because it'll be easier to solve when MP queries are supported or should we be looking now or a way to eliminate it like having a common base class?

tests/frontend/org/voltdb/TestTheHashinator.java
```diff
@@ -240,7 +240,7 @@ public void testExpectNonZeroHash() throws Exception {
                         0,
                         64*1024,
                         100,
-                        config, false);
+                        config, false,"");
```
paulmartel: Space after comma (comment also applies to a few other modules).

tests/frontend/org/voltdb/regressionsuites/TestLongRunningReadQuery.java
```diff
@@ -0,0 +1,130 @@
...
+        boolean success;
+
+        config = new LocalCluster("plansgroupby-onesite.jar", 1, 1, 0, BackendTarget.NATIVE_EE_JNI);
+        success = config.compile(project);
+        assertTrue(success);
+        builder.addServerConfig(config);
+
+        return builder;
+    }
+}
```
paulmartel: End all files with newlines.

tests/frontend/org/voltdb/regressionsuites/TestLongRunningReadQuery.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    //
+    // Suite builder boilerplate
+    //
+
+    public TestLongRunningReadQuery(String name) {
+        super(name);
+    }
+    static final Class<?>[] PROCEDURES = {
```
paulmartel: member not used.

tests/frontend/org/voltdb/regressionsuites/TestLongRunningReadQuery.java
```diff
@@ -0,0 +1,130 @@
...
+                ;
+        try {
+            project.addLiteralSchema(literalSchema);
+        } catch (IOException e) {
+            assertFalse(true);
+        }
+
+        boolean success;
+
+        config = new LocalCluster("plansgroupby-onesite.jar", 1, 1, 0, BackendTarget.NATIVE_EE_JNI);
```
paulmartel: Minor: inappropriate dummy name: plansgroupby.

tests/frontend/org/voltdb/regressionsuites/TestLongRunningReadQuery.java
```diff
@@ -0,0 +1,130 @@
...
+import org.voltdb.client.ProcCallException;
+import org.voltdb.compiler.AsyncCompilerAgent;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.jni.ExecutionEngine;
+
+public class TestLongRunningReadQuery extends RegressionSuite {
+
+    private static int tableSize = 1000000;
+
+    private void checkProcStatistics(Client client)
```
paulmartel: This function just seems to be cluttering the test output.
Is it really serving a purpose? Can it be modifed to do an actual validation rather than just dump output for a visual inspection that, practically speaking, will never happen.
In general, I LIKE keeping around PROVEN USEFUL visual diagnostic code so it can be quickly enabled when something goes wrong, BUT I like it DISABLED when all is running as expected for a less verbose successful run and the clearer identification of actual problem diagnostics.

paulmartel:
End of review round -- everything else looks good.

tests/frontend/org/voltdb/TestClientInterface.java
rharding6373: This test is only for the client interface to test whether the stored procedure is dispatched correctly. The checks for MP vs SP are later in the pipeline (in the planner), so both succeed here. Perhaps testing with a MP query is misleading, though...

src/ee/common/executorcontext.hpp
```diff
@@ -150,6 +151,30 @@ class ExecutorContext {
         return m_topEnd;
     }
 
+    std::string getHighVolumeDirPath() {
```
BillWhite: These are specialized to the file version, aren't they?  You might use std::enable_if and type traits of SerialOutputBuffer and SerialOutputFile to choose whether to define them.  But that may be more trouble than it's worth.

BillWhite:
I mostly skimmed these changes, but they seem to get at what I expected.  Looks good to me.

End of review round.

src/ee/common/declarations.h
```diff
@@ -78,6 +79,8 @@ class UndoLog;
 
 #ifndef SERIALIZE_IO_DECLARATIONS
 #define SERIALIZE_IO_DECLARATIONS
+template <class SO> class SerializeOutput;
+
```
paulmartel: fyi, declarations.h is mostly a crutch propping up superstitions and sloppiness.
It's true purpose is to make template INSTANTIATIONS available more globally -- though I question even that.
MOST of the forward declarations here should be moved to where they are specifically needed.
Many, including the ones you changed/added can safely be removed.

src/ee/common/serializeio.h
```diff
@@ -594,8 +498,240 @@ class CopySerializeOutput : public SerializeOutput {
...
+
+    template <typename T>
+    void writePrimitive(T value) {
+        ofile.write((char*)&value,sizeof(value));
+    }
+
+    template <typename T>
+    void writePrimitiveAt(T value) {
+        ofile.write((char*)&value,sizeof(value));
+    }
```
paulmartel: This function is apparently not called and not required even if its faulty signature was correct. Remove it?

src/ee/storage/persistenttable.cpp
```diff
@@ -1217,7 +1217,23 @@ std::string PersistentTable::debug() {
...
 void PersistentTable::processLoadedTuple(TableTuple &tuple,
-                                         ReferenceSerializeOutput *uniqueViolationOutput,
+                                         SerializeOutput<ReferenceSerializeOutput> *uniqueViolationOutput,
+                                         int32_t &serializedTupleCount,
+                                         size_t &tupleCountPosition,
+                                         bool shouldDRStreamRows) {
+    processLoadedTupleShared(tuple, uniqueViolationOutput, serializedTupleCount, tupleCountPosition, shouldDRStreamRows);
+}
+
+void PersistentTable::processLoadedTuple(TableTuple &tuple,
```
paulmartel: SerializeOutput can rely on SerializeOutputBuffer virtual functions to cover the differences between ReferenceSerializeOutput and FallbackSerializeOutput, etc. So, there is minimal benefit to instantiating SerializeOutput<FallbackSerializeOutput> instead of a generic SerializeOutput<SerializeOutput> -- as in VoltDBEngine.h et al.
Use of a more generic declaration avoids the problems caused by instantiations not being related by inheritance -- like the need here and in table.cpp to define explicit overloads and instantiatiations to satisfy the overly-specialized SerializeOutput declaration.
The only instantiations we need are for SerializeOutputBuffer and SerializeOutputFile and we only need the latter for specific functions motivated by your new code.

src/ee/common/NValue.hpp
```diff
@@ -312,7 +312,7 @@ class NValue {
     void deserializeFromAllocateForStorage(ValueType vt, SerializeInputBE& input, Pool* tempPool);
 
     /* Serialize this NValue to a SerializeOutput */
-    void serializeTo(SerializeOutput &output) const;
+    template<class T> void serializeTo(SerializeOutput<T> &output) const;
```
paulmartel: Is there a benefit to this vs. `template <class SO> void serializeTo(SO& output) const;`?
I suppose it's more explicit, but I think it's needlessly inflexible -- the simpler form is more open to application to minimally duck-typed SOs that need not be instantiations of the particular template. This comment applies to several signatures that got "generalized yet specialized" this way.

src/ee/common/serializeio.h
```diff
@@ -247,9 +247,9 @@ template <Endianess E> class SerializeInput {
 };
 
 /** Abstract class for writing to memory buffers. Subclasses may optionally support resizing. */
-class SerializeOutput {
+class SerializeOutputBuffer {
```
paulmartel: Minor problem that the name SerializeOutput -- though not the best of names -- at least used to match ReferenceSerializeOutput, CopySerializeOutput and have this analogous relationship with SerializeInput -- which the name SerializeOutputBuffer does not. Arguably what the SerializeOutput template provides is more type-specific serialization, so maybe SerializeOutput could keep its name and the template could be called TypedOutput?

src/ee/execution/VoltDBEngine.h
```diff
@@ -637,8 +648,10 @@ class __attribute__((visibility("default"))) VoltDBEngine {
 
 inline void VoltDBEngine::resetReusedResultOutputBuffer(const size_t headerSize)
 {
-    m_resultOutput.initializeWithPosition(m_reusedResultBuffer, m_reusedResultCapacity, headerSize);
-    m_exceptionOutput.initializeWithPosition(m_exceptionBuffer, m_exceptionBufferCapacity, headerSize);
+    m_resultOutputSerializer.initializeWithPosition(m_reusedResultBuffer, m_reusedResultCapacity, headerSize);
+    m_exceptionOutputSerializer.initializeWithPosition(m_exceptionBuffer, m_exceptionBufferCapacity, headerSize);
+    m_resultOutput.setSerializer(&m_resultOutputSerializer);
+    m_exceptionOutput.setSerializer(&m_exceptionOutputSerializer);
```
paulmartel: setSerializer doesn't change anything. remove it.

paulmartel:
End of review round.

PR #3720, opened 2016-06-23
ENG-10651: Reset DR applied tracker after command log replaying when r…

slowsoul:
…recovering an old replica cluster to a new normal cluster

**Problem and solution:**
if the cluster is recovering and has completed replaying command log, check if it is recovered as a producer cluster for active-passive DR. If it is, the DR applied trackers in sites should be reset and cleared as they are no longer valid.

Added a new type of task in ExecuteTask system procedure, RESET_DR_APPLIED_TRACKER, which will clear the DR applied trackers on every site.

### DIFF
--- a/src/frontend/org/voltdb/RestoreAgent.java
+++ b/src/frontend/org/voltdb/RestoreAgent.java
@@ -105,6 +105,8 @@ SnapshotCompletionInterest, Promotable
     // Current state of the restore agent
     private volatile State m_state = State.RESTORE;
 
+    private static final int MAX_RESET_DR_APPLIED_TRACKER_TIMEOUT_MILLIS = 30000;
+
     // Restore adapter needs a completion functor.
     // Runnable here preferable to exposing all of RestoreAgent to RestoreAdapater.
     private final Runnable m_changeStateFunctor = new Runnable() {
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -1294,6 +1296,24 @@ public void onReplayCompletion() {
...
+        if (m_isLeader && m_action.doesRecover()) {
+            VoltDBInterface instance = VoltDB.instance();
+            if (!instance.getCatalogContext().database.getIsactiveactivedred() &&
+                instance.getReplicationRole() == ReplicationRole.NONE) {
+                ByteBuffer params = ByteBuffer.allocate(4);
+                params.putInt(ExecutionEngine.TaskType.RESET_DR_APPLIED_TRACKER.ordinal());
+                try {
+                    instance.getClientInterface().callExecuteTask(30000, params.array());
+                } catch (IOException | InterruptedException e) {
+                    e.printStackTrace();
```
luyangco: Worth to put a warning here says reset DR applied tracker timeout

luyangco:
Review complete.

src/frontend/org/voltdb/RestoreAgent.java
slowsoul: Fixed

src/frontend/org/voltdb/jni/ExecutionEngine.java
```diff
@@ -63,7 +63,8 @@
         SET_DR_PROTOCOL_VERSION(3),
         SP_JAVA_GET_DRID_TRACKER(4),
         SET_DRID_TRACKER(5),
-        GENERATE_DR_EVENT(6);
+        GENERATE_DR_EVENT(6),
+        RESET_DR_APPLIED_TRACKER(7);
```
nshi: I think these are defined in the EE as well. They should be synced with each other even though some of them are only implemented in Java.

slowsoul: Yes indeed, I have made a pull request for this, https://github.com/VoltDB/voltdb/pull/3732

luyangco: Both branches were merged.

PR #3721, opened 2016-06-23
basic-upsert.sql: reduced the numbers added to the ID column (e.g.,

johnTcrawford:
changed '@idcol + 100' to '@idcol + 20'), which reduces the number of
rows created in each table (each UPSERT now does more updating and less
inserting), which solves the slow (PostgreSQL) query problem, so those
query templates are now uncommented. Also uncommented a couple of other
query templates, after tweaking them to generate fewer queries (e.g.,
changed '@cmp' to '@somecmp', and replaced '@columnpredicate' with a
clause using '@somecmp’). And changed 'SELECT *' queries to 'SELECT
@star' (which will make ENG-10464 easier).
sql_coverage_test.py: just fixed the comments for the
print_elapsed_seconds method.


wolffcm:
Looks good to me.

PR #3724, opened 2016-06-24
ENG-10639: Gracefully fail transaction on partitioning change.

nshi:
If the partitioning of a procedure is changed from MP to SP or vice
versa while there are in-flight transactions, these transactions should
be gracefully failed. Restarting them as mispartitioned invocations
enters an infinite loop of retry.

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -56,6 +56,7 @@ import org.voltdb.exceptions.EEException;
 import org.voltdb.exceptions.SerializableException;
 import org.voltdb.exceptions.SpecifiedException;
 import org.voltdb.groovy.GroovyScriptProcedureDelegate;
+import org.voltdb.iv2.MpInitiator;
 import org.voltdb.iv2.UniqueIdGenerator;
 import org.voltdb.messaging.FragmentTaskMessage;
 import org.voltdb.planner.ActivePlanRepository;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -469,6 +470,11 @@ public boolean checkPartition(TransactionState txnState, TheHashinator hashinato
                 return true;
             }
 
+            if (m_site.getCorrespondingPartitionId() == MpInitiator.MP_INIT_PID) {
+                // MP txn misrouted to SPI, possible to happen during catalog update
+                throw new ExpectedProcedureException("Single-partition procedure routed to multi-partition initiator");
```
wweiss-voltdb: The comment seems inconsistent with the error message

paulmartel:
This change looks correct, but it's strange that I don't see any other examples of ExpectedProcedureException being thrown in the current production code -- and yet there is a specific catch block for it ready made to turn these throws into GRACEFUL_FAILURE -- like the throwing of ExpectedProcedureException was a lost technique suddenly being revived in this change.

Ruth asked if there was any way to unit test this. Would it be possible to create a deterministic junit test that launched a thread to invoke a proc in a loop and then changed the partitioning of the proc?
It seems like it would be easy to observe that the loop was continuing to make progress finding the proc after the change -- possibly by checking its proc invocation stats? You'd expect to see invocation counts initially increasing on all partitions and then, after the DDL change, only on one partition (assuming that the invocations just used the same constant parameter value). But I don't know that we can GUARANTEE that one of the firehosed invocations will be caught in flight in a way that allows a deterministic test.

nshi:
@paulmartel Yeah, I was a bit surprised to find the catch block there without seeing any code constructing it except for the non-Volt backend. Maybe that was it.

I do have a unit test that consistently reproduces this without the patch in pro. You can see it here https://github.com/VoltDB/pro/pull/1403.

src/frontend/org/voltdb/ProcedureRunner.java
nshi: Good catch. I'll fix the comment.

PR #3725, opened 2016-06-24
Eng 10464 sqlcov upsert suites

johnTcrawford:
Added new SQLCoverage test suites: upsert-decimal, upsert-geo-point, upsert-geo-polygon, and upsert-timestamp,


src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -461,7 +474,32 @@ protected String replaceGroupNameVariables(String str, List<String> groupNames,
         List<String> columnValues = null;
         index = groupNames.indexOf("values");
         if (index > -1 && index < groupValues.size()) {
-            columnValues = new ArrayList<String>(Arrays.asList(groupValues.get(index).split(",")));
+            String columnValuesString = groupValues.get(index);
+            columnValues = new ArrayList<String>(Arrays.asList(columnValuesString.split(",")));
+
+            // Handle the case where one or more of the commas were enclosed in
```
wolffcm: This is the kind of thing where a real tokenizer/parser (like JavaCC or ANTLR) would simplify stuff like this a lot.

wolffcm:
I guess this incorporates changes from your other pull request?

Both look good to me.

PR #3726, opened 2016-06-24
ENG-10523: Recognize multi-partition binary log entries in the buffer splitter

slowsoul:
1. Add check for MP single hash txn so that if the txn is not for local partition, it will be skipped.
2. For DR binary log record of DR_RECORD_TRUNCATE_TABLE type, ignore the value of skipRows because the hash for this type of record is just a placeholder and meaningless. Table truncation should always be executed.
3. Adjust the behavior of getTestDRBuffer() for both normal and compatible DR tuple stream. A partition key value of -1 for a SINGLE_HASH type txn will now instruct this test method to generate this txn as a MP txn for test purpose. A partition key value of -1 for a MULTI_HASH type txn will now instruct this test method to generate this txn as a SP txn (part of a run-everywhere txns group) for test purpose.
4. Add some version constants for binary log versions from which DR starts to support different transactionality guarantee levels on the consumer cluster. (All SP & No MP, Uncoordinated MP and coordinated MP)
5. Fix a typo for CONSTRAINT_VIOLATION.

### DIFF
--- a/src/ee/storage/CompatibleDRTupleStream.cpp
+++ b/src/ee/storage/CompatibleDRTupleStream.cpp
@@ -488,7 +488,10 @@ int32_t CompatibleDRTupleStream::getTestDRBuffer(int32_t partitionId,
     // Override start sequence number
     stream.m_openSequenceNumber = startSequenceNumber - 1;
     for (int ii = 0; ii < flagList.size(); ii++) {
-        int64_t uid = UniqueId::makeIdFromComponents(ii * 5, 0, (flagList[ii] == TXN_PAR_HASH_MULTI || flagList[ii] == TXN_PAR_HASH_SPECIAL) ? 16383 : partitionId);
+        bool isMp = (flagList[ii] == TXN_PAR_HASH_MULTI && partitionKeyValueList[ii] != -1) ||
+                    flagList[ii] == TXN_PAR_HASH_SPECIAL ||
+                    (flagList[ii] == TXN_PAR_HASH_SINGLE && partitionKeyValueList[ii] == -1);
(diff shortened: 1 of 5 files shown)


src/ee/storage/CompatibleDRTupleStream.cpp
```diff
@@ -488,7 +488,10 @@ int32_t CompatibleDRTupleStream::getTestDRBuffer(int32_t partitionId,
     // Override start sequence number
     stream.m_openSequenceNumber = startSequenceNumber - 1;
     for (int ii = 0; ii < flagList.size(); ii++) {
-        int64_t uid = UniqueId::makeIdFromComponents(ii * 5, 0, (flagList[ii] == TXN_PAR_HASH_MULTI || flagList[ii] == TXN_PAR_HASH_SPECIAL) ? 16383 : partitionId);
+        bool isMp = flagList[ii] == TXN_PAR_HASH_MULTI ||
+                    flagList[ii] == TXN_PAR_HASH_SPECIAL ||
+                    (flagList[ii] == TXN_PAR_HASH_SINGLE && partitionKeyValueList[ii] == -1);
```
dongjunsun: Not obvious the intention of adding this scenario. The code seems to be only exercised by test code and most seem to only initialize the partitionKeyValueList in between 0..N

slowsoul: Actually it is used in TestDR2IdempotencyFilter to contruct an interesting edge case where a MP txn has a SINGLE flag. It is in https://github.com/VoltDB/pro/blob/485520d34fbd34d6bf13d962aa99b230e716eba2/tests/frontend/org/voltdb/dr2/TestDR2IdempotencyFilter.java#L485

PR #3727, opened 2016-06-27
ENG-10443: restore uses excessive heap gc time

dongjunsun:
Added the guard against no-op CompletionChecks

### DIFF
--- a/tests/frontend/org/voltdb/iv2/TestDurabilityListener.java
+++ b/tests/frontend/org/voltdb/iv2/TestDurabilityListener.java
@@ -26,15 +26,16 @@ package org.voltdb.iv2;
 import com.google_voltpatches.common.collect.Lists;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.voltdb.CommandLog.CompletionChecks;
+import org.voltdb.SnapshotCompletionMonitor;
 import org.voltdb.StoredProcedureInvocation;
 import org.voltdb.messaging.Iv2InitiateTaskMessage;
 
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.mock;
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
 
 public class TestDurabilityListener {
     private SpScheduler m_sched;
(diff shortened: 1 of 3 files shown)


nshi:
Looks okay to me. Pending Jenkins' approval.

@dongjunsun Can you also try a blocking snapshot save that takes a long time to finish to see if heap grows with your patch, please? Thanks.

dongjunsun:
Simulated a 5 minute delay inside SnapshotSaveAPI#Snapshotting, no heap growth is observed and heap dumps show no buildup of CompletionChecks objects.

nshi:
Cool. I imagine that it would run out of heap without your patch, is that correct?

dongjunsun:
Also tested without the fix with simulated long delay during snapshot save. Saw similar patterns of heap growth.

tests/frontend/org/voltdb/iv2/TestDurabilityListener.java
```diff
@@ -151,50 +152,110 @@ public void testNoDuplicateNotifications()
...
-    private List<Long> logSp(int...timestamps)
+    @Test
+    public void testProcessDurabilityChecks()
+    {
+        SiteTaskerQueue stq = mock(SiteTaskerQueue.class);
+        SnapshotCompletionMonitor scm = mock(SnapshotCompletionMonitor.class);
+        SpScheduler sched = spy(new SpScheduler(1, stq, scm));
+        sched.setLock(new Object());
+
+        TransactionTaskQueue taskQueue = mock(TransactionTaskQueue.class);
```
slowsoul: There's actually no need to recreate the taskQueue and listener as they are already created in setUp() as member variables. You can replace the private member dut by initializing it with the sched instance. This way you also don't have to change the signature of logSp, logMP and newInitMsg, which all uses the member variables.

tests/frontend/org/voltdb/iv2/TestDurabilityListener.java
```diff
@@ -26,15 +26,16 @@
...
 import org.voltdb.StoredProcedureInvocation;
 import org.voltdb.messaging.Iv2InitiateTaskMessage;
 
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.mockito.Mockito.mock;
+import static org.junit.Assert.*;
```
slowsoul: Intellij automatically condense normal/static imports from the same prefix into .* when there are several of them (can be configured in settings)
It is recommended to set the condense threshold to be a very high value, e.g. 100 to avoid these changes. You can set it in settings > editor > code style > Java > 'imports tab' > class count to use import with * / names count to use static import with *

PR #3728, opened 2016-06-27
ENG-9244: Partial snapshot missing replicated table name

yt14:
Fix SnapshotSaveAPI/startSnapshotting() to show table name when doing partial snapshot only for replicated table(s).

Reason for previous problem: the site who created the snapshot taskList may empty the result table because its own taskList is null. Actual snapshot work is done by other site.

### DIFF
--- a/src/frontend/org/voltdb/SnapshotSaveAPI.java
+++ b/src/frontend/org/voltdb/SnapshotSaveAPI.java
@@ -243,7 +243,10 @@ public class SnapshotSaveAPI
                             earlyResultTable.addRow(context.getHostId(), hostname,
                                     CoreUtils.getSiteIdFromHSId(context.getSiteId()), "SUCCESS", "");
                         } else {
-                            earlyResultTable = SnapshotUtil.constructNodeResultsTable();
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/SnapshotSaveAPI.java
```diff
@@ -243,7 +243,10 @@ else if (taskList == null) {
                             earlyResultTable.addRow(context.getHostId(), hostname,
                                     CoreUtils.getSiteIdFromHSId(context.getSiteId()), "SUCCESS", "");
                         } else {
-                            earlyResultTable = SnapshotUtil.constructNodeResultsTable();
+                            //If doing snapshot for only replicated table(s), earlyResultTable here
+                            //may not be empty even if the taskList of this site is null.
+                            //In that case, snapshot result is preserved by earlyResultTable.
+                            earlyResultTable = result;
```
nshi: Maybe we should consider doing the same thing to the assignment here https://github.com/VoltDB/voltdb/pull/3728/files#diff-495fb907cdc93322c12e76194931bbddR233

PR #3730, opened 2016-06-27
ENG-10481, change the covering host computing algorithm.

luyangco:
make debug output more readable.
Change-Id: I9dcdcbd26a0d9fa18437a6acddef23e5995d922b

### DIFF
--- a/src/frontend/org/voltdb/DRLogSegmentId.java
+++ b/src/frontend/org/voltdb/DRLogSegmentId.java
@@ -73,8 +73,29 @@ public class DRLogSegmentId implements Serializable {
         return makeDRIdFromComponents(clusterId, 0L) - 1L;
     }
 
+    /*
+     * Empty DR Id is used as a sentinel
+     * 1) in EndOfSnapshotBuffer to indicate the partition doesn't have any snapshot buffer,
+     * 2) in registerLastAckedSegmentId() when the partition has no tracker to recover.
+     *
+     * Empty DR Id has the highest significant bit been set, but there is one exception that
+     * if all bits are set then it belongs to cluster 0's Initial Ack DR Id, not Empty DR Id
+     * for cluster 255.
+     */
     public static boolean isEmptyDRId (long drId) {
-        return (drId >>> 63) == 1L;
+        return ((drId >>> 63) == 1L) && !(drId == -1);
+    }
+
+    /*
+     * Initial Ack DR Id is used as initial value for DR Idempotency filter
+     */
+    public static boolean isInitialAckDRId(long drId) {
+        if (drId == -1) return true;
+        int clusterId = getClusterIdFromDRId(drId);
+        if (clusterId >= 0 && clusterId < MAX_CLUSTER_ID) {
+            return ((drId >>> 63) != 1L) && (getSequenceNumberFromDRId(drId) == MAX_SEQUENCE_NUMBER);
+        }
+        return false;
     }
 
     public static boolean seqIsBeforeZero(long drId) {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -94,6 +94,12 @@ public static long getSentinelOrSeqNumFromDRId(long drId) {
     }
 
     public static String getDebugStringFromDRId(long drId) {
+        if (drId == Long.MAX_VALUE) {
+            return "QUEUE EMPTY";
+        }
+        if (drId < 0) {
+            return String.format("%d", drId);
```
nshi: It could be the empty DRID in this case, couldn't it? Aren't we still interested in the cluster ID even if it's empty?

slowsoul: Yes, I think we can add functions to check if a DR id is an empty DR id, an initial ack DR id, or a normal one, since they can be unambiguously differentiated. The isEmptyDRId() should be improved to check both the first bit and all sequence number bits equals 1, then if the cluster id bits are bigger than 127, it's not an empty DR id. These bits can only be all 1 (255), which is the initial ack DR id for cluster 0. We can also add a isInitialAckDRId() function to check. In this way we can have much clearer debug messages for DR ids.

luyangco: This is an utility function, since EMPTY DR ID is a valid input, it should be included. I'll make the change.

src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -73,8 +73,28 @@ public static long makeInitialAckDRId(int clusterId) {
...
+        return ((drId >>> 63) == 1L) && !(drId == -1);
+    }
+
+    /*
+     * Initial Ack DR Id is used as initial value for DR Idempotency filter
+     */
+    public static boolean isInitialAckDRId(long drId) {
+        if (drId == -1) return true;
+        int clusterId = getClusterIdFromDRId(drId);
+        if (clusterId >= 0 && clusterId < MAX_CLUSTER_ID) {
```
slowsoul: The second condition should be clusterId <= MAX_CLUSTER_ID

src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -73,8 +73,28 @@ public static long makeInitialAckDRId(int clusterId) {
...
+            return !((drId >>> 63) == 1L) && (getSequenceNumberFromDRId(drId) == MAX_SEQUENCE_NUMBER);
```
slowsoul: Maybe you can use !=, which is simpler

src/frontend/org/voltdb/DRLogSegmentId.java
```diff
@@ -73,8 +73,28 @@ public static long makeInitialAckDRId(int clusterId) {
...
+     * Empty DR Id is used as a sentinel
+     * 1) in EndOfSnapshotBuffer to indicate the partition doesn't have any snapshot buffer,
+     * 2) in registerLastAckedSegmentId() when the partition has no tracker to recover.
+     *
+     * Empty DR Id has the highest significant bit been set, but there is one exception that
+     * if all bits are one then it belongs cluster 0's Initial Ack DR Id, not Empty DR Id.
+     */
     public static boolean isEmptyDRId (long drId) {
-        return (drId >>> 63) == 1L;
+        return ((drId >>> 63) == 1L) && !(drId == -1);
```
slowsoul: Maybe you can use !=, which is simpler

PR #3731, opened 2016-06-27
Add EE support for PartitionByPlanNode.

BillWhite:
Add EE support for PartionByPlanNode.  The PartitionByPlanNode is a
trivial kind of AggregatePlanNode.  In the EE it's a trivial kind of
AggregateSerialPlanNode.  The only real difference between PartitionByPlanNode
and AggregateSerialPlanNode is that the PartitionByPlanNode knows that
it needs to output one row per input row, and not one per group.  The
algorithm p_execute_tuple knows this, and writes the correct number of
rows.

This supersedes the previous versions of PartitionByPlanNode in Java, and introduces
new EE support.

There is one EE plan execution tests, though more are needed.  This is in PartitionByExecutorTest, though it tests an entire plan.

Finally the facility for testing plans in the EE has been refactored quite a bit to make it easier to use.
There is a function which can be used to verify that output tables have the right values.


src/ee/executors/aggregateexecutor.cpp
```diff
@@ -505,7 +506,79 @@ class HyperLogLogsToCardAgg : public ApproxCountDistinctAgg {
...
+    // This is the peer count.  It counts the number of
+    // rows whose rank are equal to the last row.  When
+    // we see a row with a new order by expression value,
+    // we reset this to zero.
+    int64_t                            m_peerCount;
+    // This is the last value of the order by expression.  Since the
+    // default order by unit is RANGE, and we don't have any way to
+    // change that syntactically in HSQL, we only get one value for
+    // the order by.  So, we can save this in a single NValue.
+    NValue                             m_lastOrderByValue;
```
wolffcm: Historically we have had issues with storing NValues in Aggs due to how we handle inlined strings. See the longish comment in MaxAgg::advance for an explanation.  Consider if this applies here too.  Also, we should have tests for this!

src/ee/executors/aggregateexecutor.cpp
```diff
@@ -505,7 +506,79 @@ class HyperLogLogsToCardAgg : public ApproxCountDistinctAgg {
...
+    // We don't support dense rank now.  But we can easily.  So
+    // add support now.
+    bool                               m_isDenseRank;
```
wolffcm: If we are just going to suport dense_rank out of the gate because it's easy, then this comment seems unnecessary?

src/ee/executors/partitionbyexecutor.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+#ifndef SRC_EE_EXECUTORS_PARTITIONBYEXECUTOR_H_
+#define SRC_EE_EXECUTORS_PARTITIONBYEXECUTOR_H_
+
+#include "aggregateexecutor.h"
+
+namespace voltdb {
+
+/*
+ *
```
wolffcm: A Doxygen comment here would be good.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -709,10 +713,14 @@ private static AbstractExpression fromJSONObject(JSONObject obj,  StmtTableScan
      * The lists are cleared before they are filled in.  This is the inverse of toJSONArrayFromSortList.
      *
      * The JSONObject should be in object state, not array state.  It should have a member
-     * named SORT_COLUMNS, which is an array with the <expression, direction> pairs.
+     * named SORT_COLUMNS, which is an array with the <expression, direction> pairs.  The
+     * directions are optional.  However, if either every expression has a direction or
+     * no expression has a direction and the list sortDirections must be null.
```
wolffcm: I'm not following this comment.  If you have `ORDER BY x asc, y desc, z asc` then don't you need the direction list still?  Also, I think you meant to end with "...*then* the list of sortDirections must be null."

src/frontend/org/voltdb/expressions/WindowedExpression.java
```diff
@@ -157,19 +156,24 @@ protected void loadFromJSONObject(JSONObject jobj) throws JSONException {
...
-         * expressions will be serialized in the orderby node which preceeds
-         * the PartitionByPlanNode.
-         */
-        stringer.key(Members.PARTITION_BY_EXPRESSIONS.name()).array();
-        for (int idx = 0; idx < m_partitionByExpressions.size(); idx += 1) {
-            stringer.object();
-            m_partitionByExpressions.get(idx).toJSONString(stringer);
-            stringer.endObject();
+        // Be careful.  This node may have changed to be a
+        // non-windowed expression.  Don't serialize the partition by and sort
```
wolffcm: I'm confused by this comment.  Aren't you in the WindowedExpression class here?  How would this node have changed?

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -740,6 +762,10 @@ private AbstractExpression parseOperationExpression(VoltXMLElement exprNode) {
         else if (exprType == ExpressionType.OPERATOR_EXISTS) {
             expr = optimizeExistsExpression(expr);
         }
+        if (exprNode.attributes.containsKey("valuetype")) {
+            String valueType = exprNode.attributes.get("valuetype");
```
wolffcm: What is this for?  Seems unrelated to your changes?

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
```diff
@@ -287,18 +287,30 @@ public void resolveColumnIndexes()
         resolveColumnIndexesUsingSchema(input_schema);
     }
 
+    /**
+     * Gather up all the expressions whose column indices need to be
```
wolffcm: I like this approach and I wonder if it could b extended to other plan nodes as well.

I have a minor issue with the name of the method, since it's not so much that the expressions are resolvable, it's that they absolutely need to be resolved for the plan to be correct.  Maybe `getExpressionsNeedingResolution`?

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
```diff
@@ -287,18 +287,30 @@ public void resolveColumnIndexes()
...
-        for (SchemaColumn col : m_outputSchema.getColumns()) {
-            output_tves.addAll(ExpressionUtil.getTupleValueExpressions(col.getExpression()));
-        }
+        List<TupleValueExpression> output_tves = getResolvableExpressions();
         for (TupleValueExpression tve : output_tves) {
             int index = tve.resolveColumnIndexesUsingSchema(input_schema);
             if (index == -1) {
                 // check to see if this TVE is the aggregate output
-                // XXX SHOULD MODE THIS STRING TO A STATIC DEF SOMEWHERE
+                // XXX SHOULD MOVE THIS STRING TO A STATIC DEF SOMEWHERE
```
wolffcm: You now have this static def, so you could fix this and remove the comment?

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -26,210 +25,135 @@
...
+                                                            AbstractParsedStmt.TEMP_TABLE_NAME,
+                                                            AbstractParsedStmt.WINDOWED_AGGREGATE_COLUMN_NAME,
+                                                            AbstractParsedStmt.WINDOWED_AGGREGATE_COLUMN_NAME,
+                                                            0);
+        tve.setExpressionType(ExpressionType.VALUE_TUPLE);
+        tve.setValueType(m_windowedExpression.getValueType());
+        tve.setValueSize(m_windowedExpression.getValueSize());
+        // This doesn't really matter, since we will be
+        // generating this.  But it can't be negative.
+        tve.setColumnIndex(0);
```
wolffcm: I would have thought that this would be the windowed expression, instead of a TVE?  What am I missing?

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
```diff
@@ -26,210 +25,135 @@
...
+        // This could be the first order by expression.  We currently
+        // just support RANGE units, so there is only one order by expression.
+        // Furthermore, the RANK() operation does not have an argument, unlike,
+        // say, the MEAN(EXP) operation.  So, we pass the only order by
+        // expression in as the aggregate expression.  However, it seems
+        // like propagating this hack out of the EE seems wrong.  So, we
+        // don't add an expression here, and we will add order by expressions
+        // to the plan node, as if we were doing it right.  We will fix it
+        // up, which is to say, we will break it, in the EE.
+        m_aggregateExpressions.add(null);
```
wolffcm: So are you adding null to `m_aggregateExpressions` here as a placeholder for the order by expression you're passing to the RANK agg in the EE?  I don't get why you're appending null to this list here, what am I missing?

src/frontend/org/voltdb/types/ExpressionType.java
```diff
@@ -226,4 +226,24 @@ public boolean isAggregateExpression() {
...
+     * column's value.  This is true when the expression is an aggregate
+     * or a windowed aggregate.  We can't just make the windowed aggregate
+     * operations have class AggregateExpression.class because we will need
+     * the class to create WindowedExpression objects, and these have a
+     * different representation than other aggregate expression objects.
+     * For example, they have a PartitionBy list and an OrderBy list.
+     *
+     * @return true if an expression's value is generated by an early plan node.
+     */
+    public boolean isGeneratedAggregateExpression() {
```
wolffcm: I'm having trouble with this method name.  Maybe `isAggregateOrWindowedExpression`?

tests/ee/plannodes/PartitionByPlanNodeTest.cpp
```diff
@@ -89,39 +131,22 @@ class PartitionByPlanNodeTest : public Test {
     }
 };
 
+//
+// There is not much here to test.  The only difference between a
+// PartitionByPlanNode and any other Aggregate node is
```
wolffcm: Is...?

tests/ee/executors/PartitionByExecutorTest.cpp
```diff
@@ -0,0 +1,672 @@
...
+#include "storage/temptable.h"
+#include "storage/tableutil.h"
+#include "test_utils/plan_testing_baseclass.h"
+#include "test_utils/LoadTableFrom.hpp"
+
+
+namespace {
+const char *plan_strings[] = {
+    //  Plan for this query:
+    //      select ID, A, B, RANK() OVER ( PARTITION BY A ORDER BY B ) from AAA;
```
wolffcm: There's two plans here, what's the statement for the other one?

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -290,6 +297,36 @@ protected void checkQueriesPlansAreTheSame(String sql1, String sql2) {
         assertEquals(explainStr1, explainStr2);
     }
 
+    protected void verifyOrderByPlanNode(OrderByPlanNode  orderByPlanNode,
```
wolffcm: Would be  good to have a comment on how this method is used, since it may be useful outside of your new test.

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
```diff
@@ -0,0 +1,486 @@
...
+        }
+
+        // extra data
+        client.callProcedure("tu.insert", 60, 2);
+        client.callProcedure("tu.insert", 70, 3);
+        client.callProcedure("tu.insert", 80, 2);
+    }
+
+    public void notestNonSupportedCase() throws NoConnectionsException, IOException, ProcCallException {
+        System.out.println("STARTING xin......");
```
wolffcm: These are all Xin's tests from his IW branch?

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
```diff
@@ -0,0 +1,486 @@
+/* This file is part of VoltDB.
```
wolffcm: I see many licenses at the top of this file.

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
```diff
@@ -0,0 +1,486 @@
...
+                {2, 40, 2}, {2, 50, 3}, {3, 30, 1}, {3, 40, 2}});
+
+        vt = client.callProcedure("@AdHoc", "select b, a, rank() over (partition by b order by a desc) from tm order by b, a;").getResults()[0];
+        validateTableOfLongs(vt, new long[][]{{1, 10, 3},{1, 10, 3}, {1, 20, 2}, {1, 30, 1}, {2, 10, 3},
+                {2, 40, 2}, {2, 50, 1}, {3, 30, 2}, {3, 40, 1}});
+    }
+
+
+    public void notestNoRankScan_NON_UNIQUE() throws NoConnectionsException, IOException, ProcCallException {
+        System.out.println("STARTING xin......");
```
wolffcm: Seems like these tests are all disabled.  Are we able to run any of them yet?

wolffcm:
More tests are coming as part of ENG-10035 I guess?

That's all my review comments for this round.  In summary, I think the inline varchar issue in the Agg subclass in the EE could be serious.  Also, I'd like to understand the issue with passing order by exprs to the rank agg in the EE better.

src/ee/executors/aggregateexecutor.cpp
BillWhite: These can only be integer or timestamp.  So, I don't think they can be out of line.  When we support ROW units this will have to change.  But I think for now it's ok.

src/ee/executors/aggregateexecutor.cpp
BillWhite: I didn't want to do all the support right now.  I'd like to get this merged to master and then do the (minimal) changes for dense rank.

src/ee/executors/partitionbyexecutor.h
BillWhite: Done.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: No, we don't need the sort directions.  We need them in the OrderByExecutor for the windowed expression, but
that's earlier in the execution order from us.  By the time the data gets to us it's been sorted.  We need the order by expressions to calculate whether we are in the same order-by peer group.

src/frontend/org/voltdb/expressions/WindowedExpression.java
BillWhite: %%% 
This is a note to myself, not a comment.  I got to here and then I had to leave.

BillWhite: You are right.  This is a comment explaining a way I was doing this before.  I tried making the Windowed Expression look like a TVE.  I decided what I really wanted in the column was a TVE to begin with.  So, the Windowed Expression is never actually put into the plan.  It's cached in the AbstractParsedStmt.  I wanted to cache it in the ParsedSelectStmt, which seems logical.  But the method parseRankExpression is in the AbstractParsedStmt, and it's hard to move it from there.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
BillWhite: Right.  I was trying to propagate the HSQL from XML to the AbstractParsedStmt.  This is better done in finalizeType, which I now call in the right place.

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
BillWhite: Ok.  I'm easy.

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
BillWhite: Yes.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
BillWhite: It turns out that putting the windowed expression here causes problems.  This in the output schema, so this is describing the columns of the table we are writing with this partitionby plan node.  Consequently we just want to describe the generated column here.  We don't really care that it's a windowed expression.  We just want to know its type and size for the next plan node in the plan.

src/ee/executors/aggregateexecutor.cpp
wolffcm: In that case I suggest an assertion to ensure we don't forget.

src/frontend/org/voltdb/plannodes/PartitionByPlanNode.java
BillWhite: It turns out that the m_aggregateExpressions list and the m_aggregateTypes list are parallel.  I need to add something here.  It gets fixed up later on in the EE.  This is just to avoid making this look wrong in the planner.  The hack with expressions and orderby is entirely kept in the EE.

src/frontend/org/voltdb/types/ExpressionType.java
BillWhite: Actually, this bit of code is not necessary.  The WindowedExpression will never be in the display list or the output schema.  It will always be a TVE.

tests/ee/executors/PartitionByExecutorTest.cpp
BillWhite: I should put it here.  But this will all be replaced by an automagically generated test case anyway, so I'm inclined to leave it as it is.

tests/ee/plannodes/PartitionByPlanNodeTest.cpp
BillWhite: Fixed.  The PartitionByPlanNode knows how to generate one output row per input row.  Even this is not completely right.  We need to generate N output rows for each order by peer group of size N.  We don't in general know how to do that until we see the entire order by peer group.

tests/frontend/org/voltdb/planner/PlannerTestCase.java
BillWhite: Done.

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
BillWhite: I eliminated this file entirely.  It was mostly an artifact of Xin's implementation, which does not work for us at all.  In particular, there are a lot of rank expressions in where clauses and having clauses, which we don't have a prayer of supporting for quite a long time.

tests/frontend/org/voltdb/regressionsuites/TestAnalyticRank.java
BillWhite: Yes.  The ones that make sense were moved to TestWindowedAggregateSuite.

src/ee/executors/aggregateexecutor.cpp
BillWhite: Actually, I'm changing it to be an int64_t.  Then we have to figure it out eventually.

BillWhite:
Squash merged.

PR #3735, opened 2016-06-29
Improve determinism checking for set operators

wolffcm:
I looked at what we do in the EE and added logic in the planner to determine when the output of set operators are deterministic.  I also added tests for the new code I wrote.


tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -593,15 +593,36 @@ public void testMPDeterminismImpliedByParameter() {
...
+                + "union (select a, b, c from ttree_with_key) "
+                + "union (select a, b, c from ttree_with_key);",
+                false, true, DeterminismMode.FASTER);
+
+        // Both sides are deterministic; whole statement is deterministic
+        assertPlanDeterminismCore("(select a, b, c from ttree_with_key order by a, b, c) "
+                + "union (select a, b, c from ttree_with_key order by a, b, c limit 1);",
+                true, true, DeterminismMode.FASTER);
+
+        assertPlanDeterminismCore("select a from tonecolumn order by abs(a)",
```
paulmartel: This non-union test seems out of place -- turn it into a union.

tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -345,12 +345,72 @@ public void testUnionDeterminism() {
             assertEquals(false, isDeterministic);
         }
         {
+            // This is deterministic, but it's a limitation of our planner that we resolve the union's order by columns
+            // on the leftmost select statement.  It's difficult to check the RHS of the UNION, since it's possible
+            // column names are different there.
             CompiledPlan plan = compileAdHocPlan("(select B, DESC from T2 UNION select A, DESC from T1) order by B asc, DESC desc");
             boolean isDeterministic = plan.isOrderDeterministic();
+            assertEquals(false, isDeterministic);
```
paulmartel: `assertEquals(false, ` --> `assertFalse(`

paulmartel: So, bottom line, is this a regression you introduced? Are we trading off new right answers for a few new wrong answers -- that seems risky. AT A MINIMUM, new wrong answers should have a new ticket.

paulmartel: To be clear, I meant mischaracterization of determinism, not actual "wrong answer".

paulmartel: On further thought, I don't know why this is difficult. How hard is it to implement the special case: "when the ORDER BY on the union covers all the columns of the LHS, the whole union is deterministic"?

tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -345,12 +345,72 @@ public void testUnionDeterminism() {
...
+        assertFalse(plan.isOrderDeterministic());
+
+        // This is marked as deterministic even though it is, because the planner is not
+        // sophisticated enough to handle this case.
+        plan = compileAdHocPlan("(((select a from t1) union (select b from t2 order by b)) "
+                + "intersect select b from t2) order by a");
+        assertFalse(plan.isOrderDeterministic());
+
+        // intersect on LHS of union
+        // Not deterministic because LHS of union is not determinstic.
```
paulmartel: spelling: determinstic (a few places).

tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -345,12 +345,72 @@ public void testUnionDeterminism() {
...
+        plan = compileAdHocPlan("(select a from t1 order by a) intersect all select b from t2");
+
+        assertFalse(plan.isOrderDeterministic());
+        plan = compileAdHocPlan("(select a from t1 order by a ) except select b from t2");
+        assertFalse(plan.isOrderDeterministic());
+
+        plan = compileAdHocPlan("(select a from t1 order by a) except all select b from t2");
+        assertFalse(plan.isOrderDeterministic());
+
+        // A statement-level order by clause will the above statements deterministic.
```
paulmartel: "will" --> "will make"

tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -345,12 +345,72 @@ public void testUnionDeterminism() {
...
+        assertTrue(plan.isOrderDeterministic());
+
+        // More examples composing the various set operators.
+
+        // union on LHS of intersect
+        plan = compileAdHocPlan("((select a from t1 order by a) union (select b from t2 order by b)) "
+                + "intersect select b from t2");
+        assertFalse(plan.isOrderDeterministic());
+
+        // This is marked as deterministic even though it is, because the planner is not
```
paulmartel: "deterministic" --> "non-deterministic"

tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -345,12 +345,72 @@ public void testUnionDeterminism() {
...
         }
     }
 
+    public void testOtherSetOpDeterminism()
+    {
+        CompiledPlan plan;
+
+        // Output of non-union set ops is considered to be non-deterministic,
+        // since they use boost unordered containers in the EE.
+        // This is true even if sub-selects are sorted.
```
paulmartel: I'm confused -- I thought that we were going to distinguish SetOps that actually scanned a hash table from those that merely used it for membership testing.

tests/frontend/org/voltdb/planner/TestDeterminism.java
wolffcm: Fixed.

tests/frontend/org/voltdb/planner/TestUnion.java
wolffcm: Yes, this is effectively what the comment is saying.  UNION and UNION ALL use hash tables for membership testing only, and so their output can be deterministic if each input is deterministic.  But the other set ops, EXCEPT, INTERSECT and their "ALL" variants, scan hash tables to produce their output.

I hope that I've made this clear.  Let me know if you still have questions.  I'll try to update the comment to be clearer.

tests/frontend/org/voltdb/planner/TestUnion.java
wolffcm: In general determinism is tricky to get exactly right, and there will always be deterministic queries that will be marked as non-deterministic because the planner isn't smart enough.  Given this, we need to err on the side of being conservative, and consider that everything is non-deterministic unless the planner can prove otherwise---and often it can.

This change set is a significant improvement because many queries with set operations were being incorrectly marked as deterministic before:
- Unions where only the RHS is non-deterministic with no top level ORDER BY
- Non-unions with no top level ORDER BY

We get these right now with this fix.

The top level ORDER BY is very tricky to analyze because we resolve the ORDER BY column references against the leftmost query to determine the columns we're sorting by.  To really get this right we would need to:
- Resolve the order by expressions to the i-th, j-th, k-th columns of the LHS input
- Analyze the effect of the order by expressions on the RHS with the understanding that we're referring to the i-th, j-th, k-th expressions on its select list
Currently we only resolve the column offsets very late in the planning process, and that happens after the determinism check.

Here's an example of where we get into trouble with a top level ORDER BY:
```SQL
create table t (a integer primary key, b integer, c integer);

-- Statement 1 (not deterministic)
((select a, c from t) union (select c, a from t)) order by a;
```

In statement 1, it would be incorrect to test for determinism of the union's RHS using the expressions in the ORDER BY because we use names to figure out which columns are being referred to.  We would think that we are sorting by `a` when at runtime we will be sorting by `c`.

There are some deterministic queries that we got right before, but are now being marked as non-deterministic.  I will file a ticket for this.

For your last point, you are right that if we are sorting by all the columns of the LHS of the union, then we are also sorting by all the RHS columns too, and the statement must be deterministic.  I will try to make the planner smart enough to handle this case.

BillWhite: I don't understand how this select statement can ever work.  When the global and only sort happen, do we use the labels of the left hand side?  Is that standard behavior?

src/frontend/org/voltdb/planner/ParsedUnionStmt.java
```diff
@@ -166,8 +166,82 @@ void postParse(String sql, String joinOrder) {
...
+        else {
+            ParsedUnionStmt setOpStmt = (ParsedUnionStmt) stmt;
+            switch (setOpStmt.m_unionType) {
+            case EXCEPT:
+            case EXCEPT_ALL:
+            case INTERSECT:
+            case INTERSECT_ALL:
+                return orderIsDeterminedByOrderColumns(setOpStmt.m_children.get(0), orderColumns);
+            default:
+                return false;
```
BillWhite: Are we just giving up in this case?  Is it possible that a union can be deterministic by order columns?

Do we need to track the names of the order columns in the input tables and the names of the columns in the output tables as they are reassigned by unions?  I'm thinking about your example "(select a, c ...) union (select c, a...)..." case from below (or above).

BillWhite:
Sorry to be late with this.  This all seems ok with me.  I don't have any especial comments.

tests/frontend/org/voltdb/planner/TestUnion.java
wolffcm: Xin worked on this a while ago when he added support for aliases in the ORDER BY clause.  When the ORDER BY is on a set operator, we resolve the aliases based on the select list of the leftmost query.  (Note that the LHS of a set operator could be another set operator.)

My recollection is that the standard is vague on the correct behavior in this case, but this is how other vendors do it.

BillWhite:
Seems good to me.

src/frontend/org/voltdb/planner/ParsedUnionStmt.java
```diff
@@ -166,8 +166,92 @@ void postParse(String sql, String joinOrder) {
...
+
+        default:
+            return false;
+        }
+    }
+
+    private static boolean orderIsDeterminedByOrderColumns(AbstractParsedStmt stmt, List<ParsedColInfo> orderColumns) {
+
+        if (orderColumns.isEmpty()) {
+            return false;
```
paulmartel: We don't count as deterministic when the left child of an INTERSECT has its own deterministic ORDER BY and there is no top-level ORDER BY?

tests/frontend/org/voltdb/planner/TestDeterminism.java
paulmartel: Fixed? Doesn't look fixed.

src/frontend/org/voltdb/planner/ParsedUnionStmt.java
paulmartel: I just reviewed the implementation -- I am way off, here.
I need to go back over this -- after some sleep.

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -593,15 +593,36 @@ public void testMPDeterminismImpliedByParameter() {
...
-        assertPlanDeterminismCore("(select a, b, c from ttree_with_key) union (select a, b, c from ttree_with_key order by a, b, c limit 1);", ENG8790IsFixed, true, DeterminismMode.FASTER);
-        assertPlanDeterminismCore("select a from tonecolumn order by abs(a)", false, true, DeterminismMode.FASTER);
+
+        // LHS of union is not deterministic
+        assertPlanDeterminismCore("(select a, b, c from ttree_with_key) "
+                + "union (select a, b, c from ttree_with_key order by a, b, c limit 1);",
+                false, true, DeterminismMode.FASTER);
+
+        // LHS of union is not deterministic,
+        // but ORDER BY clause deterines order of whole statement.
```
paulmartel: spelling "deterines"

paulmartel:
This makes all kinds of sense now that I am awake enough to see it. Thanks for thinking this through.

PR #3736, opened 2016-06-30
FROM subquery optimization with complex display columns expressions 

alexeevm:


tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
```diff
@@ -3428,6 +3428,24 @@ public void testNPEbug() throws Exception {
...
 
+    public void testSubquerySimplification() throws Exception {
+        Client client = getClient();
+        String sql;
+
+        client.callProcedure("@AdHoc", "insert into R5 values (1,2,3)");
+        client.callProcedure("@AdHoc", "insert into R5 values (4,5,6)");
+
+        sql = "select * from (select C as D, D from R5) T;";
+        validateTableOfLongs(client, sql, new long[][] {{2, 3}, {5,6}});
```
wolffcm: This is exactly the test case that we were having issues with before, that I needed to solve by adding the differentiator.

To be certain that your optimization is working correctly, you could expand upon these test cases:

```SQL
-- multiply nested subqueries
select a from (select * from (select d as a, c, a as d from r5));

select * from (select a + c + d as acd, a * c * d as acd from R5);
```

wolffcm: Maybe there are other tests in that vein?  The basic idea is to try to trick the planner into producing an invalid plan.

I'm also curious about several levels of nested queries:
```SQL
select * from (select * from (select * from (select * from t)))
```
In this case, since we optimize from the most deeply nested up, will should be able to convert this into:
```SQL
select * from t;
```
Is that right?

wolffcm:
That's all my comments.

tests/frontend/org/voltdb/regressionsuites/TestSubQueriesSuite.java
alexeevm: I added the following cases to the planner test cases (TestPlansSubQuery.testSubquerySimplification):
select a from (select * from (select d as a, c, a as d from R1) T1) T2; == select T2.D A from R1 T2
select * from (select a + c + d as acd, a * c * d as acd from R1) T1 == select a + c + d as acd, a * c * d as acd from R1 T1
select * from (select * from (select * from R1 R1A) T1) T2 == select * from R1 T2
select * from (select * from (select * from R1 R1A) T1 LIMIT 10) T2 == select * from (select * from R1 T1 LIMIT 10)T2

alexeevm: But this one
select * from (select C AC, A AC from R5) T where AC = 15;
ant its counterpart without the optimization
select * from (select C AC, A AC from R5 LIMIT 10) T where AC = 15;
are apparently failing with the error message "user lacks privilege or object not found: AC". The error seams legitimate to me (which AC?) but the error message is misleading. Unfortunately, it's the HSQL error and  am not sure what could be done there. I have added these sals to the TestPlansSubQuery.testUnsupportedCases method

PR #3739, opened 2016-07-05
ENG-10715 VoltFile npe and ENG-10716 CSV import test flakes

paulmartel:


src/frontend/org/voltdb/utils/VoltFile.java
```diff
@@ -131,42 +131,44 @@ public static void moveSubRootContents(File fromSubRoot, File toSubRoot) throws
...
-                    }
+                moveSubRootContents(file, fInOtherSubroot);
+            }
+            else {
+                if (fInOtherSubroot.exists()) {
+                    throw new IOException(fInOtherSubroot + " already exists");
+                }
+                if (!fInOtherSubroot.createNewFile()) {
+                    throw new IOException();
+                }
```
BillWhite: Why no error message here?  Does the IOException usually contain all that is needed?  I don't know, I'm just asking.

BillWhite:
This all seems straightforward to me.  Does this sound like a description of your change:

Other than cleanups and reformatting, you added pushDataAsync and tryForResult to push data and fetch responses.  You also changed a lot of configuration code to calls to buildProperties.  Other than that nothing much has changed.

src/frontend/org/voltdb/utils/VoltFile.java
paulmartel: Good question. I probably should improve this.

PR #3741, opened 2016-07-06
ENG-10519: Generate binary log entry for multi-partition transaction

dongjunsun:
Per Jira ticket this change set makes EE generating empty BEGIN/END binary log entry for
multi-partition transaction even if there is no data changes on the partition.

* Check of begin of transaction is applied at AbstractExecutor instead of individual executors (insert/update/delete)
* DRTupleStream::transactionChecks called from existing code path remain unchanged and becomes no-op.
* BinaryLogSink is changed to allow handling of empty BEGIN/END entry at consumer side, currently is a no-op. Further handling requires coordinator changes.
* For testing "ant eecheck" all passed. Manually running multi-part transaction passes without error following BinaryLogSink changes.

### DIFF
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -19,6 +19,7 @@
 #include "common/debuglog.h"
 #include "executors/abstractexecutor.h"
 #include "storage/AbstractDRTupleStream.h"
+#include "storage/DRTupleStream.h"
 
 #include "boost/foreach.hpp"
 
@@ -250,6 +251,19 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
     int64_t oldSeqNum = m_drReplicatedStream->m_committedSequenceNumber;
     m_drReplicatedStream = drReplicatedStream;
     m_drReplicatedStream->setLastCommittedSequenceNumber(oldSeqNum);
+}
+
+void ExecutorContext::checkTransactionForDR() {
+    if (UniqueId::isMpUniqueId(m_uniqueId)) {
+        if (dynamic_cast<DRTupleStream *>(m_drStream)) {
+            m_drStream->transactionChecks(m_lastCommittedSpHandle, m_spHandle,
+                    m_uniqueId);
+            if (m_drReplicatedStream) {
+                m_drReplicatedStream->transactionChecks(m_lastCommittedSpHandle,
+                        m_spHandle, m_uniqueId);
+            }
+        }
+    }
 }
 
 } // end namespace voltdb
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -254,7 +254,7 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
 }
 
 void ExecutorContext::checkTransactionForDR() {
-    if (UniqueId::isMpUniqueId(m_uniqueId)) {
+    if (UniqueId::isMpUniqueId(m_uniqueId) && m_undoQuantum != NULL) {
         if (dynamic_cast<DRTupleStream *>(m_drStream)) {
             m_drStream->transactionChecks(m_lastCommittedSpHandle, m_spHandle,
                     m_uniqueId);
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -253,6 +253,14 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
     m_drReplicatedStream->setLastCommittedSequenceNumber(oldSeqNum);
 }
 
+/**
+ * To open DR stream to start binary logging for a transaction at this level,
+ *   1. It needs to be a multipartition transaction.
+ *   2. It is NOT a read-only transaction as it generates no data change on any partition.
+ *
+ * For single partition transactions, DR stream's binary logging is handled as is
+ * at persistenttable level.
+ */
 void ExecutorContext::checkTransactionForDR() {
     if (UniqueId::isMpUniqueId(m_uniqueId) && m_undoQuantum != NULL) {
         if (dynamic_cast<DRTupleStream *>(m_drStream)) {
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -20,6 +20,7 @@
 #include "executors/abstractexecutor.h"
 #include "storage/AbstractDRTupleStream.h"
 #include "storage/DRTupleStream.h"
+#include "storage/DRTupleStreamUndoAction.h"
 
 #include "boost/foreach.hpp"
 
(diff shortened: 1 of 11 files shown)


src/ee/executors/abstractexecutor.cpp
```diff
@@ -178,6 +179,32 @@ void AbstractExecutor::setDMLCountOutputTable(TempTableLimits* limits) {
     m_abstractNode->setOutputTable(m_tmpOutputTable);
 }
 
+/**
+ * Start binary log for transactions even if there is no data change.
+ */
+void AbstractExecutor::transactionChecksForDR() {
```
nshi: This doesn't look like the right place to put DR stream related stuff. An executor will only be executed when there is SQL (reads or writes) to run in a transaction. If there is nothing to run on this partition, no executor will be called, which defeats the purpose of the ticket.

I think you can directly generate the BEGIN record in VoltDBEngine through a JNI call when a transaction begins.

dongjunsun:
Moved opening DR stream transaction into VoltDBEngine's executePlanFragments. Also added a UT in DRBinaryLogTest.

Noticed that multi-partition transactions seem to be sent to all partitions even if a partition makes no data change within. Individual executors (insert/update/delete) have logic to filter them out.

src/ee/execution/VoltDBEngine.h
```diff
@@ -460,6 +460,8 @@ class __attribute__((visibility("default"))) VoltDBEngine {
         bool checkTempTableCleanup(ExecutorVector * execsForFrag);
         void resetExecutionMetadata();
 
+        void transactionChecksForDR();
```
nshi: Is this unused and not implemented?

src/ee/executors/abstractexecutor.cpp
```diff
@@ -51,6 +51,7 @@
 #include "plannodes/abstractscannode.h"
 #include "storage/tablefactory.h"
 #include "storage/TableCatalogDelegate.hpp"
+#include "storage/DRTupleStream.h"
```
nshi: Unused include.

src/ee/storage/persistenttable.h
```diff
@@ -206,6 +206,7 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
     friend class ::CompactionTest_BasicCompaction;
     friend class ::CompactionTest_CompactionWithCopyOnWrite;
     friend class CoveringCellIndexTest_TableCompaction;
+    friend class AbstractExecutor;
```
nshi: I believe this is also unnecessary now.

src/ee/storage/AbstractDRTupleStream.h
```diff
@@ -117,6 +119,9 @@ class AbstractDRTupleStream : public TupleStreamBase {
     int64_t m_rowTarget;
     bool m_opened;
     size_t m_txnRowCount;
+
+private:
+    virtual void transactionChecks(int64_t lastCommittedSpHandle, int64_t spHandle, int64_t uniqueId) = 0;
```
nshi: Or simply make this a public method.

src/ee/common/executorcontext.cpp
```diff
@@ -252,4 +253,17 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
     m_drReplicatedStream->setLastCommittedSequenceNumber(oldSeqNum);
 }
 
+void ExecutorContext::checkTransactionForDR() {
+    if (UniqueId::isMpUniqueId(m_uniqueId)) {
+        if (dynamic_cast<DRTupleStream *>(m_drStream)) {
```
nshi: Since the `transactionChecks()` is already in the base class, you can just call it directly without the dynamic cast and have a no-op implementation for the compatible mode.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -519,17 +519,21 @@ int64_t BinaryLogSink::applyTxn(ReferenceSerializeInputLE *taskInfo,
...
     type = static_cast<DRRecordType>(taskInfo->readByte());
-    do {
-        rowCount += apply(taskInfo, type, tables, pool, engine, remoteClusterId,
-                          txnStart, sequenceNumber, uniqueId, skipWrongHashRows);
-        type = static_cast<DRRecordType>(taskInfo->readByte());
-        if (type == DR_RECORD_HASH_DELIMITER) {
-            assert(isMultiHash);
-            partitionHash = taskInfo->readInt();
-            skipWrongHashRows = !engine->isLocalSite(partitionHash);
+    if (type != DR_RECORD_END_TXN) {
```
nshi: You can simply make the `do..while();` loop a `while()..` loop without adding the if guard.

src/ee/common/executorcontext.cpp
nshi: I see, the compatibility mode already has an implementation. I guess using the dynamic cast here is fine.

src/ee/common/executorcontext.hpp
```diff
@@ -87,6 +87,8 @@ class ExecutorContext {
         m_uniqueId = uniqueId;
         m_currentTxnTimestamp = (m_uniqueId >> 23) + VOLT_EPOCH_IN_MILLIS;
         m_currentDRTimestamp = createDRTimestampHiddenValue(static_cast<int64_t>(m_drClusterId), m_uniqueId);
+
+        checkTransactionForDR();
```
nshi: You have to guard against read-only transactions because those should not generate any DR data or increment the DR sequence number.

dongjunsun: Use undoToken to differentiate read-only transactions.

src/ee/storage/BinaryLogSink.cpp
dongjunsun: Changed but this needs to be changed again to actually count the empty begin/end pair when interfacing with coordinator.

src/ee/common/executorcontext.cpp
dongjunsun: no change here other than adding the guard against read-only transaction below.

src/ee/storage/persistenttable.h
dongjunsun: Removed.

src/ee/executors/abstractexecutor.cpp
dongjunsun: Removed. Wish Eclipse can tell me like Intellij did for unused import.

src/ee/execution/VoltDBEngine.h
dongjunsun: Removed. Compiler warns on unused variables but no methods and since it is not used linker is happy as well.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -519,21 +519,18 @@ int64_t BinaryLogSink::applyTxn(ReferenceSerializeInputLE *taskInfo,
     }
     // Read the whole txn since there is only one version number at the beginning
     type = static_cast<DRRecordType>(taskInfo->readByte());
-    if (type != DR_RECORD_END_TXN) {
-        do {
-            rowCount += apply(taskInfo, type, tables, pool, engine,
-                    remoteClusterId, txnStart, sequenceNumber, uniqueId,
-                    skipWrongHashRows);
+    while (type != DR_RECORD_END_TXN) { // FIXME empty BEGIN/END binary log entry for multi-partition transaction
```
nshi: What needs to be fixed here? The comment starts with "FIXME".

src/ee/common/executorcontext.cpp
```diff
@@ -254,7 +254,7 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
 }
 
 void ExecutorContext::checkTransactionForDR() {
-    if (UniqueId::isMpUniqueId(m_uniqueId)) {
+    if (UniqueId::isMpUniqueId(m_uniqueId) && m_undoQuantum != NULL) {
```
nshi: It's worth a comment here to explain the second condition.

src/ee/storage/BinaryLogSink.cpp
dongjunsun: Just reminder of pending integration work with coordinator. Not clear how it looks like now.

nshi: Not sure if this is the right place for that. It's so random and easy to be forgotten. The next person who comes along to edit this file may be very confused.

nshi: For reminder purpose, I would suggest writing a comment on the relevant ticket. https://issues.voltdb.com/browse/ENG-10524

PR #3744, opened 2016-07-06
Eng 10404 voltpaths

akhanzode:
DO NOT MERGE review only.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -298,6 +302,10 @@
     private final ListeningExecutorService m_es = CoreUtils.getCachedSingleThreadExecutor("StartAction ZK Watcher", 15000);
 
     private volatile boolean m_isRunning = false;
+    private boolean m_isLegacy = true;
```
nshi: What does this legacy flag mean?

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -99,32 +102,104 @@
     <xs:all>
         <xs:element name="voltdbroot" minOccurs="0" maxOccurs="1">
             <xs:complexType>
+            <xs:annotation>
+                <xs:appinfo>
+                    <ci:code>
+        <![CDATA[
+        public String getNodePath() {
+            return org.voltdb.VoltDB.instance().getPath(this, getKey());
+        }
```
nshi: Wow, really??? We are injecting Java code into the deployment file schema now???

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -381,6 +389,50 @@ public String getLicenseInformation() {
         return m_licenseInformation;
     }
 
+    @Override
+    public String getPath(Object base, String name) {
+        if (isLegacy()) {
+            try {
+                Method m = base.getClass().getMethod("getPath");
+                return (String )m.invoke(base);
```
nshi: Why is it doing a reflection to get a path???

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -406,20 +458,44 @@ private void managedPathsEmptyCheck() {
...
-        if ((path = managedPathEmptyCheck(voltDbRoot, paths.getCommandlog().getPath())) != null)
+        if ((path = managedPathEmptyCheck(voltDbRoot, paths.getCommandlog().getNodePath())) != null)
             nonEmptyPaths.add(path);
-        if ((path = managedPathEmptyCheck(voltDbRoot, paths.getCommandlogsnapshot().getPath())) != null)
+        if ((path = managedPathEmptyCheck(voltDbRoot, paths.getCommandlogsnapshot().getNodePath())) != null)
+            nonEmptyPaths.add(path);
+        return nonEmptyPaths.build();
+    }
+
+    private final List<String> managedPathsWithFilesPrimed(Configuration config, DeploymentType deployment) {
```
nshi: What's the difference between this method and the one above?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -406,20 +458,44 @@ private void managedPathsEmptyCheck() {
...
     }
 
-    private final List<String> managedPathsWithFiles(DeploymentType deployment) {
+    private final List<String> managedPathsWithFiles(Configuration config, DeploymentType deployment) {
         ImmutableList.Builder<String> nonEmptyPaths = ImmutableList.builder();
+        if (!config.m_isEnterprise) {
+            return nonEmptyPaths.build();
+        }
         PathsType paths = deployment.getPaths();
         String voltDbRoot = paths.getVoltdbroot().getPath();
```
nshi: Do you mean `paths.getVoltdbroot().getNodePath()`?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -498,6 +576,19 @@ public void initialize(Configuration config) {
...
                     return;
                 }
+                //Case where you give primed deployment with -d look in ../../ for initialized marker.
+                //Also check if parents are config and voltdbroot
+                File cfile = (new File(config.m_pathToDeployment)).getParentFile();
+                rootFH = cfile.getParentFile();
+                if (cfile.getName().equals("config") && rootFH.getName().equals(VoltDB.DBROOT)) {
+                    inzFH = new VoltFile(rootFH, VoltDB.INITIALIZED_MARKER);
+                    if (inzFH.exists()) {
+                        VoltDB.crashLocalVoltDB("cannot use legacy start action "
```
nshi: Capitalize the first letter.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -604,13 +695,18 @@ public void initialize(Configuration config) {
...
+                        managedPathsEmptyCheck(config);
+                }
 
-            if (!isRejoin && !m_joining) {
-                hostGroups = m_messenger.waitForGroupJoin(numberOfNodes);
+                if (!isRejoin && !m_joining) {
+                    hostGroups = m_messenger.waitForGroupJoin(numberOfNodes);
+                }
+            } catch (IOException ioe) {
+                //Crash
```
nshi: I think you have to actually call something to crash here.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: means you started with create/recover/rejoin verbs (old style) I will add comment.

nshi: I think it needs a new name. We have many legacy things, and putting `m_isLegacy` in the most comment place (RealVoltDB) is very confusing.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1650,6 +1798,28 @@ private void stageInitializedMarker(Configuration config) {
...
     }
 
+    private void stagePathConfiguration(Configuration config) {
+        File depFH = new VoltFile(config.m_voltdbRoot, VoltDB.INITIALIZED_PATHS);
+        try (PrintWriter pw = new PrintWriter(new FileWriter(depFH), true)) {
+            m_pathList.store(pw, "DO NOT MODIFY THIS FILE");
+            pw.flush();
+            pw.close();
+        } catch (IOException e) {
+            VoltDB.crashLocalVoltDB("Unable to stage configuration path designtion", false, e);
```
nshi: typo, "destination".

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1650,6 +1798,28 @@ private void stageInitializedMarker(Configuration config) {
...
+    }
+
+    private void loadPathConfiguration(Configuration config) {
+        File depFH = new VoltFile(config.m_voltdbRoot, VoltDB.INITIALIZED_PATHS);
+        try (FileInputStream is = new FileInputStream(depFH)) {
+            consoleLog.info("Loading Path configuration from " + depFH.getAbsolutePath());
+            m_pathList.load(is);
+            is.close();
+        } catch (IOException e) {
+            VoltDB.crashLocalVoltDB("Unable to read configuration path designtion", false, e);
```
nshi: Same here, typo.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1569,8 +1663,47 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
+            if (!path.isAbsolute()) {
+                path = new VoltFile(config.m_voltdbRoot, path.getPath());
+            }
+            m_pathList.put(dt.getPaths().getExportoverflow().getKey(), path.getCanonicalPath());
+
+            path = new VoltFile(dt.getPaths().getDroverflow().getPath());
+            if (!path.isAbsolute()) {
+                path = new VoltFile(config.m_voltdbRoot, path.getPath());
+            }
+            m_pathList.put(dt.getPaths().getDroverflow().getKey(), path.getCanonicalPath());
```
nshi: Is this whole block a duplicate of the `loadLegacyPathProperties()`?

src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -238,6 +239,8 @@ public void run() {
         public final Set<String> digestTables = new HashSet<String>();
         // Track the tables for which we found files on the node reporting this SnapshotInfo
         public final Set<String> fileTables = new HashSet<String>();
+        public final SnapthotPathType stype;
```
nshi: A more elaborate name could be picked here.

src/frontend/org/voltdb/DiskResourceChecker.java
```diff
@@ -194,21 +192,19 @@ private boolean isDiskAvailable(File filePath, int percThreshold, double sizeThr
...
 
-    private static File getPathForFeature(FeatureNameType featureName, PathsType pathsConfig)
+    private static File getPathForFeature(FeatureNameType featureName)
     {
-        // pathsConfig will never be null or something is really wrong
-        File voltDbRoot = CatalogUtil.getVoltDbRoot(pathsConfig).getAbsoluteFile();
         switch(featureName) {
         case COMMANDLOG :
-            return CatalogUtil.getCommandLog(pathsConfig.getCommandlog(), voltDbRoot);
+            return new File(VoltDB.instance().getCommandLogPath());
```
vtkstef: VoltFile?

src/frontend/org/voltdb/SnapshotDaemon.java
```diff
@@ -419,10 +419,12 @@ public void clientCallback(final ClientResponse clientResponse)
                     new TreeMap<Long, TruncationSnapshotAttempt>();
                 while (snapshots.advanceRow()) {
                     final String path = snapshots.getString("PATH");
+                    final String pathType = snapshots.getString("PATHTYPE");
```
nshi: Instead of sending another variable to snapshot save, can it leverage the URI template variable?

src/frontend/org/voltdb/SnapshotDaemon.java
```diff
@@ -581,7 +590,7 @@ public void run() {
                     loggingLog.error("Error during scan and group of truncation snapshots");
                 }
             }
-        }, 0, 1, TimeUnit.HOURS);
+        }, 0, 1, TimeUnit.MINUTES);
```
nshi: What's the motivation behind this change?

akhanzode: Oops testing need to undo.

src/frontend/org/voltdb/SnapshotDeleteAgent.java
```diff
@@ -172,14 +179,23 @@ protected void handleJSONMessage(JSONObject obj) throws Exception {
...
         new Thread("Async snapshot deletion thread") {
             @Override
             public void run() {
                 StringBuilder sb = new StringBuilder();
                 sb.append("Deleting files: ");
                 for (int ii = 0; ii < paths.length; ii++) {
-                    List<File> relevantFiles = retrieveRelevantFiles(paths[ii], nonces[ii]);
+                    String path = paths[ii];
+                    //When user calla @SnapshotDelete this will not be set to AUTO so we will delete what user requested.
+                    //If its SNAP_AUTO thenits coming from periodic delete task.
```
nshi: Many typos in this two line comment.

src/frontend/org/voltdb/SnapshotSaveAPI.java
```diff
@@ -495,7 +499,16 @@ else if (format == SnapshotFormat.INDEX) {
         else {
             throw new RuntimeException("BAD BAD BAD");
         }
-        final Callable<Boolean> deferredSetup = plan.createSetup(file_path, file_nonce, txnId,
+        SnapthotPathType stype = SnapthotPathType.valueOf(pathType);
+        if (stype == SnapthotPathType.SNAP_AUTO) {
+            file_path = VoltDB.instance().getSnapshotPath();
+            SNAP_LOG.info("Using local auto snapshot path: " + file_path);
```
nshi: Is INFO the right level for this? It might be too noisy.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: This is during the startup before properties are loaded in m_pathList for old syntax where paths are picked from deployment.

nshi: I still don't get why reflection is needed.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: refactoring to a method this one checks if its legacy.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: let me refactor these into one function.

nshi:
I think if we allow URI templates being specified in the deployment file paths, then there is no need to differentiate legacy and new. For legacy, the paths will just be absolute paths. For new, if you want to use different paths, the paths will be URI templates with variables to replace. Will that be simpler?

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: This is to avoid any places doing getPath() or getNodePath() based on if (isLegacy()) { getPath() } else { getNodePath() }

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: I think no one throws IOException so I will remove it...@stefano makes sense?

src/frontend/org/voltdb/DiskResourceChecker.java
akhanzode: Sure.

src/frontend/org/voltdb/SnapshotSaveAPI.java
akhanzode: I will remove these.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -493,11 +561,24 @@ public void initialize(Configuration config) {
...
                 if (inzFH.exists()) {
-                    VoltDB.crashLocalVoltDB("cannot use legacy start action "
+                    VoltDB.crashLocalVoltDB("Can not use legacy start action "
                             + config.m_startAction + " on voltdbroot "
                             + rootFH + " that was initialized with the init command");
                     return;
                 }
+                //Case where you give primed deployment with -d look in ../../ for initialized marker.
+                //Also check if parents are config and voltdbroot
+                File cfile = (new File(config.m_pathToDeployment)).getParentFile();
```
vtkstef:
```java
File cfile = (new VoltFile(config.m_pathToDeployment)).getParentFile();
```

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -493,11 +561,24 @@ public void initialize(Configuration config) {
...
+                rootFH = cfile.getParentFile();
+                if (cfile.getName().equals("config") && rootFH.getName().equals(VoltDB.DBROOT)) {
```
vtkstef:
```java
if ("config".equals(cfile.getName()) && VoltDB.DBROOT.equals(rootFH.getName())) {
```

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -901,24 +983,23 @@ public void initialize(Configuration config) {
...
 
-            boolean usingCommandLog = m_config.m_isEnterprise &&
-                    m_catalogContext.cluster.getLogconfig().get("log").getEnabled();
-            String clSnapshotPath = null;
-            if (m_catalogContext.cluster.getLogconfig().get("log").getEnabled()) {
-                clSnapshotPath = m_catalogContext.cluster.getLogconfig().get("log").getInternalsnapshotpath();
-            }
+            boolean usingCommandLog = m_config.m_isEnterprise
+                    && (m_catalogContext.cluster.getLogconfig() != null)
+                    && (m_catalogContext.cluster.getLogconfig().get("log") != null)
```
vtkstef: nice

src/catgen/spec.txt
```diff
@@ -6,9 +6,6 @@ begin Cluster                      "A set of connected hosts running one or more
   int httpdportno                  "The port number httpd will listen on. A 0 value implies 8080."
   bool jsonapi                     "Is the http/json interface enabled?"
   bool networkpartition            "Is network partition detection enabled?"
-  string voltRoot                  "Directory tree where snapshots, ppd snapshots, export data etc. will be output to"
```
vtkstef: nice

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -901,24 +983,23 @@ public void initialize(Configuration config) {
...
-                    m_producerDRGateway = (ProducerDRGateway) ndrgwConstructor.newInstance(new File(m_catalogContext.cluster.getDroverflow()),
-                                                                                   getSnapshotPath(m_catalogContext),
-                                                                                   m_replicationActive.get(),
-                                                                                   m_configuredNumberOfPartitions,
-                                                                                   m_catalogContext.getDeployment().getCluster().getHostcount());
+                    m_producerDRGateway =
+                            (ProducerDRGateway) ndrgwConstructor.newInstance(
+                                    new File(VoltDB.instance().getVoltDBRootPath()),
+                                    new File(VoltDB.instance().getSnapshotPath()),
+                                    m_replicationActive.get(),
```
vtkstef: new VoltFile

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1646,7 +1749,29 @@ private void stageInitializedMarker(Configuration config) {
...
-            VoltDB.crashLocalVoltDB("Unable to stage cluster name designtion", false, e);
+            VoltDB.crashLocalVoltDB("Unable to stage cluster name destination", false, e);
+        }
+    }
+
+    private void stagePathConfiguration(Configuration config) {
+        File depFH = new VoltFile(config.m_voltdbRoot, VoltDB.INITIALIZED_PATHS);
+        try (PrintWriter pw = new PrintWriter(new FileWriter(depFH), true)) {
+            m_pathList.store(pw, "DO NOT MODIFY THIS FILE");
+            pw.flush();
```
vtkstef: flush and close are done automagically when you get out from the try with resource block

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1646,7 +1749,29 @@ private void stageInitializedMarker(Configuration config) {
...
+            VoltDB.crashLocalVoltDB("Unable to stage configuration path destination", false, e);
+        }
+    }
+
+    private void loadPathConfiguration(Configuration config) {
+        File depFH = new VoltFile(config.m_voltdbRoot, VoltDB.INITIALIZED_PATHS);
+        try (FileInputStream is = new FileInputStream(depFH)) {
+            consoleLog.info("Loading Path configuration from " + depFH.getAbsolutePath());
+            m_pathList.load(is);
+            is.close();
```
vtkstef: flush and close are done automagically when you get out from the try with resource block

src/frontend/org/voltdb/CommandLog.java
```diff
@@ -28,6 +28,7 @@
 public interface CommandLog {
     /**
      *
+     * @param logSize
```
paulmartel: Better, but still short 2 @ params here. And 4 on the next function.

src/frontend/org/voltdb/RealVoltDB.java
paulmartel: This code would flow better if this flag were set about 140 lines further down nearer to the only(?) place it is being tested. The luxuries of a 700 line code pile.

src/frontend/org/voltdb/RealVoltDB.java
paulmartel: Should VoltFile be required by the constructor signature to catch this kind of oversight?

src/frontend/org/voltdb/SnapshotDeleteAgent.java
```diff
@@ -88,7 +89,7 @@ protected void collectStatsImpl(Connection c, long clientHandle, OpsSelector sel
     // return null.  Yes, ugly.  Bang it out, then refactor later.
     private String parseParams(ParameterSet params, JSONObject obj) throws Exception
     {
-        if (params.size() != 2) {
+        if (params.size() < 2) {
```
paulmartel: Message should say "either 2 or 3".

src/frontend/org/voltdb/SnapshotDeleteAgent.java
```diff
@@ -134,11 +135,17 @@ private String parseParams(ParameterSet params, JSONObject obj) throws Exception
         if (paths.length != nonces.length) {
             return "A path must be provided for every nonce";
         }
-
+        String stype = SnapthotPathType.SNAP_PATH.toString();
+        if (params.size() > 2) {
+            stype = (String )(ParameterConverter.tryToMakeCompatible(
+                        String.class,
+                        params.toArray()[2]));
```
paulmartel: Fun fact: `params.getParam(2)` indexes into a member array. `params.toArray()[2]` indexes into a short-lived clone of that member array. Yet, many -- possibly most -- system stored procs call toArray() to get their parameters -- many even calling toArray() once per parameter.

src/frontend/org/voltdb/SnapshotDeleteAgent.java
```diff
@@ -255,11 +266,12 @@ public boolean accept(File pathname) {
     }
 
     private VoltTable constructFragmentResultsTable() {
-        ColumnInfo[] result_columns = new ColumnInfo[9];
+        ColumnInfo[] result_columns = new ColumnInfo[10];
         int ii = 0;
         result_columns[ii++] = new ColumnInfo(VoltSystemProcedure.CNAME_HOST_ID, VoltSystemProcedure.CTYPE_ID);
         result_columns[ii++] = new ColumnInfo("HOSTNAME", VoltType.STRING);
         result_columns[ii++] = new ColumnInfo("PATH", VoltType.STRING);
+        result_columns[ii++] = new ColumnInfo("PATHTYPE", VoltType.STRING);
```
paulmartel: fyi, there is a convenience constructor for VoltTable that takes a variable number (one or more) ColumnInfo arguments instead of an explicit array. It makes these kinds of changes a little easier.

src/frontend/org/voltdb/SnapshotScanAgent.java
```diff
@@ -159,13 +160,14 @@ protected void dispatchFinalAggregations(PendingOpsRequest request)
             clientResults.add(new ClientResultRow(
                     (String)s.asRow()[0],
                     (String)s.asRow()[1],
-                    (long)s.asRow()[2],
+                    (String)s.asRow()[2],
```
paulmartel: Adding one more call to asRow . So how many identical calls to asRow does it take to justify factoring it out into one call and then operating on the resulting Object array?  Does it matter that asRow initializes a new array (the slow way) using calls to 4 other functions with loops and map lookups and string building?

src/frontend/org/voltdb/SnapshotScanAgent.java
```diff
@@ -269,11 +273,17 @@ private VoltTable getSnapshotScanResults(String path)
                             if (partitions.startsWith(",")) {
                                 partitions = partitions.substring(1);
                             }
-
+                            SnapthotPathType stype = SnapthotPathType.SNAP_PATH;
+                            if (f.getParent().equalsIgnoreCase(VoltDB.instance().getCommandLogSnapshotPath())) {
+                                stype = SnapthotPathType.SNAP_CL;
+                            } else if (f.getParent().equalsIgnoreCase(VoltDB.instance().getSnapshotPath())) {
```
paulmartel: Is IgnoreCase for file paths really "close enough" here? 'Cause millions of windows users can't be wrong?

src/frontend/org/voltdb/SnapshotScanAgent.java
```diff
@@ -486,6 +504,7 @@ private VoltTable getDiskFreeResults(String path)
     static class ClientResultRow implements Comparable<ClientResultRow>
```
paulmartel: Make static class ClientResultRow private.

src/frontend/org/voltdb/SnapshotScanAgent.java
```diff
@@ -789,9 +815,10 @@ private String complete() {
         }
 
         private Object[] asRow() {
-            Object row[] = new Object[9];
+            Object row[] = new Object[10];
```
paulmartel: Use `new Object[] {` and lose all this silly counting.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -675,9 +669,12 @@ public static String compileDeployment(Catalog catalog,
             }
 
             setCommandLogInfo( catalog, deployment.getCommandlog());
-
             setDrInfo(catalog, deployment.getDr(), deployment.getCluster());
+            //This is here so we can update our local list of paths.
+            //I would not have needed this if validateResourceMonitor didnt go through VoltDB
```
paulmartel: Did you mean validateResourceMonitorInfo?

tests/frontend/org/voltdb/TestMixedVersionClusters.java
```diff
@@ -35,6 +32,9 @@
 import org.voltdb.compiler.VoltProjectBuilder;
 import org.voltdb.regressionsuites.LocalCluster;
 import org.voltdb.utils.MiscUtils;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
```
paulmartel: Why are these imports moving? Did someone upgrade or reconfigure a dev tool. Is this signaling the start of a code-thrashing tools war where eclipse organizes imports  one way and the infinitely better tool that I refuse to adopt organizes imports a different way? Please no. please tell me your sentient dev tools have an "act dumb as eclipse" mode.

paulmartel:
That's the end of my unsolicited review round. I don't even remember what piqued my interest.

src/frontend/org/voltdb/CommandLog.java
akhanzode: Will add rest.

src/frontend/org/voltdb/SnapshotDeleteAgent.java
akhanzode: ok

src/frontend/org/voltdb/SnapshotDeleteAgent.java
akhanzode: sure make sense to do once.

src/frontend/org/voltdb/SnapshotScanAgent.java
akhanzode: refactoring...

src/frontend/org/voltdb/SnapshotScanAgent.java
akhanzode: Dont need ignoreCase will change.

src/frontend/org/voltdb/SnapshotScanAgent.java
akhanzode: Sure

src/frontend/org/voltdb/utils/CatalogUtil.java
akhanzode: yup will fix comment.

tests/frontend/org/voltdb/TestMixedVersionClusters.java
akhanzode: Yeah some global refactor-rename spooked my ide will revert to master as I dont intend to modify this.

akhanzode:
Thanks for looking at it.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: Sure I left things where they were but certainly looks good to keep together.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: I need both parent dirs.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: Its done via reflections so wont get caught. It used to take File originally as well. VoltFile is additional for voltFilePrefix for test after this and LocalCluster is refactored we will never need magic of VoltFile.

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: sure

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1899,6 +2010,49 @@ else if (sysType.getQuery().getTimeout() == 0) {
         }
     }
 
+    //Load paths into properties list.
+    private void loadPaths(DeploymentType deployment) throws IOException {
+        String voltdbRoot = (new VoltFile(deployment.getPaths().getVoltdbroot().getPath())).getCanonicalPath();
+        m_pathList.put(deployment.getPaths().getVoltdbroot().getKey(), voltdbRoot);
```
vtkstef: m_pathList is a good use case for using immutable map

src/frontend/org/voltdb/SnapshotDaemon.java
```diff
@@ -209,9 +211,8 @@
 
         // Register the snapshot status to the StatsAgent
         SnapshotStatus snapshotStatus = new SnapshotStatus();
-        snapshotStatus.setSnapshotPath(
-                catalogContext.cluster.getLogconfig().get("log").getInternalsnapshotpath(),
-                catalogContext.database.getSnapshotschedule().get("default").getPath());
+        snapshotStatus.setSnapshotPath(VoltDB.instance().getCommandLogSnapshotPath(),
```
vtkstef: wouldn't having the get*Path methods be better unser catalog context. It would decrease the need to use the VoltDB.instance() everywhere

src/frontend/org/voltdb/SnapshotDaemon.java
```diff
@@ -1544,7 +1553,7 @@ private void processScanResponse(ClientResponse response) {
                 final String nonce = snapshots.getString("NONCE");
                 if (nonce.startsWith(m_prefixAndSeparator)) {
                     final Long txnId = snapshots.getLong("TXNID");
-                    m_snapshots.add(new Snapshot(path, nonce, txnId));
+                    m_snapshots.add(new Snapshot(path, SnapthotPathType.SNAP_AUTO, nonce, txnId));
```
vtkstef: Can this bit can also be used to scan for command log snapshots too?

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -38,6 +39,7 @@
 {
     private static final VoltLogger compilerLog = new VoltLogger("COMPILER");
     private final LicenseApi m_licenseApi;
+    private static Pattern PATH_DELETER = Pattern.compile("(?i)\\<paths\\>(.*)\\<\\/paths\\>", Pattern.DOTALL|Pattern.MULTILINE);
```
vtkstef: instead of (?i) or it with Pattern.IGNORE_CASE .are you also removing comments too? because </path> could be embedded in a path

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -200,16 +202,21 @@ else if (work.invocationName.startsWith("@AdHoc")) {
...
-            retval.deploymentString = deploymentString;
+            //In non legacy mode discard the path element.
+            if (VoltDB.instance().isRunningWithOldVerbs()) {
+                retval.deploymentString = deploymentString;
+            } else {
+                retval.deploymentString = PATH_DELETER.matcher(deploymentString).replaceAll("");
+            }
             retval.deploymentHash =
-                CatalogUtil.makeDeploymentHash(deploymentString.getBytes(Constants.UTF8ENCODING));
+                CatalogUtil.makeDeploymentHash(retval.deploymentString.getBytes(Constants.UTF8ENCODING));
```
vtkstef: Use StandardEncoding.UTF_8

src/frontend/org/voltdb/messaging/SnapshotCheckResponseMessage.java
```diff
@@ -67,9 +75,12 @@ protected void initFromBuffer(ByteBuffer buf) throws IOException
     {
         m_path = new byte[buf.getInt()];
         buf.get(m_path);
+        m_stypeBytes = new byte[buf.getInt()];
+        buf.get(m_stypeBytes);
         m_nonce = new byte[buf.getInt()];
         buf.get(m_nonce);
         m_response = PrivateVoltTableFactory.createVoltTableFromSharedBuffer(buf);
+        m_stype = SnapthotPathType.valueOf(new String(m_stypeBytes, Charsets.UTF_8));
```
vtkstef: use StandardEncodings.UTF_8

src/frontend/org/voltdb/rejoin/Iv2RejoinCoordinator.java
```diff
@@ -290,7 +291,7 @@ private void onSiteInitialized(long HSId, long masterHSId, long dataSinkHSId,
         }
         if (data != null && !schemaHasNoTables) {
             REJOINLOG.debug("Snapshot request: " + data);
-            SnapshotUtil.requestSnapshot(0l, "", nonce, !m_liveRejoin, SnapshotFormat.STREAM, data,
+            SnapshotUtil.requestSnapshot(0l, "", nonce, !m_liveRejoin, SnapshotFormat.STREAM, SnapthotPathType.SNAP_CL, data,
```
vtkstef: Does SnapthotPathType.SNAP_CL really have a meaning in this snapshot request?

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -100,31 +100,37 @@
...
         <xs:element name="snapshots" minOccurs="0" maxOccurs="1">
             <xs:complexType>
               <xs:attribute name="path" type="xs:string" default="snapshots"/>
+              <xs:attribute name="key" type="xs:string" default="org.voltdb.path.snapshots" />
             </xs:complexType>
         </xs:element>
         <xs:element name="exportoverflow" minOccurs="0" maxOccurs="1">
             <xs:complexType>
               <xs:attribute name="path" type="xs:string" default="export_overflow"/>
+              <xs:attribute name="key" type="xs:string" default="org.voltdb.path.export_overflow" />
```
vtkstef: instead of default use fixed attribute for all key attribute definitions

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -704,7 +701,8 @@ private static void validateResourceMonitorInfo(DeploymentType deployment) {
     private static void setCommandLogInfo(Catalog catalog, CommandLogType commandlog) {
         int fsyncInterval = 200;
         int maxTxnsBeforeFsync = Integer.MAX_VALUE;
-        org.voltdb.catalog.CommandLog config = catalog.getClusters().get("cluster").getLogconfig().get("log");
+        org.voltdb.catalog.CommandLog config = catalog.getClusters().get("cluster").getLogconfig().add("log");
```
vtkstef: you are leaving this here because you want to make sure the nullified paths are the same across the cluster. Correct?

vtkstef:
I'd suggest you detach the piecemeal path knowledge from VoltDB.instance() and pass it around (eg. into the catalog context) as an ImmutableMap or Enum with path accessors

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
akhanzode: I am removing this from deployment.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
akhanzode: Yeah it picks up anything in between.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
akhanzode: I am keeping it what it was before.

src/frontend/org/voltdb/messaging/SnapshotCheckResponseMessage.java
akhanzode: Again keeping what everything else is using.

src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -65,6 +64,7 @@
 import org.voltdb.utils.MiscUtils;
 
 import com.google_voltpatches.common.collect.ImmutableSet;
+import org.voltdb.sysprocs.saverestore.SnapthotPathType;
 
 /**
```
wweiss-voltdb: SnapthotPathType?

wweiss-voltdb:
|`src/frontend/org/voltdb/SnapshotInitiationInfo.java:261`|
|---|
It would be nice if all the JSON keys were kept together in SnapshotUtil

wweiss-voltdb:
|`src/frontend/org/voltdb/SnapshotScanAgent.java:311`|
|---|
should this be equals?

src/frontend/org/voltdb/rejoin/Iv2RejoinCoordinator.java
wweiss-voltdb: I agree. There is no path for stream snapshot so using even CL does not make much sense, I would have a no-path type.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -213,6 +213,10 @@ else if (work.invocationName.startsWith("@AdHoc")) {
                 return retval;
             }
 
+            //In non legacy mode discard the path element.
+            if (!VoltDB.instance().isRunningWithOldVerbs()) {
+                dt.setPaths(null);
+            }
```
hhakimi53: If the paths element is updated, the follow-on logic still will be using the original deployment string based of original deployment where path elements are still populated. Follow-on logic to this will have to updated so that if paths elements are updated, it refreshes deployment string based on updated deployment type.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1566,28 +1668,34 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
-            if (!VoltDB.DBROOT.equals(deprootFN) && writeGeneratedDF) {
+            if (!VoltDB.DBROOT.equals(deprootFN)) {
                 consoleLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
                 hostLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
             }
             // if provided admin-mode start up is set, update it to false and update
             // the rewrite-flag reflect to rewrite the deployment instead of copy
             if(dt.getAdminMode().isAdminstartup()) {
                 dt.getAdminMode().setAdminstartup(false);
-               writeGeneratedDF = true;
```
hhakimi53: Suggestion: If we will be writing generated deployment based on dt instead of copying the original deployment file, we can unconditionally set adminStartup to false here.

src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -185,6 +185,7 @@ public void run() {
                                 " in " + m_snapshotToRestore.path);
```
wweiss-voltdb: Shouldn't this be PERSISTENT_SEQUENTIAL?

akhanzode: yeah will fix it.

akhanzode:
squash merged.

PR #3747, opened 2016-07-08
Eng 10645 sqlcov limit0

johnTcrawford:
Updated basic-select.sql, advanced-template.sql to include LIMIT 0 and OFFSET 0 as possible options, for various query templates.


wolffcm:
Looks good.  Sorry for the delay in reviewing this.

PR #3750, opened 2016-07-08
Deployment file pretty print

aquafemi:
Manually tested it by: 

* Starting a database with the new cli using 'start' without a deployment file and downloading a default deployment file through VMC (_which is a different default from the one in RealVoltDB but works the same_). 
* Making changes to the deployment through VMC to only some of the fields. 
* Catalog update through voltadmin and download new version on VMC. 

In all cases the file was indented and consistent with the change.


akhanzode:
Looks good...merge if jenkins is good.

PR #3751, opened 2016-07-10
Eliminate persistenttable_test flakiness (especially for -Dbuild=memc…

paulmartel:
…heck on mac) from over-dependence on iterator order determinism.

Also a makefile glitch in cpptest rebuild.

### DIFF
--- a/buildtools.py
+++ b/buildtools.py
@@ -504,7 +504,7 @@ def buildMakefile(CTX):
     makefile.write("\t$(AR) $(ARFLAGS) $@ $?\n")
     harness_source = TEST_PREFIX + "/harness.cpp"
     makefile.write("objects/harness.o: $(ROOTDIR)/" + harness_source + "\n")
-    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $^\n")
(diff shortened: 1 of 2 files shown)


buildtools.py
```diff
@@ -504,7 +504,7 @@ def buildMakefile(CTX):
     makefile.write("\t$(AR) $(ARFLAGS) $@ $?\n")
     harness_source = TEST_PREFIX + "/harness.cpp"
     makefile.write("objects/harness.o: $(ROOTDIR)/" + harness_source + "\n")
-    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $^\n")
+    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $(ROOTDIR)/" + harness_source + "\n")
```
xinjiacs: What's the difference between explicitly listing the prerequisites with $^ ? Or do you mean $< ?

I found this information here: http://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html#Automatic-Variables

xinjiacs:
Looks good to me.

wolffcm:
Looks good to me too.

PR #3753, opened 2016-07-11
ENG-10739 fix hang on two concurrent rejoins

vtkstef:


src/frontend/org/voltdb/VoltDB.java
```diff
@@ -642,6 +642,9 @@ else if (arg.equalsIgnoreCase("force")) {
                 m_startAction = StartAction.PROBE;
                 m_enableAdd = true;
                 checkInitializationMarker();
+                if (m_meshBrokers == null || m_meshBrokers.trim().isEmpty()) {
+                    m_meshBrokers = m_leader;
```
akhanzode: Do you need to check if m_leader is not empty like you check for legacy option?

src/frontend/org/voltdb/probe/MeshProber.java
```diff
@@ -386,6 +386,15 @@ public JSONObject decorate(JSONObject jo, Optional<Boolean> paused) {
                 return new JoinAcceptor.PleaDecision(error, false, false);
             }
             return new JoinAcceptor.PleaDecision(null, true, false);
+        } else {
+            StartAction operationalStartAction = hostCriteria.values().stream()
+                    .filter(c->c.getNodeState().operational())
+                    .map(c->c.getStartAction())
+                    .findFirst().orElse(getStartAction());
+            if (operationalStartAction == StartAction.PROBE && hc.getStartAction() != StartAction.PROBE) {
```
akhanzode: Do you need to check if cluster is full or not here?

tests/frontend/org/voltcore/messaging/TestHostMessenger.java
```diff
@@ -740,6 +740,66 @@ public void testProbedRejoinOnWholeCluster() throws Exception {
     }
 
     @Test
+    public void testProbedRejoinWithMismatchedStartAction() throws Exception {
```
akhanzode: The name does not suggests two concurrent rejoins where second is going to wait. Is that covered in different test?

src/frontend/org/voltdb/VoltDB.java
vtkstef: no because null leader will cause an error. empty leader means all host's interfaces

src/frontend/org/voltdb/probe/MeshProber.java
vtkstef: considerMeshPlea is run only one host pleas for membership. i.e. host a wants to connect to mesh where host b is. b considers host a plea.

tests/frontend/org/voltcore/messaging/TestHostMessenger.java
vtkstef: yes testTwoProbedConcuncurrentRejoins

PR #3755, opened 2016-07-18
Add consumer cluster id to DRProducer stats

aquafemi:
Add a new column to the node level and partition level table that adds an entry for the consumer cluster id that the producer is connected to. The id is the integer id of the consumer specified in the deployment.

Andrew said that in the future it would be nice if the rows were in order by the consumer cluster id.

### DIFF
--- a/src/frontend/org/voltdb/DRProducerStatsBase.java
+++ b/src/frontend/org/voltdb/DRProducerStatsBase.java
@@ -28,7 +28,7 @@ public class DRProducerStatsBase {
 
     public static interface Columns {
         // column for both tables
-        public static final String CONSUMER_CLUSTER_ID = "CLUSTER_ID";
+        public static final String CONSUMER_CLUSTER_ID = "CONSUMER_CLUSTER_ID";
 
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/DRProducerStatsBase.java
```diff
@@ -56,6 +59,7 @@ public DRProducerNodeStatsBase() {
         @Override
         protected void populateColumnSchema(ArrayList<VoltTable.ColumnInfo> columns) {
             super.populateColumnSchema(columns);
+            columns.add(new ColumnInfo(Columns.CONSUMER_CLUSTER_ID, VoltType.INTEGER));
```
nshi: Small int is probably enough for this field. Please also check with Andrew to see if having it as the first column is okay for backward compatibility.

aquafemi:
Some examples:

idexample1
idexample2
idexample3

aquafemi:
Wondering if there's a way to set the new system property in the tests?

PR #3757, opened 2016-07-18
basic-select.sql: include tests of OFFSET without LIMIT, now that

johnTcrawford:
ENG-10733 seems to be fixed (for materialized views).


BillWhite:
Seems right to me, not that I know anything.

PR #3758, opened 2016-07-18
ENG-10770: Use new temp table method for deleting tuples

wolffcm:
The FOJ changes made some changes to temp table methods for deleting tuples (and the non-inlined strings stored in them).  This resulted in an EE unit test showing a memory leak.  This was a test-only issue.

### DIFF
--- a/tests/ee/storage/serialize_test.cpp
+++ b/tests/ee/storage/serialize_test.cpp
@@ -108,14 +108,14 @@ class TableSerializeTest : public Test {
             }
 
         }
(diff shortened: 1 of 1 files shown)


BillWhite:
According to my understanding, the new method, deleteAllTempTupleDeepCopies, deletes all the outlined data, and then calls the usual deleteAllTempTuples to delete the data itself.  If that's right, then this seems right to me.

wolffcm:
Merged on separate branch.

PR #3760, opened 2016-07-19
ENG-10649 Task submission check during shutdown

zeemanhe:


akhanzode:

PR #3763, opened 2016-07-19
Fix compilation errors on Ubuntu 16.04.

BillWhite:
The GCC version 5.4.0 compiler does not have a global definition of
isnan.  This function is in std::isnan and not global.  The C6, C7
and U14.04 compilers had these functions in the global namespace and
not in std::isnan.  So, we leverage some partial support for this in
the S2 library and expand coverage to other data types which need it.


wolffcm:
This looks good to me.  Go ahead and merge when you're ready.

hhakimi53:
Looks good.

simonzhangsm:
yes, I have found this problem and have fix it in my code! Please merge it and do tests

PR #3767, opened 2016-07-20
Added a new valgrind suppression.

BillWhite:
For Ubuntu 16.04 the old STL suppression does not work.  The library
version numbers are incorrect.  I also added a diagnostic to build.py
to show how many threads are used when making the EE and the EE tests.

### DIFF
--- a/tests/ee/test_utils/vdbsuppressions.supp
+++ b/tests/ee/test_utils/vdbsuppressions.supp
@@ -9,3 +9,13 @@
    obj:/lib/x86_64-linux-gnu/ld-2.19.so
 }
 
+{
+   stl:u16.04
+   Memcheck:Leak
+   fun:malloc
+   obj:/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.21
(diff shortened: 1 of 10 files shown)


tests/ee/test_utils/vdbsuppressions.supp
```diff
@@ -9,3 +9,13 @@
    obj:/lib/x86_64-linux-gnu/ld-2.19.so
 }
 
+{
+   stl:u16.04
```
hhakimi53: Will this get applied to all stl containers or it's just the name for the suppression? Was the earlier suppression specific to vectors? Sorry, I don't remember the syntax for it.

Changes looks fine and hopefully with this false leaks reported will go away.

BillWhite: This will be applied just to the 16.04 leaks.  The leaks in the older compiler, the 5.2.X version I think, are covered by the first suppression.

BillWhite:
Squashing this on a separate branch.

PR #3768, opened 2016-07-21
ignore being admin-mode being set through deployment when using new verbs 

hhakimi53:
Logic updates to ignore admin mode setting via deployment file during voldtb init if different than default.  

Changes have been tested for the following scenario:
- User provides no deployment, the default deployment gets used same as today
- User provides a valid deployment without admin mode settings, deployment file gets copied to voldbroot if voldbroot is not changing
- User provides a valid deployment with admin mode startup set to false, the same deployment file gets copied to voldbroot if voldbroot is not changing
- User provides a valid deployment with admin mode set or different admin port , newly generated deployment file with other default setting gets written to voldbroot similar to case if the voltdbroot directory was different.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -1541,7 +1541,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
         String deprootFN = dt.getPaths().getVoltdbroot().getPath();
         File   deprootFH = new VoltFile(deprootFN);
         File   cnfrootFH = config.m_voltdbRoot;
-        boolean differingRoots = false;
+        boolean writeGeneratedDF = false;
 
         if (!cnfrootFH.exists() && !cnfrootFH.mkdirs()) {
             VoltDB.crashLocalVoltDB("Unable to create the voltdbroot directory in " + cnfrootFH, false, null);
@@ -1555,13 +1555,18 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
             }
             File cnfcanoFH = cnfrootFH.getCanonicalFile();
             if (!cnfcanoFH.equals(depcanoFH)) {
-                differingRoots = true;
+                writeGeneratedDF = true;
                 dt.getPaths().getVoltdbroot().setPath(cnfrootFH.getPath());
             }
             // root in deployment conflicts with command line voltdbroot
-            if (!VoltDB.DBROOT.equals(deprootFN) && differingRoots) {
+            if (!VoltDB.DBROOT.equals(deprootFN) && writeGeneratedDF) {
                 consoleLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
                 hostLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
+            }
+            // if provided admin-mode settings in deployment are different than the default one, update
+            // admin mode settings to default and update flag to commit updated deployment
+            if(CatalogUtil.updateAdminModeToDefaultIfNotDefault(dt)) {
+               writeGeneratedDF = true;
             }
         } catch (IOException e) {
             VoltDB.crashLocalVoltDB(
@@ -1617,8 +1622,14 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
                 return;
             }
         }
+
+        // log message unconditionally indicating that the provided host-count and admin-mode settings in
+        // deployment, if any, will be ignored
+        consoleLog.info("Hostcount, admin-mode settings specified in the deployment file will be ignored");
+        hostLog.info("Hostcount, admin-mode settings specified in the deployment file will be ignored");
+
         // see if you just can simply copy the deployment file
-        if (differingRoots) {
+        if (writeGeneratedDF) {
             File depFH = getConfigLogDeployment(config);
             try (FileWriter fw = new FileWriter(depFH)) {
                 fw.write(CatalogUtil.getDeployment(dt, true /* pretty print indent */));
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -1637,8 +1637,10 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
 
         // log message unconditionally indicating that the provided host-count and admin-mode settings in
         // deployment, if any, will be ignored
-        consoleLog.info("Hostcount, admin-mode settings specified in the deployment file will be ignored");
-        hostLog.info("Hostcount, admin-mode settings specified in the deployment file will be ignored");
+        consoleLog.info("When using the INIT command, some deployment file settings (hostcount, voltdbroot path, "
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1617,8 +1622,14 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
                 return;
             }
         }
+
+        // log message unconditionally indicating that the provided host-count and admin-mode settings in
+        // deployment, if any, will be ignored
+        consoleLog.info("Hostcount, admin-mode settings specified in the deployment file will be ignored");
```
akhanzode: Can you pass this messages by Andrew?

hhakimi53: Sure, thanks

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1555,14 +1555,19 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
             }
             // root in deployment conflicts with command line voltdbroot
-            if (!VoltDB.DBROOT.equals(deprootFN) && differingRoots) {
+            if (!VoltDB.DBROOT.equals(deprootFN) && writeGeneratedDF) {
                 consoleLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
                 hostLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
             }
+            // if provided admin-mode settings in deployment are different than the default one, update
+            // admin mode settings to default and update flag to commit updated deployment
+            if(CatalogUtil.updateAdminModeToDefaultIfNotDefault(dt)) {
```
vtkstef: a bit of an overkill to give it its own method 

```java
dt.getAdminMode().setAdminstartup(false);
```

hhakimi53: If we check only adminstartup attribute and if port does not matter for the case when adminstartup is false, than we can simply check and set it that way. I can tweak the logic to do it that way and eliminate the new function, if that's valid. Thanks

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -298,6 +298,10 @@
     private final ListeningExecutorService m_es = CoreUtils.getCachedSingleThreadExecutor("StartAction ZK Watcher", 15000);
 
     private volatile boolean m_isRunning = false;
+    private boolean m_isRunningWithOldVerb = true;
+
+    @Override
+    public boolean isRunningWithOldVerbs() { return m_isRunningWithOldVerb; };
```
vtkstef:
```java
public boolean isRunningWithOldVerbs() {
    return m_isRunningWithOldVerb; 
}
```

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1555,14 +1560,22 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
+            if (!VoltDB.DBROOT.equals(deprootFN) && writeGeneratedDF) {
                 consoleLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
                 hostLog.info("Ignoring voltdbroot \"" + deprootFN + "\"specified in the deployment file");
             }
+            // if provided admin-mode start up is different than the default one, update amin-mode
+            // startup to null for default, which is admin-mode set to false in deployment, and
+            // flag updated to reflect rewrite the deployment instead of copy
+            if(dt.getAdminMode().isAdminstartup()) {
+                dt.getAdminMode().setAdminstartup(null);
+                dt.getAdminMode().setPort(null);
```
vtkstef: leave the admin port alone please

vtkstef: also what is wrong with `dt.getAdminMode().setAdminstartup(false);` it is clearer and more readable than `dt.getAdminMode().setAdminstartup(null);`

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -200,16 +201,32 @@ else if (work.invocationName.startsWith("@AdHoc")) {
                 }
             }
 
-            result =
-                CatalogUtil.compileDeploymentString(newCatalog, deploymentString, false);
+            DeploymentType dt  = CatalogUtil.parseDeploymentFromString(deploymentString);
+            if (dt == null) {
+                retval.errorMsg = "Unable to update deployment configuration: Error parsing deployment string";
+                return retval;
```
vtkstef: nice

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -200,16 +201,32 @@ else if (work.invocationName.startsWith("@AdHoc")) {
...
                 retval.errorMsg = "Unable to update deployment configuration: " + result;
                 return retval;
             }
 
-            retval.deploymentString = deploymentString;
+            // ignore admin mode settings if operating with new start actions/verbs
+            if (!VoltDB.instance().isRunningWithOldVerbs() && dt.getAdminMode().isAdminstartup()) {
+                // set it to null, this will default to admin-mode startup to default which is false
+                dt.getAdminMode().setAdminstartup(null);
+                dt.getAdminMode().setPort(null);
```
vtkstef: leave the port alone, and use false instead of null

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
```diff
@@ -200,16 +201,32 @@ else if (work.invocationName.startsWith("@AdHoc")) {
...
+
+            result = CatalogUtil.compileDeployment(newCatalog, dt, false);
             if (result != null) {
                 retval.errorMsg = "Unable to update deployment configuration: " + result;
                 return retval;
             }
 
-            retval.deploymentString = deploymentString;
+            // ignore admin mode settings if operating with new start actions/verbs
+            if (!VoltDB.instance().isRunningWithOldVerbs() && dt.getAdminMode().isAdminstartup()) {
```
vtkstef: for future consideration do your utmost to avoid using the VoltDB.instance() singleton. -Thanks

src/frontend/org/voltdb/RealVoltDB.java
hhakimi53: Nothing wrong, just fewer wordings in generated deployment when doing that way :). Went with dt.getAdminMode().setAdminstartup(null) instead of dt.getAdminMode().setAdminstartup(false) as when generating deployment, we get <admin-mode/> when the startup was set to null instead <admin-mode adminstartup="false"/>. I can set it to false instead of null.
I will leave port as is.

src/frontend/org/voltdb/compiler/AsyncCompilerAgentHelper.java
hhakimi53: Sure.

hhakimi53:
Thanks for looking at the changes. I have pushed updates based on the review feedback.

vtkstef:
Reviewed

PR #3769, opened 2016-07-21
Prevent voltdb to recover from a partial snapshot.

yt14:
Compare tables in snapshot digest file with tables in catalog jar. If there is a dismatch, ignore that snapshot.

- Create getNormalTableNamesfromInMemoryJar() in CatalogUtil, which returns all the normal table names in a catalog in-memory jar file.

- Call getNormalTableNamesfromInMemoryJar() in RestoreAgent to compare the table names in catalog with table names in digest.

### DIFF
--- a/src/frontend/org/voltdb/RestoreAgent.java
+++ b/src/frontend/org/voltdb/RestoreAgent.java
@@ -61,6 +61,7 @@ import org.voltdb.jni.ExecutionEngine;
 import org.voltdb.sysprocs.saverestore.SnapshotUtil;
 import org.voltdb.sysprocs.saverestore.SnapshotUtil.Snapshot;
 import org.voltdb.sysprocs.saverestore.SnapshotUtil.TableFiles;
+import org.voltdb.utils.CatalogUtil;
 import org.voltdb.utils.InMemoryJarfile;
 import org.voltdb.utils.MiscUtils;
 
@@ -829,11 +830,21 @@ SnapshotCompletionInterest, Promotable
                                 .append(" because catalog CRC did not match digest.");
                 return null;
             }
+            // Make sure this is not a partial snapshot.
+            // Compare digestTableNames with all normal table names in catalog file.
+            // A normal table is one that's NOT a materialized view, nor an export table.
+            Set<String> catalogNormalTableNames = CatalogUtil.getNormalTableNamesfromInMemoryJar(jarfile);
+            if (!catalogNormalTableNames.equals(digestTableNames)) {
+                m_snapshotErrLogStr.append("\nRejected snapshot ")
+                                .append(s.getNonce())
+                                .append(" because this is a partial snapshot.");
+                return null;
+            }
         }
         catch (IOException ioe) {
             m_snapshotErrLogStr.append("\nRejected snapshot ")
                             .append(s.getNonce())
-                            .append(" because catalog CRC could not be validated");
+                            .append(" because catalog file could not be validated");
             return null;
         }
         finally {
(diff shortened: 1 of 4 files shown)


nshi:
Great! I like the patch. Could you also add a unit test for the utility method to TestCatalogUtil?

src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -829,11 +830,21 @@ private SnapshotInfo checkSnapshotIsComplete(Long key, Snapshot s)
                                 .append(" because catalog CRC did not match digest.");
```
yt14: I am suggesting that the code above can be replaced by:
`byte[] bytes = MiscUtils.fileToBytes(file);`
`InMemoryJarfile jarfile = CatalogUtil.loadInMemoryJarFile(bytes);`
For simplicity and better usage of utility functions. What do you think?

manjujames: I like that suggestion!

nshi: Fine with me, too. You should also consider making `MiscUtils.fileToBytes()` use a loop to read the bytes because it is not guaranteed that a single read will get all the bytes.

src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -95,16 +95,28 @@ public static void copyFile(String fromPath, String toPath) throws Exception {
      *             If there are errors reading the file
      */
     public static byte[] fileToBytes(File path) throws IOException {
-        FileInputStream fin = new FileInputStream(path);
-        byte[] buffer = new byte[(int) fin.getChannel().size()];
+        FileInputStream fin = null;
+        byte[] buffer = new byte[(int) path.length() + 1000];
```
nshi: I would recommend using the original `fin.getChannel().size()` to get the file size here because it's more efficient. `path.length()` does more IOs.

I'm also not sure the 1000 bytes are needed.

src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -95,16 +95,28 @@ public static void copyFile(String fromPath, String toPath) throws Exception {
...
+        int totalBytes = 0;
         try {
-            if (fin.read(buffer) == -1) {
-                throw new IOException("File " + path.getAbsolutePath() + " is empty");
+            fin = new FileInputStream(path);
+            while (readBytes >= 0) {
+                totalBytes += readBytes;
+                readBytes = fin.read(buffer, totalBytes, buffer.length - totalBytes - 1);
             }
+        } catch(IOException ioe) {
```
nshi: nitpick: missing a space after `catch`.

src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -95,16 +95,28 @@ public static void copyFile(String fromPath, String toPath) throws Exception {
...
-            fin.close();
+            if (fin != null) {
+                try {
+                    fin.close();
+                }
+                catch (Exception e) {}
+            }
         }
-        return buffer;
+        byte[] bytes = Arrays.copyOf(buffer, totalBytes);
```
nshi: You don't need to copy the bytes, just return `buffer`.

src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -88,23 +88,21 @@ public static void copyFile(String fromPath, String toPath) throws Exception {
...
+        byte[] bytes = null;
         try {
-            if (fin.read(buffer) == -1) {
-                throw new IOException("File " + path.getAbsolutePath() + " is empty");
-            }
-        } finally {
-            fin.close();
+            bytes = Files.readAllBytes(path.toPath());
+        } catch (IOException ioe) {
+            throw new IOException("File " + path.getAbsolutePath() + " is empty");
```
nshi: This error message is no longer accurate. You don't have catch it here, just let it throw.

PR #3773, opened 2016-07-21
Fix ENG-10749, 10750

yzhang1991:
I today found the two bugs are actually the same one.
The parameter problem, although the data type didn't show correct, does not affect the behavior of the EE. So I guess there are some other places in the code that can detect the DECIMAL and BIGINT type issue.

The real problem was serializing TIMESTAMP to JSON, I just added the code to make it support both long value string and date string.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
@@ -683,6 +683,7 @@ public class FunctionForVoltDB extends FunctionSQL {
                             catch (Exception e) {
                                 throw Error.error(ErrorCode.X_42561);
                             }
+                            nodes[i].valueData = paramTypes[i].castToType(session, nodes[i].valueData, nodes[i].dataType);
                             nodes[i].dataType = paramTypes[i];
                         }
                     } else if (paramTypes[i].isNumberType() && !nodes[i].dataType.isNumberType()) {
--- a/tests/frontend/org/voltdb/planner/TestFunctions.java
+++ b/tests/frontend/org/voltdb/planner/TestFunctions.java
@@ -126,4 +126,9 @@ public class TestFunctions extends PlannerTestCase {
         compile("select case when varchar_type like 'M%' then 1 end as m_state from bit;");
         compile("select case when varchar_type like '_%' then 1 end as m_state from bit;");
     }
+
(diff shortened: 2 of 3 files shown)


src/frontend/org/voltdb/expressions/ConstantValueExpression.java
```diff
@@ -158,7 +160,11 @@ public void toJSONString(JSONStringer stringer) throws JSONException {
                 stringer.value(m_value);
                 break;
             case TIMESTAMP:
-                stringer.value(Long.valueOf(m_value));
+                try {
+                    stringer.value(Long.valueOf(m_value));
+                } catch (NumberFormatException  e) {
+                    stringer.value(Timestamp.valueOf(m_value));
```
wolffcm: So sometimes we create a CVE here `m_value` is formatted as a long, and sometimes as a date and a time.  I wonder if there is a way to make the various codepaths that reach here create CVEs more consistently, picking either a long representation or a date+time representation?

wolffcm:
That's all my comments.  I would like to know if this issue could have been caught earlier be ensuring that CVEs are created more consistently.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -683,6 +683,7 @@ else if (paramTypes[i].canConvertFrom(nodes[i].dataType)) {
                             catch (Exception e) {
                                 throw Error.error(ErrorCode.X_42561);
                             }
+                            nodes[i].valueData = paramTypes[i].castToType(session, nodes[i].valueData, nodes[i].dataType);
```
wolffcm: So it seems like we were just changing the data type without changing the actual value before?

Why was this a problem for SINCE_EPOCH but not for TRUNCATE?

That's all my comments.

yzhang1991:
There are many places we call `ParameterConverter.tryToMakeCompatible()` to do the conversion, but we didn't cover function parameters.
The query that our SQL coverage utility generated happened to have the `TIMESTAMP` parameter not being parameterized in the query, so we can see the problem rise up when serializing the plan.
The simplest reproduce for error could be:
```sql
CREATE TABLE T (a INT);
explain SELECT SINCE_EPOCH(MICROSECOND, '1812-10-28 07:30:43') FROM t WHERE a>?;
```

There is no unique problem with SINCE_EPOCH, it turned out to be a false verdict.

PR #3774, opened 2016-07-21
Eng 10622

manjujames:
PersistentBinaryDeque changes to support multiple readers.
I still need to figure out what to do with cursors when we close PBD. May be this can be done with the next ticket where we make changes to InvocationBufferQueue.


src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -220,17 +243,48 @@ public void sync() throws IOException {
...
     @Override
-    public boolean isEmpty() throws IOException
-    {
+    public boolean hasAllFinishedReading() throws IOException {
         if (m_closed) throw new IOException("Segment closed");
-        return m_discardCount == m_numOfEntries;
+
+        if (m_readCursors.size() == 0) return false;
+
+        for (String cid : m_readCursors.keySet()) {
```
luyangco: Use `values()` instead of `keySet()` can save a map lookup for you.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -220,17 +243,48 @@ public void sync() throws IOException {
...
+    public boolean isCursorEmpty(String cursorId) throws IOException {
+        if (m_closed) throw new IOException("Segment closed");
+        return m_readCursors.get(cursorId).m_discardCount == m_numOfEntries;
+    }
+
+    @Override
+    public boolean isSegmentEmpty() throws IOException {
+        if (m_closed) throw new IOException("Segment closed");
+
+        for (String cid : m_readCursors.keySet()) {
```
luyangco: Same comment applies here.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -376,41 +618,38 @@ public synchronized void push(BBContainer objects[]) throws IOException {
...
-            if (s.isClosed()) {
-                s.open(false);
-            }
-
-            if (s.hasMoreEntries()) {
-                segment = s;
-                retcont = segment.poll(ocf);
-                break;
-            }
+        if (!m_readCursors.containsKey(cursorId)) {
```
luyangco: Does `readCursors` contain null entry? if not, another map lookup can be saved by calling `get()` first.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -376,41 +618,38 @@ public synchronized void push(BBContainer objects[]) throws IOException {
...
+        long firstSegmentId = peekFirstSegment().segmentId();
+        for (ReadCursor cursor : m_readCursors.values()) {
+            cursor.m_segmentId = firstSegmentId;
+        }
+    }
+
     @Override
-    public synchronized BBContainer poll(OutputContainerFactory ocf) throws IOException {
-        assertions();
+    public synchronized BinaryDequeReader openForRead(String cursorId) throws IOException {
```
luyangco: Does `openForRead()` be called from different threads? I mean `synchronized` keyword might not necessary.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+                if (m_closed) {
+                    throw new IOException("Reader " + m_cursorId + " has been closed");
+                }
+                assertions();
+
+                BBContainer retcont = null;
+                PBDSegment segment = null;
+
+                // It is possible that the last segment got removed
+                long firstSegmentId = peekFirstSegment().segmentId();
```
luyangco: `peekFirstSegment()` can be null.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+            return numObjects;
+        }
+
+        public void close() {
+            m_closed = true;
+        }
+
+        @Override
+        public BBContainer poll(OutputContainerFactory ocf) throws IOException {
+            synchronized (PersistentBinaryDeque.this) {
```
luyangco: Is there a reason why you don't allow multiple reads on the same segment in parallel?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+
+                BBContainer retcont = null;
+                PBDSegment segment = null;
+
+                // It is possible that the last segment got removed
+                long firstSegmentId = peekFirstSegment().segmentId();
+                while (m_segmentId < firstSegmentId) m_segmentId++;
+
+                long currSegmentId = m_segmentId;
+                while (m_segments.containsKey(currSegmentId)) {
```
luyangco: Same here, `containsKey()` is redundant . Use `get()`.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
```diff
@@ -0,0 +1,154 @@
...
+
+public class TestPBDMultipleReaders {
+
+    private final static VoltLogger logger = new VoltLogger("EXPORT");
+
+    private PersistentBinaryDeque m_pbd;
+
+    private class PBDReader {
+        private String m_readerId;
+        private int m_totalRead;
```
aquafemi: I think this needs to be initialized?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
aquafemi: Same as what Yang said earlier with `values`

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -568,97 +787,27 @@ public static TruncatorResponse fullTruncateResponse() {
...
-        }
-        m_segments.offer(newSegment);
-        assertions();
-    }
-
     private void addToNumObjects(int num) {
-        assert(m_numObjects >= 0);
-        m_numObjects += num;
+        for (ReadCursor cursor : m_readCursors.values()) {
+            assert(cursor.m_numObjects >= 0);
```
luyangco: Because input can be negative, better to assert after `line 793`.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: OK.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: OK

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -224,30 +360,136 @@ public boolean accept(File pathname) {
...
+        PBDSegment newSegment = newSegment(newSegmentIndex, new VoltFile(m_path, m_nonce + "." + newSegmentIndex + ".pbd"));
+        newSegment.openForWrite(true);
+        if (m_usageSpecificLog.isDebugEnabled()) {
+            m_usageSpecificLog.debug("Segment " + newSegment.file() + " has been created by PBD truncator");
         }
+        m_segments.put(newSegment.segmentId(), newSegment);
+        assertions();
+    }
+
+    private PBDSegment newSegment(long segmentId, File file) {
```
luyangco: No need to create another function to wrap `new PBDRegularSegment(segmentId, file)` , especially regular segment is the only option now.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: I read up a bit about readwrite locks, which raises questions about how much parallelism this actually gives when there if frequent writes and reads (which is our case). But, I agree that this is something we should definitely try. May be I will do it once this is reviewed.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: No, this can never be null.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: OK.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Yes, openForRead may be called from different threads.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: OK.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: OK.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
manjujames: You mean totalRead. That'll be 0.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -224,30 +360,136 @@ public boolean accept(File pathname) {
...
-        final PBDSegment last = m_segments.peekLast();
+        PBDSegment last = peekLastSegment();
         if (last != null && !last.isBeingPolled()) {
             last.close();
         }
-        m_segments.offer(newSegment);
+        m_segments.put(newSegment.segmentId(), newSegment);
+    }
+
+    private PBDSegment peekFirstSegment() {
```
luyangco: Those functions are confusing because they may return `null` but under our assumption pdb should always has at least one segment. Maybe `assert(entry != null)` is more readable.

luyangco:
Review complete.

nshi:
|`src/frontend/org/voltdb/utils/PBDSegment.java:66`|
|---|
An alternative is to return the cursor object here and have the caller pass the cursor object in to the other methods, instead of passing the cursorId string. This avoids the map lookup (and string comparison) and also guards against read before opening.

nshi:
|`src/frontend/org/voltdb/utils/PersistentBinaryDeque.java:118`|
|---|
When can this happen?

nshi:
|`src/frontend/org/voltdb/utils/PersistentBinaryDeque.java:135`|
|---|
In this case, don't you also want to increment `m_segmentId`?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: OK.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+
+                if (retcont == null) {
+                    return null;
+                }
+
+                decrementNumObjects();
+                assertions();
+                assert (retcont.b() != null);
+                return wrapRetCont(segment, retcont);
+            }
```
nshi: |`line 377`|
|---|
Why read it again here without using the return value?

nshi:
|`src/frontend/org/voltdb/utils/PersistentBinaryDeque.java:701`|
|---|
I don't think so, because `close()` doesn't delete the persisted queue. They may be read later on recover.

nshi:
|`src/frontend/org/voltdb/utils/PersistentBinaryDeque.java:734`|
|---|
This should also clear all the cursors.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -26,6 +26,8 @@
 import java.io.IOException;
```
wweiss-voltdb: |`line 41`|
|---|
An int or byte based lookup would be more efficient than a string as the key

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -198,6 +220,7 @@ public boolean isClosed()
 
     @Override
     public void close() throws IOException {
+        m_readCursors.clear();
         try {
             if (m_fc != null) {
```
slowsoul: A comment not related to this ticket, m_ras is actually the parent of the FileChannel m_fc. It would be better to close m_ras (will close m_fc according to doc) here although closing m_fc actually closes m_ras (but not documented).

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -36,16 +38,11 @@
 public class PBDRegularSegment extends PBDSegment {
     private static final VoltLogger LOG = new VoltLogger("HOST");
 
-    //Index of the next object to read, not an offset into the file
-    private int m_objectReadIndex = 0;
-    private int m_bytesRead = 0;
```
wweiss-voltdb: |`line 137`|
|---|
With a single cursor, a file is opened for read as needed and then closed and deleted when we move past the file. However, now, all the files between 2 read cursors will be left open, which is undesirable. We need a way for each read cursor to independently open and close files provided no other read cursors have the file open.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
slowsoul: Since m_segments is a TreeMap, I think you can just iterate over m_segments.entrySet() and update m_segmentId to the id of the first segment that return true for hasMoreEntries(). No containsKey(), get() or peekFirstSegment() needs to be called.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
slowsoul: I think it is to read out the line feed character which is to make readUTF() work. BTW, UpSource comments seems to have a big offset to the actual line that are being commented.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
slowsoul: I think m_segmentId for the cursor is to record the segment the cursor is currently reading, you can actually save a reference "m_segment" to the segment instead and whenever the segment has no more entries, move to the next segment by calling higherEntry() (can be null). m_segmentId is no longer needed as it can be retrieved by segmentId()

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+         */
+        @Override
+        public long sizeInBytes() throws IOException {
+            synchronized(PersistentBinaryDeque.this) {
+                if (m_closed) {
+                    throw new IOException("Reader " + m_cursorId + " has been closed");
+                }
+                assertions();
+                long size = 0;
+                for (PBDSegment segment : m_segments.values()) {
```
slowsoul: The loop range can be changed to m_segments.tailMap(m_segmentId).values(), since the segment numbered m_segmentId is currently read by this cursor, the segments before are already read and uncompressedBytesToRead(m_cursorId) will always return 0 for these segments.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+
+        @Override
+        public boolean isEmpty() throws IOException {
+            synchronized(PersistentBinaryDeque.this) {
+                if (m_closed) {
+                    throw new IOException("Closed");
+                }
+                assertions();
+
+                for (PBDSegment s : m_segments.values()) {
```
slowsoul: The loop range can be changed to m_segments.tailMap(m_segmentId).values(), since the segment numbered m_segmentId is currently read by this cursor, the segments before are already read and hasMoreEntries(m_cursorId) will always return false for these segments.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +75,141 @@ public synchronized void discard() {
...
+                        break;
+                    }
+                    currSegmentId++;
+                }
+
+                if (retcont == null) {
+                    return null;
+                }
+
+                decrementNumObjects();
```
slowsoul: m_numObjects is initially assigned in the constructor and only decremented here in poll(). However, as more objects are offered to the wrapping PBD, the cursors' m_numObjects need to be updated as well, otherwise they can't reflect the actual number of objects if called.

slowsoul: Sorry, I didn't notice the increment of cursor.m_numObjects in PBD methods.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
slowsoul: One suggestion to close segment files when there are no cursors reading them is to add an explicit closeForRead(String cursorId) method and a closeForWrite() for PBDSegment. ReadCursor.poll() calls closeForRead() when there are no entries to be read at the moment (The writer cursor (if any) may write more entries later). PBD.offer() calls closeForWrite() when the current write segment cannot hold any more entries.
For closeForRead() and closeForWrite(), they should check if there are any read/write cursors after removing itself. If no cursors are left, the segment file can be closed for now.

slowsoul: Actually a read cursor should check if there is a write cursor for the segment at the moment. If there is, do nothing as more entries will be written, otherwise, call closeForRead() to remove itself from the read cursor set for the segment.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: We will look into changing this when we work on ENG-10627.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Thanks. I will work on this.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: OK, thanks.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Yes, it is to read the newline. Technically, we don't need the newline, but I just thought it would be nice to make the file readable for us too.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: I changed this method to use segment.size() for segments that this cursor hasn't started reading (no openForRead) and just uncompressedBytesToRead for segment that is currently being read.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Changed this one as well to not open any unnecessary cursors.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
aquafemi: Ah ok. My mistake

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -19,13 +19,17 @@
 
 import org.voltcore.logging.VoltLogger;
 import org.voltcore.utils.DBBPool;
+import org.voltcore.utils.DBBPool.BBContainer;
```
nshi: What's the purpose of the `m_closedCursors` map?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -224,30 +356,149 @@ public boolean accept(File pathname) {
...
+    private int countNumObjects() throws IOException {
+        int numObjects = 0;
+        for (PBDSegment segment : m_segments.values()) {
+            numObjects += segment.getNumEntries();
         }
+
+        return numObjects;
+    }
+
+    private void readCursorFile(File file) {
```
nshi: |`line 661-662`|
|---|
Can you CRC the entries written to the cursor file and verify the checksum on read? I'm thinking about cases where server powers off on cursor file manipulation. It may read invalid cursor on recover and never advance because the cursor doesn't exist.

manjujames: Actually the whole cursor file handling needs to be done differently. It has to be a global one instead of the way I have it now. We will look into that as part of Rui's ticket.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -112,51 +102,45 @@ public BBContainer poll(OutputContainerFactory ocf) throws IOException {
...
+                long lastSegmentId = peekLastSegment().segmentId();
+                while (!segmentReader.hasMoreEntries()) {
+                    if (m_segment.segmentId() == lastSegmentId) { // nothing more to read
+                        return null;
+                    }
 
-                if (retcont == null) {
-                    return null;
+                    segmentReader.close();
+                    m_segment = m_segments.higherEntry(m_segment.segmentId()).getValue();
```
slowsoul: ReadCursor.poll(), sizeInBytes() and isEmpty() can use iterator on top of TreeMap.tailMap() instead of calling higherEntry() every time

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -112,51 +102,45 @@ public BBContainer poll(OutputContainerFactory ocf) throws IOException {
...
+                moveToValidSegment();
 
-                    if (s.hasMoreEntries(m_cursorId)) {
-                        segment = s;
-                        retcont = segment.poll(m_cursorId, ocf);
-                        m_segmentId = currSegmentId;
-                        break;
-                    }
-                    currSegmentId++;
+                PBDSegmentReader segmentReader = m_segment.getReader(m_cursorId);
```
slowsoul: SegmentReader.m_closed is only set to true but not false, no open method, maybe need to change openForRead(): if it is in m_readCursors, return it, else if it is in m_closedCursors, open it and return it, otherwise, create a new cursor and return it. getReader() does the first two steps but return null otherwise. We can simplify ReadCursor.poll() and PBD.canDeleteSegment() with the change.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -685,6 +696,26 @@ public void discard() {
         };
     }
 
+    private boolean canDeleteSegment(PBDSegment segment) throws IOException {
+        for (ReadCursor cursor : m_readCursors.values()) {
+            if (cursor.m_segment != null && (cursor.m_segment.segmentId() >= segment.segmentId())) {
+                PBDSegmentReader segmentReader = segment.getReader(cursor.m_cursorId);
+                if (cursor.m_segment.segmentId() == segment.segmentId()) { // reading this segment now
```
slowsoul: Is this check necessary? I mean "cursor.m_segment.segmentId() == segment.segmentId()" and "segmentReader == null" won't be both true because if cursor.m_segment equals to segment, it must have a segmentReader either in m_readCursors or in m_closedCursors, so the segmentReader returned by getReader() cannot be null in this case.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -826,25 +844,29 @@ private void assertions() {
...
+            try {
+                for (PBDSegment segment : m_segments.values()) {
+                    PBDSegmentReader reader = segment.getReader(cursor.m_cursorId);
+                    if (reader == null) {
+                        numObjects += segment.getNumEntries();
+                    } else {
+                        numObjects += segment.getNumEntries() - reader.readIndex();
                     }
                 }
+                assert numObjects == cursor.getNumObjects() : numObjects + " != " + cursor.getNumObjects();
```
slowsoul: ReadCursor.getNumObjects() currently returns the entries in this segment that haven’t been polled out from this cursor, this includes those segments after the cursor, the current segment but also entries of segments that has already been removed from the PBD. On the other hand, numObjects now counts the number of entries that haven’t been pulled out from the cursor for all *existing* segments. The assertion will fail for cursors that were created after some segments have been polled out from the PBD.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -73,6 +76,135 @@ public synchronized void discard() {
...
+            m_cursorId = cursorId;
+        }
+
+        public void close() {
+            m_closed = true;
+        }
+
+        @Override
+        public BBContainer poll(OutputContainerFactory ocf) throws IOException {
+            synchronized (PersistentBinaryDeque.this) {
```
wweiss-voltdb: I think that moving the wrapper for the BBContainer from the PBD to the ReadCursor would simplify accounting for the outstanding buffers, segments and cursors.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -224,30 +356,149 @@ public boolean accept(File pathname) {
...
+        try {
+            m_cursorsWriter = new RandomAccessFile(file, "rwd");
+            String cursorId = null;
+            while ((cursorId=m_cursorsWriter.readUTF()) != null) {
+                m_readCursors.put(cursorId, new ReadCursor(cursorId));
+                m_cursorsWriter.readUTF();
+            }
+        } catch(EOFException e) {
+            // Nothing to read
+        } catch(IOException e) {
```
slowsoul: I think FileNotFoundException also needs to be caught here, in the case of a fresh start.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: I had thought about doing it that way as well. However, we still have to go through all readCursors on the PBD and not just the current cursor. So it felt more like a method that should belong on the PBD. If we move it to the reader, we will be using *almost* the same exact code.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: It will not FileNotFoundException because of 'rw' mode.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: We need to keep them around until the segment is deleted. We need to look through all cursors to see if they are done reading the segment on a discard callback.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Right. I just left it there because we still have the interface and it makes it convenient to switch to a different impl.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Yes, you are right. I will change it to an assertion.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Yes, thanks. For poll though, I prefer it without tailMap because it doesn't need to iterate through all the entries.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: The way the implementation is, a reader should never open a segment more than once. Once a reader is closed, it shouldn't be reopened, or we will lose track of the fact that the reader has finished reading the segment. Only a close of the segment should clear out the cursors, closed or open.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Thanks for catching that. I will fix it and add a unit test case to cover it.

src/frontend/org/voltdb/utils/BinaryDeque.java
```diff
@@ -70,35 +70,73 @@
...
+    /**
+     * Reader class used to read entries from the deque. Multiple readers may be active at the same time,
+     * each of them maintaining their own read location within the deque.
+     */
+    public interface BinaryDequeReader {
+        /**
+         * Read and return the object at the current read position of this reader.
+         * The entry will be removed once all active readers have read the entry.
+         * @param ocf
+         * @return
```
akhanzode: @return BBContainer

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -36,16 +40,12 @@
 public class PBDRegularSegment extends PBDSegment {
     private static final VoltLogger LOG = new VoltLogger("HOST");
 
-    //Index of the next object to read, not an offset into the file
-    private int m_objectReadIndex = 0;
-    private int m_bytesRead = 0;
-    // Maintains the read byte offset
-    private long m_readOffset = SEGMENT_HEADER_BYTES;
+    private Map<String, SegmentReader> m_readCursors = new HashMap<>();
```
akhanzode: Could be final?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -36,16 +40,12 @@
...
+    private Map<String, SegmentReader> m_closedCursors = new HashMap<>();
```
akhanzode: Same as above

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -73,10 +73,12 @@ public File file()
     public void reset()
     {
         m_syncedSinceLastEdit = false;
-        m_objectReadIndex = 0;
-        m_bytesRead = 0;
-        m_readOffset = SEGMENT_HEADER_BYTES;
-        m_discardCount = 0;
+        for (SegmentReader reader : m_readCursors.values()) {
+            reader.m_objectReadIndex = 0;
```
akhanzode: Would a reset or init method on reader help?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -36,16 +40,12 @@
...
-    private long m_readOffset = SEGMENT_HEADER_BYTES;
+    private Map<String, SegmentReader> m_readCursors = new HashMap<>();
+    private Map<String, SegmentReader> m_closedCursors = new HashMap<>();
 
     //ID of this segment
     private final Long m_index;
 
-    private int m_discardCount;
     private int m_numOfEntries = -1;
     private int m_size = -1;
```
akhanzode: how come size is -1 by default maybe some comments on this will help.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -86,43 +88,62 @@ public void reset()
...
             PBDUtils.readBufferFully(m_fc, m_tmpHeaderBuf.b(), COUNT_OFFSET);
             m_numOfEntries = m_tmpHeaderBuf.b().getInt();
             m_size = m_tmpHeaderBuf.b().getInt();
-            return m_numOfEntries;
         } else {
             m_numOfEntries = 0;
             m_size = 0;
-            return 0;
         }
+        if (wasClosed) closeReadersAndFile();
```
akhanzode: should the wasClosed be isOpen sounds like if (closed) close()

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -86,43 +88,62 @@ public void reset()
...
         }
         if (m_fc.size() > 0) {
             m_tmpHeaderBuf.b().clear();
             PBDUtils.readBufferFully(m_fc, m_tmpHeaderBuf.b(), COUNT_OFFSET);
             m_numOfEntries = m_tmpHeaderBuf.b().getInt();
             m_size = m_tmpHeaderBuf.b().getInt();
-            return m_numOfEntries;
         } else {
             m_numOfEntries = 0;
             m_size = 0;
```
akhanzode: Why do they become 0 and not -1 seems odd.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -86,43 +88,62 @@ public void reset()
     @Override
     public int getNumEntries() throws IOException
     {
+        boolean wasClosed = false;
         if (m_closed) {
-            open(false);
+            wasClosed = true;
+            open(false, false);
```
akhanzode: Why not cache when closed enries and size so you dont need to open again just to get sizes?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -86,43 +88,62 @@ public void reset()
...
-        return m_objectReadIndex != 0;
+        return !m_readCursors.isEmpty();
     }
 
     @Override
-    public int readIndex()
-    {
-        return m_objectReadIndex;
+    public boolean isOpenForReading(String cursorId) {
+        return m_readCursors.containsKey(cursorId);
```
akhanzode: cursorId case sensitive?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -86,43 +88,62 @@ public void reset()
...
     }
 
     @Override
-    public void open(boolean forWrite) throws IOException
+    public PBDSegmentReader openForRead(String cursorId) throws IOException
     {
-        open(forWrite, forWrite);
+        if (m_readCursors.containsKey(cursorId) || m_closedCursors.containsKey(cursorId)) {
```
akhanzode: Do these cursors maps need synchronization or the whole assumption that this is not thread safe?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -198,9 +220,15 @@ public boolean isClosed()
 
     @Override
     public void close() throws IOException {
+        m_closedCursors.clear();
```
akhanzode: closeReadersAndFile already clears so you dont need it here?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -198,9 +220,15 @@ public boolean isClosed()
...
+        closeReadersAndFile();
+    }
+
+    private void closeReadersAndFile() throws IOException {
+        m_readCursors.clear();
         try {
-            if (m_fc != null) {
-                m_fc.close();
+            if (m_ras != null) {
+                m_ras.close();
```
akhanzode: If IOException is thrown m_readCursors will be cleared in that event. Should it be in finally?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -308,126 +337,173 @@ public int offer(DeferredSerialization ds) throws IOException
...
     {
-        if (m_closed) throw new IOException("closed");
+        int written = 0;
+        final DBBPool.BBContainer partialCont = DBBPool.allocateDirect(length);
+        try {
+            written += entry.writeTruncatedObject(partialCont.b());
+            partialCont.b().flip();
+
+            while (partialCont.b().hasRemaining()) {
+                m_fc.write(partialCont.b());
```
akhanzode: if second write fails what happens to entry object?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -308,126 +337,173 @@ public int offer(DeferredSerialization ds) throws IOException
...
+        private final String m_cursorId;
+        private long m_readOffset = SEGMENT_HEADER_BYTES;
+        //Index of the next object to read, not an offset into the file
+        private int m_objectReadIndex = 0;
+        private int m_bytesRead = 0;
+        private int m_discardCount = 0;
+        private boolean m_closed = false;
+
+        public SegmentReader(String cursorId) {
+            m_cursorId = cursorId;
```
akhanzode: precondition checks for null and assert for tests please.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -308,126 +337,173 @@ public int offer(DeferredSerialization ds) throws IOException
...
         }
 
-        final long writePos = m_fc.position();
-        m_fc.position(m_readOffset);
-        m_objectReadIndex++;
+        @Override
+        public boolean hasMoreEntries() throws IOException {
+            return m_objectReadIndex < m_numOfEntries;
```
akhanzode: what if readindex == m_numberOfEntries does it mean we have more entries?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -308,126 +337,173 @@ public int offer(DeferredSerialization ds) throws IOException
...
 
-            final DBBPool.BBContainer retcont;
-            if (compressed) {
-                final DBBPool.BBContainer compressedBuf = DBBPool.allocateDirectAndPool(length);
-                try {
-                    while (compressedBuf.b().hasRemaining()) {
-                        int read = m_fc.read(compressedBuf.b());
+            final long writePos = m_fc.position();
+            m_fc.position(m_readOffset);
+            m_objectReadIndex++;
```
akhanzode: You havent read until you have a BBContainer to return right? maybe I am misreading something.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -308,126 +337,173 @@ public int offer(DeferredSerialization ds) throws IOException
...
-    @Override
-    protected int writeTruncatedEntry(BinaryDeque.TruncatorResponse entry, int length) throws IOException
-    {
-        int written = 0;
-        final DBBPool.BBContainer partialCont = DBBPool.allocateDirect(length);
-        try {
-            written += entry.writeTruncatedObject(partialCont.b());
-            partialCont.b().flip();
+        @Override
+        public int readIndex() {
```
akhanzode: Whats the diff between readIndex ad readOffset is not clear to me.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -308,126 +337,173 @@ public int offer(DeferredSerialization ds) throws IOException
...
+                closeReadersAndFile();
             }
-        } finally {
-            partialCont.discard();
         }
-        return written;
+
+        @Override
+        public boolean isClosed() {
+            return m_closed == true;
```
akhanzode: just return m_closed.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -224,30 +357,149 @@ public boolean accept(File pathname) {
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public synchronized void parseAndTruncate(BinaryDequeTruncator truncator) throws IOException {
```
akhanzode: how come truncator is not specific to cursor?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -446,49 +697,54 @@ public void discard() {
...
-        }
-        m_closed = true;
-    }
-
-    @Override
-    public synchronized boolean isEmpty() throws IOException {
-        assertions();
-        if (m_closed) {
-            throw new IOException("Closed");
+        for (ReadCursor cursor : m_readCursors.values()) {
```
akhanzode: Shouldnt you close all cursors before setting m_closed=true

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -446,49 +697,54 @@ public void discard() {
...
-                if (wasClosed) {
-                    s.close();
-                }
-            }
+        for (PBDSegment segment : m_segments.values()) {
+            segment.close();
         }
-        return true;
+        m_cursorsWriter.close();
+        m_closed = true;
```
akhanzode: oh real close is here maybe you dont need earlier m_close setting.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
```diff
@@ -0,0 +1,278 @@
...
+        private BinaryDequeReader m_reader;
+
+        public PBDReader(String readerId) throws IOException {
+            m_readerId = readerId;
+            m_reader = m_pbd.openForRead(m_readerId);
+        }
+
+        public int readToEndOfSegment() throws Exception {
+            int end = (m_totalRead/47 + 1) * 47;
```
akhanzode: whats significance of 47? maybe make it constant?

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
```diff
@@ -0,0 +1,278 @@
...
+
+        int numSegments = 5;
+        for (int i=0; i<numSegments; i++) {
+            for (int j=0; j<47; j++) {
+                m_pbd.offer( DBBPool.wrapBB(TestPersistentBinaryDeque.getFilledBuffer(j)) );
+            }
+            assertEquals(1, m_pbd.numOpenSegments());
+        }
+
+        BinaryDequeReader reader0 = m_pbd.openForRead("reader0");
```
akhanzode: test for using same cusrsor name twice?

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
```diff
@@ -0,0 +1,278 @@
...
+    public void tearDown() throws Exception {
+        try {
+            m_pbd.close();
+        } catch (Exception e) {}
+        try {
+            TestPersistentBinaryDeque.tearDownTestDir();
+        } finally {
+            m_pbd = null;
+        }
+        System.gc();
```
akhanzode: GC and finalization what is it for?

src/frontend/org/voltdb/utils/BinaryDeque.java
manjujames: OK

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Yes

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Sure.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Yes, probably doable. But, that is a bigger change, that I didn't want to take on as part of this work.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Yes, that is the intention - 'if it was closed, close it' versus 'if it was not open, close it'.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Yes

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: The assumption is that this is only called from PersistentBinaryDeque class, which does the synchronization. So there is no need to synchronize here.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: This is clearing the closed readers as well.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Finally will do the work even if IOException is thrown. So, yes, it should be cleared.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: This is existing code. If anything fails here, truncation did not happen correctly and will need to be handled by the caller, I assume.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: OK

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
slowsoul: Yes, got it.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: No, that means we have read all that is currently there.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: This is existing code, but makes sense. I will move it to where it is incrementing bytesRead.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
slowsoul: I agree, since we will have small treemap size and possibly very few entries to go over for poll(), using higherEntry() will be better in most cases.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: Position in file versus number of objects.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
manjujames: OK

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Because truncate is on the PBD, not on a specific reader.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Yes, thanks.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
manjujames: 47 is the number of objects required to fill up a segment. Yea, I will make it a constant.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
manjujames: OK.

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
manjujames: This is for the code that ensures that the allocated buffers are discarded.

PR #3775, opened 2016-07-21
ENG-10792 cli spec fixes, and disable adds for new cli

vtkstef:


akhanzode:

PR #3779, opened 2016-07-25
ENG-10787 exit when hostcount is less then list of nodes

vtkstef:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -484,8 +484,8 @@ public void initialize(Configuration config) {
             if (config.m_startAction == StartAction.INITIALIZE) {
                 stageDeploymemtFileForInitialize(config, readDepl.deployment);
                 stageInitializedMarker(config);
-                hostLog.info("Initialized VoltDB on " + config.m_voltdbRoot.getPath());
-                consoleLog.info("Initialized VoltDB on " + config.m_voltdbRoot.getPath());
+                hostLog.info("Initialized VoltDB root directory " + config.m_voltdbRoot.getPath());
```
akhanzode: Can you print Absolute path here. It will be helpful to track.

PR #3781, opened 2016-07-26
Eng 10775 full join postgres

johnTcrawford:
Added a get_max_mismatches method, and use it to determine whether the tests have failed; and apply this to the current 'known' failures in the basic-joins, basic-index-joins, basic-compoundex-joins, numeric-decimals & numeric-ints test suites (all of these only on PostgreSQL).

### DIFF
--- a/tests/scripts/sql_coverage_test.py
+++ b/tests/scripts/sql_coverage_test.py
@@ -209,6 +209,31 @@ def run_once(name, command, statements_path, results_path, submit_verbosely, tes
     else:
         return 0
 
+def get_max_mismatches(comparison_database, suite_name):
+    """Returns the maximum number of acceptable mismatches, i.e., the number of
+    'known' failures for VoltDB to match the results of the comparison database
+    (HSQL or PostgreSQL), which is normally zero; however, there are sometimes
+    a few exceptions, e.g., for queries that are not supported by PostgreSQL.
+    """
+    max_mismatches = 0
+
+    # Kludge to not fail for known issues, when running against PostgreSQL
+    # (or the PostGIS extension of PostgreSQL)
+    if comparison_database.startswith('Post'):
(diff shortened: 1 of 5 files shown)


tests/scripts/sql_coverage_test.py
```diff
@@ -209,6 +209,31 @@ def run_once(name, command, statements_path, results_path, submit_verbosely, tes
...
+        if (config_name == 'basic-joins' or config_name == 'basic-index-joins' or
+              config_name == 'basic-compoundex-joins'):
+            max_mismatches = 4620
+        # Known failures in the numeric-decimals and numeric-ints "extended"
+        # test suites (see ENG-10546)
+        elif config_name == 'numeric-decimals':
+            max_mismatches = 1180
+        elif config_name == 'numeric-ints':
+            max_mismatches = 2820
+
```
BillWhite: These two numbers seem to be at the heart of the change, right?  You know that one test has no more than 1180 failures, and the other has no more than 2820 failures.  Is this sufficient?  Can't the actual failures change from run to run, maybe, if the code changes and we generate different plans, say?

If this is sufficient then everything else seems ok with me.

BillWhite:
I just had the one comment.  Other than that everything seems ok with me.

End of review round.

tests/scripts/sql_coverage_test.py
johnTcrawford: Actually, the heart of the change is two things. First, reorganized to make this more formal, in its own separate method, since I'm starting to think ignoring certain failures (on PostgreSQL) is going to be a regular thing, not a temporary kludge. Second, the adding of the "4620" number - the other two numbers were already there, just in a different place (see below); but yes, they sometimes change (though "4620" is consistent), but also yes, these numbers seem to be large enough.

PR #3782, opened 2016-07-26
replace the default SAFE reads option with a faster better SAFE option

xinjiacs:
We could move ShortCircuitReadLog into repair log with duplicate counter, but it will make the code review harder.  I propose to have a separate cleaning ticket with existing code.


src/frontend/org/voltdb/iv2/RepairLog.java
```diff
@@ -157,58 +156,65 @@ public void deliver(VoltMessage msg)
...
         }
-        else if (msg instanceof DumpMessage) {
+        if (msg instanceof DumpMessage) {
             String who = CoreUtils.hsIdToString(m_HSId);
             tmLog.warn("Repair log dump for site: " + who + ", isLeader: " + m_isLeader);
             tmLog.warn("" + who + ": lastSpHandle: " + m_lastSpHandle + ", lastMpHandle: " + m_lastMpHandle);
             for (Iv2RepairLogResponseMessage il : contents(0l, false)) {
                tmLog.warn("" + who + ": msg: " + il);
             }
+            return;
```
nshi: None of the changes in the RepairLog seem to be related to this patch. They are all style changes?

src/frontend/org/voltdb/iv2/ShortCircuitReadLog.java
```diff
@@ -0,0 +1,63 @@
...
+        m_shortCircuitReadSp = new ArrayDeque<InitiateResponseMessage>();
+
+        assert(mailbox != null);
+        m_mailbox = mailbox;
+    }
+
+    //  SPI offers a new message.
+    public void offerSp(InitiateResponseMessage msg, long handle)
+    {
+        if (msg.getSpHandle() >= handle) {
```
nshi: Do you mean `<=`?

xinjiacs: hmm, a code refactoring mistake. 

Do you have recommendation(existing) Junit test cases for this kind of code pass?

src/frontend/org/voltdb/iv2/RepairLog.java
xinjiacs: Yes, just code style then.

src/frontend/org/voltdb/iv2/ShortCircuitReadLog.java
nshi: This class itself could be easily unit tested, or starting a two node k=1 local cluster and construct a workload that will expose this could also do. This probably fails the existing network partition unit tests.

xinjiacs: Just kicked a jepsen test job on this branch.

src/frontend/org/voltdb/iv2/RepairLog.java
nshi: It makes the pull request noisy, considering this is an important change to a critical path.

nshi: Harder to back port, too.

src/frontend/org/voltdb/iv2/ShortCircuitReadLog.java
nshi: `TestJepsenStylePartition.java` is the test I referred to.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -767,7 +771,7 @@ private void handleBorrowTaskMessage(BorrowTaskMessage message) {
     // doesn't matter, it isn't going to be used for anything.
     void handleFragmentTaskMessage(FragmentTaskMessage message)
     {
-        boolean shortcutRead = message.isReadOnly() && (m_defaultConsistencyReadLevel == ReadLevel.FAST);
+        boolean shortcutRead = message.isReadOnly() && m_defaultConsistencyReadLevel.hasShortcutRead();
```
nshi: So real-only fragments are still being replicated?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -715,6 +714,11 @@ else if (result == DuplicateCounter.MISMATCH) {
             m_repairLogTruncationHandle = spHandle;
             m_mailbox.send(message.getInitiatorHSId(), message);
         }
+
+        if (m_defaultConsistencyReadLevel == ReadLevel.SAFE && m_isLeader && message.isReadOnly() &&
+                m_shortCircuitReadLog != null) {
+            m_shortCircuitReadLog.offerSp(message, m_repairLogTruncationHandle);
```
nshi: You also have to notify the `m_shortCircuitReadLog` whenever the truncation handle advances or the reads queued up in there will have a long latency and if there's no more txns after a read, that read will be stuck in there forever.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -78,6 +78,8 @@
     private final ArrayList<SocketChannel> m_clientChannels = new ArrayList<SocketChannel>();
     protected final String m_methodName;
 
+    public static int TIME_OUT_MILLIS = 10 * 60 * 1000;
```
nshi: This change seems irrelevant to the ticket.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -686,13 +679,22 @@ public void handleInitiateResponseMessage(InitiateResponseMessage message)
...
-            // the initiatorHSId is the ClientInterface mailbox.
-            m_mailbox.send(message.getInitiatorHSId(), message);
-            return;
+        if (message.isReadOnly()) {
+            if (m_defaultConsistencyReadLevel == ReadLevel.FAST) {
+                // the initiatorHSId is the ClientInterface mailbox.
+                m_mailbox.send(message.getInitiatorHSId(), message);
+                return;
+            }
+            if (m_defaultConsistencyReadLevel == ReadLevel.SAFE && m_isLeader) {
```
nshi: I think `m_leader` is always true here.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -803,7 +798,7 @@ void handleFragmentTaskMessage(FragmentTaskMessage message)
              * everywhere.
              * In that case don't propagate it to avoid a determinism check and extra messaging overhead
              */
-            if (m_sendToHSIds.length > 0 && (!shortcutRead || msg.isSysProcTask())) {
+            if (m_sendToHSIds.length > 0 && (!message.isReadOnly() || msg.isSysProcTask())) {
```
nshi: `FragmentResponseMessage` needs to be put into the `m_bufferedReadLog`, too.

nshi:
Can you also add unit tests for `BufferedReadLog` itself?

Sorry for nitpicking on the purely style-related changes. My concern is that at some point in the future, this may need to be back ported to some older versions. The less code it contains, the easier it is to back port and for other people to understand what the patch contains. So I would recommend backing out any changes that's not related to fixing this issue.

src/frontend/org/voltdb/iv2/SpScheduler.java
xinjiacs: Yes. I had comments three lines below indicating only test cases will be here. 
SAFE mode reads on replicas.

I need to change the test cases and assert the m_leader inside of the IF case then.

xinjiacs:
I closed this pull request, and will propose another one only with required changes, for easy backup purpose. 

For style changes or code cleanup, I will propose another separate one instead.

PR #3784, opened 2016-07-26
ENG-10830 init --force clears out deployment config paths absolute or non

vtkstef:


vtkstef:
Reviewed by @akhanzode

PR #3785, opened 2016-07-26
ENG-10358, add diagnostic code to elastic join path.

luyangco:
Change-Id: I31e71b3e0372f176c4d52787e598eecc339c2652

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -1424,6 +1424,12 @@ bool VoltDBEngine::isLocalSite(const int32_t pkHash) const
     return getPartitionForPkHash(pkHash) == m_partitionId;
 }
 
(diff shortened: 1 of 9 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1424,6 +1424,12 @@ bool VoltDBEngine::isLocalSite(const int32_t pkHash) const
     return getPartitionForPkHash(pkHash) == m_partitionId;
 }
 
+std::string VoltDBEngine::dumpCurrentHashinator() const {
+    std::ostringstream buffer;
```
nshi: You can simply return `m_hashinator.get()->debug()`.

luyangco: Yes, I'll change it.

PR #3786, opened 2016-07-26
ENG-10720: Assign currentTxnId for the task taken from SiteTaskQueue …

slowsoul:
…if the task log is emptied in the same loop before this task is handled


nshi:
Review finished.

PR #3792, opened 2016-07-27
Added a junit test point for windowed functions and partitioned tables.

BillWhite:
We test both partitioned tables in windowed partition by expressions and
in windowed order by expressions.


wolffcm:
In all these examples, the PARTITION BY clause contains just the partition key for the table.  What happens when you PARTITION BY something that is not?

In this case we need to do some work on the coordinator.  Do we get that right?

wolffcm:
Also, do we have planner tests that lock down the plans generated when we do RANK on a partitioned table where the PARTITION BY clause is not a table's partition key?

wolffcm:
That's all my comments.  I think we need some more test cases here.

BillWhite:
I added some more regression tests with mixed partition by and partitioned tables.  I also added some planner tests, but the plans for these will change quite significantly in the future.  Most of the work happens on the coordinator fragment anyway.

wolffcm:
Thanks Bill, I'll have a look when I'm done reviewing Ethan's changes.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
```diff
@@ -238,6 +239,59 @@ private void validateQueryWithSubquery(String windowedQuery) {
...
+        // Validate the coordinator fragment.
+        assertTrue(child instanceof SendPlanNode);
+        child = child.getChild(0);
+        assertTrue(child instanceof ProjectionPlanNode);
+        if (hasStatementOrderBy) {
+            child = child.getChild(0);
+            assertTrue(child instanceof OrderByPlanNode);
+        }
+        child = child.getChild(0);
+        assertTrue(child instanceof PartitionByPlanNode);
```
wolffcm: So we always put the PartitionByPlanNode on the coordinator?  When we're doing PARTITION BY on a table's partition key, then we could do the work on the distributed fragment, and then concatenate the result from all the partitions.

tests/frontend/org/voltdb/planner/testplans-windowingfunctions-ddl.sql
```diff
@@ -0,0 +1,38 @@
...
+);
+
+CREATE TABLE AAA_PA (
+    A INTEGER NOT NULL,
+    B INTEGER NOT NULL,
+    C INTEGER NOT NULL,
+);
+PARTITION TABLE AAA_PA ON COLUMN A;
+
+CREATE TABLE AAA_PB (
```
wolffcm: This table seems unused?

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
```diff
@@ -93,6 +93,14 @@ static private void setupSchema(VoltProjectBuilder project) throws IOException {
...
+                + "  A INTEGER NOT NULL,"
+                + "  AA INTEGER NOT NULL,"
+                + "  B INTEGER NOT NULL,"
+                + "  C INTEGER NOT NULL"
+                + ");\n"
+                + "PARTITION TABLE T_PAA ON COLUMN AA;"
+
                 + "CREATE TABLE T_PB (\n"
                 + "  A INTEGER NOT NULL,"
                 + "  B INTEGER NOT NULL,"
```
wolffcm: What does the table name `T_PB` mean?  Is it supposed to mean that the table is partitioned by column `B`?

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
```diff
@@ -259,7 +272,7 @@ public void testRankWithString() throws Exception {
             assertEquals(expected[rowIdx][colA], vt.getLong(0));
             assertEquals(expected[rowIdx][colB], vt.getLong(1));
             assertEquals(Long.toString(expected[rowIdx][colC], 10), vt.getString(2));
-            assertEquals(expected[rowIdx][colR1], vt.getLong(3));
+            assertEquals(expected[rowIdx][colR_A], vt.getLong(3));
```
wolffcm: Do we have any tests where you do a PARTITION BY a string column?

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
```diff
@@ -337,7 +351,29 @@ public void testRankPartitionedTable() throws Exception {
...
+        sql = "select A, B, C, rank() over (partition by AA, A order by B) as R from T_PAA ORDER BY A, AA, B, C, R;";
+        cr = client.callProcedure("@AdHoc", sql);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        vt = cr.getResults()[0];
+        for (int rowIdx = 0; vt.advanceRow(); rowIdx += 1) {
+            String msg = String.format("Row %d:", rowIdx);
+            assertEquals(msg, expected[rowIdx][colA],    vt.getLong(0));
+            assertEquals(msg, expected[rowIdx][colB],    vt.getLong(1));
+            assertEquals(msg, expected[rowIdx][colC],    vt.getLong(2));
+            assertEquals(msg, expected[rowIdx][colR_AA], vt.getLong(3));
```
wolffcm: Are there any tests where we are selecting from a partitioned table, but the table's partition key does not appear at all in the PARTITION BY clause?  Maybe I'm missing it, but I don't see it here, since T_PC is actually partitioned by column A.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
wolffcm: We do something similar to this when we handle aggregates: if the GROUP BY the partition key, then we can just concatenate the aggregations produced by the distributed fragment on the coordinator.  I'm kind of surprised we're not doing it here.

wolffcm:
That's all my comments.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
BillWhite: I had thought optimizations like this were part of the follow on ticket.

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
BillWhite: I added one of these.

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
BillWhite: Sorry.  T_PC should have been partitioned on C.  I fixed this.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
wolffcm: I see.  Using indexes to avoid sorting and pushing down PartitionByPlanNode into distributed fragments are really two separate optimizations that can be applied independently.  I think it probably needs a separate ticket.

Not executing PartitionByPlanNode in the distributed fragments is not as big a deal as it is for aggregates, but still worth doing.

wolffcm:
The latest commit looks good.

PR #3795, opened 2016-07-28
Eng 10831 better safe

xinjiacs:
Pro: https://github.com/VoltDB/pro/pull/1443


src/frontend/org/voltdb/iv2/InitiatorMailbox.java
```diff
@@ -283,11 +287,11 @@ protected void deliverInternal(VoltMessage message) {
             handleLogRequest(message);
             return;
         }
-        else if (message instanceof Iv2RepairLogResponseMessage) {
+        if (message instanceof Iv2RepairLogResponseMessage) {
```
jhugg: Please put the elses back in here and make it explicit. Sometimes messages use inheritance and it seems possible a future bug could hit two branches.

src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
```diff
@@ -272,7 +273,11 @@ public String toString() {
         else
             sb.append("\n  ROLLBACK/ABORT, ");
         sb.append("\n CLIENT RESPONSE: \n");
-        sb.append(m_response.toJSONString());
+        if (m_response == null) {
```
jhugg: So m_response was never null before and now might be? Or was this a bug fix?

src/frontend/org/voltdb/iv2/InitiatorMailbox.java
```diff
@@ -283,11 +287,11 @@ protected void deliverInternal(VoltMessage message) {
...
             m_algo.deliver(message);
             return;
         }
-        else if (message instanceof RejoinMessage) {
+        if (message instanceof RejoinMessage) {
```
nshi: I thought you removed all the pure style changes from the pull request. All changes in this file are style.

src/frontend/org/voltdb/iv2/RepairLog.java
```diff
@@ -157,58 +156,65 @@ public void deliver(VoltMessage msg)
...
-                m_logSP.add(new Item(IS_SP, m, m.getSpHandle(), m.getTxnId()));
-            }
-        } else if (msg instanceof FragmentTaskMessage) {
-            final FragmentTaskMessage m = (FragmentTaskMessage) msg;
 
-            boolean shortcutRead = m.isReadOnly() && (m_defaultConsistencyReadLevel == ReadLevel.FAST);
+            m_lastSpHandle = m.getSpHandle();
+            truncate(m.getTruncationHandle(), IS_SP);
+            m_logSP.add(new Item(IS_SP, m, m.getSpHandle(), m.getTxnId()));
+            return;
```
nshi: This one, too. Most of the changes in this file are indentation changes caused by a style change.

src/frontend/org/voltdb/AdmissionControlGroup.java
```diff
@@ -154,7 +154,7 @@ public void increaseBackpressure(int messageSize)
         checkAndLogInvariants();
         if (m_pendingTxnBytes > MAX_DESIRED_PENDING_BYTES || m_pendingTxnCount > MAX_DESIRED_PENDING_TXNS) {
             if (!m_hadBackPressure) {
-                hostLog.debug("TXN back pressure began");
+                hostLog.info("TXN back pressure began");
```
nshi: Unintended changes for debugging?

src/frontend/org/voltdb/AdmissionControlGroup.java
```diff
@@ -217,7 +217,7 @@ public void reduceBackpressure(int messageSize)
             (m_pendingTxnCount < LESS_THAN_MAX_DESIRED_PENDING_TXNS))
         {
             if (m_hadBackPressure) {
-                hostLog.debug("TXN backpressure ended");
+                hostLog.info("TXN backpressure ended");
```
nshi: Same here.

src/frontend/org/voltdb/iv2/BufferedReadLog.java
```diff
@@ -0,0 +1,125 @@
...
+import org.voltdb.messaging.FragmentResponseMessage;
+import org.voltdb.messaging.InitiateResponseMessage;
+
+public class BufferedReadLog
+{
+    private static final VoltLogger hostLog = new VoltLogger("HOST");
+
+    public static class Item {
+        InitiateResponseMessage m_initiateMsg;
+        FragmentResponseMessage m_fragmentMsg;
```
nshi: Should be final.

src/frontend/org/voltdb/AdmissionControlGroup.java
xinjiacs: I added this change to run on Jenkins to see, whether the back pressure triggered or not. At least not locally on my runs.

src/frontend/org/voltdb/iv2/BufferedReadLog.java
```diff
@@ -0,0 +1,125 @@
...
+        if (item.getSpHandle() <= handle) {
+            m_mailbox.send(item.getResponseHSId(), item.getMessage());
+        } else {
+            m_bufferedReadSp.add(item);
+        }
+        releaseBufferedRead(handle);
+
+        if (m_bufferedReadSp.size() > m_maxCount) {
+            m_maxCount = m_bufferedReadSp.size();
+            hostLog.rateLimitedLog(10, Level.INFO, null, "max buffered read log size: " + m_maxCount);
```
nshi: RateLimitedLog is not very cheap. I've benchmarked and seen significant overhead when called on each transaction.

src/frontend/org/voltdb/iv2/BufferedReadLog.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public void offerSp(FragmentResponseMessage msg, long handle)
+    {
+        offerSp(new Item(msg), handle);
+    }
+
+    //  SPI offers a new message.
+    private void offerSp(Item item, long handle) {
+        if (item.getSpHandle() <= handle) {
+            m_mailbox.send(item.getResponseHSId(), item.getMessage());
```
nshi: Can it get into this branch when `m_bufferedReadSp` is not empty? If so, this may result in out of order responses.

src/frontend/org/voltdb/iv2/InitiatorMailbox.java
```diff
@@ -64,6 +64,7 @@
     protected final Scheduler m_scheduler;
     protected final HostMessenger m_messenger;
     protected final RepairLog m_repairLog;
+    protected final BufferedReadLog m_bufferedReadlog;
```
nshi: Why is this owned by the InitiatorMailbox instead of the SpScheduler?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -980,6 +985,24 @@ else if (result == DuplicateCounter.MISMATCH) {
...
+                assert(m_isLeader);
+                assert(m_bufferedReadLog != null);
+                if (m_defaultConsistencyReadLevel == ReadLevel.SAFE) {
+                    m_bufferedReadLog.offerSp(message, m_repairLogTruncationHandle);
+                    return;
+                }
+            } else {
+                // writes may update the truncation handle
+                if (m_isLeader && message.getSpHandle() > m_repairLogTruncationHandle) {
+                    m_repairLogTruncationHandle = message.getSpHandle();
```
nshi: The truncation handle is already updated above, I think you can ignore it here.

src/frontend/org/voltdb/iv2/BufferedReadLog.java
```diff
@@ -0,0 +1,125 @@
...
+
+    BufferedReadLog(Mailbox mailbox)
+    {
+        m_bufferedReadSp = new ArrayDeque<Item>(INIT_BUFFER_CAPACITY);
+
+        assert(mailbox != null);
+        m_mailbox = mailbox;
+    }
+
+    public void offerSp(InitiateResponseMessage msg, long handle)
```
nshi: This can probably be renamed to `offer()` since it handles both SP and MP fragments.

src/frontend/org/voltdb/iv2/BufferedReadLog.java
xinjiacs: I just want to know how many reads will be buffered. This may impact the performance of using the default ArrayDeque constructor as capacity 16. 

I agree that this is not important logging, at least should be debug level.

src/frontend/org/voltdb/iv2/SpScheduler.java
xinjiacs: If not running with K-safety, the truncation handle will be updated here then.

src/frontend/org/voltdb/iv2/InitiatorMailbox.java
xinjiacs: Should I do something for node rejoin? I was wondering that. 

Yes, I see that. In the long term, the DuplicateCounter, RepairLog, BufferedReadLog will be merged inside SpScheduler.

xinjiacs: I thought it's more efficient to create the BufferedReadLog in SPI only, but not in replica, where InitiatorMailbox has context of it. 

Anyway, constructing an empty Log is not end of the world.

src/frontend/org/voltdb/iv2/BufferedReadLog.java
```diff
@@ -0,0 +1,120 @@
...
+        offerInternal(mailbox, new Item(msg), handle);
+    }
+
+    //  SPI offers a new message.
+    private void offerInternal(Mailbox mailbox, Item item, long handle) {
+        if (item.getSpHandle() <= handle) {
+            mailbox.send(item.getResponseHSId(), item.getMessage());
+        } else {
+            m_bufferedReads.add(item);
+            releaseBufferedReads(mailbox, handle);
```
nshi: Don't you want to call `releaseBufferedReads()` regardless of whether the current item can be released or not?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -700,15 +698,23 @@ public void handleInitiateResponseMessage(InitiateResponseMessage message)
...
+                // the initiatorHSId is the ClientInterface mailbox.
+                m_mailbox.send(message.getInitiatorHSId(), message);
+                return;
+            }
+
+            if (m_defaultConsistencyReadLevel == ReadLevel.SAFE) {
+                // InvocationDispatcher routes SAFE reads to SPI only
+                assert(m_isLeader);
+                assert(m_bufferedReadLog != null);
+                m_bufferedReadLog.offer(m_mailbox, message, m_repairLogTruncationHandle);
```
nshi: It's safer to use the truncationHandle in the `RepairLog` in case this code is enabled on the replicas.

nshi: It's slightly faster to use the `m_repairLogTruncationHandle` directly on the leader, because it doesn't have to wait for this value to be propagated to the RepairLog. So it's safe for now, but should be careful when we use this variable on the replicas in the future.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1256,9 +1282,16 @@ private long getRepairLogTruncationHandleForReplicas()
 
     private void setRepairLogTruncationHandle(long newHandle)
     {
-        assert newHandle >= m_repairLogTruncationHandle;
+        assert newHandle >= m_repairLogTruncationHandle : "new handle: " + newHandle + ", repairLog:" + m_repairLogTruncationHandle;
         if (newHandle > m_repairLogTruncationHandle) {
             m_repairLogTruncationHandle = newHandle;
+
+            if (! m_isLeader) {
```
nshi: You don't have to check for this here. I think removing this check and changing the following if to the following is sufficient.
```java
if (m_bufferedReadLog != null) {
    m_bufferedReadLog.releaseBufferedReads(m_mailbox, m_repairLogTruncationHandle);
}
```

PR #3796, opened 2016-07-28
Eng 10091 grammar gen

johnTcrawford:


tests/sqlgrammar/DDL.sql
```diff
@@ -0,0 +1,108 @@
+-- Define tables and other DDL, for use with the SQL grammar generator
+
+DROP VIEW V_P1 IF EXISTS;
+DROP VIEW V_P2 IF EXISTS;
```
wolffcm: Curious why you need two DDL files?

wolffcm:
This is very cool.  It's neat to see the internals.

What are pros and cons of this approach vs. sqlcoverage?

Feel free to merge when you're ready.

tests/sqlgrammar/DDL.sql
johnTcrawford: Yeah, I don't really: I'm currently only using the one in the 'sqlgrammar' directory; the stuff, including the DDL file, in the 'sql_coverage' directory is a vestige of my original innovation week on this topic, when I was also trying to pass the generated SQL statements to sqlcoverage; maybe I'll delete those files from the squashed branch, and put them in a new branch associated with ENG-10240.

johnTcrawford:
Pros and cons: SQLCoverage (as it exists now, without input from sql-grammar-gen), even though the data is partly randomized, is very good at testing queries in specific formats that we anticipate as useful and important to test, and at confirming that the results are correct (against HSQL or PostgreSQL) - as well as that they don't crash VoltDB. Sql-grammar-gen is very good at testing queries in any old complicated, crazy format (e.g., joins of sub-queries of sub-queries), and testing whether they crash VoltDB, but not at testing their correctness (though passing the results to SQLCoverage, per ENG-10240, could also do that). One is not a replacement for the other, they are both useful, in different ways.

johnTcrawford:
Closing this very old pull request. I believe a version of these changes was merged to master a long time ago.

PR #3798, opened 2016-07-29
ENG-7938 - New SQLCMD CLI option

jpiekos:
Added a hidden switch to disable the sqlcmd version check/phone home HTTP invocation, per ENG-7938.  Added a new sqlcmd test (copied existing test) to test the new command line switch, but difficult to prove that the http request was or was not invoked, so the test simply validates that the option is allowed.


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -74,6 +74,7 @@
     private static boolean m_stopOnError = true;
     private static boolean m_debug = false;
     private static boolean m_interactive;
+    private static boolean m_version_check = true;
```
zeemanhe: Since the version check is not supported, would it be better to disable the check by default?

wolffcm: Our coding convention (which is obviously not enforced everywhere) is to use camel-case, so this should be `m_versionCheck`.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1338,7 +1343,10 @@ else if ((arg.equals("--usage")) || (arg.equals("-?"))) {
         String[] servers = serverList.split(",");
 
         // Phone home to see if there is a newer version of VoltDB
-        openURLAsync();
+        if (m_version_check)
+        {
```
wolffcm: Another code convention nit: open brace should be on the same line as the IF statement.

wolffcm:
Looks good to me.  Feel free to merge assuming that Jenkins is blue (especially sqlcmdtest).

PR #3799, opened 2016-07-31
ENG-10679: Route reads to SPIs by default.

jhugg:
Offer clientconfig option to round-robin them among replicas like previous behavior.

No real tests :-(, but a pretty trivial change that should show up in perf numbers if right.

### DIFF
--- a/src/frontend/org/voltdb/client/Distributer.java
+++ b/src/frontend/org/voltdb/client/Distributer.java
@@ -85,9 +85,9 @@ class Distributer {
 
     // collection of connections to the cluster
     private final CopyOnWriteArrayList<NodeConnection> m_connections =
-            new CopyOnWriteArrayList<NodeConnection>();
+            new CopyOnWriteArrayList<>();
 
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -862,14 +863,15 @@ void drain() throws InterruptedException {
...
-                false, null);
+                false, false, null);
     }
 
     Distributer(
             boolean useMultipleThreads,
             long procedureCallTimeoutNanos,
             long connectionResponseTimeoutMS,
             boolean useClientAffinity,
+            boolean sendReadsToReplicasBytDefault,
```
akhanzode: Do these distributor changes need to go in other clients especially the one that uses client affinity?

jhugg: I don't think c++ routes reads to replicas?
I don't know about golang. Maybe?

akhanzode: Go does route. I will get a ticket in for this on go side.

PR #3800, opened 2016-07-31
Use pre-computed catalog jar hash when updating classes

wolffcm:
In profiling doing `@UpdateClasses` with a very large number of stored procedures, we were spending a lot of time computing the SHA1 hash for the new catalog jar.  This change makes it so we'll use a precomputed hash when it's available.  

In my tests I saw this decrease the amount of time we spent blocking other work on the cluster (when doing catalog update on the MP and SP threads) by around 40%.

Pro pull request:
https://github.com/VoltDB/pro/pull/1445


src/frontend/org/voltdb/Inits.java
```diff
@@ -353,13 +354,15 @@ public void run() {
             } while (catalogStuff == null || catalogStuff.catalogBytes.length == 0);
 
             String serializedCatalog = null;
-            byte[] catalogJarBytes = catalogStuff.catalogBytes;
+            byte[] catalogJarBytes = null;
+            byte[] catalogJarHash = null;
```
paulmartel: Nice catch -- eliminating the completely ineffectual/misleading assignment.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2070,12 +2071,17 @@ private static ByteBuffer makeCatalogVersionAndBytes(
...
-            versionAndBytes.put((new InMemoryJarfile(catalogBytes)).getSha1Hash());
+        if (catalogHash != null) {
+            versionAndBytes.put(catalogHash);
         }
-        catch (IOException ioe) {
-            VoltDB.crashLocalVoltDB("Unable to build InMemoryJarfile from bytes, should never happen.",
-                    true, ioe);
+        else {
+            try {
+                versionAndBytes.put((new InMemoryJarfile(catalogBytes)).getSha1Hash());
```
paulmartel: I'd rather see a conditional reassignment of catalogHash:

    if == null
        try {
           catalogHash =  new InMemoryJarfile(catalogBytes)).getSha1Hash();

followed by unconditional

            versionAndBytes.put(catalogHash);

paulmartel:
OK to merge this with or without suggested change -- AND pro changeset as well.

BillWhite:
I looked through this.  It seems right to me.

PR #3801, opened 2016-08-01
ENG-10871: fix newCLI background mode (-B)

JadenW:
simply convert all args to string type for starting as daemon.

Otherwise need to change hostcount option in "start.py" verb.


akhanzode:
Looks good. merge it.

PR #3804, opened 2016-08-02
ENG-10430 sqlcmd echo and echoerror directives

paulmartel:


wolffcm:
Maybe add a sqlcmdtest for the stdout version of `echo` as well?

Looks good otherwise.

src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -432,6 +432,20 @@ private Exception(Throwable cause, String message, Object... args)
             "([^;\\s]*)" +    // optional subcommand (group 2)
             InitiallyForgivingDirectiveTermination,
             Pattern.CASE_INSENSITIVE);
+    private static final Pattern EchoToken = Pattern.compile(
+            "^\\s*" +         // optional indent at start of line
+            "echo" +          // required ECHO command token
+            "(\\W|$)" +       // require an end to the keyword OR EOL (group 1)
```
wolffcm: I am surprised that this works since there are two `$` characters in this pattern.  If the first group matches EOL, then what will the second group be equal to---I guess in that case it's null or the empty string?  In any case your code is defensive towards this, I'm just curious.

PR #3805, opened 2016-08-02
Eng 10546 sqlcov numeric squash

johnTcrawford:


src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -114,10 +114,15 @@
             = new QueryTransformer(dayOfYearQuery)
             .initialText("EXTRACT ( ").prefix("DOY FROM").suffix(")").groups("column");
 
+    // Regex pattern for a typical column or column expression (including
+    // functions, operators, etc.),
+    private static final String COLUMN_PATTERN = "(\\s*\\w*\\s*\\()*\\s*(\\w+\\.)?(?<column>\\w+)(\\s+(AS|FROM)\\s+\\w+)?(\\s*\\))*"
+            + "\\s*((\\+|\\-|\\*|\\/)(\\s*\\w*\\s*\\()*\\s*(\\w+\\.)?\\w+(\\s+(AS|FROM)\\s+\\w+)?(\\s*\\))*)*\\s*";
```
wolffcm: Could you expand this comment to describe more specifically what this matches?  I am surprised that `AS` and `FROM` are allowed here.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -114,10 +114,15 @@
...
+
     // Captures the use of AVG(columnName), which PostgreSQL handles
     // differently, when the columnName is of one of the integer types
     private static final Pattern avgQuery = Pattern.compile(
-            "AVG\\s*\\((\\s*\\w*\\s*\\()*\\s*(\\w+\\.)?(?<column>\\w+)(\\s*\\)(\\s+(AS|FROM)\\s+\\w+)?)*\\s*\\)",
+            "AVG\\s*\\("+COLUMN_PATTERN+"\\)",
```
wolffcm: Seems like `AS` and `FROM` keywords are never allowed in this context, so COLUMN_PATTERN is too broad here?

Perhaps for your purposes you understand the input that it will be matching so it's not an issue, but it seems confusing to read.

tests/scripts/sql_coverage_test.py
```diff
@@ -225,12 +229,9 @@ def get_max_mismatches(comparison_database, suite_name):
...
               config_name == 'basic-compoundex-joins'):
             max_mismatches = 4620
-        # Known failures in the numeric-decimals and numeric-ints "extended"
-        # test suites (see ENG-10546)
+        # Known failures in the numeric-decimals "extended" test suite (see ENG-10546)
         elif config_name == 'numeric-decimals':
-            max_mismatches = 1180
-        elif config_name == 'numeric-ints':
-            max_mismatches = 2820
+            max_mismatches = 300
```
wolffcm: Are these mismatches that may occur after things have been normalized in order to minimize mismatches?

It seems like for any place where you're allowing more than 0 mismatches you'll want a comment explaining why and a ticket number to fix it.

tests/scripts/sql_coverage_test.py
```diff
@@ -241,10 +242,15 @@ def run_config(suite_name, config, basedir, output_dir, random_seed,
     # Store the current, initial system time (in seconds since January 1, 1970)
     time0 = time.time()
 
+    global precision
```
wolffcm: I see you've introduced a new global here.  Could it be included in the other global (`normalize`)?  There may be a need for some globals, but this one seems unnecessary.

wolffcm:
I like that you've deleted a lot of code, which is nice to see.  Could you make a comment on this pull request about how you were able to delete all this code?

wolffcm:
That's all my comments.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: I have added (& will check in shortly) this to the comment:
"- currently only used for AVG, though this
    // could be expanded in the future; note that AS or FROM can occur
    // (rarely) if certain functions occur within the AVG function, e.g.,
    // AVG(CAST(VCHAR AS INTEGER)) or AVG(EXTRACT(DAY FROM PAST))"

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: In this comment (& below), I've changed AVG(columnName) to AVG(columnExpression) - and see my comment above. I cannot recall whether I added the "(AS|FROM)" stuff from an overabundance of caution or because certain tests actually fail without it. (Note that that stuff was already in there before the current changes - which definitely are needed to avoid failures.)

tests/scripts/sql_coverage_test.py
johnTcrawford: Agreed, but there already is such a comment. The change here is to eliminate the allowed mismatches for the 'numeric-ints' test suite (which is now fixed) and reduce from 1180 to 300 the allowed mismatches for the 'numeric-decimals' test suite (which is now 2/3 fixed, meaning that the third category of failures described in that ticket are not yet fixed - which I may break off into a separate ticket).

tests/scripts/sql_coverage_test.py
johnTcrawford: I don't see how it could be incorporated into 'normalize', which is an external function, but I have changed it to be an argument to the run_once function, since that is the only place it is used (well, it is passed from there to the 'normalize' function, so that part is unchanged).

johnTcrawford:
Chris, regarding deleting a lot of code: by introducing the 'precision' as an argument in the config files, I was able to delete several normalizers that are no longer needed (fuzzynormalizer.py, geo-normalizer.py, geo-normalizer-4digits.py; nulls-first-normalizer.py, nulls-highest-normalizer.py, nulls-last-normalizer.py), either because their only purpose was to use a non-default precision, or they specified a NULL-sorting that was never used. The remaining normalizers (StandardNormalizer.py, not-a-normalizer.py, normalizer.py, nulls-lowest-normalizer.py) were updated to use (or ignore, in one case) the 'precision' argument.

tests/scripts/sql_coverage_test.py
wolffcm: Ah, now I see that there is.  The diff wasn't showing enough context.

wolffcm:
Makes sense.  This looks good to me.

PR #3806, opened 2016-08-03
ENG-10864 view warnings

yzhang1991:
Improve the warning logic for single table views.
Add warnings for join query views.

The ad-hoc DDL warning tests were filed as [ENG-10894](https://issues.voltdb.com/browse/ENG-10894), which will be added after Paul's fix for the warning messages lands in the product.


wolffcm:
For the case of joins with views, isn't it better to warn if there are any instances of NestLoopPlanNode (as opposed to NestLoopIndexPlanNode) in the plan?  If there are joins being evaluated without an index there could be quadratic performance, or worse.  

One can imagine a plan which contains several NestLoopPlanNodes, but contains an index scan on the leftmost table.  Such a plan would not produce a warning with your proposed change, but could still have terrible performance.

wolffcm:
That's all my comments.

wolffcm:
This looks good.  Thanks for making the change.

PR #3807, opened 2016-08-03
ENG-10472: Transaction commit notification

nshi:


src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -962,8 +962,12 @@ public void handleFragmentResponseMessage(FragmentResponseMessage message)
         if (counter != null) {
             int result = counter.offer(message);
             if (result == DuplicateCounter.DONE) {
+                final TransactionState txn = m_outstandingTxns.get(message.getTxnId());
+                if (txn != null && txn.isDone()) {
```
xinjiacs: I see you schedule the RepairLogTruncationMsg in `handleCompleteTransactionMessage `, why still need it here in FragmentResponseMessage?

Looks like for MP txn, we only update the DONESS of the txn state when the site generating InitiateResponseMessage, which is the last step. Is this for MP SAFE READs only?

PR #3808, opened 2016-08-04
Add trig functions to VoltDB SQL

wolffcm:
The first commit is what we got from our contributor.

The last two commits (notably the last one) are changes I made for more thorough testing and handling of nonfinites.  I moved the JUnit tests into TestFunctionsSuite so I could use the existing framework for testing scalar functions, which was handy.


tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -1721,11 +1726,55 @@ public void subtestSqrt() throws Exception
...
+        final Set<Double> filters = null;
+
+        Map<String, DoubleFunction<Double>> funcInfo = new HashMap<>();
+        funcInfo.put("SIN", x -> Math.sin(x));
+        funcInfo.put("COS", x -> Math.cos(x));
+        funcInfo.put("TAN", x -> Math.tan(x));
+        funcInfo.put("COT", x -> 1.0 / Math.tan(x));
+        funcInfo.put("SEC", x -> 1.0 / Math.cos(x));
+        funcInfo.put("CSC", x -> 1.0 / Math.sin(x));
+
```
BillWhite: Hmm.  Lambdas.  Wish we could use these in C++.

BillWhite:
This all seems straightforward to me.
 End of review round.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
yzhang1991: Curious if this is compilable on CentOS.
I remember I had a commit before using `auto` that can work on CentOS, but `for ( a : collection )` cannot because we are using "c++0x" instead of "c++11" on CentOS.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -506,11 +509,31 @@ private FunctionCustom(int id) {
             case FUNC_ASIN :
             case FUNC_ATAN :
             case FUNC_ATAN2 :
+            case FUNC_CSC :
```
yzhang1991: Here, if I execute `select asin(a) from t;`, I will get the following error message which I think is misleading:
>VOLTDB ERROR: UNEXPECTED FAILURE:
  Internal Error: SQL function 'csc' with ID (102) with (1) parameters is not implemented in VoltDB (or may have been incorrectly parsed)

Same for `acos()`, `atan()`, `atan2` (what is atan2?). 
I think it would be great if you can try to give a clearer error message for those unimplemented functions.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -506,11 +509,31 @@ private FunctionCustom(int id) {
...
+                break;
             case FUNC_COS :
+                name = Tokens.T_COS;
+                parseList = singleParamList;
+                break;
             case FUNC_COT :
+                name = Tokens.T_COT;
+                parseList = singleParamList;
+                break;
             case FUNC_DEGREES :
```
yzhang1991: >10> select degrees(a) from t;
VOLTDB ERROR: UNEXPECTED FAILURE:
  Internal Error: SQL function 'sin' with ID (107) with (1) parameters is not implemented in VoltDB (or may have been incorrectly parsed)

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -941,6 +964,20 @@ Object getValue(Session session, Object[] data) {
 
                 return Double.valueOf(java.lang.Math.cos(d));
             }
+            case FUNC_CSC : {
```
yzhang1991: I am confused here, why do we need a java implementation here?

yzhang1991: You can ignore this comment, we do this for HSQL backend.
FunctionForVoltDB does not implement this method, do we have a list to decide which function shall we put into tests running against HSQLbackend? Are all our geo functions not being tested against HSQL backend?

yzhang1991:
That's all my comments.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
wolffcm: This is Java code, and I believe we're on Java 8 everywhere now, so it should be okay.

I've found too that we can use `auto` in C++ but that for range-based for loops we're not there yet.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
wolffcm: Good catch, I will investigate.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
wolffcm: That's right, we didn't bother implementing geo functions in Java, which in a few cases would have required importing the Java version of S2.  

We do verify our answers against PostGIS in sqlcoverage though, so we do have wrong answer testing.

PR #3809, opened 2016-08-04
ENG-10896: Only register undo action for BEGINTXN when one has been written to the stream

slowsoul:
Make transactionChecks() return a bool indicating if stream state was switched from close to open in this function call.

Register an undo action for BEGINTXN only if it has been written to stream (i.e. transactionChecks() returns true)


luyangco:
LGTM.

PR #3810, opened 2016-08-04
ENG-10816: Close stale DR producer connections when new connection comes

slowsoul:
Add method to VoltNetworkPool to get all connections an instance is managing


nshi:
Review finished.

PR #3812, opened 2016-08-05
Long running read with suspendable sequential scan

rharding6373:
The first 4 commits integrate feedback from the last round of code review on the long_running_reads branch and add a utility function to read volt tables from files produced by the EE. The 5th commit contains the changes to the EE that allow sequential scans to suspend execution and the EE to save a transaction's state so that OLTP transactions can be interleaved with a long running read query.


src/ee/common/SavedContext.hpp
```diff
@@ -0,0 +1,132 @@
...
+namespace voltdb {
+class ExecutorContext;
+class VoltDBEngine;
+class TempTable;
+
+/*
+ * Saved state from the ExecutorContext and VoltDBEngine
+ * Used to keep the state of paused transactions
+ */
+class SavedContext {
```
wolffcm: I am just thinking aloud, but I wonder if this is a bit of encapsulation that we might want to for normal queries as well as pausable ones.  It seems like we store a lot of per-fragment context in VoltDBEngine and ExecutorContext, and neither of those classes seem like the right place for that stuff.  (No action item here, just thinking about possible refactorings to make things easier moving forward.)

src/ee/common/executorcontext.cpp
```diff
@@ -128,30 +132,40 @@ ExecutorContext* ExecutorContext::getExecutorContext() {
...
         BOOST_FOREACH (AbstractExecutor *executor, executorList) {
             assert(executor);
+            if (m_isResumed && ctr < m_executedCtr) {
+                ++ctr;
+                continue;
+            }
             // Call the execute method to actually perform whatever action
             // it is that the node is supposed to do...
             if (!executor->execute(*m_staticParams)) {
+                if (executor->isSuspendable()) {
```
wolffcm: There's a few verbs associated with this concept: pause, suspend, yield.  Might be good to choose just one for easier grepping going forward.

src/ee/common/executorcontext.hpp
```diff
@@ -237,13 +249,16 @@ class ExecutorContext {
         std::pair<std::map<int, SubqueryContext>::iterator, bool> result =
 #endif
             m_subqueryContextMap.insert(std::make_pair(subqueryId, fromCopy));
+#ifdef DEBUG
         assert(result.second);
```
wolffcm: There's no need to wrap this in a preprocessor guard, since `assert` has a macro to become a no-op in the release build.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -244,6 +247,8 @@ VoltDBEngine::initialize(int32_t clusterIndex,
                                             m_drReplicatedStream,
                                             drClusterId,
                                             highVolumeDirPath);
+    m_savedContext = new SavedContext();
```
wolffcm: Is there only ever one of these per EE?  If so you could make it a direct member of VoltDBEngine, rather than a pointer to an object that needs to be explicitly allocated.

src/ee/executors/seqscanexecutor.cpp
```diff
@@ -103,6 +104,10 @@ bool SeqScanExecutor::p_init(AbstractPlanNode* abstract_node,
     // Inline aggregation can be serial, partial or hash
     m_aggExec = voltdb::getInlineAggregateExecutor(node);
 
+    m_highVolume = node->isPauseable();
+
+    m_limit = node->getLimit();
```
wolffcm: A better name would be good here.  Scan nodes can have an inlined limit node, so there would be some confusion about which limit this is.  Maybe, `tupleLimitForPausableFragments`?  That is just a suggestion.

src/ee/storage/SnapshotContext.h
```diff
@@ -36,8 +36,9 @@ class TempTable;
 class ParsedPredicate;
 class TupleOutputStreamProcessor;
 class PersistentTableSurgeon;
+class ScanCopyOnWriteContext;
 
-class CopyOnWriteContext : public TableStreamerContext {
+class SnapshotContext : public TableStreamerContext {
```
wolffcm: Would it make sense for `ScanCopyOnWriteContext` to also be a subclass of `TableStreamerContext`?  Why or why not?

wolffcm:
That's all my comments on the EE part of the last commit.  I'll have a look at the Java part tomorrow.

I wonder if it makes sense for PersistentTable to have a vector of COW contexts (both for snapshots and for LRRs), which would simplify some of the logic in the storage layer.  This gets back to my comment about scan COW contexts and snapshot COW contexts maybe having a common base class.

There's some gaps in the kinds of plans that can be handled here, like we talked about yesterday.   However, my main concern is that you have a foundation from which to try out different implementations of COW indexes, and I think what you have here definitely achieves that.

src/ee/execution/VoltDBEngine.cpp
rharding6373: There's currently only one. Thinking ahead a bit, to support multiple LRRs, we will need to have as many SavedContexts as LRRs. If we only allow a certain number of LRRs, it should be possible to avoid explicit allocation. Otherwise, explicit allocation is unavoidable, right?

src/ee/storage/SnapshotContext.h
rharding6373: One reason not to is because a CoW used to fulfill a LRR does not stream tables in the same way that the other TableStreamers do (e.g., it does not need handleStreamMore(), the code path for reading out tuples is very different). On the other hand, it shares a lot in common with other parts of the interface (e.g., notifyTupleUpdate(), the code path for modifying the table is very similar). I agree that making the CoWs subclasses of TableStreamerContext and use the TableStreamerInterface could simplify a lot of logic, and I've gone back and forth on the idea several times. Do you think it makes sense to make it a subclass?

wolffcm: I'm not sure that I understand this pat of the code as well as you at this point.  But I can envision a future where each PersistentTable has a dynamic list of objects that implement a TableListener interface, and when DML happens to the table, the PersistentTable talks to the listeners only via this interface.  (Seems like materialized views could be maintained in this way too.)

Differences between CoW contexts for scans and snapshots could then be addressed in their respective implementations.

This could simplify code in PersistentTable and make it easier to understand and extend.

There could be performance implications too, but this is even less clear to me.

src/ee/execution/VoltDBEngine.cpp
wolffcm: Yes, that's true.  I guess whenever I see `new` called nowadays I think... "could the dynamic allocation be avoided, or could a smart pointer be used instead?"

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1673,6 +1673,11 @@ public  boolean createTransaction(
                 isShortCircuitRead = true;
             }
         }
+        boolean isReadOnlySlow = invocation.getProcName().equals("@ReadOnlySlow");
+        // Force read only slow queries to not use replicas
+        if (isReadOnlySlow) {
+            isShortCircuitRead = false;
```
wolffcm: Is a short-circuited read when you can consult a local replica to complete a RO transaction?  Why can't you use short-circuiting for LRRs?

src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -207,59 +207,59 @@ synchronized AdHocPlannedStatement planSql(String sqlIn, StatementPartitioning p
...
-								ParameterSet params = null;
-								if (planner.compiledAsParameterizedPlan()) {
-									params = planner.extractedParamValues(core.parameterTypes);
-								} else if (hasUserQuestionMark) {
-									params = ParameterSet.fromArrayNoCopy(userParams);
-								} else {
-									// No constants AdHoc queries
-									params = ParameterSet.emptyParameterSet();
-								}
+                    parsedToken = planner.parameterize();
```
wolffcm: We had tabs in this file?  I wonder how it got past licensecheck.

wolffcm: Did anything change in this file? I don't see it.

src/frontend/org/voltdb/iv2/ProcedureTask.java
```diff
@@ -130,6 +130,12 @@ protected InitiateResponseMessage processInitiateTask(Iv2InitiateTaskMessage tas
                 if ((!task.isReadOnly()) && task.isSinglePartition()) {
                     m_txnState.storeResults(cr);
                 }
+                if(task.getStoredProcedureName().equals("@ReadOnlySlow")) {
```
wolffcm: Is this executed just for the initial slice of a LRR?  If it happens to complete in the very first slice, is this code correct?

src/frontend/org/voltdb/sysprocs/ReadOnlySlow.java
```diff
@@ -19,29 +19,45 @@
...
 import org.voltdb.planner.ActivePlanRepository;
+import org.voltdb.planner.CorePlan;
+import org.voltdb.utils.CatalogUtil;
 
 import com.google_voltpatches.common.base.Charsets;
 
 
 public class ReadOnlySlow extends VoltSystemProcedure {
 
+    private Set<String> pauseableTableNames = new HashSet<String>();
```
wolffcm: "Pauseable" suggests that the pausing is optional.  A better name here... maybe cowTables?  I am not crazy about that either.

tests/ee/storage/CompactionTest.cpp
```diff
@@ -121,7 +127,9 @@ class CompactionTest : public Test {
 
     ~CompactionTest() {
         delete m_engine;
-        delete m_table;
+        if (m_table != NULL) {
```
wolffcm: Deleting a null pointer in C++ is defined as a no-op, so the the guard is unnecessary.

tests/frontend/org/voltdb/regressionsuites/TestLongRunningReadQuery.java
```diff
@@ -24,70 +24,200 @@
...
+        sql = "SELECT * FROM R1;";
+        executeLRR(client,sql);
+        while (!m_receivedResponse) {
+            // Do random ad hoc queries until the long running read returns
+            doRandomTableManipulation(client, "R1");
+
+        }
+        //System.out.println(initTableSize + " " + m_insertCount + " " + m_deleteCount);
+        VoltTable vt = client.callProcedure("@AdHoc",sql).getResults()[0];
+        assertEquals(initTableSize + m_insertCount - m_deleteCount,vt.getRowCount());
```
wolffcm: You could also assert that m_insertCount, etc are greater than zero, ensuring that the table was actually modified.

tests/frontend/org/voltdb/regressionsuites/TestLongRunningReadQuery.java
```diff
@@ -24,70 +24,200 @@
...
+        m_receivedResponse = false;
 
         sql = "SELECT * FROM R1;";
-        files = client.callProcedure("@ReadOnlySlow", sql).getResults()[0];
-        vt = LRRHelper.getTableFromFileTable(files);
-        assertEquals(tableSize,vt.getRowCount());
+        executeLRR(client,sql);
+        executeLRR(client,sql);
+        while (!m_receivedResponse) {
+            // Wait for LRR to finish
```
wolffcm: You could sleep in this loop to avoid spiking the CPU.

wolffcm:
That's all my comments.  I've learned a bunch by reviewing this.

src/frontend/org/voltdb/iv2/ProcedureTask.java
rharding6373: This piece of code just sets variables in the response to be passed up to the SPScheduler, in case the transaction does need to be suspended. Perhaps a name like prepareLRRResponse() would be more descriptive than setPaused().

src/ee/storage/SnapshotContext.h
rharding6373: Based on your comments I'll make ScanCopyOnWriteContext a subclass of TableStreamerContext and be accessible via the TableStreamerInterface and clean up PersistentTable.

src/frontend/org/voltdb/compiler/PlannerTool.java
rharding6373: It looks like the tabs were introduced in one of the intermediate commits.

build.xml
```diff
@@ -1546,7 +1546,7 @@ NATIVE EE STUFF
     description="Run testcases for C++ JNI library.">
     <exec dir='.' executable='python' failonerror='true'>
         <env key="TEST_DIR" value="${build.testobjects.dir}" />
-        <env key="EETESTSUITE" value="${eetestsuite}"/>
+        <env key="EETESTSUITE" value="storage"/>
```
wolffcm: Did you mean to hardcode this value here, or just set it temporarily?

rharding6373: This is an outdated diff. It should be fixed now after fb7c240.

wolffcm:
Looks good to me, thanks for making the changes.

PR #3813, opened 2016-08-05
allow inherited SQLStmts and Multiple inheritance

jankeirse:
- Allow extending SQLStmt so that a method that expects a specific SQLStmt as input can force compile time validation to avoid the wrong SQLStmt being passed in. 
- Also check interfaces for SQLStmts, this allows using multiple inheritance through interfaces with default implementations to avoid one enormous superclass when many stored procedures share some logic but not all stored procedures share the same methods.

To use the multiple inheritance part you need to define the SLQStmts in the interface as statics, I haven't thoroughly tested yet if this has undesirable effects.

### DIFF
--- a/src/frontend/org/voltdb/compiler/ProcedureCompiler.java
+++ b/src/frontend/org/voltdb/compiler/ProcedureCompiler.java
@@ -101,7 +101,7 @@ public abstract class ProcedureCompiler implements GroovyCodeBlockConstants {
         Field[] fields = procClass.getDeclaredFields();
         for (Field f : fields) {
             // skip non SQL fields
-            if (f.getType() != SQLStmt.class)
+            if (!SQLStmt.class.isAssignableFrom(f.getType()))
                 continue;
 
             int modifiers = f.getModifiers();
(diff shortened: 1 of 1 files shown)


jankeirse:
You may probably also want to provide an interface IVoltProcedure to inherit the interfaces from to make the default implementations pretty. I have now put this in my own package that contains my stored procedures.

`public interface IVoltProcedure {
    
    
    public void voltQueueSQL(SQLStmt stmt, Expectation expectation, Object[] args) ;

    public void voltQueueSQL(SQLStmt stmt, Object... args) ;

    public VoltTable[] voltExecuteSQL() ;

    public VoltTable[] voltExecuteSQL(boolean isFinalSQL) ;

    
}`

benjaminballard:
Hi Jan,

Thanks for submitting this.  It looks very good.

Please submit a signed Contributor Agreement (https://voltdb.com/contributor-license-agreement) to support@voltdb.com, before we can code review, test and merge this.

Thanks,
Ben

jankeirse:
@besquared I have signed and submitted the agreement in August 2016. We ended up using another solution instead of voltdb so I didn't follow up on this.

benjaminballard:
@jankeirse - I can confirm you did sign the CLA and it was fully executed. I'm not sure what happened, but this fell through the cracks. I'm creating a new ticket for someone in Engineering to review this. We will probably need to close this PR against master and move the commit to a branch for testing.

PR #3814, opened 2016-08-05
ENG-10901: Wrap debug log messages in SynchronizedStatesManager with debug enable check

slowsoul:


nshi:
We are going to land this on the integration branch.

PR #3815, opened 2016-08-05
Fix ENG-10916

yzhang1991:
Streams support views as long as the partition column is part of the view. However, for view definitions without any group by columns, there is no way you could have a partition column in the view. (aggregation on a partition column does not count!)

That is, **you cannot define a view without a group by column on a streamed table**.

There was an initializeTupleHavingNoGroupBy() function in the persistent view logic to handle the initialization for views without group-by column because there will always be one row even the source table is empty. This call was persistent view specific when I fist added it in ENG-7872 last year.

During a recent refactoring of the old code, we pulled this logic from persistent view to a common code path shared by both types of views. This accidentally invalidated a logic the persistent view had for triggering the data catching-up process.

This fix put the logic back to track. Related tests will be checked in along with ENG-10867 to avoid conflicts.


paulmartel:
Looks good.

PR #3817, opened 2016-08-08
ENG-10758: warnings if ulimit or JVM settings are insufficient

cyx11:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3031,6 +3032,8 @@ else if (m_catalogContext.catalogVersion != expectedCatalogVersion) {
 
                 checkHeapSanity(MiscUtils.isPro(), m_catalogContext.tables.size(),
                         (m_iv2Initiators.size() - 1), m_configuredReplicationFactor);
+//
+                checkThreadsSanity();
```
JadenW: nitpick: remove "//"

src/frontend/org/voltdb/importer/ImportProcessor.java
```diff
@@ -138,24 +143,33 @@ public void addProcessorConfig(ImportConfiguration config) {
...
         } catch(Throwable t) {
             m_logger.error("Failed to configure import handler for " + bundleJar, t);
             Throwables.propagate(t);
         }
     }
 
     @Override
+    public int getPartitionsCount() {
+        if (m_wrapper != null) {
+            return m_wrapper.getConfigsCount();
```
JadenW: Why only return count for m_wrapper? 
Don't you need add the count for all the wrap in m_bundles?

This also works for class-based importers, (i.e. socket importer), right?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3785,6 +3788,59 @@ static public long computeMinimumHeapRqt(boolean isPro, int tableCount, int site
...
+        int expThreadsCount = computeThreadsCount(tableCount, partitions, replicates, importPartitions, exportTableCount, exportNonceCount);
+        String[] command = {"bash", "-c" ,"ulimit -u"};
+        String cmd_rst = ShellTools.local_cmd(command);
+        int maxThreadsCount = 0;
+        try {
+            maxThreadsCount = Integer.parseInt(cmd_rst.substring(0, cmd_rst.length() - 1));
+        } catch(Exception e) {
+            maxThreadsCount = Integer.MAX_VALUE;
+        }
+
```
JadenW: maxThreadsCount properly won't change in the runtime. 
I think we can compute this early and cache it.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3785,6 +3788,59 @@ static public long computeMinimumHeapRqt(boolean isPro, int tableCount, int site
...
+        return importerBaseCount + importerPartitionCount;
+    }
+
+    private int computeExporterThreads(int exportTableCount, int partitionCount, int replicateCount, int exportNonceCount) {
+        if (exportTableCount == 0) {
+            return 0;
+        }
+        int exporterBaseCount = 1;
+        return exporterBaseCount + partitionCount * exportTableCount + exportNonceCount;
+    }
```
JadenW: Not very sure about the formulation here. Is this a upper bound estimation? If the export table is not in runevery-where mode,  only master partitions are going to export  (thus create the threads). 
I saw the replicateCount pass in,  but not used?

src/frontend/org/voltdb/RealVoltDB.java
cyx11: user may update ulimit settings during the runtime. We need to check most up-to-date setting.

PR #3819, opened 2016-08-09
versioncheck.config: added the '--stop-on-error=false' option (to the

johnTcrawford:
'--no-version-check' option), so that the test script will run to the end, even after the expected error, and therefore will clean up after itself, deleting the tables & procedures it creates (which were causing other tests to fail); hence, had to add additional text to versioncheck.outbaseline, for the part after the error; also, deleted the initial WARN, which gets "cleaned" away (which was causing this test to fail).
load_remove_classes.in: also, added drop (if exists) for the my_table_1, my_table_2 & my_table_3 tables, which theoretically could cause this test to fail (though I've not seen this happen); and added corresponding text to load_remove_classes.outbaseline.

### DIFF
--- a/tests/sqlcmd/baselines/ddl/load_remove_classes.outbaseline
+++ b/tests/sqlcmd/baselines/ddl/load_remove_classes.outbaseline
@@ -17,6 +17,15 @@ Command succeeded.
 drop table dropable if exists;
 Command succeeded.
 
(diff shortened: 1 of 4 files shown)


johnTcrawford:
Note: reviewed by Rachael (she sent email, did not post a comment here in github).

PR #3822, opened 2016-08-09
N-partition query scheduling and Debit/Credit Benchmark [Do not merge]

rharding6373:
Adds a debit/credit benchmark to voltdb examples. Allows multiple multi-partition queries to be scheduled to different site masters. The UI is a @AdHoc_NP system stored procedure that accepts a SQL query and a JSON string containing a list of keys accessed by the query and their type.


src/frontend/org/voltdb/iv2/MpScheduler.java
```diff
@@ -329,8 +345,10 @@ public void handleIv2InitiateTaskMessage(Iv2InitiateTaskMessage message)
     private boolean isNpTxn(Iv2InitiateTaskMessage msg)
     {
         return msg.getStoredProcedureName().startsWith("@") &&
-                msg.getStoredProcedureName().equalsIgnoreCase("@BalancePartitions") &&
-                (byte) msg.getParameters()[1] != 1; // clearIndex is MP, normal rebalance is NP
```
nshi: OMG, you reminded me of the horrible hack I made here!! This is terrible!

examples/stored-value-cards/client/debitcredit/DebitCreditBenchmark.java
```diff
@@ -0,0 +1,571 @@
+/* This file is part of VoltDB.
```
jhugg: What's with these licenses? Why three of them?

examples/stored-value-cards/procedures/debitcredit/Transfer.java
```diff
@@ -0,0 +1,136 @@
...
+                                                  " available_balance = available_balance + ?," +
+                                                  " last_activity = ?" +
+                                                  " WHERE pan = ?;");
+
+    public final SQLStmt insertActivity = new SQLStmt("INSERT INTO card_activity VALUES (?,?,?,?,?);");
+
+
+    public long run( String        from_pan,
+                     String        to_pan,
+             double        amount,
```
jhugg: Some whacky whitespace.

src/frontend/org/voltdb/iv2/MpTransactionTaskQueue.java
```diff
@@ -181,28 +229,55 @@ private boolean taskQueueOffer()
...
+
+            while (task != null &&
+                    ((!task.getTransactionState().isReadOnly() && !hasReadSiteConflicts(task) && !hasWriteSiteConflicts(task) && m_updateSitePool.canAcceptWork()) ||
+                    (task.getTransactionState().isReadOnly() && !hasWriteSiteConflicts(task) && m_sitePool.canAcceptWork()))
+                    ) {
+                task = m_backlog.pollFirst();
+                if (!task.getTransactionState().isReadOnly()) {
+
                     m_currentWrites.put(task.getTxnId(), task);
+                    if (task instanceof MpProcedureTask) {
```
nshi: Do you mean `NpProducerTask` here?

src/frontend/org/voltdb/iv2/MpTransactionTaskQueue.java
```diff
@@ -181,28 +229,55 @@ private boolean taskQueueOffer()
...
-            else if (m_currentWrites.isEmpty()) {
-                while (task != null && task.getTransactionState().isReadOnly() &&
-                       m_sitePool.canAcceptWork())
-                {
-                    task = m_backlog.pollFirst();
+                else {
+
                     assert(task.getTransactionState().isReadOnly());
                     m_currentReads.put(task.getTxnId(), task);
+                    if (task instanceof MpProcedureTask) {
```
nshi: Same here.

src/frontend/org/voltdb/iv2/MpTransactionTaskQueue.java
```diff
@@ -218,13 +293,37 @@ synchronized int flush(long txnId)
     {
         int offered = 0;
         if (m_currentReads.containsKey(txnId)) {
+            TransactionTask task = m_currentReads.get(txnId);
+            Iterator<Integer> iter;
+            if (task instanceof MpProcedureTask) {
```
nshi: Here, too.

src/frontend/org/voltdb/iv2/MpTransactionTaskQueue.java
```diff
@@ -218,13 +293,37 @@ synchronized int flush(long txnId)
...
+                m_currentReadSites.remove(iter.next(),txnId);
+            }
             m_currentReads.remove(txnId);
             m_sitePool.completeWork(txnId);
         }
         else {
             assert(m_currentWrites.containsKey(txnId));
+            TransactionTask task = m_currentWrites.get(txnId);
+            Iterator<Integer> iter;
+            if (task instanceof MpProcedureTask) {
```
nshi: And here.

src/frontend/org/voltdb/iv2/MpTransactionTaskQueue.java
rharding6373: NpProcedureTask extends MpProcedureTask, so NP tasks go through this code path, too. The alternate code path is for multi-part system procedures such as catalog update which don't track the IDs of the partitions they need.

nshi: This block is able to handle the case where the involved partitions are either all partitions (MP) or a subset of all the partitions (NP). My confusion actually came from the else block, where it only handles the all partition case. Isn't the else block redundant?

nshi: In fact, will it ever enter the else block?

rharding6373: It enters the else block when scheduling a system procedure that needs to run on every partition and is therefore scheduled using the MPI (e.g., catalog update). These don't provide which partitions the task should run on, unlike MP/NP transactions, presumably because it's inferred that they should run on every site. So there's a separate code path to catch this case that conservatively marks all sites as running a write or read. 

MP transactions could go through the else branch, too, but checking if a task is NP is a bit tricky, since that class is pro.

nshi: Ah, yes, you are right. I forgot about the run-everywhere sysprocs also go through this path. That makes sense now.

PR #3823, opened 2016-08-10
Add Copyright generation to EE test generator to satisfy license check.

paulmartel:
For review only.
Before merging to master, any generated EE test files currently checked into github should be re-generated based on the new template introduced here. I don't yet know how to do that.

### DIFF
--- a/tests/frontend/org/voltdb/planner/PlannerTestCase.java
+++ b/tests/frontend/org/voltdb/planner/PlannerTestCase.java
@@ -34,6 +34,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import junit.framework.TestCase;
+
 import org.apache.commons.lang3.StringUtils;
 import org.json_voltpatches.JSONException;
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -378,6 +378,28 @@ static protected void assertClassesMatchNodeChain(
      * that make the "#include" start in the middle of a line.
      */
     private static final String TESTFILE_TEMPLATE =
+        "/* This file is part of VoltDB.\n" +
+        " * Copyright (C) 2008-2016 VoltDB Inc.\n" +
+        " *\n" +
```
BillWhite: If you want to add this, and I'm not sure you have decided this, I would replace the year string, 2016, with a parameter, say @YEAR@.  Then add YEAR to the list of defined parameters on or around line 800 in this file.  That way when the year changes the generated years change.

BillWhite:
I only have the one comment, other than our email thread.

PR #3826, opened 2016-08-11
ENG-10951: failed assertions in debug build

wolffcm:
In the backend there was an index on a COUNT(*) field of a MV.  The component type for the index was INTEGER, but we expected BIGINT.  I made HSQL set the output type of COUNT(*) to be BIGINT always.

I also loosened an equality check in `copyForPersistentUpdate`.  We were expecting the same instance of `TupleSchema` when copying tuples, but they really only need to be equivalent.  In particular, if we are copying into a tuple backed by `StandAloneTupleStorage`, the schema instance will never be the same as some persistent table, since the stand alone class allocates its own copy.

### DIFF
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -773,7 +773,7 @@ inline void TableTuple::copyForPersistentUpdate(const TableTuple &source,
                                                 std::vector<char*> &oldObjects, std::vector<char*> &newObjects)
 {
     assert(m_schema);
-    assert(m_schema == source.m_schema);
+    assert(m_schema->equals(source.m_schema));
     const int columnCount = m_schema->columnCount();
     const uint16_t uninlineableObjectColumnCount = m_schema->getUninlinedObjectColumnCount();
     /*
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
@@ -31,6 +31,7 @@
 
 package org.hsqldb_voltpatches;
 
+import java.io.Serializable;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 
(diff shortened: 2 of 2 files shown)


src/ee/common/tabletuple.h
```diff
@@ -773,7 +773,7 @@ inline void TableTuple::copyForPersistentUpdate(const TableTuple &source,
                                                 std::vector<char*> &oldObjects, std::vector<char*> &newObjects)
 {
     assert(m_schema);
-    assert(m_schema == source.m_schema);
+    assert(m_schema->equals(source.m_schema));
```
paulmartel: Arguably off topic / "broken when you found it" since we don't care all that much about debug/assert performance but:
should we add an optimizing short-circuit like `if (this == other) return true;` in TupleSchema::equals (especially if this case is at all likely in production)?

paulmartel:
Looks good.

src/ee/common/tabletuple.h
yzhang1991: I also have this change in my view test branch, Paul's suggestion sounds good.

src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
```diff
@@ -601,8 +600,8 @@ private Number getStdDev() {
         }
 
         return sample ? (n == 1) ? null    // NULL (not NaN) is correct in this case
-                                 : new Double(Math.sqrt(vk / (double) (n - 1)))
-                      : new Double(Math.sqrt(vk / (double) (n)));
+                                 : new Double(Math.sqrt(vk / (n - 1)))
+                      : new Double(Math.sqrt(vk / (n)));
```
BillWhite: This is also somewhat off topic, but aren't getStdDev() and getVariance() almost exactly the same?  Isn't one just Math.sqrt() of the other, with managing null return values?

Also, I think this is a very inefficient way of computing std deviation.  You are likely to get overflow in computing the sum of squares in addDataPoint().  There's a recurrence relation that does a better job.

BillWhite:
That's all I have.

src/ee/common/tabletuple.h
wolffcm: We do the pointer equality check already in `equals` (actually a method that it calls).

src/hsqldb19b3/org/hsqldb_voltpatches/SetFunction.java
wolffcm: This was just Eclipse formatting that changed this.  I don't have the background to be able to answer your questions.  If we ever support these aggregates, then we may wish to visit this code if we want to use HSQL as a back end.

PR #3828, opened 2016-08-11
Eng 10740 planner perf

paulmartel:
Code reductions:
    encapsulate inside AbstractExpression.java former ExpressionUtil methods that take an obvious this-like AbstractExpression argument. 
    encapsulate inside AbstractExpression.java most recursive expression tree searches with specific target Expression classes and optional caller-specific filters.
  use java generics to avoid the need to downcast the results of the recursive searches in the callers.
  encapsulate copying TVE value type/size settings from SchemaColumns, ColInfos, TVEs, and other expressions.
  encapsulate resetting of TVE column indexes.

Code optimizations:
    use a short-circuiting "has any" AbstractExpression tree search method instead of calling isEmpty() on the results of a "find all" method that collects all possible matches for nothing.
   eliminate some redundant attribute settings in clone methods (which still profile as rather expensive).
   eliminate ArrayList resizing by pre-allocating at least enough elements (when known).
   eliminate temporary ArrayLists that just addAll their contents to another ArrayList, when we can pre-allocate and directly fill the final ArrayList.
  eliminate some extraneous cloning where the original object is about to become unreachable.

Other simplifications:
  avoid potentially confusing edge case mismatches between an AbstractExpression's class and its ExpressionType -- like
 `new OperatorExpression(COMPARE_EQUAL, ...` instead of the canonical
 `new ComparisonExpression(COMPARE_EQUAL, ...`

### DIFF
--- a/src/frontend/org/voltdb/expressions/ComparisonExpression.java
+++ b/src/frontend/org/voltdb/expressions/ComparisonExpression.java
@@ -41,11 +41,13 @@ public class ComparisonExpression extends AbstractExpression {
     public ComparisonExpression(ExpressionType type) {
         super(type);
         setValueType(VoltType.BOOLEAN);
+        setValueSize(m_valueType.getLengthInBytesForFixedTypes());
     }
 
     public ComparisonExpression(ExpressionType type, AbstractExpression left, AbstractExpression right) {
         super(type, left, right);
         setValueType(VoltType.BOOLEAN);
+        setValueSize(m_valueType.getLengthInBytesForFixedTypes());
     }
 
     public ComparisonExpression() {
@@ -53,6 +55,8 @@ public class ComparisonExpression extends AbstractExpression {
         // This is needed for serialization
         //
         super();
+        setValueType(VoltType.BOOLEAN);
+        setValueSize(m_valueType.getLengthInBytesForFixedTypes());
     }
 
     public void setQuantifier(QuantifierType quantifier) {
@@ -84,7 +88,6 @@ public class ComparisonExpression extends AbstractExpression {
     @Override
     public Object clone() {
         ComparisonExpression clone = (ComparisonExpression) super.clone();
-        clone.m_quantifier = m_quantifier;
         return clone;
     }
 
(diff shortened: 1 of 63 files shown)


src/frontend/org/voltdb/expressions/ComparisonExpression.java
```diff
@@ -106,6 +109,33 @@ public void toJSONString(JSONStringer stringer) throws JSONException {
         }
     }
 
+    /**
+     * Does this expression meet the criteria for a column equivalence filter,
+     * as used in determining statement partitioning and allowable
+     * multi-partition joins.
+     * @return true if this is an equality between two TVEs or between a TVE and
+     * a runtime constant.
+     **/
```
wolffcm: There is a ticket that I filed where a customer notes that variations of simple equality predicates are not used when determining statement partitioning.  It is ENG-10857.

I wonder if we could expand the definition of "column equivalence filter" to include a cast on either side where the cast does not change the value.  At some point in the future, I mean.  This seems like one place that would need the change.

src/frontend/org/voltdb/expressions/ExpressionUtil.java
```diff
@@ -75,40 +95,25 @@ public static AbstractExpression combinePredicates(Collection<AbstractExpression
         if (stack.isEmpty()) {
             return null;
         }
-        return combineStack(stack);
+        return combinePredicateStack(stack);
     }
 
-    private static AbstractExpression combineStack(Stack<AbstractExpression> stack) {
+    private static AbstractExpression combinePredicateStack(Stack<AbstractExpression> stack) {
```
wolffcm: I feel like I already reviewed this...  I guess you resurrected an old branch that did not get merged as the basis for these changes?

src/frontend/org/voltdb/expressions/TupleValueExpression.java
```diff
@@ -40,7 +42,7 @@
 
     protected int m_columnIndex = -1;
```
wolffcm: Can these all by made private?  I don't think there's any subclasses of this class.

wolffcm:
I wonder if I could convince you to use several commits for a pull request like this.  It would be much easier to review each item in your list of code reductions and optimizations as a separate commit.  Whitespace-only changes could be a separate commit too.  GitHub has a really nice "squash and merge" option now when merging to master which would make this workflow really easy.

src/frontend/org/voltdb/expressions/TupleValueExpression.java
paulmartel: Probably. There was once a CORE engineer who had this misguided idea that generally using "protected" instead of "private" was somehow a good thing like "coding for the future". Fixing this is an ongoing task.

paulmartel:
@wolffcm I could fix this now, retroactively, if you like. The recurring problem is that I often find myself half-way into a solution when I realize that I want to do a semi-related refactor -- like when I realize that I am making the same patch to N instances of the same boiler-plate code that should be called in ONE place so it can be fixed in that ONE place. Or when I realize that I am patching dead code, or when I could not wrap my head around some existing code until I eliminated layers of arbitrary complexity or formatting.

So, what do you advise I do in these cases? Options:
    A) suck it up. make the N identical patches, or add more complexity or wonky formatting to fit in with the existing code, to get the current task work reviewed and merged, and go in LATER and do a technical debt reduction patch to master.
    B) put the current task on hold. Do the technical debt reduction first, wait to get it reviewed and merged, and then rebase the current task branch.
   C) proceed on one branch until ready for review but then tease out the changes for separate review and integration in whatever way seems most likely to avoid re-work -- knowing that there likely WILL be re-work (merge conflicts). This approach is somewhat helped by the fact that it can be easier to tease out the branches when you have a reference branch for the intended "final goal"
and you can handle merge conflicts by just hacking the code to look like the "final goal" version.
Yet, given that late changes to reflect review comments can move the goal posts, even this is no sure thing.

I honestly don't know how to decide these issues in the heat of battle.

wolffcm:
@paulmartel I don't think there's a one-size-fits-all answer to this question, but for your example of patching N instances of boilerplate code, here's a workflow that would improve my experience as a code reviewer.  It's a variation of B).

1. You start to make some changes and realize that you're changing boilerplate that is the same in several different places.
2. Do a `git stash`.
3. Factor all the instances of the boilerplate into a separate method.  Make sure you can build and maybe run a test or two to make sure you didn't break anything.
4. Do a `git commit`.
5. Do a `git stash pop`.  There will probably be conflicts to resolve.
6. Repeat steps 1-5 as needed.
7. When you push your changes, don't squash them, leave them as several commits in one pull request.  The code reviewer can then has the option to review each commit individually if it seems warranted.

rmorgenstein:
The associated ticket was closed a long time ago, presumably from another merge.  Closing this pull request.

PR #3831, opened 2016-08-11
Handle null sort directions when reading JSON for windowed aggregate plans.

BillWhite:
Sometimes the sort directions list in a plan with a windowed function is
a null pointer.  In particular, if we read a plan from JSON the sort
directions in the windowed functions are not represented.  This is only
important in the Java code when the plan is explained, since that's the
only time a plan is deserialized from JSON in Java code.  This is not a
problem in the EE, since the EE does not think to look for sort
directions in windowed expressions.

### DIFF
--- a/src/frontend/org/voltdb/expressions/AbstractExpression.java
+++ b/src/frontend/org/voltdb/expressions/AbstractExpression.java
@@ -728,14 +728,18 @@ public abstract class AbstractExpression implements JSONString, Cloneable {
                                                  JSONObject               jobj) throws JSONException {
         if (jobj.has(AbstractExpression.SortMembers.SORT_COLUMNS.name())) {
             sortExpressions.clear();
-            sortDirections.clear();
+            if (sortDirections != null) {
+                sortDirections.clear();
+            }
             JSONArray jarray = jobj.getJSONArray(SortMembers.SORT_COLUMNS.name());
             int size = jarray.length();
             for (int ii = 0; ii < size; ii += 1) {
                 JSONObject tempObj = jarray.getJSONObject(ii);
-                sortExpressions.add( AbstractExpression.fromJSONChild(tempObj, SortMembers.SORT_EXPRESSION.name()) );
-                if (sortDirections != null && tempObj.has(SortMembers.SORT_DIRECTION.name())) {
-                    sortDirections.add( SortDirectionType.get(tempObj.getString( SortMembers.SORT_DIRECTION.name())) );
+                if (sortDirections != null) {
+                    sortExpressions.add( AbstractExpression.fromJSONChild(tempObj, SortMembers.SORT_EXPRESSION.name()) );
+                    if (sortDirections != null && tempObj.has(SortMembers.SORT_DIRECTION.name())) {
+                        sortDirections.add( SortDirectionType.get(tempObj.getString( SortMembers.SORT_DIRECTION.name())) );
+                    }
                 }
             }
         }
--- a/src/frontend/org/voltdb/expressions/AbstractExpression.java
+++ b/src/frontend/org/voltdb/expressions/AbstractExpression.java
@@ -735,8 +735,8 @@ public abstract class AbstractExpression implements JSONString, Cloneable {
             int size = jarray.length();
             for (int ii = 0; ii < size; ii += 1) {
                 JSONObject tempObj = jarray.getJSONObject(ii);
+                sortExpressions.add( AbstractExpression.fromJSONChild(tempObj, SortMembers.SORT_EXPRESSION.name()) );
                 if (sortDirections != null) {
-                    sortExpressions.add( AbstractExpression.fromJSONChild(tempObj, SortMembers.SORT_EXPRESSION.name()) );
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -728,14 +728,18 @@ public static void loadSortListFromJSONArray(List<AbstractExpression> sortExpres
...
+                sortDirections.clear();
+            }
             JSONArray jarray = jobj.getJSONArray(SortMembers.SORT_COLUMNS.name());
             int size = jarray.length();
             for (int ii = 0; ii < size; ii += 1) {
                 JSONObject tempObj = jarray.getJSONObject(ii);
-                sortExpressions.add( AbstractExpression.fromJSONChild(tempObj, SortMembers.SORT_EXPRESSION.name()) );
-                if (sortDirections != null && tempObj.has(SortMembers.SORT_DIRECTION.name())) {
-                    sortDirections.add( SortDirectionType.get(tempObj.getString( SortMembers.SORT_DIRECTION.name())) );
+                if (sortDirections != null) {
```
paulmartel: If no sortDirections list is provided, then we don't bother collecting sortExpressions?

More to the point, it seems like the real bug here is that one of the four callers to this function is calling it with an explicit null pointer when it could almost as easily pass it the sortDirections List that corresponds to the sortExpressions List it is using. This would make it more consistent with the closely related callers, and then we would not have to worry that a slightly different call sequence could violate the asserts that a WindowedExpression has about its sortExpressions and sortDirections having the same length.

Also, could use a test case.

BillWhite: The problem is that in the one case, the null case, there are no sort expressions to pass in.  We don't serialize the sort directions for windowed expressions into JSON.  So, when we deserialize a JSON plan we don't have sort directions to deserialize.  When we have sort directions we need to check that there are as many expressions as sort directions.  But when we have no sort directions at all that's ok as well.  The case we never want is if there are sort directions but there are different numbers of sort directions and expressions.

It could use a test case as well, I suppose.

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
```diff
@@ -481,6 +481,28 @@ public void testRankOrderbyExpressions() throws Exception {
...
+    public void testExplainPlan() throws Exception {
+        Client client = getClient();
+        String sql = "select rank() over ( partition by A, B order by C ) from T;";
+
+        ClientResponse cr;
+        try {
+            cr = client.callProcedure("@Explain", sql);
+            assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+            VoltTable vt = cr.getResults()[0];
+            assertTrue(true);
```
paulmartel: assertTrue(true)?

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
```diff
@@ -481,6 +481,28 @@ public void testRankOrderbyExpressions() throws Exception {
...
+        } catch (Exception ex) {
+            assertTrue("Exception on @Explain of windowed expression", false);
+        }
```
paulmartel: assertTrue(message, false) --> fail(message)

src/frontend/org/voltdb/expressions/AbstractExpression.java
paulmartel: Looks like I missed the "assert(false);" at the top of WindowedExpression.loadFromJSonObject and WindowedExpression.toString. 
So, the assert that the lists have the same size is subsumed by the more aggressive assert.
Don't these functions seem rather elaborate for dead code?

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
BillWhite: I wanted a assertSomething test here, since it's a jUnit test.  But if we get an explain string, that's enough.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: Sorry, I'd forgotten that.  We don't put WindowedExpression objects in the display list at all.  When we see a windowed aggregate expression in the display list we create one of these and store it in the parseSelectStatement, but we put an ordinary column TVE in the display list.  This is to keep aliases in group by and order by expressions from copying the WindowedExpression objects and getting confused by them.  I wanted to make sure that the WindowedExpression's toJSON and fromJSON were never called.  I forgot to take them out.

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
BillWhite: Ok.  Fixed.

BillWhite:
Squashed on another branch.

PR #3832, opened 2016-08-11
ENG-10938: Kafka system test pause/resume case fails with short row c…

JadenW:
…ount

apprunner pause/resume tweak always have 10s intervals. 
Another option is let client try write something to see if server is in the pause mode.

### DIFF
--- a/tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
+++ b/tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
@@ -98,6 +98,8 @@ public class KafkaImportBenchmark {
 
     private static final int END_WAIT = 10; // wait at the end for import to settle after export completes
 
+    private static final int PAUSE_WAIT = 10; // wait for server resume from pause mode
+
     static List<Integer> importProgress = new ArrayList<Integer>();
 
     static InsertExport exportProc;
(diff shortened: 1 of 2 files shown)


pdshaw:
One other thing we can look at is add a "tweak" option to the Java client that's set by the runapp.py, which is where the tweak is set. Then the client can expect server down, or pause or whatever and set specific timeouts and other behavior appropriately.

A little more work but somewhat less arbitrary.

tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
```diff
@@ -422,6 +428,16 @@ public static void main(String[] args) throws Exception {
         long importRows = MatchChecks.getImportTableRowCount(config.alltypes, client);
         long importRowCount = MatchChecks.getImportRowCount(client);
 
+        // in case of pause / resume tweak, let it drain longer
+        int trial = 3;
+        while (PauseResumeTweak.equalsIgnoreCase(config.tweaktype) && (--trial > 0) && ((importStatValues[OUTSTANDING_REQUESTS] > 0) || (importRows < config.expected_rows))) {
```
akhanzode: Why not do @SystemInformation this will tel you if cluster is paused....then wait until its resumed and outstanding requests are 0.

akhanzode:
r=akhanzode

PR #3837, opened 2016-08-12
Use rack aware algorithm for rejoin

aquafemi:
- Fixed a bug in the original rack-aware algorithm to cover more cases.
- Rejoining node calls rack-aware algorithm similar to when create is called.

Since there's significant change to the algorithm, reviewing the diff in ClusterConfig may not be straightforward. It's easier to read the new algorithm directly. I suggest start from there and review the changes in the other files afterwards.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -62,6 +62,7 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import com.google_voltpatches.common.collect.HashMultimap;
 import org.apache.cassandra_voltpatches.GCInspector;
 import org.apache.log4j.Appender;
 import org.apache.log4j.DailyRollingFileAppender;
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1513,25 +1513,27 @@ private JSONObject getTopology(StartAction startAction, Map<Integer, String> hos
...
             try {
-                topo =
-                    new JSONObject(new String(m_messenger.getZK().getData(VoltZK.topology, false, stat), "UTF-8"));
+                final Set<Integer> liveHostIds = m_messenger.getLiveHostIds();
+                Preconditions.checkArgument(hostGroups.keySet().equals(liveHostIds));
+                int hostcount = hostGroups.size();
+                ClusterConfig clusterConfig = new ClusterConfig(hostcount, sitesperhost, kfactor);
+                topo = clusterConfig.getTopology(hostGroups);
             }
             catch (Exception e) {
```
aquafemi: Should change this catch block to reflect what's really going on.

src/frontend/org/voltdb/compiler/ClusterConfig.java
```diff
@@ -736,13 +852,15 @@ public int compare(Node o1, Node o2)
 
     // Statically build a topology. This only runs at startup;
     // rejoin clones this from an existing server.
-    public JSONObject getTopology(Map<Integer, String> hostGroups) throws JSONException
+    public JSONObject getTopology(Map<Integer, String> hostGroups,
+                                  Multimap<Integer, Long> partitionReplicas,
+                                  Map<Integer, Long> partitionMasters) throws JSONException
     {
         int hostCount = getHostCount();
         int partitionCount = getPartitionCount();
```
luyangco: The return value of `getPartitionCount()` may vary depends on how many hosts in the cluster. In rejoin case, if there are more than two nodes missing, call `getPartitionCount()` will get wrong number of master partitions. Should use `partitionMasters.size()` instead.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1559,28 +1556,33 @@ private JSONObject getTopology(StartAction startAction, Map<Integer, String> hos
...
             }
-            topo = registerClusterConfig(clusterConfig, hostGroups);
-        }
-        else {
-            Stat stat = new Stat();
+
             try {
-                topo =
-                    new JSONObject(new String(m_messenger.getZK().getData(VoltZK.topology, false, stat), "UTF-8"));
+                final Multimap<Integer, Long> replicas = m_cartographer.getReplicasForPartitions(m_cartographer.getPartitions());
```
luyangco: `getReplicasForPartitions()` returns both masters and replicas, here what really need is a diff to remove master partitions from it.

xinjiacs:
Fixed in ENG-11442 AbstractTopology patch. Outdated pull requests.

PR #3842, opened 2016-08-16
Fix ENG-11000 wrong answer when evaluating min from index scan having…

yzhang1991:
… nulls

### DIFF
--- a/src/frontend/org/voltdb/planner/microoptimizations/ReplaceWithIndexLimit.java
+++ b/src/frontend/org/voltdb/planner/microoptimizations/ReplaceWithIndexLimit.java
@@ -317,6 +317,7 @@ public class ReplaceWithIndexLimit extends MicroOptimization {
                 ispn.setLookupType(IndexLookupType.GTE);
                 ispn.removeLastSearchKey();
                 ispn.addEndExpression(ExpressionUtil.uncombinePredicate(ispn.getInitialExpression()).get(numberOfExprs - 1));
+                ispn.setSkipNullPredicate(numOfSearchKeys);
                 ispn.resetPredicate();
             }
             // add an inline LIMIT plan node to this index scan plan node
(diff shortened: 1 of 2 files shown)


wolffcm:
I am reviewing this now.

src/frontend/org/voltdb/planner/microoptimizations/ReplaceWithIndexLimit.java
```diff
@@ -317,6 +317,7 @@ else if (numberOfExprs > 1) {
                 ispn.setLookupType(IndexLookupType.GTE);
                 ispn.removeLastSearchKey();
                 ispn.addEndExpression(ExpressionUtil.uncombinePredicate(ispn.getInitialExpression()).get(numberOfExprs - 1));
+                ispn.setSkipNullPredicate(numOfSearchKeys);
```
wolffcm: Any other places where we're missing the skip null predicate?

yzhang1991: Only range scan with < or <= need to have skip null predicate.
If isTableMin => current fix
If isTableMax => do not need to skip null because null is the smallest.

So I think we are not missing anything

wolffcm: Curious about where `numOfSearchKeys` would be greater than one here.  Is it possible?  If so, there should be a test case for it.

wolffcm: Seems like you need this test case:
```SQL
select min(x) from t where w = 10 and x <= ?; -- index on (w, x)
```
Just to make sure the optimization is applied correctly for multi-component indexes.  If this works, then I'm satisfied and feel free to merge when you're ready.

yzhang1991: IndexScanPlanNode.java, line 965, I think because of this we cannot find a case where `numOfSearchKeys` is greater than one. Although I am not 100% sure.

wolffcm: Seems like the `m_predicate' field is actually the post-filter for the node?  That is, those filters not implemented by the index scan.  

For the new case you added, does the optimization get applied, or not?  Just curious.

yzhang1991: Yes

wolffcm:
Looks good, merge when you're ready.

PR #3843, opened 2016-08-16
ENG-10968 run-everywhere

zeemanhe:
Added a sync and an async method which go through partitions for the stored proc to be executed.


jhugg:
Can you make the API look like https://issues.voltdb.com/browse/ENG-10296.

That has the latest feedback from Andrew. I'm sorry that didn't make it into the ticket https://issues.voltdb.com/browse/ENG-10968

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,36 @@ public boolean updateClasses(ProcedureCallback callback,
      */
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, boolean upsert, BulkLoaderFailureCallBack blfcb) throws Exception;
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkLoaderFailureCallBack blfcb) throws Exception;
+
+    /**
+     * <p>Synchronously invoke a procedure. Blocks until a result is available. A {@link ProcCallException}
```
jhugg: This doc isn't specific to the new functionality yet?

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,111 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
             return new VoltBulkLoader(m_vblGlobals, tableName, maxBatchSize, blfcb);
         }
     }
+
+    @Override
+    public PartitionClientResponse[] callAllPartitionProcedure(String procedureName, Object... params)
```
jhugg: I would suggest we make the sync version call the async version and block on the callback. This reduces duplicated code, and also doesn't serialize the calls, which isn't really what we want.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,111 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+            partitionCount--;
+            PartitionProcedureCallback cb = new PartitionProcedureCallback(responseWaiter, key, partitionCount, responses);
+            if(!callProcedure(cb, procedureName, args)){
+                return false;
+            }
+        }
+
+        try {
+            responseWaiter.await();
+            callback.clientCallback(responses);
```
jhugg: This isn't actually asynchronous. We're blocking here. 

We probably want to use a specialized callback for each call that aggregates responses, and calls the users callback if it's the last one.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,111 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+        }
+        return responses;
+    }
+
+    @Override
+    public boolean callAllPartitionProcedure(AllPartitionProcedureCallback callback, String procedureName,
+            Object... params) throws IOException, NoConnectionsException, ProcCallException {
+
+        Object[] args = new Object[params.length + 1];
+        System.arraycopy(params, 0, args, 1, params.length);
```
jhugg: Not all procedures partition on the first parameter. We could add the restriction that all-partition procs must I guess. We do have this information for client affinity purposes as well.

Also, does this work with strings or ints? Is this another restriction?

src/frontend/org/voltdb/client/PartitionClientResponse.java
```diff
@@ -0,0 +1,45 @@
...
+import org.voltdb.client.ClientResponse;
+
+/**
+ * Packages up the data to be sent back to the client as a stored
+ * procedure response for a partition
+ *
+ */
+public class PartitionClientResponse {
+
+    private final Object m_partitionKey;
```
jhugg: just make them public final and get rid of the accessors.

src/frontend/org/voltdb/client/PartitionClientResponse.java
```diff
@@ -0,0 +1,45 @@
...
+    private ClientResponse m_clientResponse;
+
+    public PartitionClientResponse(Object partitionKey, ClientResponse clientResponse) {
```
jhugg: This shouldn't be public.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,111 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+
+        setPartitions();
```
jhugg: This only fetches the keys the first time and won't work if the topology ever changes. It might be better to integrate this data with the client affinity data kept in the distributer. We might even want to extend the subscription thing that gets us periodic details from the server side to ensure it has what we want.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,111 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+    /**
+     * Set up partitions.
+     * @throws ProcCallException on any VoltDB specific failure.
+     * @throws NoConnectionsException if this {@link Client} instance is not connected to any servers.
+     * @throws IOException if there is a Java network or connection problem.
+     */
+    private void setPartitions() throws IOException, NoConnectionsException, ProcCallException {
+        if(m_partitionKeys == null){
+            m_partitionKeys = new ArrayList<Integer>();
+            VoltTable results[] = callProcedure("@GetPartitionKeys", "integer").getResults();
```
xinjiacs: Use the client Distributer Affinity to get the type of partition parameter so that we do not introduce new limitations on partition column type.  Not necessary to be integer always.

xinjiacs:
We need test cases of exercising the new API using different parameter type stored procedures.

src/frontend/org/voltdb/client/ClientImpl.java
jhugg: Here's a dumb trick. You can always use integer partition keys with quotes around them. "1" and 1 will always route to the same place, intentionally. So you shouldn't have to call getPartitionKeys for both types.

src/frontend/org/voltdb/client/ClientImpl.java
zeemanhe: will look into it.

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,38 @@ public boolean updateClasses(ProcedureCallback callback,
      */
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, boolean upsert, BulkLoaderFailureCallBack blfcb) throws Exception;
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkLoaderFailureCallBack blfcb) throws Exception;
+
+    /**
+     * <p>Synchronously invoke a single partition procedure. The call will iterate through all the partitions and
+     * execute the stored procedure one partition at a time, and return response for each partition. Blocks until a result is available.
+     * A {@link ProcCallException} is thrown if the response is anything other then success.</p>
```
jhugg: This is a really good, unanswered design question. If we have partial failure, should it throw or have some responses returned be failed. 

I kinda like the idea that it throws, but then how do you understand what failed and what succeeded? Needs thought.

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,38 @@ public boolean updateClasses(ProcedureCallback callback,
      */
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, boolean upsert, BulkLoaderFailureCallBack blfcb) throws Exception;
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkLoaderFailureCallBack blfcb) throws Exception;
+
+    /**
+     * <p>Synchronously invoke a single partition procedure. The call will iterate through all the partitions and
```
jhugg: We don't really want the sync call to run one at a time. It should send them all at once.

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,38 @@ public boolean updateClasses(ProcedureCallback callback,
...
+     * @return {@link ClientResponseWithPartitionKey} instance of procedure call results.
+     * @throws ProcCallException on any VoltDB specific failure.
+     * @throws NoConnectionsException if this {@link Client} instance is not connected to any servers.
+     * @throws IOException if there is a Java network or connection problem.
+     */
+    public ClientResponseWithPartitionKey[] callAllPartitionProcedure(String procedureName, Object... params)
+            throws IOException, NoConnectionsException, ProcCallException;
+
+    /**
+     * <p>Asynchronously invoke a single partition procedure, by providing a callback that will be invoked by the single
```
jhugg: Out of date comments. These will probably need some careful text though, so we can worry about this as we get closer to done.

src/frontend/org/voltdb/client/ClientImpl.java
zeemanhe: Form what I understand, the first arg on procedure.run(...) will be used for partition lookup. If a proc has two parameters, procedure.run(...) must have 3 arguments.  We get a list of partition keys and iterate through. The column partition parameter is not used for partition lookup. I could be wrong.

I will dig into more on the types of the partitions.

src/frontend/org/voltdb/client/ClientResponseWithPartitionKey.java
```diff
@@ -0,0 +1,39 @@
...
+import org.voltdb.client.ClientResponse;
+
+/**
+ * Packages up the data to be sent back to the client as a stored
+ * procedure response for a partition
+ *
+ */public class ClientResponseWithPartitionKey {
+
+    public final Object m_partitionKey;
+    private ClientResponse m_clientResponse;
```
jhugg: Get rid of m_ for public final members. Make clientResponse public final and get rid of the accessor.

src/frontend/org/voltdb/client/ClientImpl.java
zeemanhe: Thanks, great!

src/frontend/org/voltdb/client/AllPartitionProcedureCallback.java
```diff
@@ -0,0 +1,33 @@
...
+ * stored procedure invocation on all partitions.
+ */
+public interface AllPartitionProcedureCallback {
+    /**
+     * Implementation of callback to be provided by client applications.
+     *
+     * @param clientResponse Responses for each partition to the stored procedure invocation this callback is associated with
+     * @throws Exception on any Exception.
+     */
+    abstract public void clientCallback(ClientResponseWithPartitionKey[] clientResponse) throws Exception;
```
jhugg: Make "clientResponse" into "responses".

We might also want to add a boolean that conveys whether 100% of the responses were successful.

src/frontend/org/voltdb/client/ClientImpl.java
jhugg: If you don't specify a parameter offset when you partition a proc, we assume it's the first parameter. That's only a default though.

The rest makes sense. Feel free to ask me questions face-to-face if you want to go over more of this.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,112 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+        return private_executeAllPartitionProcedure(null, procedureName, params);
+    }
+
+    @Override
+    public boolean callAllPartitionProcedure(AllPartitionProcedureCallback callback, String procedureName,
+            Object... params) throws IOException, NoConnectionsException, ProcCallException {
+        return (private_executeAllPartitionProcedure(callback, procedureName, params) != null);
+    }
+
+    private ClientResponseWithPartitionKey[] private_executeAllPartitionProcedure(AllPartitionProcedureCallback callback,
```
jhugg: You don't need a private method here. You can literally call the async version unchanged from the sync version. Just pass in a custom callback that looks like an all-partition version of SyncCallback and block on it.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +816,112 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+            String procedureName, Object... params) throws IOException, NoConnectionsException, ProcCallException {
+
+        Object[] args = new Object[params.length + 1];
+        System.arraycopy(params, 0, args, 1, params.length);
+
+        int partitionCount = getPartitionIntegerKeys().size();
```
jhugg: Right now this is going to call GetPartitionKeys for every call? That's not very efficient, especially because we already have a mechanism in Distributor for being notified of topology changes. We should only get new partition keys when the topology changes.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +817,171 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+    }
+
+    /**
+     * Set up partitions.
+     * @throws ProcCallException on any VoltDB specific failure.
+     * @throws NoConnectionsException if this {@link Client} instance is not connected to any servers.
+     * @throws IOException if there is a Java network or connection problem.
+     */
+    private void setPartitions() throws IOException, NoConnectionsException, ProcCallException {
+        if(m_partitionIntegerKeys.isEmpty()){
```
xinjiacs: It's better to return early for non-empty case and save one indentation, like:
if(! m_partitionIntegerKeys.isEmpty()) {
    return;
}
long time = System.currentTimeMillis() - m_lastPartitionKeyFetched;
...

Also, we should have consistent space, like "if (...) {".

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +817,171 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+    * @throws NoConnectionsException if this {@link Client} instance is not connected to any servers.
+    * @throws IOException if there is a Java network or connection problem.
+    */
+    public List<Integer> getPartitionIntegerKeys() throws NoConnectionsException, IOException, ProcCallException {
+        if(m_distributer.getPartitionKeys().size() > 0){
+            return m_distributer.getPartitionKeys();
+        }
+
+        long time = System.currentTimeMillis() - m_lastPartitionKeyFetched;
+        if(time > 5 * 60 * 1000){
```
xinjiacs: It's better to have a final constant for the 5 minutes period. Not sure if that's too long after a topology change. Anyway, we can keep it a constant here for further benchmark testing or adding an configuration later.

Can you also check other places having the space format inconsistent issue?

xinjiacs: And we can early return the case less than 5 minutes, and put the "setPartitions" function inside only for case that we need to update.

"if (time <= PARTITION_KEYS_INFO_REFRESH_FREQUENCY) {
    return m_partitionIntegerKeys;
}
"

src/frontend/org/voltdb/client/ClientImpl.java
xinjiacs: Probably, we can just assert empty collection here. We have one caller knowing to update the partition keys.
assert(m_partitionIntegerKeys.isEmpty());

Maybe rename the function as "refreshPartitionKeys".

src/frontend/org/voltdb/client/ClientImpl.java
zeemanhe: Will make changes.  
With client affinity and topology update are on,   the partition keys are fetched from Distributer. The 5 min refresh applies only when client affinity is off.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +817,171 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+        final AllPartitionProcedureCallback m_cb;
+
+        /**
+         * Callback initialization
+         * @param responseWaiter The count down latch
+         * @param partitionKey  The partition where the call back works on
+         * @param index  The index for PartitionClientResponse
+         * @param responses The final result array
+         */
+        public PartitionProcedureCallback(AtomicInteger counter, Object partitionKey, int index,
```
xinjiacs: Maybe rename to "OnePartitionProcedureCallback", and make it a private internal class instead.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +817,171 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+            if(m_partitionCounter.get() == 0){
+                m_cb.clientCallback(m_responses);
+            }
+        }
+    }
+
+    /**
+     * Sync all partition procedure call back
+     */
+    class SyncAllPartitionProcedureCallback implements AllPartitionProcedureCallback {
```
xinjiacs: Make it a private internal class instead.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +817,171 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+            try{
+                if(callback instanceof SyncAllPartitionProcedureCallback){
+                    ClientResponse response =  callProcedure(procedureName, args);
+                    cb.clientCallback(response);
+                } else {
+                    if(!callProcedure(cb, procedureName, args)){
+                        cb.clientCallback(null);
+                    }
+                }
+            } catch (ProcCallException | NoConnectionsException pe){
```
xinjiacs: Do not need to catch ProcCallException | NoConnectionsException when you do the same thing for Exception. You can simply remove this catch block.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +817,171 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+
+        @Override
+        public void clientCallback(ClientResponse response) throws Exception {
+            if(response != null) {
+                m_responses[m_index] = new ClientResponseWithPartitionKey(m_partitionKey, response);
+            } else {
+                final ClientResponse r = new ClientResponseImpl(ClientResponse.GRACEFUL_FAILURE, new VoltTable[0],
+                        "The procedure is not queued for execution.");
+                m_responses[m_index] = new ClientResponseWithPartitionKey(m_partitionKey, r,
+                        "The procedure is not queued for execution.", null);
```
xinjiacs: I guess you can use "r.getStatusString()" function instead to remove the error message duplication.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -821,43 +826,38 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
         ClientResponseWithPartitionKey[] responses = new ClientResponseWithPartitionKey[partitionCount];
         AtomicInteger counter = new AtomicInteger(partitionCount);
         for (Integer key : getPartitionIntegerKeys()) {
             args[0] = key;
             partitionCount--;
             PartitionProcedureCallback cb = new PartitionProcedureCallback(counter, key, partitionCount, responses, callback);
-            try{
-                if(callback instanceof SyncAllPartitionProcedureCallback){
+            try {
+                if (callback instanceof OnePartitionProcedureCallback) {
```
xinjiacs: Sorry if I do not make it clear.

Keep "SyncAllPartitionProcedureCallback", but rename "PartitionProcedureCallback" as "OnePartitionProcedureCallback".

src/frontend/org/voltdb/client/ClientResponseWithPartitionKey.java
```diff
@@ -0,0 +1,55 @@
...
+import org.voltdb.client.ClientResponse;
+
+/**
+ * Packages up the data to be sent back to the client as a stored
+ * procedure response for a partition
+ *
+ */
+public class ClientResponseWithPartitionKey {
+
+    public final Object m_partitionKey;
```
xinjiacs: Can you make member variables protected and add getter functions for them? That make this public API more formal for users like our existing client library.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -46,6 +48,11 @@
  */
 public final class ClientImpl implements Client, ReplicaProcCaller {
 
+    /*
+     * refresh the partition key cache every 5 min
+     */
+    static long PARTITION_KEYS_INFO_REFRESH_FREQUENCY = 5 * 60 * 1000;
```
xinjiacs: The frequency is too large, our window example use once a second. We may need a similar value for it.

tests/frontend/org/voltdb/client/TestAllPartitionProcedureCalls.java
```diff
@@ -63,78 +103,103 @@ public void setUp() throws Exception
...
-        } finally {
-            if (client != null) {
-                try {
-                    client.close();
-                } catch(InterruptedException e) {
-                    e.printStackTrace();
-                }
+        Client client = getClient();
+        client.callAllPartitionProcedure(new CallBack(STRING_PARTITION_EXPECTED_COUNTS), "PartitionStringTestProc");
+        Thread.sleep(2000);
```
xinjiacs: The Async test version does not necessary validate the result. It could be the "callAllPartitionProcedure" never return anything. We want to ensure the "clientCallback" has been executed.

tests/frontend/org/voltdb/client/allpartitioncall.sql
```diff
@@ -3,17 +3,16 @@ CREATE TABLE TABLE_INT_PARTITION
   id       			 		bigint     NOT NULL,
   value_string              varchar(50) NOT NULL,
   value_number1  			bigint    NOT NULL,
-  value_number2  			bigint    NOT NULL
+  value_number2  			bigint    NOT NULL,
+  value_ts					TIMESTAMP NOT NULL
```
xinjiacs: As timestamp is not able to be partitioned, maybe we should remove it from the tests and simply the LOAD function by removing the CONVERT function.

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,102 @@ public boolean updateClasses(ProcedureCallback callback,
...
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkLoaderFailureCallBack blfcb) throws Exception;
+
+    /**
+     * <p>
+     * The method uses system procedure <strong>@GetPartitionKeys</strong> to get a set of partition values and then execute the stored procedure
+     * one partition at a time, and return an aggregated response. Blocks until results are available.
+     * </p><p>
+     * The set of partition values is cached to avoid repeated requests to fetch them. However the database partitions may be changed. The cached set of partition values will
+     * be updated when the client affinity feature {@link ClientConfig#setClientAffinity(boolean)} is enabled and database cluster topology is updated.
+     * If the client affinity is not enabled, the cached set of partition values will be synchronized with the database very 5 minutes. The database partitions are usually pretty static.
```
xinjiacs: Every 1 second now.

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,102 @@ public boolean updateClasses(ProcedureCallback callback,
...
+     * But when the cached set of partition values is out sync with the database, a procedure execution may be routed to an incorrect partition.
```
xinjiacs: I would not use "incorrect partition". We can say  " a procedure may be routed to a partition zero time or multiple times, exactly once per partition will not be guaranteed."

src/frontend/org/voltdb/client/Client.java
```diff
@@ -422,4 +422,102 @@ public boolean updateClasses(ProcedureCallback callback,
...
+     * </p><p>There may be undesirable impact on latency and throughput as a result of running a multi-partition procedure. This is particularly true for longer running procedures.
+     * Using multiple, smaller procedures can also help reducing latency and increasing throughput, for queries that modify large volumes of data, such as large deletes.
```
xinjiacs: Maybe also worth mention:
If strong consistency can be relaxed,  using multiple smaller single partition procedures call can help reducing latency and increasing throughput." For example, process like aging out large stale data does not need to be global strong consistent using a multiple partition transaction.

xinjiacs:
The rest looks good to me. We can ask Andrew to review the Java client doc later.

tests/frontend/org/voltdb/client/allpartitioncall.sql
```diff
@@ -3,16 +3,14 @@ CREATE TABLE TABLE_INT_PARTITION
   id       			 		bigint     NOT NULL,
```
xinjiacs: Can you also fix the indentation ?

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +822,167 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+        SyncAllPartitionProcedureCallback callBack = new SyncAllPartitionProcedureCallback();
+        callAllPartitionProcedure(callBack, procedureName, params);
+        return callBack.getResponse();
+    }
+
+    @Override
+    public boolean callAllPartitionProcedure(AllPartitionProcedureCallback callback, String procedureName,
+            Object... params) throws IOException, NoConnectionsException, ProcCallException {
+        if (callback == null) {
+            throw new IOException("AllPartitionProcedureCallback can not be null");
```
jhugg: Should this be an IOException or an IllegalArgumentException?

src/frontend/org/voltdb/client/AllPartitionProcedureCallback.java
jhugg: This is still an open request.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +822,167 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+
+        Object[] args = new Object[params.length + 1];
+        System.arraycopy(params, 0, args, 1, params.length);
+
+        int partitionCount = getPartitionIntegerKeys().size();
+        assert(partitionCount > 0);
+        ClientResponseWithPartitionKey[] responses = new ClientResponseWithPartitionKey[partitionCount];
+        AtomicInteger counter = new AtomicInteger(partitionCount);
+        for (Integer key : getPartitionIntegerKeys()) {
+            args[0] = key;
```
jhugg: I'm worried about the UX of calling a proc that doesn't partition on the first parameter. The user will currently get a terrible and unhelpful error. What can we do to make this better?

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +822,167 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+            partitionCount--;
+            OnePartitionProcedureCallback cb = new OnePartitionProcedureCallback(counter, key, partitionCount, responses, callback);
+            try {
+                if (callback instanceof SyncAllPartitionProcedureCallback) {
```
jhugg: Why? Why are we synchronously calling this proc for each partition instead of calling the procedure against all partitions asynchronously at once.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +822,167 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+    * @throws IOException if there is a Java network or connection problem.
+    */
+    public List<Integer> getPartitionIntegerKeys() throws NoConnectionsException, IOException, ProcCallException {
+        if (m_distributer.getPartitionKeys().size() > 0) {
+            return m_distributer.getPartitionKeys();
+        }
+
+        long time = System.currentTimeMillis() - m_lastPartitionKeyFetched;
+        if (time > PARTITION_KEYS_INFO_REFRESH_FREQUENCY) {
+            m_partitionIntegerKeys.clear();
```
jhugg: This API is really async. Why clear and then populate this list? That will leave it unpopulated while you fetch?

Fetch it once a second and then atomic swap the pointers. Or don't if the lists are identical.

You may want to consider making the list itself immutable using a Guava immutable set.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -814,4 +822,167 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, BulkL
...
+            m_partitionCounter = counter;
+            m_partitionKey = partitionKey;
+            m_index = index;
+            m_responses = responses;
+            m_cb = cb;
+        }
+
+        @Override
+        public void clientCallback(ClientResponse response) throws Exception {
+            if (response != null) {
```
jhugg: Why would response be null? What am I missing?

zeemanhe: if  an asyc call is not queued, the stored proc process will not enter clientCallback. To ensure that every partition gets its ClientResponse, client  will call clientCallback(null) to set up ClientResponse  for the partition.

src/frontend/org/voltdb/client/ClientImpl.java
zeemanhe: will make the change

src/frontend/org/voltdb/client/ClientResponseWithPartitionKey.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+    *
+    * @param partitionKey  A partition key to reach the partition
+    * @param response The ClientResponse instance from the procedure execution
+    * @param errorMessage  Error message
+    * @param cause  The exception
+    */
+   ClientResponseWithPartitionKey(Object partitionKey, ClientResponse response, String errorMessage, Exception cause) {
```
jhugg: The response already has an error message and a cause. They don't need to be here.

Still want two immutable values with no accessors.

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -185,6 +187,28 @@ public void clientCallback(ClientResponse clientResponse) throws Exception {
...
+                try {
+                    synchronized (Distributer.this) {
+                        VoltTable results[] = clientResponse.getResults();
+                        if (results != null && results.length > 0) {
+                            updatePartitioning(results[0]);
+                        }
+                    }
+                }
+                catch (Exception e) {
+                    e.printStackTrace();
```
jhugg: What do we want to do when this error happens? I doubt we want to print the exception and continue.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -539,7 +548,7 @@ else if(procedure.sql != null) {
                 String[] parameter = procedure.partitionInfo.split(":");
                 String[] token = parameter[0].split("\\.");
                 String position = "";
-                if(Integer.parseInt(parameter[1].trim()) > 0) {
+                if(parameter.length > 1 && Integer.parseInt(parameter[1].trim()) > 0) {
```
jhugg: >= 1 or > 1 ?

src/frontend/org/voltdb/client/ClientImpl.java
jhugg: I see. So you fake the client response here. Why not just fake the client response above where it catches the exception. Then you know all procs get callbacks and you have to do less work here.

jhugg: Do the same thing above when it returns false or throws and exception, and you don't need the exception callback code below either.

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -1403,4 +1448,8 @@ private ByteBuffer serializeSPI(ProcedureInvocation pi) throws IOException {
     long getProcedureTimeoutNanos() {
         return m_procedureCallTimeoutNanos;
     }
+
+   public  List<Integer> getPartitionKeys() {
```
jhugg: This has weird whitespace. Also why public?

Also why not immutable set?

src/frontend/org/voltdb/client/Distributer.java
zeemanhe: I thought it about it.  TopoUpdateCallback, ProcUpdateCallback, etc, did the same. Maybe we should not ask distributer hand over the list if exception occurs?

PR #3845, opened 2016-08-16
ENG-10913 use number of coordinators as hostcount if it is unspecified

vtkstef:


akhanzode:
r=akhanzode see you on master

PR #3847, opened 2016-08-18
ENG-10758: warnings if ulimit or JVM settings are insufficient

cyx11:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3879,6 +3882,61 @@ static public long computeMinimumHeapRqt(boolean isPro, int tableCount, int site
...
+        int expThreadsCount = computeThreadsCount(tableCount, partitions, replicates, importPartitions, exportTableCount, exportNonceCount, hostcount);
+
+        String[] command = {"bash", "-c" ,"ulimit -u"};
+        String cmd_rst = ShellTools.local_cmd(command);
+        int maxThreadsCount = 0;
+        try {
+            maxThreadsCount = Integer.parseInt(cmd_rst.substring(0, cmd_rst.length() - 1));
+        } catch(Exception e) {
+            maxThreadsCount = Integer.MAX_VALUE;
+        }
```
JadenW: We don't to need to do the" bash ulimit" every time since it won't change frequently.
One way we can cache the value at start and update every hour.
Take a look at ResourceUsageMonitor.java  for how we monitoring our memory usage.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3879,6 +3882,61 @@ static public long computeMinimumHeapRqt(boolean isPro, int tableCount, int site
...
+            consoleLog.warn(builder.toString());
+        }
+    }
+
+    private int computeThreadsCount(int tableCount, int partitionCount, int replicateCount, int importerPartitionCount, int exportTableCount, int exportNonceCount, int hostcount) {
+        final int clusterBaseCount = 5;
+        final int hostBaseCount = 56;
+        // TODO: add DR count
+        return clusterBaseCount + (hostBaseCount + partitionCount) * hostcount
+                + computeImporterThreads(importerPartitionCount)
```
JadenW: This seems to be the total threads across the cluster, while ulimit is  per machine.
We can assume that one machine run only one host (in production).
I guess you mean there are threads increase with number of site ?

PR #3848, opened 2016-08-18
ENG-10972: Resolve column indices for ORDER BY expressions of windowed aggregates

wolffcm:
I added a resolveColumnIndices method to `PartitionByPlanNode`.  I also moved some member variables to the top of classes.  This conventions seems to actually be followed pretty consistently within our code base, so I thought it would be good to follow the trend.


BillWhite:
Looks right to me, as we discussed yesterday.

BillWhite:
That seems simple to me.

PR #3852, opened 2016-08-18
DR graph not available on the replica cluster

JadenW:


akhanzode:
r=akhanzode see you on master

PR #3856, opened 2016-08-19
ENG-11025 fix sum distinct on mv wrong answer

paulmartel:


wolffcm:
@paulmartel Could you add a comment either here or in the ticket describing why we were getting a wrong answer?  What was wrong about the plan we were generating?  I'll offer my stab at what's going on here.

In the ticket that `edgeCaseQueryNoFixNeeded` was created to fix, we were aggregating a join on a materialized view MV, where MV's GB keys did not include the partition ID for the table.  The output of the join is also being aggregated, such that the re-aggregation that would normally be required for data pulled from the MV to be correct would be subsumed by the user-specified join in the query.

In cases like that we can skip the "fixing" step, but only for certain kinds of aggregates in the user SELECT query, which doesn't include SUM(DISTINCT ...), hence your fix.

src/frontend/org/voltdb/planner/MaterializedViewFixInfo.java
```diff
@@ -286,20 +289,32 @@ private boolean edgeCaseQueryNoFixNeeded(Set<String> mvDDLGroupbyColumnNames,
                 return false;
             }
             ExpressionType type = aggExpr.getExpressionType();
-            TupleValueExpression tve = (TupleValueExpression) aggExpr.getLeft();
-            String columnName = tve.getColumnName();
 
-            if (type != ExpressionType.AGGREGATE_SUM && type != ExpressionType.AGGREGATE_MIN
+            // Only MIN, MAX, and non-DISTINCT SUM
+            // can tolerate a skipped reaggregation.
+            if ((type != ExpressionType.AGGREGATE_SUM || aggExpr.isDistinct())
```
wolffcm: I find this logic a little hard to follow.  This code implies that in the presence of `COUNT(DISTINCT ...)` it would be okay to skip the fix.  But that doesn't seem right.  I think this logic needs to be updated such that the fix is always applied whenever there distinct agg.

tests/frontend/org/voltdb/regressionsuites/TestGroupByComplexMaterializedViewSuite.java
```diff
@@ -1049,6 +1049,11 @@ private void mvBasedP1_Agg() throws Exception {
                     + tb).getResults()[0];
             assertEquals(95, vt.asScalarLong());
 
+            // Test for ENG-11025 regression -- mistaken optimization of sum(distinct)
```
wolffcm: Please add a test case for `count(distinct col)` here as well.

wolffcm:
That's all my comments.  Thanks for jumping on this so quick.  In summary, think about what count(distinct) means for this case.

src/frontend/org/voltdb/planner/MaterializedViewFixInfo.java
paulmartel: I don't understand why you are misreading this in spite of the comment. How does the exclusion (return false) for anything except MIN, MAX, and non-DISTINCT SUM leave you feeling like COUNT(DISTINCT) is slipping through?

wolffcm: I saw your comment of course.  I found it difficult to verify to myself that the boolean expression in the if-statement matched what the comment said.  Re-examining it now it looks okay.  Sorry about that.

PR #3859, opened 2016-08-22
ENG-11024: Fix edge case of doing a count(*) of view with no GB keys

wolffcm:
In this situation we can end up with plan nodes that have
a zero-column output schema.  Explicitly add the count(*)
column from the view to avoid this.


paulmartel:
Looks good.

PR #3860, opened 2016-08-22
ENG-11029: Planner assertions for some seemingly innocuous RANK queries

wolffcm:
The issue here is with the generation of the synthesized RANK column produced by PartitionByPlanNode.  When we need to add an extra projection to the plan, we need to update the "differentiator" field of TVEs in order to help us disambiguate columns in some obscure cases.  The code that does this adjustment is confused by synthesized columns.

The fix here was just to avoid doing the differentiator adjustment for synthesized columns, which consistently have the table alias `VOLT_TEMP_TABLE`.

I think that we didn't find this before because we happened to have combinations of column indexes such that things happened to work.

We are close to shipping 6.6, but this seems like it could make RANK very difficult to use for some tables, and this change seems pretty safe to make.

### DIFF
--- a/src/frontend/org/voltdb/planner/AbstractParsedStmt.java
+++ b/src/frontend/org/voltdb/planner/AbstractParsedStmt.java
@@ -585,7 +585,11 @@ public abstract class AbstractParsedStmt {
             }
         }
 
+        String columnName = WINDOWED_AGGREGATE_COLUMN_NAME;
         String alias      = WINDOWED_AGGREGATE_COLUMN_NAME;
+        if (exprNode.attributes.containsKey("alias")) {
+            alias = exprNode.attributes.get("alias");
+        }
         WindowedExpression rankExpr = new WindowedExpression(ExpressionType.AGGREGATE_WINDOWED_RANK,
                                                              partitionbyExprs,
                                                              orderbyExprs,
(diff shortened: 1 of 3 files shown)


BillWhite:
Seems right to me, though it also seems odd that we depend on the name of the table to not disambiguate by column names.  Can a user create a table called VOLT_TEMP_TABLE?  Does that make a difference?

BillWhite:
That's all I have.

wolffcm:
Hmm, there's nothing keeping users from creating a VOLT_TEMP_TABLE, but it's likely to result in some unexpected behavior (even without this fix) since we resolve columns based on table aliases.

I'll have to think about if there's a better way.

wolffcm:
I have been thinking about this.  I'm not sure there is a better way to do this right now.

I've filed a ticket to close the loophole for tables with the special temp table name.

In a new commit I've pushed, I've made the code in PlanAssembler even more selective, and made sure that windowed aggregates have a special name.

@BillWhite Please have a look and let me know what you think.

BillWhite:
This seems safer, though not completely safe.  The only thing I can think of is to mangle table names, like compilers mangle variable and function names.  Then users and internal names come from different namespaces, and cannot collide.  This seems very intrusive, however.

wolffcm:
Yes, you're right.  I filed a ticket to make the temp table name we use be more magical.  That will address this completely.  

As it turns out, this issue will affect any query that uses RANK and that does not use the source table's first column (the one with column index 0).  So I think the benefit we get from fixing this is more than the very small risk that someone is doing a RANK query on a table called VOLT_TEMP_TABLE, and they have the special conditions that give us trouble when disambiguating column references, which is really vanishingly unlikely.

PR #3861, opened 2016-08-22
Eng 10821 xenial valgrind

BillWhite:
Cause valgrind to write its output in XML, and then parse the XML.

Having structured output from valgrind makes it easier to interpret the valgrind reports programmatically.


tests/frontend/org/voltdb/regressionsuites/EEProcess.java
```diff
@@ -192,11 +202,12 @@ public void run() {
         }
 
         /*
-         * Create a thread to parse Valgrind's output and populdate
+         * Create a thread to parse Valgrind's output and populate
          * m_valgrindErrors with errors.
          */
         final Process p = m_eeProcess;
         m_stdoutParser = new Thread() {
+            ValgrindXMLParser m_xmlParser = null;
```
BillWhite: This is a bug.  I think this is never referenced or used.  I'll delete it.

I just noticed it.

tests/frontend/org/voltdb/utils/TestValgrindParser.java
```diff
@@ -0,0 +1,127 @@
...
+        Arrays.sort(files);
+        for (File file : files) {
+            String baseName = file.getName();
+            // Remove the valgrind version and extension.
+            String valgrindVersion = baseName.substring(0, baseName.indexOf('_'));
+            String programName = baseName.substring(baseName.indexOf('_') + 1, baseName.length()-4);
+            System.out.printf("Testing valgrind version %s with program %s\n", valgrindVersion, programName);
+            valgrindErrors.clear();
+            ValgrindXMLParser.processValgrindOutput(file, valgrindErrors);
+            switch (programName) {
```
BillWhite: The assertions below come from inspecting the xml files.  They are the errors I expect to find in these failing tests.

tests/frontend/org/voltdb/utils/valgrind_test_files/3.10.0_definite_losses.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0"?>
```
BillWhite: This and all the xml files below are generated by valgrind.  They are not all that
interesting to read in detail.

tests/frontend/org/voltdb/regressionsuites/EEProcess.java
```diff
@@ -23,6 +23,7 @@
 package org.voltdb.regressionsuites;
 
 import java.io.BufferedReader;
+import java.io.File;
```
wolffcm: I'm not familiar with this class.  Could you add a header comment for the class?  It seems to be used only for running valgrind on JUnit tests, is that right?  If so, it seems to have a rather general name.

wolffcm:
I seem to recall you mentioning that we need two parsers, one in Python (for EE unit tests) and one in Java, for JUnit tests.  Do I have that right?

wolffcm:
So this works for both U16.04 and the other OSes with older versions of valgrind too?  And the output that we'll see in ant and Jenkins is still as good as before?

Overall, looks great, thanks for doing this.  I had questions, but mostly just for my own education.

wolffcm:
That's all my comments.

tests/frontend/org/voltdb/regressionsuites/EEProcess.java
BillWhite: I think that's true.  It has two publicly available member functions, which are noops when not running with valgrind.

BillWhite:
You are right.  There are two XML parsers.  I suppose there could be one, but it seemed less intrusive to have them both.   For example, if the python program called a Java application somehow, then one would have to ensure that they Java application was built before the EE could be tested.  This is possible, but seems tricky, and one more coordination point between the EE and the Java sides.

wolffcm:
Hey Bill, this last commit that completes the fix for ENG-10972 looks good to me.  Can you merge it to master on its own branch?  I'm assuming that the Xenial valgrind stuff doesn't need to be in for 6.6 so I don't want that to hold it up.

wolffcm:
Or, now that we have branched, could you merge this last commit to release-6.6.x?

wolffcm:
I mean, when it's blue of course.

PR #3864, opened 2016-08-23
Add logic to nestedLoppIndex executor.

yzhang1991:
We added a logic last year in index scan executor to skip pending delete tuples. This is useful when using a compiled query plan to find a new min/max value.
This year we have this on join view, but sometimes we have to use inline index scan that is in nestLoopIndex executor. But in this executor, we did not have this logic to skip pending delete tuple.


wolffcm:
Ok, I think that this looks good.

I will merge it when it's blue.

PR #3865, opened 2016-08-24
ENG-11001 snapshots are archived on init --force

vtkstef:


vtkstef:
related to https://github.com/VoltDB/pro/pull/1473

src/frontend/org/voltdb/settings/PathSettings.java
```diff
@@ -0,0 +1,196 @@
...
+import org.voltdb.VoltDB;
+import org.voltdb.utils.MiscUtils;
+import org.voltdb.utils.VoltFile;
+
+import com.google_voltpatches.common.collect.ImmutableList;
+import com.google_voltpatches.common.collect.ImmutableMap;
+import com.google_voltpatches.common.collect.ImmutableSet;
+import com.google_voltpatches.common.collect.ImmutableSortedMap;
+
+@Sources({"file:${org.voltdb.config.dir}/path.properties", "file:${org.voltdb.config.dir}/../.paths"})
```
akhanzode: Does this migrate .paths to path.properties automatically and delete .paths when done?

src/frontend/org/voltdb/settings/PathSettings.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Key(DR_OVERFLOW_PATH_KEY)
+    public File getDROverflow();
+
+    public static PathSettings create(Map<?, ?>...imports) {
+        return ConfigFactory.create(PathSettings.class, imports);
+    }
+
+    default File resolve(File path) {
+        return path.isAbsolute() ? path : new VoltFile(getVoltDBRoot(), path.getPath());
```
akhanzode: You dont have to use VoltFile as this is for new cli only right?

src/frontend/org/voltdb/settings/PathSettings.java
```diff
@@ -0,0 +1,196 @@
...
+        File renameTo = new File(snapshotDH.getPath() + ".1");
+        try {
+            Files.move(snapshotDH.toPath(), renameTo.toPath());
+        } catch (IOException e) {
+            throw new SettingsException("failed to rename " + snapshotDH  + " to " + renameTo, e);
+        }
+        if (!snapshotDH.mkdir()) {
+            throw new SettingsException("failed to create snapshot directory " + snapshotDH);
+        }
+        return true;
```
akhanzode: we should log on console when we archive that their new location is....

src/frontend/org/voltdb/settings/PathSettings.java
```diff
@@ -0,0 +1,196 @@
...
+        NavigableMap<Integer, File> snapdirs = mb.build();
+        for (Map.Entry<Integer, File> e: snapdirs.descendingMap().entrySet()) {
+            File renameTo = new File(snapshotDH.getPath() + "." + (e.getKey() + 1));
+            try {
+                Files.move(e.getValue().toPath(), renameTo.toPath());
+            } catch (IOException exc) {
+                throw new SettingsException("failed to rename " + e.getValue()  + " to " + renameTo, exc);
+            }
+        }
+        File renameTo = new File(snapshotDH.getPath() + ".1");
```
akhanzode: snapshot directory should be empty after right? So why not keep it empty and not move it. If you do What if .1 exists?

akhanzode:
review done.

src/frontend/org/voltdb/settings/PathSettings.java
vtkstef: good point about deleting the old. This list also reflects precedence so path.properties will be read before .paths in a first to set wins merge strategy

src/frontend/org/voltdb/settings/PathSettings.java
vtkstef: that was my wish but jenkins spanked me silly. I could remove the VoltFile now because VoltDBRoot is a VoltFile

src/frontend/org/voltdb/settings/PathSettings.java
vtkstef: in line 92 I don't move it if its empty

PR #3866, opened 2016-08-24
ENG-11043: EE exceptions doing updates on MV

wolffcm:
Fix for ENG-11043, which was an EE issue with loop counters on min/max aggs.

Also made it so assertTablesAreEqual does what it's supposed to.  I added an optional epsilon argument for the places that use it.

Fix assertTablesAreEqual turned up another bug that went uncaught in TestMaterializedViewSuite, for which I have filed ENG-11048.


src/ee/storage/MaterializedViewHandler.cpp
```diff
@@ -310,13 +312,18 @@ namespace voltdb {
...
                         assert(false); // Should have been caught when the matview was loaded.
                         // no break
                 }
             }
+
+            if (aggType == EXPRESSION_TYPE_AGGREGATE_MIN
+                || aggType == EXPRESSION_TYPE_AGGREGATE_MAX) {
+                minMaxColumnIndex++;
+            }
+
```
BillWhite: So the only semantic difference here is that minMaxColumnIndex is incremented even if deltaValue is NULL.  Is that right?

BillWhite:
I think  I understand why the old code would have failed, and why this fixes it.

So, it seems right to me.

src/ee/storage/MaterializedViewHandler.cpp
wolffcm: Yes, that's it exactly.

wolffcm:
Closing so I can merge to release-6.6.x instead.

PR #3873, opened 2016-08-24
ENG-11048: Fix issue with wrong answers when deleting tuples from MV source tables

wolffcm:
This is analogous to similar issues we saw with MIN/MAX and update.  Here is that change, which is already merged:

https://github.com/VoltDB/voltdb/pull/3864/commits/6ab93a73aa23b9bf2c4030c8eca74847a2260e16

### DIFF
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -991,13 +991,15 @@ void PersistentTable::deleteTuple(TableTuple &target, bool fallible) {
     // handle any materialized views, insert the tuple into delta table,
     // then hide the tuple from the scan temporarily.
     insertTupleIntoDeltaTable(target, fallible);
-    BOOST_FOREACH (auto viewHandler, m_viewHandlers) {
-        viewHandler->handleTupleDelete(this, fallible);
-    }
-
-    // This is for single table view.
     {
         SetAndRestorePendingDeleteFlag setPending(target);
+
+        // for multi-table views
+        BOOST_FOREACH (auto viewHandler, m_viewHandlers) {
+            viewHandler->handleTupleDelete(this, fallible);
+        }
+
+        // This is for single table view.
         for (int i = 0; i < m_views.size(); i++) {
             m_views[i]->processTupleDelete(target, fallible);
         }
--- a/tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
@@ -1640,16 +1640,14 @@ public class TestMaterializedViewSuite extends RegressionSuite {
             verifyViewOnJoinQueryResult(client);
         }
 
-// This part of the test commented out until ENG-11048 is fixed
-//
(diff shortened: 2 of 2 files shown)


yzhang1991:
Looks good

src/ee/storage/persistenttable.cpp
```diff
@@ -991,13 +991,15 @@ void PersistentTable::deleteTuple(TableTuple &target, bool fallible) {
...
-    // This is for single table view.
     {
         SetAndRestorePendingDeleteFlag setPending(target);
+
+        // for multi-table views
+        BOOST_FOREACH (auto viewHandler, m_viewHandlers) {
+            viewHandler->handleTupleDelete(this, fallible);
+        }
+
+        // This is for single table view.
```
paulmartel: Minor: It would be more consistent to also use BOOST_FOREACH for the single-table loop just below.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -1640,16 +1640,14 @@ public void testViewOnJoinQuery() throws IOException, ProcCallException
...
-//
-//        Collections.shuffle(dataList1);
-//        System.out.println("Now testing deleting data from the join query view source table.");
-//        for (int i=0; i<dataList1.size(); i++) {
-//            deleteRow(client, dataList1.get(i));
-//            verifyViewOnJoinQueryResult(client);
-//        }
+        // -- 5 -- Test deleting the data from the source tables.
+
+        Collections.shuffle(dataList1);
```
paulmartel: You mentioned in stand-up that reversing the list was a good deterministic stand-in for shuffle. It would be nice to see the code for that option commented out here.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -1640,16 +1640,14 @@ public void testViewOnJoinQuery() throws IOException, ProcCallException
...
+        System.out.println("Now testing deleting data from the join query view source table.");
+        for (int i=0; i<dataList1.size(); i++) {
```
paulmartel: It was broke in the comment as you found it, but... use a for-each loop.

paulmartel:
End of review round, for what it's worth.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
wolffcm: Sorry I missed your comments, Paul.  I must have merged just after you began reviewing.

If I need to change this file again I'll try to remember to make this change.

BillWhite: FWIW, you could also supply the source of the randomness to the shuffle operation.  That way you could get repeatable random shuffles.  (Not that it matters all that much, as reversing works too.)

BillWhite:
That's all I have, which is not much.

PR #3874, opened 2016-08-24
ENG-10994: update windowing example with all partition APIs

zeemanhe:


benjaminballard:
This looks good, and much much simpler.  One question I have is why is the PartitionDataTracker still necessary?  It seems all it is doing now is keeping a count of how many partitions there are, but I don't see where that is used.

zeemanhe:
The partition size is used to calculate the target rows per partition, which is then used in procedure DeleteOldestToTarget.  I agree that  PartitionDataTracker is not necessary here.  The target rows per partition can be initialized and and continuously updated with every delete. Thanks to the all partition call API which returns # of partitions.  PartitionDataTracker is removed.

examples/windowing/client/windowing/ContinuousDeleter.java
```diff
@@ -55,118 +52,87 @@
...
-                long tuplesDeleted = clientResponse.getResults()[0].asScalarLong();
-                app.addToDeletedTuples(tuplesDeleted);
-                // If the procedue deleted up to its limit, reduce the time before
-                // the deletes process runs again.
-                if (tuplesDeleted >= app.config.deletechunksize) {
-                    unifinished.set(true);
-                }
+        public void clientCallback(ClientResponseWithPartitionKey[] clientResponse) throws Exception {
+
+            if (clientResponse == null) {
```
jhugg: Why should this ever be null?

examples/windowing/client/windowing/ContinuousDeleter.java
```diff
@@ -55,118 +52,87 @@
...
-                // Un-comment to see problems on the console
-                //System.err.println(clientResponse.getStatusString());
 
-                // Track failures in a simplistic way.
-                failureCount.incrementAndGet();
+            app.updatePartitionInfo(clientResponse.length);
+
+            for (ClientResponseWithPartitionKey resp: clientResponse) {
+
+                if (resp.getResponse() != null && resp.getResponse().getStatus() == ClientResponse.SUCCESS) {
```
jhugg: Why would resp.getResponse() ever be null?

jhugg: For that matter, I thought we were getting rid of accessors? Did those get put back in?

examples/windowing/client/windowing/ContinuousDeleter.java
```diff
@@ -55,118 +52,87 @@
...
-                }
-                catch (Exception e) {
-                    // Track failures in a simplistic way.
-                    failureCount.incrementAndGet();
-                    // Make sure the latch is released for this call, even if failed
-                    latch.countDown();
-                }
+            // Send the procedure call to all partitions and get results from each partitions.
+            if (app.config.historyseconds > 0) {
+                app.client.callAllPartitionProcedure(new CallBack(unfinished),
```
jhugg: Where do we actually wait for the callback to be called? We just check unfinished below, but it's unlikely this proc will be done. The old version uses a latch.

Why use the async api at all? Why not just use the sync API?

examples/windowing/client/windowing/WindowingApp.java
```diff
@@ -162,9 +148,9 @@ TimestampType getTargetDate() {
         return new TimestampType(targetTimestampMillis * 1000);
     }
 
-    void updatePartitionInfo(Map<Long, PartitionInfo> partitionData) {
-        this.partitionData.set(partitionData);
-        targetRowsPerPartition.set(config.maxrows / partitionData.size());
+    void updatePartitionInfo(long partitionSize) {
```
jhugg: Maybe this should be called "updatePartitionCount".

examples/windowing/client/windowing/WindowingApp.java
```diff
@@ -313,18 +297,19 @@ public void run() {
         System.out.println(" Starting Processing");
         System.out.println(HORIZONTAL_RULE);
 
+        try {
```
jhugg: So this code is just to get started? There should be a comment. Also, what if it fails? Should the program stop? Currently it just keeps going.

examples/windowing/client/windowing/ContinuousDeleter.java
xinjiacs: What's the reason not to add accessor?

jhugg: Less code. Make it a struct-y thing, not a fancy class.

jhugg:
This looks a lot better to me. Thanks.

PR #3875, opened 2016-08-25
Use the windowed RANK() function in the voters example.

BillWhite:
The query to calculate the heat rank in the voters example used to
calculate the heat map by rummaging through the rollup of votes by
contestant_number and state, keeping track of the winner and the total
number of votes.  This commit uses two SQL queries in one stored
procedure to avoid the java computation.  The first statement uses
RANK() to determine which contestant is the winner in a given state.
The second vote counts the number of votes.  The GetStateHeatmap stored
procedure returns both tables.

The dashboard index.html file has been updated to take advantage of
these changes.


examples/voter/procedures/voter/GetStateHeatmap.java
```diff
@@ -35,57 +35,26 @@
...
-                    ((Result)resultArray[i]).votes,
-                    ((Result)resultArray[i]).isWinning });
-        }
+    public final SQLStmt stateHeatMap = new SQLStmt(
+                                                    "SELECT STATE, CONTESTANT_NUMBER, NUM_VOTES "
+                                                  + "FROM ( SELECT STATE, NUM_VOTES, CONTESTANT_NUMBER, "
+                                                  + "       RANK() OVER ( PARTITION BY STATE ORDER BY NUM_VOTES DESC ) VRANK "
+                                                  + "       FROM V_VOTES_BY_CONTESTANT_NUMBER_STATE ) AS SUB "
+                                                  + "WHERE SUB.VRANK = 1; ");
+    public final SQLStmt contestantTotals = new SQLStmt("SELECT * from V_VOTES_BY_CONTESTANT_NUMBER;");
```
paulmartel: Should maintain stylistically consistent with other voter sql that uses lowercased schema names and (usually) string concatenating ` + ` operators before the line break and a +8-space indent on statement continuations (vs. align with open paren).
For simplicity, decide on an ordering of `STATE, CONTESTANT_NUMBER, NUM_VOTES` and stick with it.

examples/voter/ddl.sql
```diff
@@ -65,6 +80,18 @@ AS
...
+    contestant_number
+,   num_votes
+)
+AS
+  SELECT contestant_number
+       , COUNT(*)
+    FROM votes
+GROUP BY contestant_number
+;
+
```
paulmartel: This additional view is going to have throughput impact -- maybe instead the query should take the hit of rolling up from the existing v_votes_by_contestant_number_state view.

paulmartel:
That's all my comments.

examples/voter/ddl.sql
benjaminballard: This is a good point.  I think the majority of the time people run this example app, they don't bring up the dashboard.  Unless the throughput impact is very minimal, it makes more sense to incur additional performance cost only when you show the dashboard.

examples/voter/ddl.sql
```diff
@@ -1,3 +1,18 @@
...
+DROP PROCEDURE ContestantWinningStates         IF EXISTS;
+DROP PROCEDURE GetStateHeatmap                 IF EXISTS;
+DROP VIEW  v_votes_by_phone_number             IF EXISTS;
+DROP VIEW  v_votes_by_contestant_number        IF EXISTS;
+DROP VIEW  v_votes_by_contestant_number_state  IF EXISTS;
+DROP TABLE contestants                         IF EXISTS;
+DROP TABLE votes                               IF EXISTS;
+DROP TABLE area_code_state                     IF EXISTS;
+
+END_OF_DROP_BATCH
```
benjaminballard: I like that you've introduced this DROP BATCH section.  I use it in the template that I use for building apps, and I find it really helps when you are actively developing and making frequent schema changes to have an idempotent DDL script.  While not necessary as part of showing the RANK() function, I welcome this change.

examples/voter/procedures/voter/GetStateHeatmap.java
BillWhite: By my count there are 3 lines with plusses at the end of the line,  in ContestantWinningStates.java, and 10 lines with plusses at beginning of the line, in Results.java.  I prefer the beginning of the line, because I think they are easier to see.  But I've already changed them to the end of the line, because my preference is not as strong as yours.

examples/voter/ddl.sql
BillWhite: Ok.  That's easily done.

examples/voter/ddl.sql
BillWhite: I don't usually do batches like this, but I almost always DROP all the tables, indices and PROCEDURES at the beginning of DDL files, for the reasons you already know.  I may have learned it from you, who knows?

examples/voter/ddl.sql
BillWhite: Ah, well, benchmarking seems to think this is a bad idea.  I've tested the implementation with this view, with the rollup and with version currently on master.  The average number of transactions are:

master with dashboard: 41363 T/S
master w/o dashboard: 41749 T/S
rollup with dashboard: 20814 T/S
rollup w/o dashboard: 43192 T/S
view with dashboard: 41423 T/S
view w/o dashboard: 42450 T/S

I'll attach the raw benchmark output, but it looks like the rollup causes more trouble than it solves.  In fact, comparing the view with master shows that the view costs almost nothing, but the rollup in combination with the dashboard is very expensive.  I suspect this is because the dashboard's GetStateHeatMap blocks the database periodically, and the rollup is expensive.

[voter_rank.txt](https://github.com/VoltDB/voltdb/files/481251/voter_rank.txt)

BillWhite: I have created the git branch ENG-10760-RollupRankExample-notest with the change which rolls up the contestant-by-state counts for the dashboard in the stored procedure without a view.  The branch ENG-10760-RankExample, which is this branch, contains the view and does not roll up the vote counts by state.  You can look at my rollup implementation if you wish.

BillWhite:
Ok, the best way to do this is to replace the view with global counts with a select statement
on the votes-by-state view.  I compared the performance of this branch to master, and it seems
to be almost identical, at least to my eye.  The file below has the benchmark output.

[voter_rank.txt](https://github.com/VoltDB/voltdb/files/483645/voter_rank.txt)

benjaminballard:
Bill, the further changes you made look good.  I approve.

PR #3876, opened 2016-08-25
Eng 10810 jenkinsbot better commands

jwiebema1:


jwiebema1:
merging femi's latest changes.

PR #3877, opened 2016-08-25
ENG-10968: run-everywhere updates

zeemanhe:


src/frontend/org/voltdb/client/ClientResponseWithPartitionKey.java
```diff
@@ -26,46 +26,15 @@
  */
 public class ClientResponseWithPartitionKey {
 
-    final Object m_partitionKey;
-    final ClientResponse m_response;
-    final String m_errorMessage;
-    final Exception m_cause;
+    final public Object m_partitionKey;
```
jhugg: Don't use m_ for public final vars.

src/frontend/org/voltdb/client/AllPartitionProcedureCallback.java
```diff
@@ -26,8 +26,8 @@
     /**
      * Implementation of callback to be provided by client applications.
      *
-     * @param clientResponse Responses for each partition to the stored procedure invocation this callback is associated with
+     * @param response Responses for each partition to the stored procedure invocation this callback is associated with
      * @throws Exception on any Exception.
      */
-    abstract public void clientCallback(ClientResponseWithPartitionKey[] clientResponse) throws Exception;
+     public void clientCallback(ClientResponseWithPartitionKey[] response) throws Exception;
```
jhugg: "responses". Plural.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -936,14 +946,7 @@ public OnePartitionProcedureCallback(AtomicInteger counter, Object partitionKey,
...
-            if (response != null) {
-                m_responses[m_index] = new ClientResponseWithPartitionKey(m_partitionKey, response);
-            } else {
-                final ClientResponse r = new ClientResponseImpl(ClientResponse.GRACEFUL_FAILURE, new VoltTable[0],
-                        "The procedure is not queued for execution.");
-                m_responses[m_index] = new ClientResponseWithPartitionKey(m_partitionKey, r,
-                        r.getStatusString(), null);
-            }
+            m_responses[m_index] = new ClientResponseWithPartitionKey(m_partitionKey, response);
             m_partitionCounter.decrementAndGet();
```
jhugg: Could combine this line with the next one.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -903,16 +906,11 @@ private void refreshPartitionKeys() throws IOException, NoConnectionsException,
     * @throws NoConnectionsException if this {@link Client} instance is not connected to any servers.
     * @throws IOException if there is a Java network or connection problem.
     */
-    public List<Integer> getPartitionIntegerKeys() throws NoConnectionsException, IOException, ProcCallException {
-        if (m_distributer.getPartitionKeys().size() > 0) {
+    public ImmutableSet<Integer> getPartitionIntegerKeys() throws NoConnectionsException, IOException, ProcCallException {
+        if (m_distributer.getPartitionKeys() != null && m_distributer.getPartitionKeys().size() > 0) {
```
jhugg: Why do we need two sets of partition keys that are the same structure? It seems the one in this class is used if client affinity is off, and the distributer one is used if it's on?

Why not just put the refresh logic in the distributor? Add a last checked timestamp, and refresh if CA is off and the timestamp is old in the distributor's version of getPartitionKeys()?

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -1452,7 +1454,37 @@ long getProcedureTimeoutNanos() {
...
+            try {
+                cb.waitForResponse();
+            } catch (final InterruptedException e) {
+                throw new java.io.InterruptedIOException("Interrupted while waiting for response");
+            }
+            if (cb.getResponse().getStatus() != ClientResponse.SUCCESS) {
+                throw new ProcCallException(cb.getResponse(), cb.getResponse().getStatusString(), null);
+            }
+            updatePartitioning(cb.getResponse().getResults()[0]);
+            m_lastPartitionKeyFetched.set(System.currentTimeMillis());
```
jhugg: Can you just use "time" from line 1473?

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -1452,7 +1454,37 @@ long getProcedureTimeoutNanos() {
         return m_procedureCallTimeoutNanos;
     }
 
-    ImmutableSet<Integer> getPartitionKeys() {
+    ImmutableSet<Integer> getPartitionKeys() throws NoConnectionsException, IOException, ProcCallException {
+        //null check: the topology update uses async update, thus partition info may not be ready upon client connection.
+        if (!m_useClientAffinity || m_partitionKeys.get() == null) {
+            refreshPartitionKeys();
```
jhugg: Does this ever get called more than once?

jhugg:
So one test I would like to see added (sorry for not asking for this sooner) is a test 
1. N tuples are loaded
2. Use run everywhere to count them.
3. Elastically expand cluster.
3. Use run everywhere to count tuples and expect same value.

zeemanhe:
Will add test cases to cover cluster elastic expansion

src/frontend/org/voltdb/client/Distributer.java
zeemanhe: if client affinity is on, refreshPartitionKeys may be called once here. The following updates are done upon topology update.

if client affinity is off, refreshPartitionKeys will be called if the last fetched set is more than 1 second old.

src/frontend/org/voltdb/client/Distributer.java
zeemanhe: using System.nanoTime() ?

tests/frontend/org/voltdb/client/TestAllPartitionProcedureCalls.java
```diff
@@ -161,14 +229,26 @@ public void testCallAllPartitionProcedureFailuerProc() throws Exception {
...
 
-    private void validateResults(ClientResponseWithPartitionKey[]  responses, Map<Integer, Long> expectedCounts) {
+     private void validateResults(ClientResponseWithPartitionKey[]  responses, Map<Integer, Long> expectedCounts) {
+        long total = 0;
         for (ClientResponseWithPartitionKey resp: responses) {
             VoltTable results = resp.response.getResults()[0];
             Long expected = expectedCounts.get(resp.partitionKey);
             assert(expected != null);
-            assertTrue(expected.longValue() == results.fetchRow(0).getLong(0));
+            long count = results.fetchRow(0).getLong(0);
```
jhugg: Should we count the number of rows to make sure it's the expected number?

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -1396,14 +1400,15 @@ private void updateProcedurePartitioning(VoltTable vt) {
     }
 
     private void updatePartitioning(VoltTable vt) {
-        m_partitionKeys.clear();
+        List<Integer> ketSet = new ArrayList<Integer>();
```
jhugg: is ketSet a typo? keySet? Or am I missing what ket is?

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -1449,7 +1454,37 @@ long getProcedureTimeoutNanos() {
         return m_procedureCallTimeoutNanos;
     }
 
-   public  List<Integer> getPartitionKeys() {
-        return m_partitionKeys;
+    ImmutableSet<Integer> getPartitionKeys() throws NoConnectionsException, IOException, ProcCallException {
+        //null check: the topology update uses async update, thus partition info may not be ready upon client connection.
+        if (!m_useClientAffinity || m_partitionKeys.get() == null) {
+            refreshPartitionKeys();
```
jhugg: I don't think this will get called repeatedly?

tests/frontend/org/voltdb/client/TestAllPartitionProcedureCalls.java
```diff
@@ -26,49 +26,73 @@
...
+        STRING_PARTITION_EXPECTED_COUNTS.put(new Integer(11), new Long(125));
         STRING_PARTITION_EXPECTED_COUNTS.put(new Integer(15), new Long(114));
         STRING_PARTITION_EXPECTED_COUNTS.put(new Integer(19), new Long(125));
-        STRING_PARTITION_EXPECTED_COUNTS.put(new Integer(23), new Long(118));
+        STRING_PARTITION_EXPECTED_COUNTS.put(new Integer(23), new Long(122));
+    }
+
+    static Map<Integer, Long> ELASTIC_JOIN_INT_PARTITION_EXPECTED_COUNTS = new HashMap<Integer, Long>();
+    static {
+        ELASTIC_JOIN_INT_PARTITION_EXPECTED_COUNTS.put(new Integer(0), new Long(78));
```
jhugg: Where do these numbers come from? Are these just the results of the tests?

This test seems really fragile if it relies on using the exact same hash forever. It would be enough to check the sum of the partitions is the right value, the number of the partitions is the right value, and maybe that none of them are empty. That would be a lot more robust to changes in our code.

zeemanhe: Test cases were updated.  Now tested with two clients, one with client affinity on, one with client affinity off.   Both of the clients pick up the topology change from cluster expansion.

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -210,6 +217,9 @@ public void clientCallback(ClientResponse clientResponse) throws Exception {
                     e.printStackTrace();
```
jhugg: I don't think this is really the error handling we want. Printing to stdout from the client is never something we do. If an exception is thrown, we need to make sure all-partition procedures fail. We should also retry.

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -201,11 +213,13 @@ public void clientCallback(ClientResponse clientResponse) throws Exception {
                             updatePartitioning(results[0]);
                         }
                     }
-                }
-                catch (Exception e) {
+                } catch (Exception e) {
                     e.printStackTrace();
```
jhugg: We don't want to handle errors by printing stacks to stderr. 

We want to update the internal state, and return errors when people try to run all partition procs.

src/frontend/org/voltdb/client/Distributer.java
```diff
@@ -191,20 +198,27 @@ public void clientCallback(ClientResponse clientResponse) throws Exception {
...
         @Override
         public void clientCallback(ClientResponse clientResponse) throws Exception {
             if (clientResponse.getStatus() == ClientResponse.SUCCESS) {
-                try {
-                    synchronized (Distributer.this) {
-                        VoltTable results[] = clientResponse.getResults();
-                        if (results != null && results.length > 0) {
-                            updatePartitioning(results[0]);
-                        }
+                synchronized (Distributer.this) {
```
jhugg: Does this have to be synchronized? What isn't thread-safe?

zeemanhe: error from copying TopoUpdateCallback. removed.

PR #3879, opened 2016-08-25
ENG-11056: Fix logic error in MaterializedViewHandler

wolffcm:
The symptom was that a view appeared to be nonempty even when all data was deleted from its source tables.  That is, until the transaction is committed.  After the current quantum is released, the view appears empty as it should.

This happened because logic in `handleTupleForDelete` was first deleting the tuple and then updating it.  We should do one or the other but not both.  The processing of updating the tuple changed the "pending delete" bit, such that it was visible again.  But once the transaction committed, the release of the undo action made the view appear empty again.

### DIFF
--- a/src/ee/common/NValue.cpp
+++ b/src/ee/common/NValue.cpp
@@ -131,7 +131,7 @@ std::string NValue::debug() const {
     std::ostringstream buffer;
     std::string out_val;
     const char* ptr;
-    int64_t addr;
+
     buffer << getTypeName(type) << "::";
     switch (type) {
     case VALUE_TYPE_BOOLEAN:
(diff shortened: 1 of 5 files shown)


src/ee/common/NValue.cpp
```diff
@@ -131,7 +131,7 @@ std::string NValue::debug() const {
     std::ostringstream buffer;
     std::string out_val;
     const char* ptr;
-    int64_t addr;
+
```
BillWhite: Does eliminating this temporary really save anything?  Won't the compiler do what it wants with the addresses anyway?  It might make it more difficult to see the values you are manipulating, though.  (This is a tiny comment.)

src/ee/storage/MaterializedViewHandler.cpp
```diff
@@ -381,11 +381,13 @@ namespace voltdb {
...
-            // since their keys shouldn't ever change, but do update other indexes.
-            m_destTable->updateTupleWithSpecificIndexes(m_existingTuple, m_updatedTuple,
-                                                        m_updatableIndexList, fallible);
+            else {
+                mergeTupleForDelete(deltaTuple);
+                // Shouldn't need to update group-key-only indexes such as the primary key
+                // since their keys shouldn't ever change, but do update other indexes.
+                m_destTable->updateTupleWithSpecificIndexes(m_existingTuple, m_updatedTuple,
+                                                            m_updatableIndexList, fallible);
+            }
```
BillWhite: This part is the important change, right?  We do one or the other, but not both.

BillWhite:
Makes sense to me.  But I'm not 100% sure I understand what's going on.  Somebody else should review it as well.

src/ee/common/NValue.cpp
wolffcm: For some reason we were converting addresses to int64_t so we would display them in base-10.  I'd rather see them in hexadecimal to match what my (and I think everyone else's) debugger does.

The change I made was just to show addresses in hexadecimal.

src/ee/storage/MaterializedViewHandler.cpp
wolffcm: Yes, that's right.

yzhang1991: Oh.....This makes me feel very stupid. But yeah, I think it should be in the way it is here.

PR #3880, opened 2016-08-25
ENG-11052 view sum null

yzhang1991:
SUM should be null when view source is empty.


wolffcm:
Merge this to release-6.6.x instead of master!  😃

wolffcm:
I'm seeing changes from other commits, but the change to fix the issue looks okay.

Did you write a unit test for this?  If you don't have time, let me know and I can finish it up.

yzhang1991:
In this change set I have already changed the view definition of ORDER_COUNT_GLOBAL to include SUM, MIN and MAX.
It will cover the case of no source table data while we are inserting / deleting data from the source tables.
So I think we do not need to add extra test cases for it.

wolffcm:
Ok, I see that you changed the base branch, but I still see 49 commits, which is not right.

Do you want to fix this?  If I don't hear back by the end of iteration review I'll go ahead and fix it.

yzhang1991:
Hi Chris, I rebased the branch, feel free to merge it if everything looks good to you and Jenkins is blue.

wolffcm:
Will do, thanks Ethan.

src/ee/storage/MaterializedViewHandler.cpp
```diff
@@ -290,41 +290,57 @@ namespace voltdb {
...
-                        // no break
+        NValue newValue;        
+        if (newCount.isZero()) {
+            // no group by key, no rows, aggs will be null except for count().
+            for (int aggIndex = 0, columnIndex = aggOffset; aggIndex < m_aggColumnCount; aggIndex++, columnIndex++) {
+                if (m_aggTypes[aggIndex] == EXPRESSION_TYPE_AGGREGATE_COUNT) {
+                    newValue = ValueFactory::getBigIntValue(0);
+                }
+                else {
+                    newValue = ValueFactory::getNullValue();
```
yzhang1991: I have a doubt here.
I saw in our single table view code I also did this `ValueFactory::getNullValue();`. But I am wondering should we mark explicitly the data type of this NValue. Currently this NValue's data type would be `VALUE_TYPE_NULL`

PR #3882, opened 2016-08-26
ENG-10358, don't add a pending delete tuple back to elastic index dur…

luyangco:
…ing force compaction.

If a tuple is marked as pending delete, it means COW context holds on it and no one can access it anyone.
So it should not be added back to elastic index.

Change-Id: I6ea3599b9a07e4b8221b3573231cb1c7bdfefc6a

### DIFF
--- a/src/ee/storage/ElasticContext.cpp
+++ b/src/ee/storage/ElasticContext.cpp
@@ -250,13 +250,14 @@ void ElasticContext::notifyTupleMovement(TBPtr sourceBlock,
                                          TableTuple &targetTuple)
 {
     if (m_indexActive) {
-        StreamPredicateList &predicates = getPredicates();
-        assert(predicates.size() > 0);
         if (m_surgeon.indexHas(sourceTuple)) {
             m_surgeon.indexRemove(sourceTuple);
-        }
-        if (predicates[0].eval(&targetTuple).isTrue()) {
-            m_surgeon.indexAdd(targetTuple);
+            // If the tuple is pending delete, it's held on by COW but
+            // shouldn't be accessible anymore. So don't add it back to
+            // elastic index.
+            if (!targetTuple.isPendingDelete()) {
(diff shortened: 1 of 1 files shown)


src/ee/storage/ElasticContext.cpp
```diff
@@ -255,7 +255,10 @@ void ElasticContext::notifyTupleMovement(TBPtr sourceBlock,
         if (m_surgeon.indexHas(sourceTuple)) {
             m_surgeon.indexRemove(sourceTuple);
         }
-        if (predicates[0].eval(&targetTuple).isTrue()) {
+        // If the tuple is pending delete, it's held on by COW but
+        // shouldn't be accessible anymore. So don't add it back to
+        // elastic index.
+        if (!targetTuple.isPendingDelete() && predicates[0].eval(&targetTuple).isTrue()) {
```
nshi: This certainly does the trick. I'm wondering if it makes more sense to not insert a tuple if it was not in the index, meaning that the check in the if block above fails. The reason is that we have explicit notifications for insertion, update, and deletion. When a tuple is moved, if it was not already in the index, we don't care about it.

nshi: If it's already in the index, then `predicates[0].eval(&targetTuple).isTrue()` is guaranteed to be true because moving a tuple should not change the hash of that tuple.

luyangco: Upsource complains it can't sync with Github, repost here.

Yes, I think it totally make sense. `notifyTupleMovement()` can be simplified by only checking elastic index once.
```
        if (m_surgeon.indexHas(sourceTuple)) {
            m_surgeon.indexRemove(sourceTuple);
            if (!targetTuple.isPendingDelete()) {
                m_surgeon.indexAdd(targetTuple);
            }
        }
```

luyangco:
I did a force push to address the comments.

PR #3883, opened 2016-08-26
Iteration over sets is not order deterministic in Java.

BillWhite:
The member function MaterializedViewFixInfo.processMVBasedQueryFix() iterated over a set of columns, expecting the result to be order deterministic.  There is a list which has the same set of columns available, so we now just iterate over that list.


wolffcm:
Merged on a separate branch.

PR #3884, opened 2016-08-26
Eng 10509

hhakimi53:
Pull request for initial commit of docker files. This uses voltdb community edition.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -1,0 +1,71 @@
+FROM ubuntu:14.04
+# exit on error
+RUN set -e
+
+# Install VoltDB 
+# update repo - Trusty does not seem to have add-repo command by default
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get upgrade -y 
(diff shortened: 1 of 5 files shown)


tools/docker/Dockerfile
```diff
@@ -0,0 +1,62 @@
+FROM ubuntu:14.04
```
vtkstef: is there an LTS version of this?

tools/docker/deployment.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<deployment>
+    <cluster sitesperhost="2" kfactor="0" />
+    <httpd enabled="true">
+        <jsonapi enabled="true" />
+    </httpd>
+    <!--paths>
```
vtkstef: remove commented stanza

tools/docker/Dockerfile
hhakimi53: That's an LTS image. For trusty docker image, there are tags for image named 14.04.5, 14.04, trusty-20160819, trusty and looks like they all point to the same image based on the repo.

tools/docker/deployment.xml
hhakimi53: I missed cleaning this one. It was there for testing with different permutations. Thnxs

PR #3889, opened 2016-08-28
Eng 10608 sqlcov join mview

johnTcrawford:
SqlCoverage tests of materialized views defined using joins.


tests/scripts/SQLGenerator.py
```diff
@@ -222,7 +256,7 @@ def set_min_max(self, longmin, longmax, latmin=None, latmax=None):
     def set_nullpct(self, nullpct):
         self.__nullpct = nullpct
 
-    def set_num_holes(self, num_holes):
+    def set_count(self, num_holes):
```
wolffcm: Is this a typo?  Seems like you want to set `__count` here?

wolffcm: Or maybe you are repurposing this variable to work with both randomly generated polygons and integers?  If so, this approach seems pretty confusing.

tests/scripts/examples/sql_coverage/joined-matview-DDL.sql
```diff
@@ -0,0 +1,215 @@
...
+  SELECT   T1.VCHAR, T2.NUM,   COUNT(*),   MAX(T2.ID),      SUM(T1.RATIO) FROM R1 T1 JOIN R2 T2 ON T1.NUM = T2.NUM
+  GROUP BY T1.VCHAR, T2.NUM;
+
+CREATE VIEW V23 (NUM, VCHAR, NUM2, COUNT_STAR, ID, RATIO, RATIO2) AS
+  SELECT   T1.NUM, T1.VCHAR, T2.NUM, COUNT(*), MIN(T2.ID), MAX(T1.RATIO), SUM(T2.RATIO) FROM P1 T1 JOIN R1 T2 ON T1.ID = T2.ID
+  GROUP BY T1.NUM, T1.VCHAR, T2.NUM;
+
+
+-- "Abnormal" Materialized Views using Joins that are not equi-joins (cannot use 2 partitioned tables)
+CREATE VIEW V24 (NUM, ID, NUM2) AS
```
wolffcm: These should fail to plan?  Or not?  I though that we had a restriction that joins in views must be equi-joins.

wolffcm:
That's all my comments.   I don't find anything that should keep us from merging this, although I did have some questions.

tests/scripts/examples/sql_coverage/joined-matview-DDL.sql
johnTcrawford: I have a vague recollection of someone saying in a meeting that all joins in views must be equip-joins, but that was not my experience when I tried it - I believe it works.

tests/scripts/SQLGenerator.py
johnTcrawford: Yes, exactly. And yes, you're right, it's confusing. I should definitely fix it - after version 6.6 goes out, and after my vacation (I'll create a ticket).

PR #3893, opened 2016-08-30
ENG-11074: Correctly set column size for view columns

wolffcm:
We did not set the size and "in bytes" attribute in the catalog correctly for some views.  This resulted in a difference in the output schema of the query used to populate the view and view itself.  This was a problem when the was a mismatch of inlined and non-inlined variable length columns.

The fix was just to set the size attributes properly in the DDL compiler.


src/frontend/org/voltdb/compiler/MaterializedViewProcessor.java
```diff
@@ -239,6 +239,11 @@ public void startProcessing(Database db, HashMap<Table, String> matViewMap, Tree
                     if (! col.expression.getValueType().isVariableLength()) {
                         destColumn.setSize(col.expression.getValueType().getMaxLengthInBytes());
                     }
+                    else {
```
yzhang1991: So this does not affect single-table view?
Look at line 356.

Both here and line 356 are added during the joined table view work, did this issue happen before with single table view?

yzhang1991: And line 377

tests/testprocs/org/voltdb_testprocs/regressionsuites/matviewprocs/matviewsuite-ddl.sql
```diff
@@ -378,3 +378,30 @@ CREATE VIEW V21 (NUM, VCHAR, COUNT_STAR, RATIO, ID) AS
...
+       ID INTEGER DEFAULT '0' NOT NULL,
+       VCHAR VARCHAR(64 BYTES),
+       VCHAR_INLINE_MAX VARCHAR(15),
+       VCHAR_INLINE VARCHAR(42 BYTES),
+       RATIO FLOAT NOT NULL,
+       PRIMARY KEY (ID)
+);
+PARTITION TABLE P2_ENG_11074 ON COLUMN ID;
+
+CREATE VIEW V_ENG_11074 (ID, VCHAR_INLINE) AS
```
paulmartel: For added clarity, I'd name this view VJOIN_ENG_11074.

src/frontend/org/voltdb/compiler/MaterializedViewProcessor.java
paulmartel: Normally avoid:

    if (! condition) {
        ...
    }
    else {
        ....
    }

paulmartel: I prefer using the combined setter method tve.setTypeSizeBytes. I think longer term we should retire the individual setter methods to avoid this kind of mishap  -- BTW, I think my someday perf branch renamed this to something more readable like tve.setTypeSizeAndInBytes. we could also provide an alternative tve.setFixedSizedType to internalize the fixedType.getMaxLengthInBytes() ugliness. We could add internal asserts that these setters were being used correctly for fixed/variable-length type constants.

paulmartel:
Apart from Ethan's comment (good catch, Ethan!), looks good.

paulmartel:
The new changes look OK short term. Longer term, I think we should just let the widths of view columns and view rows in particular be unbounded when the max possible width of their input values can not be determined. That way we avoid unexpected consequences when view refreshes fail.

PR #3896, opened 2016-08-31
ENG-11085 submit offset before importer calls the formatter

vtkstef:


src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -129,6 +131,13 @@ private PartitionMetadata findLeader() {
         return returnMetaData;
     }
 
+    private int nextCorrelationId() {
+        FetchRequest fr = m_fetchRequestBuilder.addFetch(m_topicAndPartition.topic(),
+                m_topicAndPartition.partition(), 1L, m_config.getFetchSize())
+                .build();
+        return fr.correlationId();
```
akhanzode: why do you need to do fetchRequest correletion id is just opaque number can be just incrementing number right?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -450,7 +456,8 @@ protected void accept() {
                     ByteBuffer payload = messageAndOffset.message().payload();
 
                     String line = new String(payload.array(),payload.arrayOffset(),payload.limit(),StandardCharsets.UTF_8);
-                    try{
+                    try {
+                        m_gapTracker.submit(messageAndOffset.nextOffset());
```
akhanzode: if callProcedure fails because of cluster is paused or pausing this should unsubmit right?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -459,11 +466,11 @@ protected void accept() {
                               if (isDebugEnabled()) {
                                  debug(null, "Failed to process Invocation possibly bad data: " + line);
                                }
-                               m_gapTracker.commit(currentOffset);
+                               m_gapTracker.commit(messageAndOffset.nextOffset());
```
akhanzode: why commit next offset and not current?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -459,11 +466,11 @@ protected void accept() {
...
                          }
                      } catch (FormatException e){
                         rateLimitedLog(Level.WARN, e, "Failed to tranform data: %s" ,line);
-                        m_gapTracker.commit(currentOffset);
+                        m_gapTracker.commit(messageAndOffset.nextOffset());
```
akhanzode: see previous comment.

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
vtkstef: so we make sure they are distinct since the fetch builder uses an incrementing atomic integer which we cannot superseed

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
vtkstef: the commit is not the current. it is what we would need to read next would we start again

PR #3901, opened 2016-09-01
Use NOTDISTINCT for searching for GB keys in fallback query

wolffcm:
This simple change fixes a wrong answer that occurs when we delete the current MAX value for a group that has a GB key that is NULL.


tests/testprocs/org/voltdb_testprocs/regressionsuites/matviewprocs/matviewsuite-ddl.sql
```diff
@@ -378,3 +378,14 @@ CREATE VIEW V21 (NUM, VCHAR, COUNT_STAR, RATIO, ID) AS
...
   FROM R2_ENG_11024 T1 JOIN P2_ENG_11024 T2 ON T1.ID = T2.NUM
   GROUP BY T2.NUM, T2.VCHAR;
+
+-- Repro for ENG-11080
+CREATE VIEW V_ENG_11080 (VCHAR, COUNT_STAR, ID) AS
+  SELECT   T1.VCHAR, COUNT(*),   MAX(T1.ID)
+  FROM R1_ENG_11024 T1
+  GROUP BY T1.VCHAR;
+
+CREATE VIEW VJOIN_ENG_11080 (VCHAR, COUNT_STAR, ID) AS
```
paulmartel: Nice trick using self-join to trivially trigger the join view code path with minimal side effect.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -1873,6 +1873,46 @@ public void testEng11047() throws Exception {
...
+
+        for (String stmt : stmts) {
+            client.callProcedure("@AdHoc", stmt);
+        }
+
+        VoltTable vt;
+
+        // SELECT list of both views is:
+        //   VCHAR, COUNT(*), MAX(ID)
+        // At this point both views have just one group, where the GB key is NULL.
```
paulmartel: Defining expectedResultBeforeDelete and expectedResultAfterDelete array constants here would make it easier to see what is expected to change and that the two views are expected to contain the same content.

src/frontend/org/voltdb/compiler/MaterializedViewProcessor.java
```diff
@@ -729,6 +729,9 @@ public void processMaterializedViewWarnings(Database db, HashMap<Table, String>
                     // there must be some min/max columns.
                     // Only check if the plan uses index scan.
                     if (needsWarningForSingleTableView( getPlanNodeTreeFromCatalogStatement(db, stmt))) {
+                        // If we are using IS NOT DISTINCT FROM as our equality operator (which is necessary
+                        // to get correct answers), then there will often be no index scans in the plan,
+                        // since we cannot optimize IS NOT DISTINCT FROM.
```
paulmartel: This is likely to hurt performance. We should at least make a ticket to provide indexscan AND NLIJ support for  IS NOT DISTINCT FROM filters.

paulmartel:
End of review round.

wolffcm:
@paulmartel I added a couple more commits.  One that incorporates your feedback and another more substantial one.

I made it so we use NOT DISTINCT for views with joins (new features so no performance regression, but should be release noted), and EQUAL for single-table views, switching at runtime to a manual scan.

This limits performance regression to just the case where we're recomputing MIN or MAX in the presence of a NULL GB column.

paulmartel:
Looks good.

PR #3909, opened 2016-09-07
ENG-11100: Fix exception safety issue with delta tuple insertion

wolffcm:
In this bug we were violating a NOT NULL constraint when inserting the delta tuple.  The resulting exception caused us to skip cleaning up the tuple being inserted in the actual, persistent table, so that the insert effectively succeeded, leading to the view and persistent table getting out of sync.


tests/testprocs/org/voltdb_testprocs/regressionsuites/matviewprocs/matviewsuite-ddl.sql
```diff
@@ -416,3 +426,9 @@ CREATE VIEW VJOIN_ENG_11080 (VCHAR, COUNT_STAR, ID) AS
   SELECT   T1.VCHAR, COUNT(*),   MAX(T1.ID)
   FROM R1_ENG_11024 T1 JOIN R2_ENG_11024 T2 ON T1.NUM = T2.NUM
   GROUP BY T1.VCHAR;
+
+-- Repro for ENG-11100
+CREATE VIEW V24_ENG_11100 (VCHAR_INLINE_MAX, ID, VCHAR_INLINE_MAX2) AS
```
paulmartel: V24?

tests/testprocs/org/voltdb_testprocs/regressionsuites/matviewprocs/matviewsuite-ddl.sql
```diff
@@ -380,6 +380,7 @@ CREATE VIEW V21 (NUM, VCHAR, COUNT_STAR, RATIO, ID) AS
   GROUP BY T2.NUM, T2.VCHAR;
 
 -- Repro for ENG-11074
+-- Thest tables are used in the "string" suite of sqlcoverage matview join tests.
```
paulmartel: typo: Thest

paulmartel:
Looks good -- I'm assuming that the delta table metadata has no distinguishing marks vs. the original table so catching the NULL on the delta table vs. the original table will give exactly the same error message detail.

wolffcm:
Good point.  Yes, the error message produced by the constraint violation thrown by inserting into the delta table is the same as the one we would get for inserting into the actual persistent table.  Seems like we should really have a different order of actions here, but the way the code is factored currently makes that tricky.

PR #3912, opened 2016-09-08
Eng 10842 mixed sph

luyangco:
Companion pro PR https://github.com/VoltDB/pro/pull/1483

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -790,6 +790,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
                     managedPathsEmptyCheck(config);
             }
 
+            // Write local sitesPerHost to ZK
             int sitesperhost = m_catalogContext.getDeployment().getCluster().getSitesperhost();
             registerSitesPerHostToZK(sitesperhost);
 
@@ -907,8 +908,9 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
                 List<Integer> partitions = null;
                 if (isRejoin) {
                     m_configuredNumberOfPartitions = m_cartographer.getPartitionCount();
+                    int sitesPerHost = clusterConfig.getSitesPerHostMap().get(m_messenger.getHostId());
                     partitions = m_cartographer.getIv2PartitionsToReplace(m_configuredReplicationFactor,
-                                                                          clusterConfig.getSitesPerHost());
+                                                                          sitesPerHost);
                     if (partitions.size() == 0) {
                         VoltDB.crashLocalVoltDB("The VoltDB cluster already has enough nodes to satisfy " +
                                 "the requested k-safety factor of " +
@@ -2449,15 +2451,15 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
         } else if (m_myHostId == 0) {
             m_clusterSettings.store(m_messenger.getZK());
         }
-        ClusterConfig config = new ClusterConfig(
-                m_clusterSettings.get().hostcount(),
-                clusterType.getSitesperhost(),
-                clusterType.getKfactor()
-                );
-
-        if (!config.validate()) {
-            VoltDB.crashLocalVoltDB("Cluster parameters failed validation: " + config.getErrorMsg());;
-        }
+//        ClusterConfig config = new ClusterConfig(
+//                m_clusterSettings.get().hostcount(),
+//                clusterType.getSitesperhost(),
+//                clusterType.getKfactor()
+//                );
+//
+//        if (!config.validate()) {
+//            VoltDB.crashLocalVoltDB("Cluster parameters failed validation: " + config.getErrorMsg());;
+//        }
         m_clusterCreateTime = m_messenger.getInstanceId().getTimestamp();
         return determination;
     }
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/compiler/ClusterConfig.java
```diff
@@ -148,19 +156,27 @@ public int getHostCount()
...
+            totalSites += entry.getValue();
+        }
+        return totalSites;
+    }
+
     public int getPartitionCount()
     {
-        return (m_hostCount * m_sitesPerHost) / (m_replicationFactor + 1);
+        return getTotalSitesCount() / (m_replicationFactor + 1);
     }
```
zeemanhe: Will the partition count be rounded here?  Say if replication factor = 1 and cluster has a total of 5 sites.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2404,15 +2443,15 @@ public Integer get() {
...
-        }
+//        ClusterConfig config = new ClusterConfig(
+//                m_clusterSettings.get().hostcount(),
+//                clusterType.getSitesperhost(),
+//                clusterType.getKfactor()
+//                );
+//
+//        if (!config.validate()) {
+//            VoltDB.crashLocalVoltDB("Cluster parameters failed validation: " + config.getErrorMsg());;
+//        }
```
zeemanhe: these comments should be removed.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1564,10 +1588,23 @@ private JSONObject getTopology(StartAction startAction, Map<Integer, String> hos
             topo = joinCoordinator.getTopology();
         }
         else if (!startAction.doesRejoin()) {
-            int sitesperhost = m_catalogContext.getDeployment().getCluster().getSitesperhost();
+            Map<Integer, Integer> sphMap = new HashMap<>();
+            try {
+                List<String> children;
+                    children = m_messenger.getZK().getChildren(VoltZK.sitesPerHost, false);
```
zeemanhe: combine line 1593 and 1594

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1553,6 +1559,24 @@ public void run() {
...
     }
 
+    private void registerSitesPerHostToZK(int sitesperhost) {
+        try {
+            ZKUtil.addIfMissing(m_messenger.getZK(), VoltZK.sitesPerHost, CreateMode.PERSISTENT, new byte[0]);
+        } catch (Exception e) {
+            VoltDB.crashLocalVoltDB("Unable to create " + VoltZK.sitesPerHost + " node to Zookeeper, dying", false, e);
+        }
+        String path = ZKUtil.joinZKPath(VoltZK.sitesPerHost, "host_" + m_messenger.getHostId());
+        try {
```
zeemanhe: Would it be better using host id instead of "host_"+ host id as the path so that no splitting is needed late on?

src/frontend/org/voltdb/compiler/ClusterConfig.java
```diff
@@ -488,23 +505,31 @@ JSONObject fallbackPlacementStrategy(
             List<Integer> hostIds,
             int hostCount,
             int partitionCount,
-            int sitesPerHost) throws JSONException{
+            Map<Integer, Integer> sitesPerHostMap) throws JSONException{
         // add all the sites
```
zeemanhe: hostIds  is not  needed since sitesPerHostMap should have the host list.

src/frontend/org/voltdb/compiler/ClusterConfig.java
```diff
@@ -124,10 +125,10 @@ public static void addPartitions(JSONObject topo, Multimap<Integer, Integer> par
         }
     }
 
-    public ClusterConfig(int hostCount, int sitesPerHost, int replicationFactor)
+    public ClusterConfig(int hostCount, Map<Integer, Integer> sitesPerHostMap, int replicationFactor)
     {
```
zeemanhe: hostCount = sitesPerHostMap.size. hostCount param can be removed.

src/frontend/org/voltdb/compiler/ClusterConfig.java
luyangco: I think unless we allow mixed K-safety, total sites count should always be multiple of `K+1`, otherwise it won't pass the check, see https://github.com/VoltDB/voltdb/pull/3912/files/8d844e0c549242a97c85136c0080dc1e9c857f3d#diff-481e240b57493a1c8d06ba24bad368a0L217

src/frontend/org/voltdb/RealVoltDB.java
luyangco: Good idea, I'll remove the prefix `host_`.

src/frontend/org/voltdb/compiler/ClusterConfig.java
luyangco: No,  I can't. The JSONObject `topo` still needs `hostId`

src/frontend/org/voltdb/compiler/ClusterConfig.java
luyangco: If we support rejoin, then `sitesPerHostMap.size()` might not equal to `hostCount`, e.g. 2 nodes left the cluster, then rejoin one back.

PR #3913, opened 2016-09-09
fix for fetching system messages file on Redhat/CentOS distribution with no lsb_release

hhakimi53:


src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -277,6 +280,37 @@ private static void parseJSONObject(JSONObject jsonObject) {
         }
     }
 
+    private static String getLinuxOSInfo() {
+        // Supported Linux OS for voltdb are CentOS, Redhat and Ubuntu
+        String versionInfo = "";
+        try {
```
akhanzode: Use try with resource

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -277,6 +280,37 @@ private static void parseJSONObject(JSONObject jsonObject) {
...
+            BufferedReader br = null;
+            // base Ubuntu distributions have the distribution info in /etc/lsb-release
+            final String ubuntuDistInfoFilePath = "/etc/lsb-release";
+            // base RedHat, CentOS distributions have the distribution info in /etc/redhat-release
+            final String rhDistInfoFilePath = "/etc/redhat-release";
```
akhanzode: Make this into a list so you can iterate till you find one. This way we can add to list.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -326,19 +360,23 @@ private static void parseJSONObject(JSONObject jsonObject) {
             }
 
             String systemLogBase;
+            final String systemLogBaseDirPath = "/var/log/";
```
akhanzode: This is only for Linux right so can be in else.

src/frontend/org/voltdb/utils/Collector.java
hhakimi53: Sure.

src/frontend/org/voltdb/utils/Collector.java
hhakimi53: Sure

src/frontend/org/voltdb/utils/Collector.java
hhakimi53: This is for both Mac and Linux.

PR #3919, opened 2016-09-19
Drop do-nothing TupleSerializer class hierarchy and declarations.h

paulmartel:
Note: declarations.h had become an arbitrary pile of random forward declares, some of them obscure or even obsolete. It's cleaner to inline forward declarations in the few places they are missing. declarations.h was apparently intended originally for template instantiations, but there are better case-specific ways to deal with those as well.

The sole (non-)effect of the TupleSerializer hierarchy was to ignore hidden columns in some unit tests that actually defined NO hidden columns that needed ignoring. 
In production, it injected virtual function wrappers around public TableTuple and TupleSchema methods trivially adding a hard-coded boolean argument to each.


paulmartel:
Merged this manually from the command line -- not sure why.

PR #3920, opened 2016-09-19
Tool to run strace on a running VoltDB process

rmorgenstein:
I know I don't need a review to push a tool we don't ship, but I appreciate any python tips, ideas, corrections that anyone wants to provide.

This tools will find the VoltDB pid, or take a pid of any JVM-based process and run strace on its children. The output from strace replaces all the child pids with the thread names.

### DIFF
--- a/tools/stracevoltdb.py
+++ b/tools/stracevoltdb.py
@@ -16,53 +16,87 @@ def find_voltdb_pid():
         return None
 
 
-desc = """Create an strace command for some or all threads in a JVM
- The best way to run this is:
-    eval $(./stracethread.py)
- If you want to add strace arguments, then it would be:
-    eval $(./stracethread.py -s "-tt -o strace.out -e trace=\!futex") &
+def find_threads(pid):
+# Jstack the pid and create a dictionary of threads/pids
+
+    threads={}
+    jstack_cmd = 'jstack ' + args.pid
+    pattern_jstack=re.compile('"(.+)".*nid=(0x\S+)\s')
+
+    try:
+        for line in run_cmd(jstack_cmd, False):
+            m = pattern_jstack.match(line)
+            if m:
+                #We need the pids in base10, but need them as strings
+                threads[m.group(1)] = str(int(m.group(2),16))
+
+    #TODO: Do some better error handling here.
+    except subprocess.CalledProcessError as e:
+        print e
+        sys.exit("Cannot jstack pid " + args.pid)
+
+    return threads
+
+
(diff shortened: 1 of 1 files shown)


tools/stracevoltdb.py
```diff
@@ -0,0 +1,125 @@
...
+except:
+    sys.exit ('ERROR: ' + args.regex + ' is not a valid regex')
+
+
+# Find the threads and make dict of pids.
+threads = find_threads(args.pid)
+
+
+#Find the ones that match the regex
+matching_threads = {key:value for key, value in threads.items() if re.search(pattern_thread, key)}
```
rmorgenstein: Centos 6 has Python2.6, which doesn't have dictionary comprehensions. Rewrite

tools/stracevoltdb.py
```diff
@@ -0,0 +1,125 @@
...
+
+if args.list:
+    exit()
+
+
+def replacepid(m):
+    return ('{0:5} - {1:25}'.format(m.group(1), pids[m.group(1)][:25]))
+
+#TODO: This could use better error handling for ctrl-c
+for line in run_cmd(strace_cmd, stderr_to_stdout=True):
```
rmorgenstein: The output leaves a lot to be desired when strace isn't installed or can't be run without sudo.

tools/stracevoltdb.py
```diff
@@ -0,0 +1,125 @@
...
+    popen.stdout.close()
+    return_code = popen.wait()
+    if return_code != 0:
+        raise subprocess.CalledProcessError(return_code, cmd)
+
+
+desc = """Run strace on VoltDB (or another JVM-based process) and show output with
+thread names for all the activity.  Threads can be narrowed down by a regex.
+    Example - strace all threads with "Snap" in the name, but exclude futex calls
+    /stracethread.py -s "-e trace=\!futex" Snap.\*
```
pdshaw: ./stracethread.py

tools/stracevoltdb.py
```diff
@@ -0,0 +1,125 @@
...
+    print
+    print "Strace command:"
+    print strace_cmd
+
+if args.list:
+    exit()
+
+
+def replacepid(m):
+    return ('{0:5} - {1:25}'.format(m.group(1), pids[m.group(1)][:25]))
```
pdshaw: I prefer not to intermix methods and "main" code. Suggest you move this up and add a explicit "main":
   #### Command line main
   if __name__ == '__main__':
   ... the main stuff

PR #3923, opened 2016-09-20
add debug message when something bad happened

xinjiacs:


nshi:
This probably shouldn't go onto master because it records every single message the SpScheduler sees. Can you ask Phil to run the system test on your branch?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1285,15 +1286,28 @@ private long getRepairLogTruncationHandleForReplicas()
...
+        m_ll.add(new RuntimeException("Updating truncation point from " +
+                TxnEgo.txnIdToString(m_repairLogTruncationHandle) +
+                "to" + TxnEgo.txnIdToString(newHandle)));
+
         assert newHandle >= m_repairLogTruncationHandle : "new handle: " + newHandle + ", repairLog:" + m_repairLogTruncationHandle;
+
         if (newHandle > m_repairLogTruncationHandle) {
             m_repairLogTruncationHandle = newHandle;
 
-            if (! m_isLeader) {
```
nshi: You should not remove this line. It keeps the replicas' truncation point static until the leader tells them to update.

xinjiacs: Should we move this IF check to the beginning of this function? The replica should not even update itself as well, only waiting leader messages. @nshi

xinjiacs:
Leaving the debug messages on branch.

PR #3927, opened 2016-09-22
analytic-template.sql: added DENSE_RANK, as an alternative to RANK, in

johnTcrawford:
tests of SQL Analytic (Window) functions.


johnTcrawford:
I'm not worried about excessive run times now, with just RANK and DENSE_RANK, but as we add more functions, the 2 current templates that use two analytic functions, one in the subquery and one outside it, may need to be changed to always use the same function, to avoid too many combinations, and therefore too many queries being run, too slowly.

PR #3928, opened 2016-09-22
simpler serializing anticipating LRR serializer integration

paulmartel:
Gutted ExportSerializeIo because its input serializer was not really needed and its output serializer had so much redundancy with the generic SerializeOutput base class -- from which it now inherits.
The export output serializer may go away or become trivial if/when we templatize the generic SerializeOutput class for endianness -- endian differences are the REAL novel feature of export serialization.

In NValue.hpp and tabletuple.h and serializeio.h, got rid of the clumsy explicit enum-based templating.
Some of this was replaced with dedicated functions for DR and non-DR that delegate to template code (only) for truly common functionality.
This allowed template functions to be called more cleanly with implied argument-driven parameterization.

In serializeio.h, I replaced the input serializers' use of explicit branching based on a dumb Endianess enum with a one-liner delegating to a smart Endian class parameter. The parameter classes use template and overload methods to encapsulate any required optional byte re-ordering.
As I just mentioned, the same kind of templating should probably be done (trivially) for the output serializers so that export output can be brought more fully into the fold.
I got rid of some unused virtual functions and the virtuality of some functions that are called only on known leaf types -- there are still a few slightly trickier cases to fix here involving optional output buffer expansion.
I got rid of some class specializations (and unit testing thereof) that weren't adding value.

### DIFF
--- a/buildtools.py
+++ b/buildtools.py
@@ -505,7 +505,10 @@ def buildMakefile(CTX):
     makefile.write("\t$(AR) $(ARFLAGS) $@ $?\n")
     harness_source = TEST_PREFIX + "/harness.cpp"
     makefile.write("objects/harness.o: $(ROOTDIR)/" + harness_source + "\n")
-    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $^\n")
+    # The dependency magic injects harness.h et. al. into the dependencies of
+    # harness.o. So, it's a good idea to explicitly repeat the source path on
+    # the compiler command line rather than trying to pick it up with $^.
+    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $(ROOTDIR)/" + harness_source + "\n")
     makefile.write("-include %s\n" % "objects/harness.d")
     makefile.write("\n")
     cleanobjs += ["objects/volt.a", "objects/harness.o", "objects/harness.d"]
--- a/src/ee/common/ExportSerializeIo.h
+++ b/src/ee/common/ExportSerializeIo.h
@@ -46,269 +46,98 @@
 #ifndef NATIVEEXPORTSERIALIZEIO_H
 #define NATIVEEXPORTSERIALIZEIO_H
 
+#include "serializeio.h"
+
 #include <cassert>
-#include <cstring>
 #include <limits>
-#include <stdint.h>
 #include <string>
 
 namespace voltdb {
 
 /*
  * This file defines a crude Export serialization interface.
- * The idea is that other code could implement these method
- * names and duck-type their way to a different Export
- * serialization .. maybe doing some dynamic symbol finding
- * for a pluggable Export serializer. It's a work in progress.
- *
- * This doesn't derive from serializeio to avoid making the
- * the serialize IO baseclass functions all virtual.
+ * It could be much simplified if the SerializeOutput implementation
+ * in serializeio.h was an Endian-aware template.
  */
 
-class ExportSerializeInput {
+class ExportSerializeOutput : public SerializeOutput {
   public:
-
-    ExportSerializeInput(const void* data, size_t length)
+    ExportSerializeOutput(void *buffer, size_t capacity)
     {
-        current_ = reinterpret_cast<const char*>(data);
-        end_ = current_ + length;
+        initialize(buffer, capacity);
     }
 
+    ~ExportSerializeOutput() { }
 
-    virtual ~ExportSerializeInput() {};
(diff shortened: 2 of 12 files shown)


buildtools.py
```diff
@@ -505,7 +505,10 @@ def buildMakefile(CTX):
     makefile.write("\t$(AR) $(ARFLAGS) $@ $?\n")
     harness_source = TEST_PREFIX + "/harness.cpp"
     makefile.write("objects/harness.o: $(ROOTDIR)/" + harness_source + "\n")
-    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $^\n")
+    # The dependency magic injects harness.h et. al. into the dependencies of
+    # harness.o. So, it's a good idea to explicitly repeat the source path on
+    # the compiler command line rather than trying to pick it up with $^.
+    makefile.write("\t$(CCACHE) $(COMPILE.cpp) -MMD -MP -o $@ $(ROOTDIR)/" + harness_source + "\n")
```
BillWhite: This is really tiny, but I would probably use the os pathname tool to join "$(ROOTDIR)" and harness_source.  I'd also format with %, but that's just me.

src/ee/common/NValue.hpp
```diff
@@ -2869,21 +2870,53 @@ inline void NValue::serializeToTupleStorage(void *storage, bool isInlined,
...
+            throwFatalException("Unexpected scale %d", scale);
+        }
+        if (precisionBytes != 16) {
+            throwFatalException("Unexpected number of precision bytes %d", precisionBytes);
+        }
+        uint64_t *longStorage = reinterpret_cast<uint64_t*>(storage);
+        //Reverse order for Java BigDecimal BigEndian
+        // Serialize to export serializes them in network byte order, have to reverse them here
+        longStorage[1] = ntohll(input.readLong());
+        longStorage[0] = ntohll(input.readLong());
```
nshi: This looks correct since it was carried over from existing code. I wonder if it makes more sense to take a `SerializeInputBE` as input since DR buffers are transferred over the network and let the input deal with the byte order conversion. It is quite confusing that only the decimal type is serialized in big endian in the existing `ExportSerializeOutput` and not other multi-byte types. It looks like that you've changed it to always serialize all types in big endian.

src/ee/common/ExportSerializeIo.h
```diff
@@ -46,269 +46,98 @@
...
- * for a pluggable Export serializer. It's a work in progress.
- *
- * This doesn't derive from serializeio to avoid making the
- * the serialize IO baseclass functions all virtual.
+ * It could be much simplified if the SerializeOutput implementation
+ * in serializeio.h was an Endian-aware template.
  */
 
-class ExportSerializeInput {
+class ExportSerializeOutput : public SerializeOutput {
```
nshi: By inheriting the `SerializeOutput` class, I think you've changed the serialization byte order from little endian to big endian for all types, except for decimal which is already in big endian. This means that for all the DR deserialization paths, we should take `SerializeInputBE` as the input stream type.

buildtools.py
paulmartel: @BillWhite I just followed the existing syntax of the previous line. I agree that preference for formatting with % IS just you.

src/ee/common/NValue.hpp
paulmartel: @nshi No I don't believe I changed anything. Is a jenkins branch run not enough to prove this? See my next comment.

src/ee/common/ExportSerializeIo.h
paulmartel: As far as I can tell, I did not change anything. The SerializeOutput methods that do the endian swapping are all re-implemented in terms of ExportSerializeOutput's own writePrimitive NOT to do the swapping. There just needs to be some care taken not to up-cast ExportSerializeOutput as SerializeOutput because the functions are not virtual. 
It's possible that this may be one of those rare use cases for private or protected inheritance.
Eventually, we should internalize this optional endian functionality into SerializeOutput as an Endian class parameter just like SerializeInput has. This will make implicit casting to the wrong type impossible.

nshi: `ExportSerializeOutput` used to just serialize data as is, which would be in little endian for x86. `SerializeOutput` forces everything to be in big endian when it passes data to `SerializeOutput::writePrimitive()`.

I agree that it'll be much nicer to internalize this into `SerializeOutput`.

The DR tests, both C++ unit tests and the Java unit tests are failing on your branch, quite possibly because of this.

PR #3932, opened 2016-09-23
ENG-11154: Internal server initiated txns should be allowed in paused…

manjujames:
… mode.

### DIFF
--- a/src/frontend/org/voltdb/compiler/AsyncCompilerWork.java
+++ b/src/frontend/org/voltdb/compiler/AsyncCompilerWork.java
@@ -28,6 +28,7 @@ public class AsyncCompilerWork implements Serializable {
     }
     private static final long serialVersionUID = 6588086204761949082L;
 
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/compiler/AsyncCompilerWork.java
```diff
@@ -28,6 +28,7 @@
     }
     private static final long serialVersionUID = 6588086204761949082L;
 
+    private boolean serverInitiated;
```
nshi: Does it make sense to treat all internal connections as admin connections?

manjujames: No, I don't think so. For example, import uses internal connections. I thought it is safe for now to just do this for CommandLogInitiator.

nshi: Makes sense.

PR #3933, opened 2016-09-23
Eng 10539 optimize fragment result processing

wweiss-voltdb:
Added an extra cache buffer between the EE and Volt that is used to supply results. One result buffer is used for the first batch result. The second buffer is used for all other results. If final is set or this is the first result, result buffer will not be copied into a heap byte buffer. Instead the volt tables are created directly on top of the result buffer. All non-final batches after the first batch are still copied.

Note that if any of the tables in the first batch result or the final batch result are returned by the stored procedure, they will be copied into a heap buffer as part of the response generation.

In addition, the conversion of the result to a volt table is deferred until we determine whether the batch is for the procedure runner or for a fragment result. This avoids the unnecessary generation of a VoltTable for Fragment Results.

Also changed the allocator for the index to use fixed size buffers (same size as table blocks) and vary the number of indexes per buffer. Also allocate from the thread local buffer pool rather than the shared heap. Finally change index to hang on to the last deleted buffer rather than returning all buffers and keeping the first buffer of the index. This avoids poor performance when the table size varies by a single row  (add one, delete one, add one, delete one) that happens to fall on an index buffer boundary.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -523,7 +523,7 @@ int VoltDBEngine::executePlanFragment(int64_t planfragmentId,
     // the result buffer
     if (last) {
         m_resultOutput.writeBoolAt(m_startOfResultBuffer, m_dirtyFragmentBatch);
-        m_resultOutput.writeIntAt(m_startOfResultBuffer, static_cast<int32_t>(m_resultOutput.position() - m_startOfResultBuffer) - sizeof(int32_t) - sizeof(int8_t) + 1);
+        m_resultOutput.writeIntAt(m_startOfResultBuffer+1, static_cast<int32_t>(m_resultOutput.position() - m_startOfResultBuffer) - sizeof(int32_t) - sizeof(int8_t));
     }
 
     return ENGINE_ERRORCODE_SUCCESS;
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -522,8 +522,8 @@ int VoltDBEngine::executePlanFragment(int64_t planfragmentId,
     // write dirty-ness of the batch and number of dependencies output to the FRONT of
     // the result buffer
     if (last) {
-        m_resultOutput.writeIntAt(m_startOfResultBuffer, static_cast<int32_t>(m_resultOutput.position() - m_startOfResultBuffer) - sizeof(int32_t) - sizeof(int8_t));
-        m_resultOutput.writeBoolAt(m_startOfResultBuffer + sizeof(int32_t), m_dirtyFragmentBatch);
+        m_resultOutput.writeBoolAt(m_startOfResultBuffer, m_dirtyFragmentBatch);
+        m_resultOutput.writeIntAt(m_startOfResultBuffer, static_cast<int32_t>(m_resultOutput.position() - m_startOfResultBuffer) - sizeof(int32_t) - sizeof(int8_t) + 1);
     }
 
     return ENGINE_ERRORCODE_SUCCESS;
@@ -1444,7 +1444,8 @@ int32_t VoltDBEngine::getPerFragmentStatsSize() const {
 
 void VoltDBEngine::setBuffers(char* parameterBuffer, int parameterBufferCapacity,
         char* perFragmentStatsBuffer, int perFragmentStatsBufferCapacity,
-        char* resultBuffer, int resultBufferCapacity,
+        char *firstResultBuffer, int firstResultBufferCapacity,
+        char *finalResultBuffer, int finalResultBufferCapacity,
         char* exceptionBuffer, int exceptionBufferCapacity) {
     m_parameterBuffer = parameterBuffer;
     m_parameterBufferCapacity = parameterBufferCapacity;
@@ -1452,8 +1453,11 @@ void VoltDBEngine::setBuffers(char* parameterBuffer, int parameterBufferCapacity
     m_perFragmentStatsBuffer = perFragmentStatsBuffer;
     m_perFragmentStatsBufferCapacity = perFragmentStatsBufferCapacity;
 
-    m_reusedResultBuffer = resultBuffer;
-    m_reusedResultCapacity = resultBufferCapacity;
+    m_firstReusedResultBuffer = firstResultBuffer;
+    m_firstReusedResultCapacity = firstResultBufferCapacity;
+
+    m_finalReusedResultBuffer = finalResultBuffer;
+    m_finalReusedResultCapacity = finalResultBufferCapacity;
 
     m_exceptionBuffer = exceptionBuffer;
     m_exceptionBufferCapacity = exceptionBufferCapacity;
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -1445,8 +1445,8 @@ int32_t VoltDBEngine::getPerFragmentStatsSize() const {
 void VoltDBEngine::setBuffers(char* parameterBuffer, int parameterBufferCapacity,
         char* perFragmentStatsBuffer, int perFragmentStatsBufferCapacity,
         char *firstResultBuffer, int firstResultBufferCapacity,
-        char *finalResultBuffer, int finalResultBufferCapacity,
-        char* exceptionBuffer, int exceptionBufferCapacity) {
+        char *nextResultBuffer, int nextResultBufferCapacity,
+        char *exceptionBuffer, int exceptionBufferCapacity) {
     m_parameterBuffer = parameterBuffer;
     m_parameterBufferCapacity = parameterBufferCapacity;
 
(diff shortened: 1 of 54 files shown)


src/ee/structures/ContiguousAllocator.cpp
```diff
@@ -21,22 +21,24 @@
 
 using namespace voltdb;
 
+#define ALLOCATOR_BLOCKSIZE 2097152
+
 ContiguousAllocator::ContiguousAllocator(int32_t allocSize, int32_t chunkSize)
     : m_count(0),
       m_allocationSize(allocSize),
-      m_numberAllocationsPerBlock(chunkSize),
+      m_numberAllocationsPerBlock((ALLOCATOR_BLOCKSIZE - sizeof(Buffer)) / allocSize),
```
wolffcm: With this change, the parameter `chunkSize` is no longer used, so it would be good to update the signature of this constructor and its callers.

wolffcm:
I've no more comments.  It will be interesting to see how this does in the performance tests.

src/ee/structures/ContiguousAllocator.cpp
```diff
@@ -21,22 +21,24 @@
...
       m_tail(NULL),
       m_blockCount(0),
       m_cachedBuffer(0) {}
 
 ContiguousAllocator::~ContiguousAllocator() {
     while (m_tail) {
         Buffer *buf = m_tail->prev;
-        free(m_tail);
+        ThreadLocalPool::freeExactSizedObject(ALLOCATOR_BLOCKSIZE, m_tail);
```
wolffcm: I don't think this is right.  Currently, the only places we use the ThreadLocalPool's "exact size" API is for StringRef objects (the objects themselves, not the string storage), and for the TupleBlock free list.  Both of these objects are small, just a handful of bytes for each instance.  This memory is managed under the hood by `boost::pool`.

I believe that `boost::pool` will attempt to allocate a buffer that can accommodate thirty-two 2MB chunks.  Moreover, our current use of `boost::pool` will not return unused memory to the OS (even though it will reuse unused chunks).

src/ee/voltdbipc.cpp
```diff
@@ -194,7 +194,8 @@ class VoltDBIPC : public voltdb::Topend {
     void setupSigHandler(void) const;
 
     int m_fd;
-    char *m_reusedResultBuffer;
+    char *m_firstReusedResultBuffer;
+    char *m_finalReusedResultBuffer;
```
paulmartel: Is the "first" reusedResultBuffer is never referenced in this module except to initialize the engine and then clean up. Either the engine somehow uses the buffer internally (even when driven by the voltdbipc which never itself looks at the result there) or it is never actually used in this scenario. If it is never used, pass it as null and/or rename it to put `unused` in the name.

src/frontend/org/voltdb/DependencyPair.java
```diff
@@ -17,15 +17,77 @@
 
 package org.voltdb;
 
-public class DependencyPair {
+import java.nio.ByteBuffer;
 
-    public final int depId;
-    public final VoltTable dependency;
+public abstract class DependencyPair {
```
paulmartel: It COULD stand a class header comment.

src/ee/voltdbjni.cpp
```diff
@@ -551,6 +559,7 @@ SHAREDLIB_JNIEXPORT jint JNICALL Java_org_voltdb_jni_ExecutionEngine_nativeExecu
 (JNIEnv *env,
         jobject obj,
         jlong engine_ptr,
+        jint buffer_hint,
```
paulmartel: We are beyond the point of pretending that the header comments for this function reflect the params in the slightest way. We had probably better delete them if we're not going to bring them and keep them up to date.

src/ee/execution/VoltDBEngine.h
```diff
@@ -551,9 +555,14 @@ class __attribute__((visibility("default"))) VoltDBEngine {
...
-        /** size of reused_result_buffer. */
-        int m_reusedResultCapacity;
+        char* m_firstReusedResultBuffer;
+        /** size of m_reusedResultBuffer. */
+        int m_firstReusedResultCapacity;
+
+        /** buffer object to receive final result tables from EE. */
+        char* m_finalReusedResultBuffer;
+        /** size of m_finalReusedResultBuffer. */
+        int m_finalReusedResultCapacity;
```
paulmartel: There are zero EE comments on the meanings of first, final, and bufferHint, and it's not at all clear. Do first and final even NEED to be named or should there just be an array of the two of them so that the bufferHint becomes a bufferSelector. Then it MAY make sense to say that the buffer selection logic is all in the java and that from the VoltDBEngine perspective, there is nothing special about either of them.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -510,8 +510,8 @@ int VoltDBEngine::executePlanFragment(int64_t planfragmentId,
     // write dirty-ness of the batch and number of dependencies output to the FRONT of
     // the result buffer
     if (last) {
-        m_resultOutput.writeIntAt(m_startOfResultBuffer, static_cast<int32_t>(m_resultOutput.position() - m_startOfResultBuffer) - sizeof(int32_t) - sizeof(int8_t));
-        m_resultOutput.writeBoolAt(m_startOfResultBuffer + sizeof(int32_t), m_dirtyFragmentBatch);
+        m_resultOutput.writeBoolAt(m_startOfResultBuffer, m_dirtyFragmentBatch);
+        m_resultOutput.writeIntAt(m_startOfResultBuffer+1, static_cast<int32_t>(m_resultOutput.position() - m_startOfResultBuffer) - sizeof(int32_t) - sizeof(int8_t));
```
wweiss-voltdb: We used to de-serialize the entire buffer in the ExecutionEngine. However, we now defer the de-serialization until we know that we are not writing it straight to the network and the purpose of the length is to facilitate de-serialization. However, the dirty flag is maintained in the ExecutionEngine so it always needs to be serialized.

src/ee/structures/ContiguousAllocator.cpp
wweiss-voltdb: The trade-off is arguable. One consideration was that mallocs are a point of contention even when the Kernel is not involved. In contrast, ThreadLocalPool is lock-free. Further if the Kernel becomes involved, it requires 2 context switches in KVMs (which most of our deployments use). The other consideration was that indexes have the same volatility as tuple blocks. Using the same size and the same pool is generally a better strategy than maintaining many separate pools. On this point I would argue that if we want to use malloc directly, we should do the same for Tuple Blocks.

src/ee/execution/VoltDBEngine.h
wweiss-voltdb: The objective of the optimization is to avoid Java heap buffers and copies if possible. To that end, a First buffer will not be copied by Java. Instead the reusable direct buffer will be wrapped by a VoltTable and handed to the stored procedure. Similarly the Final buffer will be used for the second through final batches and *will* be copied for every batch except the final batch. This implies that because we make these buffers zero-copy, we must protect these buffers if the procedure runner returns either of the buffers as a return value. Therefore we examine all results from the user procedure and swap a direct (EE) buffer with a copy in java heap space.

src/ee/voltdbjni.cpp
wweiss-voltdb: Params Removed

src/frontend/org/voltdb/TableHelper.java
```diff
@@ -1529,4 +1530,32 @@ public void clientCallback(ClientResponse clientResponse) throws Exception {
             }
         }
     }
+
+    public static ByteBuffer getBackedBuffer(VoltTable table) {
+        // This is an unsafe version of VoltTable.getBuffer() because it does not instantiate a new ByteBuffer
+        table.m_buffer.position(0);
+        table.m_readOnly = true;
```
wweiss-voltdb: If this is called it is because we are now going to serialize a table to the network. At this point we should not expect any more changes to the table that could alter the underlying buffer. Prior to this point, it is fine to add more rows to the table.

src/frontend/org/voltdb/jni/ExecutionEngineJNI.java
```diff
@@ -301,15 +315,32 @@ public void updateCatalog(long timestamp, final String catalogDiffs) throws EEEx
...
-        deserializer.clear();
+        FastDeserializer targetDeserializer = null;
+        if (bufferHint == 0) {
+            firstDeserializer.clear();
+            targetDeserializer = firstDeserializer;
+        }
+        else {
+            finalDeserializer.clear();
+            targetDeserializer = finalDeserializer;
+        }
```
wweiss-voltdb: Done

src/ee/voltdbjni.cpp
wweiss-voltdb: I have also changed all the references to final to more the accurate next (as in non-first) buffer. Also I changed buffer_hint to batchIndex (where batchIndex == 0 implies first buffer). Since only the ProcedureRunner cares about the final buffer, it is a little misleading to use final buffer in the ee.

PR #3936, opened 2016-09-27
ENG-11107 moved OSGi module management out of import manager

vtkstef:


akhanzode:

PR #3938, opened 2016-09-27
ENG-11163. Change stop to stay connected to same node.

rmorgenstein:
Previously, voltadmin stop would connect to the -H host to get SystemInformation OVERVIEW information.  With this information, it would fine the hostID of the target to stop, then select another host and connect to hostname:adminport to deliver the @StopNode command. This was a problem, since hostname might not be interface listening for the admin port.  In fact it might not even be accessible.

This change uses the -H host to deliver the @StopNode command.  If it determines that -H host is the same as the target to stop, then it gives an error, "You must connect to a different host than the one you want to stop".  

My code seems to introduce 2 new problems:
1) It can't always determine what node it is connected to in order to check the above condition . For example, if your machine has multiple interfaces and you do not use any of the *interface settings (that is, all interfaces are listening), then it is possible that this code won't correctly figure out what node it is connected to. For example, if you are using volt3a,b,c (which have 2 networks) and you _voltdb start_ without setting any interfaces, certain voltadmin commands can get confused.

Mixing names and IPs can confuse it.  This command below can connect to 192.168.0.1 (volt3a), but
the check I added can't figure out that his is a host it knows, since OVERVIEW doesn't know this address
`
voltadmin stop  -H 192.168.0.1 volt3b 
`
And the error is confusing too (I left this from the original code)
`                                                                             
ERROR: Fatal error in "stop" command.
ERROR: The entire cluster is being stopped, use "shutdown" instead.
`
2) In a 1-node cluster, if you tell it to stop node, the reply will now be "ERROR: You must connect to a different host than the one you want to stop". It used to be the error shown above in point 1.

### DIFF
--- a/lib/python/voltcli/voltadmin.d/stop.py
+++ b/lib/python/voltcli/voltadmin.d/stop.py
@@ -46,17 +46,17 @@ class Hosts(object):
             value = int(value)
         self.hosts_by_id.setdefault(host_id, Host(host_id, self.abort_func))[prop_name] = value
 
-    def get_target_and_connection_host(self, host_name):
+    def get_target_and_connection_host(self, target_host_name, connection_host_name):
         """
-        Find an arbitrary host that isn't the one being stopped.
+        Find host being stopped and current host connected to. 
         Returns a tuple with connection and target host objects.
         """
         connection_host = None
         target_host = None
         for host in self.hosts_by_id.values():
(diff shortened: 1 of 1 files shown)


rmorgenstein:
Closing.  The restriction on -H being a different node from the target to stop is not a good one.

PR #3940, opened 2016-09-28
config.py: removed the joined-matview-* test suites from the "extended"

johnTcrawford:
tests.
sql_coverage_test.py: added expected failures for the joined-matview-*
test suites to the get_max_mismatches method.
Renamed joined-matview-extended-config.py to joined-matview-config.py
(removed "extended-"), since these tests are separate from the
"extended-" tests.


tests/scripts/sql_coverage_test.py
```diff
@@ -232,6 +232,13 @@ def get_max_mismatches(comparison_database, suite_name):
         # Known failures in the numeric-decimals "extended" test suite (see ENG-10546)
         elif config_name == 'numeric-decimals':
             max_mismatches = 300
+        # Known failures in the joined-matview-* test suites
+        elif config_name == 'joined-matview-default-full':
```
wolffcm: Can you annotate the allowed mismatches with the tickets that would need to be fixed for `max_mismatches` to go to zero?

wolffcm: With these changes will we be exercising materialized views with joins at all in the standard Jenkins jobs?  It seems like the answer is "no" but I think we probably should, even if it's just a smaller number of queries.

johnTcrawford: Yes, I've now updated the file with these - after filing a couple of relevant tickets (ENG-11196, ENG-11209).

johnTcrawford: These changes remove the joined-matview-* test suites from the *extended* sqlcoverage tests, but they will now be in their own separate Jenkins job: test-nextrelease-sqlcoverage-matview-postgres (which will run against master, like the "extended" tests). Also, the joined-matview-default-quick test suite remains part of branch-2-sqlcoverage-geo-postgis, which is run on every branch. Sorry I did not make this clear originally.

johnTcrawford:
Chris, I think I've now addressed your concerns, so I plan to merge this branch, once Jenkins is sufficiently blue (aside from known/intermittent failures).

PR #3954, opened 2016-10-01
Add support for dense rank.

BillWhite:
In this commit we add support for dense rank. We also refactor the
parsing of windowed aggregates to be less rank-centric and more general.

Also, with this commit, we repair some errors in the build system.
These caused unnecessary and misleading error messages, and are more
robust to GCC versions.


BillWhite:
Three junit test failures, down from 22.  None of them are even close to these changes, so I think they are part of the problems master has been having the past few days.  I'm going to merge this, and I'll monitor over the weekend.

PR #3957, opened 2016-10-03
Eng 10878

miketex:
We're trying to do SSL with non blocking io, with NIO.  This is accomplished using SSLEngine.

SSLEngine supports encryption and decryption of individual messages, it works independently of the transport layer.  SSLEngine is used to encrypt and decrypt messages that are passed between two peers, it doesn't have a sense of client or server.

Because SSLEngine is transport independent, it doesn't touch the SocketChannel being used to read and write messages.  So our code reads from the socket, and if ssl is configured for the port involved then our code uses SSLEngine to decrypt the message once it's been read.  Similarly on writing, if ssl is configured for a port then SSLEngine is used to encrypt the message before it's written to the socket; our code actually writes the message onto the socket.

When an SSL connection is established, the two parties involved go through a handshaking process.  So if sqlcmd connects to the server, both sqlcmd and the server need to agree that the ssl is enabled; they then go through the handshake.

A number of messages are passed back and forth during a handshake, but SSLEngine encrypts and decrypts only individual messages.  So our code needs to manage the handshake process, this code is in a class org.voltdb.client.SSLHandshaker.  After the connection is established, an SSLEngine that's being used to encrypt or decrypt messages needs to have participated in the handshake.  So an SSLEngine holds state as a result of handshaking, it needs that state to know how to encrypt and decrypt messages.

So the task is to agree, on the client and server, that ssl is enabled and then to handshake at the right point in time as the connection is established.  The SSLEngine that was used to handshake then needs to be plumbed into the code that will read and write over the encrypted port that was established.  So the client and server will then encrypt and decrypt all the messages that they pass back and forth over the port.

In our case, two things make this a little more complicated.  One is that we want to know the serialized size of a message when we read it and write it, but in the case of ssl we won't know the size of a message until it's been encrypted.  But we don't want to do the encryption up front, we want to defer it so that it's not done in the network thread.  

A second complication is that TLS has a max message size of 16k.  This means that SSLEngine will encrypt/decrypt a buffer of this size or less.  When we have a message bigger than this, we need to break it into chunks, encrypt and send each chunk, and then assemble the chunks on the receiving end.  So a given message can consist of some number of chunks.

This was implemented without changing our wire protocol.  Each chunk consists of a four bytes size that's sent in clear.  This size is the size of the message chunk that follows, and the chunk that follows is encrypted.  For each chunk, we read it's size, in clear, read that many bytes into a buffer, and decrypt the buffer.

When the first chunk is read, and the encrypted part is decrypted, the first four decrypted bytes rerpresent the full size of the decrypted message.  We'll continue to decrypt chunks until we reach that size.  Necessarily the last chunk needs to decrypt exactly to that size or there's a bug.  That is, when we decrypt the last chunk and add it the bytes we previously decrypted, the full decrypted message needs to be the expected size.

The same scheme is used even when a message consists of only one chunk.  That is, the first four bytes of the message are in clear and represent the size of the encrypted chunk to follow.  That chunk is then decrypted.  The first four bytes of the decrypted chunk match the size of the rest of the decrypted chunk, which is the message in clear.

In terms of code, the handshaking for the server is done in ClientInterface:AuthRunnable.run().  For the client, handshaking is done in ConnectionUtil.getAuthenticatedConnection().

Encryption of messages, meaning the write side of things, is done in NIOWriteStream.  The instance of SSLEngine that was used to handshake is plumbed into NIOWriteStream.  If SSL is configured, encryption code is in NIOWriteStream's enqueue methods.  If serialization is deferred, then encryption of the message is also deferred.  

Note that encryption of a single message, or buffer, can result in a series of encrypted buffers that will be written onto the socket.  This concept is captured in a class SSLDeferredSerializationIterator.  Encrypting one message buffer yields an iterator, iterating then yields each encrypted buffer; though there might be just one.

Decrypting of messages, the read side of things, is done in VoltPort.  VoltPort uses a class, SSLMessageDecrypter, that decrypts a message.  In particular this class is used to read the chunks that make up a message and assemble them into the full message.

One final node, the api for DeferredSerialization is changed.  In particular, the serialize method is changed from:

void serialize(ByteBuffer buf) throws IOException;

to:

ByteBuffer serialize(ByteBuffer buf) throws IOException;

The reason for this is that, for messages that will be encrypted, we can't know the serialized size of a message until after it's been encrypted.  In this scheme, when serialize is called, we try to encrypt the message into the given buffer.  If the message doesn't fit, we allocate a new buffer, the serialize() method returns the newly allocated buffer if one was needed, null if not.  The existing code in NIOWriteStreamBase.swapAndSerializeQueuedWrites() has similar logic; if a message will fit in the given buffer it uses that buffer, if not it allocates a new buffer that will be big enough.


src/frontend/org/voltcore/network/NIOWriteStream.java
```diff
@@ -342,21 +355,58 @@ int drainTo (final GatheringByteChannel channel) throws IOException {
...
+                        break;
+                    }
+
+                    ByteBuffer plainTextMessage = ByteBuffer.allocate(messageLen);
+                    process.get(plainTextMessage.array());
+
+                    // encrypt
+                    m_encBuffer.clear();
+                    wrapMessage(plainTextMessage);
+                    ByteBuffer sendMessage = ByteBuffer.allocate(m_encBuffer.remaining() + 4);
```
wweiss-voltdb: I don't believe you need to do this copy if you set the position of m_encBuffer to 4 in wrapMessage.

src/frontend/org/voltcore/network/NIOWriteStream.java
```diff
@@ -311,6 +322,8 @@ protected void updateQueued(int queued, boolean noBackpressureSignal) {
         }
     }
 
+    private ByteBuffer remainder = null;
```
wweiss-voltdb: move to top and prefix with m_

src/frontend/org/voltcore/network/NIOWriteStream.java
```diff
@@ -378,4 +428,25 @@ int drainTo (final GatheringByteChannel channel) throws IOException {
         }
         return bytesWritten;
     }
+
+    private void wrapMessage(ByteBuffer unwrapped) throws IOException {
+        while (true) {
+            SSLEngineResult result = m_sslEngine.wrap(unwrapped, m_encBuffer);
```
wweiss-voltdb: If you use the gather version of wrap (that takes multiple bytebuffers, you won't need to combine (create and copy) when dealing with remainders

examples/voter/client/voter/AsyncBenchmark.java
```diff
@@ -455,6 +513,20 @@ public static void main(String[] args) throws Exception {
         VoterConfig config = new VoterConfig();
         config.parse(AsyncBenchmark.class.getName(), args);
 
+        SSLContext sslContext;
+        try {
+            KeyStore ks = KeyStore.getInstance("JKS");
+            ks.load(new FileInputStream("/Users/mteixeira/keystore.jks"), "myk5yst15r5".toCharArray());
```
akhanzode: remove hardcoded paths use javax system properties.

src/frontend/org/voltcore/network/NIOWriteStream.java
```diff
@@ -76,20 +81,26 @@
      */
     private long m_lastPendingWriteTime = -1;
 
+    private final boolean m_isSSLConfigured;
```
akhanzode: Do you need boolean as well as sslengine only sslengine would work and give method that returns bool based on sslengine.

src/frontend/org/voltcore/network/NIOWriteStream.java
```diff
@@ -188,29 +199,80 @@ public void enqueue(final DeferredSerialization ds) {
...
-            public void run() {
-                synchronized (NIOWriteStream.this) {
-                    updateLastPendingWriteTimeAndQueueBackpressure();
-                    m_queuedWrites.offer(ds);
-                    m_port.setInterests( SelectionKey.OP_WRITE, 0);
+        if (m_isSSLConfigured) {
+            m_port.queueTask(new Runnable() {
+                @Override
+                public void run() {
+                    Iterator<DeferredSerialization> dsIter = new SSLDeferredSerializationIterator(m_sslEngine, ds);
```
akhanzode: You can move this inside that task as opposed to big if

src/frontend/org/voltcore/network/NIOWriteStream.java
```diff
@@ -243,30 +305,44 @@ public void enqueue(final ByteBuffer b[]) {
...
             updateLastPendingWriteTimeAndQueueBackpressure();
 
-            m_queuedWrites.offer(new DeferredSerialization() {
-                @Override
-                public void serialize(ByteBuffer outbuf) {
-                    for (ByteBuffer buf : b) {
-                        outbuf.put(buf);
+            for (final ByteBuffer buf : b) {
+                if (m_isSSLConfigured) {
+                    Iterator<DeferredSerialization> dsIter = new SSLDeferredSerializationIterator(m_sslEngine, new Serializer() {
```
akhanzode: Can this be merged ? I case of no ssl the itarator will point to only one.

src/frontend/org/voltdb/Inits.java
```diff
@@ -549,6 +613,54 @@ public void run() {
...
+        /**
+         * Creates the key managers required to initiate the {@link SSLContext}, using a JKS keystore as an input.
+         *
+         * @param filepath - the path to the JKS keystore.
+         * @param keystorePassword - the keystore's password.
+         * @param keyPassword - the key's passsword.
+         * @return {@link KeyManager} array that will be used to initiate the {@link SSLContext}.
+         * @throws Exception
+         */
+        private KeyManager[] createKeyManagers(String filepath, String keystorePassword, String keyPassword) throws Exception {
```
akhanzode: This could be in some utils class so client and server can share.

src/frontend/org/voltdb/client/ConnectionUtil.java
```diff
@@ -306,6 +327,7 @@ public DelegatePrincipal apply(Principal input) {
             byte buildStringBytes[] = new byte[buildStringLength];
             loginResponse.get(buildStringBytes);
             returnArray[2] = new String(buildStringBytes, Constants.UTF8ENCODING);
+            returnArray[3] = sslEngine;
```
akhanzode: why add sslEngine here doesnt look right?

src/frontend/org/voltdb/client/SSLHandshaker.java
```diff
@@ -0,0 +1,177 @@
...
+        m_netEncData.flip();
+
+        SSLEngineResult.HandshakeStatus hs;
+        SSLEngineResult res;
+        try {
+            res = engine.unwrap(m_netEncData, m_netClearData);
+            m_netEncData.compact();
+            hs = res.getHandshakeStatus();
+        } catch (SSLException e) {
+            System.err.println("unwrap: see ssl exception " + e.getMessage());
```
akhanzode: no system out please. log it.

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -315,14 +315,14 @@
           <xs:attribute name="enabled" type="xs:boolean" default="true"/>
         </xs:complexType>
       </xs:element>
-      <xs:element name="https" minOccurs="0" maxOccurs="1" type="httpsType"/>
+      <xs:element name="ssl" minOccurs="0" maxOccurs="1" type="sslType"/>
```
akhanzode: I thought Andrew wants this outside of http.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1489,4 +1544,51 @@ private static void openURL()
...
+    /**
+     * Creates the key managers required to initiate the {@link SSLContext}, using a JKS keystore as an input.
+     *
+     * @param filepath - the path to the JKS keystore.
+     * @param keystorePassword - the keystore's password.
+     * @param keyPassword - the key's passsword.
+     * @return {@link KeyManager} array that will be used to initiate the {@link SSLContext}.
+     * @throws Exception
+     */
+    protected static KeyManager[] createKeyManagers(String filepath, String keystorePassword, String keyPassword) throws Exception {
```
akhanzode: should be from shared utility.

tests/frontend/org/voltcore/network/TestNIOWriteStream.java
```diff
@@ -75,7 +75,7 @@ public String toString() {
         }
 
         public MockPort() throws UnknownHostException {
-            super(null, null, new InetSocketAddress(InetAddress.getByName("127.0.0.1"), 21212), pool);
+            super(null, null, new InetSocketAddress(InetAddress.getByName("127.0.0.1"), 21212), pool, null);
```
akhanzode: These tests should have SSL. They dont depend on server.

tests/frontend/org/voltcore/network/TestVoltNetwork.java
```diff
@@ -43,7 +43,7 @@
 
     private static class MockVoltPort extends VoltPort {
         MockVoltPort(VoltNetwork vn, InputHandler handler) throws UnknownHostException {
-            super (vn, handler, new InetSocketAddress(InetAddress.getByName("127.0.0.1"), 21212), vn.m_pool);
+            super (vn, handler, new InetSocketAddress(InetAddress.getByName("127.0.0.1"), 21212), vn.m_pool, null);
```
akhanzode: Should have SSL in tests.

tests/frontend/org/voltdb/TestClientInterface.java
```diff
@@ -851,8 +851,9 @@ public void testSubscribe() throws Exception {
     private DeferredSerialization dsOf(final ByteBuffer buf) {
         return new DeferredSerialization() {
             @Override
-            public void serialize(final ByteBuffer outbuf) throws IOException {
+            public ByteBuffer serialize(final ByteBuffer outbuf) throws IOException {
                 outbuf.put(buf);
+                return null;
```
akhanzode: I thought for non ssl we return the same bytebuffer.

tests/frontend/org/voltdb/client/TestDistributer.java
```diff
@@ -241,7 +241,7 @@ public void run() {
                         client.configureBlocking(false);
                         channels.add(client);
                         if (handleConnection) {
-                            network.registerChannel( client, handler);
+                            network.registerChannel( client, handler, null);
```
akhanzode: Should include ssl in tests.

examples/voter/client/voter/AsyncBenchmark.java
```diff
@@ -444,6 +455,53 @@ public void runBenchmark() throws Exception {
...
+
+    /**
+     * Creates the trust managers required to initiate the {@link SSLContext}, using a JKS keystore as an input.
+     *
+     * @param filepath - the path to the JKS keystore.
+     * @param keystorePassword - the keystore's password.
+     * @return {@link TrustManager} array, that will be used to initiate the {@link SSLContext}.
+     * @throws Exception
+     */
+    protected static TrustManager[] createTrustManagers(String filepath, String keystorePassword) throws Exception {
```
miketex: Yes, I haven't quite finished with this part of the code yet.

examples/voter/client/voter/AsyncBenchmark.java
miketex: Yes, I probably should have just reverted AsyncBenchmark for now as it's not yet finished.

src/frontend/org/voltcore/network/NIOWriteStream.java
miketex: Yes, ok.

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
miketex: He does, not all all of Andrew's comments are incorporated yet.

src/frontend/org/voltdb/Inits.java
```diff
@@ -516,6 +537,49 @@ private void setupHttpServer(String httpInterface, String publicInterface,
...
         }
 
+        private SslContextFactory getSSLContextFactory(SslType sslType) {
+            SslContextFactory sslContextFactory = new SslContextFactory();
+            String value = getKeyTrustStoreAttribute("javax.net.ssl.keyStore", sslType.getKeystore(), "path", true);
+            sslContextFactory.setKeyStorePath(value);
+            String keyStorePassword = getKeyTrustStoreAttribute("javax.net.ssl.keyStorePassword", sslType.getKeystore(), "password", true);
+            sslContextFactory.setKeyStorePassword(keyStorePassword);
+            value = getKeyTrustStoreAttribute("javax.net.ssl.trustStore", sslType.getTruststore(), "path", false);
+            if (value!=null) {
```
vtkstef: spacing `if (value != null)`

src/frontend/org/voltdb/Inits.java
```diff
@@ -516,6 +537,49 @@ private void setupHttpServer(String httpInterface, String publicInterface,
...
+                sslContextFactory.setTrustStorePath(value);
+            }
+            value = getKeyTrustStoreAttribute("javax.net.ssl.trustStorePassword", sslType.getTruststore(), "password", false);
+            if (value!=null) {
```
vtkstef: spacing

src/frontend/org/voltdb/Inits.java
```diff
@@ -516,6 +537,49 @@ private void setupHttpServer(String httpInterface, String publicInterface,
...
+                    "SSL_RSA_EXPORT_WITH_RC4_40_MD5",
+                    "SSL_RSA_EXPORT_WITH_DES40_CBC_SHA",
+                    "SSL_DHE_RSA_EXPORT_WITH_DES40_CBC_SHA",
+                    "SSL_DHE_DSS_EXPORT_WITH_DES40_CBC_SHA");
+            sslContextFactory.setKeyManagerPassword(keyStorePassword);
+            return sslContextFactory;
+        }
+
+        private String getKeyTrustStoreAttribute(String sysPropName, KeyOrTrustStoreType store, String valueType, boolean throwForNull) {
+            String sysProp = System.getProperty(sysPropName);
```
vtkstef: you can use the default variant `String sysProp = System.getProperty(sysPropName,'')`

src/frontend/org/voltdb/Inits.java
```diff
@@ -516,6 +537,49 @@ private void setupHttpServer(String httpInterface, String publicInterface,
...
+            if (StringUtils.isNotBlank(sysProp)) {
```
vtkstef: .. and avoid using a third party utility method by simply `if (!sysProp.trim().isEmpty())`

src/frontend/org/voltdb/Inits.java
```diff
@@ -516,6 +537,49 @@ private void setupHttpServer(String httpInterface, String publicInterface,
...
+                return sysProp.trim();
+            } else {
+                String value = null;
+                if (store!=null) {
+                    value = "path".equals(valueType) ? store.getPath() : store.getPassword();
```
vtkstef: btw password can contain leading and trailing blanks. you need to handle them for passwords only

src/frontend/org/voltdb/Inits.java
```diff
@@ -524,7 +588,7 @@ public void run() {
             m_rvdb.m_jsonEnabled = false;
             boolean httpsEnabled = false;
             if ((m_deployment.getHttpd() != null) && (m_deployment.getHttpd().isEnabled())) {
-                if (m_deployment.getHttpd().getHttps()!=null && m_deployment.getHttpd().getHttps().isEnabled()) {
+                if (m_deployment.getHttpd().getSsl()!=null && m_deployment.getHttpd().getSsl().isEnabled()) {
```
vtkstef: spacing

src/frontend/org/voltdb/Inits.java
```diff
@@ -549,6 +613,54 @@ public void run() {
...
+         *
+         * @param filepath - the path to the JKS keystore.
+         * @param keystorePassword - the keystore's password.
+         * @param keyPassword - the key's passsword.
+         * @return {@link KeyManager} array that will be used to initiate the {@link SSLContext}.
+         * @throws Exception
+         */
+        private KeyManager[] createKeyManagers(String filepath, String keystorePassword, String keyPassword) throws Exception {
+            KeyStore keyStore = KeyStore.getInstance("JKS");
+            InputStream keyStoreIS = new FileInputStream(filepath);
```
vtkstef: use try with resource
```java
try (InputStream keyStoreIS = new FileInputStream(filepath)) {
   keyStore.load(keyStoreIS, keystorePassword.toCharArray());
} // don't need the finally
```

src/frontend/org/voltdb/Inits.java
vtkstef: big code smell when your procedure throws a generic Exception. If you want a generic exception create a descriptive runtime exception like TLSSetupException and use it to wrap check exceptions btw log them too.

src/frontend/org/voltdb/Inits.java
```diff
@@ -549,6 +613,54 @@ public void run() {
...
+
+        /**
+         * Creates the trust managers required to initiate the {@link SSLContext}, using a JKS keystore as an input.
+         *
+         * @param filepath - the path to the JKS keystore.
+         * @param keystorePassword - the keystore's password.
+         * @return {@link TrustManager} array, that will be used to initiate the {@link SSLContext}.
+         * @throws Exception
+         */
+        private TrustManager[] createTrustManagers(String filepath, String keystorePassword) throws Exception {
```
vtkstef: big code smell when your procedure throws a generic Exception. If you want a generic exception create a descriptive runtime exception like TLSSetupException and use it to wrap check exceptions btw log them too.

src/frontend/org/voltdb/client/SSLHandshaker.java
```diff
@@ -0,0 +1,177 @@
...
+        switch (res.getStatus()) {
+            case OK:
+                m_localEncData.flip();
+                while (m_localEncData.hasRemaining()) {
+                    m_sc.write(m_localEncData);
+                }
+                m_localEncData.compact();
+                break;
+            case BUFFER_OVERFLOW:
+                m_localEncData = ByteBuffer.allocate(m_localEncData.capacity() * 2);
```
vtkstef: bit shifting << 1 is faster than * 2

src/frontend/org/voltdb/client/SSLHandshaker.java
```diff
@@ -0,0 +1,177 @@
...
+    private ByteBuffer m_localClearData;
+    private ByteBuffer m_localEncData;
+    private ByteBuffer m_netClearData;
+    private ByteBuffer m_netEncData;
+
+    public SSLHandshaker(SocketChannel sc, SSLEngine engine) {
+        m_sc = sc;
+        m_sslEngine = engine;
+        SSLSession dummySession = engine.getSession();
+        m_localClearData = ByteBuffer.allocate(dummySession.getApplicationBufferSize() + 50);
```
vtkstef: can wrap and unwrap be called simultaneously by two threads, if not here you really need one pair of byte buffers

src/frontend/org/voltdb/client/SSLHandshaker.java
vtkstef: log not system.err

src/frontend/org/voltdb/client/SSLHandshaker.java
```diff
@@ -0,0 +1,177 @@
...
+            System.err.println("unwrap: see ssl exception " + e.getMessage());
+            engine.closeOutbound();
+            return engine.getHandshakeStatus();
+        }
+
+        switch (res.getStatus()) {
+            case OK:
+                break;
+            case BUFFER_OVERFLOW:
+                m_netClearData = ByteBuffer.allocate(m_netClearData.capacity() * 2);;
```
vtkstef: bit shifting << 1 is faster than * 2

src/frontend/org/voltdb/client/SSLMessageEncrypter.java
```diff
@@ -0,0 +1,65 @@
...
+    private static ByteBuffer encryptWithAllocated(SSLEngine sslEngine, ByteBuffer src) throws IOException {
+        ByteBuffer allocated = ByteBuffer.allocate((int) (src.capacity() * 1.2));
+        while (true) {
+            SSLEngineResult result = sslEngine.wrap(src, allocated);
+            switch (result.getStatus()) {
+                case OK:
+                    allocated.flip();
+                    return allocated;
+                case BUFFER_OVERFLOW:
+                    allocated = ByteBuffer.allocate(allocated.capacity() * 2);
```
vtkstef: bit shifting << 1 is faster than * 2

src/frontend/org/voltdb/utils/SQLCommand.java
vtkstef: procedure throws a generic Exception. If you want a generic exception create a descriptive runtime exception like TLSSetupException and use it to wrap check exceptions btw log them too

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1489,4 +1544,51 @@ private static void openURL()
...
+
+    /**
+     * Creates the trust managers required to initiate the {@link SSLContext}, using a JKS keystore as an input.
+     *
+     * @param filepath - the path to the JKS keystore.
+     * @param keystorePassword - the keystore's password.
+     * @return {@link TrustManager} array, that will be used to initiate the {@link SSLContext}.
+     * @throws Exception
+     */
+    protected static TrustManager[] createTrustManagers(String filepath, String keystorePassword) throws Exception {
```
vtkstef: procedure throws a generic Exception. If you want a generic exception create a descriptive runtime exception like TLSSetupException and use it to wrap check exceptions btw log them too

src/frontend/org/voltcore/utils/ssl/SSLBufferDecrypter.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private final SSLEngine m_sslEngine;
+    private final ByteBuffer m_srcBuffer;
+    private final ByteBuffer m_dstBuffer;
+    private ByteBuffer m_partialMessage;
+    private ByteBuffer m_partialLength;
+
+    public SSLBufferDecrypter(SSLEngine sslEngine) {
+        this.m_sslEngine = sslEngine;
+        this.m_srcBuffer = ByteBuffer.allocateDirect(Constants.SSL_CHUNK_SIZE + 128);
```
vtkstef: where are the corresponding cleaners for these direct byte buffer allocations?

src/frontend/org/voltcore/utils/ssl/SSLBufferEncrypter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+public class SSLBufferEncrypter {
+
+    private final SSLEngine m_sslEngine;
+    private ByteBuffer m_encBuffer;
+
+    public SSLBufferEncrypter(SSLEngine sslEngine) {
+        this.m_sslEngine = sslEngine;
+        int packetBufferSize = m_sslEngine.getSession().getPacketBufferSize();
```
vtkstef: where are the corresponding cleaners for these direct byte buffer allocations?

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -249,14 +254,15 @@ public AdmissionControlGroup initialValue() {
...
+        private final SSLContext m_sslContext;
 
         /**
          * Used a cached thread pool to accept new connections.
          */
         private final ExecutorService m_executor = CoreUtils.getBoundedThreadPoolExecutor(128, 10L, TimeUnit.SECONDS,
                         CoreUtils.getThreadFactory("Client authentication threads", "Client authenticator"));
 
-        ClientAcceptor(InetAddress intf, int port, VoltNetworkPool network, boolean isAdmin)
+        ClientAcceptor(InetAddress intf, int port, VoltNetworkPool network, boolean isAdmin, SSLContext sslContext)
```
vtkstef: see my comment about VoltPort factory. ClientInterface should be devoid of any SSL knowledge

src/frontend/org/voltdb/client/ClientConfig.java
```diff
@@ -420,4 +455,59 @@ public void enableKerberosAuthentication(final String loginContextEntryKey) {
...
+        InputStreamReader configIsr;
+        try {
+            FileInputStream configFis = new FileInputStream(configFile);
+            configIsr = new InputStreamReader(configFis);
+        } catch (FileNotFoundException e) {
+            throw new IllegalArgumentException("Can't read ssl config file " + m_sslPropsFile);
+        }
+
+        String keyStorePath = null;
+        String keyStorePassword = null;
```
vtkstef: don't need keyStore and keyStorePassword. remove them

src/frontend/org/voltcore/network/InputHandler.java
```diff
@@ -22,6 +22,14 @@
 
 public interface InputHandler {
 
+    /** The distinct exception class allows better logging of these unexpected errors. */
+    class BadMessageLength extends IOException {
+        private static final long serialVersionUID = 8547352379044459911L;
+        public BadMessageLength(String string) {
```
vtkstef: include the other constructors too, specifically the one that takes in a throwable, and a message and a throwable

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,94 @@
...
+    private final SSLBufferDecrypter m_sslBufferDecrypter;
+    private final ByteBuffer m_sslSrcBuffer;
+    private final ByteBuffer m_messageHeader;
+    private final byte[] m_twoBytes = new byte[2];
+
+    public SSLVoltPort(VoltNetwork network, InputHandler handler, InetSocketAddress remoteAddress, NetworkDBBPool pool, SSLEngine sslEngine) {
+        super(network, handler, remoteAddress, pool);
+        this.m_sslEngine = sslEngine;
+        m_sslBufferDecrypter = new SSLBufferDecrypter(sslEngine);
+        m_sslSrcBuffer = ByteBuffer.allocateDirect(Constants.SSL_CHUNK_SIZE + 128);
```
vtkstef: use BBContainer to get directly allocated byte buffers, make sure you call BBContainer discard() when you close this port

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,94 @@
...
+            while ((message = m_sslBufferDecrypter.message()) != null) {
+                m_handler.handleMessage(message, this);
+                m_messagesRead++;
+            }
+        }
+    }
+
+    protected int getMessageLengthFromHeader(ByteBuffer header) {
+        header.position(3);
+        header.get(m_twoBytes);
```
vtkstef: are these unsigned? why can't you use getShort() instead?

src/frontend/org/voltcore/network/SSLNIOWriteStream.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    synchronized public boolean isEmpty()
+    {
+        return super.isEmpty() && m_encryptedWriteBuffer == null && m_queuedWrites.isEmpty();
+    }
+
+    protected ByteBuffer getBufferToWrite() throws IOException {
+        if (m_encryptedWriteBuffer != null) {
```
vtkstef: Question: is there a chance that two threads could content for this call?

src/frontend/org/voltcore/network/SSLNIOWriteStream.java
```diff
@@ -0,0 +1,74 @@
...
+            return m_encryptedWriteBuffer;
+        }
+        ByteBuffer writeBuffer;
+        if ((writeBuffer = super.getBufferToWrite()) == null) {
+            return null;
+        }
+        ByteBuffer buffToEncrypt;
+        if (writeBuffer.remaining() <= Constants.SSL_CHUNK_SIZE) {
```
vtkstef: chunk size is a cipher attribute. we know the MAX as specified by TLS's RFC, but different ciphers may require different segment/chunk/fragment sizes

src/frontend/org/voltcore/network/SSLNIOWriteStream.java
```diff
@@ -0,0 +1,74 @@
...
+            writeBuffer.limit(writeBuffer.position() + Constants.SSL_CHUNK_SIZE);
+            buffToEncrypt = writeBuffer.slice();
+            writeBuffer.position(writeBuffer.limit());
+            writeBuffer.limit(oldLimit);
+        }
+        m_encryptedWriteBuffer = m_sslBufferEncrypter.encryptBuffer(buffToEncrypt);
+        return m_encryptedWriteBuffer;
+    }
+
+    protected int discardBuffer() {
```
vtkstef: rename the method as the code discards tentatively on conditions

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,190 @@
...
+                }
+
+                int read;
+                while (true) {
+                    if (m_nextFrameLength == 0) {
+                        if (readStream().dataAvailable() >= 5) {
+                            read = readStream().getBytes(m_frameHeader);
+                            m_frameHeader.flip();
+                            m_frameHeader.position(3);
+                            m_frameHeader.get(m_twoBytes);
```
vtkstef: any reason why you can't use getShort?

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,190 @@
...
+                        m_callables.add(new DecryptionCallable(frameCont));
+                        m_nextFrameLength = 0;
+                    } else {
+                        break;
+                    }
+                }
+
+                Callable c = m_callables.poll();
+                if (c != null) {
+                    Future<List<ByteBuffer>> f = SSLEncryptionService.instance().submit(c);
```
vtkstef: Make your executor service a listenable executor service as many other Volt executor services do. Return ListenableFutures

vtkstef: I am inferring you will have two executor services one for decrypting and one for decrypting. During our design session we specified that you would use only one fronted by two gateway objects (one for decrypt, and on for encrypt) that each would manage its submission queue to one executor service

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,190 @@
...
+                    List<ByteBuffer> messages;
+                    while (f != null) {
+                        try {
+                            messages = f.get();
```
vtkstef: why are you waiting right away for the future? Here you are basically doing one at a time sequentially.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private class DecryptionCallable implements Callable<List<ByteBuffer>> {
+        private final DBBPool.BBContainer m_srcCont;
+
+        public DecryptionCallable(DBBPool.BBContainer srcCont) {
+            this.m_srcCont = srcCont;
+        }
+
+        @Override
+        public List<ByteBuffer> call() throws IOException {
```
vtkstef: call does not throw IOException, look at other parts fo Volt for reference implmentation

src/frontend/org/voltcore/network/SSLVoltPort.java
vtkstef: you are not handling execution exceptions correctly, see other parts of Volt for reference implementations

src/frontend/org/voltcore/network/VoltNetwork.java
```diff
@@ -173,7 +176,8 @@ public Connection call() throws Exception {
                                 VoltNetwork.this,
                                 handler,
```
vtkstef: comment is still pertinent

src/frontend/org/voltcore/utils/ssl/MessagingChannel.java
```diff
@@ -0,0 +1,133 @@
...
+    private final SSLMessageParser m_sslMessageParser;
+
+    public MessagingChannel(SocketChannel m_socketChannel, SSLEngine sslEngine) {
+        this.m_socketChannel = m_socketChannel;
+        this.m_sslEngine = sslEngine;
+        if (sslEngine != null) {
+            int packetBufferSize = m_sslEngine.getSession().getPacketBufferSize();
+            // wrapMessage will overflow until the encryption buffer is this size.
+            this.m_encBuffer = ByteBuffer.allocate(packetBufferSize);
+            this.m_dstBuffer = ByteBuffer.allocateDirect(1024);
```
vtkstef: where is the cleaner for this direct allocation. Use BBContainer and add a close method that discards it

src/frontend/org/voltcore/utils/ssl/SSLEncryptionService.java
```diff
@@ -0,0 +1,65 @@
...
+    public SSLEncryptionService(int nThreads) {
+        this.m_executorService = Executors.newFixedThreadPool(nThreads);
+    }
+
+    public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
+        return m_executorService.awaitTermination(timeout, unit);
+    }
+
+    public void shutdown() {
+        m_executorService.shutdown();
```
vtkstef: see in Volt how to properly shutdown an executor service

src/frontend/org/voltcore/utils/ssl/SSLEncryptionService.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public List<Runnable> shutdownNow() {
```
vtkstef: useless

src/frontend/org/voltcore/utils/ssl/SSLEncryptionService.java
```diff
@@ -0,0 +1,65 @@
...
+        m_self = new SSLEncryptionService(nThreads);
+        return m_self;
+    }
+
+    public static SSLEncryptionService instance() {
+        return m_self;
+    }
+
+    public SSLEncryptionService(int nThreads) {
+        this.m_executorService = Executors.newFixedThreadPool(nThreads);
```
vtkstef: you should adorn it with guava MoreExecutors see how it is done in other parts of Volt

src/frontend/org/voltcore/utils/ssl/SSLEncryptionService.java
```diff
@@ -0,0 +1,65 @@
...
+    public void shutdown() {
+        m_executorService.shutdown();
+    }
+
+    public List<Runnable> shutdownNow() {
+        return m_executorService.shutdownNow();
+    }
+
+    public <T> Future<T> submit(Callable<T> task) {
+        return m_executorService.submit(task);
```
vtkstef: should return a ListenableFuture

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -404,6 +415,34 @@ public void run() {
...
+                        boolean handshakeStatus;
+                        try {
+                            handshakeStatus = handshaker.handshake();
+                        } catch (IOException e) {
+                            socket.close();
+                            networkLog.warn("Rejected accepting new connection, SSL handshake failed: " + e.getMessage());
+                            continue;
+                        }
+                        if (!handshakeStatus) {
+                            socket.close();
```
vtkstef: socket close could close too

src/frontend/org/voltcore/network/SSLNIOWriteStream.java
miketex: I don't believe so...  I believe this code will be obsoleted when the rest of this design is pushed up.

src/frontend/org/voltcore/network/SSLNIOWriteStream.java
miketex: Yes, that's true.  I've yet to make that change, but will.

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: No, there's only one executor service - this one is used for both encryption and decryption, I didn't think of a better name.

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: Sequentially ... I thought that was design?  SSLEngine can do wrap/unwrap concurrently, but not two unwraps.  Tasks could be submitted concurrently if the engine is locked for the duration of unwrap.  The messages need to be processed in the order submitted, but that's not really relevant to the submission policy.  Do you want to submit concurrently and lock the engine?

src/frontend/org/voltcore/utils/ssl/MessagingChannel.java
miketex: Yes, I think MessagingChannel should be removed, but I haven't done that yet.

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -1075,29 +1055,36 @@ public static ClientInterface create(
             InetAddress clientIntf,
             int clientPort,
             InetAddress adminIntf,
-            int adminPort) throws Exception {
+            int adminPort,
+            SSLContext sslContext) throws Exception {
```
vtkstef: isn't ssl context passed at client interface instantiation, could you use refer it a member variable instead of weaving it here as a parameter

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2835,6 +2836,11 @@ public boolean shutdown(Thread mainSiteThread) throws InterruptedException {
                 m_latencyStats = null;
                 m_latencyHistogramStats = null;
 
+                if (m_config.m_sslEncryptionService != null) {
+                    m_config.m_sslEncryptionService.shutdown();
+                    m_config.m_sslEncryptionService.awaitTermination(1, TimeUnit.SECONDS);
```
vtkstef: move it to encryption service and refer to Volt source to see how long you really need to wait

src/frontend/org/voltdb/client/ClientConfig.java
```diff
@@ -430,4 +478,66 @@ public void enableKerberosAuthentication(final String loginContextEntryKey) {
...
+            m_sslContext = null;
+            return;
+        }
+        File configFile = new File(m_sslPropsFile);
+        InputStreamReader configIsr;
+        try {
+            FileInputStream configFis = new FileInputStream(configFile);
+            configIsr = new InputStreamReader(configFis);
+        } catch (FileNotFoundException e) {
+            throw new IllegalArgumentException("Can't read ssl config file " + m_sslPropsFile);
```
vtkstef: chain the filenotfoundexception into the illegalargumentexception

src/frontend/org/voltdb/client/ConnectionUtil.java
```diff
@@ -216,52 +236,26 @@ public DelegatePrincipal apply(Principal input) {
...
-                    if (!successfulWrite) {
-                        throw new IOException("Unable to write authentication info to server");
-                    }
-                    throw new IOException("Authentication rejected");
-                }
+            ByteBuffer loginResponse;
+            try {
+                loginResponse = messagingChannel.readMessage();
+            } catch (IOException e) {
+                throw new IOException("Authentication rejected");
```
vtkstef: chain the caught exception with the one you are throwing

src/frontend/org/voltdb/common/Constants.java
```diff
@@ -78,5 +78,6 @@
     public static final int HTTP_PORT_AUTO = 0;
 
     public static final String DBROOT = "voltdbroot";
+    public static final int SSL_CHUNK_SIZE = 16 * 1024;
```
vtkstef: this is cipher dependent see my previous comment regarding this

src/frontend/org/voltcore/network/SSLVoltPort.java
vtkstef: that is the reason we were discussing the gateway objects one for encrypt and one for decrypt. each will have its own queue.

class EncryptGateway {   
   Queue<Pair<ByteBuffer,SettableFuture>> m_q;
   synchronized ListenableFuture<ByteBuffer> enque(ByteBuffer fragment) {
      boolean empty = m_queue.isEmpty();
      SettableFuture<ByteBuffer> fut = SettableFuture.create();
      Runnable task = new Runnable() {
           Pair<ByteBuffer,SettableFuture> p = m_q.poll();
           ByteBuffer fragment = p.getFirst();
           SettableFuture fut = p.getSecond();
           // decrypt buffer. on exception set it in the settable future
           fut.set(decryptedBuffer);
           synchronized(EncryptGateway.this) {
               if (!m_q.isEmpty()) {
                   executorService.submit(this);
               }
      };
      m_q.offer(Pair.create(fragment, fut));
      if (empty) {
          excutorService.submit(task);
      }
}

src/frontend/org/voltcore/utils/ssl/SSLEncryptionService.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public void shutdown() {
+        m_executorService.shutdown();
+    }
+
+    public List<Runnable> shutdownNow() {
+        return m_executorService.shutdownNow();
+    }
+
```
miketex: Not understanding this comment in particular.  What do you want here?

src/frontend/org/voltdb/Inits.java
```diff
@@ -472,6 +475,107 @@ public void run() {
...
+        private static final String DEFAULT_KEYSTORE_PASSWD = "password";
+
+        @Override
+        public void run() {
+            SslType sslType = m_deployment.getSsl();
+            if ((sslType != null && sslType.isEnabled()) || (m_config.m_sslEnable)) {
+                try {
+                    m_config.m_sslContextFactory = getSSLContextFactory(sslType);
+                    m_config.m_sslContextFactory.start();
+                    hostLog.info("SSL Enabled for HTTP. Please point browser to HTTPS URL.");
```
hhakimi53: Should this message be more clearer?

lib/python/Query.py
```diff
@@ -500,6 +505,7 @@ def help(program_name):
     print program_name, "hostname port [dump=filename] [command]"
 
 if __name__ == "__main__":
+    # TODO Add SSL arguments to command line & its help
```
hhakimi53: Is it still a todo item?

lib/python/voltdbclient.py
```diff
@@ -129,25 +131,42 @@ class FastSerializer:
     # if these assumptions are not true. it is further assumed
     # that host order is little endian. See isNaN().
 
-    def __init__(self, host = None, port = 21212, username = "",
+    # default ssl configuration
+    DEFAULT_SSL_CONFIG = {
+        'keyfile': None,
+        'certfile': None,
+        'cert_reqs': ssl.CERT_NONE,
+        'ssl_version': ssl.PROTOCOL_TLSv1,
```
hhakimi53: Is this the protocol that will be used? How about SSLv23?

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+
+    private void processReads() throws IOException {
+        if (readStream().dataAvailable() == 0) {
+            final int maxRead = m_handler.getMaxRead();
+            int nRead = fillReadStream(maxRead);
+            if (nRead > 0) {
+                m_decryptionGateway.submitDecryptionTasks();
+            }
+        } else {
+            m_decryptionGateway.submitDecryptionTasks();
```
nshi: I'm assuming that this branch is supposed to mean that there is no data read despite the fact that the if condition above says the opposite, then why submit a decryption task when nothing is read?

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+
+        public DecryptionGateway(SSLBufferDecrypter m_sslBufferDecrypter, SSLVoltPort port, SSLMessageParser sslMessageParser, ByteBuffer dstBuffer) {
+            this.m_sslBufferDecrypter = m_sslBufferDecrypter;
+            this.m_sslMessageParser = sslMessageParser;
+            this.m_dstBuffer = dstBuffer;
+            this.m_port = port;
+        }
+
+        synchronized void submitDecryptionTasks() {
+            if (m_hasOutstandingTask.compareAndSet(false, true)) {
```
nshi: I don't get the double protection here with the synchronized method and the CAS. Are they redundant?

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+                Runnable task = new Runnable() {
```
nshi: Why create a new `Runnable` every time if you are not using it to capture any local variable or state? You can reuse the same one over and over again.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+                    @Override
+                    public void run() {
+                        DBBPool.BBContainer srcC = null;
+                        boolean queuedMessages = false;
+                        SettableFuture<Void> sf = null;
+                        while ((srcC = getDecryptionFrame()) != null) {
```
nshi: This doesn't look thread-safe. `getDecryptionFrame()` has no protection at all and this calls it from a different thread. Does the threading model implicitly protects it?

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+                                    }
+                                }
+                                srcC.discard();
+                                srcC = null;
+                                m_dstBuffer.clear();
+                                m_decryptedMessages.offer(new DecryptionResult(messages, sf));
+                                sf = null;
+                                queuedMessages = true;
+
+                                if (!m_isShuttingDown.get() && queuedMessages) {
```
nshi: The check for `queuedMessages` is completely redundant here because it's always true.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+        final AtomicBoolean m_isShuttingDown = new AtomicBoolean(false);
+        private final AtomicBoolean m_hasOutstandingTask = new AtomicBoolean(false);
+
+        public EncryptionGateway(SSLBufferEncrypter m_sslBufferEncrypter, SSLVoltPort port) {
+            this.m_sslBufferEncrypter = m_sslBufferEncrypter;
+            this.m_port = port;
+        }
+
+        synchronized void submitEncryptionTasks() {
+            if (m_hasOutstandingTask.compareAndSet(false, true)) {
```
nshi: Same here, why the double protection?

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+                                m_dstBuffer.clear();
+                                return;
+                            } finally{
+                                if (srcC != null) {
+                                    srcC.discard();
+                                }
+                                if (sf != null) {
+                                    sf.setException(new Throwable("Decryption of buffer failed."));
+                                }
+                                m_hasOutstandingTask.set(false);
```
nshi: This prematurely sets the guard `m_hasOutstandingTask` to `false` before it even finishes the while loop.

src/frontend/org/voltcore/network/SSLNIOWriteStream.java
```diff
@@ -0,0 +1,115 @@
...
+
+        if (outCont != null) {
+            if (outCont.b().position() > 0) {
+                getQueuedBuffers().offer(outCont);
+            } else {
+                outCont.discard();
+            }
+        }
+        updateQueued(bytesQueued, true);
+        return processedWrites;
```
nshi: Same here, this looks almost identical to the base method. Can you reuse it?

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: This is the only place getDecryptionFrame is called, and only one thread can be running in the gateway at a time.  The state referenced by getDecryptionFrame() - m_nextFrameLength, m_frameHeader, m_frameCont are called only in that method.  With the exception that m_frameCont is called in unregistered, which is invoked after the gateway is  shutdown.

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: ok...

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: Setting m_hasOutstandingTask and submitting to the executor service need to atomic, I think, so those two statements need to be synchronized.  Also, though, m_hasOutstandingTask is touched by both the network thread and the decryption (executor service) thread, so I think that also needs to be an AtomicBoolean.  Is there a way to simplify.Maybe there's a way to  simplify that logic?

nshi: Either the synchronized keyword or the CAS can protect this block because only one thread will succeed in getting into this block.

src/frontend/org/voltcore/network/SSLVoltPort.java
nshi: It's totally unclear from the code. If this is the case, maybe that method should be a member of the nested class instead of the parent class.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+        } finally {
+            synchronized (m_lock) {
+                assert (m_running == true);
+                m_running = false;
+            }
+        }
+    }
+
+    private void processReads() throws IOException {
+        if (readStream().dataAvailable() == 0) {
```
miketex: My understanding is - 'dataAvailable' indicates how much data is in the read stream.  So the intent of this code is - fill the read stream if its empty.  The read stream is now emptied asynchronously, so it's possibly that data remains from the last time it was read.

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: ok...

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: ok...

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,453 @@
...
+
+    @Override
+    public void run() throws IOException {
+
+        try {
+            if (m_isShuttingDown) {
+                unregistered();
+                return;
+            }
+            processReads();
```
miketex: In handleEncryptedBuffers(), backpressure is turned on when the buffer isn't completely written out.  It's then turned off when the buffer is completed.  This is meant to be analogous to the logic in NIOWriteStream.drainTo().  NIOWriteStream.drainTo() doesn't turn backpressure off when a write succeeds, though; that's an oversight on my part.  
NIOWriteStream.drainTo() checks to see if backpressure can be turned off in its finally block.  I put the same logic in the finally block of handleEncryptedBuffers.  The method that does this is named 'updateWriteStats'; I'll give that a better name.

src/frontend/org/voltcore/network/SSLVoltPort.java
nshi: I see. But why does it have to wait until it's completely emptied before it can read more? Wouldn't this slow things down because they are not well pipelined?

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: ok...

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: ok...

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -195,56 +208,39 @@ public DecryptionGateway(SSLBufferDecrypter m_sslBufferDecrypter, SSLVoltPort po
             this.m_port = port;
         }
 
+        private DBBPool.BBContainer getDecryptionFrame() {
+            if (m_nextFrameLength == 0) {
+                readStream().getBytes(m_frameHeader);
```
nshi: I don't think this is thread-safe. `getBytes()` accesses the available bytes in the read stream without any protection.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,466 @@
...
+        private final Object m_encLock = new Object();
+        private final Runnable m_encTask = new Runnable() {
+            @Override
+            public void run() {
+                DBBPool.BBContainer fragCont = null;
+                boolean queuedEncBuffers = false;
+                SettableFuture<Void> sf = null;
+                while (true) {
+                    synchronized (m_encLock) {
+                        if ((fragCont = writeStream().getWriteBuffer()) == null) {
```
nshi: The write stream already uses a thread-safe queue, you can make it a blocking queue and just `take()` work in the encryption thread. Instead of submitting runnables to the encryption thread, it will always run. `take()` blocks it if there is no work.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,477 @@
...
+        public DecryptionGateway(SSLBufferDecrypter m_sslBufferDecrypter, SSLVoltPort port, SSLMessageParser sslMessageParser, ByteBuffer dstBuffer) {
+            this.m_sslBufferDecrypter = m_sslBufferDecrypter;
+            this.m_sslMessageParser = sslMessageParser;
+            this.m_dstBuffer = dstBuffer;
+            this.m_port = port;
+        }
+
+        private DBBPool.BBContainer getDecryptionFrame() {
+            if (m_nextFrameLength == 0) {
+                readStream().getBytes(m_frameHeader);
```
nshi: This is still not thread-safe. Mike, I think you have to address this issue.

src/frontend/org/voltcore/network/SSLVoltPort.java
```diff
@@ -0,0 +1,477 @@
...
+                    m_hasOutstandingTask = true;
+                } catch (RejectedExecutionException e) {
+                    // the thread pool has been shutdown.
+                }
+            }
+        }
+
+        synchronized boolean shouldContinueTask(boolean didWork) {
+            // if there's no data left on the read stream
+            if (readStream().dataAvailable() == 0) {
```
nshi: This is also not thread-safe.

src/frontend/org/voltcore/network/SSLVoltPort.java
miketex: Yes, I pushed up changes to make SSLNIOReadStream thread safe...

src/frontend/org/voltcore/network/SSLNIOReadStream.java
```diff
@@ -119,21 +151,25 @@ int read(ReadableByteChannel channel, int maxBytes, NetworkDBBPool pool) throws
...
     }
 
-    Deque<DBBPool.BBContainer> getReadContainers() {
-        return m_readBBContainers;
+    @Override
+    public void shutdown() {
+        for (DBBPool.BBContainer c : m_readBBContainers) {
+            c.discard();
+        }
+        m_readBBContainers.clear();
```
nshi: `m_totalAvailable` needs to be cleared as well.

src/frontend/org/voltcore/network/SSLNIOReadStream.java
```diff
@@ -171,5 +171,6 @@ public void shutdown() {
             c.discard();
         }
         m_readBBContainers.clear();
+        m_totalAvailable.set(0);
```
nshi: This may race with the decryption thread if it's in the middle of reading stuff. It could get into `getBytes()` thinking that there are bytes to be read and failed to read anything because they are discarded already.

PR #3958, opened 2016-10-03
ENG-11119: Creating a view with parameters crashes EE

wolffcm:
These changes implement a refactoring of how we execute plan fragments.  My intent is to make the EE more robust to unusual occurrences like long-running fragments (for which we call back into Java and log with some contextual information), and throwing exceptions.

The bug here was that we were allowing parameters to appear in the query of a CREATE VIEW statement.  I added a guard in the planner to disallow parameters.  Without the guard in place, the EE was crashing, which in part prompted the refactoring.

Another blocking bug, ENG-11180, will also make use of this refactoring.

I tried to structure my commits on this branch such that they are easy to read individually.


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -514,34 +502,50 @@ int VoltDBEngine::executePlanFragment(int64_t planfragmentId,
...
-        m_tuplesModifiedStack.pop();
+    // set this to zero for dml operations
+    m_tuplesModifiedStack.push(0);
+
+    // execution lists for planfragments are cached by planfragment id
+    try {
+        // Launch the target plan through its top-most executor list.
+        executorVector->setupContext(m_executorContext);
+        m_executorContext->executeExecutors(0);
+        m_executorContext->cleanupAllExecutors();
```
yzhang1991: I not quite clear with this, so this function will only be called when executing DML that has no output table?

wolffcm: Maybe it became clearer with later commits, but this method is for top-level execution of plan fragments (as opposed to fragments that are executed when another fragment is already running (like updating a MV or subquery, etc).

wolffcm:
Thanks for taking the time to review this Ethan.

PR #3959, opened 2016-10-04
ENG-11185: patch 1

xinjiacs:
Promotion Related Changes:
1. Remove Complete Transaction Response Message Duplicate Counter, and swollen its repair truncation message on leader; 

2. Repaired MP transaction should use SpHandle from TransactionState other than its newest message SpHandle;

Once a while, I saw a similar truncation handle going backwards during process of rejoin, which seems to be a separate problem. Still working on this issue.

### DIFF
--- a/src/frontend/org/voltdb/iv2/SpScheduler.java
+++ b/src/frontend/org/voltdb/iv2/SpScheduler.java
@@ -248,7 +248,10 @@ public class SpScheduler extends Scheduler implements SnapshotCompletionInterest
             final TransactionState txn = m_outstandingTxns.get(key.m_txnId);
             if (txn == null || txn.isDone()) {
                 m_outstandingTxns.remove(key.m_txnId);
-                setRepairLogTruncationHandle(key.m_spHandle);
+                // for MP write txns, we should use it's first SpHandle in the TransactionState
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1282,7 +1274,17 @@ private void setRepairLogTruncationHandle(long newHandle)
             if (m_defaultConsistencyReadLevel == ReadLevel.SAFE) {
                 m_bufferedReadLog.releaseBufferedReads(m_mailbox, m_repairLogTruncationHandle);
             }
-            scheduleRepairLogTruncateMsg();
+
+            if (tellReplica) {
```
nshi: This makes the truncation point hard to reason about. I'm not sure this is the best approach to solve this case. I can imagine debugging an issue where the leader's truncation point has advanced many times without telling any of its replicas, thus deadlocking the system down the road. I have to think about this more.

xinjiacs:
Merged with only one change:
-- Repaired MP transaction should use SpHandle from TransactionState other than its newest message SpHandle;

Will work on different fix for the issue of Complete Transaction Response Message Duplicate Counter.

PR #3962, opened 2016-10-04
ENG-10524: MP support for DR

nshi:
- Upped the protocol version.
- Added two message types.
- New internal mailbox ID for the coordinator.
- Pushed the producer partition ID into the tracker.


luyangco:
Review complete. All looks good to me.

PR #3963, opened 2016-10-04
ENG-11203 view crash on update

yzhang1991:
This pull request is a fix for [ENG-11203](https://issues.voltdb.com/browse/ENG-11203).
In our current version of the single view code the indices created after the view creation is not properly tracked by the `MaterializedViewTriggerForInsert`. The newly-created index may become out of sync with the view data. This will cause a fatal error on deletion.


src/ee/storage/MaterializedViewTriggerForInsert.cpp
```diff
@@ -76,6 +70,33 @@ MaterializedViewTriggerForInsert::~MaterializedViewTriggerForInsert() {
...
+}
+
+void MaterializedViewTriggerForInsert::updateDefinition(PersistentTable *destTable, catalog::MaterializedViewInfo *mvInfo)
+{
+    setTargetTable(destTable);
+    // destTable->allIndexes().size() will be at least ONE because
+    //   we always create an index on the group by keys on view destination tables.
+    // Currently, m_updatableIndexList will include all the indices except the primary key index.
+    // So destTable->allIndexes().size() == m_updatableIndexList.size() + 1
+    // Here we only test the count of indices because indices cannot be changed
```
wolffcm: It's not possible to drop an index on a materialized view?

wolffcm: Suppose you dropped one index and created another in the same DDL batch?

src/ee/storage/MaterializedViewTriggerForInsert.cpp
```diff
@@ -76,6 +70,33 @@ MaterializedViewTriggerForInsert::~MaterializedViewTriggerForInsert() {
...
 
+void MaterializedViewTriggerForInsert::initUpdatableIndexList() {
+    // Note that if the way we initialize this m_updatableIndexList changes in the future, 
+    //   we also need to change the condition to detect when the m_updatableIndexList
+    //   should be refreshed in the updateDefinition() function.
+    const std::vector<TableIndex*>& targetIndexes = m_target->allIndexes();
+    m_updatableIndexList.clear();
+    BOOST_FOREACH(TableIndex *index, targetIndexes) {
+        if (index != m_index) {
+            m_updatableIndexList.push_back(index);
```
wolffcm: (Just for my own understanding...) why do we need to keep a separate list of indexes in the view trigger, can't we just use the list of indexes for the destination table?

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -2120,6 +2120,45 @@ public void testEng11100() throws Exception {
...
+
+            // - 1 - Insert the initial data into the view source table.
+            insertRow(client, initialRow);
+            VoltTable vt = client.callProcedure("@AdHoc",
+                    "SELECT * FROM V_ENG_11203").getResults()[0];
+            assertContentOfTable(new Object[][] {{2, 1, 1}}, vt);
+
+            // - 2 - Now add a new index on the view target table.
+            try {
+                client.callProcedure("@AdHoc", "CREATE INDEX I_ENG_11203 ON V_ENG_11203(a, b);");
```
wolffcm: There is a method in `RegressionSuite` called `verifyStmtFails` that will verify an error message more concisely.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -2120,6 +2120,45 @@ public void testEng11100() throws Exception {
...
+            // - 4 - Start to delete rows.
+            // If the new index was not tracked properly, the server will start to crash
+            // because the newly-inserted row was not inserted into the index.
+            deleteRow(client, initialRow);
+            deleteRow(client, secondRow);
+            vt = client.callProcedure("@AdHoc", "SELECT * FROM V_ENG_11203").getResults()[0];
+            assertContentOfTable(new Object[][] {}, vt);
+        }
+    }
+
```
wolffcm: At standup you mentioned another test case for the views-with-joins?

src/ee/storage/MaterializedViewTriggerForInsert.cpp
yzhang1991: If we do not maintain this list, we will have to form one every time we can updateTupleWithSpecificIndexes() to update the view. To form that list we need to search in the index list for the primary key index and skip it. This may be more expensive.

src/ee/storage/MaterializedViewTriggerForInsert.cpp
yzhang1991: How to do a DDL batch?
I am not aware of that, but yes, if there is one drop and create happening at the same time, then we will not be able to detect here.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
yzhang1991: I don't get it, we are not verifying any error message here, the CREATE INDEX statement should be able to succeed.

wolffcm: Oops, you're right.  It's my reflex to say that every time I see an ad hoc call wrapped in a `try`.

src/ee/storage/MaterializedViewTriggerForInsert.cpp
wolffcm: I see.  We have this problem in UpdateExecutor, where we recalculate the index list for every time we invoke `p_execute`.

src/ee/storage/MaterializedViewTriggerForInsert.cpp
```diff
@@ -76,6 +70,24 @@ MaterializedViewTriggerForInsert::~MaterializedViewTriggerForInsert() {
...
+        }
+    }
+}
+
+void MaterializedViewTriggerForInsert::updateDefinition(PersistentTable *destTable, catalog::MaterializedViewInfo *mvInfo) {
+    setTargetTable(destTable);
+    initUpdatableIndexList();
```
yzhang1991: Here I just removed the condition to check when to initialize this index list.
As noted in Chris' previous review comment, the condition I used to have cannot catch the case when an index creation and a deletion happened in the same DDL batch. Although I do not know how to do that DDL batch.

src/ee/storage/persistenttable.cpp
```diff
@@ -2009,9 +2007,19 @@ void PersistentTable::dropViewHandler(MaterializedViewHandler *viewHandler) {
...
+    // from the table to update the joined table view handler properly.
+    //   If the current table is a view source table, adding / dropping an index
+    // may change the plan to refresh the view, therefore all the handlers that
+    // use the current table as one of their sources need to be updated.
+    //   If the current table is a view target table, we need to refresh the list
+    // of tracked indices so that the data in the table and its indices can be in sync.
     BOOST_FOREACH (auto mvHanlder, m_viewHandlers) {
         mvHanlder->pollute();
     }
+    if (m_mvHandler) {
```
yzhang1991: I thought the `polluteViews()` caught the case in this ticket but it didn't.
Thanks to Chirs for reminding me this.

yzhang1991:
Once a view handler is "polluted", in the final stage of catalog update it will be reconstructed. So it will get a brand new list of indices. Same as the single table view case.

PR #3966, opened 2016-10-06
ENG-11180: avoid crash when create view is long-running

wolffcm:
This does some refactoring of the JNI layer on the Java side to keep track of whether we're executing a RO procedure, a RW one, or a catalog update, and modifies the long running query message to suit.

On the EE side I had to bypass using the "current fragment" as tracked by the VoltDBEngine, since we don't set it for catalog updates.  The only reason we needed it is for reporting usage of temp table memory, so I made it so we stored this info in the monitor proxy.

These changes make use of a refactoring (soon to be merged) that I did here:
https://github.com/VoltDB/voltdb/pull/3958
which is why I've made this pull request against that branch.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -139,7 +139,7 @@ typedef boost::multi_index::multi_index_container<
 class EnginePlanSet : public PlanSet { };
 
 VoltDBEngine::VoltDBEngine(Topend *topend, LogProxy *logProxy)
-    : m_currentIndexInBatch(0),
+    : m_currentIndexInBatch(-1),
       m_tuplesProcessedInBatch(0),
       m_tuplesProcessedInFragment(0),
       m_tuplesProcessedSinceReport(0),
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -423,6 +423,8 @@ int VoltDBEngine::executePlanFragments(int32_t numFragments,
         m_stringPool.purge();
     }
 
+    m_currentIndexInBatch = -1;
+
     return failures;
 }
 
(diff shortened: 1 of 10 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -423,6 +423,8 @@ int VoltDBEngine::executePlanFragments(int32_t numFragments,
         m_stringPool.purge();
     }
 
+    m_currentIndexInBatch = -1;
```
paulmartel: Should this reset be made exception-safe for when executePlanFragment throws a recoverable exception?

src/ee/execution/VoltDBEngine.h
```diff
@@ -668,19 +670,21 @@ inline int64_t VoltDBEngine::pullTuplesRemainingUntilProgressReport(PlanNodeType
     return m_tupleReportThreshold - m_tuplesProcessedSinceReport;
 }
 
-inline int64_t VoltDBEngine::pushTuplesProcessedForProgressMonitoring(int64_t tuplesProcessed)
+inline int64_t VoltDBEngine::pushTuplesProcessedForProgressMonitoring(const TempTableLimits* limits,
```
paulmartel: The way this signature had to be changed in two places is my main argument against inline methods, especially short ones like these, being defined outside the class but inside the header . I'd make an exception only for the rare case where they cause circular dependencies.

src/frontend/org/voltdb/jni/ExecutionEngine.java
```diff
@@ -431,43 +440,58 @@ public long fragmentProgressUpdate(int indexFromFragmentTask,
...
+            sb.append("Catalog load");
+            break;
+        }
+
+        // Timing out (canceling) versus just reporting something taking a long time
+        sb.append(timeout ? " is timed out at " : " is taking a long time to execute -- at least ");
+        sb.append(String.format("%.2f seconds spent accessing %d tuples. ",
+                latency / 1000.0, m_lastTuplesAccessed));
+
+        // Type of plan node executing, and index in batch in known.
```
paulmartel: typo "in known"

src/frontend/org/voltdb/jni/ExecutionEngine.java
```diff
@@ -431,43 +440,58 @@ public long fragmentProgressUpdate(int indexFromFragmentTask,
...
+        sb.append("Current plan fragment " + PlanNodeType.get(planNodeTypeAsInt).name());
+        if (indexFromFragmentTask >= 0) {
+            sb.append(" in call " + m_currentBatchIndex + " to voltExecuteSQL");
+        }
+        sb.append(" on site " + CoreUtils.hsIdToString(m_siteId) + ". ");
+
+        if (m_currMemoryInBytes > 0 && m_peakMemoryInBytes > 0) {
+            sb.append("Current temp table uses " + m_currMemoryInBytes + " bytes memory,"
+                    + " and the peak usage of memory for temp table is " + m_peakMemoryInBytes + " bytes.");
```
paulmartel: fix message typo "for temp table" --> "for temp tables"

src/ee/execution/VoltDBEngine.cpp
paulmartel: This was the only possible blocker -- my other review comments in this pass (complete) were minor issues.

wolffcm: Recoverable exceptions (which will abort the current fragment and the subsequent ones but not crash the EE) are caught in executePlanFragments, which will return an error code.  So we're guaranteed to execute this line for recoverable exceptions.

wolffcm:
Merging on separate branch.

PR #3973, opened 2016-10-09
Relax if voltdb is used through links

patope:
`voltenv` script checks is it used through symblic and stops. However homebrew uses symbolic links to installed software and this do not work without tweaks. This doesn't mean classpaths generated by script would be incorrect. see. https://github.com/Homebrew/homebrew-core/pull/5713

This patch relaxes this check. Changes are:
- Write warning to stderr, but continue execution. eg. `export CP=$(voltenv clientclasspath)` would work.
- Remove duplicate `clientclasspath` help text

### DIFF
--- a/bin/voltenv
+++ b/bin/voltenv
@@ -67,9 +67,6 @@ function envhelp() {
     echo "voltenv clientclasspath"
     echo "  Print out a classpath that contains the VoltDB client jars"
     echo
-    echo "voltenv clientclasspath"
-    echo "  Print out a classpath that contains the VoltDB client jars"
-    echo
     echo "A common use for this file is to source it from other scripts"
     echo " in order to set up the VoltDB environment."
     echo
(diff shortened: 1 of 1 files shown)


bin/voltenv
```diff
@@ -97,31 +94,32 @@ function checkpath() {
...
+            echo "    $VOLTDB_BIN" >&2
+            echo >&2
         fi
     else
         # assume this is the examples folder for a kit
         echo "The VoltDB scripts are not in your PATH."
         echo "For ease of use, add the VoltDB bin directory to your path: "
         echo "    $VOLTDB_BIN"
         echo
+        exit 1
```
jhugg: Why "exit 1" here. I don't think we want to stop the script if voltdb isn't in the path. Running inside a newly untarred directory is a pretty common use case.

patope: yes, i'll remove it.

PR #3974, opened 2016-10-11
ENG-11152: Add JStack output to VoltLog of the MPI host when a possib…

wweiss-voltdb:
…le MPDeadlock is detected.

### DIFF
--- a/src/frontend/org/voltdb/VoltDB.java
+++ b/src/frontend/org/voltdb/VoltDB.java
@@ -39,6 +39,8 @@ import java.util.Queue;
 import java.util.TimeZone;
 import java.util.UUID;
 
+import javax.management.MBeanServer;
+
 import org.voltcore.logging.VoltLog4jLogger;
 import org.voltcore.logging.VoltLogger;
 import org.voltcore.messaging.HostMessenger;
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/VoltDB.java
```diff
@@ -1025,6 +1031,76 @@ private static boolean turnOffClientInterface() {
         return true;
     }
 
+    public static String GenerateThreadDump() {
```
nshi: You may want to consider wrapping the whole function in a try...catch block so that it never throws.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -1025,6 +1031,76 @@ private static boolean turnOffClientInterface() {
...
+                dump.append("\n\n");
+            }
+        }
+        return dump.toString();
+    }
+
+    /*
+     * WARNING: Use with care this may take a long time and generate a GC. It could even cause a DEAD HOST Timeout.
+     */
+    public static void GenerateHeapDump() {
```
nshi: Same for this. It should never throw.

src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -318,6 +319,7 @@ private FragmentResponseMessage pollForResponses()
                         }
                     }
                     m_mbox.send(com.google_voltpatches.common.primitives.Longs.toArray(m_useHSIds), new DumpMessage());
+                    tmLog.warn(VoltDB.GenerateThreadDump());
```
nshi: This only prints the thread dump on the MPI node, which may not be the node that's blocking the transaction.

PR #3977, opened 2016-10-11
ENG-11259: Support versioned StoredProcedureInvocation serialization.

nshi:
We already support deserializing different versions of SPI. This is to
add serialization of different versions for DR.

### DIFF
--- a/src/frontend/org/voltdb/StoredProcedureInvocation.java
+++ b/src/frontend/org/voltdb/StoredProcedureInvocation.java
@@ -215,6 +215,34 @@ public class StoredProcedureInvocation implements JSONString {
         // 2 is one byte for ext type, one for size
         int allPartitionExtensionSize = m_allPartition ? 2 : 0;
 
+        // compute the size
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/StoredProcedureInvocation.java
```diff
@@ -273,7 +289,7 @@ public static int getLoadVoltTablesMagicSeriazlizedSize(Table catTable, boolean
         // Magic size of @LoadVoltTable* StoredProcedureInvocation
         int tableNameLengthInBytes =
                 catTable.getTypeName().getBytes(Constants.UTF8ENCODING).length;
-        int metadataSize = 42 + tableNameLengthInBytes;
+        int metadataSize = 41 + tableNameLengthInBytes;
```
jhugg: Oh, we should add a comment that says this is using the original serialization format.
Also update the commented out code to use the new methods for serializing with the original version.

PR #3978, opened 2016-10-12
ENG-11110: Add a dummy transaction task message at the end of SP lead…

xinjiacs:
…er promotion. The request will be duplicated to replicas like a SP write. It does not do anything useful work, but as a signal to flush everyone's transaction queue before the newly promoted leader starting to run new transactions.

### DIFF
--- a/src/frontend/org/voltdb/iv2/DummyTransactionTask.java
+++ b/src/frontend/org/voltdb/iv2/DummyTransactionTask.java
@@ -1,0 +1,66 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2016 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.voltdb.iv2;
+
+import java.io.IOException;
+
+import org.voltcore.messaging.Mailbox;
+import org.voltdb.SiteProcedureConnection;
+import org.voltdb.dtxn.TransactionState;
+import org.voltdb.messaging.DummyTransactionResponseMessage;
+import org.voltdb.rejoin.TaskLog;
+
+/**
+ * This dummy transaction task is a SP read only transaction that will be put into transaction task queue.
+ * And it will be duplicated to followers as well. This task acts as a synchronization transaction on all
+ * partition replicas like normal SP write transaction.
+ */
+public class DummyTransactionTask extends TransactionTask {
+    final Mailbox m_initiator;
+
+    public DummyTransactionTask(Mailbox initiator,
+            TransactionState txnState, TransactionTaskQueue queue) {
+        super(txnState, queue);
+        m_initiator = initiator;
+    }
+
+    private void generateDummyResponse() {
+        doCommonSPICompleteActions();
+        DummyTransactionResponseMessage response = new DummyTransactionResponseMessage(this);
+        response.m_sourceHSId = m_initiator.getHSId();
+        m_initiator.deliver(response);
+    }
+
+    @Override
+    public void run(SiteProcedureConnection siteConnection) {
+        generateDummyResponse();
+    }
+
+    @Override
+    public void runFromTaskLog(SiteProcedureConnection siteConnection) {
+        generateDummyResponse();
+    }
+
+    @Override
+    public void runForRejoin(SiteProcedureConnection siteConnection, TaskLog rejoinTaskLog) throws IOException {
+        generateDummyResponse();
+    }
+
+    public long getSPIHSId() {
+        return m_txnState.initiatorHSId;
+    }
+}
--- a/src/frontend/org/voltdb/iv2/SpInitiator.java
+++ b/src/frontend/org/voltdb/iv2/SpInitiator.java
@@ -40,6 +40,7 @@ import org.voltdb.VoltZK;
 import org.voltdb.export.ExportManager;
 import org.voltdb.iv2.RepairAlgo.RepairResult;
 import org.voltdb.iv2.SpScheduler.DurableUniqueIdListener;
+import org.voltdb.messaging.DummyTransactionTaskMessage;
 
 import com.google_voltpatches.common.collect.ImmutableMap;
 
@@ -175,6 +176,14 @@ public class SpInitiator extends BaseInitiator implements Promotable
                     LeaderCacheWriter iv2masters = new LeaderCache(m_messenger.getZK(),
                             m_zkMailboxNode);
                     iv2masters.put(m_partitionId, m_initiatorMailbox.getHSId());
+
+                    // After SP leader promotion, a DummyTransactionTaskMessage is generated from the new leader.
+                    // This READ ONLY message will serve as a synchronization point on all replicas of this
+                    // partition, like normal SP write transaction that has to finish executing on all replicas.
+                    // In this way, the leader can make sure all replicas have finished replaying
+                    // all their repair logs entries.
+                    // From now on, the new leader is safe to accept new transactions. See ENG-11110.
+                    m_initiatorMailbox.deliver(new DummyTransactionTaskMessage());
                 }
                 else {
                     // The only known reason to fail is a failed replica during
--- a/src/frontend/org/voltdb/iv2/SpInitiator.java
+++ b/src/frontend/org/voltdb/iv2/SpInitiator.java
@@ -40,7 +40,6 @@ import org.voltdb.VoltZK;
 import org.voltdb.export.ExportManager;
 import org.voltdb.iv2.RepairAlgo.RepairResult;
 import org.voltdb.iv2.SpScheduler.DurableUniqueIdListener;
-import org.voltdb.messaging.DummyTransactionTaskMessage;
 
 import com.google_voltpatches.common.collect.ImmutableMap;
 
@@ -176,14 +175,6 @@ public class SpInitiator extends BaseInitiator implements Promotable
                     LeaderCacheWriter iv2masters = new LeaderCache(m_messenger.getZK(),
                             m_zkMailboxNode);
                     iv2masters.put(m_partitionId, m_initiatorMailbox.getHSId());
-
-                    // After SP leader promotion, a DummyTransactionTaskMessage is generated from the new leader.
-                    // This READ ONLY message will serve as a synchronization point on all replicas of this
-                    // partition, like normal SP write transaction that has to finish executing on all replicas.
-                    // In this way, the leader can make sure all replicas have finished replaying
-                    // all their repair logs entries.
-                    // From now on, the new leader is safe to accept new transactions. See ENG-11110.
-                    m_initiatorMailbox.deliver(new DummyTransactionTaskMessage());
                 }
                 else {
                     // The only known reason to fail is a failed replica during
--- a/src/frontend/org/voltdb/iv2/SpScheduler.java
+++ b/src/frontend/org/voltdb/iv2/SpScheduler.java
@@ -185,6 +185,15 @@ public class SpScheduler extends Scheduler implements SnapshotCompletionInterest
     {
         super.setLeaderState(isLeader);
         m_snapMonitor.addInterest(this);
+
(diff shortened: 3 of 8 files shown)


src/frontend/org/voltdb/iv2/DummyTransactionTask.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.IOException;
+
+import org.voltcore.messaging.Mailbox;
+import org.voltdb.SiteProcedureConnection;
+import org.voltdb.dtxn.TransactionState;
+import org.voltdb.messaging.DummyTransactionResponseMessage;
+import org.voltdb.rejoin.TaskLog;
+
+public class DummyTransactionTask extends TransactionTask {
```
nshi: Should add a comment here explaining what this is for.

src/frontend/org/voltdb/iv2/InitiatorMailbox.java
```diff
@@ -38,9 +38,9 @@
 import org.voltdb.messaging.Iv2RepairLogRequestMessage;
 import org.voltdb.messaging.Iv2RepairLogResponseMessage;
 import org.voltdb.messaging.RejoinMessage;
+import org.voltdb.messaging.RepairLogTruncationMessage;
```
nshi: This change is completely unnecessary.

src/frontend/org/voltdb/iv2/SpInitiator.java
```diff
@@ -175,6 +176,12 @@ public void acceptPromotion()
                     LeaderCacheWriter iv2masters = new LeaderCache(m_messenger.getZK(),
                             m_zkMailboxNode);
                     iv2masters.put(m_partitionId, m_initiatorMailbox.getHSId());
+
+                    // After SP leader promotion, a DummyTransactionTaskMessage is generated from the new leader.
+                    // This message will be put in transaction task queue and duplicated for response from every replica.
```
nshi: "duplicated for response from every replica" is confusing.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -411,7 +419,7 @@ private long getMaxTaskedSpHandle() {
 
     // SpScheduler expects to see InitiateTaskMessages corresponding to single-partition
     // procedures only.
-    public void handleIv2InitiateTaskMessage(Iv2InitiateTaskMessage message)
+    private void handleIv2InitiateTaskMessage(Iv2InitiateTaskMessage message)
```
nshi: These changes are completely unnecessary.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1143,6 +1151,61 @@ public void handleDumpMessage()
...
     }
 
+    private void handleDummyTransactionTaskMessage(DummyTransactionTaskMessage message)
+    {
+        DummyTransactionTaskMessage msg = message;
+        long uniqueId, newSpHandle;
+        if (m_isLeader) {
+            TxnEgo ego = advanceTxnEgo();
+            newSpHandle = ego.getTxnId();
+            uniqueId = m_uniqueIdGenerator.getNextUniqueId();
```
nshi: You don't really need a unique ID for this message.

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
```diff
@@ -0,0 +1,93 @@
...
+public class DummyTransactionTaskMessage extends TransactionInfoBaseMessage
+{
+    public DummyTransactionTaskMessage()
+    {
+        super();
+        m_isReadOnly = true;
+        m_isForReplay = false;
+    }
+
+    public DummyTransactionTaskMessage (long initiatorHSId, long coordinatorHSId, long txnId, long uniqueId) {
```
nshi: Isn't `initiatorHSId` and `coordinatorHSId` always the same? If so, one of the parameters can go away.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1143,6 +1151,61 @@ public void handleDumpMessage()
...
+
+            msg = new DummyTransactionTaskMessage(m_mailbox.getHSId(),
+                    m_mailbox.getHSId(), newSpHandle, uniqueId);
+            msg.setSpHandle(newSpHandle);
```
nshi: Why set the spHandle right after you passed it in to the constructor?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1143,6 +1151,61 @@ public void handleDumpMessage()
...
+                safeAddToDuplicateCounterMap(new DuplicateCounterKey(msg.getTxnId(), newSpHandle), counter);
+            }
+        } else {
+            setMaxSeenTxnId(msg.getSpHandle());
+        }
+        Iv2Trace.logDummyTransactionTaskMessage(msg, m_mailbox.getHSId());
+
+        DummyTransactionTask task = new DummyTransactionTask(m_mailbox,
+                new SpTransactionState(msg), m_pendingTasks);
+        m_pendingTasks.offer(task);
```
nshi: You need to pass it to the command logging thread to flush the queue there, too. It can't go into the transaction task queue directly.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1143,6 +1151,61 @@ public void handleDumpMessage()
...
+    }
+
+    private void handleDummyTransactionResponseMessage(DummyTransactionResponseMessage message) {
+        final long spHandle = message.getSpHandle();
+        final DuplicateCounterKey dcKey = new DuplicateCounterKey(message.getTxnId(), spHandle);
+        DuplicateCounter counter = m_duplicateCounters.get(dcKey);
+        if (counter == null) {
+            // this will be on SPI without k-safety or replica only with k-safety
+            setRepairLogTruncationHandle(spHandle);
```
nshi: Since the dummy transaction is a read-only one, you shouldn't advance the spHandle for the dummy message. That saves you the trouble of advancing the truncation handle.

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.voltcore.messaging.TransactionInfoBaseMessage;
+import org.voltcore.utils.CoreUtils;
+import org.voltdb.iv2.TxnEgo;
+import org.voltdb.iv2.UniqueIdGenerator;
+
+/**
+ * Message from a client interface to an initiator, instructing the
+ * site to begin executing a stored procedure, coordinating other
+ * execution sites if needed.
```
nshi: Please fix copy/pasted comment.

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
```diff
@@ -0,0 +1,93 @@
...
+        m_isForReplay = false;
+    }
+
+    @Override
+    public boolean isSinglePartition() {
+        return true;
+    }
+
+    @Override
+    public int getSerializedSize()
```
nshi: You don't have to override it if you are not changing the implementation.

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
```diff
@@ -0,0 +1,93 @@
...
+    {
+        buf.put(VoltDbMessageFactory.DUMMY_TRANSACTION_TASK_ID);
+        super.flattenToBuffer(buf);
+
+        assert(buf.capacity() == buf.position());
+        buf.limit(buf.position());
+    }
+
+    @Override
+    public void initFromBuffer(ByteBuffer buf) throws IOException {
```
nshi: Same here.

src/frontend/org/voltdb/iv2/SpScheduler.java
xinjiacs: Not really here for this case. The problem we are fixing is the next new SP RO txn will be released immediately because of the early committed last repaired SP. When the dummy txn  advances the SpHandle, the following new SP RO txns will not be released until the dummy txn gets committed (when SPI gets acked from all replicas by dummy responses).

src/frontend/org/voltdb/iv2/DummyTransactionTask.java
```diff
@@ -0,0 +1,60 @@
...
+    final Mailbox m_initiator;
+
+    public DummyTransactionTask(Mailbox initiator,
+            TransactionState txnState, TransactionTaskQueue queue) {
+        super(txnState, queue);
+        m_initiator = initiator;
+    }
+
+    private void generateDummyResponse() {
+        DummyTransactionResponseMessage response = new DummyTransactionResponseMessage(this);
```
nshi: You also have to mark the transaction state as done. Otherwise, `SpScheduler.updateReplicas()` won't repair it correctly, which may deadlock the system.

src/frontend/org/voltdb/iv2/SpInitiator.java
xinjiacs: I updated this message. Check it again. 
I still have not proved that we have a bug not picking the longest command log to replay.

src/frontend/org/voltdb/iv2/DummyTransactionTask.java
xinjiacs: Good catch.

src/frontend/org/voltdb/iv2/SpScheduler.java
xinjiacs: I think it's better being private.

src/frontend/org/voltdb/iv2/SpScheduler.java
xinjiacs: The reason is the super constructor sets SpHandle member variable as Long.MIN_VALUE as default. 

If uniqueId is not used, we can put this setter function inside of the constructor of DummyTransactionTaskMessage.

src/frontend/org/voltdb/iv2/SpScheduler.java
xinjiacs: The DummyTransactionTask is a read only task, why it needs to be command logged?

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
xinjiacs: Right. I guess it's the legacy of copy/past.

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
xinjiacs: Initially when I added extra member variables, I need it. After removing them, I end up leaving it there. Good catch.

src/frontend/org/voltdb/iv2/SpScheduler.java
nshi: It doesn't need to be command logged, but you have to flush the command log queue. Otherwise, there may be transactions sitting in the command log queue and you'll miss them and run into the same bug again.

src/frontend/org/voltdb/iv2/SpScheduler.java
nshi: You can set the spHandle directly in the constructor. Unique ID is not needed for this message.

src/frontend/org/voltdb/iv2/InitiatorMailbox.java
nshi: Yes, it's better to organize them together. On the other hand, this is completely unnecessary for this bug fix. I recommend to **not** reorganize imports because it generates a lot of unnecessary changes in commits. For example, if my IDE thinks it's better to organize imports in a slightly different way than yours, we'll constantly generate unnecessary commit changes when we touch the same files.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1154,15 +1154,12 @@ private void handleDumpMessage()
     private void handleDummyTransactionTaskMessage(DummyTransactionTaskMessage message)
     {
         DummyTransactionTaskMessage msg = message;
-        long uniqueId, newSpHandle;
+        long newSpHandle;
```
nshi: This variable can go inside the leader if branch.

src/frontend/org/voltdb/messaging/DummyTransactionTaskMessage.java
```diff
@@ -23,13 +23,10 @@
...
 import org.voltdb.iv2.TxnEgo;
-import org.voltdb.iv2.UniqueIdGenerator;
 
 /**
- * Message from a client interface to an initiator, instructing the
- * site to begin executing a stored procedure, coordinating other
- * execution sites if needed.
- *
+ * Message from an initiator just after newly leader promotion, instructing the
+ * site to begin executing a stored procedure, coordinating other execution sites if needed.
```
nshi: This comment is still incorrect. Please read the whole sentence.

PR #3979, opened 2016-10-13
grammar.sql: added IS NOT DISTINCT FROM to the list of comparison

johnTcrawford:
operators.

PostgreSQLBackend.java: subtle change to one regex expression, to make
sure that it does not confuse the FROM in IS [NOT] DISTINCT FROM with
the FROM in SELECT ... FROM T1 ... (otherwise, such confusion causes
UPSERT statements to be translated into invalid INSERT / ON CONFLICT DO
statements); and corrected a couple of nearby comments.

### DIFF
--- a/tests/scripts/sql_coverage_test.py
+++ b/tests/scripts/sql_coverage_test.py
@@ -228,17 +228,17 @@ def get_max_mismatches(comparison_database, suite_name):
         # and basic-compoundex-joins "extended" test suites (see ENG-10775)
         if (config_name == 'basic-joins' or config_name == 'basic-index-joins' or
               config_name == 'basic-compoundex-joins'):
-            max_mismatches = 4620
+            max_mismatches = 5280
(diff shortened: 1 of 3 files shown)


tests/scripts/sql_coverage_test.py
```diff
@@ -228,17 +228,17 @@ def get_max_mismatches(comparison_database, suite_name):
         # and basic-compoundex-joins "extended" test suites (see ENG-10775)
         if (config_name == 'basic-joins' or config_name == 'basic-index-joins' or
               config_name == 'basic-compoundex-joins'):
-            max_mismatches = 4620
+            max_mismatches = 5280
```
wolffcm: I guess you had to bump up the number of expected mismatches since we're now generating more queries due to the extra operator?

johnTcrawford: Yes, exactly.

PR #3980, opened 2016-10-13
VoltDecimalHelper.java: avoid negative values being passed to the

johnTcrawford:
MathContext constructor (as the precision, in the roundToScale method)
and being used as an array index (of the scaleFactors array, in the
serializeBigDecimal method), since they were causing exceptions, when
running sqlcoverage's numeric-decimals test suite against PostgreSQL.

sql_coverage_test.py: in get_max_mismatches, removed the expected
failures for the numeric-decimals test suite.

### DIFF
--- a/src/frontend/org/voltdb/types/VoltDecimalHelper.java
+++ b/src/frontend/org/voltdb/types/VoltDecimalHelper.java
@@ -206,7 +206,7 @@ public class VoltDecimalHelper {
         if (!isRoundingEnabled()) {
             throw new RuntimeException(String.format("Decimal scale %d is greater than the maximum %d", bd.scale(), kDefaultScale));
         }
-        int desiredPrecision = bd.precision() - lostScaleDigits;
+        int desiredPrecision = Math.max(0, bd.precision() - lostScaleDigits);
         MathContext mc = new MathContext(desiredPrecision, mode);
         BigDecimal nbd = bd.round(mc);
         if (nbd.scale() != scale) {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/types/VoltDecimalHelper.java
```diff
@@ -206,7 +206,7 @@ static private final BigDecimal roundToScale(BigDecimal bd, int scale, RoundingM
         if (!isRoundingEnabled()) {
             throw new RuntimeException(String.format("Decimal scale %d is greater than the maximum %d", bd.scale(), kDefaultScale));
         }
-        int desiredPrecision = bd.precision() - lostScaleDigits;
+        int desiredPrecision = Math.max(0, bd.precision() - lostScaleDigits);
```
wolffcm: I was having trouble understanding what it means to have a precision of zero.  It seems like zero precision is a special value meaning unlimited precision?

https://docs.oracle.com/javase/7/docs/api/java/math/MathContext.html#UNLIMITED

This should perhaps be a `1` instead of a `0`?

paulmartel: Looks like Chris is right -- best to use 1 as minimum allowed precision to avoid any special semantics for `(precision == 0)`. The referenced Oracle doc is pretty backhanded in how it calls out `(precision == 0)` as having a special meaning. The special meaning is strongly implied for one obscure internal use case but I don't see it documented for other more direct use cases or as a general principle. It's POSSIBLE that changing to 1 could cause regressions that may need a more complex workaround, but it's worth a try. BTW, the method comment looks garbled and incomplete. Please change to:

        /**
         * Round a BigDecimal number to a scale given the rounding mode.
         * Note that the precision of the result can depend not only on its original
         * precision and scale and the desired scale, but also on its value.
         * For example, when rounding up with scale 2,
         *     9.1999 with input scale 4 and precision 5 returns 9.20 with precision 3 (down 2).
         *     9.9999 with input scale 4 and precision 5 returns 10.00 with precision 4 (down 1).
         *     91.9999 with input scale 4 and precision 6 returns 92.00 with precision 4 (down 2).
         * @param bd the input value of arbitrary scale and precision
         * @param scale the desired scale of the return value
         * @param mode the rounding algorithm to use
         * @return the rounded value approximately equal to bd but having the desired scale 
         */

johnTcrawford:
I have taken Chris's / Paul's suggestions, and plan to merge this, once the Jenkins runs complete successfully (which I expect they will).

PR #3981, opened 2016-10-13
ENG-10626: One-to-many active-passive DR happy path, rejoin, recover

slowsoul:
This PR will include happy path, rejoin/recover handling. The DR reset functionality for individual and all conversations and the conversation collision handling which is dependent on this functionality will be addressed in another ticket and PR.

Changes in community repo:
- Add closeCursor() interface for BinaryDeque, for snapshot placeholder cursor in the case of snapshot discard and for the normal cursor in the case of conversation reset.
- Since snapshot placeholder cursor can prevent several PBD files from being deleted, after it goes away, a relatively new PBD file may be the first one eligible to be deleted. When this happens, PBD files older than this one should be deleted as well. Update the behavior in all 3 places that PBD file gets deleted, also update the m_segments map correspondingly.
- There were two synchronized functions for PBD buffer count and size separately, but there are cases where they are used together to initialize a DRPartitionStreamReader instance. Add a single synchronized function to return a pair of count and size.
- Bump the protocol version from 6 to 7 (an integration branch may be needed to avoid increasing the version prematurely, considering incorporating other associated tickets into this version)
- Remove the flag so that consumer cluster id is always in DR stats table.

### DIFF
--- a/src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
+++ b/src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
@@ -733,7 +733,6 @@ public class PersistentBinaryDeque implements BinaryDeque {
     }
 
     @Override
-    //TODO: Should we get rid of persisted cursors on close?
     public synchronized void close() throws IOException {
         if (m_closed) {
             return;
@@ -743,6 +742,7 @@ public class PersistentBinaryDeque implements BinaryDeque {
         for (PBDSegment segment : m_segments.values()) {
             segment.close();
         }
+        m_segments.clear();
         m_closed = true;
     }
 
@@ -772,13 +772,17 @@ public class PersistentBinaryDeque implements BinaryDeque {
 
     @Override
     public synchronized void closeAndDelete() throws IOException {
-        if (m_closed) return;
-        m_closed = true;
+        if (m_closed) {
+            return;
+        }
+        m_readCursors.clear();
+
         for (PBDSegment qs : m_segments.values()) {
             m_usageSpecificLog.debug("Segment " + qs.file() + " has been closed and deleted due to delete all");
             closeAndDeleteSegment(qs);
         }
         m_segments.clear();
+        m_closed = true;
     }
 
     public static class ByteBufferTruncatorResponse extends TruncatorResponse {
--- a/src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
+++ b/src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
@@ -221,11 +221,20 @@ public class PersistentBinaryDeque implements BinaryDeque {
                                 return;
                             }
                             if (canDeleteSegment(segment)) {
-                                m_segments.remove(segment.segmentId());
-                                if (m_usageSpecificLog.isDebugEnabled()) {
-                                    m_usageSpecificLog.debug("Segment " + segment.file() + " has been closed and deleted after discarding last buffer");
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -677,6 +690,18 @@ public synchronized BinaryDequeReader openForRead(String cursorId) throws IOExce
         return reader;
     }
 
+    @Override
+    public synchronized void closeCursor(String cursorId) {
+        ReadCursor reader = m_readCursors.remove(cursorId);
+        if (reader != null && reader.m_segment != null) {
+            try {
+                reader.m_segment.getReader(cursorId).close();
```
manjujames: Should we also check if segments can be closed and deleted as a result of closing cursors? I guess it's OK if we don't delete at this point because the next ack that deletes a segment after this will delete this as well.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -221,11 +222,20 @@ public void discard() {
...
+                                    Map.Entry<Long, PBDSegment> entry = iter.next();
+                                    if (entry.getKey() > segment.segmentId()) {
+                                        break;
+                                    }
+                                    PBDSegment segmentToDelete = entry.getValue();
+                                    if (m_usageSpecificLog.isDebugEnabled()) {
+                                        m_usageSpecificLog.debug("Segment " + segmentToDelete.file() + " has been closed and deleted after discarding last buffer");
+                                    }
+                                    closeAndDeleteSegment(segmentToDelete);
+                                    iter.remove();
```
nshi: This can be extracted and be reused below.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
nshi: I think that depends on the semantic of this method, which is not completely clear. If this method is meant to just close the FD, then it shouldn't delete the segment. If it means to remove the cursor, then it should.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -718,6 +742,7 @@ public synchronized void close() throws IOException {
         for (PBDSegment segment : m_segments.values()) {
             segment.close();
         }
+        m_segments.clear();
```
nshi: Is this correct? It doesn't look right to forget about all segments on segment close, which just closes the FDs. What happens to the segment files on disk? Are they just leaked?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -735,14 +760,29 @@ public synchronized long sizeInBytes() throws IOException {
         return size;
     }
 
+    public synchronized Pair<Integer, Long> getBufferCountAndSize() throws IOException {
```
nshi: Maybe you can remove the other two methods and use this one.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
manjujames: Yes, I agree. But, there is no way to open a PBD once it is closed, so the PBD is unusable once it is closed, so it doesn't really matter.

nshi:
Landed on integration branch.

PR #3983, opened 2016-10-14
ENG-11275 block transactions during shutdown

zeemanhe:
When a cluster in PAUSED mode is force-shutdown, the cluster will be set to SHUTTINGDOWN mode. There may be a window before the cluster process is killed within which transactions can go through.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2683,6 +2683,8 @@ public boolean shutdown(Thread mainSiteThread) throws InterruptedException {
         synchronized(m_startAndStopLock) {
             boolean did_it = false;
             if (m_mode != OperationMode.SHUTTINGDOWN) {
+                //block all traffic other than some Sysprocs which are allowed during shutting down.
```
xinjiacs: Maybe adding a new line comment?
"FYI, the same sysprocs allowed in prepare shutting down process".

xinjiacs:
The other looks good to me.

PR #3986, opened 2016-10-17
AAUIW sqlcmd connection exception details output if unable to connect

benjaminballard:
A customer requested this, they were getting more helpful errors if sqlcmd failed prior to v6.5 which allows sqlcmd to continue if given a list of servers, as long as it gets at least one connection, which had the side effect of hiding the original connection exception messages.


wolffcm:
I had a peek at how this branch is doing with sqlcmdtest, our test suite that covers sqlcmd.

http://ci.voltdb.lan/view/Branch-jobs/view/branch-ENG-11293/job/branch-2-sqlcmdtest-ENG-11293/

This change is causing sqlcmdtest to fail, so the test will need to be updated to merge this.

Also, I noticed that the error diagnostics don't seem all that helpful---output from sqlcmdtest:
```
Unable to connect to VoltDB cluster
    fakeserver1:21212 - fakeserver1
    fakeserver2:21212 - fakeserver2
```
I guess the `getMessage` just produces the server name that couldn't be found?  It would be good if "unknown host" appeared in here somewhere.  This was broke before you did anything, but since you're here, if you could improve the message that would be great.

benjaminballard:
I think UnknownHostException must have a blank message, while IOException is more generic and probably has some additional detail.  I can fix that.

benjaminballard:
Chris, this is now passing "ant sqlcmdtest"  I think it's ready to merge.

wolffcm:
Looks good.  Thanks for doing this!

PR #3998, opened 2016-10-19
ENG-11286:BorrowTask, ultimately FragmentTask or SysprocFragmentTas, …

xinjiacs:
… is an read only task embedded in an MP transaction. Its response is a FragmentResponseMessage that should not ever be buffered on SpScheduler. The response used to not differentiate BorrowTask with FragmentTask/SysprocFragmentTask, so SpScheduler can not decide whether it should buffer the response or not. The commit addresses this issue by adding response bufferable flag into FragmentTask/SysprocFragmentTask and FragmentResponseMessage.

### DIFF
--- a/src/frontend/org/voltdb/messaging/FragmentResponseMessage.java
+++ b/src/frontend/org/voltdb/messaging/FragmentResponseMessage.java
@@ -50,6 +50,10 @@ public class FragmentResponseMessage extends VoltMessage {
     // Not currently used; leaving it in for now
     boolean m_dirty = true;
     boolean m_recovering = false;
+    // a flag to decide whether to buffer this response or not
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/messaging/FragmentResponseMessage.java
```diff
@@ -206,6 +217,7 @@ public void flattenToBuffer(ByteBuffer buf)
         buf.put(m_status);
         buf.put((byte) (m_dirty ? 1 : 0));
         buf.put((byte) (m_recovering ? 1 : 0));
+        buf.put((byte) (m_respBufferable ? 1 : 0));
```
nshi: As I mentioned in the original commit, this flag doesn't need to be sent over the network.

PR #4000, opened 2016-10-20
ENG-11055: Infrastructure for @Rebalance

zeemanhe:
Created infrastructure for new stored proc @Rebalance.

### DIFF
--- a/lib/python/voltcli/voltadmin.d/rebalance.py
+++ b/lib/python/voltcli/voltadmin.d/rebalance.py
@@ -1,0 +1,151 @@
+# This file is part of VoltDB.
+# Copyright (C) 2008-2016 VoltDB Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+
+class Host(dict):
+
+    def __init__(self, id, abort_func):
+        self.id = id
+        self.abort_func = abort_func
+
+    # Provide pseudo-attributes for dictionary items with error checking
+    def __getattr__(self, name):
+        try:
+            return self[name]
+        except IndexError:
+            self.abort_func('Attribute "%s" not present for host.' % name)
+
+    def get_admininterface(self):
+        """
+        Return the likely admininterface.
+        Implementation note: Currently, ipaddress will be set to externalinterface
+        if set by user, otherwise server selects any interface.
+        """
+
+        if 'admininterface' in self and self['admininterface']:
+            return self['admininterface']
+        if 'ipaddress' in self and self['ipaddress']:
+            return self['ipaddress']
+        return None
+
+class Hosts(object):
+
+    def __init__(self, abort_func):
+        self.hosts_by_id = {}
+        self.abort_func = abort_func
+
+    def update(self, host_id_raw, prop_name_raw, value_raw):
+        host_id = int(host_id_raw)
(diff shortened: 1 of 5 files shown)


lib/python/voltcli/voltadmin.d/rebalance.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@VOLT.Command(
+    bundles = VOLT.AdminBundle(),
+    options = (
+        VOLT.IntegerOption('-c', '--count', 'hostcount', 'number of hosts in the cluster'),
+        VOLT.StringOption('-t', '--stop', 'stoplist', 'target nodes to be stopped'),
+        VOLT.IntegerOption('-k', '--kfactor', 'kfactor', 'kfactor in cluster'),
+    ),
+    description = 'Rebalance the cluster.',
```
xinjiacs: Adding message like ", not read to use"

lib/python/voltcli/voltadmin.d/rebalance.py
```diff
@@ -0,0 +1,151 @@
...
+    hideverb = True
```
xinjiacs: Does this property control whether "rebalance" will be shown or not when "voltadmin -h" is called?

lib/python/voltcli/voltadmin.d/rebalance.py
```diff
@@ -0,0 +1,151 @@
...
+        for host in self.hosts_by_id.values():
+            if host.hostname == host_name:
+                return host
+        return None
+
+
+@VOLT.Command(
+    bundles = VOLT.AdminBundle(),
+    options = (
+        VOLT.IntegerOption('-c', '--count', 'hostcount', 'number of hosts in the cluster'),
```
xinjiacs: --hostcount other than '--count'?

"target number of hosts in the cluster"

lib/python/voltcli/voltadmin.d/rebalance.py
```diff
@@ -0,0 +1,151 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+
+class Host(dict):
```
xinjiacs: Can you refactor out these two classes? It has been used in STOP command as well.

src/frontend/org/voltdb/SystemProcedureCatalog.java
```diff
@@ -179,6 +179,7 @@ Procedure asCatalogProcedure() {
         builder.put("@ExplainProc",             new Config(null,                                           true,  true,  false, 0,    VoltType.INVALID,   false, false, true,  true,      false,    false));
         builder.put("@SendSentinel",            new Config(null,                                           true,  false, false, 0,    VoltType.INVALID,   true,  false, false, true,      false,    false));
         builder.put("@PrepareShutdown",         new Config("org.voltdb.sysprocs.PrepareShutdown",          false, false, false, 0,    VoltType.INVALID,   false, false, true,  true,      false,    true));
+        builder.put("@Rebalance",               new Config(null,                                           true,  false, false, 0,    VoltType.INVALID,   false, false, true,  true,      false,    false));
```
xinjiacs: Not sure if it's SP, but we can just leave it here util the REBALANCE is really used.

lib/python/voltcli/voltadmin.d/rebalance.py
zeemanhe: sure. will do

lib/python/voltcli/voltadmin.d/rebalance.py
zeemanhe: ok

lib/python/voltcli/voltadmin.d/rebalance.py
zeemanhe: correct. with hideverb = True, the command will not be list on voltadmin -h

lib/python/voltcli/voltadmin.d/rebalance.py
zeemanhe: will change

PR #4003, opened 2016-10-20
Correct error messages.

BillWhite:


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -739,15 +739,19 @@ protected TablePopulationRequirements checkAddDropIfTableIsEmptyWhitelist(final
...
                 viewName,
-                ((singleTableName == null)
-                        ? ""
-                        : String.format("on table %s ", singleTableName)));
+                (singleTable
+                        ? String.format("on table %s ", singleTableName)
+                        : ""),
+                (singleTable
+                        ? "the table already contains data"
+                        : "all of the tables already contain data"));
```
wolffcm: Your original message ended with a period.  The new one does not.  I'm not sure which one is correct.

PR #4004, opened 2016-10-20
Eliminate race conditions in TestCatalogUpdateSuite.

BillWhite:
There is a variable callbackSuccess which is initialized to true and is
set to false if an asynchronous stored procedure call returns an
unexpected result for any reason.  The test does not always wait until
the procedures are done before the variable is checked.  So, the
variable's value may not reflect the correct state of the procedure
execution. Also, the server may be taken down before the message for the
last stored procedure is executed.  If this happens, the last operations
fail and the variable is set to false.

This is all a race condition in the test and does not affect the product
functionality at all, except insofar as the product is not tested
completely.

https://issues.voltdb.com/browse/ENG-11313


wolffcm:
Looks good to me.

PR #4010, opened 2016-10-24
Avoid spurious valgrind output in memcheck builds

wolffcm:
Pro pull request:
https://github.com/VoltDB/pro/pull/1543

The first commit is Ethan's work on ENG-11228, which makes it so we shutdown properly in the IPC EE, which avoids apparent memory leaks (really just failing to clean up).

The remaining commits are my work to make it so that we can detect memory leaks even when they are not part of the JUnit process, and to disable running tests in memcheck builds when they need to crash the server.  We don't have a way of making the JUnit test fail when we expect a server crash.


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -1424,20 +1442,24 @@ public void waitForNodesToShutdown() {
...
-                proc.waitForShutdown();
+                valgrindOutputFile = proc.waitForShutdown();
             } catch (InterruptedException e) {
                 log.error("Unable to wait for EEProcess to die: " + proc.toString(), e);
             }
+
+            failIfValgrindErrors(valgrindOutputFile);
         }
 
         if (templateCmdLine.target() == BackendTarget.NATIVE_EE_VALGRIND_IPC) {
```
yzhang1991: Maybe it's OK to remove this part now.
Seems like it's doing the same thing as `failIfValgrindErrors()`?

tests/frontend/org/voltdb/regressionsuites/EEProcess.java
```diff
@@ -310,11 +302,43 @@ public void destroy() {
...
+            }
+        } catch (IOException e) {}
+    }
+
+    /**
+     * Shut down the EE process and return a file containing XML valgrind output.
+     * @return a File
+     * @throws InterruptedException
+     */
+    public File waitForShutdown() throws InterruptedException {
```
yzhang1991: Does this method have to return a `File`?
We can always get back to this by calling `EEProcess.m_valgrindErrors`, which is a static container.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -1819,4 +1841,31 @@ public void setOutputWatcher(OutputWatcher watcher) {
...
+    public static void failIfValgrindErrors(File valgrindOutputFile) {
+        if (valgrindOutputFile == null) {
+            return;
+        }
+
+        List<String> valgrindErrors = new ArrayList<>();
+        ValgrindXMLParser.processValgrindOutput(valgrindOutputFile, valgrindErrors);
+        if (!valgrindErrors.isEmpty()) {
+            String failString = "";
+            for (final String error : valgrindErrors) {
```
yzhang1991: Given the observation of the `ConcurrentModificationException` we had, I recommend to rewrite this part with iterators. See my examples above.

tests/frontend/org/voltdb/regressionsuites/TestZooKeeperPort.java
```diff
@@ -51,6 +50,10 @@ public TestZooKeeperPort(String name) {
      */
     @Override
     public void setUp() throws Exception {
+        if (LocalCluster.isMemcheckDefined()) {
+            return;
+        }
+
```
yzhang1991: Do you think we need to add `config.shutDown()` for this test just like the ones I did for even though we already disabled this test in memcheck mode?

wolffcm:
Trying a different approach following discussions with Bill and Ethan.

PR #4012, opened 2016-10-25
Fix ENG-10939 stored procedure has wrong parameter count

yzhang1991:
https://issues.voltdb.com/browse/ENG-10939


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -4198,6 +4199,10 @@ ExpressionOrException readSQLFunction(FunctionSQL function, boolean preferToThro
             }
 
             rewind(position);
+            // Discard parsed parameters along with the token rewinding.
+            for (int i=compileContext.parameters.size()-1; i>=prevParamCount; i--) {
```
paulmartel: We usually add 1 space before and after = and >= operators.

PR #4016, opened 2016-10-26
ENG-11322, integrate mixed sites per host support with the new CLI framework.

luyangco:
plus some other optimizations.

Change-Id: Icf5ea9efe3e20286f2406c2585c5f4e572ff5ad9


luyangco:
This is a correction for my squash and rebase mistake.

PR #4022, opened 2016-10-27
Unconditionally and deterministically force shutdown of LocalCluster instances in JUnit framework

wolffcm:
Pro pull request:
https://github.com/VoltDB/pro/pull/1553

The first commit is Ethan's, to ensure that we can shut down a running VoltDBIPC process even if the server doesn't start (i.e., due to a bogus command line).

The last commits change all of our tests that use LocalCluster that do not inherit from RegressionSuite to instead be a JUnit 4-style test that inherits from a new class called `JUnit4LocalClusterTest`.  This class has an `@AfterClass` annotation that will shut down any instances of LocalCluster that were created.  In memcheck builds, if we find any memory errors, an assertion will fire that causes the JUnit test to fail (and in Jenkins, a red orb will appear).

Previously, if the test writer forgot to shut down a cluster, memory errors would appear in the console log, even while all the tests appear to pass.


BillWhite:
This seems plausible to me.

PR #4024, opened 2016-10-28
ENG-11374 look for catalogs on auto and command log snaphots

vtkstef:


akhanzode:

PR #4026, opened 2016-10-28
Eng 11196 sqlcov div int

johnTcrawford:


src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -366,40 +469,46 @@ private boolean isColumnType(List<String> columnTypes, String columnName,
...
 
-    /** Returns the number of occurrences of the specified character in the specified String. */
-    static private int numOccurencesOfCharIn(String str, char ch) {
-        int num = 0;
-        for (int i = str.indexOf(ch); i >= 0 ; i = str.indexOf(ch, i+1)) {
-            num++;
+    /** Returns true if the <i>columnName</i> is either an integer constant or
+     *  an integer column (including types TINYINT, SMALLINT, INTEGER, BIGINT,
+     *  or equivalents in a comparison, non-VoltDB database); false otherwise. */
+    private boolean isIntegerConstant(String columnName) {
```
wolffcm: The comment and parameter name here must be incorrect?  Seems like this will return false if the argument contains any non-digits.

tests/scripts/sql_coverage_test.py
```diff
@@ -233,7 +233,7 @@ def get_max_mismatches(comparison_database, suite_name):
         # Failures in joined-matview-default-full due to ENG-11086
         elif config_name == 'joined-matview-default-full':
             max_mismatches = 3387
-        # Failures in joined-matview-int due to ENG-11196 & (mainly) ENG-11086
+        # Failures in joined-matview-int due to ENG-11086
```
wolffcm: No change to the number of allowed mismatches here?  I would have thought that it could be ratcheted down.

src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -431,6 +471,100 @@ static private boolean hasOddNumberOfSingleQuotes(String str) {
...
+                System.out.println("\nWARNING: in PostgreSQLBackend.handleParens, "
+                        + "numDivOperators is not 1: " + numDivOperators);
+            }
+            if (numDivOperators > 0) {
+                div_index = group.indexOf('/');
+                String subgroup = group.substring(0, div_index);
+                numOpenParens  = numOccurencesOfCharIn(subgroup, '(');
+                numCloseParens = numOccurencesOfCharIn(subgroup, ')');
+                if (numOpenParens > numCloseParens) {
+                    // Put the prefix after the last unmatched open parenthesis
```
wolffcm: At some point, maybe soon, you might find it much easier to transform a SQL string into an abstract syntax tree, which can be transformed much easier.  It seems to me that this code is reaching a point where it's becoming difficult to maintain.

src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -291,6 +330,41 @@ protected QueryTransformer debugPrint() {
             return this;
         }
 
+        // Used by the toString() method.
+        @SuppressWarnings("rawtypes")
```
johnTcrawford: Good idea. Fixed, using "(Collection<?>)".

src/frontend/org/voltdb/NonVoltDBBackend.java
johnTcrawford: Yes. Fixed.

tests/scripts/sql_coverage_test.py
johnTcrawford: Sorry, thought I already said this... Well, the vast majority of the failures are due to ENG-11086, not this bug; and the number of failures for that varies widely. I supposed I could reduce the current estimate of 46440 by 3 or 6 (the typical number of failures caused by this ticket, ENG-11196), but doing so seems fairly pointless.

PR #4033, opened 2016-11-01
ENG-11297. Create a new Pro edition kit

rmorgenstein:
If you want, this can be looked at as two separate commits.
1. Refactored build_kits to use argparse and added a few switches that make debugging easier, such as --nomac, --nopackages, --nocommunity.  It can also use the internal mirror
2. Added code to create the new Pro edition.  This is done by
* un-tgzing the newly built enterprise 
* making a new license 
* re-tgzing the kit as the pro kit
3. Changed the license length

I verified these kits with the following methods:
1. I compared the output of the "new" enterprise build from this script (I rebuilt voltdb-6.7) to the real release.  The only difference was the dates.
2. I compared the output of the new Enterprise release to the new Pro release. The only difference was the license file.

### DIFF
--- a/tools/kit_tools/build_kits.py
+++ b/tools/kit_tools/build_kits.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-import argparse, os, sys, shutil, datetime
+import argparse, datetime, os, sys, shutil, traceback
 from fabric.api import run, cd, local, get, settings, lcd, put
 from fabric_ssh_config import getSSHInfoForHost
 from fabric.context_managers import shell_env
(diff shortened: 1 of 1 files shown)


tools/kit_tools/build_kits.py
```diff
@@ -78,18 +80,36 @@ def buildCommunity():
...
 ################################################
 
-def buildPro():
+def buildEnterprise():
     with cd(builddir + "/pro"):
         run("pwd")
         run("git status")
         run("git describe --dirty")
-        run("VOLTCORE=../voltdb ant -f mmt.xml -Djmemcheck=NO_MEMCHECK -Dallowreplication=true -DallowDrActiveActive=true -Dlicensedays=%d -Dkitbuild=true %s clean dist.pro" % (defaultlicensedays, build_args))
+        run("VOLTCORE=../voltdb ant -f mmt.xml -Djmemcheck=NO_MEMCHECK -Dallowreplication=true -DallowDrActiveActive=true -Dlicensedays=%d -Dkitbuild=%s %s clean dist.pro" % (defaultlicensedays, "true" if build_mac else "false", build_args))
```
jhugg: I'm warming up to Paul's obsession with changing printf-style post-lists vars to string concatenation with vars inline. Easier to see what's what. 

Not sure this is worth changing though. It could be a bit cleaner not to put if logic in the string itself.

tools/kit_tools/build_kits.py
```diff
@@ -114,10 +134,10 @@ def buildRabbitMQExport(version):
 # MAKE AN ENTERPRISE TRIAL LICENSE
 ################################################
 
-# Must be called after buildPro has been done
-def makeTrialLicense(days=30):
+# Must be called after buildEnterprise has been done
+def makeTrialLicense(days=30, dr_and_xdcr="true", nodes=12):
```
jhugg: What's with the 12? Was the old one 12?

tools/kit_tools/build_kits.py
```diff
@@ -324,55 +355,62 @@ def rmNativeLibs():
...
+    # build debian kit
+    try:
+        with settings(user=username,host_string=UbuntuSSHInfo[1],disable_known_hosts=True,key_filename=UbuntuSSHInfo[0]):
+            debbuilddir = "%s/deb_build/" % builddir
+            run("rm -rf " + debbuilddir)
+            run("mkdir -p " + debbuilddir)
+
+            with cd(debbuilddir):
+                put ("tools/voltdb-install.py",".")
+
```
jhugg: Double whitespace?

jhugg:
I guess this looks fine. Some of the if-statements inside string formatting made me read a line more than once, but it's not like this is all that complex overall. I didn't see anything whacky.

tools/kit_tools/build_kits.py
rmorgenstein: Yes. The old one was 12.

tools/kit_tools/build_kits.py
rmorgenstein: Uck.  What was I thinking.  Will fix.

PR #4036, opened 2016-11-03
ENG-11392 let DR consumer complete transactions before shutting down

zeemanhe:
In graceful shutdown, check DR consumer stats. Let consumer apply all transactions before shutdown is issued.

### DIFF
--- a/lib/python/voltcli/checkstats.py
+++ b/lib/python/voltcli/checkstats.py
@@ -208,3 +208,17 @@ def check_importer(runner):
         if outstanding == 0:
             return
         time.sleep(1)
+
+def check_dr_consumer(runner):
+     while True:
+        resp = get_stats(runner, 'DRCONSUMER')
+        outstanding = 0
+        if len(resp.table(1).tuples()) == 0:
+            return
+        for r in resp.table(1).tuples():
+            if r[7] <> r[8]:
+                outstanding += 1
+                runner.info('DR Consumer-partition %d on host %d has outstanding transactions to be applied.' %(r[4], r[1]))
+        if outstanding == 0:
+            return
+        time.sleep(2)
--- a/lib/python/voltcli/checkstats.py
+++ b/lib/python/voltcli/checkstats.py
@@ -218,7 +218,7 @@ def check_dr_consumer(runner):
         for r in resp.table(1).tuples():
             if r[7] <> r[8]:
                 outstanding += 1
-                runner.info('DR Consumer-partition %d on host %d has outstanding transactions to be applied.' %(r[4], r[1]))
+                runner.info('Partition %d on host %d has outstanding DR consumer transactions.' %(r[4], r[1]))
         if outstanding == 0:
             return
         time.sleep(2)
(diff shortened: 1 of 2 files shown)


lib/python/voltcli/checkstats.py
```diff
@@ -208,3 +208,17 @@ def check_importer(runner):
...
         time.sleep(1)
+
+def check_dr_consumer(runner):
+     while True:
+        resp = get_stats(runner, 'DRCONSUMER')
+        outstanding = 0
+        if len(resp.table(1).tuples()) == 0:
+            return
+        for r in resp.table(1).tuples():
+            if r[7] <> r[8]:
```
xinjiacs: Can you add comments here, why we need to check these two columns to determine DRConsumer is drained?

lib/python/voltcli/checkstats.py
```diff
@@ -208,3 +208,17 @@ def check_importer(runner):
...
+                outstanding += 1
+                runner.info('Partition %d on host %d has outstanding DR consumer transactions.' %(r[4], r[1]))
+        if outstanding == 0:
+            return
+        time.sleep(2)
```
xinjiacs: Why sleep(2) here and other places sleep(1)?

lib/python/voltcli/voltadmin.d/shutdown.py
```diff
@@ -44,12 +44,14 @@ def shutdown(runner):
             status = runner.call_proc('@Quiesce', [], []).table(0).tuple(0).column_integer(0)
             if status <> 0:
                 runner.abort('The cluster has failed to be quiesce with status: %d' % status)
-            runner.info('Completing outstanding export and DR transactions...')
+            runner.info('Completing outstanding export and DR producer transactions...')
```
xinjiacs: Maybe it's better to separate the exporter and dr producer check?
Let's discuss with Anish/Ning about the checking order here. I am not sure whether they matter or not,

lib/python/voltcli/voltadmin.d/shutdown.py
```diff
@@ -44,12 +44,14 @@ def shutdown(runner):
...
             checkstats.check_export_dr(runner)
             runner.info('Completing outstanding client transactions.')
             checkstats.check_clients(runner)
             runner.info('Completing outstanding importer requests.')
             checkstats.check_importer(runner)
+            runner.info('Completing outstanding DR consumer transactions...')
+            checkstats.check_dr_consumer(runner)
```
xinjiacs: Also, does the --pause need to take care of these stats checking? It already has exporter and DR producer checked.

xinjiacs: How do you test it? Run the failed tests? or Manually?

Probably we can commit this and run the failed tests over and over on master.

zeemanhe: I have manually tested it to make sure the consumer stats are clear before shutdown is issued.  We can also do system testing on the branch.

We can certainly separate export and dr producer stats.

lib/python/voltcli/checkstats.py
zeemanhe: comments added

PR #4040, opened 2016-11-06
Avoid excessive array allocation in InMemoryJarfile

wolffcm:
A customer noted long GC pauses when doing LOAD CLASSES on a largish jar with hundreds of classes.  We were doing a lot of allocating (per class in the jar) of 4MB and 8MB byte buffers.  This changeset addresses that.

The existing unit test, TestInMemoryJarfile, cover this change.

### DIFF
--- a/src/frontend/org/voltdb/utils/InMemoryJarfile.java
+++ b/src/frontend/org/voltdb/utils/InMemoryJarfile.java
@@ -88,41 +88,70 @@ public class InMemoryJarfile extends TreeMap<String, byte[]> {
         loadFromStream(new ByteArrayInputStream(bytes));
     }
 
-    private void loadFromStream(InputStream in) throws IOException {
-        JarInputStream jarIn = new JarInputStream(in);
-        JarEntry catEntry = null;
-        while ((catEntry = jarIn.getNextJarEntry()) != null) {
-            byte[] value = readFromJarEntry(jarIn, catEntry);
-            String key = catEntry.getName();
-            put(key, value);
+    // A best guess at an upper bound for the size of entries in jar files, since
+    // we do not know the uncompressed size of the entries ahead of time.
+    // This number is used to size the buffers used to read from jar streams.
+    private static final int JAR_ENTRY_SIZE_GUESS = 1024 * 1024; // 1MB
+
+    // A class that can be used to read from a JarInputStream where the size of entries
+    // is not known.  Can be reused multiple times to avoid excessive memory allocations.
+    private static class JarInputStreamReader {
+
+        private byte[] m_array0;
+        private byte[] m_array1 = null;
+
+        JarInputStreamReader(int initSize) {
+            m_array0 = new byte[initSize];
+        }
+
+        byte[] readEntryFromStream(JarInputStream jarIn) throws IOException {
+            int totalRead = 0;
+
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -88,41 +88,65 @@ public InMemoryJarfile(byte[] bytes) throws IOException {
...
-        JarInputStream jarIn = new JarInputStream(in);
-        JarEntry catEntry = null;
-        while ((catEntry = jarIn.getNextJarEntry()) != null) {
-            byte[] value = readFromJarEntry(jarIn, catEntry);
-            String key = catEntry.getName();
-            put(key, value);
+    // A best guess at an upper bound for the size of entries in jar files, since
+    // we do not know the uncompressed size of the entries ahead of time.
+    // This number is used to size the buffers used to read from jar streams.
+    private static final int JAR_ENTRY_SIZE_GUESS = 1024 * 1024; // 1MB
```
paulmartel: Minor point: it'd be slightly simpler to encapsulate this constant inside JarInputStreamReader and reference it only in the initializer for m_array vs. in the two constructor callers.

src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -88,41 +88,65 @@ public InMemoryJarfile(byte[] bytes) throws IOException {
...
-            int readSize = jarIn.read(buffer, 0, buffer.length);
-            if (readSize > 0) {
+        byte[] readEntryFromStream(JarInputStream jarIn) throws IOException {
+            int totalRead = 0;
+
+            while (jarIn.available() == 1) {
+                int bytesToRead = m_array.length - totalRead;
+                assert (bytesToRead > 0);
+                int readSize = jarIn.read(m_array, totalRead, bytesToRead);
+                if (readSize < 0) {
```
paulmartel: Use `<=` ? I don't know if `readSize == 0` is actually a possibility, but you've changed (complicated) its handling vs. the original code.

src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -88,41 +88,65 @@ public InMemoryJarfile(byte[] bytes) throws IOException {
...
-                    System.arraycopy(bytes, 0, temp, 0, bytes.length);
-                    bytes = temp;
+                // If we have filled up our buffer and there is still more to
+                // read...
+                if (readSize == bytesToRead && (jarIn.available() == 1)) {
+                        // Make a new array double the size, and copy what we've read so far
+                        // in there.
+                        byte[] tmpArray = new byte[2 * m_array.length];
+                        System.arraycopy(m_array, 0, tmpArray, 0, totalRead);
+                        m_array = tmpArray;
```
paulmartel: Over-indented block here.  End of review round.

src/frontend/org/voltdb/utils/InMemoryJarfile.java
wolffcm: The Java API is confusing because there are two signals that we are done reading the current entry, returning -1 from `read`, and `available` returning 0.  In the original code, we relied on the latter, so I'll change it back.

PR #4042, opened 2016-11-07
ENG-11432 check various stats for shutdown and pause

zeemanhe:
graceful shutdown: 
wait for in-flight txns, importer, dr consumer, async cmdlog txns fsynced.

shutdown --save and voltadmin pause --wait: 
wait for in-flight txns, importer, dr consumer, async cmdlog txns fsynced, exporter, and dr producer.

### DIFF
--- a/lib/python/voltcli/voltadmin.d/pause.py
+++ b/lib/python/voltcli/voltadmin.d/pause.py
@@ -27,7 +27,7 @@ from voltcli import checkstats
     )
 )
 def pause(runner):
-   # Check the STATUS column. runner.call_proc() detects and aborts on errors.
+    #Check the STATUS column. runner.call_proc() detects and aborts on errors.
     status = runner.call_proc('@Pause', [], []).table(0).tuple(0).column_integer(0)
     if status <> 0:
         runner.error('The cluster has failed to pause with status: %d' % status)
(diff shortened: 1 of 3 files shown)


lib/python/voltcli/voltadmin.d/shutdown.py
```diff
@@ -44,18 +44,19 @@ def shutdown(runner):
...
             if runner.opts.save:
                columns = [VOLT.FastSerializer.VOLTTYPE_BIGINT]
                shutdown_params =  [zk_pause_txnid]
+               #save option, check more stats
+               checkstats.check_exporter(runner)
+               checkstats.check_dr_producer(runner)
         except (KeyboardInterrupt, SystemExit):
             runner.info('The cluster shutdown process has stopped. The cluster is still in a paused state.')
             runner.abort('You may shutdown the cluster with the "voltadmin shutdown --force" command, or continue to wait with "voltadmin shutdown".')
+    runner.info('Cluster is ready for shutdown')
```
xinjiacs: This message is only printed when doing non-force shutdown before. So you should move it inside the IF case.

Also the message should be something: "Cluster is safe to shutdown, everything has been drained."

lib/python/voltcli/voltadmin.d/pause.py
```diff
@@ -27,7 +27,7 @@
     )
 )
 def pause(runner):
-    # Check the STATUS column. runner.call_proc() detects and aborts on errors.
+   # Check the STATUS column. runner.call_proc() detects and aborts on errors.
```
xinjiacs: on purpose?

lib/python/voltcli/voltadmin.d/pause.py
```diff
@@ -39,5 +39,9 @@ def pause(runner):
             runner.error('The cluster has failed to quiesce with status: %d' % status)
             return
         runner.info('The cluster is quiesced.')
-        #check the dr stats
-        checkstats.check_export_dr(runner)
+        checkstats.check_clients(runner)
```
xinjiacs: Can you do similar catch for CTRL-C like shutdown?

zeemanhe: ctrl-c handling was added

lib/python/voltcli/voltadmin.d/shutdown.py
zeemanhe: updated.

PR #4044, opened 2016-11-08
Eng 11264

manjujames:


manjujames:
New control response message with structured message types for different response types.

src/protobuf/dragent.proto
```diff
@@ -69,6 +69,46 @@ message Response {
...
+message ConnectResponse {
+    optional fixed64 catalogCRC = 1;
+    optional string catalogSignature = 2;
+    optional int32 protocolVersion = 3;
+    repeated ClusterInfo clusterInfo = 4;
+}
+
+message ClusterInfo {
+    required int32 clusterId = 1;
+    required fixed64 clusterPersistedId = 2;
```
nshi: Is this the creation instance ID?

manjujames: Yes. I am fine with using whatever the agreed terminology is. Let me know.

nshi: `clusterCreationId` is probably okay.

PR #4045, opened 2016-11-08
analytic-template.sql: added COUNT(*) and COUNT(_columnorexpr) to the

johnTcrawford:
list of analytic functions to be tested; and modified some (though not all) of the query templates to have optional ORDER BY clauses, since the old analytic functions (RANK, DENSE_RANK) always require an ORDER BY clause, but this new one does not. (Also tweaked the definition of _analyticfunc to include the parentheses, to support this; plus improved comments and spacing.)

### DIFF
--- a/tests/scripts/examples/sql_coverage/analytic-template.sql
+++ b/tests/scripts/examples/sql_coverage/analytic-template.sql
@@ -18,32 +18,38 @@
 DELETE FROM @dmltable
 INSERT INTO @dmltable VALUES (@insertvals)
 
--- Define expressions and clauses used to test analytic functions, such as RANK, below
+-- Define expressions and clauses used to test windowed analytic functions (such as RANK)
 {_columnorexpr |= "_variable"}
 {_columnorexpr |= "@onefun(_variable[@comparabletype])@plus10"}
 {_orderbycolumnorexpr |= "_variable[#ord @rankorderbytype]"}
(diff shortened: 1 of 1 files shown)


BillWhite:
I question whether it's right to follow Postgresql and not the SQL 2003 standard in this case.
RANK and DENSE_RANK require ORDER BY, at least by my reading of the standard.  CUME_DIST and PERCENT_RANK do as well, though we don't support these yet, and may never support them.  None of the other window functions we are likely to support soon, max, min, sum, count, and avg, require order by, and none of the standard ones we are likely to support require ORDER BY.  But these four do.  I am not clear about ROW_NUMBER, which we might want to support soonish.

johnTcrawford:
Bill, can you please clarify for me: RANK, DENSE_RANK, CUME_DIST & PERCENT_RANK (& maybe ROW_NUMBER) all require ORDER BY (right?); but do the others (COUNT, MIN, MAX, SUM, AVG) disallow ORDER BY, or is it optional (according to the SQL 2003 standard)?

johnTcrawford:
Bill's (emailed) answer to me was that RANK and DENSE_RANK are the only ones that require ORDER BY; all the others (CUME_DIST, PERCENT_RANK, ROW_NUMBER, COUNT, MIN, MAX, SUM, AVG) should (& will, once implemented) work with or without ORDER BY. So I'm going to leave the tests as they are, at least for now.

PR #4047, opened 2016-11-08
ENG-10316: Improve logging for consolidated Digest generation

wweiss-voltdb:


src/frontend/org/voltdb/ExtensibleSnapshotDigestData.java
```diff
@@ -205,7 +205,19 @@ private void mergeDRTupleStreamInfoToZK(JSONObject jsonObj) throws JSONException
                 partitionStateInfo = e.getValue().replicatedInfo;
             }
             JSONObject existingStateInfo = stateInfoMap.optJSONObject(partitionId);
-            if (existingStateInfo == null || partitionStateInfo.drId > existingStateInfo.getLong("sequenceNumber")) {
+            boolean addEntry = false;
+            if (existingStateInfo == null) {
+                addEntry = true;
+            }
+            else if (partitionStateInfo.drId > existingStateInfo.getLong("sequenceNumber")) {
```
nshi: What if it's smaller than? Do we want to know that, too?

PR #4048, opened 2016-11-08
ENG-11149 explainview vmc

yzhang1991:
This pull request addressed the following issues:
* Integrated Andrew's comments about the presentation of the table.
* Refactored the code for generating the `ExplainView` table. A new class `ViewExplainer` was created so that both `ReportMaker` and `ExplainView` system procedure and share the same code for table generation.
* Added a ReportMaker test case to verify the HTML table generated in VMC is correct.

### DIFF
--- a/src/frontend/org/voltdb/compilereport/ViewExplainer.java
+++ b/src/frontend/org/voltdb/compilereport/ViewExplainer.java
@@ -35,7 +35,8 @@ import org.voltdb.utils.Encoder;
 public class ViewExplainer {
 
     private static String getIndexNameUsedInStatement(Statement stmt) {
-        // We should get at most one index from the statement.
+        // Get the name of the index used in a statement (query plan)
+        // that is used for refreshing a min/max column in a single-table view.
         String[] indicesUsedInStatement = stmt.getIndexesused().split(",");
         assert(indicesUsedInStatement.length <= 1);
         for (String tableDotIndexPair : indicesUsedInStatement) {
--- a/src/frontend/org/voltdb/compilereport/ViewExplainer.java
+++ b/src/frontend/org/voltdb/compilereport/ViewExplainer.java
@@ -1,0 +1,122 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2016 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/compilereport/ViewExplainer.java
```diff
@@ -0,0 +1,122 @@
...
+import org.voltdb.catalog.Table;
+import org.voltdb.types.ExpressionType;
+import org.voltdb.utils.CatalogUtil;
+import org.voltdb.utils.Encoder;
+
+// Generate a report that can explain to the users how we maintain a view.
+public class ViewExplainer {
+
+    private static String getIndexNameUsedInStatement(Statement stmt) {
+        // We should get at most one index from the statement.
```
wolffcm: For views with joins and/or where clauses, isn't it possible to have two indexes?

wolffcm: Oh, I think this is only for MIN/MAX refresh, not for initial materialization... still it seems possible to have more than one index used?

tests/frontend/org/voltdb/compilereport/TestReportMaker.java
```diff
@@ -244,6 +244,32 @@ public void testTableWithoutIndexGetFullTableScanWarning() throws IOException {
         assertTrue(report.contains("Table " + tableName + " doesn't have any unique index, it will cause full table scans to update/delete DR record and may become slower as table grow."));
     }
 
+    public void testExplainViewTable() throws IOException {
```
wolffcm: This seems like quite sufficient testing to me, for the changes you're making here.

src/frontend/org/voltdb/compilereport/ViewExplainer.java
yzhang1991: This `getIndexNameUsedInStatement()` function is only used to extract the name of the index that was used in a statement in single-table view cases. For views with joins, we always directly output the plan explanation text.
The queries we used to refresh a MIN/MAX column in single-table view cases are always like this:
```sql
SELECT MIN(col) FROM T WHERE g1 = ? AND g2 = ? AND col >= ?;
```
I cannot think of a case where such queries can reference more than one index for evaluation. Do you think I missed anything?

wolffcm: If it's only for single-table view cases, I think it's fine.  You can only use one index per table instance in a statement.  Thanks for clarifying.  Maybe update the comment?

PR #4055, opened 2016-11-10
ENG-10294 part 1 EE supports SWAP TABLES

paulmartel:
Also tweaked truncate table to simplify its undo action and to share (renamed) its underlying mechanism for dealing with streaming in progress against past state of the table(s) (such as snapshotting).

Error handling is still a little bit stubbed with some asserts where there should be runtime errors.

Handling of SWAP TABLES when the tables have different index or view definitions is guarded against and stubbed for now -- not sure how soon we'll get around to phasing in at least edge cases like that.

### DIFF
--- a/src/ee/executors/swaptablesexecutor.cpp
+++ b/src/ee/executors/swaptablesexecutor.cpp
@@ -1,0 +1,109 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2016 VoltDB Inc.
+ *
+ * This file contains original code and/or modifications of original code.
+ * Any modifications made by VoltDB Inc. are licensed under the following
+ * terms and conditions:
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/* Copyright (C) 2008 by H-Store Project
+ * Brown University
+ * Massachusetts Institute of Technology
+ * Yale University
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT
+ * IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include "swaptablesexecutor.h"
+
+#include "plannodes/swaptablesnode.h"
+//#include "storage/temptable.h"
+#include "storage/persistenttable.h"
+
+using namespace std;
+using namespace voltdb;
+
+bool SwapTablesExecutor::p_init(AbstractPlanNode *abstract_node, TempTableLimits* limits)
+{
+    VOLT_TRACE("init SwapTable Executor");
+#ifndef NDEBUG
+    SwapTablesPlanNode* node = dynamic_cast<SwapTablesPlanNode*>(m_abstractNode);
+    assert(node);
+    assert(node->getTargetTable1());
+    assert(node->getTargetTable2());
+    assert(node->getInputTableCount() == 0);
+#endif
+
+    setDMLCountOutputTable(limits);
+    return true;
+}
+
+bool SwapTablesExecutor::p_execute(const NValueArray &params) {
+    // target tables should be persistent tables
+    // update target table references from table delegates
+    SwapTablesPlanNode* node = static_cast<SwapTablesPlanNode*>(m_abstractNode);
+    assert(dynamic_cast<SwapTablesPlanNode*>(m_abstractNode));
+    PersistentTable* targetTable1 = node->getTargetTable1();
+    assert(targetTable1);
+    PersistentTable* targetTable2 = node->getTargetTable2();
+    assert(targetTable2);
+
+    int64_t modified_tuples = 0;
+
+    VOLT_TRACE("swap tables %s and %s",
+               targetTable1->name().c_str(),
+               targetTable2->name().c_str());
+    // count the active tuples in both tables as modified
+    modified_tuples = targetTable1->visibleTupleCount() +
+            targetTable2->visibleTupleCount();
+
+    VOLT_TRACE("Swap Tables: %s with %d active, %d visible, %d allocated"
+               " and %s with %d active, %d visible, %d allocated",
+               targetTable1->name().c_str(),
+               (int)targetTable1->activeTupleCount(),
+               (int)targetTable1->visibleTupleCount(),
+               (int)targetTable1->allocatedTupleCount(),
+               targetTable2->name().c_str(),
+               (int)targetTable2->activeTupleCount(),
+               (int)targetTable2->visibleTupleCount(),
+               (int)targetTable2->allocatedTupleCount());
+
+    // Swap the table catalog delegates and corresponding indexes and views.
+    targetTable1->swapTable(targetTable2, m_engine);
+
+    TableTuple& count_tuple = m_tmpOutputTable->tempTuple();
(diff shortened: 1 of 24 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -545,6 +548,461 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
     }
 }
 
+struct Swappability {
+    vector<TableIndex*>* m_theIndexesToSwap;
+    vector<TableIndex*>* m_otherIndexesToSwap;
+    vector<TableIndex*>* m_theIndexesToRepopulate;
+    vector<TableIndex*>* m_otherIndexesToRepopulate;
+    vector<MaterializedViewTriggerForWrite*>* m_theScanViewsToSwap;
+    vector<MaterializedViewTriggerForWrite*>* m_otherScanViewsToSwap;
```
wolffcm: `other` makes more sense inside `PersistentTable`.

src/ee/storage/persistenttable.cpp
```diff
@@ -545,6 +548,461 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+                m_theIndexesToSwap->push_back(theIndex);
+                m_otherIndexesToSwap->push_back(otherIndex);
+                continue;
+            }
+
+            bool unmatched;
+
+            unmatched = true;
+            auto otherBacklogCount = m_otherIndexesToRepopulate->size();
+            for (int jj = 0; jj < otherBacklogCount; ++jj) {
```
wolffcm: So this loop will be O(n^2) in the number of indexes on the table?  I guess we don't expect to have that many indexes.

I was going to ask why we can't do a lot of this determination in the planner.  But I guess it's because we don't have a numeric ID for our indexes.  Maybe we should have numeric index IDs?

src/ee/storage/persistenttable.cpp
```diff
@@ -545,6 +548,461 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+                    break;
+                }
+            }
+            if (unmatched) {
+                m_otherIndexesToRepopulate->push_back(otherIndex);
+            }
+        }
+    }
+
+    void analyzeScanViews(PersistentTable* theTable, PersistentTable* otherTable) {
```
wolffcm: If we had numeric view IDs, would it just be possible to define a mapping from this table's views to the other table's views in the planner, and just swap pointers to the storage for the materializations?  You've thought about this more than me, so I'm probably missing something obvious.

src/ee/storage/persistenttable.cpp
```diff
@@ -545,6 +548,461 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+        ScanViewVector& theScanViewsToRepopulate,
+        ScanViewVector& otherScanViewsToRepopulate,
+        JoinViewVector& joinViewsToSwap,
+        JoinViewVector& otherJoinViewsToSwap,
+        JoinViewVector& joinViewsToRepopulate,
+        JoinViewVector& otherJoinViewsToRepopulate) {
+    auto tcd1 = engine->getTableDelegate(m_name);
+    if (tcd1->getTable() == this) {
+        cout << "Found this " << endl;
+    tcd1->setTable(otherTable);
```
wolffcm: This line should be indented, plus a few others below.

src/ee/storage/persistenttable.cpp
```diff
@@ -545,6 +548,461 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+    if (namedTable != this) {
+        cout << "Other table mismatch: " << (void*)namedTable << " vs " << (void*)this << " vs " << (void*)otherTable << endl;
+    }
+
+    // Swap the table attributes that must continue to be associated with each
+    // table's name/identity, not its swapped content.
+    // We MIGHT want to consider making these attributes of TableCatalogDelegate
+    // instead of PersistentTable?
+
+    auto name1 = m_name;
```
wolffcm: I was thinking that you could leave all the attributes of the `PersistentTable` as-is, save for the pointer(s) to the table blocks?  Is that possible?

src/ee/storage/MaterializedViewHandler.cpp
```diff
@@ -412,4 +415,34 @@ namespace voltdb {
...
+    // didacted. If needed for catalog performance, this information would
+    // be easy to encode into a compressed string and decode upon EE
+    // deserialization of the view definition.
+    // For one thing, doing the work in DDLCompiler would save runtime here.
+    // For another, the pattern matching tools in java are more familiar
+    // and possibly just better.
+    std::string MaterializedViewHandler::getSwappableSQL(
+            PersistentTable* sourceTable) const {
+        std::string const& tableName = sourceTable->name();
+        return MaterializedViewUtil::sourceTableDidact(m_createQuerySQL, tableName);
```
BillWhite: Is this required because table names are embedded in the MV plan fragments for updating aggregate columns and populating the tables?  We have to swap table names in the plans, I think.  But can't we just swap the plans themselves, or is that too hard?

How different are the two plans for the swapped tables?  Do the swapped tables have to have the same materialized views?  is that the source of this?

paulmartel: Yes, the swapped tables critically need identical views. I apologize for including some vapor-ware stubs that hinted otherwise.  Cached plan fragments contain executors that may reference (DML) target tables and (in general) source tables and source indexes all by name. As long as we correctly rename all the tables and indexes, all plan fragments remain valid. The continued validity of all existing plan fragments is crucial to the implementation of SWAP TABLES as a low-cost DML statement vs. a high-impact DDL statement.

The need for didacted query text is in the analysis of which views are in fact "corresponding pairs, differing only in a reference to the target table of a swap in one view where the other view has a reference to the other table in the swap. We determine this by redacting (oops, not "didacting" -- pardon the vocabulary fail) the differing table names from the differing queries and then testing for exact match.

src/ee/storage/persistenttable.cpp
```diff
@@ -90,6 +91,8 @@
 #include <cstdio>
 #include <sstream>
 
+using namespace std;
```
paulmartel: Will fix.

src/ee/executors/swaptablesexecutor.cpp
```diff
@@ -0,0 +1,109 @@
...
+    setDMLCountOutputTable(limits);
+    return true;
+}
+
+bool SwapTablesExecutor::p_execute(const NValueArray &params) {
+    // target tables should be persistent tables
+    // update target table references from table delegates
+    SwapTablesPlanNode* node = static_cast<SwapTablesPlanNode*>(m_abstractNode);
+    assert(dynamic_cast<SwapTablesPlanNode*>(m_abstractNode));
+    PersistentTable* targetTable1 = node->getTargetTable1();
```
paulmartel: The short answer is "yes I could, if I cared about debug build micro-optimization".

The long answer is that after reviewing all the different ways we handle this down-cast in the various executors, I purposely settled on this one for the following reasons:

    Use of m_abstractNode rather than the identical value passed as a p_init argument allows a greater degree of consistency between p_init and p_exec which has no such argument. Eventually, the argument to p_init can be eliminated.

    Use of static_cast in the normal code path  has 0 run time cost unlike dynamic_cast.

    Use of dynamic_cast in an assert where we don't mind the debug build runtime cost catches two unlikely but tragic logic errors that would likely crash the static_cast code path -- a null plan node pointer OR the wrong plan node type. With the debug build suitably guarding against these cases, the release build is freed to run unencumbered without fear of crashes.

And finally, addressing your specific question, it may be possible that other p_init methods have no need for a local node variable, so the assert is written to apply generally where there isn't one. Also, I am not sure whether the compiler is obligated to run a dynamic_cast when the DECLARED type of the input is already the target type of the cast -- though I'd guess it probably is obligated. Still, I could see a dumb compiler or half-clever engineer mistakenly discarding the cast as redundant because of that. The dynamic cast from the generically typed m_abstractNode benefits from LOOKING like it is doing what it is actually doing and would be doing anyway to node.

src/ee/storage/MaterializedViewUtil.h
```diff
@@ -0,0 +1,64 @@
...
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MATERIALIZEDVIEWUTIL_H
+#define MATERIALIZEDVIEWUTIL_H
+
+namespace voltdb {
+
+struct MaterializedViewUtil {
+    static std::string sourceTableDidact(std::string const& statementSQL,
```
paulmartel: Sorry for the vocab fail. Meant "redact". Fixing this.

Point taken (somewhat reluctantly) on the dangers of `using namespace std;`
A shame that we do this in > 100 modules.

As for plan fragments, yes they do reference their tables and indexes by name and typically do a name lookup at the top of p_execute -- as I recall, this was forced by the implementation of Truncate Table.

src/ee/storage/persistenttable.cpp
paulmartel: I don't get it. Numeric ID based on what? 
Anyway, can we optimize this later, as needed? I don't consider O(n^2) a non-starter when n is the number of indexes on a table.
Besides, I am guessing that the O(n^2) worst case here will often if not always degenerate into the O(n) best case because the indexes happen to be cataloged in the same order. If needed, we could augment the catalog to contain an index signature string that encapsulated all its details EXCEPT its table -- then the EE PersistentTable could store a signature to index map. Parallel iteration over these maps would give an O(n) match.

src/ee/storage/persistenttable.cpp
paulmartel: I don't see any preferable alternative to holding over from PersistentTable the parallel naming of theXXX and otherXXX, extending it triviallly to use theTable in place of PersistentTable's this.

src/ee/storage/persistenttable.cpp
paulmartel: Again, I am confused. ID based on what? Either the IDs must be identical between two objects that are supposed to correspond in a swap -- though we don't even know a priori which pairs of tables we will eventually try to swap -- OR an ID-to-ID mapping must be determined by the planner or EE -- and there IS an obscure reason to make the EE capable of doing this. Identical IDs seems like it would be hard to implement in the DDL compiler, and ID-to-ID mapping seems little better than the name-to-name mapping I am doing already. We can argue whether the SWAP TABLE planner should (ALSO?) be able to generate mappings, but I don't see use of IDs vs. names changing that question very much.

src/ee/storage/persistenttable.cpp
paulmartel: I've had the thought but eventually became convinced that it wasn't that easy. Maybe that was a failure of imagination -- maybe there were issues with that, but I just missed easy workarounds for them. I'll try to reconstruct my thought process so we can discuss it.

PR #4063, opened 2016-11-14
ENG-11263: DR query response format change.

nshi:
Removed unused fields and added support for multiple clusters.

**PLEASE DON'T DELETE THE BRANCH AFTER MERGE.**

### DIFF
--- a/src/protobuf/dragent.proto
+++ b/src/protobuf/dragent.proto
@@ -40,6 +40,11 @@ message Subscribe {
     optional bool isSyncing = 5;
 }
 
+message Query {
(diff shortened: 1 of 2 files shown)


src/protobuf/dragent.proto
```diff
@@ -40,6 +40,11 @@ message Subscribe {
     optional bool isSyncing = 5;
 }
 
+message Query {
```
manjujames: This would be an optional field in CtrlEnvelope?

nshi: Good catch. I fixed it in the commit.

PR #4064, opened 2016-11-14
ENG-11471: skip pip install if the module already installed

JadenW:
It seems venv_start always try to install required packages. We should skip that if already system wide installed (by root).


akhanzode:
Can we install all modules for --user so that we dont have to worry about system modules?

PR #4066, opened 2016-11-15
Eng 11396 window refactor

BillWhite:
Refactor the Window function code.

The code for window functions was completely separated from Aggregate functions.  New plans and new executors are defined.  The plans. in WindowFunctionPlanNode, are pretty much identical with the previous ones, though they don't depend on AggregatePlanNode.  The C++ executor class, WindowFunctionExecutor is similar to but simpler than the serial aggregation executor.

The C++ and jUnit tests are all blue in Jenkins.


src/ee/plannodes/abstractplannode.cpp
```diff
@@ -393,9 +393,11 @@ AbstractPlanNode::TableOwner::~TableOwner() { delete m_tempTable; }
...
 AbstractPlanNode::OwningExpressionVector::~OwningExpressionVector()
 {
-    size_t each = size();
-    while (each--) {
-        delete (*this)[each];
+    for (size_t each = 0; each < size(); each += 1) {
+        AbstractExpression *expr = (*this)[each];
+        if (expr) {
+            delete expr;
+        }
```
BillWhite: The existing code didn't work for empty vectors.

tests/ee/executors/PartitionByExecutorTest.cpp
```diff
@@ -1,672 +0,0 @@
-/* This file is part of VoltDB.
```
BillWhite: This test is now called TestRank.

tests/ee/plannodes/PartitionByPlanNodeTest.cpp
```diff
@@ -1,158 +0,0 @@
-/* This file is part of VoltDB.
```
BillWhite: This test is now called WindowFunctionPlanTest.  It's essentially the same, but with the newly refactored plan node class.  The new tests looks at the schema of the output columns more, but not in great detail.

tests/ee/test_utils/LoadTableFrom.hpp
```diff
@@ -61,56 +68,55 @@ namespace voltdb {
...
-        printf("  msg size:              %d\n",   msg_size);
-        printf("  status:                %hhd\n", status);
-        printf("  inter cluster latency: %d\n",   icl);
-        printf("  serialized exception:  %d\n",   serialized_exp);
-        printf("  table length:          %d\n",   tbl_len);
-        printf("  table metadata length: %d\n",   tbl_metadata_len);
-        printf("  table_status:          %hhd\n", tbl_status);
-        printf("  column count:          %d\n",   column_count);
-    }
+    // These varaibles are only used if
```
BillWhite: Here I mostly replaced the static variable debug_print with VOLT_TRACE.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+    // types.  They are more like expression operator types, like
+    // MIN, MAX etc.
+    protected List<ExpressionType> m_aggregateTypes = new ArrayList<>();
+    // a list of whether the aggregate is over distinct elements
+    protected List<Boolean> m_aggregateDistinct = new ArrayList<>();
+    // a list of column offsets/indexes not plan column guids.
+    protected List<Integer> m_aggregateOutputColumns = new ArrayList<>();
+    // List of the input TVEs into the aggregates.  Maybe should become
+    // a list of SchemaColumns someday
+    protected List<List<AbstractExpression>> m_aggregateExpressions = new ArrayList<>();
```
paulmartel: The comment does not match the code. Justifiably, AFAIK, because the code is wrong.
Preface: I don't know of any plan to support windowedexpressions that take more than one argument.
If there is such a plan, let me know and disregard this review comment.
It used to be that just the WindowedExpression contained a List of arguments even though it just needed one or zero -- easily represented as a valid AbstractExpression or null.
Presumably to fit in better with generic AggregatePlanNode, the original PartitionByPlanNode only used the first element of the List or null if the list was empty.
Now that WindowFunctionPlanNode is divorced from AggregatePlanNode -- the unwarranted complexity of the List<AbstractExpression> vs simple (possibly null) AbstractExpression is getting propagated into WindowFunctionPlanNode and on into the EE where it is reconstituted as a vector so that we can assert that the vector is empty for the current *RANK functions and presumably assert that we have only one element in future window functions. Let's not. Let's instead go back to the argument for each agg being a single AbstractExpression or null and instead propagate this simplification BACK UP into WindowedExpression.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
+/* This file is part of VoltDB.
```
paulmartel: Why is it WindowedExpression but WindowFunctionPlanNode. I think I'd prefer simply WindowedPlanNode for consistency and brevity.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+    // At the moment these are guaranteed to be TVES.  This might always be true
+    protected List<AbstractExpression> m_partitionByExpressions = new ArrayList<>();
+    // This is the list of TVEs for the window functions.  There
+    // will only be one of them for now.
+    private List<TupleValueExpression> m_tve = new ArrayList<>();
+
+    // List of the lists of order by expressions.  If there
+    // are no order by expressions in one aggregate the list
+    // is empty, not null.
+    protected List<AbstractExpression> m_orderByExpressions = new ArrayList<>();
```
paulmartel: This is NOT a list of lists as described in the comment. Maybe it will be one day or maybe not if we have separate plannodes whenever we have separate order by clauses -- which MAY be easier to deal with. Change the comment. Explain that these order by expressions apply to the one windowed expression or for all the windowed expressions of which there is currently only one -- depending on your best predictions for the medium term.
Make it clear that this list is not one-element-per-windowed-function -- I don't THINK it is or will be. I think that could be mistakenly inferred given how some of the other lists work.

src/ee/plannodes/abstractplannode.cpp
paulmartel: HOW didn't this work for empty vectors? It's an idiom.
size() returns 0. each-- evaluates to 0, THEN decrements each but who cares. The while loop exits. What could possibly go wrong?

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
```diff
@@ -127,7 +127,7 @@ private void validateWindowedFunctionPlan(String windowedQuery, int nSorts, int
         assertTrue(projPlanNode instanceof ProjectionPlanNode);
 
         AbstractPlanNode partitionByPlanNode = projPlanNode.getChild(0);
-        assertTrue(partitionByPlanNode instanceof PartitionByPlanNode);
+        assertTrue(partitionByPlanNode instanceof WindowFunctionPlanNode);
```
paulmartel: This test name uses yet another variant on Window/Windowed Function/Expression. Rename for consistency.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+        AGGREGATE_EXPRESSIONS,
+        PARTITIONBY_EXPRESSIONS
+
+    };
+    // A list of aggregate types.  These are not programming language
+    // types.  They are more like expression operator types, like
+    // MIN, MAX etc.
+    protected List<ExpressionType> m_aggregateTypes = new ArrayList<>();
+    // a list of whether the aggregate is over distinct elements
+    protected List<Boolean> m_aggregateDistinct = new ArrayList<>();
```
paulmartel: The switch from List<Integer> to List<Boolean> for m_aggregateDistinct here makes sense.
It makes as much sense for AggregatePlanNode, and avoiding arbitrary differences like this between the two modules makes even more sense.
Back-fit this type change to AggregatePlanNode.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+    // List of the input TVEs into the aggregates.  Maybe should become
+    // a list of SchemaColumns someday
+    protected List<List<AbstractExpression>> m_aggregateExpressions = new ArrayList<>();
+    // There is one list of partition by expressions for this WindowFunctionPlanNode.
+    // If there is more than one window we will need more than one WindowFunctionPlanNode.
+    // At the moment these are guaranteed to be TVES.  This might always be true
+    protected List<AbstractExpression> m_partitionByExpressions = new ArrayList<>();
+    // This is the list of TVEs for the window functions.  There
+    // will only be one of them for now.
+    private List<TupleValueExpression> m_tve = new ArrayList<>();
```
paulmartel: This name is ambiguous and the comment helps only a little: suggest alternative m_outputTVE.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+        m_aggregateDistinct.add(winExpr.getIsDistinct());
+        if (winExpr.getAggregateArguments().size() > 0) {
+            m_aggregateExpressions.add(winExpr.getAggregateArguments());
+        }
+        else {
+            m_aggregateExpressions.add(null);
+        }
+        m_partitionByExpressions = winExpr.getPartitionByExpressions();
+        m_orderByExpressions = winExpr.getOrderByExpressions();
+        m_tve.add(winExpr.getDisplayListExpression());
```
paulmartel: The name `getAggregateOutputTVE()` would be more accurate than `getDisplayListExpression()`.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+    // This is the list of TVEs for the window functions.  There
+    // will only be one of them for now.
+    private List<TupleValueExpression> m_tve = new ArrayList<>();
+
+    // List of the lists of order by expressions.  If there
+    // are no order by expressions in one aggregate the list
+    // is empty, not null.
+    protected List<AbstractExpression> m_orderByExpressions = new ArrayList<>();
+
+    private int getNumberAggregateFunctions() {
```
paulmartel: getNumberOfAggregateFunctions() or getAggregateFunctionCount() would be more readable, conventional.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+     * be in the order by plan node in any case.
+     */
+    @Override
+    public void toJSONString(JSONStringer stringer) throws JSONException {
+        super.toJSONString(stringer);
+        stringer.key("AGGREGATE_COLUMNS")
+                .array();
+        for (int ii = 0; ii < m_aggregateTypes.size(); ii++) {
+            stringer.object();
+            stringer.key(Members.AGGREGATE_TYPE.name()).value(m_aggregateTypes.get(ii).name());
```
paulmartel: The idiom:
`stringer.key(Members...name()).value(`
has been replaced with the faster:
`stringer.keySymbolValuePair(Members...name(), `.
You missed this in a merge from master after the refactor.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -0,0 +1,368 @@
...
+namespace voltdb {
+
+class ProgressMonitorProxy;
+/**
+ * A WindowAggregate is used to calculate one kind of
+ * window function.  The algorithm
+ */
+class WindowAggregate {
+public:
+    virtual ~WindowAggregate() {
```
paulmartel: Is there a point to defining all these no-op virtual constructors and then not calling them anywhere?

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -0,0 +1,368 @@
...
+#include "plannodes/windowfunctionnode.h"
+
+namespace voltdb {
+
+class ProgressMonitorProxy;
+/**
+ * A WindowAggregate is used to calculate one kind of
+ * window function.  The algorithm
+ */
+class WindowAggregate {
```
paulmartel: There is no reason to expose WindowAggregate and WindowAggregateRow in this header. 
Put a forward declaration to WindowAggregateRow in the voltdb namespace here and move 
these classes into the .cpp as implementation details.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -0,0 +1,368 @@
...
+        m_lastOrderByKeyTuple(m_lastOrderByKeyStorage),
+        m_inputSchema(NULL),
+        m_aggregateRow(NULL)
+        {
+            dynamic_cast<WindowFunctionPlanNode *>(abstract_node)->collectOutputExpressions(m_outputColumnExpressions);
+        }
+    virtual ~WindowFunctionExecutor();
+
+
+    /**
```
paulmartel: I'm not seeing any point to these accessors being declared public. I'm not seeing much point in their being defined at all. Direct access to the private members within the class' methods seems like fair game.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -0,0 +1,368 @@
...
+     * used to calculate the pass through columns.
+     */
+    const std::vector<AbstractExpression*>& getOutputColumnExpressions() const {
+        return m_outputColumnExpressions;
+    }
+
+    /**
+     * Initiate the member variables for execute the window function.
+     */
+    TableTuple p_execute_init(const NValueArray& params,
```
paulmartel: There's an argument that these p_execute_* sub-functions were motivated mostly by the inlining of aggregates
and could be swallowed up back into p_execute for this class.
I suppose that a more index-aware windowed plan node COULD some day be inlined, so maybe it's worth keeping them as is.

src/ee/executors/windowfunctionexecutor.cpp
```diff
@@ -0,0 +1,491 @@
...
+}
+
+/**
+ * This RAII class sets a pointer to null when it is goes out of scope.
+ * We could use boost::scoped_ptr with a specialized destructor, but it
+ * seems more obscure than this simple class.
+ */
+template <typename T>
+struct ScopedNullingPointer {
+    ScopedNullingPointer(T *&ptr, T *value)
```
paulmartel: This is useful, but overkill. This class body could be three lines long. The class needn't to know about value. If the caller wants to assign a value to the original pointer before or after it is associated with this guard class, it can use an assignment statement. Similarly, there's no point in the fancy operator to expose the original pointer because the caller can access the original pointer directly. Finally, not setting the pointer to null if it is already null doesn't add much but code bulk. And all that pointless whitespace after T                                            just looks ugly.

tests/ee/plannodes/WindowFunctionPlanNodeTest.cpp
```diff
@@ -0,0 +1,262 @@
...
+        EXPECT_EQ(jsonDescr->nAggs, pn->getAggregates().size());
+        for (int aggIdx = 0; aggIdx < pn->getAggregates().size(); aggIdx += 1) {
+            aggDescription &aggDescr = jsonDescr->aggDescr;
+            EXPECT_EQ(aggDescr.aggType, pn->getAggregates()[aggIdx]);
+            EXPECT_EQ(aggDescr.nAggArgs, pn->getAggregateInputExpressions()[aggIdx].size());
+        }
+        EXPECT_EQ(jsonDescr->nPartitionByExprs, pn->getPartitionByExpressions().size());
+        EXPECT_EQ(jsonDescr->nOrderByExprs, pn->getOrderByExpressions().size());
+        EXPECT_EQ(jsonDescr->nOutputColumns, pn->getOutputSchema().size());
+    EXPECT_EQ(jsonDescr->nOutputColumns, jsonDescr->colDescriptions.getNumColumns());
```
paulmartel: Strange indent.

src/ee/executors/windowfunctionexecutor.cpp
```diff
@@ -0,0 +1,491 @@
...
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/*
+ * partitionbyexecutor.cpp
+ */
+
+#include "executors/windowfunctionexecutor.h"
+
+#include "../plannodes/windowfunctionnode.h"
+#include "../execution/ProgressMonitorProxy.h"
```
BillWhite: Fixed.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -0,0 +1,368 @@
...
+    }
+    void recordPassThroughTuple(const TableTuple &nextTuple) {
+        m_passThroughTuple.copy(nextTuple);
+    }
+    TableTuple &getPassThroughTuple() {
+        return m_passThroughTuple;
+    }
+private:
+    PoolBackedTupleStorage  m_passThroughStorage;
+    TableTuple             &m_passThroughTuple;
```
BillWhite: Done.  I also changed the m_inProgress*Tuple and m_last*Tuple to functions which access m_inProgress*Storage and m_last*Tuple.  The tuple references are mostly local variables in the aggregate executors, and I found them confusing.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
BillWhite: Well, we call them WindowFunctions.  The standard calls them window functions.  So, I have changed them to WindowFunctionExpression.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
BillWhite: These functions are binary functions:
COVAR_POP
COVAR_SAMP
CORR
REGR_SLOPE
REGR_INTERCEPT
REGR_COUNT
REGR_R2
REGR_AVGX
REGR_AVGY
REGR_SXX
REGR_SYY
REGR_SXY

We don't have plans to implement anything except for COUNT.  But maybe our plans will change.

This doesn't seem more complex to me, in any case.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
BillWhite: True.  This is a false start which leaked out.  I have fixed the comment.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
BillWhite: Fixed.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
BillWhite: Changed to getAggregateFunctionCount().

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+        m_partitionByExpressions = winExpr.getPartitionByExpressions();
+        m_orderByExpressions = winExpr.getOrderByExpressions();
+        m_tve.add(winExpr.getDisplayListExpression());
+    }
+
+    /**
+     * Serialize to JSON.  We only serialize the expressions, and not the
+     * directions.  We won't need them in the executor.  The directions will
+     * be in the order by plan node in any case.
```
BillWhite: This is not so.  These are sort criteria.  For order by nodes we need to serialize the sort directions, ASC or DESC.  Here we don't need the directions, as we don't actually sort anything in the EE in a WindowFunctionExecutor.  But we do compute the sort expressions in the WindowFunctionExecutor, so we need to serialize them.  We need to explain here why we don't serialize the sort directions.  This was the cause of a bug earlier on, and I wanted to warn against making it again in the future.

The comment about amphibians is ad hoc and misplaced.  A better choice might have been: No amphibolies have been created in the typing of this comment.  But that would have been in category and perhaps relevant.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
BillWhite: Sorry.  I forgot this.  I've fixed it.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
BillWhite: I think these are all WindowFunction* now.  That is, WindowFunctionExpression, WindowFunctionPlanNode and WindowFunctionExecutor.

src/ee/plannodes/plannodeutil.cpp
```diff
@@ -67,9 +67,8 @@
 #include "plannodes/tuplescannode.h"
 #include "plannodes/unionnode.h"
 #include "plannodes/updatenode.h"
-#include "plannodes/partitionbynode.h"
-
 #include <sstream>
+#include "windowfunctionnode.h"
```
BillWhite: Fixed.

src/ee/executors/windowfunctionexecutor.h
BillWhite: Done.

src/ee/executors/windowfunctionexecutor.h
BillWhite: I think this is a destructor.  The compiler complains if a function with virtual functions does not have a virtual destructor.

src/ee/executors/windowfunctionexecutor.cpp
BillWhite: But what if an exception is thrown through this function?  Then the function can't assign NULL to m_pmp, since it's an exception, and m_pmp points to a local variable in a function which has exited.  This seems like a problem.

tests/ee/plannodes/WindowFunctionPlanNodeTest.cpp
BillWhite: Fixed.

PR #4067, opened 2016-11-15
ENG-11448: Allocate initiateTaskBuffer in setStateForDurability() onl…

slowsoul:
…y once

### DIFF
--- a/src/frontend/org/voltdb/messaging/FragmentTaskMessage.java
+++ b/src/frontend/org/voltdb/messaging/FragmentTaskMessage.java
@@ -456,6 +456,10 @@ public class FragmentTaskMessage extends TransactionInfoBaseMessage
         if (m_initiateTaskBuffer == null) {
             m_initiateTaskBuffer = ByteBuffer.allocate(initiateTask.getSerializedSize());
         }
+        else {
+            m_initiateTaskBuffer.position(0);
+            m_initiateTaskBuffer.limit(initiateTask.getSerializedSize());
+        }
         try {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/messaging/FragmentTaskMessage.java
```diff
@@ -453,7 +453,9 @@ public void setStateForDurability(Iv2InitiateTaskMessage initiateTask,
                                       Collection<Integer> involvedPartitions) {
         m_initiateTask = initiateTask;
         m_involvedPartitions = ImmutableSet.copyOf(involvedPartitions);
-        m_initiateTaskBuffer = ByteBuffer.allocate(initiateTask.getSerializedSize());
+        if (m_initiateTaskBuffer == null) {
+            m_initiateTaskBuffer = ByteBuffer.allocate(initiateTask.getSerializedSize());
+        }
```
nshi: If it's not null, don't you have to reset the buffer so that the position is at 0 and limit is at `initiateTask.getSerializedSize()`?

slowsoul: I think since every time flip() is called on m_initiateTaskBuffer, it should be fine. But that's under the assumption that the passed-in initiateTask is of the same size for a single FragmentTaskMessage instance. If this doesn't hold, we should add the reset, this is also more defensive.

slowsoul: Fixed. the only assumption now is that the subsequent initiateTask must be smaller or equal to the first initiateTask passed in.

PR #4068, opened 2016-11-15
ENG-11448: Allocate initiateTaskBuffer in setStateForDurability() only once

slowsoul:


src/frontend/org/voltdb/messaging/FragmentTaskMessage.java
```diff
@@ -453,7 +453,13 @@ public void setStateForDurability(Iv2InitiateTaskMessage initiateTask,
                                       Collection<Integer> involvedPartitions) {
         m_initiateTask = initiateTask;
         m_involvedPartitions = ImmutableSet.copyOf(involvedPartitions);
-        m_initiateTaskBuffer = ByteBuffer.allocate(initiateTask.getSerializedSize());
+        if (m_initiateTaskBuffer == null) {
```
nshi: I think it may be worth putting a big comment here explaining why this is done and why the new InitiateTaskMessage won't be bigger than the already-allocated ByteBuffer.

slowsoul: Done.

PR #4069, opened 2016-11-15
ENG-11475 determine local replicas first before client interface acce…

vtkstef:
determine local replicas first before client interface accepts connections


akhanzode:

PR #4070, opened 2016-11-16
ENG-11480 add timeout for DR consumer stats

zeemanhe:
The DR consumer stats checking will be timeout and warning will be presented to users.

### DIFF
--- a/lib/python/voltcli/checkstats.py
+++ b/lib/python/voltcli/checkstats.py
@@ -252,8 +252,8 @@ def check_dr_consumer(runner):
         if outstanding == 0:
             return
         timelast = time.time() - start
-        #if the DR consumer stats are not cleaned in 15 min, return and warn user
-        if timelast > 900 :
+        #if the DR consumer stats are not cleaned in 2 seconds, return and warn user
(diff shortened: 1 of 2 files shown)


lib/python/voltcli/checkstats.py
```diff
@@ -77,7 +77,7 @@ def get_stats(runner, component):
         response = runner.call_proc('@Statistics', [voltdbclient.FastSerializer.VOLTTYPE_STRING,
                                     voltdbclient.FastSerializer.VOLTTYPE_INTEGER], [component, 0])
         status = response.status()
-        if status <> 1 and "timeout" in response.statusString:
+        if status <> 1 and "timeout" in response.response.statusString:
```
nshi: Was this getting the wrong object? It looked like it should have never worked.

zeemanhe: yeah, it was a wrong object.  The error was never caught (runner.call_proc checks status by default, if status is not 1, abort).

xinjiacs:
Not going to happen in production by shutting down consumer cluster first.

PR #4074, opened 2016-11-17
Don't enforce create view restrictions when dropping views

wolffcm:


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -470,16 +470,23 @@ protected String checkAddDropWhitelist(final CatalogType suspect, final ChangeTy
...
-                }
+
+            // Remember the name of the new table.
+            m_newTables.add(tableName.toUpperCase());
+            if (CatalogUtil.isTableExportOnly((Database)tbl.getParent(), tbl)) {
+                // Remember that it's a new export table.
+                m_newTablesForExport.add(tbl.getTypeName());
             }
+
+
```
yzhang1991: Remove the blank lines?

PR #4079, opened 2016-11-18
Check for unsafe operations in join and where expressions.

BillWhite:
When creating a materialized view we need to check join and where
expressions for operations which are unsafe when the source tables are
all non-empty.


src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -2523,4 +2523,15 @@ public boolean isPartitionColumnInWindowedAggregatePartitionByList() {
...
     }
 
+    /**
+     * Gather up all the join expressions.  This happens before planning.
+     * So, all the innerouter and innerinner lists are not yet populated.
+     *
+     * @param checkExpressions
+     */
+    public void gatherJoinExpressions(List<AbstractExpression> checkExpressions) {
+        // TODO Auto-generated method stub
```
wolffcm: I guess you can remove this TODO comment.

Also, in the future, if we call this method at an earlier point in the planner flow, is it possible that we'll think there are no join expressions when there really are (or will be)?  Or is it the case that before we analyze the joins, `m_joinTree` will be null and we'll just get a NPE?

template.classpath
```diff
@@ -1,75 +1,75 @@
...
-        <classpathentry kind="lib" path="third_party/java/jars/guava-18.0.jar"/>
-        <classpathentry kind="lib" path="third_party/java/jars/protobuf-java-2.6.1.jar"/>
-        <classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER">
-                <attributes>
-                        <attribute name="org.eclipse.jdt.launching.CLASSPATH_ATTR_LIBRARY_PATH_ENTRY" value="voltdb/voltdb"/>
-                </attributes>
-        </classpathentry>
-        <classpathentry kind="lib" path="lib/tomcat-juli.jar"/>
-        <classpathentry kind="output" path="obj/eclipse"/>
+	<classpathentry kind="src" path="src/frontend"/>
```
wolffcm: Looks like you replaced spaces for tabs in this file?

wolffcm: Whitespace issues aside, there's more changes in this file than I would have expected?  Could be from different versions of Eclipse?  Maybe it's fine though.

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -2237,6 +2237,40 @@ public void testEng11314() throws Exception {
...
+        }
+        client.callProcedure("@AdHoc", "drop view mv if exists");
+        assertFalse("Unexpected compilation success", success);
+        assertTrue("Did not find pattern \"" + pattern + "\" in error message.",
+                   msg.contains(pattern));
+    }
+
+    public void testNowFailed() throws Exception {
+        String sql;
+        sql = "CREATE VIEW MV AS SELECT L.ID, COUNT(*) FROM ENG11495 AS L JOIN ENG11495 AS R ON L.TS = NOW GROUP BY L.ID";
```
wolffcm: There are also very similar tests for DDL that is expected to fail in TestDDLCompiler, or TestVoltCompiler.  

On the one hand I think having our tests spread out over several files to be not so good as it makes tests harder to find.

On the other hand TestVoltCompiler and TestDDLCompiler are big sprawling tests that cover lots of features, so I'm not sure they need to be any bigger.

I don't know that I feel strongly one way or the other, this is just something to keep in mind.

BillWhite: Well, this is a materialized view test, so I felt it should be in the MV test suite.  But it could be rehomed.  I'll keep it in mind.

template.classpath
BillWhite: I think I just changed the version numbers on two jar files: scala_parser_combinator and scala_xml.  But they get put in a different place in the file sometimes.  It doesn't matter as far as Eclipse is concerned, so it doesn't matter all that much.

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
BillWhite: Sure.  I'll do that later on.

PR #4080, opened 2016-11-18
ENG-11339:

manjujames:
Added datasourcecluster commandline option to legacy voltdb cmdline.
  This will be added to 'init' of new cmdline utility in a different ticket.

Also added cluster recovery time to connect request, which was missed in the previous ticket.


src/frontend/org/voltdb/utils/CommandLine.java
```diff
@@ -774,6 +775,11 @@ public void vemTag(String tag) {
             cmdline.add("voltdbroot"); cmdline.add(m_voltdbRoot.getPath());
         }
 
+        if (m_datasourceClusterId != -1)
+        {
+            cmdline.add("datasourcecluster"); cmdline.add(Integer.toString(m_datasourceClusterId));
```
manjujames: Yes, absolutely

PR #4083, opened 2016-11-21
Do all-at-once deallocation of relocatables at transaction boundaries

wolffcm:
A user noticed that TRUNCATE TABLE could take a long time to commit if the table contained a lot of string data.  I found that this lag was due to releasing memory for string data managed by ContiguousAllocator.  It's normally a hole-filling-allocator, but when many objects are released at the same time, we might be doing a lot of memcpys for things that may be deleted anyway.

These changes make it so we defer actual deallocation at transaction end until we know all the items that will be deallocated, and then we can avoid doing any needless copying.

This does not eliminate the problem, but is a significant improvement.

### DIFF
--- a/src/ee/common/ThreadLocalPool.cpp
+++ b/src/ee/common/ThreadLocalPool.cpp
@@ -408,7 +408,7 @@ ScopedPoolDeferredReleaseMode::~ScopedPoolDeferredReleaseMode() {
         entry.second->freePendingAllocations(pendingReleaseSet);
     }
     assert (pendingReleaseSet.empty());
-    getPerThreadPools()->setRelocatablePoolReleaseMode(DEFERRED_RELEASE);
+    getPerThreadPools()->setRelocatablePoolReleaseMode(IMMEDIATE_RELEASE);
 }
 
 }
--- a/src/ee/common/ThreadLocalPool.cpp
+++ b/src/ee/common/ThreadLocalPool.cpp
@@ -46,79 +46,109 @@ const int ThreadLocalPool::POOLED_MAX_VALUE_LENGTH = 1024 * 1024;
 /**
  * Thread local key for storing thread specific memory pools
  */
-static pthread_key_t m_key;
-static pthread_key_t m_stringKey;
-static pthread_key_t m_releaseModeKey;
-/**
- * Thread local key for storing integer value of amount of memory allocated
- */
-static pthread_key_t m_keyAllocated;
+static pthread_key_t m_perThreadPoolsKey;
 static pthread_once_t m_keyOnce = PTHREAD_ONCE_INIT;
 
 typedef boost::pool<voltdb_pool_allocator_new_delete> PoolForObjectSize;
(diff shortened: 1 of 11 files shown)


src/ee/common/ThreadLocalPool.cpp
```diff
@@ -232,8 +262,14 @@ void ThreadLocalPool::freeRelocatable(Sized* sized)
...
-    // Free the raw allocation from the found pool.
-    iter->second->free(sized);
+
+    if (! isDeferredReleaseMode()) {
+        // Free the raw allocation from the found pool.
+        iter->second->free(sized);
+    }
+    else {
+        iter->second->markAllocationAsPendingRelease(sized);
+    }
```
paulmartel: Is it possible/plausible to detect the no-copy case when sized is last() for its contiguous allocator and execute the free immediately, avoiding the boost::unordered_set entry overhead for this edge case. I'm generally concerned about short-term memory growth spikes at transaction end. Is there some way of unit test benchmarking RSS effect as well? Also, is there a hidden multi-thread performance cost of multiple site threads contending over the OS memory allocator in/for the ordered_set implementation?

src/ee/common/ThreadLocalPool.cpp
```diff
@@ -20,12 +20,14 @@
 #include "common/SQLException.h"
 
 #include "structures/CompactingPool.h"
+#include "structures/CompactingSet.h"
```
yzhang1991: Not needed.

src/ee/common/ThreadLocalPool.cpp
```diff
@@ -46,64 +48,121 @@ const int ThreadLocalPool::POOLED_MAX_VALUE_LENGTH = 1024 * 1024;
...
+    void decrementRefCount() {
+        --m_refCount;
+        assert(m_refCount >= 0);
+    }
+
+    int getRefCount() const {
+        return m_refCount;
+    }
+
+    CompactingStringStorage& getStringPoolMap() {
```
yzhang1991: Why don't we rename `m_relocatablePools` to something like `m_stringPoolMap`?
I keep getting myself confused distinguishing those names, although for people who are familiar with them it might be easy.

src/ee/common/ThreadLocalPool.h
```diff
@@ -112,9 +114,37 @@ class ThreadLocalPool {
...
+ * This is desirable because the hole-filling algorithm CompactingPool
+ * uses to keep memory contiguous can incur needless memcpying.
+ * Freeing many objects at once can allows us to avoid copying objects
+ * that are soon to be deleted anyway.
+ */
+struct ScopedPoolDeferredReleaseMode {
+    ScopedPoolDeferredReleaseMode();
+    ~ScopedPoolDeferredReleaseMode();
+private:
+    bool m_origMode;
```
yzhang1991: `m_orignalMode` is more readable, and it is not much longer.

PR #4090, opened 2016-11-25
ENG-11532: Make a copy of connections when retrieving them from VoltNetwork instances

slowsoul:


src/frontend/org/voltcore/network/VoltNetwork.java
```diff
@@ -549,8 +551,18 @@ int numPorts() {
...
+        final SettableFuture<Set<Connection>> connectionsFuture = SettableFuture.create();
+        m_tasks.offer(new Runnable() {
+            @Override
+            public void run() {
+                // Make a copy of m_ports to avoid concurrent modification
+                connectionsFuture.set(new HashSet<Connection>(m_ports));
+            }
+        });
+        try {
+            return connectionsFuture.get();
```
slowsoul: Done. Actually there's a queueTask() method there, I didn't see it :(

src/frontend/org/voltcore/network/VoltNetworkPool.java
```diff
@@ -156,7 +156,13 @@ public Connection registerChannel(
     public Set<Connection> getConnections() {
         Set<Connection> conns = new HashSet<>();
         for (VoltNetwork vn : m_networks) {
-            conns.addAll(vn.getConnections());
+            Set<Connection> connsForNetwork;
+            try {
+                connsForNetwork = vn.getConnections().get();
```
nshi: It may be faster to save all the `Future`s here and iterate through them to get the final results in another loop below. That way you submit the work to all network threads without waiting.

slowsoul: Yes, addressed it in the latest commit.

PR #4091, opened 2016-11-25
ENG-11532: Make a copy of connections when retrieving them from VoltNetwork instances

slowsoul:


src/frontend/org/voltcore/network/VoltNetwork.java
```diff
@@ -549,8 +549,7 @@ int numPorts() {
     }
 
     public Set<Connection> getConnections() {
-        Set<Connection> conns = new HashSet<>();
-        conns.addAll(m_ports);
-        return conns;
+        // Make a copy of m_ports to avoid concurrent modification
+        return new HashSet<>(m_ports);
```
nshi: The copy constructor also calls `addAll()` internally, so this has no difference. I wonder what the other modification was when `getConnections()` was called. The for loop should only call this at the beginning. Maybe it was the heartbeats sent out in the previous accept?

slowsoul: Oh yes, I didn't think about that. I think it might be that the VoltNetwork is unregistering some stale connections and removing from m_ports, which is triggered by the heartbeats

nshi: I think the culprit is that `VoltNetwork.getConnections()` is not thread-safe. It should use the network thread to access the `m_ports`.

slowsoul: I see, yes, it should, so a good way is to offer task into each VoltNetwork's thread and use futures to get the connections, right? Since it's only the acceptor thread that needs to wait for the futures, it won't affect DR producer thread.

nshi: Yes, I think that might be the way to fix it.

slowsoul: Done.

PR #4093, opened 2016-11-26
ENG-11528: skip EE JNI call when there are no diff commands to apply

xinjiacs:
EE JNI can take time to return, especially with high number of sites per host. This is a easy performance improvement.


src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -1433,6 +1433,11 @@ public boolean updateCatalog(String diffCmds, CatalogContext context, CatalogSpe
             return true;
         }
 
+        if (diffCmds.length() == 0) {
+            // empty diff cmds for the EE to apply, so skip the JNI call
+            return true;
+        }
+
```
zeemanhe: Look good to me.

src/catgen/in/javasrc/Catalog.java
```diff
@@ -109,9 +109,29 @@ public void execute(final String commands) {
...
+        public boolean isProcedureRelatedCmd() {
+            if (path.indexOf("procedures#") != -1) return true;
+            if ("procedures".equals(arg1) && path.endsWith("#database")) return true;
+
+            return false;
+        }
+    }
 
+    public static CatalogCmd parseStmt(String stmt) {
+        // command comes before the first space (add or set)
```
zeemanhe: It would be of help if you can list the pattens of stmt.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2689,4 +2691,35 @@ public static DeploymentType updateRuntimeDeploymentPaths(DeploymentType deploym
...
+
+        StringBuilder sb = new StringBuilder();
+        String[] cmds = diffCmds.split("\n");
+        boolean skip = false;
+        for (int i = 0; i < cmds.length; i++) {
+            String stmt = cmds[i];
+
+            CatalogCmd catCmd = Catalog.parseStmt(stmt);
+            if (catCmd.cmd == 'a' || catCmd.cmd == 'd') { // add, del
+                skip = catCmd.isProcedureRelatedCmd() ? true : false;
```
zeemanhe: remove "? true : false".  
It seems CatalogUtil.getDiffCommandsForEE() is invoke only once on Site.  If so, I would recommend to rename this method and return a boolean as soon as a delete or add command is identified.

src/catgen/in/javasrc/Catalog.java
xinjiacs: Yes, I will add sample diff commands text.

src/frontend/org/voltdb/utils/CatalogUtil.java
xinjiacs: It's called per site per diff statement. There is likely an enhancement that can do less parsing work. It's in my new commit.

PR #4099, opened 2016-11-30
Eng 11366 windowed count

BillWhite:
Implement count(*) and count(EXPR) window functions.

https://issues.voltdb.com/browse/ENG-11366


build.py
```diff
@@ -497,6 +497,7 @@
     MergeReceiveExecutorTest
     TestGeneratedPlans
     TestRank
+    TestCount
```
wolffcm: This test just does windowed count, right?  So maybe make the name reflect this.

BillWhite: Fixed.  Also TestRank is renamed to TestWindowedRank.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -34,27 +34,47 @@ class WindowFunctionExecutor: public AbstractExecutor {
...
         m_pmp(NULL),
         m_orderByKeySchema(NULL),
         m_partitionByKeySchema(NULL),
+        m_inputTable(NULL),
         m_inputSchema(NULL),
-        m_aggregateRow(NULL)
+        m_aggregateRow(NULL),
+        m_tableWindow(NULL)
         {
             dynamic_cast<WindowFunctionPlanNode *>(abstract_node)->collectOutputExpressions(m_outputColumnExpressions);
```
BillWhite: I take it back.  These are references.  They need to be initialized in the constructor.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -34,27 +34,47 @@ class WindowFunctionExecutor: public AbstractExecutor {
...
+     * depend on the order by peers of the row.  So, we need to
+     * scan the input forward to the next edge between order by
+     * groups.  Edges between partition by groups are a kind of
+     * edge between order by groups as well.
+     *
+     * This enum type gives the type of kinds of edges between groups
+     * of rows.
+     */
+    enum EdgeType {
+        InvalidEdgeType,         /** No Type. */
```
BillWhite: Fixed.

src/ee/executors/windowfunctionexecutor.h
```diff
@@ -148,27 +162,45 @@ class WindowFunctionExecutor: public AbstractExecutor {
...
+     * Call endGroup for each aggregate.  This will happen
+     * for each group and cannot be disabled.
+     */
+    void endGroupForAggs(TableWindow *window, EdgeType edgeType);
+    /**
+     * Insert the output tuple.
+     */
+    void insertOutputTuple();
 
     TupleSchema* constructSchemaFromExpressionVector(const AbstractPlanNode::OwningExpressionVector &exprs);
```
BillWhite: Ok, I moved it, and renamed it createTupleSchema, which is what the rest of the similar factory functions are called.

tests/frontend/org/voltdb/regressionsuites/TestWindowedAggregateSuite.java
```diff
@@ -693,6 +693,174 @@ public void testEng11029() throws Exception {
         }, vt);
     }
 
+    private void validateCount(Client client, String sql, long expected[][]) throws Exception {
```
BillWhite: Actually, I ran the memcheck tests and they all passed.  I'm writing tests with more data types and partitioned tables.  But as far as data types go, the WindowAggregate objects, which do all the calculation, contain only NValues with integers in them.  We don't actually save any strings there.  Consequently I'm not sure what a good set of tests would be.   A white box test would be one which tried to create an NValue string which didn't get collected, or which got garbage collected some how before it was used.  I'm not sure how that could happen.  But I'm writing the tests as well as I can.

PR #4100, opened 2016-12-01
ENG-11434 voltadmin shutdown/pause timeout

zeemanhe:
added a mechanism to monitor the draining of outstanding transactions.  The command will be timed out if no transactions have been drained within timeout period after the last successfully drained transaction.

### DIFF
--- a/lib/python/voltcli/checkstats.py
+++ b/lib/python/voltcli/checkstats.py
@@ -137,8 +137,7 @@ def dr_producer_stats(runner, partition_min_host, partition_min, partition_max):
         else:
             last_acked = r[9]
 
-        #The partition is in the initial state. no stats updated yet-no transactions are queued and acknowledged 
-        #thus no need to track. 
+        # Initial state, no transactions are queued and acknowledged.
         if last_queued == -1 and last_acked == -1:
             continue
 
--- a/lib/python/voltcli/checkstats.py
+++ b/lib/python/voltcli/checkstats.py
@@ -80,20 +80,18 @@ def check_dr_producer(runner, timeout=-1):
         LastPartitionMin = partition_min.copy()
         lastPartitionMax = partition_max.copy()
 
-
 def verifyDrProducerTimeout(lastPartitionMin, currentPartitionMin,
                             lastPartitionMax, currentPartitionMax, lastUpdatedTime, timeout):
     currentTime = time.time()
-    #not timeout check
+    #no timeout check
     if timeout == -1:
         return currentTime
-    #are stats moved?
-    updated = (comp(lastPartitionMin,currentPartitionMin) <> 0)
-    if updated:
-        updated = (comp(lastPartitionMax,currentPartitionMax) <> 0)
+    #any stats progress?
+    partitionMinProgressed = comp(lastPartitionMin,currentPartitionMin)
+    partitionMaxprogressed = comp(lastPartitionMax,currentPartitionMax)
 
     #stats moved
-    if updated == False:
+    if partitionMinProgressed <> 0 or partitionMaxprogressed <> 0:
         return currentTime
 
     timeSinceLastUpdate = currentTime - lastUpdatedTime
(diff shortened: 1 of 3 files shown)


lib/python/voltcli/checkstats.py
```diff
@@ -269,7 +340,49 @@ def check_command_log(runner):
...
+        lastUpdatedTime = monitorStatisticsProgress(lastValidationParamms, currentValidationParams, lastUpdatedTime, runner, 'COMMANDLOG')
+        lastValidationParamms = [outstandingByte, outstandingTxn]
         time.sleep(1)
+
+def monitorStatisticsProgress(lastUpdatedParams, currentParams, lastUpdatedTime, runner, component):
+    currentTime = time.time()
+
+    timeout = -1;
+    if runner.opts.timeout:
+        timeout = (runner.opts.timeout) * 60
```
xinjiacs: Do we have any negative timeout input checking?

lib/python/voltcli/checkstats.py
```diff
@@ -108,6 +142,10 @@ def dr_producer_stats(runner, partition_min_host, partition_min, partition_max):
             last_acked = -1
         else:
             last_acked = r[9]
+
+        if last_queued == -1 and last_acked == -1:
```
xinjiacs: What's the case for -1? Can you add comments for it?

zeemanhe: no transactions have been registered  for the partition in the queue yet.

lib/python/voltcli/checkstats.py
```diff
@@ -351,15 +345,7 @@ def check_command_log(runner):
...
-
-    timeout = -1;
-    if runner.opts.timeout:
-        timeout = (runner.opts.timeout) * 60
-
-    #no timeout check
-    if timeout == -1:
-        return currentTime
-
+    timeout = (runner.opts.timeout) * 60
```
xinjiacs: What if the user does not specify the timeout option? Previously you handled this situation.

lib/python/voltcli/voltadmin.d/pause.py
```diff
@@ -25,10 +25,13 @@
         VOLT.BooleanOption('-w', '--wait', 'waiting',
                            'wait for all DR and Export transactions to be externally processed',
                            default = False),
-        VOLT.IntegerOption('-t', '--timeout', 'timeout', 'The timeout value in minute if @Statistics is not progressing.', default = 2),
+        VOLT.IntegerOption('-t', '--timeout', 'timeout', 'The timeout value in minutes if @Statistics is not progressing.', default = 2),
     )
 )
 def pause(runner):
+    if runner.opts.timeout and runner.opts.timeout < 0:
+        runner.abort_with_help('You cannot specify timeout less than zero minutes.')
```
xinjiacs: Maybe better with "Timeout value less than zero is invalid."

Does it make sense to have a 0 value timeout? Do we want to give the unit in SECONDS?

zeemanhe: updated with seconds, default=120.

lib/python/voltcli/checkstats.py
```diff
@@ -70,23 +76,45 @@ def check_dr_producer(runner):
...
+    partitionMinProgressed = cmp(lastPartitionMin, currentPartitionMin)
+    partitionMaxprogressed = cmp(lastPartitionMax, currentPartitionMax)
+    #stats moved
+    if partitionMinProgressed <> 0 or partitionMaxprogressed <> 0:
+        return currentTime
+
+    timeSinceLastUpdate = currentTime - lastUpdatedTime
+    #stats timeout
+    if timeSinceLastUpdate > timeout:
+         msg = "The cluster has not drained any transactions for DRPRODUCER in last %d minutes. There are outstanding transactions."
```
xinjiacs: @zeemanhe Please also update the message with seconds. And let Andrew know to document this new timeout shutdown stuff.

lib/python/voltcli/checkstats.py
```diff
@@ -269,7 +335,41 @@ def check_command_log(runner):
...
+    #stats progressed, update lastUpdatedTime
+    if statsProgressed:
+        return currentTime
+
+    #stats has not made any progress since last check
+    timeSinceLastUpdate = currentTime - lastUpdatedTime
+
+    #stats timeout
+    if timeSinceLastUpdate > timeout:
+         msg = "The cluster has not drained any transactions for %s in last %d minutes. There are outstanding transactions."
```
xinjiacs: Also here.

PR #4101, opened 2016-12-01
Eng 11442

luyangco:
Integration branch of partition group, features include:
1) Replace ClusterConfig with AbstractTopology, which is faster in non-ideal configuration
2) Expandable PicoNetwork, nodes within the same partition group can add connections between them.

The pro side PR is https://github.com/VoltDB/pro/pull/1603


src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1438,28 +1549,42 @@ public JoinAcceptor getAcceptor() {
...
-            ForeignHost fh = m_foreignHosts.get(hostId);
-            if (fh != null && fh.isUp()) {
-                fh.sendPoisonPill(err, ForeignHost.CRASH_SPECIFIED);
+            Iterator<ForeignHost> it = m_foreignHosts.get(hostId).iterator();
+            // No need to overdose the poison pill
+            if (it.hasNext()) {
+                ForeignHost fh = it.next();
+                if (fh.isUp()) {
+                    fh.sendPoisonPill(err, ForeignHost.CRASH_SPECIFIED);
+                }
```
zeemanhe: If hostIds contain local host id and local host is first poisoned,  is it possible that other poison pills do not reach destinations? just curious.

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -577,6 +606,206 @@ else if (!remoteAcceptsLocalVersion) {
...
+                }
+                try {
+                    Thread.sleep(250); //  milliseconds
+                }
+                catch (InterruptedException dontcare) {}
+            }
+            ++connectAttempts;
+        }
+        return socket;
+    }
```
zeemanhe: if not ConnectStrategy.PROBE, it may not get out of the retry loop.  Any risk associated with it?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1639,48 +1640,42 @@ else if (!startAction.doesRejoin()) {
...
+        int localHostId = m_messenger.getHostId();
+        Set<Integer> peers = Sets.newHashSet();
+        if (m_configuredReplicationFactor > 0 && partitionGroupCount > 1) {
+            Set<Integer> buddyHostIds = m_cartographer.getBuddyHostIds(localHostId);
+            if (isRejoin) {
+                peers.addAll(buddyHostIds);
+            } else {
+                for (Integer host : buddyHostIds) {
+                    if (host > localHostId) {
+                        peers.add(host);
```
zeemanhe: can you add comments why host ids of peers must be larger than local host id?

src/frontend/org/voltcore/messaging/SocketJoiner.java
luyangco: This is expected behavior, the node keeps retrying until it connects to the leader, otherwise the cluster won't start.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1639,48 +1640,42 @@ else if (!startAction.doesRejoin()) {
...
-        try
-        {
-            byte[] data = m_messenger.getZK().getData(VoltZK.topology, false, null);
-            topo = new JSONObject(new String(data, "UTF-8"));
-        }
-        catch (Exception e)
-        {
-            VoltDB.crashLocalVoltDB("Unable to read topology from ZK, dying",
-                    true, e);
+            m_messenger.createAuxiliaryConnections(peers, secondaryConnections);
```
luyangco: Good point. I think losing node is OK. When cluster loses node, remaining nodes close connection to the dead node. On the other hand the expanding case might have problem, I don't think clustersetting will get updated after node joinning, I'll check on that.

src/frontend/org/voltdb/RealVoltDB.java
luyangco: Sure, the idea here is since connection is bidirectional, I don't want every node create connections to all its buddies. Only allowing node connects to nodes that have a higher hostId could avoid creating more-than-expected connections.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1223,4 +1303,73 @@ private static void assignLeadersToPartitionsThatNeedThem(
...
+            int hostId = e.getKey();
+            int sitesCount = e.getValue();
+            hosts[i++] = new HostDescription(hostId, sitesCount, hostGroups.get(hostId));
+        }
+        // partition descriptions
+        int totalSites = 0;
+        for (Map.Entry<Integer, Integer> entry : sitesPerHostMap.entrySet()) {
+            totalSites += entry.getValue();
+        }
+        int partitionCount = totalSites / (kfactor + 1);
```
xinjiacs: Does this means that we still assume the same K for each partition? This assumption applies for INIT, Rejoin, Elastic Join?

src/frontend/org/voltdb/iv2/Cartographer.java
```diff
@@ -292,6 +294,32 @@ public int getPartitionCount()
     }
 
     /**
+     * Convenient method, given a hostId, return the hostId of its buddies (including itself) which both
+     * belong to the same partition group.
+     * @param hostId
+     * @return A set of host IDs that both belong to the same partition group
+     */
+    public Set<Integer> getBuddyHostIds(int hostId) {
```
xinjiacs: Use PartitionGroup instead of Buddy?

src/frontend/org/voltdb/AbstractTopology.java
zeemanhe: AbstractTopology supports varied Ks per partition. But support for  init, rejoin and elastic join with varied Ks has not been done yet.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1494,17 +1617,47 @@ public void setDeadHostTimeout(int timeout) {
...
+            while (it.hasNext()) {
+                ForeignHost fh = it.next();
                 fh.cutLink();
             }
         }
     }
 
+    // Create connections to nodes within the same partition group
+    public void createAuxiliaryConnections(Set<Integer> peers, int secondaryConnections) {
+        m_numberOfConnections = secondaryConnections + 1;
```
xinjiacs: @luyangco Better to update m_numberOfConnections at the end of this function when all secondaryConnections are successfully created, in case of exceptions happened.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1659,48 +1658,43 @@ else if (!startAction.doesRejoin()) {
...
-                    true, e);
-        }
+    private void createSecondaryConnections(boolean isRejoin) {
+        int partitionGroupCount = m_clusterSettings.get().hostcount() / (m_configuredReplicationFactor + 1);
+        int localHostId = m_messenger.getHostId();
+        Set<Integer> peers = Sets.newHashSet();
+        if (m_configuredReplicationFactor > 0 && partitionGroupCount > 1) {
+            Set<Integer> buddyHostIds = m_cartographer.getBuddyHostIds(localHostId);
+            if (isRejoin) {
+                peers.addAll(buddyHostIds);
```
xinjiacs: @luyangco I think if you remove the m_localHostId here, you will not need to do IF CHECK "if (hostId == m_localHostId) " in createAuxiliaryConnections() function.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1659,48 +1658,43 @@ else if (!startAction.doesRejoin()) {
...
+            } else {
+                for (Integer host : buddyHostIds) {
+                    // Connection is bidirectional, so create connections on one side is enough
+                    if (host > localHostId) {
```
xinjiacs: @luyangco Can you explain more/add comments about this IF check? Why it's one side then?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1659,48 +1658,43 @@ else if (!startAction.doesRejoin()) {
...
+                    }
+                }
+            }
+            // Basic goal is each host should has the same number of connections compare to the number
+            // without partition group layout.
+            int maxConnection = m_clusterSettings.get().hostcount() - 1;
+            int existingConnections = buddyHostIds.size() - 1;
+            int numberOfConnections = Math.min( maxConnection, CoreUtils.availableProcessors() / 4);
+            // exclude primary connection, round up the result.
+            int secondaryConnections = (numberOfConnections - 1) / existingConnections;
```
xinjiacs: Can you explain or add comments about what is the magic number 4? And how does the division for "secondaryConnections" help?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1612,36 +1613,34 @@ public void run() {
...
             try {
-                topo =
-                    new JSONObject(new String(m_messenger.getZK().getData(VoltZK.topology, false, stat), "UTF-8"));
+                return AbstractTopology.topologyFromJSON(topoJson);
+            } catch (JSONException e) {
+                VoltDB.crashLocalVoltDB("Unable to get topology from Json object", true, e);
             }
-            catch (Exception e) {
-                VoltDB.crashLocalVoltDB("Unable to get topology from ZK", true, e);
+        } else if (!startAction.doesRejoin()) {
```
xinjiacs: @zeemanhe This IF/ELSE IF/ELSE is hard to read. 
Can we do something like:
if (startAction == StartAction.JOIN) {
} else if (startAction.doesRejoin()) {
} else {
...
}

src/frontend/org/voltdb/iv2/Cartographer.java
luyangco: Sure, I'll rename all the occurrences.

src/frontend/org/voltdb/RealVoltDB.java
luyangco: Ok I updated the comment.
For the number `4`, I use the same equation for creating `VoltNetwork` thread count, check here 
https://github.com/VoltDB/voltdb/blob/master/src/frontend/org/voltcore/messaging/HostMessenger.java#L137

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1135,8 +1196,13 @@ public String getHostnameForHostID(int hostId) {
         if (hostId == m_localHostId) {
             return CoreUtils.getHostnameOrAddress();
         }
-        ForeignHost fh = m_foreignHosts.get(hostId);
-        return fh == null ? "UNKNOWN" : fh.hostname();
+        Iterator<ForeignHost> it = m_foreignHosts.get(hostId).iterator();
+        if (it.hasNext()) {
+            ForeignHost fh = it.next();
+            return fh.hostname();
+        } else {
```
xinjiacs: @luyangco You can remove the ELSE and return "UNKNOWN".

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1164,20 +1230,33 @@ ForeignHost presend(long hsId, VoltMessage message)
...
         }
-
-        if (!fhost.isUp())
-        {
+        ForeignHost fhost = null;
+        if (CoreUtils.getSiteIdFromHSId(hsId) < 0 ) {
+            // special mailbox, always use primary connection
+            fhost = getPrimary(fhosts, hostId);
+        } else {
+            if (fhosts.size() != m_numberOfConnections) {
```
xinjiacs: What's the case here? Can you explain and add comments.

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -577,6 +606,206 @@ else if (!remoteAcceptsLocalVersion) {
...
+        SocketChannel socket = connectToHost(hostAddr);
+        /*
+         * Get the clock skew value
+         */
+        ByteBuffer currentTimeBuf = ByteBuffer.allocate(8);
+        while (currentTimeBuf.hasRemaining()) {
+            socket.read(currentTimeBuf);
+        }
+        currentTimeBuf.flip();
+        long skew = System.currentTimeMillis() - currentTimeBuf.getLong();
```
xinjiacs: @luyangco "skew" never used?

zeemanhe:
AbstractTopology should have taken care of rack aware partition assignments at both startup and rejoin.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1244,6 +1243,12 @@ ForeignHost presend(long hsId, VoltMessage message)
             // special mailbox, always use primary connection
             fhost = getPrimary(fhosts, hostId);
         } else {
+            // Because the secondary connections are created rather late, just after cluster mesh network has
+            // established, but before the whole cluster has been initialized. It's possible that some non-transactional
+            // iv2 messages (agreementSiteMessag, heartBeatMessage etc.) to be sent through foreign host when
+            // there is only one primary connection, So in case of binding all sites to the primary connection,
+            // this check has been added to prevent it.
+            // FIXME: Maybe I am wrong, do all those non-transactional messages sent to well-known siteIds < 0?
```
xinjiacs: @luyangco My assumption is that "fhosts.size() != m_numberOfConnections" will never evaluate to be TRUE. I think all those non-transactional messages sent to well-known siteIds < 0.

Any non-negative site id will be handled before. agreementSiteMessag, heartBeatMessage are special messages sent to non-negative site id.

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -334,28 +333,18 @@ public void run() {
                         final InputHandler handler = authenticate(m_socket, timeoutRef);
                         if (handler != null) {
                             m_socket.configureBlocking(false);
-                            if (handler instanceof ClientInputHandler) {
```
xinjiacs: In this case, it's better to change the "handler" type and the return type of `authenticate ()` to ClientInputHandler.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1494,17 +1622,44 @@ public void setDeadHostTimeout(int timeout) {
...
 
+    // Create connections to nodes within the same partition group
+    public void createAuxiliaryConnections(Set<Integer> peers, int secondaryConnections) {
+        for (int hostId : peers) {
+            for (int ii = 0; ii < secondaryConnections; ii++) {
+                Iterator<ForeignHost> it = m_foreignHosts.get(hostId).iterator();
+                if (it.hasNext()) {
+                    ForeignHost fh = it.next();
+                    try {
+                        SocketChannel socket = m_joiner.requestForConnection(fh.m_listeningAddress);
```
zeemanhe: @luyangco Clock skews are not checked for these connections.  I would believe the check is needed.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1677,10 +1682,14 @@ private void createSecondaryConnections(boolean isRejoin) {
...
-            // exclude primary connection, round up the result.
-            int secondaryConnections = (numberOfConnections - 1) / existingConnections;
+            int existingConnections = hostIdsWithinGroup.size() - 1;
+            int totalNumberOfConnections = Math.min( maxConnection, CoreUtils.availableProcessors() / 4);
+            // (totalNumberOfConnections - existingConnections) is the the total number of secondary
+            // connections we try to create, I want the secondary connections to have an even
+            // distribution on all existing connections, and round up the result because this is
+            // integer division, there is a trick to do this:  (a + (b - 1)) / b
+            // so it becomes (totalNumberOfConnections - existingConnections) + (existingConnections - 1)
+            int secondaryConnections = (totalNumberOfConnections - 1) / existingConnections;
```
xinjiacs: @luyangco I suggest to update the variable names as:
// All the number are per node basis, PG is short for Partition Group
maxConnection - > connectionsWithoutPG
existingConnections -> existingConnectionsWithinPG
totalNumberOfConnections -> targetConnectionsWithinPG
secondaryConnections -> secondaryConnectionsWithinPG

Because we use " CoreUtils.availableProcessors() / 4", is it possible to have different secondaryConnectionsWithinPG for different Partition Group?

src/frontend/org/voltdb/ClientInterface.java
luyangco: Good idea.

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -577,6 +606,206 @@ else if (!remoteAcceptsLocalVersion) {
...
+        /*
+         * Get the clock skew value
+         */
+        ByteBuffer currentTimeBuf = ByteBuffer.allocate(8);
+        while (currentTimeBuf.hasRemaining()) {
+            socket.read(currentTimeBuf);
+        }
+        currentTimeBuf.flip();
+        currentTimeBuf.getLong();
+        assert(currentTimeBuf.remaining() == 0);
```
zeemanhe: @luyangco  should the clock skew be postponed here till all the connections are created in HostMessenger.createAuxiliaryConnections() and do a clock skew check there?

src/frontend/org/voltcore/messaging/HostMessenger.java
luyangco: Clock skews and version check are done during primary connection request -- a.k.a. `requestHostId()`. I feel it's kind redundant to do it again in secondary connection request.

src/frontend/org/voltdb/RealVoltDB.java
luyangco: It's totally possible, for example, a 5-node cluster with K=1 will have two partition groups, the group with three nodes which each of them has one secondary connection, and the group with two nodes which each of them has three secondary connections.

I'll rename those variables.

src/frontend/org/voltcore/messaging/SocketJoiner.java
luyangco: I feel there is no need to do multiple clock skew checks because system time won't change dramatically between the primary connection and following secondary connection.  If the first clock skew check passes, we should let it go.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1243,22 +1243,25 @@ ForeignHost presend(long hsId, VoltMessage message)
...
-            // there is only one primary connection, So in case of binding all sites to the primary connection,
-            // this check has been added to prevent it.
-            // FIXME: Maybe I am wrong, do all those non-transactional messages sent to well-known siteIds < 0?
-            if (fhosts.size() != m_numberOfConnections) {
-                fhost = getPrimary(fhosts, hostId);
-            } else {
+            /**
+             *  Because the secondary connections are created rather late, just after cluster mesh network has
+             *  established, but before the whole cluster has been initialized. It's possible that some non-transactional
+             * iv2 messages (agreementSiteMessag, heartBeatMessage etc.) to be sent through foreign host when
```
xinjiacs: Outdated "agreementSiteMessag, heartBeatMessage etc."

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1866,22 +1866,18 @@ public  boolean createTransaction(
...
                 isShortCircuitRead = true;
             }
         }
-
-        if (initiatorHSId == null) {
-            hostLog.error("Failed to find master initiator for partition: "
-                    + Integer.toString(partition) + ". Transaction not initiated.");
-            return false;
-        }
+        Preconditions.checkArgument(initiatorHSId != null,
```
vtkstef: checkArgument does its own formatting

```java
Preconditions.checkArgument(initiatorHSId != null,
    "Failed to find master initiator for partition: %s. Transaction not initiated.", partition");
```

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1078,6 +1124,21 @@ private static int parseHostId(String name) {
...
+    public Map<Integer, String> getHostGroupsFromZK() {
+        Map<Integer, String> hostGroups = Maps.newHashMap();
+        try {
+            List<String> children = m_zk.getChildren(CoreZK.hosts, false);
+            for (String child : children) {
+                byte[] payload = m_zk.getData( ZKUtil.joinZKPath(CoreZK.hosts, child), false, new Stat());
+                final HostInfo info = HostInfo.fromBytes(payload);
+                hostGroups.put(parseHostId(child), info.m_group);
+            }
+        } catch (Exception e) {
```
vtkstef: please list the type of exception you are expecting

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -480,10 +507,12 @@ private void processSSC(ServerSocketChannel ssc) throws Exception {
             }
 
             hostLog.info("Received request type " + type);
-            if (type.equals(REQUEST_HOSTID)) {
-                m_joinHandler.requestJoin( sc, listeningAddress, jsObj);
-            } else if (type.equals(PUBLISH_HOSTID)){
+            if (type.equals(ConnectionType.REQUEST_HOSTID.toString())) {
```
vtkstef: use .name() instead

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -577,6 +606,206 @@ else if (!remoteAcceptsLocalVersion) {
...
+        }
+        currentTimeBuf.flip();
+        long skew = System.currentTimeMillis() - currentTimeBuf.getLong();
+        skews.add(skew);
+
+        VersionChecker versionChecker = m_acceptor.getVersionChecker();
+        activeVersions.add(versionChecker.getVersionString());
+
+        JSONObject jsObj = new JSONObject();
+        jsObj.put(TYPE, ConnectionType.REQUEST_HOSTID.toString());
```
vtkstef: .name()

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -577,6 +606,206 @@ else if (!remoteAcceptsLocalVersion) {
...
+        byte[] jsBytes = jsObj.toString(4).getBytes(StandardCharsets.UTF_8);
+        ByteBuffer addConnection = ByteBuffer.allocate(4 + jsBytes.length);
+        addConnection.putInt(jsBytes.length);
+        addConnection.put(jsBytes).flip();
+        while (addConnection.hasRemaining()) {
+            socket.write(addConnection);
+        }
+        // read the json response from socketjoiner with version info and validate it
+        final String remoteAddress = socket.socket().getRemoteSocketAddress().toString();
+        processJSONResponse(socket, remoteAddress, Sets.newHashSet());
```
vtkstef: why are we checking version for REQUEST_CONNECTION?

luyangco:
@vtkstef has some comments after we merge it into master, I'll address them in a separate ticket.
Thanks Stefano!

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -941,12 +940,11 @@ public void initialize(Configuration config) {
...
              * is trying to rejoin, it should rely on the cartographer's view to pick the partitions to replace.
              */
-            JSONObject topo = getTopology(config.m_startAction, hostGroups, m_joinCoordinator);
+            AbstractTopology topo = getTopology(config.m_startAction, hostGroups, m_joinCoordinator);
             m_partitionsToSitesAtStartupForExportInit = new ArrayList<>();
             try {
                 // IV2 mailbox stuff
-                ClusterConfig clusterConfig = new ClusterConfig(topo);
-                m_configuredReplicationFactor = clusterConfig.getReplicationFactor();
+                m_configuredReplicationFactor = m_catalogContext.getDeployment().getCluster().getKfactor();
```
luyangco: I'll replace it with `AbstractTopology.getReplicationFactor()`

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1494,17 +1624,44 @@ public void setDeadHostTimeout(int timeout) {
...
+                if (it.hasNext()) {
+                    ForeignHost fh = it.next();
+                    try {
+                        SocketChannel socket = m_joiner.requestForConnection(fh.m_listeningAddress);
+                        // Auxiliary connection never time out
+                        ForeignHost fhost = new ForeignHost(this, hostId, socket, Integer.MAX_VALUE,
+                                fh.m_listeningAddress, new PicoNetwork(socket));
+                        putForeignHost(hostId, fhost);
+                        fhost.enableRead(VERBOTEN_THREADS);
+                    } catch (Exception e) {
```
luyangco: A very good advice, I'll fix all occurrences.

PR #4103, opened 2016-12-01
ENG-11294: Add protobuf message types for 'start cursor' request

slowsoul:


src/protobuf/dragent.proto
```diff
@@ -49,6 +49,17 @@ message Query {
     optional bool includeMesh = 1 [default = false];
 }
 
+message ConversationInfo {
+    optional int32 clusterId = 1;
+    optional fixed64 clusterCreationId = 2;
+    optional int32 clusterProtocolVersion = 3;
+    repeated NodeInfo nodeInfo = 4;
+}
```
nshi: This seems to largely duplicate the `ClusterInfo` message.

src/protobuf/dragent.proto
```diff
@@ -109,11 +113,13 @@ message ClusterInfo {
...
     required fixed64 creationId = 2;
 
-    optional fixed64 recoverTime = 3;
+    optional fixed64 recoveryId = 3;
     optional int32 globalPartitionCount = 4;
 
     repeated NodeInfo nodeInfo = 5;
     repeated PartitionInfo partitionInfo = 6;
+
+    optional int32 protocolVersion = 7;
```
nshi: You can actually move this up. It doesn't matter if the indices of the fields change now because we haven't released yet.

PR #4105, opened 2016-12-02
ENG-11453 plan fragment execution refactor

yzhang1991:
NOT READY FOR REVIEW YET.

### DIFF
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -332,10 +332,4 @@ void ExecutorContext::checkTransactionForDR() {
     }
 }
 
-void TempTableTupleDeleter::operator()(TempTable* tbl) const {
-    if (tbl != NULL) {
-        tbl->deleteAllTempTuples();
-    }
-}
-
 } // end namespace voltdb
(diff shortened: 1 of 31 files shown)


src/ee/common/executorcontext.cpp
```diff
@@ -77,17 +77,17 @@ ExecutorContext::ExecutorContext(int64_t siteId,
                 UndoQuantum *undoQuantum,
                 Topend* topend,
                 Pool* tempStringPool,
-                NValueArray* params,
```
yzhang1991: Moving the params array from VoltDBEngine to ExecutorContext.
Now ExecutorContext actually owns this NValueArray, instead of maintaining a pointer.

wolffcm:
This seems to move things in a good direction.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -536,28 +517,41 @@ UniqueTempTableResult VoltDBEngine::executePlanFragment(ExecutorVector* executor
...
 
-    resetExecutionMetadata(executorVector);
+    m_executorContext->resetExecutionMetadata(executorVector);
 
     VOLT_DEBUG("Finished executing successfully.");
     return result;
 }
 
-void VoltDBEngine::resetExecutionMetadata(ExecutorVector* executorVector) {
+void VoltDBEngine::releaseUndoToken(int64_t undoToken)
```
wolffcm: I'm never sure if methods like this were made inline intentionally for performance, or just out of habit.  I think it makes sense to be in the .cpp file as you've done here.

PR #4107, opened 2016-12-02
Eng 10091 grammar gen squash8

johnTcrawford:


tests/sqlgrammar/DDL.sql
```diff
@@ -0,0 +1,108 @@
...
+  FROM R1 WHERE ID > 5
+  GROUP BY BIG;
+
+CREATE VIEW VP2 (SMALL, TINY, TOTAL_COUNT, ID_SUM, BIG_SUM) AS 
+  SELECT SMALL, TINY, COUNT(*), SUM(ID), SUM(BIG)
+  FROM P2
+  GROUP BY SMALL, TINY;
+
+CREATE VIEW VR2 (SMALL_ABS, TINY, TOTAL_COUNT, ID_SUM, BIG_SUM) AS 
+  SELECT ABS(SMALL), TINY, COUNT(*), SUM(ID), SUM(BIG)  FROM R2
```
wolffcm: You could have a view with a join here now, I don't think it was supported in your first pass.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -0,0 +1,507 @@
...
+
+        # Parse a symbol name and its definition, out of the current grammar rule
+        grammar_defn = __SYMBOL_DEFN.search(grammar_rule)
+        if not grammar_defn:
+            print 'ERROR: Unrecognized grammar rule, ignored:\n', grammar_rule
+            continue
+        symbol_name = grammar_defn.group('symbolname').strip()
+        definition  = grammar_defn.group('definition').strip()
+
+        weighted_xor = re.search(__WEIGHTED_XOR, definition)
```
BillWhite: Where is this variable used?

tests/sqlgrammar/sql-ddl-grammar.txt
```diff
@@ -0,0 +1,127 @@
...
+index-definition        ::= {index-type} ({column-list})
+limit-definition        ::= LIMIT PARTITION ROWS {non-negative-int-value} [EXECUTE {delete-statement}]
+index-or-limit-defn     ::= {index-definition} | {limit-definition}
+
+################################################################################
+# Grammar rules for CREATE VIEW statements:
+################################################################################
+create-view-statement   ::= CREATE VIEW {view-name} ({column-name-list}) AS \
+                            SELECT {column-expression} [[AS] {column-alias}] \
+                            FROM {table-name} \
```
wolffcm: You could also have a rule for generating multi-table views here.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -0,0 +1,507 @@
...
+        weights = []
+        for wxor in __WEIGHTED_XOR.finditer(definition):
+            weight = wxor.group('weight')
+            if weight:
+                weights.append(int(weight))
+                definition = definition.replace(weight+'|', '|', 1)
+            else:
```
BillWhite: What if the length of weight is empty in the grammar.  Don't you loop then, replacing "|" with "|".

I haven't tried this, I just was thinking.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -0,0 +1,507 @@
...
+            # When there are are no alternative options (i.e., no '|'),
+            # the entry for the current symbol name is just a string
+            grammar[symbol_name] = definition.strip()
+        elif all(w is 1 for w in weights):
+            # When all weights are 1 (equally weighted), the entry is a List
+            grammar[symbol_name] = [options[i].strip() for i in range(len(options))]
+        else:
+            # When weights differ, the entry is a Dictionary, including each weight
+            grammar[symbol_name] = dict([ (options[i].strip(), weights[i]) for i in range(len(options)) ])
+
```
BillWhite: It seems to me that these three cases are really the same, and it would be easier to represent them with one dictionary.  Isn't it a PITA to have three different ways to represent a RHS?

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -0,0 +1,507 @@
...
+        definition  = grammar.get(symbol_name)
+        #print 'DEBUG: bracketed_name:', str(bracketed_name)
+        #print 'DEBUG: symbol_name :', str(symbol_name)
+        #print 'DEBUG: definition:', str(definition)
+        if not definition:
+            print "ERROR: Could not find symbol_name '" + str(symbol_name) + "' in grammar dictionary!!!"
+            break
+        # Check how deep into a recursive definition we're going
+        if depth.get(symbol_name):
+            depth[symbol_name] += 1
```
BillWhite: I think there's a python function "depth.get(symbol_name, 0)" which sets depth[symbol_name] to 0 if it's not defined.  Or maybe it just returns 0 if symbol_name is not defined.  In any case, that might simplify away this conditional.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -0,0 +1,507 @@
...
+            if (depth[symbol_name] > max_depth and bracketed_name in definition[random_index] and
+                    any(bracketed_name not in definition[i] for i in range(len(definition)) ) ):
+                while bracketed_name in definition[random_index]:
+                    random_index = randrange(0, len(definition))
+                    #print 'DEBUG: random_index:', str(random_index)
+            definition = definition[random_index]
+        elif isinstance(definition, dict):
+            random_selection = randrange(0, sum(definition.values()))
+            # TODO: combine the max depth stuff above with the weights thing below ?!?
+            sum_weights = 0
```
BillWhite: This is what I was saying.  You have lists and dictionaries all representing the same thing.
So, you have two cases, one of which is not implemented here.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -0,0 +1,507 @@
...
+            for key in definition:
+                sum_weights += definition[key]
+                if random_selection < sum_weights:
+                    definition = key
+                    break
+        #print 'DEBUG: definition:', definition
+        # Check for any optional text [in brackets], and decide whether to include it or not
+        optional = __OPTIONAL.search(definition)
```
BillWhite: FWIW, people usually represent an optional rule:
 A ::= b [ c ] d;

by

A ::= b g0001 d;
g0001 ::= /* empty */ | c ;

where g0001 is a generated non-terminal symbol.  That way you don't need a special case here.

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: It's not; that logic got moved to the for loop definition just below; I've now deleted it.

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: No, if weight has length zero, the if condition fails, so you're in the else clause instead.

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: Yes, you're right. I don't want to take the time to fix this now, but I'll add it to my list of improvements (with Chris's suggestions above).

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: Thanks, fixed (or rather, simplified).

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: Not quite. I have 3 cases: the "definition" variable represents a string (simple defn. with no options), a list (equally weighted options), or a dict (options with different weights). All 3 are implemented here, but only the "list" option uses the max_depth code to avoid recursive definitions going too deep. This max_depth stuff complicates unifying all 3 cases, which is why (as I said above) I'm not dealing with this now.

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: Interesting. I'll add this simplification to my to-do list, as well.

johnTcrawford:
Thanks, Chris & Bill, for your comments. I have (as noted) fixed a couple of Bill's smaller suggestions; his larger, and Chris's, suggestions I'm adding to my to-do list (which I'll put in a Jira ticket and/or deal with during Innovation Week). I'll run it in Jenkins a couple more times tonight, and then merge it to master.

PR #4109, opened 2016-12-02
ENG-11447: Refactor DR producer initialization sequence.

nshi:
- Instead of creating the DR gateways when the initiators are
  constructed, which is way early than when the DR producer is ready to
  do anything, I moved them out of the initiator constructor and create
  them when the DR producer is ready.

- The SpScheduler really doesn't care about the DR gateway. Only the
  transactions care about it when they are run on the site. So I removed
  the reference to the DR gateway from SpScheduler and created a getter
  in the Site so that transactions can have access to it.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -1121,8 +1121,6 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
                                     new VoltFile(VoltDB.instance().getSnapshotPath()),
                                     m_replicationActive.get(),
                                     m_configuredNumberOfPartitions,m_catalogContext.getClusterSettings().hostcount());
-                    m_producerDRGateway.start();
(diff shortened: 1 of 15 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3565,9 +3556,17 @@ public ListeningExecutorService getComputationService() {
     private void prepareReplication() {
         try {
             if (m_producerDRGateway != null) {
-                m_producerDRGateway.initialize(m_catalogContext.cluster.getDrproducerenabled(),
-                        VoltDB.getReplicationPort(m_catalogContext.cluster.getDrproducerport()),
-                        VoltDB.getDefaultReplicationInterface());
+                m_producerDRGateway.startAndWaitForGlobalAgreement();
```
nshi: Thanks for catching this! I missed that part. Maybe what I'll do is to cache the truncation information and use it when all the streams are created.

nshi: I addressed this comment in the PR in pro.

PR #4113, opened 2016-12-08
ENG-11573 populate joined table view on snapshot restore

yzhang1991:
When tuples are inserted into a base table of a multi-table view, the normal code path is:
`PersistentTable::insertTuple()` => `PersistentTable::insertPersistentTuple()` => `PersistentTable::insertTupleCommon()`

The function call to insert the new tuple into the delta table is in `PersistentTable::insertPersistentTuple()`. 

However, on snapshot restore, `PersistentTable::processLoadedTuple()` directly call `PersistentTable::insertTupleCommon()`, so the new tuple is not inserted into the delta table for view mantainence.

This pull request fixed the bug by adding a `insertTupleIntoDeltaTable()` call in `PersistentTable::processLoadedTuple()`.

The regression test will be added to the pro repository.

### DIFF
--- a/src/ee/storage/MaterializedViewHandler.cpp
+++ b/src/ee/storage/MaterializedViewHandler.cpp
@@ -166,9 +166,9 @@ namespace voltdb {
     // to catch up with the existing data.
     //TODO: *non-grouped views could instead set up a hard-coded initial
     // row as they do in the single-table case to avoid querying empty tables.
-    void MaterializedViewHandler::catchUpWithExistingData(VoltDBEngine *engine, bool fallible) {
+    void MaterializedViewHandler::catchUpWithExistingData(bool fallible) {
         ExecutorContext* ec = ExecutorContext::getExecutorContext();
-        UniqueTempTableResult viewContent = engine->executePlanFragment(m_createQueryExecutorVector.get());
+        UniqueTempTableResult viewContent = ec->getEngine()->executePlanFragment(m_createQueryExecutorVector.get());
         TableIterator ti = viewContent->iterator();
         TableTuple tuple(viewContent->schema());
         while (ti.next(tuple)) {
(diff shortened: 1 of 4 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -1364,6 +1364,7 @@ void PersistentTable::processLoadedTuple(TableTuple &tuple,
                                          size_t &tupleCountPosition,
                                          bool shouldDRStreamRows) {
     try {
+        insertTupleIntoDeltaTable(tuple, true);
```
wolffcm: I wonder if the call to `insertIntoDeltaTable` should go into `insertTupleCommon`, since it is really something that should always happen when inserting any tuple into a persistent table?

src/ee/storage/MaterializedViewHandler.cpp
```diff
@@ -166,9 +166,9 @@ namespace voltdb {
     // to catch up with the existing data.
     //TODO: *non-grouped views could instead set up a hard-coded initial
     // row as they do in the single-table case to avoid querying empty tables.
-    void MaterializedViewHandler::catchUpWithExistingData(VoltDBEngine *engine, bool fallible) {
+    void MaterializedViewHandler::catchUpWithExistingData(bool fallible) {
         ExecutorContext* ec = ExecutorContext::getExecutorContext();
-        UniqueTempTableResult viewContent = engine->executePlanFragment(m_createQueryExecutorVector.get());
+        UniqueTempTableResult viewContent = ec->getEngine()->executePlanFragment(m_createQueryExecutorVector.get());
```
wolffcm: I think this will invoke a `pthread_` routine to get the thread local engine.  I wonder if calls like that are expensive.  I really have no idea.

src/ee/storage/persistenttable.cpp
yzhang1991: I was hoping to avoid recursive call because insertTupleIntoDeltaTable will call another insertTupleCommon, but it seems to be OK now. I will put it into insertTupleCommon.

yzhang1991:
So I moved insertTupleIntoDeltaTable into insertTupleCommon, and it passed TestMaterializedViewSuite and memcheck on my server. Will make the final merge after Jenkins is blue.

PR #4126, opened 2016-12-16
ENG-11616 delay param between node startup

zeemanhe:
Add an optional delay between the node startup to help matching the host ids cached on LocalCluster to those in real cluster.


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -119,6 +119,12 @@ public void setHasLocalServer(boolean hasLocalServer) {
     ArrayList<CommandLine> m_cmdLines = null;
     ServerThread m_localServer = null;
     ProcessBuilder m_procBuilder;
+
+    //wait before next node is started up in millisecond
+    //to help matching the host id on the real cluster with the host id on the local
+    //cluster
+    private long m_deplayBetweenNodeStartup = 0;
```
luyangco: It's better to put `ms` in the variable name, so others know this is for millisecond.

PR #4127, opened 2016-12-19
Make SQLUpdateSuite 10x faster

rmorgenstein:
10x faster by only bringing up the server once.  This sacrifices test isolation for speed and may not
be what we want to do everywhere, but this particular suite makes calls to a single method, executeAndTestUpdate(), so changes needed for between-testcase cleanup are in that test method.
- On Jenkins this went from 4+ minutes to 25 seconds
- All SQL is in 1 testcase per backend
- Added some println to show what is happening
- Added more info to assert to show failed query
- Added TRUNCATE TABLE at the end of the executeAndTestUpdate call


wolffcm:
This looks good.  There's probably other tests there this approach would work too.

PR #4131, opened 2016-12-19
ENG-11580 try harder to complete shutdown snapshot

vtkstef:


akhanzode:
Looks good merge when jenkins is good.

PR #4133, opened 2016-12-20
Eng 10139 more aggs

BillWhite:


build.py
```diff
@@ -161,16 +161,17 @@
 else:
     LOG_LEVEL = "500"
 
+CTX.LOG_LEVEL = LOG_LEVEL
```
BillWhite: This change is so that we can change the log level just for building the EE.

src/ee/expressions/dateconstants.h
```diff
@@ -0,0 +1,35 @@
+/* This file is part of VoltDB.
```
BillWhite: I needed some date constants but not all of the date functions.  So I split
these off into their own include file.

src/frontend/org/voltdb/expressions/WindowFunctionExpression.java
```diff
@@ -129,14 +130,16 @@ public int hashCode() {
...
-        hash += m_aggArguments.hashCode();
+        hash += m_args.hashCode();
         return hash;
     }
 
     @Override
     public void finalizeValueTypes() {
-        m_valueType = VoltType.BIGINT;
-        m_valueSize = VoltType.BIGINT.getLengthInBytesForFixedTypes();
+        // This is a kind of logical aggregate
```
BillWhite: There is a lot of common functionality in AggregateExpression and WindowFunctionExpression.
So I made these both call the same function to finalize their types.

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
```diff
@@ -474,11 +474,7 @@ protected String explainPlanForNode(String indent) {
         for (ExpressionType e : m_aggregateTypes) {
             sb.append(sep).append(e.symbol());
             sep = ", ";
-            if (e == ExpressionType.AGGREGATE_WINDOWED_RANK || e == ExpressionType.AGGREGATE_WINDOWED_DENSE_RANK) {
-                sb.append("()");
-            }
-            else if (e != ExpressionType.AGGREGATE_COUNT_STAR
-                    && e != ExpressionType.AGGREGATE_WINDOWED_COUNT) {
+            if (e != ExpressionType.AGGREGATE_COUNT_STAR) {
```
BillWhite: We don't use windowed aggregates in AggregatePlanNode any more.  This should have been removed quite a few weeks ago.

tests/frontend/org/voltdb/planner/EECountPlanTestGenerator.java
```diff
@@ -1,167 +0,0 @@
-/* This file is part of VoltDB.
```
BillWhite: All the test plan generators are unified in one java file.

tests/frontend/org/voltdb/planner/EEPlanTestGenerator.java
```diff
@@ -169,6 +169,632 @@ public void testGeneratedPlan() throws Exception {
         generateTests("executors", "TestGeneratedPlans", db);
     }
 
+    public void testGeneratedMaxPlan() throws Exception {
+        TableConfig TConfig = new TableConfig("T",
```
BillWhite: All of the test plan generators are unified in this source file.  The other test plan generator source files are deleted.

src/ee/executors/windowfunctionexecutor.cpp
```diff
@@ -257,16 +334,33 @@ class MaxAgg : public WindowAggregate {
     virtual void lookaheadOneRow(TableWindow &window, NValueArray &argVals) {
         assert(argVals.size() == 1);
         if ( ! argVals[0].isNull()) {
-            if (argVals[0].op_greaterThan(m_value)) {
+            if (m_isEmpty || argVals[0].op_greaterThan(m_value).isTrue()) {
                 m_value = argVals[0];
+                if (m_value.getSourceInlined()) {
+                    m_value.allocateObjectFromInlinedValue(&m_pool);
```
wolffcm: Glad you added this code.  Forgetting to add this has been a source of bugs.  I wish there was a mechanism that would do this automatically, because it's so easy to forget.

src/ee/executors/windowfunctionexecutor.cpp
```diff
@@ -280,28 +374,33 @@ class SumAgg : public WindowAggregate {
     }
 };
 
+#if  0
```
wolffcm: Is there value to having code that is never compiled in here?  Personally I think it can be misleading (like when searching for uses of a symbol, this may make it look like something is live when it is not).

src/ee/expressions/dateconstants.h
```diff
@@ -0,0 +1,35 @@
...
+#include "boost/date_time/gregorian/gregorian.hpp"
+
+/*
+ * Some date and time constants needed in datafunctions
+ * and also other places.
+ */
+static const boost::posix_time::ptime EPOCH(boost::gregorian::date(1970,1,1));
+static const int64_t GREGORIAN_EPOCH = -12212553600000000;  //  1583-01-01 00:00:00
+static const int64_t NYE9999         = 253402300799999999;  //  9999-12-31 23:59:59.999999
+#endif /* defined(SRC_EE_PLANNODES_WINDOWFUNCTIONNODE_H_) */
```
wolffcm: Minor: the symbol in this comment doesn't match the `#define` symbol.

src/frontend/org/voltdb/expressions/AbstractExpression.java
```diff
@@ -1471,4 +1471,18 @@ public static void toJSONArray(JSONStringer stringer, String keyString, List<Abs
         }
         stringer.endArray();
     }
+
+    /**
+     * Ferret out the first argument.  This can be m_left or else
+     * the first element of m_args.
+     */
+    public AbstractExpression getFirstArgument() {
+        if (m_left != null) {
```
wolffcm: Can we have values in both `m_left` and `m_args`?  If not, is there a debug assert somewhere?

src/frontend/org/voltdb/expressions/AggregateExpression.java
```diff
@@ -58,60 +58,74 @@ public int hashCode() {
...
         case AGGREGATE_VALS_TO_HYPERLOGLOG:
-            m_valueType = VoltType.VARBINARY;
-            m_valueSize = 65537;
+            expr.m_valueType = VoltType.VARBINARY;
+            expr.m_valueSize = 65537;
             break;
         case AGGREGATE_AVG:
         case AGGREGATE_MAX:
         case AGGREGATE_MIN:
+        case AGGREGATE_WINDOWED_MIN:
```
wolffcm: Do windowed MIN and MAX take more than one argument?  Maybe this will become clearer as I read on...

src/frontend/org/voltdb/expressions/WindowFunctionExpression.java
wolffcm: Ok.  The code reuse is good.

src/frontend/org/voltdb/plannodes/AggregatePlanNode.java
wolffcm: Glad you caught it.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserBase.java
```diff
@@ -655,6 +655,16 @@ IntervalType readIntervalType() {
...
 		int type = windowedExpressionTypeMap.get(tokenT, -1);
+		// This is an error.  But we
+		// can't actually throw a good error
+		// message here, so we will
+		// try and do the best we can.
+		// This will be picked up later if
+		// we are trying to use an unsupported
+		// aggregate type.
+		if (type == -1) {
+			type = expressionTypeMap.get(tokenT, -1);
```
wolffcm: Is the body of the following if-statement dead code then?

wolffcm: Also, it's not clear to me why you can't produce a good error message here?

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuite.java
```diff
@@ -0,0 +1,901 @@
...
+import org.voltdb.BackendTarget;
+import org.voltdb.VoltTable;
+import org.voltdb.client.Client;
+import org.voltdb.client.ClientResponse;
+import org.voltdb.client.NoConnectionsException;
+import org.voltdb.client.ProcCallException;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.types.TimestampType;
+
+public class TestWindowFunctionSuite extends RegressionSuite {
```
wolffcm: This is a totally new file?  Didn't you add these tests already?

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuite.java
```diff
@@ -0,0 +1,901 @@
...
+            assertTrue(success);
+            builder.addServerConfig(config);
+        }
+        catch (IOException excp) {
+            fail();
+        }
+
+        return builder;
+    }
+}
```
wolffcm: I guess what I'm getting at with my previous comment, is should I be reviewing this 900-line file, or has it already been reviewed?

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
```diff
@@ -0,0 +1,1687 @@
...
+                + "  B INTEGER NOT NULL,"
+                + "  C VARCHAR "
+                + ");"
+                + "PARTITION TABLE T_STRING_C_NULL_PB ON COLUMN B;"
+
+                // Nothing nullable, partition on C
+                +"CREATE TABLE T_STRING_PC ("
+                + "  A INTEGER NOT NULL,"
+                + "  B INTEGER NOT NULL,"
+                + "  C VARCHAR NOT NULL"
```
wolffcm: Is there value to testing on tables that are the same except for NOT NULL constraints on some columns?  As long as you are testing null and non-null values, it seems like that's enough.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
```diff
@@ -0,0 +1,1687 @@
...
+        cr = client.callProcedure(insertProcName,  2,  3,    null);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure(insertProcName,  2,  3,    null);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure(insertProcName,  2,  3,    null);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+    }
+
+    private void validateStringTable(VoltTable voltTable, Object[][] expect) {
+        assertEquals(expect.length, voltTable.getRowCount());
```
wolffcm: We have various very similar code in the RegressionSuites base class.  You could use it here maybe?

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
```diff
@@ -0,0 +1,1687 @@
...
+                + "PARTITION TABLE P2 ON COLUMN TINY;"
+
+                + "CREATE TABLE P1_ENG_11029 ("
+                + "        ID INTEGER NOT NULL,"
+                + "        TINY TINYINT NOT NULL,"
+                + "        SMALL SMALLINT NOT NULL,"
+                + "        BIG BIGINT NOT NULL,"
+                + "        PRIMARY KEY (ID)"
+                + ");"
+                + "PARTITION TABLE P1_ENG_11029 ON COLUMN ID;"
```
wolffcm: No tables with inlined VARCHAR or VARBINARY fields?  It's pretty important to test that.  They especially need to be tested as arguments to MAX and MIN.  (This is the comment that will make we request changes.)

src/frontend/org/voltdb/expressions/AggregateExpression.java
wolffcm: My understanding is now that the regular aggregrates have their argument in m_left, but that windowed aggregate have their argument in m_args?  Is that right, and if so why are they different?

BillWhite: I just wanted to put them in an array rather than two separate variables.

src/ee/executors/windowfunctionexecutor.cpp
BillWhite: I thought it was useful from before.  But I removed it.

src/frontend/org/voltdb/expressions/AbstractExpression.java
BillWhite: I don't think that's possible.  Having an assert here seems ok, but it doesn't seem like we'd get much more coverage.

I added some asserts, though.

src/frontend/org/voltdb/expressions/AggregateExpression.java
BillWhite: No, MIN and MAX take exactly one argument.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserBase.java
BillWhite: We don't have any way to map from tokenT to a name, so we can't throw an exception with the name here.  If we are given an aggregate which we don't support, we have opportunities later on to find this out and give a good error message.  For example, if we are given "SELECT AVG(ID) OVER () FROM TBL" we would have an integer representing AVG here, but no way to recover AVG from this integer.  We keep this as AVG for now.  Later on we will see this and give the error message that AVG is not a supported window function.

BillWhite: The body of the if statement here is taken if we somehow get a tokenT value which is not a windowed aggregate and is not anything we actually know as an expression type.  This would be a VoltDB error, and not a user error.

src/ee/expressions/dateconstants.h
BillWhite: Fixed.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuite.java
BillWhite: This isn't new.  I don't know what git thinks it is.  I may have renamed it from TestWindowedFunctionSuite.java to TestWindowFunctionSuite.java.  That might be it.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuite.java
BillWhite: Already reviewed.  Sorry.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
BillWhite: I don't have any varbinary tests.  But the string table tests all have short and long strings, and I take min and max of them.  So I think I get test coverage of this.  Look at makeLongString(n).  It makes n copies of the string "abcdefgh".  So, makeLongString(5) should make a 40 character string.  Isn't that long enough to make the string out-of-line?  I thought any string longer than 16 bytes is good enough.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
BillWhite: I have a specific requirement for tables with nullable strings and partitioned other columns.  I didn't see anything appropriate.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
```diff
@@ -1350,105 +1486,216 @@ public void testMax() throws Exception {
...
-                             expectStringTableSomeNull);
+                             LONG_REPL,
+                             expectLongStringTableSomeNull);
+        testStringsSomeNulls(client,
+                             "t_string_inline_c_null",
+                             "max(C)",
+                             SHORT_REPL,
+                             expectShortStringTableSomeNull);
+        testStringsSomeNulls(client,
+                             "t_string_c_null_pb",
```
wolffcm: These changes look good.  It's especially important that this test passes in memcheck mode.  If this doesn't run in Jenkins automatically it would be worth running manually.

PR #4136, opened 2016-12-20
ENG-11447: Extract DRConflictManager interface from PartitionDRGateway.

nshi:
So that the conflict manager can exist without the dependency on PDRG.

### DIFF
--- a/src/frontend/org/voltdb/PartitionDRGateway.java
+++ b/src/frontend/org/voltdb/PartitionDRGateway.java
@@ -28,6 +28,7 @@ import org.voltdb.iv2.SpScheduler.DurableUniqueIdListener;
 import org.voltdb.jni.ExecutionEngine.EventType;
 
 import com.google_voltpatches.common.collect.ImmutableMap;
+import org.voltdb.utils.MiscUtils;
 
 /**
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -182,15 +189,21 @@ public static int reportDRConflict(int partitionId, int remoteClusterId, long re
                                        ByteBuffer expectedMetaTableForDelete, ByteBuffer expectedTupleTableForDelete,
                                        int insertConflict, ByteBuffer existingMetaTableForInsert, ByteBuffer existingTupleTableForInsert,
                                        ByteBuffer newMetaTableForInsert, ByteBuffer newTupleTableForInsert) {
-        final PartitionDRGateway pdrg = m_partitionDRGateways.get(partitionId);
-        if (pdrg == null) {
-            VoltDB.crashLocalVoltDB("No PRDG when there should be", true, null);
+        if (m_conflictManager != null) {
```
slowsoul: I agree, you mean BinaryLogSink::apply() instead of pushDRBuffer(), right?

PR #4138, opened 2016-12-20
ENG-11559: Fix multiply overflow in REPEAT output length check

wolffcm:


BillWhite:
Seems right to me.

paulmartel:
Seems right to me, too.

PR #4141, opened 2016-12-22
Eng-11644 Fix up some HSQLDB tests that were taking 2 min. to timeout the getClient() call

rmorgenstein:
Six tests were calling getClient() prior to checking isHSQL(), resulting in the tests waiting on
the getClient() timeout rather than quickly returning with a PASS for testcases that the HSQL
backend can't run.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
@@ -2900,12 +2900,13 @@ public class TestFunctionsSuite extends RegressionSuite {
     // Unicode character to UTF8 string character
     public void testChar() throws NoConnectionsException, IOException, ProcCallException {
         System.out.println("STARTING test CHAR");
-        Client client = getClient();
-        ClientResponse cr;
-        VoltTable result;
 
         // Hsql has wrong answers.
         if (isHSQL()) return;
+
+        Client client = getClient();
+        ClientResponse cr;
+        VoltTable result;
 
         cr = client.callProcedure("P1.insert", 1, "Xin@VoltDB", 1, 1.0, new Timestamp(1000000000000L));
         assertEquals(ClientResponse.SUCCESS, cr.getStatus());
--- a/tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
@@ -2902,7 +2902,9 @@ public class TestFunctionsSuite extends RegressionSuite {
         System.out.println("STARTING test CHAR");
 
         // Hsql has wrong answers.
-        if (isHSQL()) return;
+        if (isHSQL()) {
+            return;
+        }
 
         Client client = getClient();
         ClientResponse cr;
--- a/tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
@@ -25,8 +25,6 @@ package org.voltdb.regressionsuites;
 
 import java.io.IOException;
 
-import junit.framework.Test;
-
 import org.voltdb.BackendTarget;
 import org.voltdb.VoltTable;
 import org.voltdb.client.Client;
@@ -34,6 +32,8 @@ import org.voltdb.client.ClientResponse;
 import org.voltdb.client.NoConnectionsException;
 import org.voltdb.client.ProcCallException;
 import org.voltdb.compiler.VoltProjectBuilder;
+
+import junit.framework.Test;
 
 public class TestPartialIndexesSuite extends RegressionSuite {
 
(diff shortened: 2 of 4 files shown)


tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -2900,13 +2900,14 @@ public void subtestOverlay(Client client, String overlayProc, String overlayFull
     // Unicode character to UTF8 string character
     public void testChar() throws NoConnectionsException, IOException, ProcCallException {
         System.out.println("STARTING test CHAR");
-        Client client = getClient();
-        ClientResponse cr;
-        VoltTable result;
 
         // Hsql has wrong answers.
         if (isHSQL()) return;
```
BillWhite: While you are working on this file you could put braces around the return.  It's not necessary, but it's a good idea.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -56,11 +56,6 @@ public TestPartialIndexesSuite(String name) {
     public void testPartialUniqueIndex() throws Exception {
         Client client = getClient();
 
-        if (isHSQL()) {
-            // HSQL doesn't support partial indexes
-            return;
-        }
-
```
BillWhite: What happened to this test?

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -229,10 +224,6 @@ public void testPartialUniqueIndex() throws Exception {
     public void testPartitionPartialUniqueIndex() throws Exception {
         Client client = getClient();
 
-        if (isHSQL()) {
-            // HSQL doesn't support partial indexes
-            return;
-        }
```
BillWhite: What happened to this test?

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -508,9 +494,9 @@ static public Test suite() {
         builder.addServerConfig(config);
 
         // HSQLDB
-        config = new LocalCluster("testpartialindexes-cluster.jar", 1, 1, 0, BackendTarget.HSQLDB_BACKEND);
-        if (!config.compile(project)) fail();
-        builder.addServerConfig(config);
```
BillWhite: Are we no longer testing with HSQL?  That's probably ok since we have sqlcoverage, I guess.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -508,9 +494,9 @@ static public Test suite() {
...
+        //config = new LocalCluster("testpartialindexes-cluster.jar", 1, 1, 0, BackendTarget.HSQLDB_BACKEND);
+        //if (!config.compile(project)) fail();
+        //builder.addServerConfig(config);
```
paulmartel: Code should not be commented out without an additional comment justifying why it might be useful in the future.
At most, if the original code, as in this case, followed a common pattern but now deviates, all there should be here is a comment justifying the deviation e.g.

    // The HSQLDB backend does not support partial indexes, so don't run a HSQLDB configuration.

If it is not likely to be useful in the future, the MOST you need is an explanation of why

tests/frontend/org/voltdb/regressionsuites/TestProcedureAPISuite.java
```diff
@@ -155,8 +155,9 @@ public void testCallArrayOfArrays() throws IOException, ProcCallException {
     }
 
     public void testMultiPartitionCURRENT_TIMESTAMP() throws IOException, ProcCallException {
-        Client client = getClient();
+
```
paulmartel: A positive test would be clearer here:

    if (isHSQL()) {
        return;
    }

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
paulmartel: @BillWhite No -- you can't test a non-feature of HSQL.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
paulmartel: @BillWhite It's suspenders got replaced with a belt.

johnTcrawford: Yeah, it seems, based on the comment that HSQL does not support partial indexes, that this test will now fail, when run against HSQL.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
```diff
@@ -341,11 +332,6 @@ public void testPartitionPartialUniqueIndex() throws Exception {
     public void testPartialIndex() throws Exception {
         Client client = getClient();
 
-        if (isHSQL()) {
-            // HSQL doesn't support partial indexes
-            return;
-        }
-
```
johnTcrawford: Again, won't removing this (and the one above, and probably the one below) cause this test to fail, when run against HSQL? Or am I missing something.

tests/frontend/org/voltdb/regressionsuites/TestPartialIndexesSuite.java
johnTcrawford: Oh, I think I see now. We're no longer running just this test suite against HSQL, since HSQL does not support partial indexes, which is what this test suite tests. Seems reasonable, but I suggest adding a comment explaining that reasoning (assuming I have it right), here where the call to HSQL is commented out. (It's possible that, at some point in the not-too-distant future, that these tests might be switched to run against the PostgreSQL backend, rather than HSQL.)

PR #4144, opened 2016-12-22
Johnc innov sqlcov dir

johnTcrawford:
Created and restructured the /tests/sqlcoverage/ directory, with several subdirectories; split config files into smaller ones to be run in parallel, per ENG-11324 (& code changes to make that work); renamed certain Geo test files, more consistently; removed the need for the annoying -Ddebug=transform_sql arg.


wolffcm:
This looks good to me.  To get the better parallelism in Jenkins, I guess you'll need to create a bunch of new jobs there?

johnTcrawford:
Chris: yes, although I have versions of those Jenkins jobs already, so it will just be a matter of renaming and tweaking them.

PR #4151, opened 2016-12-27
ENG-11661 use test switch to speedup unit test

zeemanhe:
use env variable VOLT_JUSTATEST in the build as a switch to speed up unit tests.

### DIFF
--- a/src/frontend/org/voltdb/client/ClientUtils.java
+++ b/src/frontend/org/voltdb/client/ClientUtils.java
@@ -91,14 +91,4 @@ public class ClientUtils {
         }
         return buffer;
     }
-
-    // Utility to try to figure out if this is a test case.  Various junit targets in
-    // build.xml set this environment variable to give us a hint
-    public static boolean isThisATest() {
-        String test = System.getenv().get("VOLT_JUSTATEST");
-        if (test == null) {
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/client/ClientUtils.java
```diff
@@ -91,4 +91,14 @@ public static final long cheesyBufferCheckSum(ByteBuffer buffer) {
...
     }
+
+    // Utility to try to figure out if this is a test case.  Various junit targets in
+    // build.xml set this environment variable to give us a hint
+    public static boolean isThisATest() {
+        String test = System.getenv().get("VOLT_JUSTATEST");
+        if (test == null) {
+            test = System.getProperty("VOLT_JUSTATEST");
+        }
+        return ("YESYESYES".equalsIgnoreCase(test));
```
luyangco: How about using `true` and `false`, `System.getProperty("VOLT_JUSTATEST", false)` allows you to give a default. You can use `Boolean.valueOf()` to convert it to a boolean.

zeemanhe: updated and moved this utility to CoreUtils so both client and server can use it.

PR #4153, opened 2016-12-27
EE Cleanup prior to swap table implementation.

paulmartel:
Changes to names and whitespace for greater consistency, understandability.

Removed a few dead code paths, especially around testing boolean return values that are always true. Even if we decide to detect legit error cases in the future, they should be handled with asserts and/or exceptions at the point of detection rather than passing the buck via mysterious error return values. There's still far to go on all of these fronts.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -46,6 +46,7 @@
 #include "VoltDBEngine.h"
 
 #include "ExecutorVector.h"
+
 #include "catalog/catalog.h"
 #include "catalog/catalogmap.h"
 #include "catalog/cluster.h"
@@ -59,6 +60,7 @@
 #include "catalog/planfragment.h"
 #include "catalog/statement.h"
 #include "catalog/table.h"
+
 #include "common/ElasticHashinator.h"
 #include "common/executorcontext.hpp"
 #include "common/FailureInjection.h"
@@ -69,20 +71,25 @@
 #include "common/SerializableEEException.h"
 #include "common/TupleOutputStream.h"
 #include "common/TupleOutputStreamProcessor.h"
+
 #include "executors/abstractexecutor.h"
+
 #include "indexes/tableindex.h"
 #include "indexes/tableindexfactory.h"
+
 #include "plannodes/abstractplannode.h"
 #include "plannodes/plannodefragment.h"
+
 #include "storage/AbstractDRTupleStream.h"
-#include "storage/tablefactory.h"
-#include "storage/persistenttable.h"
-#include "storage/streamedtable.h"
-#include "storage/MaterializedViewHandler.h"
-#include "storage/MaterializedViewTriggerForWrite.h"
-#include "storage/TableCatalogDelegate.hpp"
 #include "storage/CompatibleDRTupleStream.h"
 #include "storage/DRTupleStream.h"
+#include "storage/MaterializedViewHandler.h"
+#include "storage/MaterializedViewTriggerForWrite.h"
+#include "storage/persistenttable.h"
+#include "storage/streamedtable.h"
+#include "storage/TableCatalogDelegate.hpp"
+#include "storage/tablefactory.h"
+
 #include "org_voltdb_jni_ExecutionEngine.h" // to use static values
 
 #include "boost/foreach.hpp"
@@ -137,7 +144,7 @@ typedef boost::multi_index::multi_index_container<
 /// This class wrapper around a typedef allows forward declaration as in scoped_ptr<EnginePlanSet>.
 class EnginePlanSet : public PlanSet { };
 
-VoltDBEngine::VoltDBEngine(Topend *topend, LogProxy *logProxy)
+VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
     : m_currentIndexInBatch(-1),
       m_currentUndoQuantum(NULL),
       m_partitionId(-1),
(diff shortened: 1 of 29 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -695,9 +689,9 @@ VoltDBEngine::processCatalogDeletes(int64_t timestamp)
          */
         assert(tcd);
         if (tcd) {
-            Table *table = tcd->getTable();
+            Table* table = tcd->getTable();
             m_delegatesByName.erase(table->name());
-            StreamedTable *streamedtable = dynamic_cast<StreamedTable*>(table);
+            auto streamedtable = dynamic_cast<StreamedTable*>(table);
```
BillWhite: This is probably kvetching, but what evidence do we have that auto actually works for us?  We don't actually support C++11, which is where auto is introduced.  The C6 gcc compiler is the only pre-11 compiler we use, I think, but that's what we need to use on C6.  The C6 gcc compiler has auto, but I think it's marked as experimental.

I'm not saying these should be done.  I'm just wondering if it's a good idea to mix up standards and old compilers this way.

src/ee/storage/MaterializedViewTriggerForInsert.h
```diff
@@ -58,7 +58,9 @@ class MaterializedViewTriggerForInsert {
      */
     void processTupleInsert(const TableTuple &newTuple, bool fallible);
 
-    PersistentTable * targetTable() const { return m_target; }
+    PersistentTable * targetTable() const { return m_dest; }
```
BillWhite: Why is this targetTable and not destTable, since you've changed m_target to m_dest?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1059,24 +1045,24 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp)
             // See initMaterializedViewsAndLimitDeletePlans
 
             for (int ii = 0; ii < survivingInfos.size(); ++ii) {
-                catalog::MaterializedViewInfo * currInfo = survivingInfos[ii];
-                PersistentTable* oldTargetTable = survivingViews[ii]->targetTable();
+                auto currInfo = survivingInfos[ii];
```
paulmartel: I think there are many cases where types are obvious and others where having to type and scan the type names just adds not-very-helpful complexity.
Replacing specific long detailed types with auto makes the code more compact, arguably more readable, more friendly to type declaration changes with or without copy/paste like to apply the same algorithm or idiom to other types, or when templatizing.

I'm still testing the waters here.

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -49,46 +54,44 @@
...
         return persistentTable->deltaTable();
     }
+
     return m_table;
 }
 
-TupleSchema *TableCatalogDelegate::createTupleSchema(catalog::Database const &catalogDatabase,
-                                                     catalog::Table const &catalogTable) {
+TupleSchema* TableCatalogDelegate::createTupleSchema(catalog::Database const& catalogDatabase,
+                                                     catalog::Table const& catalogTable) {
```
BillWhite: I have to say, moving the ptr_operator from the declarator to the decl_specifier always seems weird to me, and contrary to the C++ standard grammar.  But people seem to like it, so who am I to judge.

src/ee/execution/VoltDBEngine.h
```diff
@@ -378,43 +413,39 @@ class __attribute__((visibility("default"))) VoltDBEngine {
...
+        void getUSOForExportTable(size_t& ackOffset, int64_t& seqNo, std::string tableSignature);
 
         /**
          * Retrieve a hash code for the specified table
          */
         size_t tableHashCode(int32_t tableId);
 
-        void updateHashinator(HashinatorType type, const char *config,
-                              int32_t *configPtr, uint32_t numTokens);
+        void updateHashinator(HashinatorType type, char const* config,
```
paulmartel: Also testing the waters, here. The "const later" style first came to my attention when I saw it being used in complicated compiler diagnostics. Something I like about it is that the common case of adding " const&" to a declaration is one change to the line instead of two. Also, I think that it puts the more interesting aspects of the type first on the line -- I often don't much care if something is const.  Also, I prefer the punctuation closer to the end of the type than the beginning of the variable because it makes easier to scan quickly to the variable name, the other critical piece of information.

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -173,35 +169,30 @@ bool TableCatalogDelegate::getIndexScheme(catalog::Table const &catalogTable,
  */
 static std::string
 getIndexIdFromMap(TableIndexType type, bool countable, bool isUnique,
-                  const std::string& expressionsAsText, vector<int32_t> columnIndexes,
-                  const std::string& predicateAsText) {
+                  std::string const& expressionsAsText, std::vector<int32_t> columnIndexes,
+                  std::string const& predicateAsText) {
```
BillWhite: Aren't these two completely identical?  Why make this const T -> T const change?

src/ee/storage/MaterializedViewTriggerForInsert.h
paulmartel: I was on the fence about invalidating all callers, but I probably DID go through with it in the end.

src/ee/storage/TableCatalogDelegate.cpp
paulmartel: @BillWhite  In case you missed by earlier reply:
The "const later" style first came to my attention when I saw it being used in complicated compiler diagnostics. Something I like about it is that the common case of adding " const&" to a declaration is one change to the line instead of two. Also, I think that it puts the more interesting aspects of the type first on the line -- I often don't much care if something is const. Also, I prefer the punctuation closer to the end of the type than the beginning of the variable because it makes easier to scan quickly to the variable name, the other critical piece of information.

paulmartel:
Merged via another branch.

PR #4154, opened 2016-12-27
Implement EE swap table

paulmartel:


src/ee/storage/persistenttable.cpp
```diff
@@ -555,6 +556,144 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+    tcd1->setTable(otherTable);
+
+    auto tcd2 = engine->getTableDelegate(otherTable->m_name);
+    assert(tcd2->getTable() == otherTable);
+    tcd2->setTable(this);
+
+    // Swap the table attributes that must continue to be associated with each
+    // table's name/identity, not its swapped content.
+    // We MIGHT want to consider making these attributes of TableCatalogDelegate
+    // instead of PersistentTable?
```
wolffcm: I like this idea.  But it's also nice for debugging for a persistent table to have a reference to its name.  Maybe the persistent table should have a pointer back to its delegate, so it's possible to  easily to get its current name.

src/ee/storage/persistenttable.cpp
```diff
@@ -555,6 +556,144 @@ void PersistentTable::truncateTable(VoltDBEngine* engine, bool fallible) {
...
+
+    auto heldName = m_name;
```
wolffcm: You could use `std::swap` here.

tests/ee/storage/persistenttable_test.cpp
```diff
@@ -160,10 +165,74 @@ class PersistentTableTest : public Test {
             "set $PREV oncommit \"\"\n"
             "set $PREV index /clusters#cluster/databases#database/tables#T/indexes#VOLTDB_AUTOGEN_IDX_PK_T_PK\n"
             "set $PREV foreignkeytable null\n"
+
+            "add /clusters#cluster/databases#database tables X\n"
```
wolffcm: It would be good to provide the SQL DDL that produced this payload, in case it ever needs to be regenerated.

tests/ee/storage/persistenttable_test.cpp
```diff
@@ -160,10 +165,74 @@ class PersistentTableTest : public Test {
...
+                        size_t nTuples, size_t nDupTuples) {
+        validateCounts(table, nTuples, nIndexes);
+        validateCounts(dupTable, nDupTuples, nIndexes);
+    }
+
+    void validateCounts(PersistentTable* table, size_t nTuples, size_t nIndexes) {
+        auto iterator = table->iterator();
+        ASSERT_EQ(nTuples > 0, iterator.hasNext());
+        ASSERT_EQ(nTuples, table->activeTupleCount());
+        ASSERT_EQ(nIndexes, table->indexCount());
```
wolffcm: Do you need the `nIndexes` parameter?  Can't you just assert that `allIndexes` produces a collection of the same size for both tables?

tests/ee/storage/persistenttable_test.cpp
```diff
@@ -313,6 +382,407 @@ TEST_F(PersistentTableTest, TruncateTableTest) {
...
+    ASSERT_NE(NULL, dupTable);
+
+    validateCounts(1, table, dupTable, tuplesToInsert, tuplesToInsert*3);
+
+    dupTable->truncateTable(engine);
+    dupTable = engine->getTableDelegate("X")->getPersistentTable();
+
+    ASSERT_NE(NULL, dupTable);
+
+    ////W-hy T-he F-ailure? validateCounts(1, table, dupTable, tuplesToInsert, 0);
```
wolffcm: I seem to recall that the `activeTupleCount` persistent table attribute was billed in a comment as being 
"for debugging only."  Could have something to do with this strange behavior?

src/ee/plannodes/swaptablesnode.h
```diff
@@ -0,0 +1,109 @@
...
+     // Other Target Table
+     // These tables are different from the input and the output tables
+     // The plannode can read in tuples from the input table(s) and apply them to the target table
+     // The results of the operations will be written to the the output table
+     TableCatalogDelegate* m_otherTcd;
+     std::string m_otherTargetTableName;
+
+     std::vector<std::string> m_theIndexes;
+     std::vector<std::string> m_otherIndexes;
+     std::vector<std::string> m_theViews;
```
BillWhite: If these statements are executed:

CREATE TABLE T2 WITH SWAP TABLE T1;
CREATE PROCEDURE P AS SWAP TABLE T1 WITH TABLE T2;
CREATE INDEX III ON T1 AS (T1.A, T1.B);

will procedure P be replanned?  Since there are different indexes now, III on T1 and presumably III on T2, though there seems to be a name conflict, the old list of indexes will be wrong, won't it?

There is a similar question about views.

paulmartel: @BillWhite 
Views are off the table for now (I do need to remove this remnant from the code).
The name conflict will be detected when we try to establish the tables as swappable.
When there is no name conflict, the catalog will be updated and the cached and cataloged plans will be re-planned as usual. This will use the exact same mechanism as statements like "INSERT INTO T1 SELECT ... FROM T1, T2 ..." which also must be replanned theoretically on every drop or add of an index on either table, in case the statement IS using an obsolete index or SHOULD BE using a newly defined index.

paulmartel:
Merged directly to master as 2 commits (refactor and new code).

PR #4155, opened 2016-12-28
ENG-11667 remove blocker for rejoin

zeemanhe:
Remove rejoin blocker on Zk when a rejoin is done.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2524,6 +2524,10 @@ private void onRejoinCompletion() {
             // so there is no need to wait for the truncation snapshot requested
             // above to finish.
             if (logRecoveryCompleted || m_joining) {
+                if (m_rejoining) {
+                   final String node = VoltZK.rejoinNodesBlockerHost + m_myHostId;
+                   VoltZK.removeRejoinNodeIndicator(m_messenger.getZK(), node);
```
xinjiacs: Do you need to move "m_rejoining = false;" into this IF case ?

I see it moved in current master.

zeemanhe: Will make change for consistency.

PR #4160, opened 2016-12-29
ENG-11587, fix the rack-aware rejoin issue on partition group algorithm.

luyangco:
Change-Id: Ie9074faf89254deaa4a58f02c0dfad18fc41bcfc

### DIFF
--- a/src/frontend/org/voltdb/AbstractTopology.java
+++ b/src/frontend/org/voltdb/AbstractTopology.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -1027,15 +1028,15 @@ public class AbstractTopology {
         String[] token2parts = token2.split("\\.");
 
         // trim shared path prefix
-        while (token1parts.length > 0) {
-            if (token2parts.length == 0) break;
-            if (!token1parts[0].equals(token2parts[0])) break;
-            token1parts = Arrays.copyOfRange(token1parts, 1, token1parts.length);
-            token2parts = Arrays.copyOfRange(token2parts, 1, token2parts.length);
+        int index = 0;
+        while (index < token1parts.length) {
+            if (index == token2parts.length) break;
+            if (!token1parts[index].equals(token2parts[index])) break;
+            index++;
         }
 
-        // distance is now the sum of the two path lengths
-        return token1parts.length + token2parts.length;
+        // distance is the sum of the two diverting path lengths
+        return token1parts.length + token2parts.length - 2 * index;
     }
 
     /**
@@ -1381,6 +1382,22 @@ public class AbstractTopology {
     public static List<List<Integer>> sortHostIdByHGDistance(int hostId, Map<Integer, String> hostGroups) {
         String localHostGroup = hostGroups.get(hostId);
         Preconditions.checkArgument(localHostGroup != null);
+        List<List<Integer>> result = Lists.newArrayList();
+
+        // special case, only one node to compare
+        if (hostGroups.keySet().size() == 2 && hostGroups.keySet().contains(hostId)) {
+            List<Integer> aList = Lists.newArrayList();
+            Iterator<Integer> iter = hostGroups.keySet().iterator();
+            while (iter.hasNext()) {
+                Integer hid = iter.next();
+                if (hid != hostId) {
+                    aList.add(hid);
+                    result.add(aList);
+                    return result;
+                }
+            }
+        }
+        // sort by distance
         List<Integer> sortedList = hostGroups.keySet().stream()
                                                       .filter(hid -> hid != hostId)
                                                       .sorted((hid1, hid2) ->
@@ -1388,11 +1405,8 @@ public class AbstractTopology {
                                                           computeHADistance(localHostGroup, hostGroups.get(hid1))
                                                       )
                                                       .collect(Collectors.toList());
-        List<List<Integer>> result = Lists.newArrayList();
-        if (sortedList.size() == 1) {
-            result.add(sortedList);
-            return result;
-        }
+
+        // group by rack-aware group
         List<Integer> subgroup = Lists.newArrayList();
         subgroup.add(sortedList.get(0));
         for (int i = 1; i < sortedList.size(); i++) {
--- a/src/frontend/org/voltdb/AbstractTopology.java
+++ b/src/frontend/org/voltdb/AbstractTopology.java
@@ -1405,6 +1405,7 @@ public class AbstractTopology {
     }
     /**
      * Best effort to find the matching host on the existing topology from ZK
+     * Use the placement group of the recovering host to match a lost node in the topology
      * @param topology The topology
      * @param liveHosts The live host id
      * @param localHostId The rejoining host id
@@ -1442,6 +1443,7 @@ public class AbstractTopology {
             return null;
         }
 
+        //update placement groups with recovering host
         for (Map.Entry<String, Set<Integer>> entry : haGroupMaps.entrySet()) {
             HAGroup haGroup = new HAGroup(entry.getKey(), entry.getValue().stream().mapToInt(Number::intValue).toArray());
             for (Integer hostId : entry.getValue()) {
@@ -1450,6 +1452,7 @@ public class AbstractTopology {
             }
         }
 
+        //move partitions to host
         for (Partition partition : topology.partitionsById.values()) {
             MutablePartition mp = new MutablePartition(partition.id, partition.k);
             mutablePartitionMap.put(mp.id, mp);
--- a/src/frontend/org/voltdb/AbstractTopology.java
+++ b/src/frontend/org/voltdb/AbstractTopology.java
@@ -37,9 +37,11 @@ import org.json_voltpatches.JSONException;
 import org.json_voltpatches.JSONObject;
 import org.json_voltpatches.JSONStringer;
 
+import com.google_voltpatches.common.base.Preconditions;
 import com.google_voltpatches.common.collect.ImmutableMap;
 import com.google_voltpatches.common.collect.ImmutableSet;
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1013,24 +1015,23 @@ private MutableHAGroup(String token) {
...
-            if (token1parts.equals(token2parts)) {
-                token1parts = Arrays.copyOfRange(token1parts, 1, token1parts.length);
-                token2parts = Arrays.copyOfRange(token2parts, 1, token2parts.length);
-            }
+            if (!token1parts[0].equals(token2parts[0])) break;
+            token1parts = Arrays.copyOfRange(token1parts, 1, token1parts.length);
+            token2parts = Arrays.copyOfRange(token2parts, 1, token2parts.length);
         }
 
         // distance is now the sum of the two path lengths
```
zeemanhe: It seems to me that the method tries to find out the sum of two diverting paths. The algorithm can then be simplified and be more efficient.
(1) split the two tokens into two arrays, token1Parts and token2Parts.  
(2)find the shorter one and iterate through two arrays
(3)find break out the loop at the first index where their values are different.  remember the index
(4) the distance should be token1parts.length + token2parts.lengh - 2 * index.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1370,4 +1371,47 @@ public int getReplicationFactor() {
...
+    /**
+     * Sort all nodes in reverse hostGroup distance order, then group by rack-aware group, local host id is excluded.
+     * @param hostId the local host id
+     * @param hostGroups a host id to group map
+     * @return sorted grouped host ids from farthest to nearest
+     */
+    public static List<List<Integer>> sortHostIdByHGDistance(int hostId, Map<Integer, String> hostGroups) {
+        String localHostGroup = hostGroups.get(hostId);
+        Preconditions.checkArgument(localHostGroup != null);
+        List<Integer> sortedList = hostGroups.keySet().stream()
```
zeemanhe: Is the check needed?  Is a host is always associated with a group?

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1370,4 +1371,47 @@ public int getReplicationFactor() {
...
+                                                      .sorted((hid1, hid2) ->
+                                                          computeHADistance(localHostGroup, hostGroups.get(hid2)) -
+                                                          computeHADistance(localHostGroup, hostGroups.get(hid1))
+                                                      )
+                                                      .collect(Collectors.toList());
+        List<List<Integer>> result = Lists.newArrayList();
+        if (sortedList.size() == 1) {
+            result.add(sortedList);
+            return result;
+        }
```
zeemanhe: it would be better off to put the above before sorting.  If hostGroups have only one entry excluding hostId, return right away.

src/frontend/org/voltdb/iv2/Cartographer.java
```diff
@@ -429,21 +434,34 @@ public int getReplicaCountForPartition(int partition) {
                 }
             }
         }
-        return partitions;
     }
```
zeemanhe: computeReplacementPartitions finds the partitions for a host from the least replicated partitions.  Will this algorithm ensure the partition groups be reconstituted after rejoining?

src/frontend/org/voltdb/AbstractTopology.java
luyangco: Since this is a public method, I feel more safe to have a sanity check.

src/frontend/org/voltdb/AbstractTopology.java
luyangco: I agree, it's a better way to implement.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1370,4 +1372,60 @@ public int getReplicationFactor() {
...
+                                                      )
+                                                      .collect(Collectors.toList());
+
+        // group by rack-aware group
+        List<Integer> subgroup = Lists.newArrayList();
+        subgroup.add(sortedList.get(0));
+        for (int i = 1; i < sortedList.size(); i++) {
+            int hid1 = sortedList.get(i - 1);
+            int hid2 = sortedList.get(i);
+            int distance1 = computeHADistance(localHostGroup, hostGroups.get(hid1));
```
xinjiacs: I find this computeHADistance() called multiple times on same value. It seems some of certain inefficiency. 

I need more time on this algorithm.

zeemanhe: It would be better to pre-calculate all the distances from hostId to others and do a sort to avoid duplicate calculations.

src/frontend/org/voltdb/iv2/Cartographer.java
luyangco: You catch a good point.

The answer is yes and no. Rejoin won't change the partition group property. In a balanced PG layout, each node within a PG always has all the replicas of the partitions belong to this PG, so least replicated partitions are always aligned on host boundary. That's the reason why the rejoining node can always become "peer" of another node. 

But this is an issue when PG is not existed, least replicated partitions no longer align on host boundary, even worse is the rejoining node may not find the right rack group to join. Let me think about a way to fix it.

zeemanhe: @luyangco If there are multiple lost hosts, rejoining is processed one at a time. With the algorithm that the least replicated partitions will be put onto the recovering host, the partition assignments does not follow the partitioning group mechanism.  
ZK has the topology before hosts are down.  We could use the topology to recover the partition assignments and update ZK with the recovered topology.  In this way, the partition group structure can be recovered after rejoining.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1437,6 +1447,32 @@ public Boolean get() {
...
+     * @return A list of partitions if recover effort is a success.
+     */
+    private List<Integer> recoverPartitions(AbstractTopology topology, String haGroup) {
+
+        List<Integer>partitions = null;
+        AbstractTopology recoveredTopo = AbstractTopology.mutateRecoverTopology(topology,
+                m_messenger.getLiveHostIds(),
+                m_messenger.getHostId(),
+                haGroup);
+        if (recoveredTopo != null) {
```
xinjiacs: early return with NULL.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1370,4 +1367,104 @@ public int getReplicationFactor() {
...
+     * @return sorted grouped host ids from farthest to nearest
+     */
+    @SuppressWarnings("unchecked")
+    public static List<Collection<Integer>> sortHostIdByHGDistance(int hostId, Map<Integer, String> hostGroups) {
+        String localHostGroup = hostGroups.get(hostId);
+        Preconditions.checkArgument(localHostGroup != null);
+        List<Collection<Integer>> result = Lists.newArrayList();
+
+        // special case, cluster has two nodes and one is the rejoining node.
+        if (hostGroups.keySet().size() == 2 && hostGroups.keySet().contains(hostId)) {
```
xinjiacs: @zeemanhe This special case only contains two nodes and one of the nodes is the rejoining node. So it does not need to complicate the code by using set difference.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1013,28 +1019,19 @@ private MutableHAGroup(String token) {
      * Not the most efficient way to do this, but even n^2 for 100 nodes is computable
      */
     private static int computeHADistance(/*final Map<String, Integer> distanceCache,*/ String token1, String token2) {
```
xinjiacs: Remove "/*final Map<String, Integer> distanceCache,*/" ?

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1393,33 +1390,81 @@ public int getReplicationFactor() {
...
-                result.add(subgroup);
-                subgroup = Lists.newArrayList();
-                subgroup.add(hid2);
+        Multimap<Integer, Integer> distanceMap = LinkedListMultimap.create();
+        for (Map.Entry<Integer, String> entry : hostGroups.entrySet()) {
+            if (hostId == entry.getKey()) continue;
+            distanceMap.put(computeHADistance(localHostGroup, entry.getValue()), entry.getKey());
+        }
+
+        result = distanceMap.asMap().entrySet().stream()
```
xinjiacs: Can you add a comment on this lambda calculating the result?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1459,16 +1459,17 @@ public Boolean get() {
...
                 m_messenger.getLiveHostIds(),
                 m_messenger.getHostId(),
                 haGroup);
-        if (recoveredTopo != null) {
-            partitions = recoveredTopo.getPartitionIdList(m_messenger.getHostId());
-            if (partitions != null && partitions.size() == m_catalogContext.getNodeSettings().getLocalSitesCount()) {
-                TopologyZKUtils.updateTopologyToZK(m_messenger.getZK(), recoveredTopo);
-            }
+        if (recoveredTopo == null) {
+            return partitions;
```
xinjiacs: "return null" here and delay creating "partitions" list after this IF block.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1459,16 +1459,17 @@ public Boolean get() {
...
+        }
+        partitions = recoveredTopo.getPartitionIdList(m_messenger.getHostId());
+        if (partitions != null && partitions.size() == m_catalogContext.getNodeSettings().getLocalSitesCount()) {
```
xinjiacs: What will happen when "partitions != null" but the sites per host does not match? Is it possible?

zeemanhe: The sites per host can be passed from command line or deployment file, which can be different from what was on the topology. It is a defensive check.

PR #4161, opened 2016-12-29
ENG-11647: Pre-populate site trackers for consumer clusters requested in snapshot request and start cursor request

slowsoul:
1. Add a new ExecuteTask task type INIT_DRID_TRACKER which is used to populating empty trackers in XDCR
2. Add partition count in snapshot request protobuf, include partition count in snapshot requests and start cursor requests (they are not really used for now, will remove them if not needed)

### DIFF
--- a/src/frontend/org/voltdb/ClientInterface.java
+++ b/src/frontend/org/voltdb/ClientInterface.java
@@ -1867,23 +1867,9 @@ public class ClientInterface implements SnapshotDaemon.DaemonInitiator {
      * @throws IOException
      * @throws InterruptedException
      */
-    public ClientResponse callExecuteTask(long timeoutMS, byte[] params ) throws IOException, InterruptedException {
-        final String procedureName = "@ExecuteTask";
-        Config procedureConfig = SystemProcedureCatalog.listing.get(procedureName);
-        Procedure proc = procedureConfig.asCatalogProcedure();
-        StoredProcedureInvocation spi = new StoredProcedureInvocation();
-        spi.setProcName(procedureName);
-        spi.setParams(params);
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -1907,11 +1893,12 @@ public void callExecuteTaskAsync(SimpleClientResponseAdapter.Callback cb, byte[]
         if (spi.getSerializedParams() == null) {
             spi = MiscUtils.roundTripForCL(spi);
         }
-        createTransaction(m_executeTaskAdpater.connectionId(), spi,
-                proc.getReadonly(), proc.getSinglepartition(), proc.getEverysite(),
-                0 /* Can provide anything for multi-part */,
-                spi.getSerializedSize(), System.nanoTime());
-
+        synchronized (m_executeTaskAdpater) {
```
nshi: Is this necessary? I think we already synchronize the calls in `DRDispatcherAdapter`.

slowsoul: This is because callExecuteTask() and callExecuteTaskAsync() use a special connection Id and adapter (m_executeTaskAdapter) to call ClientInterface.createTransaction() while DRDispatcherAdapter uses its own connection Id. Also, the callers of these two methods reside in many threads and DRDispatcherAdapter is not involved in these calls. So synchronize on this adapter can help us order the "@ExecuteTask" transactions.

manjujames: This means that we can take out the synchronized from DRDispatcherAdapter methods executeSetSiteDRIdTrackers and attemptTruncationSnapshot.

PR #4165, opened 2017-01-05
ENG-11693: DRROLE for multi-cluster XDCR

nshi:
- Return one row per conversation in DR producer and consumer node
  stats.

- Roll DR producer and consumer node stats into DRROLE stats for each
  conversation.

- This patch doesn't contain the work for the STOPPED state on the
  producer. It requires break replication to work on the producer.

### DIFF
--- a/src/frontend/org/voltdb/DRProducerNodeStats.java
+++ b/src/frontend/org/voltdb/DRProducerNodeStats.java
@@ -18,6 +18,13 @@
 package org.voltdb;
 
 public class DRProducerNodeStats {
+    public static final DRProducerNodeStats DISABLED_NODE_STATS =
+        new DRProducerNodeStats((short) -1, DRRoleStats.State.DISABLED, "",
+                                -1, -1, 0);
+    public static final DRProducerNodeStats PENDING_NODE_STATS =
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/DRProducerNodeStats.java
```diff
@@ -18,6 +18,13 @@
 package org.voltdb;
 
 public class DRProducerNodeStats {
+    public static final DRProducerNodeStats DISABLED_NODE_STATS =
+        new DRProducerNodeStats((short) -1, DRRoleStats.State.DISABLED, "",
```
slowsoul: Maybe the state string in these 2 constants should be "NONE" for consistency.

nshi: Do you mean for the sync snapshot state or the general node state? I think we used to leave it empty for the snapshot state.

slowsoul: Sync snapshot state. Per the doc "During normal operation, this value is "NONE" indicating either that replication is not active or that transactions are actively being replicated. If a synchronization snapshot is in progress, this value provides additional information about the specific activity underway." and old code, I think we used to provide at least "NONE" in this field.

PR #4171, opened 2017-01-06
ENG-11726 - remove <admin-mode>

akhanzode:
Dont merge untill system tests are looked at.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2895,12 +2893,12 @@ public boolean shutdown(Thread mainSiteThread) throws InterruptedException {
                  * Various scheduled tasks get crashy in unit tests if they happen to run
                  * while other stuff is being shut down
                  */
-                for (ScheduledFuture<?> sc : m_periodicWorks) {
-                    sc.cancel(false);
-                    try {
+                try {
+                    for (ScheduledFuture<?> sc : m_periodicWorks) {
```
vtkstef: don't you need to try each one, instead of failing on the first and not cancelling the remaining?

src/frontend/org/voltdb/sysprocs/SystemInformation.java
```diff
@@ -538,14 +538,6 @@ static public VoltTable populateDeploymentProperties(
 
         results.addRow("heartbeattimeout", Integer.toString(cluster.getHeartbeattimeout()));
 
-        results.addRow("adminport", Integer.toString(VoltDB.instance().getConfig().m_adminPort));
-        String adminstartup = "false";
-        if (cluster.getAdminstartup())
```
vtkstef: you can get the admin port from node settings which is available form system procedure context

akhanzode: Good point.. I should leave adminport and just remove adminstartup

PR #4180, opened 2017-01-10
ENG-11457: Move "--replica" CLI arg to the deployment file.

nshi:
- Print Deprecate warning when --replica is used.

- Removed "--replica" from the VoltDB.java CLI parser.

- Added "role" attribute to the <dr> tag in the deployment
  file. Currently only takes either "master" or "replica". The default
  is "master" if not specified.

- @Promote now uses @UAC to handle the promotion.


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -916,6 +916,21 @@ protected String checkModifyWhitelist(final CatalogType suspect,
...
+            } else if (field.equals("drRole")) {
+                final String prevRole = (String) prevType.getField("drRole");
+                final String newRole = (String) suspect.getField("drRole");
+                // Promote from replica to master
+                if (prevRole.equals("replica") && newRole.equals("master")) {
+                    return null;
+                }
+                // Enabling DR from none to master
+                else if (prevRole.equals("replica") && newRole.equals("master")) {
+                    return null;
```
JadenW: what's the difference of this condition with previous one? Should this be preRole == null according to the comment?

PR #4188, opened 2017-01-12
ENG-11729

akhanzode:
Remove old export syntax and DDL


src/frontend/org/voltcore/network/InputHandler.java
```diff
@@ -42,6 +42,7 @@
      * Handle the incoming message produced by retrieve next message
      *
      * @param message
+     * @param c
      * @param channelFacade
```
xinjiacs: @akhanzode Remove the "channelFacade"?

xinjiacs:
Do you test that "EXPORT TABLE" syntax  can not be recognized?

akhanzode:
No I dont its as good as garbage ddl syntax. I can add a test easily if you like.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -945,7 +945,13 @@ private void compileDatabase(
         // Actually parse and handle all the DDL
         // DDLCompiler also provides partition descriptors for DDL PARTITION
         // and REPLICATE statements.
-        final DDLCompiler ddlcompiler = new DDLCompiler(this, hsql, voltDdlTracker, m_classLoader);
+        final DDLCompiler ddlcompiler;
+        try {
+            ddlcompiler = new DDLCompiler(this, hsql, voltDdlTracker, m_classLoader);
+        } catch (Throwable t) {
```
akhanzode: good catch leftover for debug.

zeemanhe:
There are tons of comments or debug statements in the codes using "export table" instead of "stream". It is not worth updating them.  

Should the following be updated with "stream"?
line 213 in  ReportMaker.generateSchemaRow() 
line 282 in CatalogBuilder.setTableAsExportOnly()

PR #4189, opened 2017-01-13
Fix several problems with strings, and with aliases in order by lists.

BillWhite:
The SPACE sql function did not check for arguments being too large.  As
a result we would try to allocate a large amount of string space, which
caused allocator failures.

These same queries showed up problems in the planner with aliases in
order by lists.  If a select list element is a constant, and there is a
reference to it as an alias in the order by list, we didn't expect it
and we threw an error. There are also cases where data types for select
list aliases are not set in the Volt expression structures when we
expect them to be.  This can cause NPEs when converting from strings to
BIGINT values.  These have been fixed.

FWIW, this fixes ENG-11683, ENG-11684 and ENG-11762.

### DIFF
--- a/src/ee/common/ThreadLocalPool.h
+++ b/src/ee/common/ThreadLocalPool.h
@@ -48,7 +48,10 @@ public:
         Sized(int32_t requested_size) : m_size(requested_size) { }
     };
 
-    static const int POOLED_MAX_VALUE_LENGTH;
+    // This needs to be >= the VoltType.MAX_VALUE_LENGTH defined in java, currently 1048576.
+    // The rationale for making it any larger would be to allow calculating wider "temp"
+    // values for use in situations where they are not being stored as column values.
+    static const int POOLED_MAX_VALUE_LENGTH = 1024 * 1024;
 
     /**
      * Allocate space from a page of objects of the requested size.
--- a/src/ee/expressions/stringfunctions.h
+++ b/src/ee/expressions/stringfunctions.h
@@ -79,10 +79,15 @@ template<> inline NValue NValue::callUnary<FUNC_SPACE>() const {
 
     int32_t count = static_cast<int32_t>(castAsBigIntAndGetValue());
     if (count < 0) {
-        char msg[1024];
-        snprintf(msg, 1024, "data exception: substring error");
         throw SQLException(SQLException::data_exception_string_data_length_mismatch,
-            msg);
+                           "The argument to the SPACE function is negative");
+    } else if (ThreadLocalPool::POOLED_MAX_VALUE_LENGTH < count) {
+        std::ostringstream oss;
+        oss << "The argument to the SPACE function is larger than the maximum size allowed for strings ("
+            << ThreadLocalPool::POOLED_MAX_VALUE_LENGTH << " bytes). "
+            << "Reduce either the string size or repetition count.";
+        throw SQLException(SQLException::data_exception_string_data_length_mismatch,
+                           oss.str().c_str());
     }
 
     std::string spacesStr(count, ' ');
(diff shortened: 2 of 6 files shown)


src/ee/common/ThreadLocalPool.h
```diff
@@ -48,7 +48,10 @@ class ThreadLocalPool {
         Sized(int32_t requested_size) : m_size(requested_size) { }
     };
 
-    static const int POOLED_MAX_VALUE_LENGTH;
+    // This needs to be >= the VoltType.MAX_VALUE_LENGTH defined in java, currently 1048576.
+    // The rationale for making it any larger would be to allow calculating wider "temp"
+    // values for use in situations where they are not being stored as column values.
+    static const int POOLED_MAX_VALUE_LENGTH = 1024 * 1024;
```
BillWhite: Moving this to here makes this an inlined value.  It saves a memory reference.

src/ee/expressions/stringfunctions.h
```diff
@@ -79,10 +79,15 @@ template<> inline NValue NValue::callUnary<FUNC_SPACE>() const {
 
     int32_t count = static_cast<int32_t>(castAsBigIntAndGetValue());
     if (count < 0) {
-        char msg[1024];
-        snprintf(msg, 1024, "data exception: substring error");
         throw SQLException(SQLException::data_exception_string_data_length_mismatch,
-            msg);
+                           "The argument to the SPACE function is negative");
```
BillWhite: This is the most important fix in this commit.  We catch out-of-range values for the SPACE function.  Lack of this was causing EE crashes.

src/ee/expressions/stringfunctions.h
```diff
@@ -218,10 +223,8 @@ template<> inline NValue NValue::call<FUNC_LEFT>(const std::vector<NValue>& argu
     }
     int32_t count = static_cast<int32_t>(startArg.castAsBigIntAndGetValue());
     if (count < 0) {
-        char msg[1024];
-        snprintf(msg, 1024, "data exception: substring error");
         throw SQLException(SQLException::data_exception_string_data_length_mismatch,
-            msg);
+                           "The argument to the LEFT function is negative");
```
BillWhite: This is a better error message.

src/ee/expressions/stringfunctions.h
```diff
@@ -251,10 +254,8 @@ template<> inline NValue NValue::call<FUNC_RIGHT>(const std::vector<NValue>& arg
 
     int32_t count = static_cast<int32_t>(startArg.castAsBigIntAndGetValue());
     if (count < 0) {
-        char msg[1024];
-        snprintf(msg, 1024, "data exception: substring error");
         throw SQLException(SQLException::data_exception_string_data_length_mismatch,
-            msg);
+                           "The argument to the RIGHT function is negative.");
```
BillWhite: This is a better error message.

src/frontend/org/voltdb/planner/ParsedColInfo.java
```diff
@@ -139,6 +139,7 @@ static public ParsedColInfo fromOrderByXml(AbstractParsedStmt parsedStmt,
                     (child.name.equals("win_aggregation") == false) &&
                     (child.name.equals("function") == false) &&
                     (child.name.equals("rank") == false) &&
+                    (child.name.equals("value") == false) &&
```
BillWhite: I noticed this when testing.  If an alias to a select list constant is used in an order by, or presumably in a group by, this test failed.  Inserting "value" children in the test lets this pass.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionValue.java
```diff
@@ -115,7 +114,9 @@ public Object getValue(Session session) {
      * @return true for a successful conversion and false otherwise.
      */
     public static boolean voltMutateToBigintType(Expression maybeConstantNode, Expression parent, int childIndex) {
-        if (maybeConstantNode.opType == OpTypes.VALUE && maybeConstantNode.dataType.isBinaryType()) {
+        if (maybeConstantNode.opType == OpTypes.VALUE
```
BillWhite: The field maybeConstantNode.dataType may be null.  If it is we can't convert the string to a BIGINT in any case, and we don't want to throw an NPE.

BillWhite: I think it's when the actual value is NULL.

PR #4191, opened 2017-01-13
Eng 11724 uacbench apprunner

jwiebema1:


tests/test_apps/uacbench/uacbench.py
```diff
@@ -146,9 +146,9 @@ def generate_add_batch_jars(procedurecount, invocations, batchsize):
 
 # cleaning the input dir
 def clean_dirpath(dirpath):
-    if os.path.exists(dirpath):
-        shutil.rmtree(dirpath)
-    os.makedirs(dirpath, 0755)
+    #if os.path.exists(dirpath):
+    #    shutil.rmtree(dirpath)
+    #os.makedirs(dirpath, 0755)
```
xinjiacs: @jwiebema1 are these on purpose?

jwiebema1: yes, I need to run this more then one time and if I do , it will remove the generated files from the previous run,  I need to keep the files around each time it's run.  I moved the cleanup and makedirs part of the code into run.sh

PR #4193, opened 2017-01-13
ENG-10491: restrict range of timestamps in scalar functions

wolffcm:


src/ee/expressions/datefunctions.h
```diff
@@ -596,7 +614,9 @@ template<> inline NValue NValue::call<FUNC_VOLT_DATEADD_YEAR>(const std::vector<
     try {
         ts += boost::gregorian::years(static_cast<int>(interval));
         boost::posix_time::time_duration td = ts - EPOCH;
-        return getTimestampValue(td.total_microseconds());
+        int64_t epochMicros = td.total_microseconds();
```
BillWhite: Maybe it's me, but it seems like this sequence of code is repeated a lot of times.  Maybe making it an inline function would be appropriate?  What if it has to be uniformly altered?

PR #4194, opened 2017-01-14
ENG-11693: Fix voltadmin pause --wait after DR stats schema change.

nshi:


slowsoul:
check_dr_consumer() may also need to be changed, two columns are used.

nshi:
Thanks for noticing that. I've addressed it.

PR #4198, opened 2017-01-18
ENG-11459: Move "set dr=active;" from schema to the deployment file.

nshi:
To simplify the process of configuring an XDCR cluster, the "set
dr=active;" directive has been deprecated from the DDL. It generates a
warning if used, but the DDL will not fail. It is essentially a no-op
now.

To configure an XDCR cluster, user should use "<dr role='xdcr'></dr>" in
the deployment file. Once set, it cannot be changed. This has several
benefits.

- Cluster always starts with the correct role. Role will not change from
  or to XDCR at runtime. As a result, a lot of code to check for online
  role change has been removed.

- This has consolidated the code for determining the cluster role such
  as in the DRROLE statistics.

There were some complications in implementation mainly due to the fact
that DDL compilation now depends on the role specified in the deployment
file. To simplify this slightly, I pass the role flag into the
VoltCompiler's constructor and cache it as a member.

I search for all occurences of "set dr=active;" in tests and fixed them
using the new approach.

One thing I haven't done, for the purpose of minimizing (cough) change
impact on stability, is to consolidate all usages of role in the
cluster. There is still a ReplicationRole enum that only has REPLICA and
NONE, and there is the new proper enum DrRoleType. Ideally, all places
should use the DrRoleType.


src/frontend/org/voltdb/JdbcDatabaseMetaDataGenerator.java
```diff
@@ -241,10 +240,6 @@ else if (selector.equalsIgnoreCase("CLASSES"))
         {
             result = getClasses();
         }
-        else if (selector.equalsIgnoreCase("CONFIG"))
-        {
-            result = getConfigs();
-        }
```
nshi: I'm unsure about this removal. @slowsoul You may know more since you added this. Is this in the JDBC standard? Should I keep it but return an empty table instead?

src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -1090,25 +1090,6 @@ else if (isTableLimitDeleteStmt(suspect)) {
...
-                            new TablePopulationRequirements(tableName,
-                                                            tableName,
-                                                            errorMessage);
-                        retval.add(entry);
-                    }
-                }
-                return retval;
-            }
-
             return null;
```
wweiss-voltdb: Where was this code moved to?

nshi: Not moved. It's removed because you can no longer flip between passive and active online.

src/frontend/org/voltdb/JdbcDatabaseMetaDataGenerator.java
slowsoul: This is not in the JDBC standard, also never documented in our documentation, so we can remove it.

slowsoul: CONFIG_SCHEMA should also be removed.

tests/frontend/org/voltdb/planner/PlannerTestAideDeCamp.java
```diff
@@ -71,7 +71,7 @@ public PlannerTestAideDeCamp(URL ddlurl, String basename) throws Exception {
         assert(ddlurl != null);
         String pth = ddlurl.getPath();
         String schemaPath = URLDecoder.decode(ddlurl.getPath(), "UTF-8");
-        VoltCompiler compiler = new VoltCompiler();
+        VoltCompiler compiler = new VoltCompiler(false);
```
slowsoul: I think we could have added a VoltCompiler constructor that takes no parameter and call this(false, false), that can greatly reduce the number of changed lines.

src/frontend/org/voltdb/compiler/DeploymentBuilder.java
```diff
@@ -418,6 +426,14 @@ public String getXML() {
         deployment.setExport(export);
         export.setEnabled(m_elenabled);
 
+        // <dr>
+        if (m_drRole != DrRoleType.NONE) {
+            final DrType drType = factory.createDrType();
+            deployment.setDr(drType);
+            drType.setRole(m_drRole);
+            drType.setId(1);
```
nshi: Yes. It's only used for tests. I just picked a random number so that the ID is not left empty.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -670,14 +672,15 @@ public static String compileDeployment(Catalog catalog,
             // set the HTTPD info
             setHTTPDInfo(catalog, deployment.getHttpd());
 
+            setDrInfo(catalog, deployment.getDr(), deployment.getCluster());
```
nshi: Unfortunately, no. `setExportInfo()` below will generate additional export configuration for DR conflict tables if XDCR is enabled, so I have to move this line here to set the DR role properly.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1980,6 +1985,7 @@ private static void setDrInfo(Catalog catalog, DrType dr, ClusterType clusterTyp
                 hostLog.info("Configured connection for DR replica role to host " + drSource);
             }
         } else {
+            cluster.setDrrole(DrRoleType.NONE.value());
```
nshi: No. The enum class `DrRoleType.value()` actually returns lower case string for each enum.

tests/frontend/org/voltdb/planner/PlannerTestAideDeCamp.java
nshi: I debated that but decided to go with this way because I wanted to have all the places that constructs a VoltCompiler to be aware that the compiler result actually depends on the DR role. I was afraid that if we have a default constructor, other people may not realize that DR role has an impact and may forget to pass it in.

PR #4203, opened 2017-01-18
ENG-11827: Expand user dir

akhanzode:


lib/python/voltcli/voltdb.d/start.py
```diff
@@ -43,7 +45,8 @@
 )
 def start(runner):
     if runner.opts.directory_spec:
-        runner.args.extend(['voltdbroot', runner.opts.directory_spec])
+        upath = os.path.expanduser(runner.opts.directory_spec)
```
xinjiacs: @akhanzode 
Do you care about any input space, like "   ~/my_dir"? 
Will this work?

akhanzode: this is exactly to address ~/dir situation are you asking if someone gives it in ""?

xinjiacs: "~/mydir" is a little different from "     ~/mydir"? Not sure if that's possible. I know "os.path.expanduser" does not work with any dir leading without "~".

akhanzode: That would not be a userdir expandable dir.

PR #4204, opened 2017-01-19
Fix ENG-11172 invalid boolean and/or right operand causes NPE

yzhang1991:


tests/testprocs/org/voltdb_testprocs/regressionsuites/fixedsql/sql-update-ddl.sql
```diff
@@ -150,3 +150,21 @@ CREATE VIEW P1_VIEW (ID, ID_COUNT, NUM_SUM) AS
...
+  DEC DECIMAL, 
+  VCHAR VARCHAR(500), 
+  VCHAR_INLINE_MAX VARCHAR(63 BYTES), 
+  VCHAR_INLINE VARCHAR(14), 
+  TIME TIMESTAMP, 
+  VARBIN VARBINARY(100), 
+  POINT GEOGRAPHY_POINT, 
+  POLYGON GEOGRAPHY, 
+  PRIMARY KEY (ID) 
+);
```
paulmartel: Seems a little overboard to define all of these columns, when just the single TIME TIMESTAMP is used, but maybe it's a reasonable investment and we will be able to reuse it as we need other column types for future tests .

yzhang1991: Yeah, I was actually looking for an existing table in the DDL that can be used but failed.
This can be used for future test cases.

PR #4205, opened 2017-01-19
DML statements with expression subqueries

alexeevm:


tests/frontend/org/voltdb/planner/TestPlansDML.java
```diff
@@ -262,6 +263,90 @@ public void testDMLPredicate() {
...
+            node = node.getChild(0);
+        }
+        assertNotNull(node);
+
+        // Verify DML Predicate
+        if (filterType != null) {
+            AbstractExpression predicate = ((SeqScanPlanNode) node).getPredicate();
+            assertNotNull(predicate);
+            assertEquals(filterType, predicate.getExpressionType());
+            if (hasExpressionSubquery) {
```
paulmartel: hasExpressionSubquery is never false here. Eliminate this flag.

tests/frontend/org/voltdb/planner/TestPlansDML.java
```diff
@@ -262,6 +263,90 @@ public void testDMLPredicate() {
...
+                "select A,C from R2 " +
+                "where not exists (select A from P1 AP1 where R2.A = AP1.A);",
+                "Subquery expressions are only supported for single partition procedures and AdHoc queries referencing only replicated tables.");
+
+        // Distributed expression subquery with inferred partitioning
+        failToCompile("DELETE FROM R1 WHERE C > ALL (SELECT A FROM P2);", PlanAssembler.IN_EXISTS_SCALAR_ERROR_MESSAGE);
+
+    }
+
+    void checkDMLPlanNodeAndSubqueryExpression(AbstractPlanNode pn, PlanNodeType dmlPlanNodeType, ExpressionType filterType, boolean hasExpressionSubquery) {
```
paulmartel: Callers could be simplified by changing the first argument to "String statement" and moving the following boilerplate into this function:

    AbstractPlanNode pn = compileToFragments(statement).get(1).getChild(0));

You could also test that the statement starts with the resulting planNodeType's symbol string -- and then not have to send in an expected dmlPlanNodeType.

tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -206,6 +214,10 @@ public void testNonSupportedUnions()
         // nonsense syntax in place of union ops (trying various internal symbol names meaning n/a)
         failToCompile("select A from T1 NOUNION select B from T2");
         failToCompile("select A from T1 TERM select B from T2");
+        failToCompile("(select A from T1 UNION select B from T2) where A in (select A from T2)");
```
paulmartel: Add comment: why this is expected to fail.

src/frontend/org/voltdb/plannodes/MaterializePlanNode.java
```diff
@@ -49,6 +53,13 @@ public void generateOutputSchema(Database db) {
         // MaterializePlanNodes have no children
         assert(m_children.size() == 0);
         // MaterializePlanNode's output schema is pre-determined, don't touch
+        // except when its output column(s) has a scalar expression
```
paulmartel: change comment to say "scalar subquery expression".

tests/frontend/org/voltdb/planner/TestPlansDML.java
alexeevm: But then you would have to specify whether the statement should produce a SP or MP plan, right?  
`AbstractPlanNode pn = compileToFragments(statement).get(1).getChild(0));`
won't work for a SP plan
For example, two DELETE statements against replicated and partitioned tables using their PK
1> explain delete from r1 where id = 3;
EXECUTION_PLAN                 
 LIMIT 1
  RECEIVE FROM ALL PARTITIONS
    DELETE R1
     INDEX SCAN of "R1" using its primary key index
      uniquely match (ID = ?0)

2> explain delete from p1 where id = 3;
EXECUTION_PLAN                 
DELETE P1
 INDEX SCAN of "P1" using its primary key index
  uniquely match (ID = ?0)

tests/frontend/org/voltdb/planner/TestPlansDML.java
alexeevm: But then you would have to specify whether the statement should produce a SP or MP plan, right?  
`AbstractPlanNode pn = compileToFragments(statement).get(1).getChild(0));`
would not work for a SP plan
For example, two DELETE statements against replicated and partitioned tables using their PK

DELETE FROM R1 WHERE ID = 3;

 LIMIT 1
  RECEIVE FROM ALL PARTITIONS
    DELETE R1
     INDEX SCAN of "R1" using its primary key index
      uniquely match (ID = ?0)

and 

DELETE FROM  P1 WHERE ID = 3;

DELETE P1
 INDEX SCAN of "P1" using its primary key index
  uniquely match (ID = ?0)

tests/frontend/org/voltdb/planner/TestPlansDML.java
alexeevm: But then you would have to specify whether the statement should produce a SP or MP plan, right?  
`AbstractPlanNode pn = compileToFragments(statement).get(1).getChild(0));`
would not work for a SP plan
For example, two DELETE statements against replicated and partitioned tables using their PK

DELETE FROM R1 WHERE ID = 3;

 LIMIT 1
  RECEIVE FROM ALL PARTITIONS
    DELETE R1
     INDEX SCAN of "R1" using its primary key index
      uniquely match (ID = ?0)

and 

DELETE FROM  P1 WHERE ID = 3;

DELETE P1
 INDEX SCAN of "P1" using its primary key index
  uniquely match (ID = ?0)

Instead of passing a plan node and an expected type you would have to pass a String and an index of a fragment we need to verify. Do you think it will simplify the code?

tests/frontend/org/voltdb/planner/TestPlansDML.java
alexeevm: Deleted

tests/frontend/org/voltdb/planner/TestUnion.java
alexeevm: Added the comment
"invalid syntax - the WHERE clause is illegal"

src/frontend/org/voltdb/plannodes/MaterializePlanNode.java
alexeevm: Corrected

tests/frontend/org/voltdb/planner/TestPlansDML.java
paulmartel: I see your point. I missed that you were sometimes  sending `compileToFragments(statement).get(0).getChild(0))` for SP cases and sometimes `compileToFragments(statement).get(1).getChild(0))` for MP cases. So, my idea to internalize this step could still work, but only if you used "the last fragment" in all cases, like

    pns = compileToFragments(statement);
    pn = pns.get(pns.size()-1).getChild(0);

This version of the test wouldn't require you to predict if each statement is SP or MP -- it wouldn't care. I'm not sure if this is a good thing or a bad thing. It's not a big deal to me, either way. I leave it up to you. What do you think?

The rest of the changes look good. I've submitted them to the branch for system testing before merge w/ master.

alexeevm:
Paul,

It does look cleaner with your suggestion. I just created a new pull
request.

https://github.com/VoltDB/voltdb/pull/4236

Mike

paulmartel:
Merged to master via another branch.

PR #4207, opened 2017-01-19
ENG-11594: Guard against nothing on deque and PBD

akhanzode:


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -315,12 +315,13 @@ public synchronized void updateAckMailboxes(final Pair<Mailbox, ImmutableList<Lo
 
     private void releaseExportBytes(long releaseOffset) throws IOException {
         // if released offset is in an already-released past, just return success
-        if (!m_committedBuffers.isEmpty() && releaseOffset < m_committedBuffers.peek().uso()) {
+        if (!m_committedBuffers.isEmpty() && (m_committedBuffers.peek() != null)
```
xinjiacs: What happens when "m_committedBuffers.peek() == null"?

xinjiacs: Can you do early return on " If (m_committedBuffers.isEmpty())"? Then you do not need to change them in the next while loop?

akhanzode: we just wait for export push to offer() and then when release comes after we cleanup. This is a very edge case and not easy to reproduce.

xinjiacs: All right, I am familiar with code base. But this PR seems to be better on NULL check case. ""m_committedBuffers.peek() == null"".

akhanzode: Didnt quite follow what you mean.

PR #4209, opened 2017-01-19
ENG-11840: Make DR stats column names consistent.

nshi:


slowsoul:
Looks good to me. It's just the DR producer and consumer stats name already kind of diverged, producer more often doesn't use underscore, while consumer always uses underscore. At least we can start to make them consistent beginning from cluster id fields as in this PR.

PR #4222, opened 2017-01-24
ENG-11779: Fix refineValueTypes for CASE WHEN expression

wolffcm:
CASE WHEN isn't like other operators, which are mostly binary or unary.

RHS of the CASE WHEN operator is the OPERATOR_ALTERNATIVE, which is what really helps define type of output, so defer to RHS when refining.


paulmartel:
Looks good.

PR #4224, opened 2017-01-24
Eng 11764

akhanzode:


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -557,6 +557,7 @@ public void run() {
     }
 
     private void handleChildUpdate(final WatchedEvent event, final HostMessenger messenger) {
+        if (shutdown || m_drainedSources.get() == m_numSources) return;
```
nshi: Not sure I quite understand this line.

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -574,7 +575,11 @@ public void run() {
                         try {
                             if (shutdown) return;
                             KeeperException.Code code = KeeperException.Code.get(rc);
-                            if (code != KeeperException.Code.OK) {
+                            //Other node must have drained so ignore.
+                            if (code == KeeperException.Code.NONODE) {
+                                exportLog.info("Path not found generation drain most likely finished on other node: " + path);
```
nshi: This might be too noisy. Is it informative to let the user know this? If not, I think it should be debug.

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -749,6 +754,17 @@ private void cleanup(final HostMessenger messenger) {
         shutdown = true;
         //We need messenger NULL guard for tests.
         if (m_mbox != null && messenger != null) {
+            for (Integer partition : m_dataSourcesByPartition.keySet()) {
+                final String partitionDN =  m_mailboxesZKPath + "/" + partition;
+                String path = partitionDN + "/" + m_mbox.getHSId();
+                try {
+                    messenger.getZK().delete(path, 0);
```
nshi: Do you want to use recursive deletion here in case there are children nodes?

akhanzode: This does not have any children.

src/frontend/org/voltdb/export/ExportGeneration.java
akhanzode: Yeah I can make it debug should only happen if generation rolls on other node faster than this node and sees callback fired before this node shuts down the generation.

src/frontend/org/voltdb/export/ExportGeneration.java
akhanzode: This is if we are shutting down or fully drained and watcher is fired we dont need to create and setup watch as we will do our cleanup and delete the generation object.

PR #4225, opened 2017-01-25
extended-config.py: just reduced the numeric precision of the

johnTcrawford:
numeric-decimals test suite, from 9 to 8 (significant digits), because
Ruth noticed that it was failing often, with tiny decimal differences.


wolffcm:
Looks good.

PR #4227, opened 2017-01-25
Make system build on GCC 6.x and on U16.10.

jhugg:
Also make ee tests build in non-debug mode.

### DIFF
--- a/tests/ee/storage/persistenttable_test.cpp
+++ b/tests/ee/storage/persistenttable_test.cpp
@@ -43,6 +43,8 @@
 
 #include "boost/scoped_ptr.hpp"
 
+#define _unused(x) ((void)(x))
+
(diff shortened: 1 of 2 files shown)


tests/ee/storage/persistenttable_test.cpp
```diff
@@ -362,6 +364,7 @@ TEST_F(PersistentTableTest, TruncateTableTest) {
     const int tuplesToInsert = 10;
     added = tableutil::addRandomTuples(table, tuplesToInsert);
     assert(added);
+    _unused(added);
```
wolffcm: Can't you just include common/FixUnusedAssertHack.h to fix this?

jhugg: Oh. didn't know about that. I suppose so.

PR #4228, opened 2017-01-25
Eng 11175 check columns resolved cast exception

yzhang1991:


src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -779,22 +779,29 @@ String getTableName() {
         return "";
     }
 
-    static void checkColumnsResolved(HsqlList set) {
+    static void checkUnresolvedColumns(HsqlList set) {
```
yzhang1991: This function is used in multiple places, some of which are not in the context of a "WHERE" clause.
But those cases are not supported by VotlDB yet:
* CALL statement
* CREATE TRIGGER statement
* SET statement
* MERGE statement

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -782,19 +782,26 @@ String getTableName() {
     static void checkColumnsResolved(HsqlList set) {
```
yzhang1991: This function is used in multiple places, some of which are not in the context of a "WHERE" clause.
But some of those cases are not supported by VotlDB yet:
* CALL statement
* CREATE TRIGGER statement
* SET statement
* MERGE statement

The rest, including:
* CASE WHEN
* INSERT

are not affected by this change.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionColumn.java
```diff
@@ -782,19 +782,26 @@ String getTableName() {
...
+                    sb.append(e.tableName + '.');
+                }
 
-            throw Error.error(ErrorCode.X_42501,
-                              sb.toString() + e.getColumnName());
+                throw Error.error(ErrorCode.X_42501,
+                                  sb.toString() + e.getColumnName());
+            }
+            else {
+                assert(obj instanceof ExpressionAggregate);
```
wolffcm: Why must it be an aggregate if it's not a column?  I'd like to understand this change.

yzhang1991: ExpressionAggregates directly put themselves into the unresolved list in its overriding resolveColumnReferences() method. In a normal aggregate query context, this list will be handled by QuerySpecification to remove the aggregate columns before hitting checkColumnsResolved(). But in a WHERE clause we do not have this step because aggregate columns should not appear in WHERE clauses.

This is the only case non-column expression can get into the list.

wolffcm:
Looks good to me.

PR #4231, opened 2017-01-27
ENG-11648 accelerate regression suites

yzhang1991:
Waiting for Jenkins to verify this works.

On my own Macbook, TestMaterializedViewSuite took 215 seconds to run on master, but it took only 52 seconds on this branch (**75.8%** improvment).

Jenkins found several failures, will tune this branch a little bit to address them.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
+++ b/tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
@@ -28,11 +28,11 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.voltdb.utils.MiscUtils;
+
 import junit.framework.Test;
 import junit.framework.TestCase;
 import junit.framework.TestSuite;
-
-import org.voltdb.utils.MiscUtils;
 
 /**
  * A subclass of TestSuite that multiplexes test methods across
@@ -145,7 +145,8 @@ public class MultiConfigSuiteBuilder extends TestSuite {
 
         // add a test case instance for each method for the specified
         // server config
-        for (String mname : methods) {
+        for (int i = 0; i < methods.size(); i++) {
+            String mname = methods.get(i);
             RegressionSuite rs = null;
             try {
                 rs = (RegressionSuite) cons.newInstance(mname);
@@ -154,6 +155,9 @@ public class MultiConfigSuiteBuilder extends TestSuite {
                 return false;
             }
             rs.setConfig(config);
+            // The last test method for the current cluster configuration will need to
+            // shutdown the cluster completely after finishing the test.
+            rs.m_completeShutdown = (i == methods.size() - 1);
             super.addTest(rs);
         }
 
--- a/tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
+++ b/tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
@@ -56,7 +56,7 @@ public class MultiConfigSuiteBuilder extends TestSuite {
      * @return A list of the names of each JUnit test method.
      */
     static List<String> getTestMethodNames(Class<? extends TestCase> testCls) {
-        ArrayList<String> retval = new ArrayList<String>();
+        ArrayList<String> retval = new ArrayList<>();
 
         for (Method m : testCls.getMethods()) {
             if (m.getReturnType() != void.class)
@@ -87,6 +87,10 @@ public class MultiConfigSuiteBuilder extends TestSuite {
      * @param config A Server Configuration to run this set of tests on.
      */
     public boolean addServerConfig(VoltServerConfig config) {
+        return addServerConfig(config, true);
+    }
+
+    public boolean addServerConfig(VoltServerConfig config, boolean reuseServer) {
 
         // near silent skip on k>0 and community edition
         if (!MiscUtils.isPro()) {
@@ -157,7 +161,7 @@ public class MultiConfigSuiteBuilder extends TestSuite {
             rs.setConfig(config);
             // The last test method for the current cluster configuration will need to
             // shutdown the cluster completely after finishing the test.
-            rs.m_completeShutdown = (i == methods.size() - 1);
+            rs.m_completeShutdown = ! reuseServer || (i == methods.size() - 1);
             super.addTest(rs);
         }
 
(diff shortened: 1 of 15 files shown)


tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
```diff
@@ -28,12 +28,12 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import org.voltdb.utils.MiscUtils;
+
```
paulmartel: Maybe you should reconfigure your eclipse to keep `import junit.*` before `import org.voltdb.*` to avoid this kind of thrashing.

tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
```diff
@@ -154,6 +155,9 @@ public boolean addServerConfig(VoltServerConfig config) {
                 return false;
             }
             rs.setConfig(config);
+            // The last test method for the current cluster configuration will need to
+            // shutdown the cluster completely after finishing the test.
+            rs.m_completeShutdown = (i == methods.size() - 1);
```
paulmartel: I had to look at the JUnit TestSuite source code to confirm that the test run order is the same as the addTest order, but that appears to be the case -- and is unlikely to change in future versions?

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -110,7 +114,25 @@ public void setUp() throws Exception {
...
+        }
+        else {
+            Client client = getClient();
+            VoltTable tableList = client.callProcedure("@SystemCatalog", "TABLES").getResults()[0];
+            ArrayList<String> tableNames = new ArrayList<String>(tableList.getRowCount());
+            int tableNameColIdx = tableList.getColumnIndex("TABLE_NAME");
+            int tableTypeColIdx = tableList.getColumnIndex("TABLE_TYPE");
+            while (tableList.advanceRow()) {
+                String tableType = tableList.getString(tableTypeColIdx);
+                if (tableType.equalsIgnoreCase("TABLE")) {
```
paulmartel: I don't think that this check will cover the case of a view on a stream which will also need to be truncated to get back to a clean database state. We just want to ignore things that can't be truncated like streams and normal views. This code could use a comment explaining exactly what it is trying to allow and to disallow.

tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
yzhang1991: I also checked this. I think this is OK.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
yzhang1991: Thanks for reminding me that streamed views are editable.

tests/frontend/org/voltdb/regressionsuites/MultiConfigSuiteBuilder.java
yzhang1991: I will look into Eclipse preferences for this option and will revert this change.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -116,7 +120,28 @@ public void setUp() throws Exception {
...
+        }
+        else {
+            Client client = getClient();
+            VoltTable tableList = client.callProcedure("@SystemCatalog", "TABLES").getResults()[0];
+            ArrayList<String> tableNames = new ArrayList<>(tableList.getRowCount());
+            int tableNameColIdx = tableList.getColumnIndex("TABLE_NAME");
+            int tableTypeColIdx = tableList.getColumnIndex("TABLE_TYPE");
+            while (tableList.advanceRow()) {
+                String tableType = tableList.getString(tableTypeColIdx);
+                if (! tableType.equalsIgnoreCase("EXPORT")) {
```
wolffcm: What about views, don't you want to skip them too?  Is "view" a table type?  I'm not sure.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -116,7 +120,28 @@ public void setUp() throws Exception {
...
+                    tableNames.add(tableList.getString(tableNameColIdx));
+                }
+            }
+            for (String tableName : tableNames) {
+                try {
+                    client.callProcedure("@AdHoc", "TRUNCATE TABLE " + tableName);
+                }
+                catch (Exception ex) {}
```
wolffcm: If the truncate fails, you could fail the test with a helpful message.  If we fail to truncate, any resulting test failures will be confusing and hard to debug.

tests/frontend/org/voltdb/regressionsuites/statistics/TestStatisticsSuiteDatabaseElementStats.java
```diff
@@ -211,6 +211,7 @@ public void testProcedureStatistics() throws Exception {
         // this plus R/W replication should ensure that every site on every node runs this transaction
         // at least once
 
+        client.callProcedure("@Statistics", "proceduredetail", 1);
```
wolffcm: Curious why you had to dd this call?

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
yzhang1991: It is just not easy to tell if a view is a streamed view
We need to truncate the streamed view as well.

tests/frontend/org/voltdb/regressionsuites/statistics/TestStatisticsSuiteDatabaseElementStats.java
yzhang1991: I want to set the starting point for the incremental stats.
Otherwise the stats will carry the numbers from the previous test case and fail this one.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
yzhang1991: Here I want to ignore the error message for truncating the persistent table views. Because from the stats I have no way to tell whether the view is defined on a stream or a persistent table.

wolffcm: I see.  In that case, you could check for an expected "truncating a view not allowed" message, and fail the test otherwise.

yzhang1991:
I made the change to throw exceptions that are unrelated to truncating the persistent table views.

PR #4232, opened 2017-01-27
ENG-11918: Do range checking when casting timestamps

wolffcm:


tests/testprocs/org/voltdb_testprocs/regressionsuites/fixedsql/sql-update-ddl.sql
```diff
@@ -150,3 +150,10 @@ CREATE VIEW P1_VIEW (ID, ID_COUNT, NUM_SUM) AS
 SELECT ID, COUNT(*), SUM(NUM)
 FROM P1
 GROUP BY ID;
+
+CREATE TABLE ENG_11918 (
+  ID INTEGER,
+  INT INTEGER,
+  VCHAR VARCHAR(64),
```
wolffcm: The original repro for this had an inline varchar field, but I found that if I set the "bytes" flag on the column declaration the repro hit a separate issue in the planner.  I'll file a separate bug for this which is not a 7.0 blocker.

PR #4241, opened 2017-01-31
ENG-11927: Add check only for non-persisted blocks.

akhanzode:


src/frontend/org/voltdb/export/StreamBlockQueue.java
```diff
@@ -323,8 +323,18 @@ public TruncatorResponse parse(BBContainer bbc) {
 
     @Override
     public void finalize() {
-        if (!m_memoryDeque.isEmpty()) {
-            exportLog.error("Finalized StreamBlockQueue with items in the memory deque");
+        try {
+            if (!m_memoryDeque.isEmpty()) {
+                m_memoryDeque.stream().filter((block) -> (!block.isPersisted())).forEachOrdered((_item) -> {
+                    exportLog.error("Finalized StreamBlockQueue with items in the memory deque that are not persisted.");
```
nshi: Maybe logging the count instead of the same message for each item would be sufficient?

akhanzode: Yes I can do one time log with count.

PR #4243, opened 2017-01-31
ENG-11733: Use CRC instead of table signature.

akhanzode:


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -149,8 +150,12 @@ public void run() {
         };
         m_database = db;
         m_tableName = tableName;
+        m_signature = signature;
+        m_signatureBytes = m_signature.getBytes(Constants.UTF8ENCODING);
```
vtkstef: maybe use the nio StandardCharsets constant instead

PR #4254, opened 2017-02-02
Remove even the machinery behind logging when using SQLCommand

jhugg:


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -1232,6 +1232,8 @@ private static String extractArgInput(String arg) {
     // Application entry point
     public static void main(String args[])
     {
+        System.setProperty("voltdb_client_app", "true");
```
zeemanhe: This property is not needed.

src/frontend/org/voltcore/logging/VoltLogger.java
```diff
@@ -354,6 +363,13 @@ public static void configure(String xmlConfig) {
      * @param classname The id of the logger.
      */
     public VoltLogger(String classname) {
+        // quick out for when we want no logging whatsoever
+        if ("true".equals(System.getProperty("voltdb_no_logging"))) {
+            m_logger = new VoltNullLogger();
+            return;
+        }
+
+        boolean isClientApp = "true".equals(System.getProperty("voltdb_client_app"));
```
zeemanhe: This check is not needed.

PR #4257, opened 2017-02-03
ENG-11866:

wweiss-voltdb:
1. Move initialization of DR ConsumerGateway to the cluster initialization time. This then allows the producer to leak connections and start cursors to the cluster even if no data source has been specified.
2. Enable breaking of producer and consumer connections to isolate a ServerThread or LocalCluster host with hasLocalServer set to true.
3. Change QueryInterval in DRConnectionInterface from static final to instance final so that it can be set to a low value for unit tests that use ServerThread. This accelerates unit tests that require discovery to detect rejoined or recovered hosts.
3. Add a unit test to multiCluster to verify that a joiner cluster and a network isolated mesh leader cluster can find eachother and sync up after the network isolation has been removed.

### DIFF
--- a/src/frontend/org/voltdb/ConsumerDRGateway.java
+++ b/src/frontend/org/voltdb/ConsumerDRGateway.java
@@ -37,6 +37,15 @@ public interface ConsumerDRGateway extends Promotable {
 
     Map<Byte, DRRoleStats.State> getStates();
 
+    /**
+     * If this cluster was a joiner (at some point) dataSourceCluster will not be -1. If this is the case
+     * the trackers for this clusterId must exist. If they don't exist, this cluster did not finish loading
+     * the snapshot as a joiner.
+     * @param dataSourceCluster
+     * @param expectedClusterMembers
+     * @return false if this cluster is a joiner and the sync snapshot did not finish loading form the
+     *         leader cluster
+     */
     boolean isSyncSnapshotComplete(byte dataSourceCluster, List<MeshMemberInfo> expectedClusterMembers);
 
     void initialize(boolean resumeReplication);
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/ConsumerDRGateway.java
```diff
@@ -37,7 +37,9 @@
 
     Map<Byte, DRRoleStats.State> getStates();
 
-    void initialize(boolean resumeReplication, final byte dataSourceCluster, List<MeshMemberInfo> expectedClusterMembers);
+    boolean isSyncSnapshotComplete(byte dataSourceCluster, List<MeshMemberInfo> expectedClusterMembers);
```
manjujames: It might be nice to have javadocs here with a short explanation of what this method does.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3872,20 +3863,13 @@ private boolean shouldInitiatorCreateMPDRGateway(Initiator initiator) {
     }
 
     private boolean createDRConsumerIfNeeded() {
-        if (!m_config.m_isEnterprise
-                || (m_consumerDRGateway != null)
-                || !m_catalogContext.cluster.getDrconsumerenabled()) {
```
manjujames: I forget now, where is this flag value used now? That is, if the user explicitly disables consumer using this flag, it is honored where?

src/frontend/org/voltdb/ConsumerDRGateway.java
wweiss-voltdb: Sure. I can add that.

src/frontend/org/voltdb/RealVoltDB.java
wweiss-voltdb: You make a good point but I don't have a good answer. We need to keep both gateways up all the time so that we learn about connections when we are an isolated cluster.  Perhaps we should discuss this tomorrow. I am inclined to handle the flag in the ConsumerGateway like we do in the DrProducer.

PR #4260, opened 2017-02-04
Fix ENG-11935 SIGSEGV crash on string comparison

yzhang1991:
If we have an expression in an index definition, and the data type for that expression is `VARCHAR` or `VARBINARY`, we will set:
```c++
// tableindexfactory.cpp, line 269
declaredLength = TupleSchema::COLUMN_MAX_VALUE_LENGTH;
```
This will make this column **OUTLINED** in the *index key tuple schema*.
However, in **single table** materialized views, when we use an index to locate the MIN/MAX fallback value, we use a StandAloneTupleStorage to hold a temporary search key tuple, and that tuple still uses the original *table tuple schema*. The *table tuple schema* and the *index key tuple schema* may conflict each other in the inline/outline setting. Therefore, during index lookup, the comparison function may retrieve an inline NValue in a outlined way according to the information stored in the *index key tuple schema* and cause an SIGSEGV exception.

This fix uses a dedicated search key tuple which follows the index search key schema for index lookups. This dedicated search key tuple was only used by the case where the selected index covers both GROUP-BY columns and the MIN/MAX column, now it is also used by the sub-optimal case where the selected index only covers the GROUP-BY columns.

This seems to be a bug that has existed since forever, I will do a double check.


src/ee/storage/MaterializedViewTriggerForWrite.cpp
```diff
@@ -172,13 +172,12 @@ void MaterializedViewTriggerForWrite::allocateMinMaxSearchKeyTuple() {
...
     BOOST_FOREACH(TableIndex *index, m_indexForMinMax) {
-        // Because there might be a lot of indexes, find the largest space they may consume
-        // so that they can all share one space and use different schemas. (ENG-8512)
-        if (minMaxIndexIncludesAggCol(index, m_groupByColumnCount)) {
-            nextIndexStoreLength = index->getKeySchema()->tupleLength() + TUPLE_HEADER_SIZE;
-            if (nextIndexStoreLength > minMaxSearchKeyBackingStoreSize) {
-                minMaxSearchKeyBackingStoreSize = nextIndexStoreLength;
-            }
+        if (! index) {
+            continue;
```
wolffcm: I wonder if the other the other callsite for `minMaxIndexIncludesAggCol` can be optimized in the same way, removing the need for it altogether?

src/ee/storage/MaterializedViewTriggerForWrite.cpp
```diff
@@ -259,7 +258,7 @@ NValue MaterializedViewTriggerForWrite::findMinMaxFallbackValueIndexed(const Tab
     }
     else {
         // Use sub-optimal index (only group-by columns).
-        selectedIndex->moveToKey(&m_searchKeyTuple, minMaxCursor);
```
wolffcm: I see there's other uses of `m_searchKeyTuple` in `MateriazliedViewTriggerForInsert`.  Are these uses safe, or could a corresponding problem exist there too?

yzhang1991: `m_searchKeyTuple` is used for index lookup on `m_index` in `findExistingTuple()`.
The schemas are identical:
```c++
// MaterializedViewTriggerForInsert.cpp, line 222
m_searchKeyTuple = TableTuple(m_index->getKeySchema());
```
So it is safe.

src/ee/storage/MaterializedViewTriggerForWrite.cpp
yzhang1991: No, we have different logic for indexes that containing both GROUP-BY columns and MIN/MAX columns and indexes only covering GROUP-BY columns.
We still need this `minMaxIndexIncludesAggCol()` function for selecting different lookup logics.

PR #4262, opened 2017-02-07
Add PROCEDUREDETAIL selector

yzhang1991:
In this pull request, I added the `PROCEDUREDETAIL` selector and it will now print the same output as the `PROCEDURE` selector.

The current output of the `PROCEDURE` selector is the base table for `PROCEDUREINPUT`, `PROCEDUREOUTPUT`, and `PROCEDUREPROFILE`.


yzhang1991:
I have another branch to replace this.

PR #4267, opened 2017-02-07
TLS refactor without netty changes

vtkstef:


src/frontend/org/voltcore/network/CipherExecutor.java
```diff
@@ -0,0 +1,244 @@
...
+    public final static int FRAME_SHIFT = 14; // 16384 (max TLS fragment)
+    public final static int FRAME_SIZE = 1 << FRAME_SHIFT;
+
+    volatile ListeningExecutorService m_es;
+    AtomicBoolean m_active = new AtomicBoolean(false);
+    final int m_threadCount;
+
+    private CipherExecutor(int nthreads) {
+        m_threadCount = nthreads;
+        m_es = CoreUtils.LISTENINGSAMETHREADEXECUTOR;
```
nshi: I wonder if it makes sense to reuse the computation service executor in RealVoltDB instead of creating new threads.

src/frontend/org/voltcore/network/CipherExecutor.java
```diff
@@ -0,0 +1,244 @@
...
+    final public <T> ListenableFuture<T> submit(Callable<T> c) {
+        try {
+            return m_es.submit(c);
+        } catch (RejectedExecutionException e) {
+            return CoreUtils.LISTENINGSAMETHREADEXECUTOR.submit(c);
+        }
+    }
+
+    public void startup() {
+        if (m_active.compareAndSet(false, true)) synchronized(this) {
```
nshi: Why the atomic boolean and synchronization? Why not just synchronize and use a boolean inside?

src/frontend/org/voltcore/network/EncryptFrame.java
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Container and tracking class used during message encryptions
+ */
+class EncryptFrame {
+
+    final static int UNCHUNKED_SIGIL = -1;
+    /**
+     * chunk number in a sequence of chunks. If a message is split into 3 chunks
+     * then their respective chunk numbers would be 1, 2, and 3. Each chunk
```
nshi: Any reason that it's not zero-based?

src/frontend/org/voltcore/network/EncryptFrame.java
```diff
@@ -0,0 +1,119 @@
...
+     * chunks value would be 3
+     */
+    final int chunkno;
+    /**
+     * chunk number in a sequence of chunks. If a message is split into 3 chunks
+     * then their respective chunk numbers would be 1, 2, and 3. Each chunk
+     * chunks value would be 3
+     */
```
nshi: Wrong comment.

src/frontend/org/voltcore/network/EncryptFrame.java
vtkstef: because the last chunk test is is chunkno == chunks also when you have a three part tv miniseries you say part 1 part 2 and part 3. I am following semantic common practice

src/frontend/org/voltcore/network/CipherExecutor.java
vtkstef: another twist on double checked singletons that support active/non-active states. In my many hours of debugging TestSSL and TestClientClose that bring up and tear down connections intensively I found that this approach yielded the results I  was aiming for

src/frontend/org/voltcore/network/CipherExecutor.java
vtkstef: that would introduce as cyclical dependency between realvoltdb and CipherExecutor

src/frontend/org/voltcore/network/EncryptFrame.java
vtkstef: ok

PR #4269, opened 2017-02-07
Eng 10474 Use indexes when evaluating window functions.

BillWhite:
There some themes in this request.

1. I replaced some constant strings "VOLT_TEMP_TABLE" with references to a single constant string in AbstractParsedStatement.
2. I pretty much blew up the function SubPlanAssembler.determineIndexOrdering.  I now keep a scoreboard with one score for each window function, and one for the statement level order by if there is one.  I apply each index expression (or column reference) to each of these scores to decide if the index expression (or column reference) matches, is an order spoiler or causes the score to be unsuitable.  When all index expressions (or column references) have been applied, I extract the result, which is a set of order spoilers and an access path.  The access path knows whether the index is usable for the window function or the statement level order by or both.  It also knows the sequence of expressions forced by the index.  (Think of partition by clauses in window functions, where the order of the partition expressions is not fixed.)  I didn't change the use of the order spoilers or the access path, so the caller to determineIndexOrdering should be pretty much unchanged.
3. The SubPlanAssembler chooses a join order and an access path for each table scan and tries to generate a plan for these choices.  This plan may contain index scans or sequential scans and perhaps some joins.  For the scans, we window function and order by information into the scan node.  For joins we propagate information from the outermost scan up the join tree.
4. When deciding to generate a WindowFunctionPlanNode in PlanAssembler.handleWindowFunction, I find the subplan, which is the highest scan or join node in the tree, and decide whether to generate an order by for the window function.  I need to generate an order by node if the plan is a two fragment plan, even if there is an index which supplies the order.  This is because there is a micro-optimization which turns a RECEIVE node into a MERGERECEIVE node with an inline ORDERBY node, and I need something to hang the ORDERBY expressions on.
5. The RECEIVE-to-MERGERECEIVE micro-optimization needs a special optimization to turn a WINDOWFUNCTION->ORDERBY->RECEIVE sequence into WINDOWFUNCTION->MERGERECEIVE(ORDERBY), where the second ORDERBY is inlined in the MERGERECEIVE node.
6. A couple of regression tests were made faster by explicitly combining former test points into one larger test point, saving the expense of restarting a server.


tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -420,6 +420,284 @@ static protected void assertClassesMatchNodeChain(
         assertTrue("Actual plan longer than expected", pn == null);
     }
 
+    /**
```
BillWhite: There are really no changes here.  This body of code was at the end of the class, which will cause some confusion when I refactor the C++ unit test generation code away from this class.  So I just swapped them.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
```diff
@@ -544,11 +545,539 @@ public void testExplainPlanText() {
         String windowedQuery = "SELECT RANK() OVER (PARTITION BY A ORDER BY B DESC) FROM AAA;";
         AbstractPlanNode plan = compile(windowedQuery);
         String explainPlanText = plan.toExplainPlanString();
-        String expected = "WindowFunctionPlanNode: ops: AGGREGATE_WINDOWED_RANK()";
+        String expected = "WINDOW FUNCTION AGGREGATION: ops: AGGREGATE_WINDOWED_RANK()";
         assertTrue("Expected to find \"" + expected + "\" in explain plan text, but did not:\n"
                 + explainPlanText, explainPlanText.contains(expected));
     }
 
+    // These two can be used to disable particular tests.
```
BillWhite: This is really the important test here to bind specific index usage to statements.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -554,18 +555,20 @@ static protected void validateRowOfLongs(VoltTable vt, long [] expected) {
      *
      * @param input
      */
-    static protected void shuffleArrayOfLongs(long [][] input) {
```
BillWhite: Generalize this.  I wanted to be able to shuffle arrays of anything.

tests/frontend/org/voltdb/regressionsuites/TestOrderBySuite.java
```diff
@@ -1078,27 +1078,82 @@ private void subtestOrderByMP_Subquery() throws Exception
         vt = client.callProcedure("@AdHoc", sql).getResults()[0];
         expected = new long[][] {{1}, {4}, {6}, {11}};
         validateTableOfLongs(vt, expected);
-}
+    }
+
+    /*
```
BillWhite: This makes debugging easier.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuiteMinMaxSum.java
```diff
@@ -793,7 +733,26 @@ private void testStringsAllNulls(Client client,
         validateStringTable(cr.getResults()[0], expected);
     }
 
-    public void testMin() throws Exception {
+    private static final boolean IS_ENABLED = true;
+    private static final boolean ISNOT_ENABLED = false;
+
```
BillWhite: This test was unnecessarily slow.  Changing it in this way cuts its execution time by about 3.

tests/frontend/org/voltdb/regressionsuites/TestWindowFunctionSuite.java
```diff
@@ -185,8 +185,70 @@ private void initUniqueTableExtra(Client client, boolean append)
         client.callProcedure("tu.insert", 70, 3);
         client.callProcedure("tu.insert", 80, 2);
     }
+    private static final boolean IS_ENABLED = true;
+    private static final boolean ISNOT_ENABLED = false;
 
-    public void testRank_UNIQUE() throws NoConnectionsException, IOException, ProcCallException {
+    public void testAll() throws Exception {
```
BillWhite: This file is mostly renamed from TestWindowedAggregates.java.  However, we combine all the tests into one test function.  This combination makes the test run in about 1/3 the time.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1045,6 +1096,523 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+
+            // The statement level order by expressions
+            // are always the last in the list.  So, if there
+            // are no window functions which match this
+            // index, and the order by expressions match,
+            // then the last time through this list will pick
+            // the order by expressions.  If a window function
+            // matches, we will pick it up first.  We
+            // can only match the statement level order by
+            // statements if on score from a window function
```
BillWhite: Should be "if no score is from a window function".

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1045,6 +1096,523 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+                        // If there are expressions later on in the
+                        // m_partitionByExprs or m_unmatchedOrderByExprs
+                        // which match this expression we need to
+                        // delete them.  We can safely ignore them.
+                        //
+                        // If there are more than one instances of
+                        // an expression, say with "PARTITION BY A, A",
+                        // we need to remove them all.  But guard against
+                        // an infinite loop here;
+                        for (idx = 0; m_partitionByExprs.remove(eorc) && idx < MAX_LIST_REMOVAL; idx += 1) {
```
wolffcm: Surely there is a better way to avoid an infinite loop?  Can't you make a call to `contains` to see if you're done?

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1045,6 +1096,523 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+            // are always the last in the list.  So, if there
+            // are no window functions which match this
+            // index, and the order by expressions match,
+            // then the last time through this list will pick
+            // the order by expressions.  If a window function
+            // matches, we will pick it up first.  We
+            // can only match the statement level order by
+            // statements if on score from a window function
+            // matches.
+            for (WindowFunctionScore score : m_winFunctions) {
```
wolffcm: So this loop has at most one iteration, is that correct?  Because there can only be one window function.

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -114,14 +115,48 @@ public static final void resetPlanNodeIds() {
...
     protected boolean m_isInline = false;
 
     /**
      * The textual explanation of why the plan may fail to have a deterministic result or effect when replayed.
      */
     protected String  m_nondeterminismDetail = "the query result does not guarantee a consistent ordering";
 
+    // Flag marking the sub-query plan
+    protected boolean m_isSubQuery = false;
+    protected StmtTableScan m_tableScan = null;
```
paulmartel: What are these suddenly doing here, newly arrived from AbstractScanPlanNode where they are actually used?
As far as I can tell, muddying the waters. No. We know better than to turn our most abstract base classes into pigpiles of members that are only used/useful by select subclass instances. This goes also for all the new state below supporting this changeset. Find some way to factor it out of here.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -1975,6 +1975,13 @@ public boolean hasOrderByColumns() {
         return null;
     }
 
+    /**
+     * Return the number of window function expressions.  This
+     * is non-zero for ParsedSelectStmt only.
+     */
+    public int getWindowFunctionExpressionCount() {
+        return 0;
+    }
```
BillWhite: Sorry.  Somehow I thought that m_windowFunctionExpressions was in ParsedSelectStmt.  Since it's in the base class there's no reason to do this.

src/frontend/org/voltdb/planner/AccessPath.java
```diff
@@ -35,20 +43,46 @@
...
+    final List<AbstractExpression> indexExprs = new ArrayList<>();
+    final List<AbstractExpression> endExprs = new ArrayList<>();
+    final List<AbstractExpression> otherExprs = new ArrayList<>();
+    final List<AbstractExpression> joinExprs = new ArrayList<>();
+    final List<AbstractExpression> bindings = new ArrayList<>();
+    final List<AbstractExpression> eliminatedPostExprs = new ArrayList<>();
+    //
+    // If a window function uses the index, then this will be set
+    // to the number of the window function which uses this index.
+    // If it is set to -1 then no window function uses the index, but
```
BillWhite: I started out using -1 and -2, and then changed to the named constant.  I thought I caught all of them, but apparently I missed one.  There are probably some more as well.  Thanks.

src/frontend/org/voltdb/planner/microoptimizations/InlineOrderByIntoMergeReceive.java
```diff
@@ -73,6 +80,15 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
                         return planNode; // Do not apply the optimization.
                     }
                 }
+            } else if (PlanNodeType.WINDOWFUNCTION == nodeType) {
```
BillWhite: I thought it was simpler to have two cases.  The algorithm for window functions is similar but much simpler.

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
BillWhite: m_isSubQuery and m_tableScan must have come in from a bad merge.  They are defined in AbstractScanPlanNode on master.

As for the rest of these, all the references to them are from AbstractPlanNode references.  If I moved them to AbstractScanPlanNode I would have to cast all the AbstractPlanNode references to AbstractScanPlanNodes, after doing a runtime check to determine that they are actually AbstractScanPlanNodes and not something else.  This seems more straightforward to me.

src/frontend/org/voltdb/plannodes/WindowFunctionPlanNode.java
```diff
@@ -276,4 +291,11 @@ public final void setAggregateTypes(List<ExpressionType> aggregateTypes) {
     public final List<List<AbstractExpression>> getAggregateExpressions() {
         return m_aggregateExpressions;
     }
+
+    public void setIsCompatibleWithStmtOrderBy() {
```
BillWhite: Nope.  This is an earlier version which leaked out.  Thanks.

tests/frontend/org/voltdb/planner/EEPlanTestGenerator.java
```diff
@@ -692,6 +696,61 @@ public void testGeneratedCountPlan() throws Exception {
...
 
+    /**
+     * Make a shuffled array.  Each row has two elements.  The
+     * first element indexed by idIndex is a the row index.
+     * The other element is a shuffled row index.
+     *
+     * @param nRows
+     * @return A shuffled array.
+     */
+    public static int[][] makeShuffledArray(int nRows, int idIndex) {
```
BillWhite: Done.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -1203,6 +1206,20 @@ protected static void truncateTable(Client client, String tb)
         validateTableOfScalarLongs(client, "select count(*) from " + tb, new long[]{0});
     }
 
+    protected static void truncateAllTables(Client client) throws Exception {
```
BillWhite: Perhaps.  But I think this will always be faster.  In any case, I got tired waiting for this test to run.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1058,86 +1626,157 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        findBindingsForOneIndexedExpression(ExpressionOrColumn nextStatementEOC,
+                                            ExpressionOrColumn indexEntry) {
+        assert(nextStatementEOC.m_expr != null);
+        AbstractExpression nextStatementExpr = nextStatementEOC.m_expr;
+        if (indexEntry.m_colRef != null) {
+            ColumnRef indexColRef = indexEntry.m_colRef;
+            // This is a column.  So try to match it
+            // with the expression in nextStatementEOC.
+            if (matchExpressionAndColumnRef(nextStatementExpr, indexColRef, indexEntry.m_tableScan)) {
+                return m_emptyBindings;
```
BillWhite: Sorry, didn't see that.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1058,86 +1626,157 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        //       such that S is a permutation of P and S+WO is a singular prefix
+        //       of IS.
+        //
+        // So, in both cases, order by and window function, we are looking for
+        // a sequence of expressions, S1, and a list of IS indices, OS, such
+        // that S1 is a prefix of IS with OS-singular values.
+        //
+        // If the statement level order by expression list is not a prefix of
+        // the index, we still care to know about window functions.  The reverse
+        // is not true.  If there are window functions but they all fail,
```
BillWhite: Fixed.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1058,86 +1626,157 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        //   1.) The expressions from the statement level order by clause, OO.
+        //       This sequence of expressions must be a singular prefix of IS.
+        //   2.) The expressions in a window function's partition by list, WP,
+        //       followed by the expressions in the window function's
+        //       order by list, WO.  The partition by functions are a set not
+        //       a sequence.  We need to find a sequence of expressions, S,
+        //       such that S is a permutation of P and S+WO is a singular prefix
+        //       of IS.
+        //
+        // So, in both cases, order by and window function, we are looking for
```
BillWhite: Fixed.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1058,86 +1626,157 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        //
+        // What expressions must we match?
+        //   1.) We have the parameters indexedExpressions and indexedColRefs.
+        //       These are the expressions or column references in an index.
+        //       Exactly one of them is non-null.  Since these are both an
+        //       indexed sequence of expression-like things, denote the
+        //       non-null one IS.
+        // What expressions do we care about?  We have two kinds.
+        //   1.) The expressions from the statement level order by clause, OO.
+        //       This sequence of expressions must be a singular prefix of IS.
```
BillWhite: Reworded.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1058,86 +1626,157 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        if (moreBindings != null) {
+            return moreBindings;
         }
-        return nSpoilers;
+        return null;
     }
 
+    private boolean matchExpressionAndColumnRef(AbstractExpression statementExpr,
+                                                ColumnRef          indexColRef,
+                                                StmtTableScan      tableScan) {
```
BillWhite: Fixed, though I believe you are completely incorrect.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1058,86 +1626,157 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        if (statementExpr instanceof TupleValueExpression) {
+            TupleValueExpression tve = (TupleValueExpression)statementExpr;
+            if (tve.getTableAlias().equals(tableScan.getTableAlias()) &&
+                    tve.getColumnName().equals(indexColRef.getColumn().getTypeName())) {
+                // NB: m_emptyBindings is a null list.  That means
+                //     we matched, but with no bindings.
```
BillWhite: It described a previous version of the code.  Sorry.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -769,7 +769,7 @@ private StmtTargetTableScan simplifierForSubquery(AbstractParsedStmt subquery) {
         int tableCount = 0;
         StmtTargetTableScan simpler = null;
         for (Map.Entry<String, StmtTableScan> entry : selectSubquery.m_tableAliasMap.entrySet()) {
-            if (entry.getKey().contains("VOLT_TEMP_TABLE")) {
+            if (entry.getKey().contains(AbstractParsedStmt.TEMP_TABLE_NAME)) {
```
BillWhite: Fixed.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -942,18 +955,56 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+     * If a candidate runs out of expressions, and the sort orders
+     * are sensible, it is marked done and is usable.  We currently
+     * support only one window function for a select statement.  But the
+     * same set of matching logic applies to window functions and any statement
+     * level order by clause, so generalizing to multiple window functions
+     * is essentially cost free.
      *
      * TODO: In theory, we COULD leverage index ordering when the index covers only a prefix of
-     * the ORDER BY list, such as the "ORDER BY A, B, C" case listed above.
+     * the ordering requirement list, such as the "ORDER BY A, B, C, D" case listed above.
```
BillWhite: I don't understand this.  That's not what I meant to say.  I did remove the extra conjunction "or".

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1045,6 +1096,523 @@ private static boolean isAGeoExpressionIndex(List<AbstractExpression> indexedExp
...
+        // This is the expression or column index of this expression or
+        // ColumnRef in the candidate index.
+        int                m_indexEntryNumber;
+        // This is the sort direction of a statement
+        // expression.
+        public SortDirectionType m_sortDirection;
+
+        ExpressionOrColumn(int indexEntryNumber, AbstractExpression expr, SortDirectionType sortDir) {
+            this(indexEntryNumber, null, expr, sortDir, null);
+        }
```
BillWhite: I don't understand this.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
BillWhite: I did this initially.  The loop then look like:
  while (m_partitionByExprs.remove(eorc)) {
    /* do nothing */
  }

This looks completely dubious to me, since there is no loop body.  I wanted to avoid looping forever.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1610,6 +2249,10 @@ private static AbstractPlanNode getIndexAccessPlanForTable(StmtTableScan tableSc
         scanNode.setBindings(path.bindings);
         scanNode.setEndExpression(ExpressionUtil.combinePredicates(path.endExprs));
         scanNode.setPredicate(path.otherExprs);
+        scanNode.setWindowFunctionUsesIndex(path.m_windowFunctionUsesIndex);
+        scanNode.setSortOrderFromIndexScan(path.sortDirection);
+        scanNode.setWindowFunctionIsCompatibleWithOrderBy(path.m_stmtOrderByIsCompatible);
+        scanNode.setFinalExpressionOrderFromIndexScan(path.m_finalExpressionOrder);
```
BillWhite: An index could provide ordering compatible with both the statement level order by and the window function.  Then windowFunctionUsesIndex == 0, and I still need to denote that the statement level order by is compatible with the index ordering.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
```diff
@@ -1610,6 +2249,10 @@ private static AbstractPlanNode getIndexAccessPlanForTable(StmtTableScan tableSc
         scanNode.setBindings(path.bindings);
         scanNode.setEndExpression(ExpressionUtil.combinePredicates(path.endExprs));
         scanNode.setPredicate(path.otherExprs);
+        scanNode.setWindowFunctionUsesIndex(path.m_windowFunctionUsesIndex);
```
BillWhite: Done.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+ *               Expressions which reference both branches of a join
+ *               might be outer expressions of a higher level
+ *               join.</li>
+ *           </ul>
+ *       </li>
+ *     </ol>
+ *   </li>
+ *   <li>
+ *     The planner calculates a set of access paths for the tablescans.
+ *     Note that this does not depend on the join order.  Each access
```
paulmartel: Actually, it DOES depend on the join order. For example, for "SELECT * FROM A JOIN B ON ID;" with tables A and B both indexed on ID, then, for join order (A, B), the access paths are (table A: (seqscan A), table B: (seqscan B, indexscan B on ID)) but for join order (B, A), the access paths are  (table A: (seqscan A, indexscan A on ID), table B: (seqscan B)). That is, multi-table predicates can generally only be index-optimized on the last-referenced table(scan) IN JOIN ORDER. This is actually the primary motivation for trying different join orders.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+ *   So,  the aggregate node’s calculation can be done row-wise.  This
+ *   means the aggregate node can be inlined in some cases.</li>
+ *   <li>
+ *    Replace an aggregate plan node with an index count plan node or
+ *    a table count plan node.</li>
+ *   <li>
+ *     Replace an aggregate plan node with an index scan plan node with
+ *    an inline limit node.</li>
+ * </ol>
+ * </p>
```
paulmartel: Nice overview. I'm not sure where/how this doc gets published , but we should make sure that there is some kind of link/reference to it from the github wiki (from the 10,000 ft overview or some such place) and to that github wiki page from the in-house developer wiki page.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -1850,15 +1850,14 @@ private static boolean isOrderByNodeRequired(AbstractParsedStmt parsedStmt, Abst
...
             }
         }
-        if (probe != null) {
-            nonAggPlan = probe;
-        } else {
+        if (probe == null) {
             // No idea what happened here.  We can't find a
             // scan or join node at all.  This seems unlikely
             // to be right.  Maybe this should be an assert?
             nonAggPlan = root;
```
paulmartel: Remove pointless assignment to local variable prior to return.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
```diff
@@ -554,18 +555,20 @@ static protected void validateRowOfLongs(VoltTable vt, long [] expected) {
...
+    static protected <T> T[][] shuffleArray(T [][] input) {
+        T[][] output = input.clone();
         Integer [] indices = new Integer[input.length];
         for (int idx = 0; idx < indices.length; idx += 1) {
             indices[idx] = Integer.valueOf(idx);
         }
         List<Integer> permutation = Arrays.asList(indices);
         Collections.shuffle(permutation);
-        long[] tmp = input[permutation.get(0)];
+        T[] tmp = input[permutation.get(0)];
```
paulmartel: This was in your original code, but puzzling nonetheless. What is gained by shifting the elements of the permuted list by 1 position before assigning them into the array? Why not a more straightforward element by element assignment at the same position? Is this known to somehow compensate for some known skew to the Collections.shuffle output? It seems arbitrary.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
paulmartel: Fine, but you should generalize the function comment to match -- and it's odd that I don't see any code using this generalization, so for now it's looking like an imagineered change for its own sake.

src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -1799,35 +1802,134 @@ private static boolean isOrderByNodeRequired(AbstractParsedStmt parsedStmt, Abst
...
+        AbstractPlanNode probe;
+        for (probe = root;
+                ! ((probe instanceof AbstractJoinPlanNode)
+                    || (probe instanceof AbstractScanPlanNode))
+                && (probe != null);
+            probe = (probe.getChildCount() > 0) ? probe.getChild(0) : null) {
+            // Count the number of window functions between the
+            // root and the join/scan node.  Note that we know we
+            // have a statement level order by (SLOB) here.  If the SLOB
+            // can use the index for ordering the scan or join node
```
BillWhite: Actually, I left a word out.  But I reworked this.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -48,14 +42,16 @@
...
+ * An inner join computation will be given two tables which must be scanned
  * in two nested loops.  For each row of the left hand table, which we call the <em>outer</em>
  * table, we look at each row of the right hand table, which we call the <em>inner</em>
  * table.  If the pair of rows satisfy the join condition, we write the row to the
  * output table.</br>
  * Note:  The computation for NESTED_LOOP is somewhat more complicated
  * than this, as we may be able to skip the scan of the inner table, and there
- * may be other optimizations.  We are just giving the flavor of the computation here.</br>
+ * may be other optimizations.  Also, NESTED_LOOP can computer outer
+ * joins.  We are just giving the flavor of the computation here.</br>
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -83,28 +79,19 @@
...
+ * aggregate functions can be computed row-wise.  For example, the sum
  * aggregate function needs only a running sum and values from the current row
- * to update its state.  Other operations, such as a join, may take access to
- * an entire table.  For row-by-row operations, we avoid creating extra temporary
+ * to update its state.  Not all operations can be computed row-wise.
+ * For example, a sort requires an entire table.
+ * For row-wise operations, we avoid creating extra temporary
  * tables by combining scans. The row-by-row operations are placed in <em>inline</em>
  * nodes in a scan node.  Each row of the scan which passes the filters are
  * applied to the inline nodes during the scan.  The inline nodes have
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
  * hierarchy is the AbstractPlanNode class.
  * </p>
+ *
+ * <h1>How Does The Planner Work In Detail?</h1>
+ * <p>
+ * The planner takes a data structure describing a DML or DQL SQL
+ * statement.  It produces a plan of execution which can be used to
+ * evaluate the statement.  The Execution Engine executes the plan.
+ * Note that DDL is processed entirely differently, and is not part of
+ * this discussion.</p>
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+
+ * <h2>The Input</h2>
+ * <p>
+ * The input to the planner is ultimately a SQL statement, which is a
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+ *           <li>Joined tables,</li>
+ *           <li>Derived tables, and</li>
+ *           <li>expressions used in WHERE and ON conditions,</li>
+ *         </ol>
+ *       </li>
+ *       <li>The list of GROUP BY expressions,</li>
+ *       <li>The HAVING expression, if any,</li>
+ *       <li>Any LIMIT or OFFSET constants.</li>
+ *       <li>An indication of whether the statement is a SELECT, DELETE
+ *           or INSERT statement.  DDL is not part of this process.</li>
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+ *      <ul>
+ *        <li>Nodes to compute aggregates and GROUP BY.  The aggregation algorithms include
+ *          <ul>
+ *            <li>Serial Aggregation, which is aggregation based on sorting,</li>
+ *            <li>Hash Aggregation, which is aggregation based on hash tables,</li>
+ *            <li>Partial Aggregation, which is a form of parallelized
+ *                serial aggregation.</li>
+ *          </ul>
+ *       </li>
+ *       <li>Nodes to compute UNION,</li>
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+ *       <li>Nodes to compute ORDERBY,</li>
+ *       <li>Nodes to compute PROJECTION,</li>
+ *       <li>Nodes to compute MATERIALIZE,</li>
+ *       <li>Nodes to compute LIMIT and OFFSET,</li>
+ *       <li>Nodes to compute WINDOWFUNCTION.</li>
+ *       <li>Nodes to compute window functions.</li>
```
BillWhite: OK.

src/frontend/org/voltdb/planner/package-info.java
```diff
@@ -114,5 +101,310 @@
...
+ * </ul>
+ * </p>
+ *
+ * <p>
+ * Plan nodes have different kinds of children.  A plan node can have
+ * nodes which are out-of-line or inline.  The latter are used to
+ * compute which can be computed row-wise, in a single pass through
+ * the input table along with the parent node.  For example, a LIMIT
+ * or OFFSET node is often made inline, since it just counts how many
+ * trailing or initial rows to ignore.  PROJECTION nodes can also be
```
BillWhite: Ok.

src/frontend/org/voltdb/planner/package-info.java
BillWhite: I don't understand this.  We only create access paths (1) at the root of the join tree, when we create an empty path, (2) in getRelevantAccessPathForIndex, and (3) at the join nodes when we create naive paths.  I think (1) and (3) are irrelevant, because they are pretty much constants.  The path created in (2) is only created for a leaf node.  I guess I don't see how the leaf node knows the join order.  Do you get different filter expressions for different join orders?  But the filter expressions on a given leaf node are on the tablescan of the leaf node, and those will be the same for any join order.

I do see that we generate access paths from the join trees, so they are generated along with each join order.  I just don't see how they can be different for different join orders.  We might generate them and reject different ones in different join order.  But that's different.

src/frontend/org/voltdb/planner/microoptimizations/InlineOrderByIntoMergeReceive.java
```diff
@@ -83,6 +99,63 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
...
     /**
+     * Convert ReceivePlanNodes into MergeReceivePlanNodes.  We won't
+     * have any inline limits or aggregates here, so this is somewhat
+     * simpler than the order by case.
+     *
+     * @param plan
+     * @return
+     */
+    private AbstractPlanNode applyOptimization(WindowFunctionPlanNode plan) {
+        AbstractPlanNode child = (plan.getChildCount() > 0) ? plan.getChild(0) : null;
```
BillWhite: I have put the asserts in.

src/frontend/org/voltdb/planner/microoptimizations/InlineOrderByIntoMergeReceive.java
```diff
@@ -83,6 +99,63 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
...
+        }
+        // SP Plans which have an index which can provide
+        // the window function ordering don't create
+        // an order by node.
+        if ( ! ( child instanceof OrderByPlanNode ) ) {
+            return plan;
+        }
+        OrderByPlanNode onode = (OrderByPlanNode)child;
+        child = onode.getChild(0);
+        // The order by node needs to be followed by a
```
BillWhite: I rewrote this to use child-of rather than follows.

src/frontend/org/voltdb/planner/microoptimizations/InlineOrderByIntoMergeReceive.java
```diff
@@ -83,6 +99,63 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
...
+        // Receive node.
+        if ( ! ( child instanceof ReceivePlanNode) ) {
+            return plan;
+        }
+        ReceivePlanNode receiveNode = (ReceivePlanNode)child;
+        // No inline nodes are expected in the window function plan node.
+        assert(plan.getChildCount() == 1);
```
BillWhite: Moved up to the beginning of the function.

src/frontend/org/voltdb/planner/microoptimizations/InlineOrderByIntoMergeReceive.java
```diff
@@ -83,6 +99,63 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
...
+        // No inline nodes are expected in the Receive node.
+        assert(receiveNode.getChildCount() == 1);
+        child = receiveNode.getChild(0);
+        // The Receive node must be followed by a Send node.
+        // Maybe this should be an assert.
```
BillWhite: Done.

src/frontend/org/voltdb/planner/microoptimizations/InlineOrderByIntoMergeReceive.java
```diff
@@ -83,6 +99,63 @@ protected AbstractPlanNode recursivelyApply(AbstractPlanNode planNode)
...
+        if (! ( child instanceof SendPlanNode) ) {
+            return plan;
+        }
+        SendPlanNode sendNode = (SendPlanNode)child;
+        child = sendNode.getChild(0);
+        // If this window function does not use the
+        // index then this optimization is not possible.
+        if (child.getWindowFunctionUsesIndex() != 0) {
```
BillWhite: I added a comment explaining why == 0 is the right choice.

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -114,14 +115,44 @@ public static final void resetPlanNodeIds() {
...
+    //
+    // This will be propagated into a scan plan from the access
+    // path and up the outer branch of a join.
+    private List<AbstractExpression> m_finalExpressionOrderFromIndexScan;
+    // Set the order direction for an index scan.  There
+    // is only one of these.
+    //
+    // This will be propagated into a scan plan from the access
+    // path and up the outer branch of a join.
+    private SortDirectionType m_sortDirectionFromIndexScan = SortDirectionType.INVALID;
```
BillWhite: I thought you didn't like interfaces or accessor functions.  If it's ok to use an interface, or an overridden accessor function I would prefer that.  I'm just trying to dodge your unwritten code review rules.

tests/frontend/org/voltdb/planner/EEPlanTestGenerator.java
```diff
@@ -692,6 +696,67 @@ public void testGeneratedCountPlan() throws Exception {
...
+     *                r, a[r][0] == r and the vector <a[s][1] | 0 <= s < 100>
+     *                is a permutation of the set {0, ..., 99}.  Similarly, if
+     *                b == makeShuffledArray(100, 1) then for each r,
+     *                b[r][1] == r and <b[s][0] | 0 <= s < 100> is a
+     *                permutation of the set {0, ..., 99}.
+     * @return A shuffled array.
+     */
+    public static int[][] makeShuffledArray(int nRows, int idIndex) {
+        assert(idIndex == 0 || idIndex == 1);
+        List<Integer> list = new ArrayList<>();
```
BillWhite: All this is not adding anything, since this test is not enabled anyway.  I deleted this function and the test.  I'm going to be working on this stuff soon anyway, so I'll revive it. sometime in the future.

tests/frontend/org/voltdb/planner/EEPlanTestGenerator.java
```diff
@@ -692,6 +696,67 @@ public void testGeneratedCountPlan() throws Exception {
...
+        for (int idx = 0; idx < nRows; idx += 1) {
+            list.add(idx);
+        }
+
+        Collections.shuffle(list);
+        int answer[][] = new int[nRows][];
```
BillWhite: Deleted.

tests/frontend/org/voltdb/planner/TestSelfJoins.java
```diff
@@ -339,7 +339,8 @@ public void testIndexedSelfJoin() {
         //* for debug */ System.out.println(apn.toExplainPlanString());
         // Project-first case: GROUP BY column that is not in the order by or the display column list
         pn = apn.getChild(0);
-        assertTrue(pn instanceof OrderByPlanNode);
+        assertTrue(String.format("Expected PlanNodeType.ORDERBY not %s", pn.getPlanNodeType()),
+                   pn instanceof OrderByPlanNode);
```
BillWhite: I only made the one change to print the node type on failure.  I broke this test and wanted to know what I needed to fix.  I've fixed it to be equal to master.

I don't see how assertLeftChain helps here anyway.  There's a lot of testing of the intermediate nodes.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
```diff
@@ -544,11 +545,539 @@ public void testExplainPlanText() {
...
+                         PlanNodeType.SEND,
+                         PlanNodeType.PROJECTION,
+                         PlanNodeType.WINDOWFUNCTION,
+                         PlanNodeType.INDEXSCAN);
+        }
+        //   8: No SLOB, one WF, MP Query, index (Can order the WF)
+        //      Expect WinFun -> MrgRecv(WF) -> SEND -> IndxScan
+        //   select a, b, max(b) over ( partition by a ) from vanilla_pb_idx where a = 1;
+        if (IS_ENABLED) {
+            validatePlan("select a, b, max(b) over ( partition by a ) from vanilla_pb_idx where a = 1;",
```
BillWhite: Ok, I added tests for inequality.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
```diff
@@ -544,11 +545,539 @@ public void testExplainPlanText() {
...
+                         PlanNodeType.INVALID, // Fragment Marker.
+                         PlanNodeType.SEND,
+                         PlanNodeType.INDEXSCAN);
+        }
+
+        // This is one of the queries from the regression test.
+        // It is here because it tests that the window function
+        // and order by function have the same expressions but
+        // different sort directions.
+        if (ISNOT_ENABLED) {
```
BillWhite: Sorry.  Forgot to change it back.

tests/frontend/org/voltdb/planner/TestWindowedFunctions.java
```diff
@@ -544,11 +545,539 @@ public void testExplainPlanText() {
...
+                         1,
+                         PlanNodeType.SEND,
+                         PlanNodeType.PROJECTION,
+                         PlanNodeType.WINDOWFUNCTION,
+                         PlanNodeType.INDEXSCAN);
+        }
+        if (IS_ENABLED) {
+            // Check to see that order information is
+            // propagated through outer branches of
+            // joins.  Join order should not matter, so
```
BillWhite: Fixed.

tests/frontend/org/voltdb/regressionsuites/RegressionSuite.java
BillWhite: I don't remember why I did that.  There was some reason, but I don't recall it.

BillWhite: Now I remember. It was originally written to shuffle an array in-place.  So I had to apply the permutation to the original array.  Later on I decided it was more useful to return the shuffled array, but the permutation wasn't taken out.  In any case I've just copied the input (shuffled) to the output.

BillWhite: This function is used for T = Long in org.voltdb.regressionsuite.TestWindowFunctionSuite and T = Integer in org.voltdb.regressionsuite.TestWindowFunctionsWithIndexes.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
paulmartel: Functions like this one in Java can have side effects. This should not be news. This one has the side effect of removing from the finite list every time it returns true. Conversely, it returns false if it does not make progress, just as if this might have been designed to anticipate this case as an idiomatic usage. I find your lack of faith disturbing.

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
paulmartel: I guess the unwritten rule here is to minimize the scope and visibility of changes to work WITH the java typing instead of around it or against it. Except now it's written. You heard it here first? I don't recall ever objecting to accessor functions that are actually called and not just defined as a knee-jerk every-attribute-must-need-them basis -- assuming that they are defined at appropriate scope and visibility. The one edge case I'm currently on the fence about is use of trivial accessors in scopes where the members they access are clearly visible anyway. But I'm generally pro-accessor. 

As for interfaces, I have nothing against using an interface to reduce linkage between jars or to model classes that have a shared behavior that is not reflected in their inheritance (i.e. not shared by their co-heirs). I object to interfaces when they are redundant with or can easily be replaced with abstract base classes that are doing things interfaces cannot, like defining useful default implementations. I object to interfaces methods that are only actually required of SOME implementers and the others fake it with dummy returns or exceptions -- there are idiomatic ways to avoid this. Otherwise I have nothing against interfaces,that I can recall. You must be thinking of some other arbitrary task-master.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
BillWhite: Ok, I've changed it, but I'm still dubious.  I guess I've spent too much time dealing with libraries that are not reliably implemented.

src/frontend/org/voltdb/planner/SubPlanAssembler.java
paulmartel: OK. I'll try again to show you how this code unraveled for me.
Start with findBindingsForOneIndexedExpression. How does it use `ExpressionOrColumn nextStatementEOC` ? It asserts nextStatementEOC.m_expr != null and from there goes on to treat m_expr as the only useful attribute of ExpressionOrColumn. It raises the small question "why not just pass in the AbstractExpression and thus reduce the complexity of the function?". So I looked at the callers, three of them.

I discovered that one caller applies the function to m_partitionByExprs elements. A quick search shows that there are very few references to these elements. They are constructed with an interesting AbstractExpression and layers of uninteresting hard-coded dummy values, put into this list and then removed based on an equality match that looks like it is expected to return true iff the AbstractExpressions are equal. In short, ExpressionOrColumn really (at best) adds nothing to this process. For simplicity, m_partitionByExprs should just be List<AbstractExpression>. 

The other caller in the same method is only slightly different and slightly more problematic. It passes elements of m_unmatchedOrderByExprs. These elements have essentially the same usage as m_partitionByExprs elements except that they also participate in sort direction determination. So, they have one other useful (non-dummied) attribute aside from the AbstractExpression. That's one, not the several defined on ExpressionColumn. Further, reasonable semantics for the equality-based removal from the List would be to ignore the sort direction. So, here, it looks like the simpler List<AbstractExpression> would give us most of what we need here. The short-fall could be made up by a parallel list of sort directions. This just slightly complicates the removal process because now it needs to expose the position of each element being matched/removed so that we can remove the corresponding sort direction list element -- a minor adjustment. As I tried to say in my earlier comment, you COULD define another list element type that encapsulated both the AbstractExpression and sort direction, and implemented equals to just do AbstractExpression equals, but I think that's slightly overkill, though not nearly at the level of overkill of using ExpressionOrColumn with its three additional unused dummied columns.

This brings us to the third caller of findBindingsForOneIndexedExpression, the ExpressionOrColumn implementation of equals that we were just discussing. Recall that the two List.remove use cases for this equals function we've been discussing A) were expected to use dummy values for all but m_expr and m_sortDirection and B) were in particular only interested in m_expr equality, either m_really wanted to be determined solely by AbstractExpression equals.
So all the considerable complexity of this equals implementation is to cover some other use case, or it's just clutter and can be eliminated once the two known use cases are fixed to use AbstractExpression equals. So I searched the module for other use cases -- non-trivial since equals is most often invoked indirectly as a "callback" from some Collection method. But the remaining (non-trivial) use cases of ExpressionOrColumn for modeling the indexed columns/expressions  -- as opposed to the statement's expressions -- always expressions -- don't appear to be used by any such methods. This was borne out by some quick debugging with the known use cases diverted and a never-reached breakpoint. So, in short, ExpressionOrColumn.equals is a fraud as it only ever boils down to m_expr.equals. Also its first constructor, the one with all the dummying is not actually useful, and since it is only used for indexed columns/expressions which don't (yet?) support ASC/DESC, it doesn't even need its m_sortDirection. I've got to give you credit for writing fake code convincing enough that I had to refactor it to be convinced that it was safe to do so. So I guess I may as well push the result of that to a branch, too "for your review". See branch refactored-10474-Index4Windows .

PR #4271, opened 2017-02-08
ENG-11789, fix fault resolution algorithm, don't waiting for remote decisions we may have already received.

luyangco:
Change-Id: I88ad00952eb6a1d021c034e4bbf03381eb2098ed

### DIFF
--- a/src/frontend/org/voltcore/agreement/MeshArbiter.java
+++ b/src/frontend/org/voltcore/agreement/MeshArbiter.java
@@ -29,7 +29,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-import com.google_voltpatches.common.collect.Lists;
 import org.voltcore.logging.Level;
 import org.voltcore.logging.VoltLogger;
 import org.voltcore.messaging.FaultMessage;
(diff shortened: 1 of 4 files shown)


luyangco:
Ticket description can be found here https://issues.voltdb.com/browse/ENG-11789.

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -406,8 +406,21 @@ protected boolean notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
             }
         }
 
-        long start = System.currentTimeMillis();
         boolean allDecisionsMatch = true;
+        // ENG-11789, It's possible to have all necessary decisions already at this point,
+        // in that case don't wait if all decisions are matched.
+        if (m_decidedSurvivors.keySet().containsAll(expectedSurvivors)) {
+            for (SiteFailureMessage remoteDecision : m_decidedSurvivors.values()) {
+                if (!sfm.m_survivors.equals(remoteDecision.m_survivors)) {
```
vtkstef: are the sets equal if if the both contain the same elements but are ordered differently?

vtkstef: the end of the do loop uses containsAll

PR #4272, opened 2017-02-08
Eng-11689: snmp trap for DRRole stopped

JadenW:
1. extract channelDistributer from Importer
2. register channel callbacks for ResourceUsageMonitor to distribute DRRole query
3. periodically check DRRole state in ResourceUsageMonitor and alert if State changed to "Stopped"

### DIFF
--- a/src/frontend/org/voltdb/HealthMonitor.java
+++ b/src/frontend/org/voltdb/HealthMonitor.java
@@ -145,7 +145,7 @@ public class HealthMonitor implements Runnable, Promotable
                     m_logger.error("Interrupted while pausing cluster for resource overusage", e);
                     return;
                 }
-                ClientResponseImpl r = ClientResponseImpl.class.cast(cb.getResponse());
+                ClientResponse r = cb.getResponse();
                 if (r.getStatus() != ClientResponse.SUCCESS) {
                     m_logger.error("Unable to pause cluster for resource overusage: " + r.getStatusString());
                 }
@@ -179,7 +179,7 @@ public class HealthMonitor implements Runnable, Promotable
                 m_logger.error("Interrupted while retrieving cluster for DRROLE STATS", e);
                 return;
             }
-            ClientResponseImpl r = ClientResponseImpl.class.cast(cb.getResponse());
+            ClientResponse r = cb.getResponse();
             if (r.getStatus() != ClientResponse.SUCCESS) {
                 m_logger.error("Unable to retrieve DRROLE STATS: " + r.getStatusString());
                 return;
--- a/src/frontend/org/voltdb/HealthMonitor.java
+++ b/src/frontend/org/voltdb/HealthMonitor.java
@@ -17,6 +17,7 @@
 
 package org.voltdb;
 
+import org.apache.zookeeper_voltpatches.KeeperException;
 import org.voltcore.logging.VoltLogger;
 import org.voltcore.utils.CoreUtils;
 import org.voltdb.AuthSystem.AuthUser;
@@ -26,9 +27,6 @@ import org.voltdb.client.SyncCallback;
 import org.voltdb.compiler.deploymentfile.DrRoleType;
 import org.voltdb.compiler.deploymentfile.ResourceMonitorType;
 import org.voltdb.compiler.deploymentfile.SystemSettingsType;
-import org.voltdb.importer.ChannelChangeCallback;
-import org.voltdb.importer.ImporterChannelAssignment;
-import org.voltdb.importer.VersionedOperationMode;
 import org.voltdb.snmp.FaultFacility;
 import org.voltdb.snmp.SnmpTrapSender;
 import org.voltdb.snmp.ThresholdType;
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -156,6 +170,51 @@ private AuthUser getInternalUser()
         return VoltDB.instance().getCatalogContext().authSystem.getInternalAdminUser();
     }
 
+    private void checkDRRole() {
+        SyncCallback cb = new SyncCallback();
+        if (getConnectionHadler().callProcedure(getInternalUser(), false, BatchTimeoutOverrideType.NO_TIMEOUT, cb, "@Statistics", "DRROLE",0)) {
+            try {
+                cb.waitForResponse();
+            } catch (InterruptedException e) {
+                m_logger.error("Interrupted while retrieving cluster for DRROLE STATS", e);
```
vtkstef: `return;`

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -156,6 +170,51 @@ private AuthUser getInternalUser()
...
+            }
+            ClientResponseImpl r = ClientResponseImpl.class.cast(cb.getResponse());
+            if (r.getStatus() != ClientResponse.SUCCESS) {
+                m_logger.error("Unable to retrieve DRROLE STATS: " + r.getStatusString());
+            } else {
```
vtkstef: you could `return` before the `} else {` and unindent the else section

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -156,6 +170,51 @@ private AuthUser getInternalUser()
...
+                        m_logger.debug("DRROLE stats: Role:" + drRole + " State:" + state + " Remote Cluster ID:" + remoteCluster);
+                    }
+                    if (drRole == DrRoleType.NONE) {
+                        continue;
+                    }
+
+                    if (DRRoleStats.State.STOPPED == state) {
+                        if (!m_snmpDRTrapSent.getOrDefault(remoteCluster, false)) {
+                            m_snmpTrapSender.statistics(FaultFacility.DR, String.format("Database Replication ROLE: %s break with Remote Cluster %d.",
+                                    drRole, remoteCluster));
```
vtkstef: when we send disk traps we keep sending it until it is cleared

src/frontend/org/voltdb/ResourceUsageMonitor.java
```diff
@@ -242,4 +301,25 @@ public static String getValueWithUnit(long value)
...
+    @Override
+    public void onChange(ImporterChannelAssignment assignment) {
+        if (m_logger.isDebugEnabled()) {
+            m_logger.debug("ImporterChannelAssignment: " + assignment);
+        }
+        if (assignment.getAdded().size() > 0) {
+            m_isDRRoleChecker = true;
+        } else {
+            if (assignment.getRemoved().size() > 0) {
+                m_isDRRoleChecker = false;
```
vtkstef: instead of a boolean save the uri that has the distributer cluster state, remove yourself only if the removed matches what you saved

src/frontend/org/voltdb/HealthMonitor.java
```diff
@@ -156,6 +170,51 @@ private AuthUser getInternalUser()
...
+    private void checkDRRole() {
+        SyncCallback cb = new SyncCallback();
+        if (getConnectionHadler().callProcedure(getInternalUser(), false, BatchTimeoutOverrideType.NO_TIMEOUT, cb, "@Statistics", "DRROLE",0)) {
+            try {
+                cb.waitForResponse();
+            } catch (InterruptedException e) {
+                m_logger.error("Interrupted while retrieving cluster for DRROLE STATS", e);
+                return;
+            }
+            ClientResponseImpl r = ClientResponseImpl.class.cast(cb.getResponse());
```
nshi: What's the purpose of the cast?

src/frontend/org/voltdb/HealthMonitor.java
```diff
@@ -156,6 +170,51 @@ private AuthUser getInternalUser()
...
+                }
+                if (drRole == DrRoleType.NONE) {
+                    continue;
+                }
+
+                if (DRRoleStats.State.STOPPED == state) {
+                    if (!m_snmpDRTrapSent.getOrDefault(remoteCluster, false)) {
+                        m_snmpTrapSender.statistics(FaultFacility.DR, String.format("Database Replication ROLE: %s break with Remote Cluster %d.",
+                                drRole, remoteCluster));
+                        m_snmpDRTrapSent.put(remoteCluster, true);
```
nshi: Is this flag reset somewhere? What happens when the connection is fixed later and it breaks again? Will this prevent new alerts for the same remote cluster being issued?

JadenW: It should be reset to false when the state changed to any none stop state. (in the following else case)
It should always resending new alerts unless the stable state is too short (i.e. if the connection is fixed but broken again before healthMonitor checks it).

src/frontend/org/voltdb/HealthMonitor.java
JadenW: seems not need. will drop the cast.

PR #4274, opened 2017-02-09
Eng 11743 stats no reset

xinjiacs:
Keep UAC stats after catalog update.

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -212,6 +212,8 @@ public class ProcedureRunner {
                 m_site.getCorrespondingSiteId(),
                 m_site.getCorrespondingPartitionId(),
                 m_catProc);
+        // Normally m_statsCollector is returned as it is and there is no affect to assign it to itself.
+        // Sometimes when this procedure statistics needs to reuse the existing one, the old stats gets returned.
         m_statsCollector = VoltDB.instance().getStatsAgent().registerProcedureStatsSource(
                 site.getCorrespondingSiteId(),
                 m_statsCollector);
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -208,15 +208,15 @@ public class ProcedureRunner {
 
         m_procedure.init(this);
 
-        m_statsCollector = new ProcedureStatsCollector(
-                m_site.getCorrespondingSiteId(),
-                m_site.getCorrespondingPartitionId(),
-                m_catProc);
         // Normally m_statsCollector is returned as it is and there is no affect to assign it to itself.
         // Sometimes when this procedure statistics needs to reuse the existing one, the old stats gets returned.
         m_statsCollector = VoltDB.instance().getStatsAgent().registerProcedureStatsSource(
                 site.getCorrespondingSiteId(),
-                m_statsCollector);
+                new ProcedureStatsCollector(
+                        m_site.getCorrespondingSiteId(),
+                        m_site.getCorrespondingPartitionId(),
+                        m_catProc)
+                );
 
         reflect();
     }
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -133,14 +133,14 @@
...
+        m_procName = catProc.getClassname();
     }
 
     /**
      * Called when a procedure begins executing. Caches the time the procedure starts.
      */
-    public final void beginProcedure() {
-        if (m_invocations % timeCollectionInterval == 0) {
+    public final void beginProcedure(boolean isSystemProc) {
+        if (m_invocations % timeCollectionInterval == 0 || (isSystemProc && isProcedureUAC())) {
```
luyangco: `isProcedureUAC()` has a string comparison so still some performance impact on transactional path if system procedures run frequently.

src/frontend/org/voltdb/StatsAgent.java
```diff
@@ -572,26 +610,39 @@ public VoltTable getStatsAggregate(
...
     {
         assert selector != null;
-        final NonBlockingHashMap<Long, NonBlockingHashSet<StatsSource>> siteIdToStatsSources = registeredStatsSources.get(selector);
+        NonBlockingHashMap<Long, NonBlockingHashSet<StatsSource>> siteIdToStatsSources =
+                m_registeredStatsSources.get(selector);
 
         // There are cases early in rejoin where we can get polled before the server is ready to provide
         // stats.  Just return null for now, which will result in no tables from this node.
-        if (siteIdToStatsSources == null || siteIdToStatsSources.isEmpty()) {
+        if (selector == StatsSelector.PROCEDURE) {
```
luyangco: Please comment on why `StatsSelector.PROCEDURE` needs separate handling

luyangco: Also the comments above should move down the line 640.

src/frontend/org/voltdb/ProcedureStatsCollector.java
xinjiacs: I put "isSystemProc" first as boolean operation hoping the early return helps. 
So for non-system procedure call, it should not affect the performance. For system procedure call, string prefix check may affect. Let's check performance benchmark jenkins first.

If bad on AdHoc (frequently called), we can skip prefix string "org.voltdb.sysprocs.". It should be sufficient enough to check a couple characters.

src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -163,7 +163,7 @@ public final void endProcedure(
             {
                 if (Math.abs(delta) > 1000000000)
                 {
-                    log.info("Procedure: " + m_catProc.getTypeName() +
+                    log.info("Procedure: " + m_procName +
```
xinjiacs: getTypeName() is null, and this function is a generic function from catalog item. 
getClassname() has real meaning, specifically for PROCEDURE.

src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -377,4 +377,43 @@ public long getInvocations() {
...
+
+    @Override
+    public int hashCode() {
+        return super.hashCode() + m_partitionId + m_procName.hashCode();
+    }
+
+    /**
+     * @return true if this procedure statistics should be reset
+     */
+    public boolean resetAfterCatalogChange() {
```
xinjiacs: I did it on purpose with TODO next if we can allow more procedures to be kept.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -273,7 +274,7 @@ public ClientResponseImpl call(Object... paramListIn) {
         assert(m_batch.size() == 0);
 
         try {
-            m_statsCollector.beginProcedure();
+            m_statsCollector.beginProcedure(isSystemProcedure());
```
yzhang1991: I guess this `isSystemProcedure` can be stored as a field of statsCollector?

PR #4278, opened 2017-02-14
Eng 11983 swap txnid2 args

johnTcrawford:
I intended this to just modify txnid2's run.sh to allow the async-benchmark (test) client to accept arguments (e.g. "--swapratio=0"), and TruncateTableLoader.java to lower some of the routine output from "info" to "debug"; but it looks like it also includes a much earlier commit of mine, as well as some of Paul's changes.


wolffcm:
I cherry-picked your latest commit on this branch and put it on mine.

johnTcrawford:
Yes, assuming that your branch already has my much earlier commit (from Jan. 18), then cherry-picking just my latest is the right thing to do.

wolffcm:
Yes, your earlier commit is on there.

PR #4280, opened 2017-02-16
Eng 11642 code cleanup

xinjiacs:
This PR does not have functional changes, but only code cleanup. It tries to simplify the LoadedProcedureSet.java and make it less dependent or correlated with other members. 
It has a corresponding PR on Pro.

### DIFF
--- a/src/frontend/org/voltdb/LoadedProcedureSet.java
+++ b/src/frontend/org/voltdb/LoadedProcedureSet.java
@@ -73,7 +73,7 @@ public class LoadedProcedureSet {
     public LoadedProcedureSet(SiteProcedureConnection site) {
         m_site = site;
 
-        m_runnerFactory = new ProcedureRunnerFactory(site, site.getSystemProcedureExecutionContext());
+        m_runnerFactory = new ProcedureRunnerFactory(m_site, m_site.getSystemProcedureExecutionContext());
         m_csp = null;
         m_defaultProcCache = new HashMap<>();
         m_defaultProcManager = null;
@@ -86,7 +86,7 @@ public class LoadedProcedureSet {
         }
     }
 
-    public void registerPlanFragment(final long pfId, final ProcedureRunner proc) {
+    private void registerPlanFragment(final long pfId, final ProcedureRunner proc) {
         synchronized (m_registeredSysProcPlanFragments) {
             assert(m_registeredSysProcPlanFragments.containsKey(pfId) == false);
             m_registeredSysProcPlanFragments.put(pfId, proc);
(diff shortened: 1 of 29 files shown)


src/frontend/org/voltdb/LoadedProcedureSet.java
```diff
@@ -143,8 +157,9 @@ public void loadProcedures(
                     procedure = lang.accept(procedureInstantiator, procClass);
                 }
                 catch (final Exception e) {
+                    // TODO: remove the extra meaningless parameter "0"
                     hostLog.l7dlog( Level.WARN, LogKeys.host_ExecutionSite_GenericException.name(),
-                                    new Object[] { m_siteId, m_siteIndex }, e);
+                                    new Object[] { m_site.getCorrespondingSiteId(), 0}, e);
```
luyangco: Since m_siteIndex is always zero, how about removing this parameter in `voltdb_logstrings.properties`

src/frontend/org/voltdb/LoadedProcedureSet.java
```diff
@@ -78,31 +95,29 @@ public void registerPlanFragment(final long pfId, final ProcedureRunner proc) {
...
     {
-        // default proc caches clear on catalog update
-        m_defaultProcCache.clear();
+        m_registeredSysProcPlanFragments.clear();
 
-        m_defaultProcManager = catalogContext.m_defaultProcs;
         m_csp = csp;
+        m_defaultProcManager = catalogContext.m_defaultProcs;
+        // default proc caches clear on catalog update
+        m_defaultProcCache.clear();
```
luyangco: I am not sure whether changing the order of these statements introduces any side effects.

src/frontend/org/voltdb/LoadedProcedureSet.java
```diff
@@ -197,7 +200,8 @@ private void loadSystemProcedures(
                     hostLog.l7dlog(
                             Level.WARN,
                             LogKeys.host_ExecutionSite_GenericException.name(),
-                            new Object[] { m_siteId, m_siteIndex },
+                            // TODO: remove the extra meaningless parameter "0"
+                            new Object[] { m_site.getCorrespondingSiteId(), 0 },
```
luyangco: My comment above also applies here.

src/frontend/org/voltdb/LoadedProcedureSet.java
```diff
@@ -207,26 +211,38 @@ private void loadSystemProcedures(
                 }
                 catch (final InstantiationException e) {
                     hostLog.l7dlog( Level.WARN, LogKeys.host_ExecutionSite_GenericException.name(),
-                            new Object[] { m_siteId, m_siteIndex }, e);
+                            new Object[] { m_site.getCorrespondingSiteId(), 0 }, e);
```
luyangco: Here too.

src/frontend/org/voltdb/LoadedProcedureSet.java
luyangco: Or remove all `siteIndex` in the l7log if it doesn't have any meaning.

xinjiacs: Looking at `voltdb_logstrings.properties`, we have so many places using "SiteId" and "SiteIndex". 

What we really need is to pass "HostId" and "SiteId" into these buckets, making the information readable. Currently it's just logging as a bit shifted long value with 0 index.

It's worthy to have a separate commit addressing all issues in `voltdb_logstrings.properties`.

PR #4281, opened 2017-02-17
ENG-11019  SPI migration and  transaction reroute [REVIEW ONLY, DO NOT MERGE] 

zeemanhe:
This pull request is part of spi migration from master to a replica.  The work includes a new sys proc BalanceSPI (from @xinjiacs ) and sp transaction reroute from master to new master without pause when command logging is off.  Transaction reroute for mp transactions and reroute with command logging enabled will be in separate pull requests.


src/frontend/org/voltcore/messaging/NodeFailureNotificationMessage.java
```diff
@@ -0,0 +1,67 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.voltcore.messaging;
```
jhugg: I still don't understand why we need this message.

src/frontend/org/voltcore/zk/ZKUtil.java
```diff
@@ -100,27 +128,6 @@ public static File getUploadAsTempFile( ZooKeeper zk, String path, String prefix
         return tempFile;
     }
 
-//    /*
```
jhugg: What's with the commented out bit?

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -1194,15 +1196,17 @@ private void failOverConnection(Integer partitionId, Long initiatorHSId, Connect
...
         for (Iv2InFlight inFlight : transactions) {
             ClientResponseImpl response =
                     new ClientResponseImpl(
                             ClientResponseImpl.RESPONSE_UNKNOWN,
                             ClientResponse.UNINITIALIZED_APP_STATUS_CODE,
                             null,
-                            new VoltTable[0],
-                            "Transaction dropped due to change in mastership. " +
-                            "It is possible the transaction was committed");
+                            new VoltTable[0], DROP_TXN_MASTERSHIP);
```
jhugg: Nit: Probably should still put one param on each line to match style.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -506,6 +509,35 @@ else if ("@Rebalance".equals(procName)) {
...
+                    "@BalanceSPI: requires integer partition id, target host id.",
+                    task.clientHandle);
+        }
+
+        int pid = (Integer) params[0];
+        int newHostId = (Integer) params[1];
+        Long hsid = VoltDB.instance().getCartograhper().getHSIdForMaster(pid);
+        if (newHostId == CoreUtils.getHostIdFromHSId(hsid)) {
+            return  gracefulFailureResponse(
+                    "@BalanceSPI: SPI is already the host.", task.clientHandle);
```
jhugg: Why do any of this param checking work here instead of at the actual site? Is it possible it could be out of date by the time it arrives at the site?

src/frontend/org/voltdb/VoltZK.java
```diff
@@ -308,4 +363,22 @@ public static boolean removeCatalogUpdateBlocker(ZooKeeper zk, String node, Volt
         }
         return true;
     }
+
+    public static boolean updateLeaderCacheNode(ZooKeeper zk, String rootDir, int partitionId, String hsidStr, VoltLogger log) {
```
jhugg: Can we add header comments to methods like this. What's it used for and how should it be used?

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -690,6 +690,7 @@ public boolean backpressureBarrier(final long start, long timeoutNanos) throws I
         String m_hostName;
         int m_clientPort;
         int m_adminPort;
+        boolean m_initialized = false;
```
jhugg: @The bug here is that the server advertises a node you can't connect to. The fix is to delay the advertisement, not make the client smarter than it has to be. Make another ticket for this bug and we'll fix there.

src/frontend/org/voltcore/zk/ZKUtil.java
zeemanhe: The whole method was removed since it was commented on  master.

src/frontend/org/voltcore/messaging/NodeFailureNotificationMessage.java
zeemanhe: The message was removed. It is not worth adding a new message type and using once

src/frontend/org/voltcore/zk/ZKUtil.java
```diff
@@ -84,6 +84,34 @@ public static String path(String... components)
         return path;
     }
 
+    private static final String BALANCE_SPI_SUFFIX = "_BALANCE_SPI_REQUEST";
```
zeemanhe: moved.

src/frontend/org/voltdb/ClientInterface.java
zeemanhe: updated.

src/frontend/org/voltdb/VoltZK.java
zeemanhe: this method was removed after code reorganization.

src/frontend/org/voltdb/client/ClientImpl.java
zeemanhe: the change was rolled back. a separate ticket has been submitted.

src/frontend/org/voltcore/zk/ZKUtil.java
jhugg: Oh I see. I feel dumb now.

src/frontend/org/voltdb/InvocationDispatcher.java
zeemanhe: will move these to the site

src/frontend/org/voltdb/utils/TopologyZKUtils.java
```diff
@@ -69,4 +74,51 @@ public static void updateTopologyToZK(ZooKeeper zk, AbstractTopology topology) {
             VoltDB.crashLocalVoltDB("Unable to update topology to ZK, dying", true, e);
         }
     }
+
+    /**
+     * Given a partition ID, return a list of HSIDs of all the sites with copies of that partition
+     */
+    public static List<Long> getReplicasForPartition(ZooKeeper zk, int partition) {
```
jhugg: I find it weird that there's no other code to do this somewhere. The Cartographer doesn't have this info?

src/frontend/org/voltdb/utils/TopologyZKUtils.java
```diff
@@ -69,4 +74,51 @@ public static void updateTopologyToZK(ZooKeeper zk, AbstractTopology topology) {
...
+                    true, e);
+        }
+
+        return retval;
+    }
+
+    /**
+     * Given a partition ID, return a list of HSIDs of all the sites on a host with copies of that partition
+     */
+    public static List<Long> getReplicasForPartition(ZooKeeper zk, int hostid, int partition) {
```
jhugg: The name should be more specific. "getReplicasForPartitionOnGivenHost" maybe?

How often would we call this? Would it make more sense to get the replicas for a partition, then filter at the calling code using a lambda?

src/frontend/org/voltdb/messaging/Iv2InitiateTaskMessage.java
```diff
@@ -52,6 +52,8 @@
     // not serialized.
     AtomicBoolean m_isDurable;
 
+    boolean m_isForReplica = false;
```
jhugg: What does this flag do? There needs to be a comment. Especially because the term "replica" is overloaded in VoltDB (partitions, dr, etc...).

src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
```diff
@@ -52,6 +53,13 @@
     private StoredProcedureInvocation m_invocation;
     private Pair<Long, byte[]> m_currentHashinatorConfig;
 
+    // Mis-routed invocation due to SPI change, send back to ClientInterface for restart
+    private boolean m_misrouted;
```
jhugg: Let's encode this info in the ClientResponse (you already seem to do this) rather than adding a byte to every response. Make the rare case slightly more expensive without adding bulk to the fast path.

src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
```diff
@@ -52,6 +53,13 @@
...
+
+    // a flag used for SPI balance operation, indicating that the task was created
+    //when the site was leader partition
+    boolean m_createdFromLeader = false;
```
jhugg: This seems to add a byte to the fast path just to support an assertion. I like the thought, but I'm not sure adding a byte is worth it.

src/frontend/org/voltdb/iv2/TransactionTask.java
```diff
@@ -33,12 +33,23 @@
     final protected TransactionTaskQueue m_queue;
     protected ListenableFuture<Object> m_durabilityBackpressureFuture = CoreUtils.COMPLETED_FUTURE;
 
+    // a flag used for SPI balance operation, indicating that the task was created
+    //when the site was leader partition
+    protected boolean m_createdFromLeader = false;
```
jhugg: Is this ever used? Should I be looked at pro code? I can't find use in community.

src/frontend/org/voltdb/utils/TopologyZKUtils.java
zeemanhe: The method is not used any more.  will be removed

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -984,6 +984,8 @@ private static void loadSystemProcedures()
                 ImmutableMap.<Integer, List<String>>builder().put( 0, new ArrayList<String>()).build());
         Procedures.put("@ResetDR",
                 ImmutableMap.<Integer, List<String>>builder().put( 0, new ArrayList<String>()).build());
+        Procedures.put("@BalanceSPI",
```
zeemanhe: agree. will roll it back

src/frontend/org/voltdb/utils/TopologyZKUtils.java
zeemanhe: Cartographer is the sole caller of this method. It makes more sense to  move this to Cartographer.

src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
zeemanhe: great idea.  the byte was removed.

PR #4282, opened 2017-02-17
ENG-11662

akhanzode:


src/frontend/org/voltdb/ClientInterfaceHandleManager.java
```diff
@@ -63,9 +62,9 @@
 
     private volatile boolean m_wantsTopologyUpdates = false;
 
-    private HandleGenerator m_shortCircuitHG = new HandleGenerator(SHORT_CIRCUIT_PART_ID);
+    private final HandleGenerator m_shortCircuitHG = new HandleGenerator(SHORT_CIRCUIT_PART_ID);
 
-    private final Map<Long, Iv2InFlight> m_shortCircuitReads = new HashMap<Long, Iv2InFlight>();
+    private final Map<Long, Iv2InFlight> m_shortCircuitReads = Collections.synchronizedMap(new HashMap<>());
```
vtkstef: this will slow down all transactions

src/frontend/org/voltdb/ClientInterfaceHandleManager.java
```diff
@@ -77,7 +76,7 @@
             m_partitionId = partitionId;
         }
 
-        public long getNextHandle()
+        public synchronized long getNextHandle()
```
vtkstef: same here

akhanzode:
How would this be slower than the existing global synchronized on getHandle() regardless of partition. This will only synchronize same partition getHandle other partition calls should continue without locking.

PR #4283, opened 2017-02-17
ENG-12023: keep system procedures but only reloading user procedures …

xinjiacs:
…when UAC


luyangco:
Looks good.

src/frontend/org/voltdb/LoadedProcedureSet.java
```diff
@@ -73,62 +72,94 @@ public VoltProcedure visitGroovy(Class<?> p) throws Exception {
...
+            // reload all system procedures from beginning
+            m_sysProcs = loadSystemProcedures(catalogContext, m_site, csp);
+        }
+
+    }
+
+    /**
+     * When catalog updates, only user procedures needs to be reloaded. System procedures can be left without changes.
+     */
+    public void loadProceduresForCatalogUpdate(
```
zeemanhe: This method is empty. should  be removed.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -221,6 +222,20 @@ public String visitGroovy(VoltProcedure p) {
         reflect();
     }
 
+    public void reInitSysProc(CatalogContext catalogContext, CatalogSpecificPlanner csp) {
+        assert(m_procedure != null);
+        if (! m_isSysProc) {
+            return;
+        }
+        assert(m_procedure instanceof VoltSystemProcedure);
```
zeemanhe: This assert is not needed since it will hit line 228 if the procedure is not a VoltSystemProcedure.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -1104,46 +1119,46 @@ else if (type == VoltType.NUMERIC) {
...
-            }
-        };
+       @Override
+       public Map<String, SQLStmt> visitJava(ProcedureRunner p) {
+           Map<String, SQLStmt> stmtMap = null;
+           try {
+               stmtMap = ProcedureCompiler.getValidSQLStmts(null, p.m_procedureName, p.m_procedure.getClass(), p.m_procedure, true);
+           } catch (Exception e1) {
+               // shouldn't throw anything outside of the compiler
+               e1.printStackTrace();
```
zeemanhe: Should this be logged instead of printing stack trace?

xinjiacs: Code moved up here unchanged, not sure what does this comment mean. Leave it untouched.

PR #4293, opened 2017-02-20
ENG-11889 proceduredetail data structure

yzhang1991:


src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -37,87 +40,101 @@
...
+        /**
+         * Total result size for calculating averages
+         */
+        private long m_totalResultSize = 0;
+        private long m_lastTotalResultSize = 0;
+
+        /**
+         * Smallest parameter set size
+         */
+        private int m_minParameterSetSize = Integer.MAX_VALUE;
```
yzhang1991: Not that I changed the type for parameter set types from long to int.
It is strange to assume that the maximum value for a result size can fit in integer but the maximum value for a parameter set has to be long.

yzhang1991:
This pull request is for review only.

PR #4294, opened 2017-02-20
ENG-11929: Remove DR compatibility mode (protocol version 3)

slowsoul:
* Removed code in EE that is related to CompatibleDRTupleStream and CompatibleBinaryLogSink
* In EE handling of SET_DR_PROTOCOL_VERSION task, save the set version and set DR streams. Since V7.1 will only be able to produce version 7 binary log, for now, this is enough. We may need to expose version to the streams or make the version a member of stream class if we will support online binary log version change
* Added a new COMPATIBLE_PROTOCOL_VERSION of 7 to both Java and EE to indicate the new compatible version, in EE it is used in BinaryLogSinkWrapper to check for unsupported binary log version. Renamed the old version 3 to FIRST_COMPATIBLE_PROTOCOL_VERSION as it is still used in ProtocolBuf and DRMessageTranslator
* Removed compatible flag from parameters of JNI method getTestDRBuffer()
* Fixed a typo of MUTLI instead of MULTI in MULTICLUSTER_PROTOCOL_VERSION


src/frontend/org/voltdb/dr2/DRProtocol.java
```diff
@@ -20,7 +20,12 @@
 public interface DRProtocol {
     // Also update DRTupleStream.h if version changes
     public static final int PROTOCOL_VERSION = Integer.getInteger("DR_PROTOCOL_VERSION", 7);
```
wweiss-voltdb: Even though it is useful to set configure the PROTOCOL_VERSION, we should bound it to the valid set of versions supported by the software.

slowsoul: Sorry for the late response. Actually, I now think there's no point in overriding the PROTOCOL_VERSION as an environment variable. Maybe when this was introduced, DR protocol version only differed in binary log binary format, but now they also differ in control message format and semantics. Should we remove the overriding completely and just give it a number?

PR #4295, opened 2017-02-21
Catch UnknownHostException for run in docker

man4j:
When first container starts and I specify "HOSTS=voltdb1,voltdb2,voltdb3" voltdb process exit with UnknownHostException because other containers not started and they unreacheable by dns resolution. After this fix all works fine.


man4j:
In Voltdb 6.x all works fine by default. I think this is regression in 7.x.

PR #4298, opened 2017-02-21
ENG-11602

akhanzode:
and ENG-11991
Pro: https://github.com/VoltDB/pro/pull/1771


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1241,11 +1251,13 @@ public Boolean get() {
                 try {
                     Class<?> ndrgwClass = null;
                     ndrgwClass = Class.forName("org.voltdb.dr2.DRProducer");
-                    Constructor<?> ndrgwConstructor = ndrgwClass.getConstructor(File.class, File.class, boolean.class, int.class, int.class);
+                    Constructor<?> ndrgwConstructor = ndrgwClass.getConstructor(File.class, File.class, boolean.class, boolean.class, boolean.class, int.class, int.class);
                     m_producerDRGateway =
                             (ProducerDRGateway) ndrgwConstructor.newInstance(
                                     new VoltFile(VoltDB.instance().getDROverflowPath()),
                                     new VoltFile(VoltDB.instance().getSnapshotPath()),
+                                    (durable || m_config.m_startAction.doesRecover() || determination.terminusNonce != null),
```
nshi: Won't this still trick the DRProducer into thinking that it can resume even if command logging is turned off and nothing is restored?

akhanzode: Yes I should remove doesRecover() which means either CL is on or terminus is present = recoverable.

nshi: I think you want to only pass `true` if it's recovering from some on-disk artifact, either command log or shutdown snapshot, like the following.

`m_config.m_startAction.doesRecover() && (durable || determination.terminusNonce != null)`

akhanzode: Yes I corrected my comment with "and"

PR #4306, opened 2017-02-22
ENG-11783: SwapTables system procedure

wolffcm:
This implements a SwapTables system procedure.  This is work that I inherited from Paul several weeks ago.  His original pull request is here:

https://github.com/VoltDB/voltdb/pull/4199

The work I added fixed an issue with double-adding indexes to the list of indexes to be swapped, and I also fixed an issue with SwapTables during an ongoing snapshot.  I also added unit tests.

JIRA tickets:
https://issues.voltdb.com/browse/ENG-11783
https://issues.voltdb.com/browse/ENG-11981
https://issues.voltdb.com/browse/ENG-11982
https://issues.voltdb.com/browse/ENG-11810


tests/frontend/org/voltdb/regressionsuites/TestSystemProcedureSuite.java
```diff
@@ -26,45 +26,348 @@
...
 
-public class TestSystemProcedureSuite extends RegressionSuite {
+import junit.framework.Test;
 
+public class TestSystemProcedureSuite extends RegressionSuite {
     private static int SITES = 3;
     private static int HOSTS = MiscUtils.isPro() ? 2 : 1;
     private static int KFACTOR = MiscUtils.isPro() ? 1 : 0;
-    private static boolean hasLocalServer = false;
+    private static boolean hasLocalServer = true; //// false;
```
paulmartel: It's PROBABLY OK? that we changed this setting -- might make debugging easier by default?

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -762,9 +762,10 @@ private static void executeStatement(String statement)
             m_startTime = System.nanoTime();
             SQLParser.ExecuteCallResults execCallResults = SQLParser.parseExecuteCall(statement, Procedures);
             if (execCallResults != null) {
+                String procName = execCallResults.procedure;
```
BillWhite: Is this really an optimization?  Won't the compiler do this anyway?

Not that I'm complaining.

tests/frontend/org/voltdb/TestAdHocQueries.java
```diff
@@ -897,7 +897,8 @@ public void testTimestampInsert() throws Exception {
                 fail("Compilation should have failed.");
             }
             catch(ProcCallException e) {
-                assertTrue(e.getMessage().contains("Error compiling"));
+                String expectedMsg = "Error compiling";
```
BillWhite: Is expectedMsg used here?

src/frontend/org/voltdb/utils/SQLCommand.java
yzhang1991: I guess it's just making the code reads nicer.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1809,10 +1825,6 @@ public  boolean createTransaction(
                 isShortCircuitRead = true;
             }
         }
-        if (initiatorHSId == null) {
```
wolffcm: I'm not sure.  This was part of Paul's code.  I'll ask him.

paulmartel: Yes this is safe. Eclipse caught this as dead code, and marked this line as a warning.
I was skeptical until I looked into it.
And then I was impressed by the apparent depth of its analysis.
Basically, the initial null value of the variable is overwritten by a non-null value in every possible code path.
Only the re-assignment in the FIRST block uses an Integer return value for a possible null.
But the result of that re-assignment is checked for null, and if null, it is overwritten with another assignment. This latter assignment and all the others in the other code paths are from non-null int-typed values. So, there's no possibility of a null value where we were checking.
Maybe it would be a little clearer to use two variables, change the current one to int and declare a separate Integer scoped to the first code block that is used with the == null conditional to set the int variable.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -237,13 +237,11 @@ public void createConnectionWithHashedCredentials(
...
-    public final ClientResponse callProcedure(
-            String procName,
-            Object... parameters)
-                    throws IOException, NoConnectionsException, ProcCallException
+    public final ClientResponse callProcedure(String procName, Object... parameters)
+            throws IOException, NoConnectionsException, ProcCallException
     {
-        return callProcedureWithClientTimeout(BatchTimeoutOverrideType.NO_TIMEOUT, procName,
-                Distributer.USE_DEFAULT_CLIENT_TIMEOUT, TimeUnit.SECONDS, parameters);
+        return callProcedureWithClientTimeout(BatchTimeoutOverrideType.NO_TIMEOUT, false,
```
paulmartel: Rule of thumb: if you put a system library call inline in the argument list of a call to your method, you make it harder to step into your method -- you have to step into the system call and step out of it, and step in again GRR.

src/frontend/org/voltdb/InvocationDispatcher.java
wolffcm: Wow, that is impressive.

tests/frontend/org/voltdb/regressionsuites/TestSystemProcedureSuite.java
```diff
@@ -493,75 +813,256 @@ public void testPause() throws Exception {
...
+                        (ii == 15 && jj == 16) || (ii == 16 && jj == 15) ||
+                        (ii == 17 && jj == 18) || (ii == 18 && jj == 17)) {
+
+                    VoltTable[] results;
+
+                    populateSwappyTables(client, theTable, otherTable);
+
+                    results = client.callProcedure("@AdHoc",
+                            "SELECT COUNT(*) FROM " + theTable).getResults();
+                    long preCount = results[0].asScalarLong();
```
johnTcrawford: Seems like it would be natural here to also get the otherCount, i.e., the COUNT(*) for otherTable, and then compare it to postCount below

tests/frontend/org/voltdb/regressionsuites/TestSystemProcedureSuite.java
```diff
@@ -493,75 +813,256 @@ public void testPause() throws Exception {
         }
     }
 
+    private void populateSwappyTables(Client client, String thisTable, String thatTable) throws Exception {
+        // Populate the tables with traceable rows.
+        // After each swap, confirm that both tables are populated, that
+        // the total number of rows is unchanged, and that the first table's
+        // column sum HAS changed.
```
johnTcrawford: This comment describes the table rows as "traceable", yet you never (as far as I can see) actually check the values in the tables, only the number of rows. Maybe you should check the values too.

tests/frontend/org/voltdb/regressionsuites/TestSystemProcedureSuite.java
wolffcm: I think it would be better to change it back.  The original author must have chosen a non-local server for a reason?  It's not clear from the history.

tests/frontend/org/voltdb/TestAdHocQueries.java
wolffcm: Fixed, thanks.

src/frontend/org/voltdb/planner/ParsedSwapStmt.java
```diff
@@ -0,0 +1,79 @@
...
+
+package org.voltdb.planner;
+
+import org.hsqldb_voltpatches.VoltXMLElement;
+import org.voltdb.catalog.Database;
+import org.voltdb.catalog.Table;
+
+/**
+ *
+ *
```
wolffcm: Fixed.

src/frontend/org/voltdb/planner/ParsedSwapStmt.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    @Override
+    void parse(VoltXMLElement stmtNode) {
+        // parseTablesAndParameters may have been called on this
+        // SWAP TABLE statement, but the simplified VoltXML representation
+        // for SWAP TABLE gave that method nothing to do.
+        assert(stmtNode.children.isEmpty());
+        assert(m_tableList.isEmpty());
+
+        String theName = stmtNode.attributes.get("thetable");
```
wolffcm: Yes, good idea.  Fixed.

tests/frontend/org/voltdb/regressionsuites/TestSystemProcedureSuite.java
wolffcm: Thanks John.  I have modified the test to use the method `assertContentOfTable` which validates the counts as well as the contents, both before and after the swaps.

PR #4307, opened 2017-02-22
ENG-12041: override the equals method for DRConsumerNodeStats and DRConsumerPart…

JadenW:
…itionStats

We registerStatsSource each time we create a new DRConsumerDispatcher.

Each new instances of DRConsumerNodeStats and DRConsumerPartitionStats need to be treated as unequal.

### DIFF
--- a/src/frontend/org/voltdb/DRConsumerStatsBase.java
+++ b/src/frontend/org/voltdb/DRConsumerStatsBase.java
@@ -64,6 +64,12 @@ public class DRConsumerStatsBase {
         protected Iterator<Object> getStatsRowKeyIterator(boolean interval) {
             return ImmutableSet.of().iterator();
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            return false;
+        }
     }
 
     public static class DRConsumerPartitionStatsBase extends StatsSource {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/DRConsumerStatsBase.java
```diff
@@ -64,6 +64,12 @@ protected void populateColumnSchema(ArrayList<VoltTable.ColumnInfo> columns) {
         protected Iterator<Object> getStatsRowKeyIterator(boolean interval) {
             return ImmutableSet.of().iterator();
         }
+
+        @Override
+        public boolean equals(Object o) {
```
xinjiacs: Two issues:
1. Overriding the equals and hashCode methods together is usually a good practice. 
[See stack overflow](http://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java)
2. Is DRConsumerStats unique by "cluster id"? A cluster can have many DR consumers, but each consumer will have a different unique id.  And there can only be one consumer stats per DR consumer for a procedure cluster. Is this right? If so, it would make sense overriding the "equals" and "hashcode" functions using "cluster id"?

JadenW: 1. I will override that too. 
2. Yes, unique consumer stats per producer cluster. However, I previously haven't store this remote cluster id in the stats, we just populate that value each time we updateStatsRow. Since this value is final in ConsumerDispatcher, I will add it to the ConsumerStats. This helps generate unique hashcode too.

JadenW:
Pro side change:
https://github.com/VoltDB/pro/pull/1775

PR #4311, opened 2017-02-25
Eng 11977 Implement IPv4 and IPv6 internet address stringification functions.

BillWhite:
These are the functions requested from Sakura.

https://issues.voltdb.com/browse/ENG-11977


tests/ee/expressions/function_test.cpp
```diff
@@ -551,6 +557,15 @@ TEST_F(FunctionTest, NaturalModTest) {
     ASSERT_EQ(sawException, true);
 }
 
+TEST_F(FunctionTest, inet6NtoATest) {
```
BillWhite: This test is really minimal.  I was mostly using it to debug a problem with inet6_ntoa.  But I thought better to leave it than delete it.  If we decide we need more tests of in6_addr it will be useful.

PR #4312, opened 2017-02-27
Eng 11806 upgrade command

luyangco:


src/frontend/org/voltdb/sysprocs/SystemInformation.java
```diff
@@ -454,6 +454,9 @@ static public VoltTable populateOverviewTable()
             vt.addRow(hostId, "LICENSE", VoltDB.instance().getLicenseInformation());
         }
         populatePartitionGroups(hostId, vt);
+
+        // root path
+        vt.addRow(hostId, "VOLTDBROOT", VoltDB.instance().getVoltDBRootPath());
```
vtkstef: please don't use the singleton this information is available in system procedure context

lib/python/voltcli/cli.py
```diff
@@ -306,6 +306,7 @@ def print_help(self):
         optparse.OptionParser.print_help(self)
         if not self.format_epilog_called:
             sys.stdout.write(self.on_format_epilog())
+        sys.stdout.write('\n')
```
xinjiacs: Haha, I want this change !!!

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -0,0 +1,336 @@
...
+
+def generateCommands(opts, hosts, kfactor):
+    halfNodes = len(hosts.hosts_by_id) / 2
+    (killSet, surviveSet) = pickNodeToKill(hosts, kfactor, halfNodes)
+    
+    # find a survivor node and a killSet node
+    survivor = surviveSet[0]
+    victim = killSet[0]
+    
+    #0 generate deployment file
```
xinjiacs: I found that it's worthy reviewing these messages with Andrew.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -0,0 +1,336 @@
...
+                                [VOLT.FastSerializer.VOLTTYPE_STRING],
+                                ['OVERVIEW'])
+    
+     # Convert @SystemInformation results to objects.
+    hosts = Hosts(runner.abort)
+    for tuple in response.table(0).tuples():
+        hosts.update(tuple[0], tuple[1], tuple[2])
+    
+    numberOfNodes = len(hosts.hosts_by_id)
+    if numberOfNodes % 2 == 1 and runner.opts.newNode is None:
```
zeemanhe: This should be the live node count.  Should the command give errors if the cluster has missing nodes?

lib/python/voltcli/voltadmin.d/stop.py
```diff
@@ -19,12 +19,13 @@
...
 from voltcli.hostinfo import Host
 from voltcli.hostinfo import Hosts
+from voltcli import utility
 
 @VOLT.Command(
     bundles = VOLT.AdminBundle(),
     description = 'Stop one host of a running VoltDB cluster.',
     arguments = (
-        VOLT.StringArgument('target_host', 'the target HOST name or address'),
+        VOLT.StringArgument('target_host', 'the target HOST[:PORT] name or address. (default PORT=3021)'),
```
zeemanhe: Would it be better to use a separate option for port# with default = 3021?

src/frontend/org/voltdb/sysprocs/SystemInformation.java
luyangco: I tried to avoid using this pattern, but then gave up because this is a static function be called in many places, and contains 10+ use of `VoltDB.instance()` .

lib/python/voltcli/voltadmin.d/stop.py
luyangco: I could, but using host:port together seems a convention, like the voltadmin -H option.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
luyangco: I emailed Andrew, he'll take a look.

src/frontend/org/voltdb/sysprocs/SystemInformation.java
akhanzode: Why add VOLTDBROOT here? Why do you need it it can be different on different nodes and should not be consistently useful for upgrade.

PR #4316, opened 2017-02-27
ENG-12045: print a warning message if the security setting is not on

xinjiacs:


akhanzode:
You may want to add warning in UAC also if security is turned off which I believe can be done.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2275,6 +2276,12 @@ else if (sysType.getQuery().getTimeout() == 0) {
                 }
             }
 
+            // log a warning on console log if security setting is turned off, like durability warning.
+            SecurityType securityType = deployment.getSecurity();
+            if (securityType == null || !securityType.isEnabled()) {
+                consoleLog.warn("Security is turned off.");
```
jhugg: I would suggest:
"Authentication-based security is disabled. This cluster may be read, updated or administers by anyone with network access."

Ask Andrew?

xinjiacs:
@akhanzode Security logging during UAC will be handled in a separate ticket: ENG-11880

PR #4321, opened 2017-02-28
Eng 12010 dont count streamed table in memory stats

brians127:
This change has a test added in Pro - since exporters are needed to achieve a StreamedTable TUPLE_COUNT > 0.


src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -1255,7 +1255,10 @@ private void statsTick(long time)
                 while (stats.advanceRow()) {
                     //Assert column index matches name for ENG-4092
                     assert(stats.getColumnName(7).equals("TUPLE_COUNT"));
-                    tupleCount += stats.getLong(7);
+                    assert(stats.getColumnName(6).equals("TABLE_TYPE"));
+                    if (stats.getString(6).equals("PersistentTable")){
```
vtkstef:
```java
if ("PersistentTable".equals(stats.getString(6)) {
```
Question why is this form better than

```java
if (stats.getString(6).equals("PersistentTable")){
```

PR #4325, opened 2017-03-02
Eng 10294 swap tables dr

manjujames:
Please don't merge! This is for review only.

Swap table DR events are generated from EE within swap table transaction. DR binary log data for swap table txns will be this event within begin and txn.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -1220,6 +1220,41 @@ void VoltDBEngine::rebuildTableCollections() {
     resetDRConflictStreamedTables();
 }
 
+void VoltDBEngine::swapDRActions(PersistentTable* table1, PersistentTable* table2) {
+    TableCatalogDelegate* tcd1 = getTableDelegate(table1->name());
+    TableCatalogDelegate* tcd2 = getTableDelegate(table2->name());
+    assert(!tcd1->materialized());
+    assert(!tcd2->materialized());
+    // Point the Map from signature hash point to the correct persistent tables
+    int64_t hash1 = *reinterpret_cast<const int64_t*>(tcd1->signatureHash());
+    m_tablesBySignatureHash[hash1] = table2;
+    int64_t hash2 = *reinterpret_cast<const int64_t*>(tcd2->signatureHash());
+    m_tablesBySignatureHash[hash2] = table1;
+
+    // Generate swap table DREvent
+    //TODO: Do this only if the tables are drEnabled
+    int64_t lastCommittedSpHandle = m_executorContext->lastCommittedSpHandle();
+    int64_t spHandle = m_executorContext->currentSpHandle();
+    int64_t uniqueId = m_executorContext->currentUniqueId();
+    // Following are stored: name1.length, name1, name2.length, name2, hash1, hash2
+    int32_t bufferLength = 4 + table1->name().length() + 4 + table2->name().length() + 8 + 8;
+    char* payloadBuffer[bufferLength];
+    ExportSerializeOutput io(payloadBuffer, bufferLength);
+    io.writeBinaryString(table1->name().c_str(), table1->name().length());
+    io.writeBinaryString(table2->name().c_str(), table2->name().length());
+    io.writeLong(hash1);
+    io.writeLong(hash2);
+    ByteArray payload(io.data(), io.size());
+
+    quiesce(lastCommittedSpHandle);
+    m_executorContext->drStream()->generateDREvent(SWAP_TABLE, lastCommittedSpHandle,
+            spHandle, uniqueId, payload);
+    if (m_executorContext->drReplicatedStream()) {
+        m_executorContext->drReplicatedStream()->generateDREvent(SWAP_TABLE, lastCommittedSpHandle,
+                spHandle, uniqueId, payload);
+    }
+}
+
 void VoltDBEngine::resetDRConflictStreamedTables() {
     if (getIsActiveActiveDREnabled()) {
         // These tables that go by well-known names SHOULD exist in an active-active
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -1227,12 +1227,19 @@ void VoltDBEngine::swapDRActions(PersistentTable* table1, PersistentTable* table
     assert(!tcd2->materialized());
     // Point the Map from signature hash point to the correct persistent tables
     int64_t hash1 = *reinterpret_cast<const int64_t*>(tcd1->signatureHash());
-    m_tablesBySignatureHash[hash1] = table2;
     int64_t hash2 = *reinterpret_cast<const int64_t*>(tcd2->signatureHash());
-    m_tablesBySignatureHash[hash2] = table1;
+    // Most swap action is already done.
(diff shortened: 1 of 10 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1220,6 +1220,41 @@ void VoltDBEngine::rebuildTableCollections() {
...
+    assert(!tcd1->materialized());
+    assert(!tcd2->materialized());
+    // Point the Map from signature hash point to the correct persistent tables
+    int64_t hash1 = *reinterpret_cast<const int64_t*>(tcd1->signatureHash());
+    m_tablesBySignatureHash[hash1] = table2;
+    int64_t hash2 = *reinterpret_cast<const int64_t*>(tcd2->signatureHash());
+    m_tablesBySignatureHash[hash2] = table1;
+
+    // Generate swap table DREvent
+    //TODO: Do this only if the tables are drEnabled
```
nshi: Why is this a TODO? I think it's important that we only generate events for DR tables, otherwise the consumer may never generate matching events to unpause the DR stream.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1220,6 +1220,41 @@ void VoltDBEngine::rebuildTableCollections() {
     resetDRConflictStreamedTables();
 }
 
+void VoltDBEngine::swapDRActions(PersistentTable* table1, PersistentTable* table2) {
```
nshi: Both parameters can be const.

src/ee/storage/DRTupleStream.cpp
```diff
@@ -623,7 +632,21 @@ void DRTupleStream::generateDREvent(DREventType type, int64_t lastCommittedSpHan
 
         extendBufferChain(0);
 
-        pushPendingBlocks();
+        pushPendingBlocks(); //TODO: extendBufferChain does pushPendingBlocks. So do we need this here again?
+        break;
+    }
+    case SWAP_TABLE : { // Similar to other events, except that
+        // this happens within a begin and end transaction, so actions done there are not needed here.
+        m_eventTxn = true;
```
nshi: Why does the swap table event need this special flag?

src/ee/storage/DRTupleStream.cpp
```diff
@@ -623,7 +632,21 @@ void DRTupleStream::generateDREvent(DREventType type, int64_t lastCommittedSpHan
...
+        // Make sure current block is empty
+        extendBufferChain(0);
+        ExportSerializeOutput io(m_currBlock->mutableDataPtr(), m_currBlock->remaining());
+        io.writeBinaryString(payloads.data(), payloads.length());
+        m_currBlock->consumed(io.position());
+        m_uso += io.position();
+
+        m_currBlock->startDRSequenceNumber(m_openSequenceNumber);
+        m_currBlock->markAsEventBuffer(type);
```
nshi: This completely duplicates part of the code block above, can they be combined or at least extracted into a function?

src/ee/storage/DRTupleStream.cpp
```diff
@@ -623,7 +632,21 @@ void DRTupleStream::generateDREvent(DREventType type, int64_t lastCommittedSpHan
 
         extendBufferChain(0);
 
-        pushPendingBlocks();
+        pushPendingBlocks(); //TODO: extendBufferChain does pushPendingBlocks. So do we need this here again?
+        break;
+    }
+    case SWAP_TABLE : { // Similar to other events, except that
+        // this happens within a begin and end transaction, so actions done there are not needed here.
```
nshi: Why does swap table happen within a begin/end pair and not the other cases?

src/ee/storage/persistenttable.cpp
```diff
@@ -635,6 +635,7 @@ void PersistentTable::swapTable(PersistentTable* otherTable,
     swapTableIndexes(otherTable,
             compiled.m_theIndexes,
             compiled.m_otherIndexes);
+    ExecutorContext::getEngine()->swapDRActions(this, otherTable);
```
nshi: Are the two tables already swapped at this point? It probably doesn't matter since they are commutative, but for the purpose of generating log messages, it may make sense to retain the ordering.

src/ee/execution/VoltDBEngine.cpp
manjujames: Oh, I will do it. I just wanted to make the PR so that we can verify that this approach will work.

src/ee/storage/DRTupleStream.cpp
manjujames: Because we are calling generate swap table event from with swap table transaction. The others are called from outside.

src/ee/storage/DRTupleStream.cpp
manjujames: I need this to push the buffer in endTransaction.

src/ee/storage/DRTupleStream.cpp
manjujames: Yes, I will.

src/ee/storage/persistenttable.cpp
manjujames: Yes. And this also uncovers a bug :)

src/ee/storage/DRTupleStream.cpp
nshi: Isn't schema change also done within a transaction?

src/ee/storage/DRTupleStream.cpp
nshi: Why does it have to be flushed immediately?

src/ee/storage/DRTupleStream.cpp
manjujames: I am not exactly sure. But, all the other generate dr events happen outside of EE's drstream's beginTxn and endTxn. I can look more to figure out exactly how they are called.

src/ee/storage/DRTupleStream.cpp
manjujames: If I understand correctly, that is the say we ensure that an event is in its own invocation buffer.

src/ee/execution/VoltDBEngine.cpp
manjujames: We realized that hash value is also stored in PersistentTable, which we need to update here. So we cannot make it const anymore.

src/ee/storage/DRTupleStream.cpp
slowsoul: I looked into this a bit, for UAC (generates catalog update event) and ExecuteTask (generates DR stream start event), we don't call Site.executePlanFragments(). For @SwapTables, since it's planned and then wrapped in a @AdHoc_RW_MP, Site.executePlanFragments() is called. What happens in this function is that it calls EE through JNI and if the transaction is a Read-write MP transaction, checkTransactionForDR is called to make sure at least an empty begin-end txn pair will be generated. In @SwapTables' case, every dr stream will have an empty begin-end txn pair. Since we do need to generate DR event in this case, additional flags and processing are needed. Haven't thought of a good way to reconcile this yet. One way is to remove the begin txn when we see that it's actually generating a dr event, so that when it is pushed to Java side, the begin txn portion doesn't need to be skipped, but this still requires additional flags and processing.

PR #4329, opened 2017-03-06
Eng 11881 updatelogging

luyangco:


lib/python/voltcli/voltadmin.d/log4j.py
```diff
@@ -17,21 +17,27 @@
...
+    arguments=(
         VOLT.PathArgument('log4j_xml_path', 'the Log4J configuration file', exists=True),
     )
 )
 def log4j(runner):
     log4j_file = utility.File(runner.opts.log4j_xml_path)
+    sockname = runner.client.socket.getsockname();
+    hostname = sockname[0] + ':' + str(sockname[1])
+    if runner.opts.username is None:
+        username = 'An anonymous user'
```
xinjiacs: Discuss with Andrew about it? Probably mentioning "security is off".

src/frontend/org/voltcore/logging/VoltLog4jLogger.java
```diff
@@ -67,14 +67,14 @@ public void run() {
...
      */
-    public static final int all = 0;
-    public static final int trace = 1;
-    public static final int debug = 2;
-    public static final int info = 3;
-    public static final int warn = 4;
-    public static final int error = 5;
-    public static final int fatal = 6;
-    public static final int off = 7;
+    public static final long all = 0;
```
xinjiacs: why long?

luyangco: Those static values are used in `getLogLevels()`, in the function `logLevels` is a long but it does logical OR with a integer.  When the integer overflows, to a negative number, the higher 32 bit of `logLevels` will set to 1. 
I want to utilize all the 64 bits, so I changed these values to long.

src/frontend/org/voltcore/logging/VoltLoggerDiffUtils.java
```diff
@@ -0,0 +1,98 @@
...
+            "GC",               /* 15 */
+            "DRAGENT",
+            "DR",
+            "DBBPOOL",
+            "CSVLOADER",
+            "LOADER",           /* 20 */
+            "CONSOLE"
+    };
+
+    /* Each logger takes 3 bits, so a 64 bit long is perfect for 21 loggers */
```
xinjiacs: @luyangco How could we extend more loggers easily not limited with 21 loggers?

src/frontend/org/voltcore/logging/VoltLoggerDiffUtils.java
```diff
@@ -0,0 +1,98 @@
...
+            "REJOIN",
+            "JOIN",
+            "SNAPSHOT",
+            "GC",               /* 15 */
+            "DRAGENT",
+            "DR",
+            "DBBPOOL",
+            "CSVLOADER",
+            "LOADER",           /* 20 */
+            "CONSOLE"
```
xinjiacs: How do you make sure these 21 loggers contains all the loggers that we currently used in VoltDB?

src/frontend/org/voltdb/sysprocs/UpdateLogging.java
```diff
@@ -37,6 +41,9 @@
  */
 public class UpdateLogging extends VoltSystemProcedure
 {
+    private static final VoltLogger authLog = new VoltLogger("AUTH");
```
xinjiacs: I do not think this should be logged in "AUTH", I think it should be on "HOST" log, just like UAC. 
"AUTH" is for authentication system only in our system currently.

src/frontend/org/voltdb/sysprocs/UpdateLogging.java
```diff
@@ -59,11 +66,37 @@ public DependencyPair executePlanFragment(
      */
     @SuppressWarnings("deprecation")
     public VoltTable[] run(SystemProcedureExecutionContext ctx,
+                           String host, String username,
                            String xmlConfig)
     {
-        VoltDB.instance().logUpdate(xmlConfig, DeprecatedProcedureAPIAccess.getVoltPrivateRealTransactionId(this));
+        long oldLevels  = 0;
+        if (ctx.isLowestSiteId()) {
```
xinjiacs: Maybe add a comment saying that "log the UpdateLogging event before it changes".

src/frontend/org/voltdb/sysprocs/UpdateLogging.java
```diff
@@ -59,11 +66,37 @@ public DependencyPair executePlanFragment(
...
+            long newLevels = VoltLoggerDiffUtils.getLoggerLevels();
+            if (newLevels != oldLevels) {
+                // Try best to log the message
+                Level before = VoltLoggerDiffUtils.parseLevel(oldLevels, "AUTH");
+                if (before.ordinal() > Level.INFO.ordinal()) {
+                    authLog.info(String.format("%s from %s changed the log4j settings", username, host));
+                }
+                // Log the difference
+                authLog.info(VoltLoggerDiffUtils.diff(oldLevels, newLevels));
```
xinjiacs: What John Hugg mentioned was that you can just print all the new XML config file.  Make the problem simpler.

src/frontend/org/voltcore/logging/VoltLoggerDiffUtils.java
luyangco: If logger number exceeds 21 then we can use `BigInteger`, which basically has no limit on how many bits it can store.

src/frontend/org/voltdb/sysprocs/UpdateLogging.java
luyangco: You're right, it may not worth to put much effort on printing out the actual difference.

src/frontend/org/voltdb/sysprocs/UpdateLogging.java
```diff
@@ -59,12 +69,42 @@ public DependencyPair executePlanFragment(
      */
     @SuppressWarnings("deprecation")
     public VoltTable[] run(SystemProcedureExecutionContext ctx,
+                           String username,
+                           String remoteHost,
                            String xmlConfig)
     {
+        long oldLevels  = 0;
+        if (ctx.isLowestSiteId()) {
```
xinjiacs: @luyangco Can you add a usage comment on the use of "ctx.isLowestSiteId()" and "doneSignal"?

PR #4330, opened 2017-03-06
sql_grammar_generator.py: fixed to allow using --seed= (i.e., a blank

johnTcrawford:
seed), without crashing (the python test code).


johnTcrawford:
This ("--seed=", i.e., a blank seed) does not fail, because an empty string is considered False in python; if someone wrote "--seed=arglebargle" then it will indeed crash, as I believe it should, since that is nonsense, so the person ought to resubmit the Jenkins job, with a sensible seed value (possibly blank), rather than the test program making some assumption about what the user meant.

PR #4332, opened 2017-03-07
ENG-11662: Relieve internal adapter lock contention.

nshi:
Instead of using a global internal adapter for all importer
transactions, create one adapter for each partition lazily. This
relieves the contention on the ClientInterfaceHandleManager because the
usage will be bound to the same thread for each partition.

The adapters are created when partition leaders are elected so that it
can also pick up elastically added partitions.

### DIFF
--- a/src/frontend/org/voltdb/ClientInterface.java
+++ b/src/frontend/org/voltdb/ClientInterface.java
@@ -79,6 +79,7 @@ import org.voltdb.common.Constants;
 import org.voltdb.dtxn.InitiatorStats.InvocationInfo;
 import org.voltdb.iv2.Cartographer;
 import org.voltdb.iv2.Iv2Trace;
+import org.voltdb.iv2.MpInitiator;
 import org.voltdb.messaging.FastDeserializer;
 import org.voltdb.messaging.InitiateResponseMessage;
 import org.voltdb.messaging.Iv2EndOfLogMessage;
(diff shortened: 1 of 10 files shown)


src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -127,7 +128,7 @@
     public static final long SNAPSHOT_UTIL_CID          = Long.MIN_VALUE + 2;
     public static final long ELASTIC_JOIN_CID           = Long.MIN_VALUE + 3;
     // public static final long UNUSED_CID (was DR)     = Long.MIN_VALUE + 4;
-    public static final long INTERNAL_CID               = Long.MIN_VALUE + 5;
+    // public static final long UNUSED_CID              = Long.MIN_VALUE + 5;
```
akhanzode: Can we remove commented code?

src/frontend/org/voltdb/InternalConnectionHandler.java
```diff
@@ -194,7 +194,7 @@ public boolean callProcedure(InternalConnectionContext caller, InternalConnectio
     }
 
     private boolean hasBackPressure() {
-        final boolean b = m_adapter.hasBackPressure();
+        final boolean b = m_adapters.values().stream().anyMatch(InternalClientResponseAdapter::hasBackPressure);
```
akhanzode: Nice 🥇

src/frontend/org/voltdb/ClientInterface.java
nshi: I think the purpose of leaving it here just like the line above is to leave a trace of why this connection ID should not be reused in the short term.

PR #4333, opened 2017-03-07
Swaptable dr eng 12005

manjujames:


src/ee/storage/persistenttable.cpp
```diff
@@ -635,12 +637,26 @@ void PersistentTable::swapTable(PersistentTable* otherTable,
...
+    assert(isDREnabled() == otherTable->isDREnabled());
+
+    if (!isUndo && isDREnabled()) { // Also called as undo action. Generate DR event if it isn't undo.
+        ExecutorContext::getEngine()->swapDRActions(otherTable, this);
+    }
+
     if (fallible) {
+        DRTupleStreamUndoAction *drUndo = NULL;
         UndoQuantum *uq = ExecutorContext::currentUndoQuantum();
+        if (isDREnabled()) {
```
wweiss-voltdb: What if isUndo is true?

src/ee/storage/persistenttable.cpp
```diff
@@ -635,12 +637,26 @@ void PersistentTable::swapTable(PersistentTable* otherTable,
...
+            ExecutorContext* ec = ExecutorContext::getExecutorContext();
+            AbstractDRTupleStream* drStream = getDRTupleStream(ec);
+            size_t drMark = drStream->m_uso - DRTupleStream::BEGIN_RECORD_SIZE;
+            drUndo = new (*uq) DRTupleStreamUndoAction(drStream, drMark, 0);
+        }
         uq->registerUndoAction(
```
wweiss-voltdb: Same here.

src/ee/storage/persistenttable.cpp
slowsoul: Notes from the discussion: fallible is actually now always true for swapTable (only false when called from the corresponding undo action). fallible for swapTable can be considered always true if we treat it as a DML (which may be the case in the future) as then failure in other sql statements in the same transaction may undo it. However, for the current implementation, swapTable is actually always a standalone transaction so it should never fail (catalog version mismatch would have failed it in the Java portion of AdHoc), so in this sense, we can say fallible is always false. In either case, fallible would only have one value, so maybe we can remove the fallible parameter and simplify the code a bit.

src/ee/storage/persistenttable.cpp
slowsoul: The "drMark" used for DRTupleStreamUndoAction should be the value of drStream->m_uso saved before swapDRActions() is called, because at this time m_uso already counts in the generated swap table event. Also, it doesn't have to subtract BEGIN_RECORE_SIZE because there is a dedicated undo action for the beginTxn already in ExecutorContext::checkTransactionForDR().

src/ee/storage/persistenttable.cpp
manjujames: For undo action, this method is always called with fallible=false. So, this will not get inside this 'if' block for undo action.

manjujames: Rui, the fallible flag and if is Chris' work. He did talk about considering some change in undo area. I will mention to him again.

src/ee/storage/persistenttable.cpp
manjujames: Yes, you are right. I'll fix it.

manjujames: Actually, now that I am thinking more about it, I don't think we need an undo here at all. Because the undo for begin txn will rewind it to where it should be and hence this is unnecessary.

PR #4335, opened 2017-03-07
Eng 11952 specify schema at init

brians127:
Follow on work required, see ENG-11444.


lib/python/voltcli/voltdb.d/init.py
```diff
@@ -26,10 +26,17 @@
                           'Specifies the root directory for the database. The default is voltdbroot under the current working directory.',
                           default = None),
         VOLT.BooleanOption('-f', '--force', 'force',
-                           'Start a new, empty database even if the VoltDB managed directories contain files from a previous session that may be overwritten.')
+                           'Start a new, empty database even if the VoltDB managed directories contain files from a previous session that may be overwritten.'),
+        VOLT.StringOption('-s', '--schema', 'schema',
+                           'Specifies a file containing the data definition (as SQL statements) to be loaded when starting the database.'
+                           ' voltdbroot/startup/bootstrap/existing/ddl.sql is used if it exists and no schema is specified, and overwritten with the new schema if provided.'),
```
akhanzode: This may be too verbose. Ask Andrew for these help messages.

lib/python/voltcli/voltdb.d/init.py
```diff
@@ -26,10 +26,17 @@
...
+        VOLT.StringOption('-p', '--procedures', 'procedures_jarfile',
+                           'Specifies a .jar file containing stored procedures to be loaded upon starting the database.'
+                           ' The default is voltdbroot/startup/bootstrap/existing/procedures.jar is used if it exists and no file is specified, and overwritten with the new file if provided.')
```
akhanzode: Same as above

src/frontend/org/voltdb/Inits.java
```diff
@@ -284,6 +284,7 @@ public void run() {
 
         @Override
         public void run() {
+            // FIXME - will need to compare catalog with leader's if we aren't the leader
```
akhanzode: Remove FIXME also wrong place for the comment you added for fixing.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -831,7 +845,8 @@ public void initialize(Configuration config) {
                     consoleLog.info(msg);
                     hostLog.info(msg);
                 }
-                stageDeploymemtFileForInitialize(config, readDepl.deployment);
+                stageDeploymentFileForInitialize(config, readDepl.deployment);
```
akhanzode: What changed for this line?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2142,6 +2157,28 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
+     * Takes the schema and stored procedures files given at initialization and performs the following tasks:
+     * <p><ul>
+     * <li>creates if necessary the voltdbroot/starter/bootstrap/existing directory tree
+     * <li>fail if database artifacts already exist and --force was not supplied (this is unexpected due to earlier checks)
+     * <li>moves the specified files to the 'existing' directory
+     * </ul>
+     * @param config VoltDB configuration
+     */
+    private void stageSchemaFiles(Configuration config){
+        if (config.m_userSchema == null){
```
akhanzode: add check for trim.isEmpty()

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2142,6 +2157,28 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
+            return; // nothing to do
+        }
+        File stagedCatalogFH = new VoltFile(getStagedCatalogPath());
```
akhanzode: What happens if stagedCatalogFH exists do you overwrite? init with --schema twice.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -641,8 +644,13 @@ else if (arg.equalsIgnoreCase("force")) {
                     m_getOutput = args[++i].trim();
                 } else if (arg.equalsIgnoreCase("forceget")) {
                     m_forceGetCreate = true;
-                }
-                else {
+                } else if (arg.equalsIgnoreCase("schema")){
+                    m_userSchema = new File(args[++i].trim());
+                    if (!m_userSchema.isFile() || !m_userSchema.canRead()){
+                        System.err.println("FATAL: Supplied schema file " + m_userSchema + " is not a readable file.");
```
akhanzode: Check with Andrew for messaging.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -1158,7 +1166,7 @@ public static void crashLocalVoltDB(String errMsg, boolean stackTrace, Throwable
         wasCrashCalled = true;
         crashMessage = errMsg;
         if (ignoreCrash) {
-            throw new AssertionError("Faux crash of VoltDB successful.");
+            throw new CrashCalledException();
```
akhanzode: Why refactor this? Avoid non required changes.

src/frontend/org/voltdb/RealVoltDB.java
brians127: Spelling correction - "Deployment" was misspelled.

src/frontend/org/voltdb/RealVoltDB.java
brians127: As long as stageSchemaFiles() happens after stageDeploymentFileForInitialize(), this cannot happen unless the user supplies --force. Therefore overwriting the file is the correct behavior. I'll make it more clear that this order of events needs to be preserved when other people modify the code.

src/frontend/org/voltdb/VoltDB.java
brians127: I needed a way to detect that these error paths got called during negative tests, and there are many other ways that AssertionError can get thrown. Being unable to distinguish between a pre- or post-condition violation and a deliberate abort is problematic from a test perspective, so I added a new exception type that extends AssertionError (for compatibility with existing tests). I'm open to suggestions on how to best detect that the product reacted correctly to bad input.

src/frontend/org/voltdb/common/Constants.java
```diff
@@ -79,4 +79,6 @@
 
     public static final String DBROOT = "voltdbroot";
     public static final String CONFIG_DIR = "config";
+    public static final String CATALOG_FILE_NAME = "catalog.jar";
```
hhakimi53: For catalog related constants, can put them in CatalogUtil.java

src/frontend/org/voltdb/VoltDB.java
hhakimi53: Suggestion: can update the assertion error message to "Faux crash of local VoltDB successful" and test for that particular msg in test. It may impact other tests too though.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2145,6 +2160,29 @@ private void stageDeploymemtFileForInitialize(Configuration config, DeploymentTy
...
+     * Takes the schema and stored procedures files given at initialization and performs the following tasks:
+     * <p><ul>
+     * <li>creates if necessary the voltdbroot/starter/bootstrap/existing directory tree
+     * <li>fail if database artifacts already exist and --force was not supplied (this is unexpected due to earlier checks)
+     * <li>moves the specified files to the 'existing' directory
+     * </ul>
+     * @param config VoltDB configuration
+     */
+    private void stageSchemaFiles(Configuration config){
+        if ((config.m_userSchema == null) || (config.m_userSchema.equals(""))){
```
hhakimi53: Instead of config.m_userSchema.equals("")  can use config.m_userSchema.trim().isEmpty(). It will cover additional cases.

lib/python/voltcli/voltdb.d/init.py
akhanzode: Did you check with Andrew on messages?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -558,12 +567,17 @@ private void managedPathsEmptyCheck(Configuration config) {
...
-        }
         PathsType paths = deployment.getPaths();
         String voltDbRoot = getVoltDBRootPath(paths.getVoltdbroot());
         String path;
+
+        if ((path = managedPathEmptyCheck(voltDbRoot, getStagedCatalogPath())) != null){
+            nonEmptyPaths.add(path);
+        }
+
+        if (!config.m_isEnterprise) {
```
akhanzode: We should allow in community version priming artifacts.

src/frontend/org/voltdb/VoltDB.java
akhanzode: Still not sure what you mean way to detect is it for junits?

src/frontend/org/voltdb/RealVoltDB.java
brians127: We do. There are other artifacts to check for that are Enterprise specific, and this is why the Community code bails out here.

src/frontend/org/voltdb/VoltDB.java
brians127: I customarily avoid relying on error strings for behavioral decisions, under the assumption that the message may undergo UX refinements after the initial commit. I'll change this to be more consistent with the rest of VoltDB.

lib/python/voltcli/voltdb.d/init.py
ajgent: This is unrelated to the task at hand, but the opening message for --force is wrong. It should be "Initialize a new, empty database..." Not "Start a new..." This is the init command, not the start command.

For the --schema option, I'd keep it simple:
"Specifies a file containing the schema (DDL statements) to be loaded when starting the database."
the second sentence is confusing and unnecessary. (We don't tell people where we put things in the root directory.)

lib/python/voltcli/voltdb.d/init.py
ajgent: Same as above... :-)

By that I mean, the first sentence is fine. The second is unnecessary and should be dropped.

I also have serious concerns about the option name. At the least, it should be reduced to two (-p and--procedures). More importantly, I'd actually suggest changing it to -c and --classes to be consistent with all our other class loading syntax.  However, if we do change it to --classes, we should change the description to match. Something like: 
"Specifies a .jar file containing classes used to declare stored procedures. The classes are loaded automatically when the database starts."

But a decisions about whether to go with --procedures or --classes probably needs wider agreement.

brians127: Interesting point - I assumed (perhaps incorrectly) that discussion had already taken place. I'll drop the second sentence since that's uncontroversial. My plan is to demo this tomorrow if it's in master - thoughts?

akhanzode: I would go with --classes as per your document which had a good compelling argument about sql syntax similarity load classes, remove classes and such. I am OK not having a -c to avoid confusion with -C but I am ok if we want to go with -c

akhanzode: Also consistent with voltdb get classes

src/frontend/org/voltdb/RealVoltDB.java
brians127: m_userSchema is a file, not a string. I found that out when trying to make the change and left an assert+comment behind with my thoughts.

lib/python/voltcli/voltdb.d/init.py
hhakimi53: Yes, by having classes will have similarity with "voltdb get " - how the user would fetch procs from current database. For short hand notation for --classes, can use "-p"

lib/python/voltcli/voltdb.d/init.py
```diff
@@ -26,10 +26,15 @@
                           'Specifies the root directory for the database. The default is voltdbroot under the current working directory.',
                           default = None),
         VOLT.BooleanOption('-f', '--force', 'force',
-                           'Start a new, empty database even if the VoltDB managed directories contain files from a previous session that may be overwritten.')
+                           'Initialize a new, empty database. Any previous session will be overwritten.'),
+        VOLT.StringOption('-s', '--schema', 'schema',
+                           'Specifies a file containing the data definition (as SQL statements) to be loaded when starting the database.'),
+        VOLT.StringOption('-j', '--classes', 'classes_jarfile',
+                          'Specifies a .jar file containing classes used to declare stored procedures. The classes are loaded automatically from a saved copy when the database starts.')
```
akhanzode: The part of help where it says "The classes are loaded automatically from a saved copy" is misleading its only used once in certain cases. I would just stop at first full stop.

brians127: What I am hoping to make clear is that "start" does not depend on the supplied jar file. You're right though - if I can't find a better wording I'll remove the second sentence.

src/frontend/org/voltdb/VoltDB.java
brians127: @ajgent Have you reviewed this message?

brians127:
During the demo, I received feedback that "voltdb init --classes" should handle the same schema files that 'sqlcmd' does. I am restructuring things so that will work, but otherwise am assuming the review feedback has been addressed.

src/frontend/org/voltdb/VoltDB.java
ajgent: Hi brian. What does "not readable" mean? Does it mean we can't open it or it is ill-formatted (like if I specify a deployment file rather than a schema file)?  Does this also cover if the file doesn't exist?

brians127: This error covers:
* File doesn't exist
* File isn't a file (for example, it's a directory) 
* File does not have read permissions set

It does not cover files being in the wrong format - the VoltCompiler is assumed to handle that variety of bad input.

src/frontend/org/voltdb/RealVoltDB.java
ajgent: Does the user get any more information that "could not compile..."? Do they get an error indicating why it did not compile?

brians127:
Confirmed verbally with Anish and Andrew that this is ready to merge.

PR #4336, opened 2017-03-07
sql_grammar_generator.py: added (to the summary file) messages giving

johnTcrawford:
the number of certain types of error messages found in the log or
console output files, e.g.:
Number of 'Error compiling query' errors: 2
Number of 'ERROR: IN: NodeSchema' errors: 5
Number of all other error messages      : 0
Total Number of all 'ERROR' messages    : 7
(that example is from a one-minute run on my laptop).


BillWhite:
Seems plausible to me.  The logic for getting the number of errors seems fragile, but not so terrible that it needs to be reworked until it breaks, if it ever breaks.  Getting output from processes in python is always kind of a crapshoot anyway.

johnTcrawford:
Thanks. Yeah, I agree that it's fragile, but I'm basically interested in whether there are a large number, a small number, or zero errors, in each of these categories (compiling, NodeSchema, or others we're not tracking explicitly), so, as you imply, this seems good enough.

PR #4338, opened 2017-03-08
Eng 12011 bshaw investigation

brians127:
This attempts to address a problem where an exception is logged by the site thread because it queried the size of a closed PBD. Hypothesis is that a statistics poll asynchronously collided with shutdown in a way that isn't currently handled. Changes resolve issues found via code inspection and leave breadcrumbs that will make the issue less ambiguous should it recur.


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -455,9 +455,13 @@ public Long call() throws Exception {
             }
         } catch (RejectedExecutionException e) {
             return 0;
-        } catch (Throwable t) {
-            Throwables.propagate(t);
+        } catch (IOException e){
+            assert e.getMessage().contains("has been closed") : e.getMessage();
             return 0;
+        } catch (Throwable t) {
+            // IOException is expected if the committed buffer was closed when stats are requested.
```
akhanzode: move this comment in previous catch block.

PR #4340, opened 2017-03-08
ENG-11536

akhanzode:
Allow to use Binary data for importer formatter.


src/frontend/org/voltdb/importer/formatter/Formatter.java
```diff
@@ -29,4 +29,5 @@
      * @throws FormatException
      */
     public Object[] transform(T sourceData) throws FormatException;
+    public Object[] transform(T sourceData, int offset, int length) throws FormatException;
```
nshi: I don't think you need this overloaded method. I'm fine getting rid of the template and just use ByteBuffer as the parameter type. With a ByteBuffer, you can set the current position to the offset and set limit to offset+length, then you don't need to pass the offset and length in.

akhanzode: I will do that...Makes it easy.

PR #4343, opened 2017-03-09
ENG-11890 proceduredetail output

yzhang1991:
I spent some time putting voltdbipc to work, there is still some work in the `TestProcedureStats`.
But I think I should sent the rest part to review first.


src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -132,6 +137,24 @@
...
+         * coordinator fragments.
+         * The worker fragments will first be sent to all the participating SP (single partition)
+         * sites for execution.
+         * Then we will select one SP site to execute the coordinator fragment.
+         * Every MpTransactionState has a m_buddyHSId to indicate which SP site is selected.
+         *
+         * Therefore, on the SP site *m_buddyHSId*, the per-fragment statistics include the statistics
+         * from two fragments (worker + coordinator).
+         * On this site, the worker fragment statistics will be cached and the calculations will be postponed.
+         * We will then do the calculation when we receive the coordinator fragment statistics.
```
wolffcm: Thanks for this comment, things are becoming clearer to me now.

src/ee/voltdbipc.cpp
```diff
@@ -786,8 +790,9 @@ void VoltDBIPC::executePlanFragments(struct ipc_command *cmd) {
     ReferenceSerializeInputBE serialize_in(offset, sz);
 
     // and reset to space for the results output
-    m_engine->resetReusedResultOutputBuffer(1);//1 byte to add status code
-    m_engine->resetPerFragmentStatsOutputBuffer();
+    m_engine->resetReusedResultOutputBuffer(1); // 1 byte to add status code
+    cout << "perFragmentTimingEnabled " << queryCommand->perFragmentTimingEnabled << endl;
```
yzhang1991: Yeah, I forgot this line.

src/frontend/org/voltdb/ProcedureDetailResultTable.java
```diff
@@ -0,0 +1,177 @@
...
+            new ColumnInfo("AVG_PARAMETER_SET_SIZE", VoltType.INTEGER),
+            new ColumnInfo("ABORTS", VoltType.BIGINT),
+            new ColumnInfo("FAILURES", VoltType.BIGINT));
+
+    public ProcedureDetailResultTable(VoltTable table) {
+        assert(table != null);
+        m_rows = new ArrayList<ProcedureDetailResultRow>(table.getRowCount());
+        table.resetRowPosition();
+        while (table.advanceRow()) {
+            m_rows.add(new ProcedureDetailResultRow(
```
yzhang1991: Using `VoltTableRow` will not let me do in-place sorting conveniently.
Basically `VoltTableRow` is a pointer to the `VoltTable` with a memory offset and a wrapper for reading values.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -212,7 +214,7 @@ public String visitGroovy(VoltProcedure p) {
 
         // Map each statement that was defined in the stored procedure with its SQLStmt variable name.
         // The variable name is used in the granular statistics for stored procedures.
-        Map<SQLStmt, String> reversedStmtMap = reflect();
+        ArrayList<String> stmtList = reflect();
```
yzhang1991: Fixed it

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -1504,8 +1548,8 @@ void addStatement(int index, SQLStmt stmt, ByteBuffer params, SiteProcedureConne
                m_depsForLocalTask[index] = outputDepId;
                // Add local and distributed fragments.
                if (stmt.inCatalog) {
-                   m_localTask.addFragment(stmt.aggregator.planHash, m_depsToResume[index], params);
-                   m_distributedTask.addFragment(stmt.collector.planHash, outputDepId, params);
+                   m_localTask.addFragment(stmt.aggregator.planHash, stmt.getStmtName(), m_depsToResume[index], params);
+                   m_distributedTask.addFragment(stmt.collector.planHash, stmt.getStmtName(), outputDepId, params);
```
yzhang1991: Today's benchmarking result shows that this is not a performance killer.

tests/frontend/org/voltdb/regressionsuites/statistics/TestProcedureDetails.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+            finally {
+                // Note that pass 1 as the second parameter to get incremental statistics.
+                VoltTable procedureDetail = client.callProcedure("@Statistics", "PROCEDUREDETAIL", 1).getResults()[0];
+                System.out.println(procedureDetail.toFormattedString());
+                validateProcedureDetail(testConfig, procedureDetail);
+            }
+            // The test configuration says an exception is expected, but we did not get it.
+            if (testConfig.expectsException()) {
+                throw new Exception(String.format("Expects an exception from exec %s %d '%s', but did not get it.",
```
yzhang1991: Thanks for pointing out.

tests/frontend/org/voltdb/iv2/TestMpTransactionState.java
```diff
@@ -303,8 +315,10 @@ public void testMultiSitePartitionedRead() throws IOException
 
         // This will be passed a FragmentTaskMessage with no deps
         dut.createAllParticipatingFragmentWork(plan.remoteWork);
+        non_local = configureHSIdsForMailbox(hsids, buddyHSId);
```
yzhang1991: This is no longer needed in a later commit I made. It is related to a behavioral change I made in MpTransactionState.

src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -164,13 +164,34 @@ public void createAllParticipatingFragmentWork(FragmentTaskMessage task)
...
-            // send to all non-local sites
-            if (non_local_hsids.length > 0) {
-                m_mbox.send(non_local_hsids, m_remoteWork);
+            if (m_useHSIds.size() > 0) {
+                // For all SP sites other than site *m_buddyHSId*,
+                // send the FragmentTaskMessage with commitPerFragmentStats = true (default).
+                // Those sites does not need to execute another coordinator task,
+                // so they should immediately commit their statistics changes.
+                // For some system stored procedures, the m_buddyHSId may not be present in m_useHSIds.
+                // So we need to handle that differently.
```
yzhang1991: This is also reverted. We have a better way to let this part of the code remain simple.

src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -38,157 +38,41 @@
...
                                    ArrayList<String> stmtNames) {
         super(siteId, false);
         m_partitionId = partitionId;
         m_procName = catProc.getClassname();
-        // Use LinkedHashMap to have a fixed element order.
-        m_stats = new LinkedHashMap<String, ProcedureStmtStat>();
-        // Use one ProcedureStmtStat instance to hold the procedure-wide statistics.
-        m_procStat = new ProcedureStmtStat("<ALL>");
+
+        m_stmtStatsMap = new HashMap<String, StatementStats>();
```
wolffcm: The keys in this map represent the statement name from the Java code?  Then why use `null` key entry instead of `"<ALL>"` for the procedure-wide entry?

src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -38,157 +38,41 @@
...
-        m_stats.put(null, m_procStat);
-        // Add statistics for the individual SQL statements.
+        m_stmtStatsMap.put(null, procedureWideStats);
+        // Add stats entry for each of the individual SQL statements.
         if (stmtNames != null) {
             for (String stmtName : stmtNames) {
-                m_stats.put(stmtName, new ProcedureStmtStat(stmtName));
+                // If the procedure is a multi-partition one, its statements will have coordinator tasks.
+                boolean hasCoordinatorTask = ! catProc.getSinglepartition();
+                m_stmtStatsMap.put(stmtName, new StatementStats(stmtName, hasCoordinatorTask));
```
wolffcm: There is something called "sendless DML" that I don't understand thoroughly which might not have a coordinator fragment even though its collector fragment is executed on all sites.

wolffcm: Maybe it's the case that there is no coordinator fragment, but there is a coordinator task though.

src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -197,128 +81,115 @@ public ProcedureStatsCollector(long siteId, int partitionId, Procedure catProc,
...
+            m_procStatsData.m_currentStartTime = System.nanoTime();
         }
     }
 
+    /**
+     * @return a boolean values indicating whether the current running procedure is sampled.
+     */
     public final boolean isRecording() {
-        return m_procStat.m_currentStartTime > 0;
+        return m_procStatsData.m_currentStartTime > 0;
```
wolffcm: This won't work if someone runs VoltDB before 1970?  Just joking, this is fine.

src/frontend/org/voltdb/ProcedureStatsCollector.java
```diff
@@ -197,128 +81,115 @@ public ProcedureStatsCollector(long siteId, int partitionId, Procedure catProc,
...
+        // Cannot use the isRecording() value here because SP sites can have values different from the MP Site.
         if (! perFragmentStatsRecording) {
             return;
         }
         // This is a sampled invocation.
         // Update timings and size statistics below.
         if (duration < 0) {
             if (Math.abs(duration) > 1000000000) {
-                log.info("Statement: " + stat.m_stmtName + " in procedure: " + m_procName +
+                log.info("Statement: " + stmtStats.m_stmtName + " in procedure: " + m_procName +
```
wolffcm: Negative execution time?  When would this happen?

wolffcm: I guess it could be clock skew between nodes?

src/frontend/org/voltdb/StatementStats.java
```diff
@@ -0,0 +1,345 @@
...
+
+    public StatementStats(String stmtName, boolean hasCoordinatorTask) {
+        m_stmtName = stmtName;
+        m_workerTask = new StatsData();
+        if (hasCoordinatorTask) {
+            m_coordinatorTask = new StatsData();
+        }
+    }
+
+    // Maybe the worker task got executed and timed multiple times, but all failed.
```
wolffcm: I would think that if any of the worker tasks fail, then the coordinator won't execute.  Do I have that right?  But that seems at odds with what's here.

src/frontend/org/voltdb/StatementStats.java
```diff
@@ -0,0 +1,345 @@
...
+         * Number of timed invocations
+         */
+        long m_timedInvocations = 0;
+        long m_lastTimedInvocations = 0;
+
+        /**
+         * Total amount of timed execution time
+         */
+        long m_totalTimedExecutionTime = 0;
+        long m_lastTotalTimedExecutionTime = 0;
```
wolffcm: Are you sometimes using "last" to indicate the incremental value?  I think I prefer "incr" as a prefix, but it might be good to standardize on one or the other?

yzhang1991: Our code used "last" everywhere here, but I found they have different meanings.
For values other than min and max, the "last" value means the value of the last query.
but for min/max, the "last" value actually means the min/max value **since** the last query.
So for min/max variables here, I changed "last" to "incr" to make it more clear.

yzhang1991:
Merged with command line tools.

PR #4344, opened 2017-03-09
ENG-12009:

wweiss-voltdb:
The @SwapTables could not be executed on a replica cluster because it actually executes the @AdHoc_RW_MP stored procedure, which is not allowed to execute on a replica. This patch actualizes the @SwapTalbes stored procedure and calls the AdHocBase code in the same manner as @AdHoc_RW_MP.

Having this Stored Procedure will also allow us to Extend the ExecuteFragmentTask of AdHocBase to notify each host's DR Consumer that an SwapTables transaction has completed, in the same way that @CatalogUpdate notifies the DRConsumer of a catalog change. This will be done in a seperate patch.


src/frontend/org/voltdb/sysprocs/SwapTables.java
```diff
@@ -0,0 +1,54 @@
...
+@ProcInfo(singlePartition = false)
+public class SwapTables extends AdHocBase {
+
+    Database m_db = null;
+
+    /**
+     * System procedure run hook.
+     * Use the base class implementation.
+     *
+     * @param ctx  execution context
```
wolffcm: This parameter list does not match the method's actual parameters...

PR #4347, opened 2017-03-09
ENG-11662: Importer back pressure and lock contention

nshi:


nshi:
Not backporting to v6.9

PR #4350, opened 2017-03-10
ENG-10627-preSnapshot: reset single DR

JadenW:
Pro Side: https://github.com/VoltDB/pro/pull/1790
This PR only handles the case when target cluster haven't finished applied syncSnapshot.
This guarantees the target cluster won't send any binary logs to any cluster in the mesh. So it guarantees there won't be any divergences in the remaining clusters.
The way to detect syncSnapshot haven't applied is by looking at Producer Cluster's Consumer Side partitionReceivers' existence. (Since we on-demand create them.) As well as site's local applied tracker.(Mainly for the recovery case when the target cluster is also down).

changes including following

## ResetDR sysProc now taken two parameters: clusterId and force
  * executing global reset if clusterId == -1 **or** there's only one consumerCluster Connected
  * for single reset, if force == 0, it will pause and check **all** bufferReceivers for whether received "STREAM_START_EVENT" or not. If anyone hasreceived, it won't be able to safe reset.
  * if force == 1 or it's safe to reset, sysProc calls DRProducer.deactiveDR()
  * DRProducer then request a breakReplication Task to the State Machine. For now, it keeps DR_STATUS and using REMOVE_SINGLE as the DR_STATE_CHANGE_TYPE.
  * This task will then call DRProducer.breakReplication method on each host for doing the real cleanup works, which includes removing conversion and metadata.
  * If this cluster is Active-Active, it will then stop ConsumerDispatcher for the target cluster.

## PAUSE mode in DRBinaryLogPartitionStreamReader
This pauses adding InvocationBuffer to toSend queue.
The aim of adding this is for delaying stream_start_event till sync snapshot applied.

the pause mode now using two different flags:
"m_mode" is used for pauseAllReaders()/resumeAllReaders() to enable the abiltiy to pause readers externally.

"m_covering" follows the original "m_mode" logic, it will be set once resetToDRId() being executed

## PAUSE state for DRConsumerDispatcher
This pauses Dispatcher offer InvocationBuffer to m_dispatcherState.
This aim for adding this is for preventing bufferReceivers received stream_start_event while we are checking whether it's safe to reset or not.

The PAUSE state is a special case of RECEIVE state. The only difference is the m_dispatcherState.m_state changes into PAUSE.
PAUSE state only supports transfer from and to RECEIVE state.

## new DR_STATE_CHANGE_TYPE in ProducerDRState:
 REMOVE_SINGLE

## ConsumerDRGateway new methods:
  * isSafeForReset, pauseConsumerDispatcher, resumeConsumerDispatcher: used in ResetDR sysProc for preChecking
  * deactiveConsumerDispatcher: used from DRProducer to stop its target Consumer Dispatcher

## ProducerDRGateway new methods:
  * deactivateDRProducer(byte clusterId): called from sysProc
  * pauseAllReaders(): this called from (Conumser Cluster) syncSnapshotPreconditions to pause all DRBinaryLogPartitionStreamReader
  * resumeAllReaders(): this called from (Conumser Cluster) transitionToNormalBuffers to resume all DRBinaryLogPartitionStreamReader

## Site can reset per cluster applied tracker
via resetDrAppliedTracker(byte clusterId)

## DRConusmer Stats Enhancement
unregister per cluster stats when removed

### DIFF
--- a/lib/python/voltcli/voltadmin.d/dr.py
+++ b/lib/python/voltcli/voltadmin.d/dr.py
@@ -15,7 +15,7 @@
 # along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
 
 def reset(runner):
-    status = runner.call_proc('@ResetDR', [], []).table(0).tuple(0).column_integer(0)
+    status = runner.call_proc('@ResetDR', [VOLT.FastSerializer.VOLTTYPE_TINYINT, VOLT.FastSerializer.VOLTTYPE_TINYINT], [-1, 0]).table(0).tuple(0).column_integer(0)
     if status == 0:
         runner.info('Conversation log is reset.')
     else:
--- a/lib/python/voltcli/voltadmin.d/dr.py
+++ b/lib/python/voltcli/voltadmin.d/dr.py
@@ -14,8 +14,15 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
 
-def reset(runner):
-    status = runner.call_proc('@ResetDR', [VOLT.FastSerializer.VOLTTYPE_TINYINT, VOLT.FastSerializer.VOLTTYPE_TINYINT], [-1, 0]).table(0).tuple(0).column_integer(0)
+def reset_local(runner):
+    status = runner.call_proc('@ResetDR', [VOLT.FastSerializer.VOLTTYPE_TINYINT, VOLT.FastSerializer.VOLTTYPE_TINYINT], [runner.opts.clusterId, runner.opts.clusterId * 1]).table(0).tuple(0).column_integer(0)
+    if status == 0:
+        runner.info('Conversation log is reset.')
+    else:
+        runner.error('The cluster failed to reset conversation log with status: %d' % status)
+
+def reset_remote(runner):
+    status = runner.call_proc('@ResetDR', [VOLT.FastSerializer.VOLTTYPE_TINYINT, VOLT.FastSerializer.VOLTTYPE_TINYINT], [runner.opts.clusterId, runner.opts.clusterId * 1]).table(0).tuple(0).column_integer(0)
     if status == 0:
         runner.info('Conversation log is reset.')
     else:
@@ -24,7 +31,14 @@ def reset(runner):
 @VOLT.Multi_Command(
     bundles = VOLT.AdminBundle(),
     description = 'DR control command.',
-    modifiers = VOLT.Modifier('reset', reset, 'Clear queued DR buffers.')
+    options = (
+            VOLT.BooleanOption('-f', '--force', 'forcing', 'bypass precheck', default = False),
+            VOLT.BooleanOption('-c', '--cluster', 'clusterId', 'drCluster Id', default = -1),
+    ),
+    modifiers = (
+            VOLT.Modifier('resetlocal', reset_local, 'remove local cluster from its dr cluster mesh.'),
+            VOLT.Modifier('reset', reset_remote, 'reset one/all remote dr cluster(s).'),
+    )
 )
 
 def dr(runner):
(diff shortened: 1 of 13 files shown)


lib/python/voltcli/voltadmin.d/dr.py
```diff
@@ -15,7 +15,7 @@
 # along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
 
 def reset(runner):
-    status = runner.call_proc('@ResetDR', [], []).table(0).tuple(0).column_integer(0)
```
manjujames: You should check with Andrew about the parameters.

src/frontend/org/voltdb/jni/ExecutionEngine.java
```diff
@@ -66,7 +66,8 @@
         GENERATE_DR_EVENT(6),
         RESET_DR_APPLIED_TRACKER(7),
         SET_MERGED_DRID_TRACKER(8),
-        INIT_DRID_TRACKER(9);
+        INIT_DRID_TRACKER(9),
+        RESET_DR_APPLIED_TRACKER_SINGLE(10);
```
slowsoul: Add a corresponding entry to TaskType enum in types.h to make them consistent in EE and Frontend.

src/frontend/org/voltdb/sysprocs/ExecuteTask_SP.java
```diff
@@ -54,7 +54,7 @@ public DependencyPair executePlanFragment(
      * @param partitionParam  key for routing stored procedure to correct site
      * @param taskId          type of the task to execute
      */
-    public void run(SystemProcedureExecutionContext ctx, byte[] partitionParam, byte taskId)
+    public void run(SystemProcedureExecutionContext ctx, byte[] partitionParam, byte taskId, byte clusterId)
```
slowsoul: Currently it's good enough. I just wonder if it's better to pass in a general-purpose array and interpret the content based on the taskId (task type) because not every task needs the cluster id.

src/frontend/org/voltdb/jni/ExecutionEngine.java
JadenW: good catch, thanks.

src/frontend/org/voltdb/sysprocs/ExecuteTask_SP.java
JadenW: Combine taskId and clusterId and any future parameters into a byte[] (similar to that in ExecuteTask) ?

lib/python/voltcli/voltadmin.d/dr.py
```diff
@@ -24,7 +31,14 @@ def reset(runner):
 @VOLT.Multi_Command(
     bundles = VOLT.AdminBundle(),
     description = 'DR control command.',
-    modifiers = VOLT.Modifier('reset', reset, 'Clear queued DR buffers.')
+    options = (
+            VOLT.BooleanOption('-f', '--force', 'forcing', 'bypass precheck', default = False),
+            VOLT.BooleanOption('-c', '--cluster', 'clusterId', 'drCluster Id', default = -1),
```
slowsoul: The description may be "dr cluster id" or just "cluster id". Also, all these descriptions will need to be reviewed by Andrew, right?

src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -453,16 +453,10 @@ public DRIdempotencyResult isExpectedApplyBinaryLog(int producerClusterId, int p
             Map<Integer, DRConsumerDrIdTracker> clusterSources = m_maxSeenDrLogsBySrcPartition.get(producerClusterId);
             if (clusterSources == null) {
                 // Don't have a tracker for this cluster
-                if (DRLogSegmentId.isEmptyDRId(lastReceivedDRId)) {
```
JadenW: revoke this change as it will be addressed in ENG-12179

PR #4351, opened 2017-03-10
Eng 12089 Validate BIGINT, DOUBLE and TIMESTAMP constant values from HSQL.

BillWhite:
Some constants come to us from HSQLDB through VoltXML, to create an AbstractParsedStatement.  If these strings were not suitable for their types, say a BIGINT was too big, we did not validate this.  The EE could see a bad 64 bit value and crash.  This commit fixes this by validating when VoltXML is parsed.


src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -443,6 +446,34 @@ private AbstractExpression parseValueExpression(VoltXMLElement exprNode) {
...
+                            VoltDecimalHelper.stringToDecimal(valueStr);
+                            break;
+                        default:
+                            break;
+                        }
+                    } catch (PlanningErrorException ex) {
+                        // We're happy with these.
+                        throw ex;
+                    } catch (NumberFormatException ex) {
+                        throw new PlanningErrorException("Numeric conversion error to type " + vt.name() + ex.getMessage().toLowerCase());
```
wolffcm: Why convert the message to lower case?  Also, seems like there won't be any whitespace between the type name and the exception message?

BillWhite: The string from NumberFormatException is not helpful, and seems to want to be the tail of a string.  It starts with "For input string: ".  So I convert to lower case to get something that can finish the error message.

I added the extra space, though.

BillWhite:
I don't know where my comment went.  The message for NumberFormatException is not very useful.  It starts with the preposition "For".  It seems as if it wants to be the tail of a better error message, so that's what I did.  I made it lower case so that the beginning of the error message could flow into it.

I did add the missing space, though.

PR #4353, opened 2017-03-10
ENG-11941: create undo action for delete before updating views

wolffcm:
If we don't do this, and updating views fails for some reason, havoc results.


wolffcm:
@BillWhite Not quite.  The call to `appendTuple` is a special action for DR, not part of the main codepath for deleteTuple.

The real important change here (perhaps obscured by my refactoring and not obvious in the diff) is that we are calling `registerUndoAction` immediately after calling `deleteFromAllIndexes`.

Previously, we postponed registering the undo action until after we had also updated any views to reflect the tuple delete.  If updating the views threw an exception, then the undo action would never get registered, and the tuple delete would not be undone when the transaction failed as a result.  This caused the view and source table to get out of sync and havoc resulted.

BillWhite:
Ok, now I think I understand.  I still approve, though.

PR #4355, opened 2017-03-13
ENG-12132: Fix junit test so that it passes in memcheck build

wolffcm:


wolffcm:
Your understanding is not quite right.  There's logic in PersistentTable that decides whether to do a row-by-row delete when a truncate is requested, or whether to replace the PersistentTable with a new, empty table instead---call this the "switcharoo" method.  In the latter case, we do not hit up against the view column size limit, since we just delete all the rows in any views referencing the table being truncated.

The logic for how to perform the truncate depends on the number of tuple blocks allocated for the table.  If it's more than one, then we don't delete row-by-row, we do the switcharoo method.  Since memcheck mode makes it so there will be only one tuple per block, we always to the switcharoo method of truncation in memcheck mode if there is more than one row.

wolffcm:
And yes, you are correct that if we could turn off view maintenance in this case, it would be possible to do the truncate for the non-memcheck case.

PR #4356, opened 2017-03-13
Make it possible to format TimestampType independent of time zone.

BillWhite:
The test TestVoltType.testTimestampToStringBeforeEpoch failed if it was
run on a machine whose timezone was not America/New_York.  This change
allows the user to choose a SimpleDateFormatter for the toString method,
which lets the test be time zone independent.

This is not actually the most general one could do, but it's better than
before.

### DIFF
--- a/src/frontend/org/voltdb/types/TimestampType.java
+++ b/src/frontend/org/voltdb/types/TimestampType.java
@@ -19,6 +19,7 @@ package org.voltdb.types;
 
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 
 import org.json_voltpatches.JSONString;
 import org.voltdb.common.Constants;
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/types/TimestampType.java
```diff
@@ -143,11 +143,28 @@ public boolean equals(Object o) {
     }
 
     /**
-     * toString for debugging and printing VoltTables
+     * An implementation of toString for debugging and printing VoltTables.  The string
+     * is formatted using the standard ODBC date string format, with
```
wolffcm: Would be good to say something about time zone here.

src/frontend/org/voltdb/types/TimestampType.java
```diff
@@ -143,11 +143,28 @@ public boolean equals(Object o) {
...
+        return toString(sdf);
+    }
+
+    /**
+     * An implementation of toString which allows specification of the
+     * format of the Date part of the string.  The number of microseconds
+     * in the TimestampType which are not counted for in the Date part
+     * of the string are appended to this string as three digit,
+     * zero filled integer.  If the number of milliseconds ends the
+     * format, then the result string will have microsecond accuracy.
```
wolffcm: So if someone passes a format that does not end with millis, the output of this method would be not very sensible... is there a way to check if format ends with millis?  

Since this is a user-facing class, it'd be nice if there was a solution that either produces sensible output or throws an exception.

wolffcm: Throwing an exception doesn't seem right to me either after some consideration... you could also just decline to append micros if format doesn't end with millis.

BillWhite: Ok.  I was not 100% happy with that, but (1) it seemed easiest and (2) It's what we currently do.  I think the better choices are (a) pass in a way to combine the Date part of the timestamp, which only has millis, and the microsecond part, which only has the microsecond remainder, and (b) my preferred solution, which is to pass in a lambda which does the formatting.

BillWhite:
Jenkins is all blue here.

PR #4357, opened 2017-03-14
ENG-12139: Fix internal adapter initialization error.

nshi:
Rejoining node won't get notified of partition master promotions. The
internal adapters have to be created during initialization. They are now
created in the constructor of ClientInterface and later can be added if
not exist during master promotion.


akhanzode:
r=akhanzode if Jenkins is happy merge.

PR #4358, opened 2017-03-14
ENG-12144: Fix memory leak on @UAC in export.

nshi:
On @UAC or during shutdown, some of the export decoder executor services
may be shutdown when the buffer is polled. Setting the buffer on the
future will throw RejectedExecutionException in this case because the
listener tries to schedule itself on those executor services. It has to
catch the exception and properly discard the containers that was just
created to avoid leaking memory.


akhanzode:
r=akhanzode if Jenkins is happy merge.

PR #4359, opened 2017-03-14
Eng 12133 skip loading hsqldb ctx

xinjiacs:
1. Skip default procedures building work for update classes
2. Skip hsqldb reloading work for update classes
3. Remove unnecessary cluster/database objects and code cleanup

### DIFF
--- a/src/frontend/org/voltdb/CatalogContext.java
+++ b/src/frontend/org/voltdb/CatalogContext.java
@@ -107,7 +107,10 @@ public class CatalogContext {
             byte[] catalogBytesHash,
             byte[] deploymentBytes,
             int version,
-            HostMessenger messenger)
+            HostMessenger messenger,
+            boolean hasSchemaChange,
+            DefaultProcedureManager defaultProcManager,
+            PlannerTool plannerTool)
     {
         m_transactionId = transactionId;
         m_uniqueId = uniqueId;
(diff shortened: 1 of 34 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -86,8 +83,9 @@ public PlannerTool(final Cluster cluster, final Database database, byte[] catalo
                 throw new RuntimeException("Error creating hsql: " + e.getMessage() + " in DDL statement: " + decoded_cmd);
             }
         }
-
-        hostLog.debug("hsql loaded");
+        if (hostLog.isDebugEnabled()) {
+            hostLog.debug("hsql loaded");
```
jhugg: Why this change. Smells of premature optimization.

src/frontend/org/voltdb/ReplicaInvocationAcceptancePolicy.java
```diff
@@ -49,13 +49,13 @@ private ClientResponseImpl shouldAcceptHelper(
...
         if (isReadOnly) {
             return null;
-        } else {
-            return new ClientResponseImpl(ClientResponseImpl.UNEXPECTED_FAILURE,
-                    new VoltTable[0],
-                    "Write procedure " + invocation.getProcName() +
-                    " is not allowed in replica cluster",
-                    invocation.clientHandle);
         }
+
```
jhugg: I wouldn't typically make style changes like this to files if you're not changing anything else about the code. It makes the diff bigger, and it's not one of our key style rules. Lots of people prefer if/else over if-return/stmt.

src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -159,10 +162,16 @@ public CatalogContext(
         this.deploymentHashForConfig = CatalogUtil.makeDeploymentHashForConfig(deploymentBytes);
         m_memoizedDeployment = null;
 
-        m_defaultProcs = new DefaultProcedureManager(database);
+        if (hasSchemaChange) {
```
jhugg: Maybe a comment? This is the meat of the change, no?

src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -176,6 +185,21 @@ public CatalogContext(
         }
     }
 
+    public CatalogContext(
```
jhugg: Feels like this constructors should have comments explaining why they exist and when to pick each.

PR #4360, opened 2017-03-15
ENG-12147: Fix race in internal adapter creation.

nshi:
Create the per-partition adapters before creating the ClientInterface
mailbox. Once the mailbox is created, the master promotion notification
may be called before the ClientInterface is fully constructed. What a
mess!


akhanzode:
Looks good.

PR #4361, opened 2017-03-15
ENG-12128:

wweiss-voltdb:
The Consumer was not started on rejoined hosts because they were only initialized on create and recover. In addition, as part of initialization, the DR code determines whether this cluster is the original data source, and if not, it verifies that there are valid site trackers for the data source cluster (implying that the snapshot completed). However, for rejoin, this check would be done on the Site thread causing a deadlock. Therefore, Starting the DR services is now always done on a seperate executor thread.

Note that this is the periodic work thread, which will block until the site trackers are returned. Therefore, scheduled work will also block until DR initialization completes.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -3959,27 +3959,36 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
     }
 
     private void prepareReplication() {
-        try {
-            boolean okToStartDR = true;
-            if (m_consumerDRGateway != null) {
-                if (m_config.m_startAction == StartAction.RECOVER) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3959,27 +3959,36 @@ private void initializeDRProducer() {
...
+        Runnable t = new Runnable()
+        {
+            @Override
+            public void run()
+            {
+                // Warning: This is called on the site thread if this host is rejoining
+                try {
+                    boolean okToStartDR = true;
+                    if (m_consumerDRGateway != null) {
+                        if (m_config.m_startAction != StartAction.CREATE) {
```
manjujames: Should this use 'm_config.m_startAction.doesRecover() || m_config.m_startAction.doesRejoin()' instead?

PR #4363, opened 2017-03-15
Aggregate functions and window functions together are not supported.

BillWhite:
We never intended to support group by with window functions.  Aggregate
functions are a kind of implicit group by, so they are not supported
either.

https://issues.voltdb.com/browse/ENG-12146

### DIFF
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -938,7 +938,10 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
                 throw new PlanningErrorException(
                         "Only one windowed function call may appear in a selection list.");
             }
-
+            if (m_hasAggregateExpression) {
+                throw new PlanningErrorException(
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -938,7 +938,10 @@ private void verifyWindowFunctionExpressions() {
                 throw new PlanningErrorException(
                         "Only one windowed function call may appear in a selection list.");
             }
-
+            if (m_hasAggregateExpression) {
+                throw new PlanningErrorException(
+                        "Use of both a windowed function call and aggregate operators in a single query is not supported.");
```
wolffcm: I'm not sure users will make a distinction between aggregation functions and windowed functions, since, say, MAX or MIN are supported on both contexts.  Andrew might have better feedback, but here's my stab at a message: "Use of analytic windowed functions (in an OVER clause) isn't supported with other aggregate functions on the SELECT list."

BillWhite: That's better.  I'll change it.

PR #4365, opened 2017-03-15
ENG-12156: Skip generation roll when catalog update does not involve EE

akhanzode:


xinjiacs:
In future, we want to move getDiffCommandsForEE() into one, always keeping them consistent.

src/frontend/org/voltdb/export/ExportManager.java
```diff
@@ -622,22 +623,26 @@ public int getConnCount() {
...
 
         updateProcessorConfig(connectors);
-        if (m_processorConfig.size() == 0) {
+        if (m_processorConfig.isEmpty()) {
             m_lastNonEnabledGeneration = catalogContext.m_uniqueId;
             return;
         }
 
+        if (CatalogUtil.getDiffCommandsForEE(diffCommands).length() == 0) {
+            // empty diff cmds means java specific catalog changes EE not touched.
```
xinjiacs: Maybe add another comment: EE catalog update will update Stream table's generation transaction id. And Java needs to sync Stream table with "catalogContext.m_uniqueId" as well.  @akhanzode

PR #4369, opened 2017-03-16
Disable subqueries in DML.

BillWhite:
The use of subqueries in DML needs to be more extensively tested before
we can release it.  This commit adds a guard to disable this use, but
leaves the code in place.  There is a final static boolean defined in
ParsedUnionStmt which can be used to find all the places that need to be
examined when the testing is done.

https://issues.voltdb.com/browse/ENG-12161

### DIFF
--- a/tests/frontend/org/voltdb/planner/TestPlansSubQueries.java
+++ b/tests/frontend/org/voltdb/planner/TestPlansSubQueries.java
@@ -65,6 +65,31 @@ public class TestPlansSubQueries extends PlannerTestCase {
         setupSchema(TestPlansSubQueries.class.getResource("testplans-subqueries-ddl.sql"), "ddl", false);
     }
 
+    public void testSelectOnlyGuard() {
+        if ( ! ParsedUnionStmt.ENG6281_FIXED) {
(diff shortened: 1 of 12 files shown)


tests/frontend/org/voltdb/planner/TestPlansSubQueries.java
```diff
@@ -65,6 +65,31 @@ protected void setUp() throws Exception {
         setupSchema(TestPlansSubQueries.class.getResource("testplans-subqueries-ddl.sql"), "ddl", false);
     }
 
+    public void testSelectOnlyGuard() {
+        if ( ! ParsedUnionStmt.ENG6281_FIXED) {
```
BillWhite: Note that this is executed when ParsedUnionStmt.ENG6281_FIXED is **_FALSE_**.  That is to say, before ENG-6281 is fixed.  When ENG-6281 is fixed, along with the other testing tickets, this can just be eliminated entirely.  Until then, this is not a DEFERREDtest, but should remain active.

BillWhite:
Abandoning this and just reverting the commits.

PR #4373, opened 2017-03-17
ENG-12175: bypass import ssl if python don't support it

JadenW:


lib/python/voltdbclient.py
```diff
@@ -29,6 +29,14 @@
 except ImportError:
     from sha import sha
 
+try:
+    import ssl
+    ssl_available = True
+except ImportError:
+    print "Cannot import ssl module, operate in non-ssl mode"
```
akhanzode: Can you get correct message from Andrew or Ruth. Rest of the changes look good. Push when message is updated.

lib/python/voltdbclient.py
```diff
@@ -176,7 +187,7 @@ def __init__(self, host = None, port = 21212, usessl = False, username = "",
         self.socket = None
         if self.host != None and self.port != None:
             ss = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            if (self.usessl):
+            if (self.usessl and ssl_available):
```
akhanzode: Can you add message here if someone says usessl and ssl_available is false and also bail out with ERROR

lib/python/voltdbclient.py
akhanzode: INFO: Cannot import ssl module.  To connect to a cluster with SSL enabled, install the Python ssl module.

lib/python/voltdbclient.py
akhanzode: Something like:
ERROR: To use SSL functionality please Install the Python ssl module.

akhanzode:
To make this better only print message if ssl is not available if someone says usessl. For non-ssl folks even if ssl module is not present dont print any message it will be annoying message after few run.

PR #4374, opened 2017-03-17
sql-grammar.txt: 3 major changes:

johnTcrawford:
added Window functions (RANK, etc.), and associated syntax; 
expanded the use of aggregate functions, used as Window functions and
otherwise; 
expanded the use of sub-queries, scalar and otherwise.
And multiple minor changes:
added very occasional use of nonexistent table, view, and column names; 
added symbols for byte columns and expressions, and use them as args for
the REPEAT function, to avoid ENG-12118; 
improved UPDATE statements, to get the column types right most of the
time; 
expanded boolean expressions, to include multiple AND or OR, optionally
using parentheses; 
fixed USING clauses, by adding parentheses; 
added EXISTS and IN clauses; 
removed the nonexistent % operator (SQL has the MOD function, but no
remainder op); 
removed the use of CONCAT with only one argument, which does not work; 
fixed use of the TRIM function, by adding single-quotes and fixing an
unmatched '['; 
added the beginnings of varbinary, point, and polygon expressions
(though more work is needed in this area); 
and other minor tweaks, e.g. to likelihoods of different grammar
options.

sql_grammar_generator.py: 
added the ability to specify SQL statements containing multiple phrases
(e.g., 'INSERT' and 'SELECT') to be echoed to a file (with their
results), and optionally, with the grammar symbols that were used to
create them, and the ability to count the number and percent of such
statements that were valid versus invalid (all of which helps to debug
changes to the grammar file);
added (to the summary file) counts of the number of certain Java
Exceptions (e.g. NPE's) found in the VoltDB log or console output; 
and tweaked various comments and debug print.

### DIFF
--- a/tests/sqlgrammar/sql-grammar.txt
+++ b/tests/sqlgrammar/sql-grammar.txt
@@ -315,9 +315,11 @@ window-function         ::= {window-agg-func-expr} 4| COUNT(*) | RANK() | DENSE_
 window-agg-func-expr    ::= {non-num-aggregate-func}({column-expression}) 3| \
                             {num-only-aggregate-func}({numeric-expression})
 
-# It is possible for this to be empty, but that is less interesting, so we make it less likely
-partition-or-order-by   ::=  PARTITION BY {partition-by-list}   ORDER BY {window-order-by-list} 2| \
-                            [PARTITION BY {partition-by-list}]  ORDER BY {window-order-by-list}  | \
+# This is somewhat redundant, but since some window functions (RANK, DENSE_RANK)
+# require an ORDER BY, we make that more likely; and while other window functions
+# do allow this to be empty, that is less interesting, even when legal, so we
+# make it less likely
+partition-or-order-by   ::= [PARTITION BY {partition-by-list}]  ORDER BY {window-order-by-list} | \
                             [PARTITION BY {partition-by-list}] [ORDER BY {window-order-by-list}]
 partition-by-list       ::= {column-expression} [[, {partition-by-list}]]
 
--- a/tests/sqlgrammar/sql-grammar.txt
+++ b/tests/sqlgrammar/sql-grammar.txt
@@ -47,13 +47,18 @@
 
 ################################################################################
 # Table and view names pre-defined in the DDL associated with this test
-table-name              ::= P1 | R1 | P2 | R2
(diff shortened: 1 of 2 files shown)


tests/sqlgrammar/sql-grammar.txt
```diff
@@ -226,46 +239,105 @@ table-references        ::= {table-reference} [[, {table-references}]]
...
+################################################################################
+# Window functions - a.k.a. analytic functions
+#
+window-function-expr    ::= {window-function} OVER ({partition-or-order-by})
+window-function         ::= {window-agg-func-expr} 4| COUNT(*) | RANK() | DENSE_RANK()
+
+# SUM can only take numeric argument; others can take any type;
+# (AVG not actually supported here, but it should not crash)
+window-agg-func-expr    ::= {non-num-aggregate-func}({column-expression}) 3| \
+                            {num-only-aggregate-func}({numeric-expression})
```
BillWhite: Maybe add functions of column expressions?  Probably a nit, though, and easy enough to add later.

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -226,46 +239,105 @@ table-references        ::= {table-reference} [[, {table-references}]]
...
+
+# It is possible for this to be empty, but that is less interesting, so we make it less likely
+partition-or-order-by   ::=  PARTITION BY {partition-by-list}   ORDER BY {window-order-by-list} 2| \
```
BillWhite: Do you need both the second and third alternative?  Doesn't the third alternative cover the second?

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: The definition of "column-expression" actually does include functions of columns (among other things) - which I already knew, but before replying to you, I had to search down several levels, through multiple definitions, to confirm that. So, very non-obvious.

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: The third alternative actually covers both the first and second; but (as the comment sort of implies) I was trying to make valid statements more likely, and for some functions, ORDER BY is required, IIRC. I'll tweak this a bit.

PR #4375, opened 2017-03-17
ENG-12094: Add log message when restarting database with no command log.

nshi:
If there is no command log or snapshot data to recover from when
restarting the database, print an INFO level message to the log file and
the console indicating that the database will start empty.


nshi:
@ajgent, I didn't include you as a reviewer by accident :-) Could you check to see if the message make sense? Please let me know if you want me to tweak it. Thanks.

PR #4377, opened 2017-03-20
Eng 10957 ssl example

brians127:
Adds a tutorial for using SSL with VoltDB.


jhugg:
Why isn't this just in /examples/HOWTOs? Why does it need it's own code? Why not just explain how to modify the current examples to use SSL?

brians127:
@jhugg I think it's better to show the user what they need to do via an example that "just works" without any script or command line modification. Prose documentation doesn't work as well, and mainly should be there to provide concepts and context. @akhanzode You're welcome to ask Anish for his opinion as well - he set the requirements in the ticket and our stand-up meetings.

ajgent:
Well, I agree with both, sort of.  I think having an example "that works" is good. But having an explanationof the moving parts would be extremely beneficial.  Also, /doc/tutorials is kind of a graveyard where they've moved examples they don't want cluttering up the examples folder but still need because they might be referenced. (e.g. the C++ client doc assumes the helloworld example is there.)

So I'd suggets an example in examples and a description in Howto. Or put both in Howto.

examples/HOWTOs/enable-ssl/README.md
```diff
@@ -0,0 +1,74 @@
+# Demo: How to Enable SSL
+
+This example demonstrates how to configure and run VoltDB while using SSL to secure connections to the database.
+
+VoltDB servers using SSL must be provided with a private key and a certificate which establishes trust for that key. Clients, including 'sqlcmd', are provided a one or more certificates that establishes that the server can be trusted.
+
```
pdshaw: Fix English in "are provided a one or more certificates that establishes that the server can be trusted."

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+     * declaratively state command line options with defaults
+     * and validation.
+     */
+    static class Parameters extends CLIConfig {
+        @Option(desc = "Comma separated list of the form server[:port] to connect to.")
+        String servers = "localhost";
+
+        @Option(desc = "SSL configuration file.")
+        String sslFile = "";
+
```
pdshaw: Readme (and elsewhere too) calls it  properties. Use sslProps and "SSL properties file" to be consistent?

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+            VoltTable[] results = response.getResults();
+            final int nameIndex = results[0].getColumnIndex("name");
+
+            System.out.println("\nNames in the database: (limit 10)");
+            while (results[0].advanceRow()){
+                System.out.println(results[0].get(nameIndex, VoltType.STRING));
+            }
+
+            System.out.println("Success!");
```
pdshaw: drain and close client connection? not sure if the examples are consistent but good practice.

examples/HOWTOs/enable-ssl/README.md
```diff
@@ -0,0 +1,74 @@
...
+    ./run.sh server
+
+Wait until you see "Server completed initialization."
+Open a new shell in the same directory and run the following to load the schema:
+
+    ./run.sh init
+
+In the same shell, run the following script several times. This will run the demo client application:
+
+    ./run.sh client
```
akhanzode: I would rather see the minimal java code that we use which enables SSL. Would be like 5 lines of code with comments. running this demo client is less important.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package simplessl;
+
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
+import java.util.StringTokenizer;
+
+import javax.crypto.Cipher;
```
akhanzode: Ignore high grade encryption.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+            if (sslFile.equals("")){
+                exitWithMessageAndUsage("must provide SSL configuration file");
+            }
+        }
+    }
+
+    /** Used to warn the user if strong cryptography is not installed.
+     * @return Maximum allowed key length
+     */
+    public static int getMaxKeyLength() {
```
akhanzode: Please remove this...make is very very simple.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            return Cipher.getMaxAllowedKeyLength("AES/CBC/PKCS5Padding");
+        } catch (final NoSuchAlgorithmException e) {
+            throw new RuntimeException("AES/CBC/PKCS5Padding is not available, despite it being mandatory for Java implementations!", e);
+        }
+    }
+
+    public static void main(String[] args)
+    {
+        final int maxCryptoKeyLength = getMaxKeyLength();
```
akhanzode: remove this.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+
+        Parameters config = new Parameters();
+        config.parse(Client.class.getName(), args);
+
+        System.out.println("Connecting to VoltDB");
+        ClientResponse response;
+        try {
+            ClientConfig clientConfig = new ClientConfig();
+            clientConfig.setTrustStoreConfigFromPropertyFile(config.sslFile);
+            clientConfig.enableSSL();
```
akhanzode: Do you need to call this or createClient will take care of doing the right thing?

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+
+import org.voltdb.CLIConfig;
+import org.voltdb.VoltTable;
+import org.voltdb.VoltType;
+import org.voltdb.client.*;
+
+
+public class ExampleClient
+{
+    static final String[] firstNames = { "Adam", "Bob", "Clara", "Denise", "Edgar", "Flores", "Gregory" };
```
akhanzode: You dot need all names and generateNames() this should be like a hello ssl

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+            clientConfig.setTrustStoreConfigFromPropertyFile(config.sslFile);
+            clientConfig.enableSSL();
+
+            final Client client = ClientFactory.createClient(clientConfig);
+            StringTokenizer serverParser = new StringTokenizer(config.servers, ",");
+            while (serverParser.hasMoreTokens()){
+                client.createConnection(serverParser.nextToken(), Client.VOLTDB_SERVER_PORT);
+            }
+
+            String person = generateName();
```
akhanzode: Remove this. remember hello ssl.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
```diff
@@ -0,0 +1,141 @@
...
+            try {
+                response = client.callProcedure("@AdHoc", "INSERT INTO people VALUES '" + person + "';");
```
akhanzode: Hardcode person as 'Hello SSL' and do a select and print and if people see 'Hello SSL' ssl worked.

examples/HOWTOs/enable-ssl/ddl.sql
```diff
@@ -0,0 +1,6 @@
+DROP TABLE people IF EXISTS;
+
+CREATE TABLE people
```
akhanzode: call this message table where message saved is 'Hello SSL'

examples/HOWTOs/enable-ssl/deployment.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<!--
```
akhanzode: The comments does not match deployment. Either remove comments or fix them to match.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
brians127: Voter SimpleBenchmark does not do this, and that's what I copied from. I'm adding these to my example since you're right; other examples consistently drain and close.

brians127:
@jhugg verbally expressed concern that the example itself offers no value beyond a purely text HOWTO. I am adding a section to the README that addresses someone converting an existing application to use SSL, but want to hear from @ajgent before deleting anything.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
brians127: I believe this adds value by helping an unsophisticated user make sure they have SSL properly configured as per the README.

examples/HOWTOs/enable-ssl/client/simplessl/ExampleClient.java
brians127: It's needed. There's a boolean m_enableSSL that only gets set by this method. Possible API oversight?

jhugg:
The #1 goal here should be to convey that adding SSL to an app is easy. The good news is that it _is_ easy. You can make Voter SSL by adding literally 2 lines of code to the client java code, and by using a provided deployment file in /examples/HOWTOs/deployment-file-examples that adds the SSL config. 

The hard part (covered in the readme) is dealing with the trust store. There should also be a note about using CLI tools when the server has enabled SSL. Honestly, we should probably consider making that easier to do in the future.

PR #4380, opened 2017-03-21
ENG-12187: Use sha256 as default.

akhanzode:


lib/python/voltdbclient.py
```diff
@@ -24,11 +24,7 @@
 import struct
 import datetime
 import decimal
-try:
-    from hashlib import sha1 as sha
-except ImportError:
```
nshi: Was this to support Python 2.4 and older?

PR #4387, opened 2017-03-22
Update voltdb collect cli

hhakimi53:
Update voltdb collect cli logic so that options for specifying database directory and output file is consistent with voltdb get logic. VEM and fileonly (used by VEM) options and associated logic has been removed. Newly added and updated messages have been sent to Andrew for review.


src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -108,26 +117,65 @@
...
 
         @Override
         public void validate() {
             if (days < 0) exitWithMessageAndUsage("days must be >= 0");
-            if (voltdbroot == "") exitWithMessageAndUsage("voltdbroot cannot be null");
+            if (voltdbroot.trim().isEmpty()) exitWithMessageAndUsage("Invalid database directory");
+        }
+    }
+
+    private static void populateVoltDBCollectionPaths() {
```
akhanzode: remove static. private static does not make useful.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -108,26 +117,65 @@
...
+            VoltDB.exit(-1);
+        }
+        m_config.voltdbroot = m_voltdbRoot.getAbsolutePath();
+
+        // files to collect from config dir
+        String configLogDirPath = m_voltdbRoot.getAbsolutePath() + File.separator + Constants.CONFIG_DIR + File.separator;
+        m_configInfoPath = configLogDirPath + "config.json";
+        m_catalogJarPath = configLogDirPath + "catalog.jar";
+        m_deploymentPath = configLogDirPath + "deployment.xml";
+        m_systemCheckPath =  m_voltdbRoot.getAbsolutePath() + File.separator + "systemcheck";
```
akhanzode: Lets add path.properties as well to collection.

akhanzode: We should collect logs under voltdbroot/log as well.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -606,9 +587,7 @@ private static void generateCollection(Set<String> paths, boolean copyToVEM) {
                         if (delLocalCopy) {
                             try {
                                 new File(collectionFilePath).delete();
-                                if (!m_config.calledFromVEM) {
-                                    System.out.println("Local copy "  + collectionFilePath + " deleted");
-                                }
+                                System.out.println("Local copy "  + collectionFilePath + " deleted");
```
akhanzode: Lets remove all VEM options. delete of local copy will not be needed then. Basically we will never delete anything local in collect.

src/frontend/org/voltdb/utils/Collector.java
hhakimi53: need to make it static as the current pattern followed in the logic was to not instantiate collector and instead have internal use routines be static. So followed same approach in here so that don't have to redo others.

src/frontend/org/voltdb/utils/Collector.java
hhakimi53: Sure, will add those to the collect.

src/frontend/org/voltdb/utils/Collector.java
hhakimi53: Sure. So host, username, password and associate apis for upload can also be taken out than? Thnxs

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -49,52 +47,46 @@
...
+import org.voltdb.settings.NodeSettings;
 import org.voltdb.types.TimestampType;
 
-import com.google_voltpatches.common.base.Charsets;
 import com.google_voltpatches.common.base.Throwables;
-import com.google_voltpatches.common.net.HostAndPort;
 
 public class Collector {
+    private final static String ZIP_ENTRY_FOLDER_BASE_SUFFIX =  "_volt_collect_logs_";
+    final static String DEFAULT_COLLECT_FILENAME = "collect.zip";
```
akhanzode: collect.zip is too generic check with Ruth what name she wants I would prefer voltdb somewhere in there.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -533,16 +510,38 @@ private static void generateCollection(Set<String> paths, boolean copyToVEM) {
...
-                    PathsType.Exportoverflow exportPath = deploymentPaths.getExportoverflow();
-                    if (exportPath != null)
-                        exportOverflowPath = VoltDB.instance().getExportOverflowPath(exportPath);
+
+                if (m_pathPropertiesPath != null && !m_pathPropertiesPath.trim().isEmpty()
+                        && (new File(m_pathPropertiesPath)).exists()) {
+                    Properties prop = new Properties();
+                    InputStream input = null;
+
+                    try {
```
akhanzode: use try with resources

PR #4404, opened 2017-03-30
ENG-11012: Online Tracing System.

nshi:
Introducing an online tracing system that can be turned on and off at
runtime without significantly impacting the workload. Trace events will
be cached in a ring buffer in memory unless instructed to be written to
file. Since the ring buffer is bounded, old trace events will be removed
if the buffer is filled up. This allows tracing to be turned on without
worrying about running out of heap. The ring buffer size can be tweaked
using the Java system property "VOLTTRACE_QUEUE_SIZE". By default, it is
set to 4096 event batches, each batch can contain several events.

The tracing system can be turned on/off for each of VoltDB's subsystems
individually. Trace events are currently available for the following
subsystems.

- CI: Client Interface.
- MPI: Multi-partition Initiator.
- MPSITE: Multi-partition Site.
- SPI: Single-partition Initiator.
- SPSITE: Single-partition Site.
- EE: Execution Engine. Please note that turning this category on may
  introduce significant performance impact because each statement in
  each fragment batch will result in a JNI call and some new objects
  being created on the heap.
- DRPRODUCER: DR Producer.
- DRCONSUMER: DR Consumer.

The trace points available now are sufficient to diagnose most
transaction related problems such as slow query/transaction, overhead
incurred by command logging (sync or async), or spikes in transaction
latency, etc. More trace points can be added to other parts of the
system.

The output file format is JSON. The JSON content conforms to the Chrome
Trace Event format, so it can easily be loaded into Chrome's tracing
tool for visualization. It is also easy to parse and analyze in
programs.

There is a system procedure, @Trace, to manage the tracing system at
runtime. It has the following parameters.

- @Trace on: Turn on the tracing system if it's not turned on already.
- @Trace off: Turn the tracing system off if it is on.
- @Trace dump: Write all trace events in the ring buffer to file and
  return the file path.
- @Trace enable category: Enable tracing for the given category.
- @Trace disable category: Disable tracing for the given category.

SQLCmd currently recognizes the @Trace sysproc, but VMC does not. It
could also be added to voltadmin for ease of use.

Based on preliminary benchmark results, turning the tracing system on
for all existing categories doesn't add noticeable overhead to
throughput but adds several milliseconds to the latency. If tracing is
turned off, it adds almost zero overhead to the system.

If you are a Volt developer and would like to add some trace events,
here are some guidelines.

1. Get a TraceEventBatch by calling VoltTrace.log(category) with the
   appropriate category listed above. If the category you have in mind
   doesn't exist yet, add it to the Category enum first. If a non-null
   TraceEventBatch is returned, that means the trace event logging for
   this category is enabled. If the returned value is null, either trace
   event logging is completely disabled or this category is
   disabled. Skip logging in this case.

2. Always wrap your trace event logging code with a null check of the
   returned TraceEventBatch object for both safety and performance
   reasons.

3. The returned TraceEventBatch object can contain multiple trace
   events. Use the same batch as long as you are still in the same
   thread logging for the same category. Call TraceEventBatch.add() to
   add events. The method takes a Supplier, which can easily be created
   using Java 8's lambda expression. The reason for using a Supplier
   here is to delay potentially expensive materialization of the
   parameters on performance sensitive paths. Materialization will be
   delayed until they are written to file, or they will never
   materialize if they are never written to file. For example, to create
   an instant event that captures some interesting variables, you can do
   the following,

       batch.add(() -> VoltTrace.instant("myInstantEvent",
                                         "argName", "argValue"));

4. Always add some meta events to record a human friendly process name
   and a thread name if the event you are logging could be the first
   event in this VoltDB process, e.g. when a stored procedure invocation
   request first arrives at a server node. This helps visualization
   later.

5. For the async events, async begin/end and async instant, you have to
   provide a unique ID to tie them together. The ID has to be globally
   unique across threads and processes. You can use string as ID. For
   example, in order to track transaction messages across the cluster, I
   use (sender HSID, receiver HSID, transaction ID) as the identifier
   for the corresponding async events. This ID has to be consistent for
   the events in the series, or they will not be linked together in the
   visualization.

### DIFF
--- a/src/frontend/org/voltdb/utils/TraceFileWriter.java
+++ b/src/frontend/org/voltdb/utils/TraceFileWriter.java
@@ -1,0 +1,105 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2017 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.voltdb.utils;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.util.Queue;
+import java.util.zip.GZIPOutputStream;
+
+import org.codehaus.jackson.map.ObjectMapper;
+import org.voltcore.logging.VoltLogger;
+
+
+/**
+ * Reads trace events from VoltTrace queue and writes them to files.
+ */
+public class TraceFileWriter implements Runnable {
+    private static final VoltLogger s_logger = new VoltLogger("TRACER");
+
+    private final File m_path;
+    private final Queue<VoltTrace.TraceEventBatch> m_events;
+
+    public TraceFileWriter(File path, Queue<VoltTrace.TraceEventBatch> events) {
+        m_events = events;
+        m_path = path;
+    }
+
+    @Override
+    public void run() {
+        final ObjectMapper jsonMapper = new ObjectMapper();
+        BufferedWriter fileWriter = null;
+        long firstEventTime = 0;
+        long count = 0;
+
+        try {
+            VoltTrace.TraceEventBatch eventSupplier;
+            while ((eventSupplier = m_events.poll()) != null) {
+                VoltTrace.TraceEvent event;
+                while ((event = eventSupplier.nextEvent()) != null) {
+                    if (fileWriter == null) {
+                        fileWriter = startTraceFile(m_path);
+                        firstEventTime = event.getNanos();
+                    } else {
+                        fileWriter.write(",");
+                    }
+
+                    event.setSyncNanos(firstEventTime);
+                    String json = jsonMapper.writeValueAsString(event);
+                    fileWriter.newLine();
+                    fileWriter.write(json);
+                    fileWriter.flush();
+
+                    count++;
+                }
+            }
+        } catch(IOException e) { // also catches JSON exceptions
+            s_logger.info("Unexpected IO exception in trace file writer. Stopping trace file writer.", e);
+        }
+
+        if (fileWriter != null) {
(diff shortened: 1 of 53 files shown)


src/frontend/org/voltdb/utils/TraceFileWriter.java
```diff
@@ -0,0 +1,127 @@
...
+ * Reads trace events from VoltTrace queue and writes them to files.
+ */
+public class TraceFileWriter implements Runnable {
+    private static final VoltLogger s_logger = new VoltLogger("TRACER");
+
+    private final ObjectMapper m_jsonMapper = new ObjectMapper();
+    private final File m_path;
+    private final Queue<VoltTrace.TraceEventBatch> m_events;
+    private BufferedWriter m_fileWriter = null;
+    private Long m_firstEventTime = null;
```
xinjiacs: `m_jsonMapper , m_fileWriter , m_firstEventTime ` all can be local variables in `run()` function.

src/frontend/org/voltdb/utils/TraceFileWriter.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        close();
+
+        s_logger.info("Wrote " + count + " trace events to " + m_path.getAbsolutePath());
+    }
+
+    private void close() {
+        BufferedWriter bw = m_fileWriter;
+        if (bw==null) return;
```
xinjiacs: Is it more readable with `if (m_fileWriter == null) return;`? or even at the caller 
`if (m_fileWriter != null) {
    close();
}`

src/frontend/org/voltdb/utils/VoltTrace.java
```diff
@@ -0,0 +1,621 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2016 VoltDB Inc.
```
xinjiacs: Same bad license issue?

manjujames:
Looks great to me! I didn't look at all the trace points, but looked at the other classes. Just a couple of questions:
- Why did you choose to use the stats framework to enable/disable tracing?
- If a second dump command comes in while the first one is still being processed, will it return a dump file path that does not exist?

nshi:
@manjujames For the first question, I picked the OpsAgent framework because it's an established mechanism to distribute non-transactional work to all nodes easily. I don't want to write a sysproc because I don't need the transacationality.

Good point in the second question, that could happen. In fact, I have to add a null pointer check there. Thanks.

nshi:
@manjujames @xinjiacs Thanks for the review. I addressed all of your comments and also rebased the branch.

PR #4405, opened 2017-03-31
ENG-11486  update clients upon stored procedure changes

zeemanhe:
Server will notify clients when stored procedures are added, deleted or updated. Client will then respond to the changes  to ensure client affinity.


src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -644,6 +644,7 @@ private final ClientResponseImpl dispatchSubscribe(InvocationClientHandler handl
 
             if ("TOPOLOGY".equals(param)) {
                 cihm.setWantsTopologyUpdates(true);
+                cihm.setWantsProcUpdates(true);
```
xinjiacs: Should "ProcedureUpdates" always be on when "TopologyUpdates" affinity is on? Is there any other case they may not be consistent like " false, true" or "true, false"? @zeemanhe 

If not, why there is a need to introduce a new option?

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -117,6 +117,7 @@
 
     //Same as in Distributer.java
     public static final long ASYNC_TOPO_HANDLE = Long.MAX_VALUE - 1;
+    public static final long ASYNC_PROC_HANDLE = Long.MAX_VALUE - 2;
```
xinjiacs: Maybe add a comment:
Client's cached Procedure information will be updated upon topology changes or any UAC called.

PR #4410, opened 2017-04-03
Update message severity level during validation of start action

hhakimi53:
During validation of start action, there is a rare window where the node can be disconnected in time frame where it was reported it exists to when it queried for data and currently this gets reported at error level. This showed up in rejoin system tests once where nodes were getting killed in cluster. After logging it at error level, the cluster was operational though test failed due to msg seen at error level. Dropping the severity for the msg at which it gets logged for this particular case. Updated logic to have try/catch block around the accessors which generate the handled exception cases.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -75,6 +75,7 @@ import org.apache.log4j.FileAppender;
 import org.apache.log4j.Logger;
 import org.apache.zookeeper_voltpatches.CreateMode;
 import org.apache.zookeeper_voltpatches.KeeperException;
+import org.apache.zookeeper_voltpatches.KeeperException.Code;
 import org.apache.zookeeper_voltpatches.WatchedEvent;
 import org.apache.zookeeper_voltpatches.Watcher;
 import org.apache.zookeeper_voltpatches.ZooDefs.Ids;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1711,39 +1712,59 @@ public void run() {
...
-                                    + "Join and rejoin are for adding nodes to an already running cluster.");
-                        }
-                    }
-                }
-            }
+            initCompleted = zk.exists(VoltZK.init_completed, false) != null;
+            children = zk.getChildren(VoltZK.start_action, new StartActionWatcher(), null);
         } catch (KeeperException e) {
             hostLog.error("Failed to validate the start actions", e);
+            return;
```
akhanzode: This needs to be debug log right? Is this the exception that got fired? I thought it was from watcher?

hhakimi53: The exception that got fired was NoNodeException (node disappeared), happened on ZK::getdata() (#1733). This was earlier getting handled under the base KeeperException catch block. Currently have separated them out and the msg from  NoNodeException is getting logged only in debugged mode for that specific case only.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1711,39 +1712,59 @@ public void run() {
...
+
+        if (children != null && !children.isEmpty()) {
+            for (String child : children) {
+                byte[] data = null;
+                try {
+                    data = zk.getData(VoltZK.start_action + "/" + child, false, null);
+                } catch (KeeperException excp) {
+                    if (excp.code() == Code.NONODE) {
+                        if (hostLog.isDebugEnabled())
+                            hostLog.warn("Failed to validate the start action as node "
```
akhanzode: Change this to debug and squash merge.

PR #4413, opened 2017-04-04
ENG-12008: DR consumer streams should pause on swap table event and unpause on local swap table callback

slowsoul:
* Use a set of Pair<String, Long> instead of two parameters to make equality check more straightforward and code clearer.
* Add a variant of createLocalCluster() helper function to allow manual specification of callingMethodName, which is used in the output file name.
* Add a helper function to create a client that connects to the admin port of the cluster, which is used to perform some tasks when the cluster is paused.

### DIFF
--- a/src/frontend/org/voltdb/ConsumerDRGateway.java
+++ b/src/frontend/org/voltdb/ConsumerDRGateway.java
@@ -19,6 +19,7 @@ package org.voltdb;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
 import org.voltcore.utils.Pair;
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/ConsumerDRGateway.java
```diff
@@ -36,7 +37,7 @@
      */
     void updateCatalog(CatalogContext catalog, String newConnectionSource);
 
-    void swapTables(final Pair<String, Long> oneTable, final Pair<String, Long> otherTable);
+    void swapTables(final Set<Pair<String, Long>> swappedTables);
```
manjujames: Why the change to Set?

PR #4415, opened 2017-04-05
Eng 10240 grammar sqlcov1

johnTcrawford:
Innovation Week improvements, part 1 of getting SQL statements generated by the SQL-grammar-generator to run in SqlCoverage (against PostgreSQL/PostGIS). Includes moving to python2.7 (from 2.6).


src/frontend/org/voltdb/PostgreSQLBackend.java
```diff
@@ -506,7 +653,7 @@ static private int indexOfNthOccurrenceOfCharIn(String str, char ch, int n) {
      *  in which case both the <i>prefix</i> and the <i>suffix</i> need to be
      *  placed so as to not cause mismatched parentheses. */
     @Override
-    protected String handleParens(String group, String prefix, String suffix) {
+    protected String handleParens(String group, String prefix, String suffix, boolean debugPrint) {
```
wolffcm: Every time I look at this code I wonder, "hmm, at some point it might be good to use a grammar to do the transformation between SQL dialects.  I wonder if John has put any thought into that."  I know I bring it up every time.  

Something like that would work by producing an abstract syntax tree using a parser generated from tools like JavaCC and JJTree (or maybe ANTLR) and translating to the target SQL dialect by traversing the tree.  Ultimately this might be much easier to maintain.  If you're okay with how things are now I don't want to bug you about it too much though.

Even just using a tokenizer of some kind might let you avoid having to use stuff like "has odd number of single quotes."

If you ever want to go this route I'd be happy to consult.

tests/sqlcoverage/ddl/sql-grammar-gen-DDL.sql
```diff
@@ -0,0 +1,391 @@
+-- Define "standard" DDL (tables, views, indexes), for use with the SQL grammar generator
+
+-- Drop all items first, in case they already exist
+DROP TABLE R0 IF EXISTS CASCADE;
```
wolffcm: You could just point sqlcoverage at the grammar generator's copy of the DDL.  That way they can't get out of sync.

src/frontend/org/voltdb/PostgreSQLBackend.java
johnTcrawford: As you suggest, I don't want to get into this right now, but I have created ENG-12283 to consider doing this as some point, and I quoted the above review comment. Might make a good Innovation Week project??

tests/sqlcoverage/ddl/sql-grammar-gen-DDL.sql
johnTcrawford: Although github does not make it clear, this "new" file actually is a symbolic link to tests/sqlgrammar/DDL.sql.

PR #4416, opened 2017-04-06
ENG-12105: Fix memory issues in C++ unit tests

wolffcm:
This makes the EE more robust if we happen to get an invalid plan from Java.  If a plan node destructor fired before a plan node is completely initialized (due to wonky JSON), then we were trying to call delete on uninitialized pointers.

I'll make another pull request to fix the bogus plan.

### DIFF
--- a/src/ee/plannodes/abstractplannode.cpp
+++ b/src/ee/plannodes/abstractplannode.cpp
@@ -58,9 +58,18 @@ using namespace std;
 namespace voltdb {
 
 AbstractPlanNode::AbstractPlanNode()
-    : m_planNodeId(-1),
-      m_isInline(false),
-      m_validOutputColumnCount(0) { }
+    : m_planNodeId(-1)
+    , m_children()
+    , m_childIds()
+    , m_executor()
+    , m_inlineNodes()
+    , m_isInline(false)
+    , m_outputTable()
+    , m_inputTables()
+    , m_validOutputColumnCount(0)
+    , m_outputSchema()
+{
+}
 
 AbstractPlanNode::~AbstractPlanNode()
 {
--- a/src/ee/plannodes/limitnode.h
+++ b/src/ee/plannodes/limitnode.h
@@ -65,7 +65,7 @@ public:
         , offset(0)
         , limitParamIdx(-1)
         , offsetParamIdx(-1)
-        , limitExpression(nullptr)
+        , limitExpression(NULL)
     {
     }
 
--- a/src/ee/plannodes/limitnode.h
+++ b/src/ee/plannodes/limitnode.h
@@ -60,8 +60,15 @@ class Table;
  */
 class LimitPlanNode : public AbstractPlanNode {
 public:
-    LimitPlanNode() : limit(-1), offset(0), limitParamIdx(-1), offsetParamIdx(-1)
-    {}
+    LimitPlanNode()
+        : limit(-1)
+        , offset(0)
+        , limitParamIdx(-1)
+        , offsetParamIdx(-1)
+        , limitExpression(nullptr)
+    {
+    }
+
     ~LimitPlanNode();
     PlanNodeType getPlanNodeType() const;
 
--- a/tests/ee/common/PerFragmentStatsTest.cpp
+++ b/tests/ee/common/PerFragmentStatsTest.cpp
@@ -72,8 +72,8 @@ protected:
         ASSERT_EQ(valueA, voltdb::ValuePeeker::peekInteger(tuple.getNValue(0)));
         ASSERT_EQ(valueB, voltdb::ValuePeeker::peekDouble(tuple.getNValue(1)));
 
-	int32_t length = 0;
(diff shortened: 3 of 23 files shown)


src/ee/plannodes/abstractplannode.cpp
```diff
@@ -58,9 +58,18 @@ using namespace std;
...
-    : m_planNodeId(-1),
-      m_isInline(false),
-      m_validOutputColumnCount(0) { }
+    : m_planNodeId(-1)
+    , m_children()
+    , m_childIds()
+    , m_executor()
+    , m_inlineNodes()
+    , m_isInline(false)
+    , m_outputTable()
```
BillWhite: Isn't C++ supposed to use the default constructor to initialize the non-POD objects?  I would think this change is not doing anything, though it's not bad to ensure it.

src/ee/plannodes/limitnode.h
```diff
@@ -60,8 +60,15 @@ class Table;
...
 class LimitPlanNode : public AbstractPlanNode {
 public:
-    LimitPlanNode() : limit(-1), offset(0), limitParamIdx(-1), offsetParamIdx(-1)
-    {}
+    LimitPlanNode()
+        : limit(-1)
+        , offset(0)
+        , limitParamIdx(-1)
+        , offsetParamIdx(-1)
+        , limitExpression(NULL)
```
BillWhite: Is there a compiler option that complains if POD data members are not explicitly initialized by the constructor?  I think there should be.

tests/ee/common/PerFragmentStatsTest.cpp
```diff
@@ -71,7 +71,10 @@ class PerFragmentStatsTest : public PlanTestingBaseClass<EngineTestTopend> {
     void validateRow(voltdb::TableTuple &tuple, int32_t valueA, double valueB, const char* valueC) {
         ASSERT_EQ(valueA, voltdb::ValuePeeker::peekInteger(tuple.getNValue(0)));
         ASSERT_EQ(valueB, voltdb::ValuePeeker::peekDouble(tuple.getNValue(1)));
-        ASSERT_EQ(0, strcmp(valueC, voltdb::ValuePeeker::peekObjectValue(tuple.getNValue(2))));
+
+	int32_t length = 0;
+	const char* data = voltdb::ValuePeeker::peekObject(tuple.getNValue(2), &length);
+        ASSERT_EQ(0, strncmp(valueC, data, length));
```
BillWhite: What if peekObject returns NULL.  Won't the strncmp crash?  While that's a kind of test failure, maybe it's too harsh.

tests/ee/executors/OptimizedProjectorTest.cpp
```diff
@@ -53,7 +53,7 @@
 #include "storage/temptable.h"
 
 
-static int64_t NUM_ROWS = 10000;
+static int64_t NUM_ROWS = 100;
```
BillWhite: Was this supposed to be put back to 10000?  Is this supposed to test beyond one tuple block?

tests/ee/test_utils/LoadTableFrom.hpp
```diff
@@ -129,7 +129,7 @@ TempTable *loadTableFrom(ReferenceSerializeInputBE& result, bool skipMsgHeader =
                                        schema, // Transfers ownership to the table.
                                        columnNames,
                                        NULL);
-    table->loadTuplesFromNoHeader(result);
+    table->loadTuplesFromNoHeader(result, ExecutorContext::getTempStringPool());
```
BillWhite: Good catch.

tests/ee/test_utils/plan_testing_baseclass.h
```diff
@@ -133,10 +133,19 @@ class PlanTestingBaseClass : public Test {
...
+	::memset(m_parameter_buffer.get(), 0, m_smallBufferSize);
+
         m_per_fragment_stats_buffer.reset(new char[m_smallBufferSize]);
+	::memset(m_per_fragment_stats_buffer.get(), 0, m_smallBufferSize);
+
         m_result_buffer.reset(new char[m_resultBufferSize]);
+	::memset(m_result_buffer.get(), 0, m_resultBufferSize);
+
         m_exception_buffer.reset(new char[m_smallBufferSize]);
+	::memset(m_exception_buffer.get(), 0, m_smallBufferSize);
```
BillWhite: Good catch, all of these memsets.

src/ee/plannodes/abstractplannode.cpp
wolffcm: Yes, you are right.  I made the calls to default constructors explicit so that adding new members later on would make it clear where to put them in the initializer list.

tests/ee/executors/OptimizedProjectorTest.cpp
wolffcm: No, I meant to leave it at 100.  10000 rows does not take a long time running normally, but in valgrind it's very slow.  I had set the number of rows high to get more stable performance numbers, but that doesn't matter for verifying correctness.

PR #4420, opened 2017-04-06
close the socket when exception happens

xinjiacs:
Close the socket when exception happens by adding a try catch block.

Sadly the diff is not so clear.

### DIFF
--- a/src/frontend/org/voltcore/messaging/SocketJoiner.java
+++ b/src/frontend/org/voltcore/messaging/SocketJoiner.java
@@ -447,84 +447,97 @@ public class SocketJoiner {
     private void processSSC(ServerSocketChannel ssc) throws Exception {
         SocketChannel sc = null;
         while ((sc = ssc.accept()) != null) {
-            sc.socket().setTcpNoDelay(true);
-            sc.socket().setPerformancePreferences(0, 2, 1);
-            final String remoteAddress = sc.socket().getRemoteSocketAddress().toString();
+            try {
+                sc.socket().setTcpNoDelay(true);
+                sc.socket().setPerformancePreferences(0, 2, 1);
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -407,38 +407,43 @@ private void doBind() throws Exception {
...
-        }
-        messageBytes.flip();
+            messageBytes.flip();
 
-        JSONObject jsObj = new JSONObject(new String(messageBytes.array(), StandardCharsets.UTF_8));
-        return jsObj;
+            JSONObject jsObj = new JSONObject(new String(messageBytes.array(), StandardCharsets.UTF_8));
+            return jsObj;
+        } catch (IOException | JSONException ex) {
+            sc.close();
```
nshi: I wonder if `processSSC()` is a better place to catch and close the socket.

src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -447,84 +447,97 @@ private JSONObject readJSONObjFromWire(SocketChannel sc, String remoteAddressFor
...
+                    m_joinHandler.notifyOfConnection(jsObj.getInt(HOST_ID), sc, listeningAddress);
+                } else {
+                    throw new RuntimeException("Unexpected message type " + type + " from " + remoteAddress);
+                }
+            } catch (Exception ex) {
+                // do not leak sockets when exception happens
+                try {
+                    sc.close();
+                } catch (IOException ioex) {
+                    LOG.error("Failed to close socket in connect accept loop when having exception");
```
nshi: Please check with Andrew on the error message.

PR #4421, opened 2017-04-06
ENG-11993.  Add sha256sum to build for enterprise and pro.

rmorgenstein:
ENG-11993. Add sha256sum calculation to the builds for enterprise and pro. Kit builds will now produce new files:
- voltdb-ent-x.y.SHA256SUM	
- voltdb-pro-x.y.SHA256SUM	

After downloadking the kit and downloading the .SHA256SUM file to the same directory, the user can then use sha256sum -c to have it check the sum.  For example:

```
>sha256sum -c voltdb-pro-7.2.SHA256SUM
voltdb-pro-7.2.tar.gz: OK
```
Note: There is also a small logic bug fixed. This only affects when the command-line switches are used to limit the kits being built.


rmorgenstein:
To see this running, goto http://ci/job/kit-tools-oneoff-build-distribution/554/

PR #4423, opened 2017-04-07
ENG-11012: Fix trace points and add null check for arguments.

nshi:


src/frontend/org/voltdb/iv2/MpScheduler.java
```diff
@@ -235,7 +235,9 @@ public void handleIv2InitiateTaskMessage(Iv2InitiateTaskMessage message)
                     .add(() -> VoltTrace.meta("thread_sort_index", "sort_index", Integer.toString(100)))
                     .add(() -> VoltTrace.beginAsync("initmp", mpTxnId,
                                                     "txnId", TxnEgo.txnIdToString(mpTxnId),
-                                                    "ciHandle", Long.toString(message.getClientInterfaceHandle())));
+                                                    "ciHandle", message.getClientInterfaceHandle(),
```
akhanzode: Can we get better and consistent names of these selectors? If we are going to document it. e.g. some in camel case some not.

src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -288,8 +290,9 @@ public void createAllParticipatingFragmentWork(FragmentTaskMessage task)
             msg = pollForResponses();
             final FragmentResponseMessage finalMsg = msg;
             if (traceLog != null) {
+                final int batchIdx = m_localWork.getCurrentBatchIndex();
```
akhanzode: m_localWork can be null right?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -988,16 +991,17 @@ private void doLocalFragmentOffer(FragmentTaskMessage msg)
                                  m_pendingTasks, msg, null);
         }
         if (logThis) {
-            if (traceLog != null) {
+            ListenableFuture<Object> durabilityBackpressureFuture =
+                    m_cl.log(msg.getInitiateTask(), msg.getSpHandle(), Ints.toArray(msg.getInvolvedPartitions()),
+                             m_durabilityListener, task);
+
+            if (durabilityBackpressureFuture != null && traceLog != null) {
```
akhanzode: nit: traceLog != check should be first as it will be false most of the time.

src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -1000,13 +1000,14 @@ public static String hsIdTxnIdToString(long hsId, long txnId) {
         return sb.toString();
     }
 
-    public static String hsIdPairTxnIdToString(long srcHsId, long destHsId, long txnId) {
+    public static String hsIdPairTxnIdToString(long srcHsId, long destHsId, long txnId, long uniqID) {
```
akhanzode: all params can be final.

src/frontend/org/voltdb/utils/VoltTrace.java
```diff
@@ -146,7 +145,7 @@ private void mapFromArgArray() {
 
             for (int i=0; i<m_argsArr.length; i+=2) {
                 if (i+1 == m_argsArr.length) break;
-                m_args.put(m_argsArr[i].toString(), m_argsArr[i+1].toString());
+                m_args.put(String.valueOf(m_argsArr[i]), String.valueOf(m_argsArr[i+1]));
```
akhanzode: if m_argsArr has nulls do you expect them to get "null" and "null"

src/frontend/org/voltdb/iv2/MpScheduler.java
nshi: Aren't they consistent now? The arguments are all in camel case.

I don't think any of these will be documented because they are not meant to be read or understood by users. It's a developer debugging tool.

src/frontend/org/voltdb/utils/VoltTrace.java
nshi: I guess so. These are arbitrary arguments developers would like to get logged. So they can put anything they want here.

src/frontend/org/voltdb/iv2/MpTransactionState.java
nshi: No, it cannot. Every batch has local work. You'll actually see that we use `m_localWork` several lines above. It's set to null after the current block of while loop.

nshi:
@akhanzode I addressed two of your comments and answered the rest.

src/frontend/org/voltdb/iv2/MpScheduler.java
akhanzode: initmp looked odd but since its dev only. I would say looks good.

nshi: Yeah...The first string is the name of the trace points, the rest are parameters. I made the trace point names all lower case throughout the code base for no particular reason.

PR #4424, opened 2017-04-07
ENG-12170

akhanzode:
When there are no stream changes in UAC dont roll generation.
Stream changes are considered to be present if

1. Add/Drop stream
3. Connector or Connector properties changes.

CatalogDiff detects it.
Added test in CatalogDIffEngine


src/catgen/spec.txt
```diff
@@ -131,6 +131,7 @@ begin Table                                  "A table (relation) in the database
   string signature                           "Catalog version independent signature of the table consisting of name and schema"
   int tuplelimit                             "A maximum number of rows in a table"
   bool isDRed                                "Is this table DRed?"
+  bool isStream                              "Is this table a Stream?"
```
nshi: We must already have a way to identify stream tables, right? Why do we need an extra flag?

Also note that the catalog has to be backward compatible because we support migrating older catalogs to new one. DR also needs to be able to parse old/new catalog commands.

src/ee/voltdbjni.cpp
```diff
@@ -361,7 +361,7 @@ SHAREDLIB_JNIEXPORT jint JNICALL Java_org_voltdb_jni_ExecutionEngine_nativeLoadC
 SHAREDLIB_JNIEXPORT jint JNICALL
 Java_org_voltdb_jni_ExecutionEngine_nativeUpdateCatalog(
     JNIEnv *env, jobject obj,
-    jlong engine_ptr, jlong timestamp, jbyteArray catalog_diffs) {
+    jlong engine_ptr, jlong timestamp, jboolean stremChanges, jbyteArray catalog_diffs) {
```
nshi: Typo, `streamChanges`.

src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -503,6 +507,10 @@ else if (suspect instanceof Table) {
                     }
                 }
             } else if (tbl.getMaterializer() != null) {
+                //If this is a view on stream lets roll generation.
```
akhanzode: Technically yes but on EE side it rebuilds the view collection, let me handle that case and ignore view addition to ignore generation.

akhanzode: I will ignore view deletions as well.

src/catgen/spec.txt
akhanzode: Yes thats more like going through connectors. I can get rid of it. I needed it for loadCatalog in EE which if it creates new STREAM will update generation and force and roll anyway.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -867,15 +865,18 @@ bool VoltDBEngine::processCatalogAdditions(int64_t timestamp) {
...
-                if (!tcd->exportEnabled()) {
-                    // Evaluate export enabled or not and cache it on the tcd.
-                    tcd->evaluateExport(*m_database, *catalogTable);
-                    // If enabled hook up streamer
-                    if (tcd->exportEnabled() && streamedTable->enableStream()) {
-                        //Reset generation after stream wrapper is created.
-                        streamedTable->setSignatureAndGeneration(catalogTable->signature(), timestamp);
-                        m_exportingTables[catalogTable->signature()] = streamedTable;
+                //Dont update and roll generation if this is just a non stream table update.
+                if (isStreamUpdate) {
```
xinjiacs: Better to concat it with the above IF? `if (streamedTable  && isStreamUpdate )` to remove one indentation.

PR #4425, opened 2017-04-07
Add ability to start VoltDB with a staged catalog

brians127:
Part 2 of PR 4335 - adds full support for the following usage model:

voltdb init --schema (schema) --classes (jarfile)
voltdb start

The targeted customer use case is being able to initialize a new cluster using the results of 'voltdb get'. I expect others will use this approach to initialize databases with their own schemas as well. Not only is this approach convenient, any warnings or errors emitted by VoltCompiler are delivered to the user during 'init' rather than the server daemon's stderr.

'init' is already covered by previous work. 'start' has to consider two sets of cases:
1. Should the database's copy of the user's schema persist?
   a. If command logging or any other durability is on, no. 
   b. Otherwise, yes, but don't keep the staged schema up to date with all DDL changes (it's not a substitute for durability)
   c. Staged schemas are always ignored during recovery from a command log or snapshot.
2. How does the rest of the cluster impact the new node?
   a. During cluster create, all nodes must agree on the same staged schema. In the future, we may allow empty nodes to assume the cluster's staged schema.
   b. During a rejoin, ignore any staged schema. Use cluster's schema.
   c. When restarting a cluster, command logs and snapshots will be used if possible even if staged schemas exist.

The code is fairly aggressive about deleting staged schemas that are not being used, such as when command logging is on. These deletes do not prevent the cluster from starting if they fail.

The term "initialized schema" is used consistently in user visible messages, but the term "staged catalog" is used widely by the code. The initialized schema is a catalog emitted by VoltCompiler (minus a few artifacts like the report) and is staged in the voltdbroot/config directory.

UX: https://docs.google.com/document/d/14UmfRidF7ggIqOpoVP-4syQNM22EkO7z7V21p95Gefs/edit#
Pro: https://github.com/VoltDB/pro/pull/1833


tests/frontend/org/voltdb/TestInitStartAction.java
```diff
@@ -384,4 +391,27 @@ public void testInitWithSchemaInvalidMissingClass() throws Exception {
         assertTrue(VoltDB.crashMessage.contains("Could not compile specified schema"));
         assertEquals(true, schemaFile.delete());
     }
+
+    /** Tests that a cluster can start with each node using an identical staged catalog.
+     * Since durability is off, the staged catalog will be kept around but not reflect any DDL changes.
+     * Restarting the database will reload the previous staged catalog.
+     */
+    @Test
```
hhakimi53: Another place you may want to add the tests are to TestInitStartLocalClusterAllOutOfProcess and TestInitStartLocalClusterInProcess. In there, "testclusterup()" initializes cluster, seeds it with sql and procedures thn collects schema and procedure from them and verifiies the data. The existing test logic can be updated so that reinit cluster with the schema and classes resources it got from the get command. To verify round trip, the existing tables and procedures from database can be checked once it has been reinited and started.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -425,6 +427,8 @@ public void startSampler() {
 
     private LatencyHistogramStats m_latencyHistogramStats;
 
+    private boolean m_durable;
```
hhakimi53: move the comment "durable means commandlogging is enabled." from initialize to here.

tests/frontend/org/voltdb/TestInitStartAction.java
brians127: @akhanzode what do you think? Is there any reason to keep tests for "init + start" without the staged schema mechanism? I am fine with Huzefa's suggestion since compiling a catalog is not how users should load a schema into VoltDB with NewCLI.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -426,6 +428,9 @@ public void startSampler() {
 
     private LatencyHistogramStats m_latencyHistogramStats;
 
+    // durable means command logging is enabled.
+    private boolean m_durable;
```
akhanzode: Isnt this already there? why turn it into member?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -987,7 +988,17 @@ public void initialize(Configuration config) {
 
             m_clusterSettings.set(clusterSettings, 1);
 
-            MeshProber.Determination determination = buildClusterMesh(readDepl);
+            UUID startupCatalogHash = null;
+            if (m_pathToStartupCatalog != null){
```
akhanzode: Why not move this into mesh and let mesh build their hashes all it needs is a path to startup catalog.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2210,12 +2220,16 @@ private void stageSchemaFiles(Configuration config){
             return; // nothing to do
         }
         assert( config.m_userSchema.isFile() ); // this is validated during command line parsing and will be true unless disk faults
-        File stagedCatalogFH = new VoltFile(getStagedCatalogPath());
-        assert( !stagedCatalogFH.exists() || config.m_forceVoltdbCreate ); // managedPathsWithFiles() checks for the staged catalog
+        File stagedCatalog = new VoltFile(getStagedCatalogPath());
```
akhanzode: What is the change here dont change variable names unnecessarily. makes diff big.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3859,7 +3874,25 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
         /*
          * Remove the terminus file if it is there, which is written on shutdown --save
          */
-        new File(m_nodeSettings.getVoltDBRoot(), VoltDB.TERMINUS_MARKER).delete();
+        final File terminusMarker = new File(m_nodeSettings.getVoltDBRoot(), VoltDB.TERMINUS_MARKER);
```
akhanzode: Why are you deleting things here especially terminus. This is not needed.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3859,7 +3874,25 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
...
+        final boolean terminusExists = terminusMarker.exists();
+        if (terminusExists){
+            boolean success = terminusMarker.delete();
+            if (!success){
+                hostLog.warn("Could not delete terminus marker");
+            }
+        }
+        if (m_durable || terminusExists){
+            File stagedCatalog = new VoltFile(RealVoltDB.getStagedCatalogPath(m_nodeSettings.getVoltDBRoot().getAbsolutePath()));
```
akhanzode: This should happen after the cluster says its has successfully initialized.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -836,6 +843,26 @@ public boolean compile(
                        ppdEnabled, snapshotPath, ppdPrefix);
     }
 
+    private void validateVoltDbRoot(String voltRootPath){
```
akhanzode: Dont move around code again diff gets panful to review.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -904,27 +920,53 @@ public boolean compile(final VoltCompiler compiler,
...
+                                        int hostCount,
+                                        int sitesPerHost,
+                                        int replication,
+                                        int clusterId)
+    {
+        DeploymentInfo deployment = new DeploymentInfo(hostCount, sitesPerHost, replication, clusterId);
+        return compileDeploymentOnly(voltDbRoot, deployment);
+    }
+
+    private String compileDeploymentOnly(String voltDbRoot, DeploymentInfo deployment)
```
akhanzode: Same as previous comment.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -1011,9 +1053,11 @@ private String writeDeploymentFile(
         // <paths>
         PathsType paths = factory.createPathsType();
         deployment.setPaths(paths);
-        Voltdbroot voltdbroot = factory.createPathsTypeVoltdbroot();
-        paths.setVoltdbroot(voltdbroot);
-        voltdbroot.setPath(voltRoot);
+        if ((voltRoot != null) && !voltRoot.trim().isEmpty()){
```
akhanzode: Is this a required change?

src/frontend/org/voltdb/probe/HostCriteria.java
```diff
@@ -235,6 +248,13 @@ public boolean isUndefined() {
             ilb.add("Servers have different startup snapshots nonces: "
                     + m_terminusNonce + " vs. " + o.m_terminusNonce);
         }
+        if ((m_startupCatalogHash == null) ^ (o.m_startupCatalogHash == null)){
+            ilb.add("Staged catalog is missing on at least one server. "
```
akhanzode: These messages will be shown to user ask Andrew and Ruth what term we want to use "staged catalog" does not look right. Most likely we should say schema and classes explicitly to tie it to cli options.

src/frontend/org/voltdb/probe/HostCriteria.java
```diff
@@ -235,6 +248,13 @@ public boolean isUndefined() {
...
+                    + "If any server has a staged catalog, all must have identical staged catalogs.");
+        } else if ((m_startupCatalogHash != null) && !m_startupCatalogHash.equals(o.m_startupCatalogHash)){
+            ilb.add("Servers have different staged catalog hashes: "
```
akhanzode: Same as above.

src/frontend/org/voltdb/probe/HostCriteria.java
```diff
@@ -301,6 +323,12 @@ public boolean equals(Object obj) {
             return false;
         if (m_nodeState != other.m_nodeState)
             return false;
+        if ((m_startupCatalogHash == null) ^ (other.m_startupCatalogHash == null)){
```
akhanzode: Why not compare it like m_meshHash?

src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -268,6 +263,24 @@ public long getCRC() {
         return md.digest();
     }
 
+    public byte[] getSha1Hash() {
+        try {
+            return getChecksum("SHA-1");
+        } catch (NoSuchAlgorithmException e) {
```
akhanzode: you can catch this in getChecksum and have only one crash.

src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -268,6 +263,24 @@ public long getCRC() {
...
+            VoltDB.crashLocalVoltDB("Bad JVM has no SHA-1 hash.", true, e);
+            throw new RuntimeException("crashLocalVoltDB didn't work - this should be impossible, but exception is required to make compiler happy");
+        }
+    }
+
+    public UUID getMD5Checksum() {
+        try {
+            return Digester.md5AsUUID(getChecksum("MD5"));
+        } catch (NoSuchAlgorithmException e) {
```
akhanzode: same as above only one crash place.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -343,8 +378,16 @@ public LocalCluster(String jarFileName,
             buildDir = System.getProperty("user.dir") + "/obj/release";
         }
 
-        String classPath = System.getProperty("java.class.path") + ":" +
-                buildDir + File.separator + m_jarFileName + ":" + buildDir + File.separator + "prod";
+        String classPath = System.getProperty("java.class.path");
```
akhanzode: what is this change?

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -2035,6 +2073,68 @@ public static LocalCluster createLocalCluster(String schemaDDL, int siteCount, i
...
+     * Catalog compilation is taken care of by VoltDB itself - no need to do so explicitly.
+     * @param schemaDDL
+     * @param proceduresJar
+     * @param siteCount
+     * @param hostCount
+     * @param kfactor
+     * @return a running VoltDB cluster
+     * @throws IOException
+     */
+    public static LocalCluster createLocalClusterViaStagedCatalog(
```
akhanzode: why a need for static methods to construct? Add a new ctor if needed dont create multiple ways of instantiating LocalCluster "follow the pattern".

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -2046,4 +2146,21 @@ public Client createClient(ClientConfig config) throws IOException {
     public void setDeplayBetweenNodeStartup(long deplayBetweenNodeStartup) {
         m_deplayBetweenNodeStartupMS = deplayBetweenNodeStartup;
     }
+
+    /** Counts how many nodes contain a given file in their voltDbRoots
+     * @param relativePathFromVoltDBRoot
+     * @return number of nodes who have that file
+     */
+    public int countNodesWithFile(String relativePathFromVoltDBRoot){
```
akhanzode: Have no idea whats the purpose of this in LocalCluster

src/frontend/org/voltdb/RealVoltDB.java
brians127: Terminus was already getting deleted. I added logging to note when there's a failure to remove it, so terminus is consistent with the staged catalog. I can revert that change if need be.

src/frontend/org/voltdb/probe/HostCriteria.java
brians127: You're right, I will switch terminology to reference staged schemas.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
brians127: I use it to verify how many nodes have a staged catalog. This is an important check since it differentiates the "no durability" and "with durability" cases. Putting the logic here means I don't have multiple copies of the same code in different tests.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
brians127: jarFileName can be null now, so I modified the classpath building to handle this.

src/frontend/org/voltdb/RealVoltDB.java
brians127: I'll look into this

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
brians127: I had to do this due to a test failure. I'll see if I can eliminate this when I address your comments on excessive code changes.

src/frontend/org/voltdb/utils/InMemoryJarfile.java
brians127: OK

tests/frontend/org/voltdb/TestStagedCatalogStart.java
```diff
@@ -0,0 +1,118 @@
...
+            fail();
+        }
+        nodesWithStagedCatalog = cluster.countNodesWithFile(Constants.CONFIG_DIR + File.separator + CatalogUtil.STAGED_CATALOG_FILE_NAME);
+        assertEquals(nodesWithStagedCatalog, hostCount);
+        clearLocalDataDirectories = false;
+        skipInit = true;
+        cluster.startUp(clearLocalDataDirectories, skipInit);
+        // FIXME how to verify the error message is as expected? It is, but test doesn't know it is.
+
+        // TODO: test that different stored procedure code results in startup failure
```
brians127: FIXMEs and TODOs are known deficiencies to be addressed before this work is considered complete.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -2035,6 +2076,58 @@ public static LocalCluster createLocalCluster(String schemaDDL, int siteCount, i
...
+                schemaDDL,
+                null,
+                siteCount,
+                hostCount,
+                kfactor,
+                clusterID,
+                BackendTarget.NATIVE_EE_JNI,
+                FailureState.ALL_RUNNING, false, false, null);
+
+        assert(proceduresJar == null) : "TODO: add ability to insert a procedures JAR into the classpath";
```
brians127: TODO that Eclipse probably won't pick up

src/frontend/org/voltdb/RealVoltDB.java
akhanzode: You can make that determination upfront keep it in boolean if you need to delete stage catalog. Also it needs to be deleted just before we start accepting connections see startAcceptingConnections

brians127: As per our discussion earlier today, I'll be sure that recovery + durability deletes staged catalogs as well. This code will remain here but be conditional on an earlier check to see if the terminus exists.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2767,6 +2768,15 @@ public Integer get() {
             }
         };
 
+        UUID startupCatalogHash = null;
+        if (m_pathToStartupCatalog != null){
+            try {
+                startupCatalogHash = new InMemoryJarfile(m_pathToStartupCatalog).getMD5Checksum();
+            } catch (IOException e){
+                VoltDB.crashLocalVoltDB("Failed to load schema and classes from staging location " + m_pathToStartupCatalog, false, e);
```
brians127: You're right - I added this to the document, with a change I plan to make.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3949,6 +3964,21 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
         m_messenger.getZK().create(VoltZK.init_completed, null, Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT, new ZKUtil.StringCallback(), null);
     }
 
+    private void deleteStagedCatalogIfNeeded() {
+        if (((m_commandLog != null) && m_commandLog.isEnabled()) || (m_terminusNonce != null)){
+            assert((m_terminusNonce == null) || !m_terminusNonce.trim().isEmpty());
+            File stagedCatalog = new VoltFile(RealVoltDB.getStagedCatalogPath(m_nodeSettings.getVoltDBRoot().getAbsolutePath()));
+            if (stagedCatalog.exists()){
```
brians127: What are you looking for? A space before the { ?

src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -234,38 +235,45 @@ public long getCRC() {
...
-            else {
-                md.update(e.getKey().getBytes(Constants.UTF8ENCODING));
-                md.update(e.getValue());
             }
+            return md.digest();
+
+        } catch (NoSuchAlgorithmException e) {
+            VoltDB.crashLocalVoltDB("JVM has no " + checksumType + " hash.", true, e);
         }
+        throw new RuntimeException("crashLocalVoltDB didn't work - this should be impossible");
```
brians127: Compiler requires it because it doesn't recognize crashLocalVoltDB() as a System.exit() call. I could use "return null" instead but this is more descriptive.

src/frontend/org/voltdb/RealVoltDB.java
hhakimi53: a space between closing bracket and curly bracket - "if (...)<space>{"

src/frontend/org/voltdb/utils/InMemoryJarfile.java
hhakimi53: The message provided in the RuntimeException is confusing as it's conveying that voltdb crash didn't work, which should not happen. If was compiler, can add comment around it. Once the try-catch block is wrapped around the resource that generates it, see the prior comment for L#266, this throw will not be needed.

src/frontend/org/voltdb/probe/HostCriteria.java
```diff
@@ -106,18 +109,21 @@ public HostCriteria(JSONObject jo) {
...
             int hostCount, NodeState nodeState, boolean addAllowed,
-            boolean safeMode, String terminusNonce) {
+            boolean safeMode, String terminusNonce, UUID startupCatalogHash) {
         checkArgument(configHash != null, "config hash is null");
         checkArgument(meshHash != null, "mesh hash is null");
         checkArgument(startAction != null, "start action is null");
         checkArgument(hostCount > 0, "host count %s is less then one", hostCount);
         checkArgument(terminusNonce == null || !terminusNonce.trim().isEmpty(),
                 "terminus should not be blank");
+        // startupCatalogHash may be null
```
hhakimi53: how about something on the lines of "checkArgument(startupCatalogHash == null || uuid is good, "msg" );" ?

src/frontend/org/voltdb/probe/MeshProber.java
```diff
@@ -227,6 +229,7 @@ protected MeshProber(NavigableSet<String> coordinators,
                 hostCountSupplier.get(), coordinators.size());
         checkArgument(terminusNonce == null || !terminusNonce.trim().isEmpty(),
                 "terminus should not be blank");
+        // stagedConfigHash can be null - that means there isn't a staged catalog
```
hhakimi53: check if uuid is not null, if it's not null -> it's good?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2211,12 +2208,17 @@ private void stageSchemaFiles(Configuration config){
...
         assert( config.m_userSchema.isFile() ); // this is validated during command line parsing and will be true unless disk faults
         File stagedCatalogFH = new VoltFile(getStagedCatalogPath());
-        assert( !stagedCatalogFH.exists() || config.m_forceVoltdbCreate ); // managedPathsWithFiles() checks for the staged catalog
+
+        // this check cannot be part of managedPathsWithFiles(), since "voltdb start" can get translated into "voltdb create" after probing the mesh.
+        if (!config.m_forceVoltdbCreate && stagedCatalogFH.exists() && stagedCatalogFH.canRead()){
+            VoltDB.crashLocalVoltDBNoArtifacts("A previous database was initialized with a schema. You must init with --force to overwrite the schema.");
+        }
         final boolean standalone = true;
         final boolean isXCDR = false;
```
brians127: Misspelled

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2211,12 +2208,17 @@ private void stageSchemaFiles(Configuration config){
...
-        VoltCompiler compiler = new VoltCompiler(standalone, isXCDR);
+        final boolean generateReports = false; // this will be overridden if debug mode is enabled
```
brians127: I find it makes the code easier to read. Why do you think I should change it?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2211,12 +2208,17 @@ private void stageSchemaFiles(Configuration config){
         }
         assert( config.m_userSchema.isFile() ); // this is validated during command line parsing and will be true unless disk faults
         File stagedCatalogFH = new VoltFile(getStagedCatalogPath());
-        assert( !stagedCatalogFH.exists() || config.m_forceVoltdbCreate ); // managedPathsWithFiles() checks for the staged catalog
+
+        // this check cannot be part of managedPathsWithFiles(), since "voltdb start" can get translated into "voltdb create" after probing the mesh.
+        if (!config.m_forceVoltdbCreate && stagedCatalogFH.exists() && stagedCatalogFH.canRead()){
+            VoltDB.crashLocalVoltDBNoArtifacts("A previous database was initialized with a schema. You must init with --force to overwrite the schema.");
```
brians127: Already discussed with Andrew. I'll make the other minor changes

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2767,6 +2769,15 @@ public Integer get() {
             }
         };
 
+        UUID startupCatalogHash = null;
+        if (m_pathToStartupCatalog != null) {
+            try {
+                startupCatalogHash = new InMemoryJarfile(m_pathToStartupCatalog).getMD5Checksum();
+            } catch (IOException e){
+                VoltDB.crashLocalVoltDB("Failed to load schema from staging location within voltdbroot: " + e.getMessage(), false, e);
```
brians127: OK

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3950,6 +3966,19 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
...
     }
 
+    private void deleteStagedCatalogIfNeeded() {
+        if (((m_commandLog != null) && m_commandLog.isEnabled()) || (m_terminusNonce != null)) {
+            File stagedCatalog = new VoltFile(RealVoltDB.getStagedCatalogPath(m_nodeSettings.getVoltDBRoot().getAbsolutePath()));
+            if (stagedCatalog.exists()) {
+                if (stagedCatalog.delete()){
+                    hostLog.info("Deleted VoltDB's copy of the initialized schema because durability is present.");
+                } else {
+                    hostLog.warn("Could not delete VoltDB's copy of the initialized schema.");
```
brians127: I don't believe it's necessary. This is unexpected, and really should not happen unless the filesystem is bad or the user has been messing with file permissions. If the schema is left around, it'll be ignored and deleted next time Volt starts.

src/frontend/org/voltdb/RealVoltDB.java
hhakimi53: there is only one consumer of all those variables. if they are not getting reused or getting based on certain condition, it's not that useful except for debugging and it makes bit verbose. That's how it's currently done at other places in the same file also.

src/frontend/org/voltdb/probe/HostCriteria.java
brians127: Any 128-bit sequence is a potentially valid MD5 hash, and UUIDs are documented to be 128-bit sequences. What check can I perform?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -195,8 +195,7 @@
 
     private static boolean m_exportEnabled = false;
     public static final String CATALOG_FILE_NAME = "catalog.jar";
-
-    public static final String STAGED_CATALOG_FILE_NAME = "staged-catalog.jar";
+    public static final String STAGED_CATALOG = Constants.CONFIG_DIR + File.separator + "staged-catalog.jar";
```
brians127: OK

src/frontend/org/voltdb/RealVoltDB.java
hhakimi53: OK. there is a check in there to check on the status and msg getting logged for it. If the deleition operation failed, it will be better to try to delete it again for the deletion failure case

src/frontend/org/voltdb/probe/HostCriteria.java
hhakimi53: What if sataged path was empty and it computed hash on that? So the check can verify that logic is not operating on empty string hash (if that can happen).

brians127: I'm relying on InMemoryJarfile and by extension Java's JarInputStream to validate that the file being loaded is a real .jar file. I can add tests to InMemoryJarfile to ensure it throws exceptions on bad input, such as an empty or nonexistent file. Hash comparison is not the right place for this.

brians127: Discussed in person. I agreed to add stricter checks that the staged catalog file exists and is a file earlier in the code. Once the MD5 hash is taken, too much of the input data has been removed to do proper validation.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -606,17 +673,27 @@ else if (clearLocalDataDirectories) {
...
+        }
+        m_localServer = new ServerThread(cmdln, alwaysForceCreate);
         m_localServer.start();
     }
 
-    public String getServerSpecificRoot(String hostId) {
+    /** Gets the voltdbroot directory for the specified host.
+     * Formerly "getServerSpecificRoot"; name changed since behavior is inconsistent with {@link VoltFile#getServerSpecificRoot(String, boolean)}.
+     */
+    public String getServerSpecificVoltDBRoot(String hostId) {
```
akhanzode: Dont change the method name.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -890,16 +971,22 @@ private void initOne(int hostId, boolean clearLocalDataDirectories) throws IOExc
                 cmdln.setJavaProperty(name, this.m_additionalProcessEnv.get(name));
             }
         }
+        File root = null;
         try {
             //If clear clean VoltFile.getServerSpecificRoot(String.valueOf(hostId))
-            File root = VoltFile.getServerSpecificRoot(String.valueOf(hostId), clearLocalDataDirectories);
-            cmdln = cmdln.voltdbRoot(root);
+            root = VoltFile.getServerSpecificRoot(String.valueOf(hostId), clearLocalDataDirectories);
```
akhanzode: Avoid unnecessary changes it becomes very to review which is related to functionality.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -961,13 +1048,15 @@ private void initOne(int hostId, boolean clearLocalDataDirectories) throws IOExc
...
+        if ( root != null ) {
+            String hostIdStr = cmdln.getJavaProperty(clusterHostIdProperty);
+            m_hostRoots.put(hostIdStr, root.getPath());
+        }
     }
 
-    private void startOne(int hostId, boolean clearLocalDataDirectories,
-            StartAction startAction, boolean waitForReady, String placementGroup)
-    throws IOException {
+    public void startOne(int hostId, boolean clearLocalDataDirectories,
```
akhanzode: Why does it need to be public? Does not look like it has any staged catalog change addition to signature.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -2071,4 +2166,35 @@ public Client createAdminClient(ClientConfig config) throws IOException {
     public void setDeplayBetweenNodeStartup(long deplayBetweenNodeStartup) {
         m_deplayBetweenNodeStartupMS = deplayBetweenNodeStartup;
     }
+
+    /** Obtains location of a file in a node's voltdbroot.
+     * @param relativePathFromVoltDBRoot
+     * @return absolute file path
+     */
+    public String getFileFromVoltDbRoot(String nodeID, String relativePathFromVoltDBRoot) {
```
akhanzode: Why do you need to add method to LocalCluster which you can do in your test. It just makes already too big file bigger.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -2071,4 +2166,35 @@ public Client createAdminClient(ClientConfig config) throws IOException {
...
+        assert( rootpath.contains(Constants.DBROOT) == false ) : rootpath;
+        File testFile = new VoltFile(rootpath + File.separator + Constants.DBROOT + File.separator + relativePathFromVoltDBRoot);
+        return testFile.getAbsolutePath();
+    }
+
+    /** Counts how many nodes contain a given file in their voltDbRoots
+     * @param relativePathFromVoltDBRoot
+     * @return number of nodes who have that file
+     */
+    public int countNodesWithFile(String relativePathFromVoltDBRoot) {
```
akhanzode: Same as above. For such things if you have multiple tests using it create a base test class to keep functions like this.

tests/frontend/org/voltdb/utils/TestCollector.java
```diff
@@ -105,7 +105,7 @@ public void setUp() throws Exception {
         cluster.startUp(true);
         //Get server specific root after startup.
         if (cluster.isNewCli()) {
-            voltDbRoot = new File(cluster.getServerSpecificRoot("0"));
+            voltDbRoot = new File(cluster.getServerSpecificVoltDBRoot("0"));
```
akhanzode: Remove all these changes.

src/frontend/org/voltdb/Inits.java
```diff
@@ -280,6 +280,23 @@ public void run() {
         return Arrays.copyOf(buffer, totalBytes);
     }
 
+    // NOTE: This MUST be deterministic. Separate nodes may verify that the SHA-1 hashes of empty catalog files match.
+    private void createEmptyStartupJarFileIfNeeded(){
```
akhanzode: This function does not make any sense leave it in the DistributeCatalog. Or it it has to be a function should be something useful.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2207,18 +2200,23 @@ private void stageDeploymentFileForInitialize(Configuration config, DeploymentTy
...
+        if (config.m_userSchema == null) {
             return; // nothing to do
         }
         assert( config.m_userSchema.isFile() ); // this is validated during command line parsing and will be true unless disk faults
-        File stagedCatalogFH = new VoltFile(getStagedCatalogPath());
-        assert( !stagedCatalogFH.exists() || config.m_forceVoltdbCreate ); // managedPathsWithFiles() checks for the staged catalog
+        File stagedCatalogFH = new VoltFile(getStagedCatalogPath(getVoltDBRootPath()));
+
+        // this check cannot be part of managedPathsWithFiles(), since "voltdb start" can get translated into "voltdb create" after probing the mesh.
+        if (!config.m_forceVoltdbCreate && stagedCatalogFH.exists()) {
```
akhanzode: You should come into this function on INIT only right? The comment suggests voltdb start which is not accurate right?

src/frontend/org/voltdb/ServerThread.java
```diff
@@ -32,6 +32,10 @@
     VoltDB.Configuration m_config;
 
     public ServerThread(VoltDB.Configuration config) {
+        this(config, true);
+    }
+
+    public ServerThread(VoltDB.Configuration config, boolean alwaysForceCreate) {
```
akhanzode: Why do you need additional param if config.alwaysForceCreate is available. and why do you need |= when you just assigned m_config = config this is too cryptic.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -708,6 +708,21 @@ else if (arg.equalsIgnoreCase("force")) {
                 referToDocAndExit();
             }
 
+            final File stagedCatalogLocation = new VoltFile(RealVoltDB.getStagedCatalogPath(m_voltdbRoot.getAbsolutePath()));
```
akhanzode: This can all be in RealVoltDB. Only keep validation that require us to exit. also all cases of what start action then remains in one file.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -1171,10 +1189,23 @@ private static void sendCrashSNMPTrap(String msg) {
...
+    }
+
     /**
      * Exit the process with an error message, optionally with a stack trace.
      */
     public static void crashLocalVoltDB(String errMsg, boolean stackTrace, Throwable thrown) {
+        crashLocalVoltDB(errMsg, stackTrace, thrown, true);
+    }
+
+    private static void crashLocalVoltDB(String errMsg, boolean stackTrace, Throwable thrown, boolean allowArtifacts) {
```
akhanzode: remove all crashLocalVoltDB or crash changes.

tests/frontend/org/voltdb/TestSSL.java
```diff
@@ -260,7 +260,7 @@ public void ntestLocalClusterRejoin() throws Exception {
         config.m_startAction = m_cluster.isNewCli() ? StartAction.PROBE : StartAction.REJOIN;
         config.m_pathToCatalog = Configuration.getPathToCatalogForTest("sslRejoin.jar");
         if (m_cluster.isNewCli()) {
-            config.m_voltdbRoot = new File(m_cluster.getServerSpecificRoot("0"));
+            config.m_voltdbRoot = new File(m_cluster.getServerSpecificVoltDBRoot("0"));
```
akhanzode: Unnecessay changes.

tests/frontend/org/voltdb/TestStartWithSchema.java
```diff
@@ -0,0 +1,201 @@
...
+
+    static final String mismatchSchema =
+            "create table TEST (myval bigint not null, PRIMARY KEY(myval));";
+
+    @Test
+    public void testMatch() throws Exception
+    {
+        // Creates a cluster on the local machine using NewCLI, staging the specified schema.
+        // Catalog compilation is taken care of by VoltDB itself - no need to do so explicitly.
+        LocalCluster cluster = new LocalCluster(
```
akhanzode: Use setup and tearDown

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -708,6 +708,21 @@ else if (arg.equalsIgnoreCase("force")) {
...
+                switch (m_startAction) {
+                case PROBE:
+                    assert (m_pathToCatalog == null) : m_pathToCatalog;
+                    m_pathToCatalog = stagedCatalogLocation.getAbsolutePath();
+                    break;
+                case INITIALIZE:
+                    // existence of previous staged catalog will be addressed later
+                    break;
+                default:
+                    hostLog.warn("Initialized schema is present, but is being ignored and may be removed.");
```
hhakimi53: which scenario will this case happen in? does it need to at warning level?

src/frontend/org/voltdb/ServerThread.java
brians127: I am being backward compatible for other tests. Adding the new param means this change only happens for tests I added.

src/frontend/org/voltdb/VoltDB.java
brians127: If I recall, @ajgent specifically requested that well articulated user errors do not drop a crash file.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
hhakimi53: does it need to be public? if it's getting used by test in regression test suite, can make it package private

tests/frontend/org/voltdb/TestStartWithSchema.java
```diff
@@ -0,0 +1,201 @@
...
+            cluster.setExpectedToCrash(false); // ensure we get the exception
+            cluster.startUp(clearLocalDataDirectories, skipInit);
+            fail("Cluster started with mismatched schemas");
+        } catch (Exception e){
+            if (e.getMessage() != null &&
+                    e.getMessage().contains("external processes failed to start"))
+            {
+                // expected
+            } else {
+                throw e;
```
hhakimi53: if (e.getMessage() == null ||  !e.getMessage().contains("external processes failed to start")) { fail() }

src/frontend/org/voltdb/VoltDB.java
akhanzode: In those case you dont want it to go through crash you can simply system out and do VoltDB.exit() not required to do crash path changes.

tests/frontend/org/voltdb/TestStartWithSchema.java
```diff
@@ -0,0 +1,201 @@
...
+            cluster.setExpectedToCrash(false); // ensure we get the exception
+            cluster.startUp(clearLocalDataDirectories, skipInit);
+            fail("Cluster started with a node missing the staged schema");
+        } catch (Exception e){
+            if (e.getMessage() != null &&
+                e.getMessage().contains("external processes failed to start"))
+            {
+                // expected
+            } else {
+                throw e;
```
hhakimi53: same as previous comment

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
brians127: Last time I looked, it is being used in a test. I'll double check if it is still used, and if so go with package scope.

src/frontend/org/voltdb/Inits.java
```diff
@@ -364,6 +368,25 @@ public void run() {
                 }
             } while (catalogStuff == null || catalogStuff.catalogBytes.length == 0);
 
+            assert( m_rvdb.m_config.m_startAction != StartAction.PROBE );
+            if (m_rvdb.m_config.m_startAction == StartAction.CREATE){
+                // We may have received a staged catalog from the leader.
+                // Check if it matches ours.
+                if (m_rvdb.m_pathToStartupCatalog == null) {
```
nshi: You can skip the check if this is null.

brians127: Isn't it 'null' on a bare node with neither a staged schema, a user-specified catalog, nor recovery? I could simplify the code by performing this check once in DistributeCatalog regardless of whether or not the node in question is the leader - is that what you are suggesting?

nshi: If it's `null`, there's no point generating an empty catalog *just* for the check. In that case, just use whatever catalog is in ZK.

nshi: Sorry for the confusion, what I meant was the hash check below. If there is no staged catalog locally, then there is no need to generate an empty catalog for the hash check.

brians127: How then does the node know if ZK has an empty catalog? There is no reserved value for an empty catalog that each node can check for - the leader will produce a hash.

Being able to join empty nodes with non-empty ones if/only if the leader is non empty will confuse users, because they now have to try and control who is the leader. I'd rather have consistent behavior regardless of who is the leader, even if it's more restrictive.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
brians127: This particular change is necessary, because I need to consistently know whether or not m_hostRoots contains the 'voltdbroot' directory in the path. Before my changes, it could vary depending on how the cluster is set up.

brians127:
I believe I have addressed review feedback at this point. Please take a final look-through and adjust your reviews accordingly. Thanks!

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2198,27 +2211,22 @@ private void stageDeploymentFileForInitialize(Configuration config, DeploymentTy
...
+        if (config.m_userSchema == null) {
             return; // nothing to do
         }
         assert( config.m_userSchema.isFile() ); // this is validated during command line parsing and will be true unless disk faults
-        File stagedCatalogFH = new VoltFile(getStagedCatalogPath());
-        assert( !stagedCatalogFH.exists() || config.m_forceVoltdbCreate ); // managedPathsWithFiles() checks for the staged catalog
+        File stagedCatalogFH = new VoltFile(getStagedCatalogPath(getVoltDBRootPath()));
+
+        if (!config.m_forceVoltdbCreate && stagedCatalogFH.exists()) {
+            VoltDB.crashLocalVoltDB("A previous database was initialized with a schema. You must init with --force to overwrite the schema.");
```
hhakimi53: should this be: log an error message and do voltdb.exit(-1) instead of crash?

brians127: Sadly, this is consistent with 'voltdb init' behavior when other artifacts are found.

brians127: Upon further investigation - this should not be reachable under normal circumstances, because the initialized marker will be checked first.

PR #4426, opened 2017-04-07
Eng 11937 jdbc ssl

pdshaw:
Add SSL option to the JDBC benchmark test (Voter)


examples/voter/client/voter/JDBCBenchmark.java
```diff
@@ -115,6 +116,9 @@
         @Option(desc = "Number of concurrent threads synchronously calling procedures.")
         int threads = 40;
 
+        @Option(desc = "Enable SSL, Optionally provide configuration file.")
+        String sslfile = "";
```
hhakimi53: May need to revise the message here. The file path in-reference here is the path to the trustore. 
Also will be nice to have a switch so that the user can provide password to the supplied truststore.

PR #4436, opened 2017-04-12
Eng 12286 fix big batches for ipc

wweiss-voltdb:


src/ee/voltdbipc.cpp
```diff
@@ -817,9 +818,8 @@ void VoltDBIPC::executePlanFragments(struct ipc_command *cmd) {
     if (errors == 0) {
         // write the results array back across the wire
         const int32_t size = m_engine->getResultsSize();
-        char *resultBuffer = m_engine->getReusedResultBuffer();
-        resultBuffer[0] = kErrorCode_Success;
-        writeOrDie(m_fd, (unsigned char*)resultBuffer, size);
+        const unsigned char *resultBuffer = m_engine->getResultsBuffer();
```
yzhang1991: Here you replaced `getReusedResultBuffer()` with `getResultsBuffer()`.
There are other places in this file where we used `getReusedResultBuffer()`, do they need to be replaced?
Line 1238, 1586

tests/testprocs/org/voltdb_testprocs/regressionsuites/fallbackbuffers/SPMultiStatementQuery.java
```diff
@@ -42,18 +43,37 @@
     partitionInfo = "P1.NUM: 0"
 )
 
+/**
+ * This stored procedure accesses multiple sets of rows in separate results based on the
+ * a different modulo value for each SQL statement. Smaller modulo's will return more rows
+ * while large modulo's will return fewer rows. This procedure will verify that the returned
```
yzhang1991: "large" => "larger"

PR #4450, opened 2017-04-18
ENG-12204: Fix issue with EXISTS simplification

wolffcm:
I fixed an issue in simplifyExistsSubqueryStmt that caused a non-sensical statement to be generated: a groupless agg with passthrough columns.  This was causing the EE to crash.  The fix was to be less aggressive in simplifying exists subqueries.  In particular I removed code that attempted to do away with aggregation altogether.  There are ripple effects to what the old code was doing that seem really difficult to get a handle on.

I encapsulated ParsedSelectStmt.m_groupByColumns and m_distinctGroupByColumns in order to aid my debugging.

### DIFF
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -59,8 +59,8 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
     public ArrayList<ParsedColInfo> m_displayColumns = new ArrayList<>();
     private ArrayList<ParsedColInfo> m_orderColumns = new ArrayList<>();
     private AbstractExpression m_having = null;
-    public ArrayList<ParsedColInfo> m_groupByColumns = new ArrayList<>();
-    public ArrayList<ParsedColInfo> m_distinctGroupByColumns = null;
+    private List<ParsedColInfo> m_groupByColumns = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -1481,49 +1481,8 @@ protected AbstractExpression simplifyExistsSubqueryStmt(
...
-        }
-
-        if (m_aggResultColumns.isEmpty()) {
-            m_hasAggregateExpression = false;
-            m_hasAverage = false;
-        }
-
+        // Redo TVE placement since we've cleared the ORDER BY columns.
         placeTVEsinColumns();
-        needComplexAggregation();
```
BillWhite: Why is this dropped?  Is there anyplace where we explain what a complex aggregation is?  Something more than a real aggregation maybe?

wolffcm: The `needComplexAggregation()` call is used to set bits in the `ParsedSelectStmt` instance that describe if we do things like `MIN(f) - MAX(f)`, that is, expressions that are more than just an agg function.  This may affect how we generate output schema for the agg node.

This function is poorly named since it has the side effect of affecting state in the parsed stmt object, which is why its being called here.  

Since we're no longer modifying whether or not we do aggregation as part of, there's no longer a reason to call it.

PR #4452, opened 2017-04-18
stop kafka loader if any of it's consumer thread encounters unexpected condition

hhakimi53:


akhanzode:
I think you have to handle ClassNotFound in getConsumerExecutor also asl constructor of KafkaConsumer can throw it and executor should stop without leaving any hanging threads.

PR #4457, opened 2017-04-19
ENG-12331, check the version of VoltDB before voltadmin status command doing the work.

luyangco:


lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -25,9 +25,6 @@
 import os
 import sys
 
-RELEASE_MAJOR_VERSION = 7
-RELEASE_MINOR_VERSION = 2
```
xinjiacs: How does the paln_upgrade work after "RELEASE_MAJOR_VERSION"  removed? Not used before?

luyangco: Yeah, unused code. It was in the python side, however I moved it into NT proc.

PR #4465, opened 2017-04-20
Eng 11586 new ee gen tests

BillWhite:
Improve the C++ unit test generation procedure.

We use the actual catalog now, allow the use of integer and string data.  We also create generated tests on the fly, using the makefile.


build.py
```diff
@@ -498,12 +498,6 @@
     CTX.TESTS['executors'] = """
     OptimizedProjectorTest
     MergeReceiveExecutorTest
-    TestGeneratedPlans
```
BillWhite: These are now generated whenever the makefile is created.

tests/ee/executors/TestGeneratedPlans.cpp
```diff
@@ -1,900 +0,0 @@
-/* This file is part of VoltDB.
```
BillWhite: All of these files in tests/ee/executors/*.cpp are generated when the makefile is created.

tests/frontend/org/voltdb/planner/EEPlanTestGenerator.java
```diff
@@ -1,807 +0,0 @@
-/* This file is part of VoltDB.
```
BillWhite: This is all moved out to a subpackage.

tests/ee/test_utils/plan_testing_baseclass.h
```diff
@@ -288,14 +394,30 @@ class PlanTestingBaseClass : public Test {
...
-                int64_t v1 = voltdb::ValuePeeker::peekAsBigInt(tuple.getNValue(col));
-                VOLT_TRACE("Row %02d, col %02d: expected %04d, got %04ld (%s)",
-                           row, col,
-                           expected, v1,
-                           (expected != v1) ? "failed" : "ok");
-                if (expected != v1) {
-                    failed = true;
+                int32_t expected = answer->m_contents[row * nCols + col];
+                if (answer->m_types[col] == voltdb::VALUE_TYPE_VARCHAR) {
+                    const char *actualStr = voltdb::ValuePeeker::peekObjectValue(tuple.getNValue(col));
```
wolffcm: This char pointer is not null-terminated, I believe, so the call to `strcmp` below should trigger an invalid read if you run this in valgrind.

tests/ee/test_utils/plan_testing_config.h
```diff
@@ -55,11 +55,18 @@ struct TableConfig {
...
-    const       char        **m_columnNames;
+    const char               *m_tableName;
+    const char              **m_columnNames;
+    const voltdb::ValueType  *m_types;
+    const int32_t            *m_typeSizes;
     int                       m_numRows;
     int                       m_numCols;
+    // Some of the ints in the m_contents are
+    // really ints.  Some are offsets into the
+    // string table, which is m_strings.
```
wolffcm: I wonder if a more readable, concise and easy-to-extend way to manage this would be to use something like `boost::variant`?

buildtools.py
```diff
@@ -352,6 +391,49 @@ def buildThirdPartyTools(CTX, makefile):
...
+               os.path.join(rootdir, output_prefix)]
+    if not doGenerateFiles:
+        tcArgs += ["--names-only"]
+    for tc in testClasses:
+        tcArgs += ["--test-class", tc.strip()]
+    p = Popen(tcArgs, stdout=PIPE)
+    p.wait()
+    #
+    # Sometimes, in a pro build, the command above
+    # makes the folder obj/${BUILD}/testoutput.  We
```
BillWhite: I think there is something in setting up the LocalServer that makes this folder.  Then the Jenkins job tries to make the folder, which fails because it's already made.  I don't think it's expected that the product is invoked during the build phase, though it seems that we do it with this change.

tests/ee/test_utils/plan_testing_baseclass.h
BillWhite: Right.  I'll fix this.

tests/ee/test_utils/plan_testing_config.h
BillWhite: Well, this is mostly glue code which I didn't think anyone would really read that much, unless it's wrong.  The more interesting part would be inside the executors which are being tested out.  Getting the executors started is not so interesting.  Since the different cases are generated, I'm not sure it matters.

buildtools.py
```diff
@@ -581,7 +675,7 @@ def buildMakefile(CTX):
         makefile.write("#\n")
         makefile.write("########################################################################\n")
         makefile.write("%s: $(ROOTDIR)/%s | build-third-party-tools \n" % (objectname, sourcename))
-        makefile.write("\t$(CCACHE) $(COMPILE.cpp) -I$(ROOTDIR)/%s -MMD -MP -o $@ $(ROOTDIR)/%s\n" % (TEST_PREFIX, sourcename))
+        makefile.write("\t$(CCACHE) $(COMPILE.cpp) -I$(ROOTDIR)/${TEST_PREFIX} -MMD -MP -o $@ $(ROOTDIR)/%s\n" % sourcename)
```
BillWhite: Yes.  At one time they were constants, but at one point I wanted to use the same makefile to do something else.  So I converted the makefile-generation-time constants to be makefile-variables.  Then they can be changed at make time.  I think I wanted to change the log level, not this TEST_PREFIX variable.  But I changed a lot of them.

PR #4479, opened 2017-04-25
Patch for voltdbclient.py

ng-labo:
Using voltdbclient.py to call stored procedure, I noticed that I get a error in some case.

Usually we use java client program to call stored procedures in our system.
To make a test our stored procedure easy, I have tried to write a script by python.
I got error result sometimes, not always.
Compared to voltdb-client-cpp code, I got idea some sequence mismatch on reading VoltTable from buffer in voltdbclient.py

I summalized that I have faced to situation below.
- The stored procedure on this case, written by java, returns 4 table.
- In the error,the first table returns no record.
- The table's size on wire-protocol message is about more 500 byte by tracing.
- In the timing voltdbclient.py complete to read first table, the buffer pointer do not indicate next table header.
- (It seems no error when first table returns one record or more.)

By reference to voltdb-cpp-client code(InvocationResponse.hpp), I wrote a patch for voltdbclient.py to fix this situation.

### DIFF
--- a/lib/python/voltdbclient.py
+++ b/lib/python/voltdbclient.py
@@ -854,6 +854,7 @@ class VoltColumn:
     def __init__(self, fser = None, type = None, name = None):
         if fser != None:
             self.type = fser.readByte()
+            self.name = None
         elif type != None and name != None:
             self.type = type
             self.name = name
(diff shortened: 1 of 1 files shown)


akhanzode:
Thanks for the patch.

PR #4481, opened 2017-04-25
Eng 11992 support kerberos in python client

wweiss-voltdb:


lib/python/voltdbclient.py
```diff
@@ -271,7 +293,11 @@ def __init__(self, host = None, port = 21212, usessl = False, username = "",
 
         self.read_buffer = ReadBuffer()
 
-        if not username is None and not password is None and not host is None:
+        if self.usekerberos and kerberos_available and self.has_ticket():
+            assert not self.socket is None
+            self.socket.settimeout(connect_timeout)
+            self.authenticate(str(self.kerberosprinciple), "")
+        elif not username is None and not password is None and not host is None:
```
akhanzode: It looks like if you dont have ticket but kerberos is enabled you are authenticating with regular auth scheme, which is not right. You probably need nested ifs here.

lib/python/voltdbclient.py
```diff
@@ -408,6 +437,52 @@ def authenticate(self, username, password):
...
+                    self.writeByte(self.AUTH_HANDSHAKE_VERSION)
+                    self.writeByte(self.AUTH_HANDSHAKE)
+                    self.wbuf.extend(out_token)
+                    self.prependLength()
+                    self.flush()
+
+                    try:
+                        self.bufferForRead()
+                    except IOError, e:
+                        print "ERROR: Connection failed. Please check that the host, port and ssl settings are correct."
```
akhanzode: do you allow kerberos over ssl? if not this message may not be correct.

lib/python/voltdbclient.py
```diff
@@ -408,6 +437,52 @@ def authenticate(self, username, password):
...
+                    if version != self.AUTH_HANDSHAKE_VERSION or status != self.AUTH_HANDSHAKE:
+                        raise RuntimeError("Authentication failed.")
+
+                    in_token = self.readVarbinaryContent(self.read_buffer.remaining()).tostring()
+                    out_token = ctx.step(in_token)
+
+                try:
+                    self.bufferForRead()
+                except IOError, e:
+                    print "ERROR: Connection failed. Please check that the host, port and ssl settings are correct."
```
akhanzode: Same as above.

lib/python/voltdbclient.py
```diff
@@ -418,6 +493,23 @@ def authenticate(self, username, password):
         for x in range(self.readInt32()):
             self.readByte()
 
+    def has_ticket(self):
```
akhanzode: Is has_ticket actually getting a ticket? if yes may be call it get_ticket()

lib/python/voltdbclient.py
```diff
@@ -418,6 +493,23 @@ def authenticate(self, username, password):
...
+            default_cred = gssapi.creds.Credentials(usage='initiate')
+            if default_cred.lifetime > 0:
+                self.kerberosprinciple = str(default_cred.name)
+                retval = True
+            else:
+                print "ERROR: Kerberos principal found but login expired."
+        except gssapi.raw.misc.GSSError as e:
+            print "ERROR: unable to find default principal from Kerberos cache."
+        return retval
+
```
akhanzode: This voltdb client and the client in seperate repo should match. Ask Ruth she was planning to merge them together and get rid of duplicate.

lib/python/voltdbclient.py
wweiss-voltdb: No. It is verifying that the user has an active ticket with kerberos (that the user is already performed single sign-on)

lib/python/voltdbclient.py
wweiss-voltdb: I did not update the voltdb-client-python repo because it does not have SSL either. If we do one, we should do both.

lib/python/voltdbclient.py
wweiss-voltdb: no. but is is common code so host and port are always correct and ssl is only applicable when enabled. Before SSL, the error was "check that the host and port settings are correct"

lib/python/voltdbclient.py
wweiss-voltdb: done

PR #4483, opened 2017-04-26
sql_coverage_test.py: better diagnostics to go with the "Unable to

johnTcrawford:
connect/create client" error message, to make it easier to track down
problems like ENG-12319 (which was a ports problem - the 3021 and 7181
ports on the volt12i-U14.04 Jenkins machine were being used by an old
process).

### DIFF
--- a/tests/sqlcoverage/sql_coverage_test.py
+++ b/tests/sqlcoverage/sql_coverage_test.py
@@ -129,17 +129,23 @@ def run_once(name, command, statements_path, results_path,
         server = subprocess.Popen(command + " backend=" + name, shell=True)
 
     client = None
+    clientException = None
     for i in xrange(30):
         try:
             client = VoltQueryClient(host, port)
             client.set_quiet(True)
             client.set_timeout(5.0) # 5 seconds
             break
-        except socket.error:
+        except socket.error as e:
+            clientException = e
             time.sleep(1)
 
     if client == None:
-        print >> sys.stderr, "Unable to connect/create client"
+        print >> sys.stderr, "Unable to connect/create client: there may be a problem with the VoltDB server or its ports:"
+        print >> sys.stderr, "name:", str(name)
+        print >> sys.stderr, "host:", str(host)
+        print >> sys.stderr, "port:", str(port)
+        print >> sys.stderr, "client (socket.error) exception:", str(clientException)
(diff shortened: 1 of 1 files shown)


tests/sqlcoverage/sql_coverage_test.py
```diff
@@ -129,17 +129,23 @@ def run_once(name, command, statements_path, results_path,
...
     client = None
+    clientException = None
     for i in xrange(30):
         try:
             client = VoltQueryClient(host, port)
             client.set_quiet(True)
             client.set_timeout(5.0) # 5 seconds
             break
-        except socket.error:
+        except socket.error as e:
```
wolffcm: Are there other exceptions that you might want to catch here so you can print out a diagnostic?

tests/sqlcoverage/sql_coverage_test.py
```diff
@@ -129,17 +129,23 @@ def run_once(name, command, statements_path, results_path,
...
+            clientException = e
             time.sleep(1)
 
     if client == None:
-        print >> sys.stderr, "Unable to connect/create client"
+        print >> sys.stderr, "Unable to connect/create client: there may be a problem with the VoltDB server or its ports:"
+        print >> sys.stderr, "name:", str(name)
+        print >> sys.stderr, "host:", str(host)
+        print >> sys.stderr, "port:", str(port)
+        print >> sys.stderr, "client (socket.error) exception:", str(clientException)
```
wolffcm: I guess that str(None) does something reasonable? I don't know enough python too be sure.

tests/sqlcoverage/sql_coverage_test.py
johnTcrawford: No, this exception is expected while the VoltDB server starts up (for which we allow up to 30 seconds), which is why we catch it; we only care about it if it does NOT go away within 30 seconds. Any other exception is unexpected, and will (and should) kill the test.

tests/sqlcoverage/sql_coverage_test.py
johnTcrawford: Yes, it returns "None" - I tested this (by temporarily changing the previous loop to go for 0 seconds rather than 30), but it's a good question.

PR #4484, opened 2017-04-26
ENG-12375 fix vmc with staged schema in 7.2.x

brians127:
Make sure catalog-report.html always ends up in catalog.jar since VMC needs it there.

### DIFF
--- a/src/frontend/org/voltdb/compiler/VoltCompiler.java
+++ b/src/frontend/org/voltdb/compiler/VoltCompiler.java
@@ -156,7 +156,6 @@ public class VoltCompiler {
     private final HashSet<Class<?>> m_cachedAddedClasses = new HashSet<>();
 
     private final boolean m_isXDCR;
-    boolean m_generateReports;
 
     /**
      * Represents output from a compile. This works similarly to Log4j; there
@@ -334,13 +333,8 @@ public class VoltCompiler {
     }
 
     public VoltCompiler(boolean standaloneCompiler, boolean isXDCR) {
-        this(standaloneCompiler, isXDCR, true);
-    }
-
-    public VoltCompiler(boolean standaloneCompiler, boolean isXDCR, boolean generateReports) {
         this.standaloneCompiler = standaloneCompiler;
         this.m_isXDCR = isXDCR;
-        this.m_generateReports = generateReports || DEBUG_MODE;
 
         // reset the cache
         m_cachedAddedClasses.clear();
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -375,12 +369,12 @@ void addErr(final String msg) {
     void addInfo(final String msg, final int lineNo) {
         final Feedback fb = new Feedback(Severity.INFORMATIONAL, msg, m_currentFilename, lineNo);
         m_infos.add(fb);
-        if (standaloneCompiler) {
+        //if (standaloneCompiler) {
             compilerLog.info(fb.getLogString());
```
brians127: I'm uncertain about what to do here, but it doesn't affect the fix much. In my mind it doesn't make sense to pass around an "if staged catalog use case" boolean just for INFO compiler notifications, but I don't want to remove these notices from staged catalog compilation. I'm partial to including them in the log but open to discussion.

akhanzode:
Add test that makes sure that the junit covers the case of existence of catalog-report.html

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -375,12 +369,12 @@ void addErr(final String msg) {
     void addInfo(final String msg, final int lineNo) {
         final Feedback fb = new Feedback(Severity.INFORMATIONAL, msg, m_currentFilename, lineNo);
         m_infos.add(fb);
-        if (standaloneCompiler) {
+        //if (standaloneCompiler) {
```
akhanzode: Instead of commented code can you remove it.

tests/frontend/org/voltdb/regressionsuites/TestStartWithSchema.java
```diff
@@ -71,17 +72,30 @@
     static final String mismatchSchema =
             "create table TEST (myval bigint not null, PRIMARY KEY(myval));";
 
+    static void ensureAllCatalogDefaultArtifactsExists(InMemoryJarfile jarFile, String nameForDebugging) {
+        Set<String> files = jarFile.keySet();
+        assertTrue(nameForDebugging + " is empty!", files.size() > 0);
+        for (String artifact : CatalogUtil.CATALOG_DEFAULT_ARTIFACTS) {
+            // NOTE: autogen-ddl.sql IS required, as it contains the schema
+            if (!artifact.equals(CatalogUtil.CATALOG_EMPTY_DDL_FILENAME)) {
```
akhanzode: Instead of putting comment about autoget-ddl it should say staged catalog does not contain ddl.sql.

PR #4485, opened 2017-04-27
ENG-12118: repeat('', <large number>) makes EE needlessly pause

wolffcm:


tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -2696,6 +2696,12 @@ public void testRepeat() throws NoConnectionsException, IOException, ProcCallExc
                 expectedError,
                 "REPEAT", 4611686018427387903L, 1);
         }
+
+        // Make sure that repeating an empty string a lot of times completes
+        // in a reasonable time.  It's tough to validate execution in a JUnit test,
+        // but trust me this would take a long time if the EE was being naive.
+        cr = client.callProcedure("@AdHoc", "select repeat('', 100000000000) from P1 limit 1");
+        assertContentOfTable(new Object[][] {{""}}, cr.getResults()[0]);
```
yzhang1991: But that just makes the test take much longer to finish?
If it does not fail the test we may not notice it.

BillWhite:
Looks good to me as well.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
wolffcm: Yes, that's true.  I could make it fail if it takes longer than say a minute?  Would that be long enough?

yzhang1991: Well, I would expect that with your optimization, this query can finish with no time right?
One minute may still be much longer than needed.

PR #4486, opened 2017-04-28
Update CSVLoader.java

cloudscope2015:
Add characterSet to support the import of multi-character set files, using the system character set by default.
Example:
csvloader TargetTable --separator='|' --characterSet='gb2312' -f TargetTable.unl


cloudscope2015:
merge into master

PR #4488, opened 2017-04-28
ENG-12399 Fix TestSQLFeaturesNewSuite

yzhang1991:
It is indeed related to the regression suite optimization.
There is one test case that uses "ALTER TABLE" statements to change the constraint.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestSQLFeaturesNewSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestSQLFeaturesNewSuite.java
@@ -26,8 +26,6 @@ package org.voltdb.regressionsuites;
 import java.io.IOException;
 import java.util.UUID;
 
-import junit.framework.Test;
-
 import org.voltdb.BackendTarget;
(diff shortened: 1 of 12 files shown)


tests/frontend/org/voltdb/regressionsuites/TestSQLFeaturesNewSuite.java
```diff
@@ -475,6 +475,14 @@ public void testAlterTableLimitPartitionRows() throws IOException, ProcCallExcep
 
         vt = client.callProcedure("@AdHoc", "select dept from capped3_limit_exec_complex order by dept asc").getResults()[0];
         validateTableOfScalarLongs(vt, new long[] {5, 6, 8});
+
```
wolffcm: Looks good.  I thought it might be something like this.  Thanks for fixing it.

I wonder if there's a way to make tests fail when they leave the catalog in a different state than when a test method started?  Just a thought.

wolffcm:
Latest commits look good too.  It will be interesting to see how many tests fail with this change in.

PR #4489, opened 2017-05-01
Eng 12207

hhakimi53:
Add support to Kafkaloader via new utility kafkaloader10 to use Kafka 10.2.1 API. Kafkloader10 uses consumer from kafka client 10.2.1 to ingest data from Kafka cluster. Kafkaloader uses default consumer properties for most part except for auto commit reset, which it sets to earliest (similar to approach taken by KafkaLoader based on Kafka 8 API). KafkaLoader is hard wired to use auto commit feature and byte array deserializer when fetching key/value pair from consumer record. In case if user supplied consumer property are different than the hard wired values, those values are overriden by the one that KafkaLoader uses. 

Kafkaloader based on Kafka API 8 is still supported and can be invoked through kafkaloader8 instead of kafkaloader


build-client.xml
```diff
@@ -72,9 +72,12 @@
     <fileset dir='${vendor.lib.dir}'>
         <include name='javaee-api-6.0.jar' />
     </fileset>
-	<fileset dir='${vendor.lib.dir}'>
+    <fileset dir='${vendor.lib.dir}'>
         <include name='javassist-3.20.0-GA.jar' />
     </fileset>
+    <fileset dir='${vendor.lib.dir}'>
```
akhanzode: You dont need to include this in build-client.xml

build-client.xml
```diff
@@ -174,6 +177,8 @@
                   "
 />
 
+<!--org/voltdb/utils/KafkaLoader10.java -->
```
akhanzode: You dont need this.

build-importers.xml
```diff
@@ -71,7 +71,7 @@
             <zipgroupfileset dir="${base.dir}/lib/" includes="scala-parser-combinators_2.11-1.0.2.jar" />
             <zipgroupfileset dir="${base.dir}/lib/" includes="scala-xml_2.11-1.0.2.jar" />
             <zipgroupfileset dir="${base.dir}/lib" includes="kafka_2.11-0.8.2.2.jar" />
-            <zipgroupfileset dir="${base.dir}/lib/" includes="kafka-clients-0.8.2.2.jar" />
+            <zipgroupfileset dir="${base.dir}/lib/" includes="kafka-clients-0.10.1.1.jar" />
```
akhanzode: Why do you need kafka 10 here in existing bundle. kafka 8 bundle should remain same as before.

build-importers.xml
```diff
@@ -118,6 +118,16 @@
 	        </jar>
 	 </target>
 
+    <target name="KafkaLoader10">
+        <!-- regular build will compile the required classes -->
+        <mkdir dir="${bundles.dir}" />
+        <jar destfile="${bundles.dir}/kafkaLoader10.jar" basedir="${build.prod.dir}">
+            <include name="org/voltdb/utils/KafkaLoader10*.class"/>
```
akhanzode: You should keep this bundle clean of any loader or CSVParser code. It should be possible to use this bundle in importer. Dont you need snappy/compression related jars here?

build-client.xml
hhakimi53: Yes, forgot to clean the client build file. I will clean it up Thanks for pointing out.

build-client.xml
hhakimi53: ditto

build-importers.xml
hhakimi53: Done. The bundle will be called kafkastream10 and will have only kafka client jars based on 10 api and compression jar files

build-importers.xml
hhakimi53: yes, it should be 8 and not 10 and 10 is not needed. thnxs

src/frontend/org/voltdb/utils/KafkaLoader.java
```diff
@@ -120,7 +120,7 @@ public void processKafkaMessages() throws Exception {
         m_loader.setFlushInterval(m_config.flush, m_config.flush);
         m_consumer = new KafkaConsumerConnector(m_config);
         try {
-            m_es = getConsumerExecutor(m_config, m_consumer, m_loader);
+            m_es = getConsumerExecutor(m_consumer, m_loader);
```
brians127: "es" is not a very descriptive name. This is optional since you didn't introduce the variable, but I prefer names like "m_consumerExecutor" that make it clear at a glance what the variable does.

build.xml
```diff
@@ -1408,7 +1421,8 @@ JAVA COMPILATION
                   io/netty_voltpatches/**/*.java
                   org/jctools_voltpatches/**/*.java"/>
     <!-- non-voltcore voltdb code -->
-    <invoke-javac srcdir="${src.gpl.dir}" excludes="org/voltcore/**/*.java" />
+    <invoke-javac srcdir="${src.gpl.dir}" excludes="org/voltcore/**/*.java
+                                                    org/voltdb/utils/KafkaLoader10.java" />
```
brians127: Why do you need to do this?

build.xml
```diff
@@ -126,7 +126,7 @@
 
 <!-- Overridden in the Hudson test script. -->
 <property name='junit.haltonfailure'    value='false' />
-<property name="j2se_api" value="http://docs.oracle.com/javase/7/docs/api/"/>
+<property name="j2se_api" value="http://docs.oracle.com/javase/8/docs/api/"/>
```
hhakimi53: For server Java 8 is needed and the change above is correcting the doc link. What can be the side effect of it?

build.xml
hhakimi53: kafkaloader10 needs to build with kafka 10 apis and can't be build with kafka 8 apis, which is what build.xml uses to build the subsystem like export which requires kafka 8 apis. kafkaloader10 is compiled using build-kafkaloader.xml and package in build in build-importer.xml

bin/kafkaloader10
```diff
@@ -0,0 +1,149 @@
...
+# add libs to CLASSPATH
+for f in $VOLTDB_LIB/*.jar; do
+  CLASSPATH=${CLASSPATH}:$f;
+done
+
+# Add formatter lib if using --formatter option.
+for f in $FORMATTER_LIB/*.jar; do
+  CLASSPATH=${CLASSPATH}:$f;
+done
+
```
akhanzode: Kafkaloader 10 does not require ZKLIB? if not thats good and we should document it.

build-importers.xml
```diff
@@ -118,6 +118,16 @@
 	        </jar>
 	 </target>
 
+    <target name="kafkastream10">
+        <!-- regular build will compile the required classes -->
+        <mkdir dir="${bundles.dir}" />
+        <jar destfile="${bundles.dir}/kafkastream10.jar" basedir="${build.prod.dir}">
+            <zipgroupfileset dir="${base.dir}/third_party/java/jars" includes="kafka-clients-0.10.2.1.jar" />
```
akhanzode: Check with Ruth and update README.thirdparty

src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -0,0 +1,517 @@
...
+        }
+    }
+
+    private Properties kafkaConfigProperties() throws IOException {
+        //Get group id which should be unique for table so as to keep offsets clean for multiple runs.
+        String groupId = "voltdb-" + (m_cliOptions.useSuppliedProcedure ? m_cliOptions.procedure : m_cliOptions.table);
+
+        Properties props = new Properties();
+        if (m_cliOptions.config.trim().isEmpty()) {
+            props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
akhanzode: Does this get picked and overridden by consumer properties file?

src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -0,0 +1,517 @@
...
+                + Client.VOLTDB_SERVER_PORT + ")")
+        String servers = "localhost:" + Client.VOLTDB_SERVER_PORT;
+
+        @Option(desc = "Username when connecting to the database")
+        String user = "";
+
+        @Option(desc = "Password to use when connecting to database")
+        String password = "";
+
+        @Option(shortOpt = "b", desc = "Kafka bootstrap server / broker to connect to. (format: broker:port)")
```
akhanzode: This is also a comma separated list right.

bin/kafkaloader10
hhakimi53: No, kafka loader 10 does not require zookeeper library to be in classpath. We have documented this in our UX document and Andrew is aware of this.  Will check with him on the same so that it's documented.

build-importers.xml
hhakimi53: Sure, thnxs.  Kafka10 API are under Apache 2 license

src/frontend/org/voltdb/utils/KafkaLoader10.java
hhakimi53: yes

src/frontend/org/voltdb/utils/KafkaLoader10.java
hhakimi53: Yes, if the property is in supplied consumer property file, we will honor that. If it's not present, the earliest policy for offset reset policy is used.

bin/kafkaloader10
```diff
@@ -0,0 +1,149 @@
...
+if [ -d $VOLTDB_HOME/../../lib/voltdb ]; then
+    # Linux package installation puts all libraries in /usr/lib/voltdb
+    VOLTDB_VOLTDB=$(dirname $(dirname "$VOLTDB_HOME"))/lib/voltdb
+    VOLTDB_LIB=$VOLTDB_VOLTDB
+	BUNDLES_DIR=$(dirname $(dirname "$VOLTDB_HOME"))/bundles
+else
+    # distribution puts third party jars in lib and voltdb libraries in voltdb
+    VOLTDB_VOLTDB=$VOLTDB_HOME/voltdb
+    VOLTDB_LIB=$VOLTDB_HOME/lib
+	BUNDLES_DIR=$VOLTDB_HOME/bundles
```
brians127: Please use spaces instead of tab

bin/kafkaloader10
```diff
@@ -0,0 +1,149 @@
...
+popd > /dev/null
+# the root of the VoltDB installation
+export VOLTDB_HOME=$(dirname $(dirname "$this"))
+
+# detect and adjust for Linux package installation vs. VoltDB distribution
+if [ -d $VOLTDB_HOME/../../lib/voltdb ]; then
+    # Linux package installation puts all libraries in /usr/lib/voltdb
+    VOLTDB_VOLTDB=$(dirname $(dirname "$VOLTDB_HOME"))/lib/voltdb
+    VOLTDB_LIB=$VOLTDB_VOLTDB
+	BUNDLES_DIR=$(dirname $(dirname "$VOLTDB_HOME"))/bundles
```
brians127: Please use spaces instead of tab

bin/kafkaloader10
```diff
@@ -0,0 +1,149 @@
...
+  fi
+fi
+
+# add kakfka jar file to classpath
+KAFKA_LOADER10_BUNDLE=${BUNDLES_DIR}/kafkastream10.jar
+if [ ! -f ${KAFKA_LOADER10_BUNDLE} ]; then
+  echo "Couldn't find kafka jar file."
+  exit 1
+fi
+CLASSPATH=${KAFKA_LOADER10_BUNDLE}:${CLASSPATH}
```
brians127: In kafkaloader8 the CLASSPATH from the user's environment is ignored. Here it is being used, though VoltDB is being searched first. Is this change intentional? I don't think it is a problem, but want to avoid unintended side effects.

build.xml
brians127: I didn't realize this was just for documentation. Change is fine. Thanks for doing the cleanup.

src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -0,0 +1,525 @@
...
+        String servers = "localhost:" + Client.VOLTDB_SERVER_PORT;
+
+        @Option(desc = "Username when connecting to the database")
+        String user = "";
+
+        @Option(desc = "Password to use when connecting to database")
+        String password = "";
+
+        @Option(shortOpt = "b", desc = "Comma separated list of kafka broker(s) to connect to. (format: broker:port)")
+        String brokers = ""; //No default here as default will clash with local voltdb cluster
```
hhakimi53: Yes

build.xml
brians127: OK. Thanks for the explanation. I was confused by the Ant syntax "src.gpl.dir"

src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -0,0 +1,525 @@
...
+            groupId = props.getProperty("group.id", groupId);
+
+            // get kafka broker connections from properties file if present - supplied brokers, if any, overrides
+            // the supplied command line
+            m_cliOptions.brokers = props.getProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, m_cliOptions.brokers);
+
+            String autoCommit = props.getProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG);
+            if (autoCommit != null && !autoCommit.trim().isEmpty() &&
+                    !("true".equals(autoCommit.trim().toLowerCase())) ) {
+                m_log.warn("Auto commit policy for Kafka loader will be set to \'true\' instead of \'" + autoCommit +"\'");
```
hhakimi53: Logic operates with consumer's auto-commit policy to be true and does not do manual offset management. If user supplied consumer config has auto commit policy set to false, it is not supported by the logic - warning message is logged and the property is overriden to true.

src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -0,0 +1,525 @@
...
+            // Only byte array are used by Kafka Loader. If there are any deserializer supplied in config file
+            // log warning message about it
+            String deserializer = props.getProperty(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG);
+            if (deserializer != null && KEY_DESERIALIZER.equals(deserializer.trim()) ) {
+                m_log.warn("Key deserializer \'" + deserializer.trim() + "\' not supported. \'"
+                        + KEY_DESERIALIZER + "\' will be used for deserializering keys");
+            }
+            deserializer = props.getProperty(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG);
+            if ( deserializer != null && VALUE_DESERIALIZER.equals(deserializer.trim())) {
+                m_log.warn("Value deserializer \'" + deserializer.trim() + "\' not supported. \'"
```
hhakimi53: sure

src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -0,0 +1,525 @@
...
+
+        }
+
+        @Override
+        public void run() {
+            String smsg = null;
+            try {
+                m_consumer.subscribe(Arrays.asList(m_config.topic));
+                while (!m_closed.get()) {
+                    ConsumerRecords<byte[], byte[]> records = m_consumer.poll(1000); // 1 second
```
hhakimi53: That's the amount of time cluster will wait before ti respond back with zero records if no data is available. I will add hidden property so that it's tweakable

bin/kafkaloader10
hhakimi53: yes, I clean that up. thanks for pointing it out.

bin/kafkaloader10
hhakimi53: sure

PR #4490, opened 2017-05-01
ENG-12246 UDF validation

yzhang1991:


yzhang1991:
Will open a pull request against master.

PR #4494, opened 2017-05-03
ENG-12246 Validate user-defined functions

yzhang1991:


src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
```diff
@@ -60,6 +94,88 @@ protected boolean processStatement(DDLStatement ddlStatement, Database db, DdlPr
...
+                    // if not null, then we've got more than one run method
+                    if (functionMethod != null) {
+                        String msg = "Class " + shortName + " has multiple methods named " + methodName;
+                        msg += ". Only a single function method is supported.";
+                        throw m_compiler.new VoltCompilerException(msg);
+                    }
+                    functionMethod = m;
+                }
+                else {
+                    m_compiler.addWarn(warningMessage.toString());
```
wolffcm: So you only warn for void, static or non-public methods if a compatible method is not found?  You might want to warn even if a compatible method *is* found?

src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
```diff
@@ -60,6 +94,88 @@ protected boolean processStatement(DDLStatement ddlStatement, Database db, DdlPr
...
+                }
+                if (Modifier.isStatic(m.getModifiers())) {
+                    warningMessage.append("static ");
+                    found = false;
+                }
+                if (m.getReturnType().equals(Void.TYPE)) {
+                    warningMessage.append("void ");
+                    found = false;
+                }
+                warningMessage.append(methodName);
```
wolffcm: Any other modifiers you should be checking for here, `abstract` maybe?

src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
wolffcm: Wait, never mind.  This looks good.

PR #4495, opened 2017-05-03
{TESTS RUNNING} Fix Kafka Importer occasionally losing small numbers of rows upon pause

brians127:
This fixes a race between the VoltDB procedure callback threads trying to set the paused offset. It was reliably set to the FIRST value seen, but not reliably set to the LOWEST value seen across all threads.

I introduce a synchronization mechanism to ensure the paused offset is reliable before using it to commit any offsets to Kafka.

Some design decisions behind the PendingWorkTracker object, which should be reusable:
* Routine adjustment of pending callback counters is lockless and fast
* Producer needs to reliably know when all callbacks are done without periodic sleeps
* Need to be able to get the number of callbacks called at the end of the run, regardless of whether or not waitForWorkToFinish() has been called.

PR #4497, opened 2017-05-04
ENG-12348: fix KafkaImporter occasionally losing small numbers of rows when server transitions to unavailable

brians127:
{ SAME FIX AS https://github.com/VoltDB/voltdb/pull/4495 BUT JIRA TICKET WAS CORRECTED }

This fixes a race between the VoltDB procedure callback threads trying to set the paused offset. It was reliably set to the FIRST value seen, but not reliably set to the LOWEST value seen across all threads.

I introduce a synchronization mechanism to ensure the paused offset is reliable before using it to commit any offsets to Kafka.

Some design decisions behind the PendingWorkTracker object, which should be reusable:
* Routine adjustment of pending callback counters is lockless and fast
* Producer needs to reliably know when all callbacks are done without periodic sleeps
* Need to be able to get the number of callbacks called at the end of the run, regardless of whether or not waitForWorkToFinish() has been called. This is why there are two counters that count up instead of a single counter which must be 0 at the end.

Performance wise I expect this code to be, during normal operation, indistinguishable from the prior revision. Shutdown due to a pause will definitely be slower - and this depends on how long it takes for the stored procedures to finish.

### DIFF
--- a/src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
+++ b/src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.LongAdder;
 import java.util.function.LongBinaryOperator;
 
 import org.voltcore.logging.Level;
@@ -76,6 +77,7 @@ public class KafkaTopicPartitionImporter extends AbstractImporter
     private final AtomicBoolean m_dead = new AtomicBoolean(false);
     //Start with invalid so consumer will fetch it.
     private final AtomicLong m_currentOffset = new AtomicLong(-1);
+    // CAUTION: m_pauseOffset is not reliable until all callbacks have completed.
     private final AtomicLong m_pauseOffset = new AtomicLong(-1);
     private long m_lastCommittedOffset = -1;
     private final AtomicReference<BlockingChannel> m_offsetManager = new AtomicReference<>();
@@ -375,8 +377,7 @@ public class KafkaTopicPartitionImporter extends AbstractImporter
     protected void accept() {
         info(null, "Starting partition fetcher for " + m_topicAndPartition);
         long submitCount = 0;
-        AtomicLong cbcnt = new AtomicLong(0);
-        @SuppressWarnings("unchecked")
+        PendingWorkTracker callbackTracker = new PendingWorkTracker();
         Formatter formatter = m_config.getFormatterBuilder().create();
         try {
             //Start with the starting leader.
@@ -465,14 +466,18 @@ public class KafkaTopicPartitionImporter extends AbstractImporter
                         params = formatter.transform(payload);
                         Invocation invocation = new Invocation(m_config.getProcedure(), params);
                         TopicPartitionInvocationCallback cb = new TopicPartitionInvocationCallback(messageAndOffset.offset(),
-                                messageAndOffset.nextOffset(), cbcnt, m_gapTracker, m_dead, m_pauseOffset);
-                         if (!noTransaction && !callProcedure(invocation, cb)) {
-                              if (isDebugEnabled()) {
-                                 debug(null, "Failed to process Invocation possibly bad data: " + Arrays.toString(params));
-                              }
-                              m_gapTracker.commit(messageAndOffset.nextOffset());
+                                messageAndOffset.nextOffset(), callbackTracker, m_gapTracker, m_dead, m_pauseOffset);
+                        if (!noTransaction) {
+                            if (callProcedure(invocation, cb)) {
+                                callbackTracker.produceWork();
+                            } else {
+                                if (isDebugEnabled()) {
+                                    debug(null, "Failed to process Invocation possibly bad data: " + Arrays.toString(params));
+                                }
+                                m_gapTracker.commit(messageAndOffset.nextOffset());
+                            }
                          }
-                     } catch (FormatException e) {
+                    } catch (FormatException e) {
                         rateLimitedLog(Level.WARN, e, "Failed to tranform data: %s" , Arrays.toString(params));
                         m_gapTracker.commit(messageAndOffset.nextOffset());
                     }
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -687,6 +699,47 @@ public synchronized long commit(long offset) {
...
+        private volatile boolean m_doneProducing = false;
+
+        public void produceWork() {
+            m_workProduced++;
+        }
+
+        public void consumeWork() {
+            // The very last consumer is responsible for waking the producer
+            long workConsumed = m_workConsumed.incrementAndGet();
+            if (m_doneProducing && m_workProduced == workConsumed) {
```
akhanzode: if (m_doneProducing && m_workConsumed.incrementAndGet() == workConsumer) no need for local variable.

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -687,6 +699,47 @@ public synchronized long commit(long offset) {
...
+
+        public synchronized void waitForWorkToFinish() {
+            m_doneProducing = true;
+            // only one consumer calls notify(), but loop just to safeguard against being interrupted
+            while (m_workProduced != m_workConsumed.longValue()) {
+                try {
+                    wait();
+                } catch (InterruptedException unexpected) {
+                }
+                assert m_workProduced == m_workConsumed.longValue();
```
akhanzode: Why assert?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -687,6 +699,47 @@ public synchronized long commit(long offset) {
...
+                }
+            }
+        }
+
+        public synchronized void waitForWorkToFinish() {
+            m_doneProducing = true;
+            // only one consumer calls notify(), but loop just to safeguard against being interrupted
+            while (m_workProduced != m_workConsumed.longValue()) {
+                try {
+                    wait();
```
akhanzode: There should be timed wait in case if lost response after Pause on other node. We dont want to be stuck in while loop forever.

brians127: How do I handle the timer expiration? If I commit a paused offset that is incorrect, that is a data loss event. I think the best choice is to skip committing entirely - forcing resume to re-import some of the same data is better than a node crash or missing data. Also, should the timeout be set to the same value as our stored procedure run time limit?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
brians127: This condition is guaranteed to be true after the callback has called notify(). I will move the assert so it can't be triggered by a spurious interrupt.

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
akhanzode: timeout does not need to be procedure timeout. You can pick 60 sec default with hidden property. If you get timeout dont skip commit if paused offset is smaller than last commit offset so you will rewind a bit otherwise the offset in future is already committed so you have lost the transaction.

brians127: OK. I'll add an ERROR message alerting the user to the timeout. Or should this be a WARNING like the gap tracker jumping ahead?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
brians127: That won't work because m_workConsumed.incrementAndGet() increments the counter, and that must be done regardless.

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -494,14 +499,21 @@ protected void accept() {
...
                 }
                 if (shouldCommit()) {
-                    commitOffset();
+                    long pausedOffset = -1; // ignores paused offset since we don't want to wait for all callbacks to complete
+                    commitOffset(pausedOffset);
                 }
             }
         } catch (Exception ex) {
             error(ex, "Failed to start topic partition fetcher for " + m_topicAndPartition);
+            ex.printStackTrace();
```
hhakimi53: this can be removed

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -494,14 +499,21 @@ protected void accept() {
...
         } finally {
-            //Dont care about return as it wil force a commit.
-            commitOffset();
+            long pausedOffset = m_pauseOffset.get();
+            if (pausedOffset != -1) {
+                callbackTracker.waitForWorkToFinish();
+                pausedOffset = m_pauseOffset.get();
+            }
+            // Force a commit.
+            commitOffset(pausedOffset);
```
hhakimi53: How about:
```
            if (m_pauseOffset.get() != -1) {
                callbackTracker.waitForWorkToFinish();
            }
            // Force a commit.
            commitOffset(m_pauseOffset.get()); 
```

Still thinking about the logic for should commit in the loop where the pause is set to -1 to skip the pause mechanism in commitoffset()

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
hhakimi53: Do you want assert for debugging or ... ? Assert may be disabled for production release builds. In production, for server side, would you want to assert on the server side for such condition?

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
akhanzode: Add it as WARN and not error

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
brians127: The assert is a debugging and testing aid. It is not needed for production.

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -27,6 +27,7 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.LongAdder;
```
brians127: I was going to do that, but due to a design change it is now being used.

src/frontend/org/voltdb/importclient/kafka/KafkaTopicPartitionImporter.java
```diff
@@ -494,14 +502,29 @@ protected void accept() {
...
         } finally {
-            //Dont care about return as it wil force a commit.
-            commitOffset();
+            final boolean usePausedOffset = m_pauseOffset.get() != -1;
+            boolean skipCommit = false;
+            if (usePausedOffset) {
+                // Paused offset is not guaranteed reliable until all callbacks have been called.
+                if (callbackTracker.waitForWorkToFinish() == false) {
+                    if (m_pauseOffset.get() < m_lastCommittedOffset) {
+                        warn(null, "Committing paused offset even though a timeout occurred waiting for pending stored procedures to finish.");
```
brians127: Do these warnings need to get run by Andrew? I think they are analogous to the "gap tracker overflow" condition - internal details which are intended to explain loss of data in KafkaImporter.

PR #4499, opened 2017-05-05
ENG-12280 improvement of partition master assignment  with missing node startup

zeemanhe:
assign masters more evenly among the nodes during cluster startup with missing nodes.

### DIFF
--- a/src/frontend/org/voltdb/AbstractTopology.java
+++ b/src/frontend/org/voltdb/AbstractTopology.java
@@ -43,6 +43,7 @@ import com.google_voltpatches.common.collect.ImmutableMap;
 import com.google_voltpatches.common.collect.ImmutableSet;
 import com.google_voltpatches.common.collect.ImmutableSortedSet;
 import com.google_voltpatches.common.collect.LinkedListMultimap;
+import com.google_voltpatches.common.collect.Lists;
 import com.google_voltpatches.common.collect.Maps;
 import com.google_voltpatches.common.collect.Multimap;
 import com.google_voltpatches.common.collect.Sets;
@@ -81,7 +82,7 @@ public class AbstractTopology {
     public static class Partition implements Comparable<Partition>{
         public final int id;
         public final int k;
-        public final int leaderHostId;
+        public int leaderHostId;
         public final ImmutableSortedSet<Integer> hostIds;
 
         private Partition(int id, int k, int leaderHostId, Collection<Integer> hostIds) {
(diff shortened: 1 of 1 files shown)


zeemanhe:
Unit test on pro: https://github.com/VoltDB/pro/pull/1881

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1403,17 +1404,32 @@ public static AbstractTopology shiftPartitionLeaders(AbstractTopology topology,
                 mutableHost.partitions.add(mp);
             }
             int leaderId = partition.leaderHostId;
-            if (missingHosts.contains(partition.leaderHostId)) {
-                //find the first non-missing host that has the partition
+            if (missingHosts.contains(leaderId)) {
+                List<Host> perspectiveHosts = Lists.newArrayList();
                 for (Host host : topology.hostsById.values()) {
                     if (!missingHosts.contains(host.id)) {
                         List<Integer> list = topology.getPartitionIdList(host.id);
```
xinjiacs: maybe rename "list" with "partitionListOnNonMissingHost"?

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -1403,17 +1404,32 @@ public static AbstractTopology shiftPartitionLeaders(AbstractTopology topology,
...
+                            perspectiveHosts.add(host);
                         }
                     }
                 }
+
+                //Place the partition master to a node which hosts the partition and has the least masters.
+                assert(!perspectiveHosts.isEmpty());
+                perspectiveHosts.sort((Host a, Host b) -> {
+                    int result = 0;
+                    for (Partition p : topology.partitionsById.values()) {
```
xinjiacs: int partitionLeaderCountForHostA = 0;
for (Partition p: a.partitions) {
    if (p.leaderHostId == a.id) {
        partitionLeaderCountForHostA++;
    }
}

// do the same calculation for host B
// Compare CountA with CountB

The above calculation minimize the number of iterations on ALL partitions, maybe better to read than using one "result" to compare.

Also, do you need to sort them all to find the least partition leader contained Host?

src/frontend/org/voltdb/AbstractTopology.java
zeemanhe: ok

src/frontend/org/voltdb/AbstractTopology.java
zeemanhe: The algo has been updated.  only go through the partitions on host and find the number of leaders.  The sort is for quick access.  The perspective list is very small (<= K).

PR #4504, opened 2017-05-08
ENG-9346 fix junit

brians127:
Fix 100% failure of TestStatisticsSuitePlannerStats by updating the test to reflect the promotion of several statistics from INTEGER to BIGINT.


benjaminballard:
I went over the tests with Phil.  This fixed the junit test that was failing, and I'm confident the other failures have nothing to do with the changes in this branch.

PR #4505, opened 2017-05-09
PostgreSQLBackend.java: modified the nowQueryTransformer, to strictly

johnTcrawford:
add parentheses after NOW when it's a separate word (presumably the NOW
SQL function), rather than in the middle of a word (e.g., in a randomly
generated VARCHAR value).

NonVoltDBBackend.java: modified to limit the number of messages that are
printed about caught exceptions (in a few cases, the console was swamped
with them).

killstragglers.sh: added code to kill any processes that are using any
of the usual VoltDB ports (3021, 7181, 21211, 21212, 8080, 5555).


prosegay:
the old killstragglers logged to syslog the processes it (tried to) kill, the additional attempts to kill straggling processes that you introduced does not log what it kills.

johnTcrawford:
Thanks, Phil. I added that (and a few other tweaks).

PR #4506, opened 2017-05-09
ENG-12247 Add function ID

yzhang1991:


src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
```diff
@@ -166,23 +176,29 @@ protected boolean processStatement(DDLStatement ddlStatement, Database db, DdlPr
...
 
-        Class<?>[] paramTypes = functionMethod.getParameterTypes();
-        for (int i = 0; i < paramTypes.length; i++) {
-            Class<?> paramType = paramTypes[i];
+        Class<?>[] paramTypeClasses = functionMethod.getParameterTypes();
+        Type[] parameterTypes = new Type[paramTypeClasses.length];
+        for (int i = 0; i < paramTypeClasses.length; i++) {
+            Class<?> paramType = paramTypeClasses[i];
             if (! m_allowedDataTypes.contains(paramType)) {
                 String msg = String.format("Method %s.%s has an unspported parameter type %s at position %d",
```
wolffcm: This is not a new message, but should we be using 0- or 1-based indexing to identify parameter positions?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -420,13 +425,24 @@ public int getTypeParameter() {
             return m_typeParameter;
         }
 
+        private static int udfCount = 0;
+
+        static void addUserDefinedFunctionId(String functionName, Type returnType, Type[] parameterTypes, short[] syntax) {
+            FunctionId fid = by_LC_name.get(functionName);
+            if (fid == null) {
```
wolffcm: Should this always be null?  If so it could be an assert.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -420,13 +425,24 @@ public int getTypeParameter() {
             return m_typeParameter;
         }
 
+        private static int udfCount = 0;
```
wolffcm: Static non-final member?  Any race conditions possible accessing this?

wolffcm: I think we might do all of our parsing on one thread, but I'm not certain.

src/hsqldb19b3/org/hsqldb_voltpatches/Tokens.java
```diff
@@ -2222,7 +2222,7 @@
         commandSet.put(T_CLOSEBRACKET, CLOSEBRACKET);
     }
 
-    static int get(String token) {
+    public static int get(String token) {
```
wolffcm: This looks like it can go back to having package visibility now.

src/hsqldb19b3/org/hsqldb_voltpatches/Types.java
```diff
@@ -642,7 +646,7 @@ public static String getTypeString(int type, int precision, int scale) {
      * @return java.sql.Types int value
      * @throws  HsqlException
      */
-    static int getParameterSQLTypeNumber(Class c) {
+    public static int getParameterSQLTypeNumber(Class c) {
```
wolffcm: You could do the translation from Class to HSQL type number in FunctionForVoltDB, which would let this method stay package-visible and also reduce the number of dependencies between components.

tests/sqlcmd/baselines/ddl/load_remove_classes.outbaseline
```diff
@@ -292,7 +292,7 @@ create function IntFunction from method sqlcmdtest.IntFunction.run;
 Command succeeded.
 
 --- User-defined Functions -----------------------------------
-IntFunction         scalar function     sqlcmdtest.IntFunction.run                        
+intfunction         scalar function     sqlcmdtest.IntFunction.run
```
wolffcm: What happened that made this all lower-case now?

tests/frontend/org/voltdb/regressionsuites/TestUDFSuite.java
```diff
@@ -0,0 +1,116 @@
...
+
+import junit.framework.Test;
+
+public class TestUDFSuite extends RegressionSuite {
+
+    public void testAddRemoveUDF() throws IOException, ProcCallException {
+        Client client = getClient();
+        String[] functionNamesToTry = new String[] {"testfunc", "TESTFUNC", "testFunc"};
+        for (String functionName : functionNamesToTry) {
+            verifyStmtFails(client, String.format("SELECT %s(a) FROM T;", functionName), "user lacks privilege or object not found: TESTFUNC");
```
wolffcm: It would be good to test with 0, 1, and 2 parameters.

src/ee/expressions/functionexpression.h
```diff
@@ -302,6 +302,10 @@ static const int FUNC_TRIM_LEADING_CHAR               = SQL_TRIM_VOLT_FUNC_OFFSE
 static const int FUNC_TRIM_TRAILING_CHAR              = SQL_TRIM_VOLT_FUNC_OFFSET + SQL_TRIM_TRAILING;
 static const int FUNC_TRIM_BOTH_CHAR                  = SQL_TRIM_VOLT_FUNC_OFFSET + SQL_TRIM_BOTH;
 
+static const int FUNC_VOLT_UDF_ID_START                = 1000000;
```
yzhang1991: Well, you actually made me think about this.
I removed the FUNC_VOLT_UDF_ID_END in my next commit.
You see, so far we have only used 21023 function IDs for our built-in functions. Far less than 1000000.
Anything greater than 1000000 can be considered as a UDF function ID. I think it is even less likely that we will have 1000000 built-in functions.

src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
yzhang1991: Let me ask Andrew

yzhang1991: He says:

>Since we do zero-based positioning for stored procs, I would do the same for UDFs.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
yzhang1991: One thread called "Ad Hoc Planner"

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
yzhang1991: No.
If it is replaying the CREATE FUNCTION statement when executing other DDLs, this fid will not be NULL.

src/hsqldb19b3/org/hsqldb_voltpatches/Types.java
yzhang1991: Good point! Thanks

tests/sqlcmd/baselines/ddl/load_remove_classes.outbaseline
yzhang1991: To be consistent with built-in functions.

wolffcm: Are there any conflicts that can arise if there are two functions whose names differ only in upper/lower case?  If I try to define a function called `foo` and then one called `FOO`, the second one would fail, correct?  Would be good to test.

wolffcm: Actually I was confused about this.  There is a SQL name that is different from the Java name.  Never mind.

wolffcm:
Looks good, thanks for making those changes.

PR #4516, opened 2017-05-11
Don't allow @SwapTables from an @AdHoc invocation.

BillWhite:
If we try to execute
  client.callProcedure("@SwapTables", "t1", "t2")
we turn this into an AdHoc so we can plan it.  The invocationName of
this AdHoc work is "@SwapTables".  So it looks a little like we've
called
  client.callProcedure("@AdHoc", "@SwapTables t1 t2")
The only difference is the work.invocationName. in AsyncCompilerAgent.
So, we look for this there in AsyncCompilerAgent an throw an error
message if we see this case.


src/frontend/org/voltdb/compiler/AsyncCompilerAgent.java
```diff
@@ -390,6 +391,23 @@ else if (work.userPartitionKey == null) {
...
+                 * "@SwapTables T1 T2".  But if we actually see a call like
+                 *   (2) client.callProcedure("@AdHoc", "@SwapTables T1 T2")
+                 * we fall into the same code path.  So, we will not be able to tell
+                 * line (1) from  line (2) above in the planner.  This is as it
+                 * should be, since the planner should not necessarily know
+                 * the secrets of its caller.  In case (1) above the work.invocationName
+                 * will be "@SwapTables" and in case (2) it will be "@AdHoc".  So
+                 * we throw an error here in case (2).
+                 */
+                if ( ! "@SwapTables".equals(work.invocationName)
```
wolffcm: What happens if you have a statement in a stored procedure?
```SQL
CREATE PROCEDURE myproc AS @SwapTables T1 T2;
```
Will that fail with the same error?

BillWhite: Bummer.   I think I'll be able to test this later today, between the interview and the planning meeting.

BillWhite: You're right.  This is incorrect.  I'll try to fix it later today.

BillWhite:
Close to create a new pull request.

wolffcm:
Looks good.

PR #4519, opened 2017-05-11
run_voltdb_server.sh: in GEB tests of the VMC, added the new

johnTcrawford:
org/voltdb_testfuncs/IntFunction.class to the jar file, since this is
now required by fullDDL.sql (and therefore by the GEB/VMC's
FullDdlSqlBasicTest).


akhanzode:
Looks good to me.

PR #4522, opened 2017-05-12
ENG-12384: Make autogenerated index names consistent

wolffcm:
We have some code that generates index names, since we create indexes as the data structures that enforce PRIMARY KEY and UNIQUE constraints.  This code was a little wonky and it would generate a different name based on whether the constraint was added in CREATE TABLE or in an ALTER TABLE statement.  When this happened no DDL could be executed after that.

There was also a bug where sometimes the HSQL "SYS_IDX_SYS_PK..." name would leak out into VoltDB.

It says there was 347 files changed but mostly these are planner tester baseline files since the index names have changed.


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1218,26 +1218,28 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
             }
 
             if (subNode.name.equals("indexes")) {
-                // do non-system indexes first so they get priority when the compiler
-                // starts throwing out duplicate indexes
```
yzhang1991: What are the duplicate indexes, the system-generated indexes?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1681,8 +1683,7 @@ else if (subNode.name.equals("predicate")) {
         if (has_geo_col) {
             index.setType(IndexType.COVERING_CELL_INDEX.getValue());
         }
-        else if (( ! indexNameNoCase.contains("tree") ) && indexNameNoCase.contains("hash") &&
-                 ! indexNameNoCase.startsWith(HSQLInterface.AUTO_GEN_PRIMARY_KEY_PREFIX.toLowerCase())) {
+        else if (( ! indexNameNoCase.contains("tree") ) && indexNameNoCase.contains("hash")) {
```
yzhang1991: Why can you remove this condition checking?
See example below:

```
1> create table somethingwithhash (a int primary key);
Command succeeded.
2> create index abc on somethingwithhash(a);
Command succeeded.
3> exec @SystemCatalog INDEXINFO;
TABLE_CAT  TABLE_SCHEM  TABLE_NAME         NON_UNIQUE  INDEX_QUALIFIER  INDEX_NAME                      TYPE  ORDINAL_POSITION  COLUMN_NAME  ASC_OR_DESC  CARDINALITY  PAGES  FILTER_CONDITION 
---------- ------------ ------------------ ----------- ---------------- ------------------------------- ----- ----------------- ------------ ------------ ------------ ------ -----------------
NULL       NULL         SOMETHINGWITHHASH            1 NULL             ABC                                 3                 1 A            A                    NULL   NULL NULL             
NULL       NULL         SOMETHINGWITHHASH            0 NULL             VOLTDB_AUTOGEN_IDX_PK_SOMETHINGWITHHASH     2                 1 A            NULL                 NULL   NULL NULL             

(Returned 2 rows in 0.00s)
```
`VOLTDB_AUTOGEN_IDX_PK_SOMETHINGWITHHASH` has index type = 2 while `abc` has index type =3.

wolffcm: Yes, you are right.  I had removed the check since:
- There was a bug where HSQL index names were sometimes leaking into VoltDB (dependent on how the PK constraint is created)
- This made it so the above check was not detecting primary keys
- As a result it was possible to create a primary key backed by a hash index (we had a test actually locking this down)
- So I removed the above check to preserve existing behavior (but as you demonstrated, I did not succeed)
 
But it seems like if we have the word "hash" in the table name, then VoltDB will try to create a hash index, which is not what we want.  The code and our documentation says that only a constraint name or index name (and not a table or column name) should trigger creating of a hash index.

I had added the table name to the index to help make the name unique, but it turns out that both indexes and constraints occupy the global namespace (unlike columns with are in the table's namespace).  It is not possible to create indexes or constraints with duplicate names on different tables.  So it seems to make sense to remove the table name from both constraint and index names.

src/frontend/org/voltdb/compiler/DDLCompiler.java
wolffcm: We want to avoid dropping system-generated indexes because they will have Constraint objects in the catalog that reference them.  It is not possible to have duplicate constraints---we will generate an error.

The way this code was before, it would prefer to drop system-generated indexes, which is not what we want.

PR #4524, opened 2017-05-12
ENG-12430: Make debug crash an error.

akhanzode:


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -809,7 +809,9 @@ private void pollImpl(SettableFuture<BBContainer> fut) {
                 m_pollFuture = null;
 
                 if (m_drainTraceForDebug != null) {
-                    VoltDB.crashLocalVoltDB("Rolling generation " + m_generation + " before it is fully drained. " +
+                    //Making this an ERROR. Looks like this is happening when ack starts drains and pollImpl is submitted to execute.
+                    //Since this does not happen outside of our system test I am going to make ENG-12430 non blocker.
```
brians127: I don't think we should keep the second line of this comment. If you feel we should reference a ticket, you should state that this message is to aid with debugging ENG-12282.

PR #4527, opened 2017-05-15
Update kafkastream 10 package with latest compression jar files from kafka 10

hhakimi53:
Licensing info in third party readme  is fine, no updates there


akhanzode:
Looks like kafkaloader10 uses all libs from voltdb/lib which is not what its supposed to do just client and kafkastream10 jar.

hhakimi53:
Ok, will update it. The approach taken was load in kafkastream10 and than the jar from libs.

akhanzode:
You will need voltdb-client in classpath though for it to work.

hhakimi53:
Kafkloader10 puts voltdb jars i.e. client and server jar in classpath of kafkloader and with the current changes loader works ok. Server jar is also needed to be in classpath as CSV parser and formatter files which are packaged in server jar and are not in client jar.

PR #4528, opened 2017-05-15
ENG-12476, plan_upgrade command supports SSL-enabled VoltDB cluster.

luyangco:
Change-Id: I985bde85f992bed638a3b96dbaab42ef6ede83f4

### DIFF
--- a/lib/python/voltcli/voltadmin.d/plan_upgrade.py
+++ b/lib/python/voltcli/voltadmin.d/plan_upgrade.py
@@ -28,9 +28,6 @@ import sys
 import subprocess
 import ssl
 
-HTTP = 'http://'
-HTTPS = 'https://'
-
 @VOLT.Command(
     bundles=VOLT.AdminBundle(),
     description="Generate a checklist before performing online upgrade.",
@@ -446,11 +443,11 @@ def writeCommands(file, subject, command):
 # get deployment file through rest API
 def getCurrentDeploymentFile(runner, host):
     sslContext = None
-    if runner.opts.ssl_config is not None:
-        protocol = HTTPS
-        sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
+    if runner.opts.ssl_config is None:
+        protocol = "http://"
     else:
-        protocol = HTTP
+        protocol = "https://"
+        sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
     url = protocol + getHostnameOrIp(host) + ':' + str(host.httpport) + '/deployment/download/'
     request = Request(url)
     base64string = base64.b64encode('%s:%s' % (runner.opts.username, runner.opts.password))
--- a/lib/python/voltcli/voltadmin.d/plan_upgrade.py
+++ b/lib/python/voltcli/voltadmin.d/plan_upgrade.py
@@ -26,6 +26,10 @@ import base64
 import os
 import sys
 import subprocess
+import ssl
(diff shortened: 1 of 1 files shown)


zeemanhe:
look good to me

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -439,14 +443,23 @@ def writeCommands(file, subject, command):
         file.write(command)
         file.write('\n\n')
 
+# get deployment file through rest API
 def getCurrentDeploymentFile(runner, host):
-    # get deployment file through rest API
-    url = 'http://' + getHostnameOrIp(host) + ':' + str(host.httpport) + '/deployment/download/'
+    sslContext = None
+    if runner.opts.ssl_config is not None:
```
xinjiacs: The same code pattern prefers:
`If X is None:
    ...
else:
    ...`

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -439,14 +443,23 @@ def writeCommands(file, subject, command):
...
+        protocol = HTTPS
+        sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
+    else:
+        protocol = HTTP
```
xinjiacs: Is that better to put the actually string content here for easy comparison? The global variables with `://` are only used once.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -439,14 +443,23 @@ def writeCommands(file, subject, command):
...
         file.write('\n\n')
 
+# get deployment file through rest API
 def getCurrentDeploymentFile(runner, host):
-    # get deployment file through rest API
-    url = 'http://' + getHostnameOrIp(host) + ':' + str(host.httpport) + '/deployment/download/'
+    sslContext = None
+    if runner.opts.ssl_config is not None:
+        protocol = HTTPS
+        sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
```
xinjiacs: @luyangco If this will not work on python version 2.6, you may want to add comments here for the reason/behavior of TLSv1.

luyangco: I choose PROTOCOL_TLSv1 because it supports python 2.6. The protocols don't support python 2.6 are TLSv1_1 and TLSv1_2.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -439,14 +440,23 @@ def writeCommands(file, subject, command):
...
+# get deployment file through rest API
 def getCurrentDeploymentFile(runner, host):
-    # get deployment file through rest API
-    url = 'http://' + getHostnameOrIp(host) + ':' + str(host.httpport) + '/deployment/download/'
+    sslContext = None
+    if runner.opts.ssl_config is None:
+        protocol = "http://"
+    else:
+        protocol = "https://"
+        sslContext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
```
akhanzode: See how voltadmin gets the context use the same code here.

akhanzode: See how voltdbclient.py uses ssl parameter.

PR #4529, opened 2017-05-15
Fix ENG-11832 error in UPDATE stream view

yzhang1991:
The fix for this bug is disabling some reference replacement code in HSQL.
The team had the concern that it may affect the answers we collect from the HSQL backend, we will see what the Jenkins say. My current guess is it won't.

Manual test completed:
* TestGroupByComplexMaterializedViewSuite
* TestMaterializedViewNonemptyTablesSuite
* TestMaterializedViewSuite
* TestStreamView

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDML.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDML.java
@@ -654,6 +654,7 @@ public class ParserDML extends ParserDQL {
             }
         }
 
+        /* Disabled for ENG-11832
         if (!targetTable.isView) {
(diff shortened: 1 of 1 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDML.java
```diff
@@ -654,6 +654,7 @@ void resolveUpdateExpressions(Table targetTable,
             }
         }
 
+        /* Disabled for ENG-11832
```
wolffcm: If this code isn't necessary for anything, you can just delete it.

PR #4530, opened 2017-05-15
ENG-12422 and ENG-12502: Fix a buffer overflow on the client port and handle end-of-stream errors silently.

brians127:
When an end-of-stream is seen during client authentication, the server will stop attempting to authenticate and close the socket. The previous behavior did this also, but it threw a stack trace into the log every time.

Anyone else who wants to use MessagingChannel (I don't think there's anybody) can handle end of stream explicitly or treat it like a generic IOException.


src/frontend/org/voltcore/utils/ssl/MessagingChannel.java
```diff
@@ -50,16 +52,19 @@ public ByteBuffer readMessage() throws IOException {
...
         int len = lengthBuffer.getInt();
+        if (len > VoltPort.MAX_MESSAGE_LENGTH) {
+            throw new IOException("Packet exceeds maximum allowed size");
+        }
         ByteBuffer message = ByteBuffer.allocate(len);
         while (message.hasRemaining()) {
             read = m_socketChannel.read(message);
             if (read == -1) {
-                throw new IOException("Failed to read message");
+                throw new IOException("End of stream while reading message");
```
akhanzode: Original failed to read was correct.

PR #4532, opened 2017-05-16
ENG-12500, resolve the MP peformance regression introduced by ENG-10599.

luyangco:
Change-Id: Idf2e29b2211effaae1ee51e1a42fac25e738268b

### DIFF
--- a/src/frontend/org/voltdb/SQLStmt.java
+++ b/src/frontend/org/voltdb/SQLStmt.java
@@ -97,20 +97,9 @@ public class SQLStmt {
     }
 
     /**
-     * Construct a SQLStmt instance from a SQL statement.
-     * @param sqlText Valid VoltDB compliant SQL with question marks as parameter
-     * place holders.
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/SQLStmt.java
```diff
@@ -93,21 +93,34 @@ public SQLStmt(String sqlText) {
...
+    }
+
+    /**
+     * Construct a SQLStmt instance from a SQL statement.
+     * @param sqlText Valid VoltDB compliant SQL with question marks as parameter
+     * place holders.
+     * @param joinOrder separated list of tables used by the query specifying the order they should be joined in
+     * @param computeCRC whether to ignore the computation of SQL text CRC.
+     */
+    public SQLStmt(String sqlText, String joinOrder, boolean computeCRC) {
```
jhugg: Are we changing the public interface that user's see? I don't want to do that.

PR #4534, opened 2017-05-17
ENG-12504 make 'init --classes' compile using the specified jar file rather than simply adding it to the CLASSPATH.

brians127:
You can now use 'init --classes' when a base class exists.
If --schema is supplied to init but the schema does not use all of the available classes and/or resources, the unused resources are now in the catalog for later use.

Arguably this is just a workaround for the base class issue - VoltCompiler still isn't smart enough to resolve all the dependencies. We just ensure the output staged-catalog.jar contains everything the user's jar did.

The 'fake use case' in the test is not currently 100% functional as a stand-alone app, but it does contain:
- base classes
- utility classes (package-info.java)
- non-class artifacts
This means that the JUnit test is able to cover these cases regardless.

### DIFF
--- a/src/frontend/org/voltdb/VoltDB.java
+++ b/src/frontend/org/voltdb/VoltDB.java
@@ -316,8 +316,11 @@ public class VoltDB {
         /** apply safe mode strategy when recovering */
         public boolean m_safeMode = false;
 
-        /** location of user supplied DDL */
+        /** location of user supplied schema */
         public File m_userSchema = null;
+
+        /** location of user supplied classes and resources jar file */
+        public File m_stagedClassesPath = null;
 
         public int getZKPort() {
             return MiscUtils.getPortFromHostnameColonPort(m_zkInterface, org.voltcore.common.Constants.DEFAULT_ZK_PORT);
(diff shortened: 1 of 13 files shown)


tests/testprocs/org/voltdb_testprocs/fakeusecase/greetings/package-info.java
```diff
@@ -0,0 +1,28 @@
...
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * A fake use case intended for use verifying that 'voltdb init --classes' works with non-trivial use cases.
+ * The theme is querying translations for 'Hello' in different languages.
+ */
+package fakeusecase.greetings;
```
akhanzode: Whats the use of this empty java file?

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -420,6 +420,58 @@ public boolean compileFromDDL(
         return compileInternalToFile(jarOutputPath, null, null, ddlReaderList, null);
     }
 
+    /** Compiles a catalog from a user provided schema and (optional) jar file. */
+    public boolean compileFromSchemaAndClasses(
+            final File schemaPath,
+            final File classesJarPath,
+            final File catalogOutputPath)
+    {
+        if (schemaPath == null || !schemaPath.exists()) {
```
akhanzode: If schema is not there we should still be able to compile and load classes in. There is diff ticket for it so we can get this fixed in that ticket.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -665,20 +668,34 @@ else if (arg.equalsIgnoreCase("force")) {
...
+                    if (!m_stagedClassesPath.exists()){
+                        System.err.println("FATAL: Supplied classes jar file " + m_stagedClassesPath + " does not exist.");
+                        referToDocAndExit();
+                    }
+                    if (!m_stagedClassesPath.canRead()) {
+                        System.err.println("FATAL: Supplied classes jar file " + m_stagedClassesPath + " can't be read.");
+                        referToDocAndExit();
+                    }
+                    if (!m_stagedClassesPath.isFile()) {
+                        System.err.println("FATAL: Supplied classes jar file " + m_stagedClassesPath + " is not an ordinary file.");
```
hhakimi53: If the supplied jar file does not exisits, is not readable or a regular file, should the error be reported at Error level instead of being fatal. 

Same thought about for schema file too.

src/frontend/org/voltdb/compiler/VoltCompiler.java
hhakimi53: Yes, for that case may be returning from here with success will do as there is nothing to no ddl to compile against.

tests/testprocs/org/voltdb_testprocs/fakeusecase/greetings/package-info.java
brians127: package-info.java provides Javadoc for the package as a whole. My test also uses it to verify that utility classes are also included in the staged catalog.

src/frontend/org/voltdb/compiler/VoltCompiler.java
brians127: I agree. I was just keeping last minute changes to a minimum since ENG-12391 isn't a blocker.

src/frontend/org/voltdb/VoltDB.java
brians127: It looks like all the argument parsing errors are FATAL and result in Volt not starting. I'm open to discussing this with Andrew, but as a user, I'd rather have the command fail than succeed with unexpected behavior.

hhakimi53: We should fail on this, as it's failure case and not success case. The thing in question here is should the reporting be at error or fatal level. Yes, will be good to clarify with Andrew for such a failure which result possibly due to wrong user input, should the error be reported at Error or Fatal level.

PR #4537, opened 2017-05-17
ENG-12500, avoid another performance bottleneck of unnecessary memory allocation.

luyangco:
Change-Id: Icebb0ab8d591176d07968591e34467ea24c18fb1

### DIFF
--- a/src/frontend/org/voltdb/SQLStmtAdHocHelper.java
+++ b/src/frontend/org/voltdb/SQLStmtAdHocHelper.java
@@ -58,14 +58,14 @@ public class SQLStmtAdHocHelper {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/SQLStmtAdHocHelper.java
```diff
@@ -58,14 +58,12 @@ public static SQLStmt createWithPlan(byte[] sqlText,
     }
 
     /**
-     * Construct a SQLStmt instance from a SQL statement.
+     * Construct a raw SQLStmt instance from a SQL statement.
```
jhugg: Maybe add a bit more context to this comment to explain how this is used? Like what does "raw" mean. Could call out the helper class here too.

PR #4539, opened 2017-05-18
Eng 12536 revert uac opt

xinjiacs:


tests/frontend/org/voltdb/regressionsuites/TestUpdateClasses.java
```diff
@@ -833,4 +834,62 @@ public void testCreateProceduresBeforeUpdateClasses() throws Exception {
...
+            config.m_pathToCatalog = pathToCatalog;
+            config.m_pathToDeployment = pathToDeployment;
+            startSystem(config);
+
+            ClientResponse resp;
+            InMemoryJarfile boom = new InMemoryJarfile(TestProcedure.class.getResource("addSQLStmt.jar"));
+            resp = m_client.callProcedure("@UpdateClasses", boom.getFullJarBytes(), null);
+            assertEquals(ClientResponse.SUCCESS, resp.getStatus());
+
+            resp = m_client.callProcedure("@AdHoc", "create procedure partition ON TABLE tt COLUMN pid from class voter.TestProcedure;");
```
zeemanhe: is voter.TestProcedure correct?

xinjiacs: It's just a package name. I get the jar file by putting this procedure into a voter package. You can see it by extracting the jar package. 

I guess I should add comments that how I get these jar files.

This tests really needs same class under same package to test, which is not easy to define in java Junit class.

luyangco:
The patch looks good to me.

PR #4541, opened 2017-05-19
ENG-12255: Parser refactor

jeffreycrump:
Add a switch that filters the parser input stream to remove things like batching constructs and other commands that aren't supported during "voltdb init".  Previously, DDL needed to be scrubbed of these items before they could be used. This will allow our examples (which have been annotated with batching constructs) to be sent "as is" through the command line.


src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -405,6 +406,10 @@ public boolean compileFromDDL(
             final String jarOutputPath,
             final String... ddlFilePaths)
     {
+        return compileFromDDL(false, jarOutputPath, ddlFilePaths);
+    }
+
+    public boolean compileFromDDL(boolean initClasses, final String jarOutputPath, final String... ddlFilePaths) {
```
akhanzode: initClasses is too specific for variable name. Add javadoc or comment explaining what this new param is. Is there a list to ignore when this bool is true? If someone adds new to list then without making more changes this should work.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -1807,7 +1822,7 @@ public void compileInMemoryJarfile(InMemoryJarfile jarfile) throws IOException
         try {
             m_classLoader = jarfile.getLoader();
             // Do the compilation work.
-            InMemoryJarfile jarOut = compileInternal(null, null, ddlReaderList, jarfile);
+            InMemoryJarfile jarOut = compileInternal(null, null, ddlReaderList, jarfile, false);
```
akhanzode: maybe add version of compile methods without flag which calls in with default false.

src/frontend/org/voltdb/utils/LineReaderAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public String readBatchLine() throws IOException {
+        m_lineNum++;
+        return m_reader.readLine();
+    }
+
+    void close() {
+        try {
+            m_reader.close();
+        } catch (IOException e) { }
```
akhanzode: Why not throw IOException and let caller handle and log etc.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -118,8 +119,15 @@ private static ClientResponse callProcedureHelper(String procName, Object... par
         return response;
     }
 
-    private static void executeDDLBatch(String batchFileName, String statements) {
+    private static void executeDDLBatch(String batchFileName, String statements, int batchEndLineNumber, DDLParserCallback callback) {
+
         try {
+
```
akhanzode: You put too much white spaces and empty lines.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -586,12 +574,12 @@ void close() {
...
      * @param parentLineReader  The current input stream, to be used for "here documents".
      */
-    static void executeScriptFile(FileInfo fileInfo, SQLCommandLineReader parentLineReader)
+    static void executeScriptFile(FileInfo fileInfo, SQLCommandLineReader parentLineReader, DDLParserCallback callback)
     {
         LineReaderAdapter adapter = null;
         SQLCommandLineReader reader = null;
 
-        if ( ! m_interactive) {
+        if ( ! m_interactive && callback == null) {
```
akhanzode: why check callback == null?

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -632,7 +620,8 @@ static void executeScriptFile(FileInfo fileInfo, SQLCommandLineReader parentLine
      * @param fileInfo  The FileInfo object describing the file command (or stdin)
      * @throws Exception
      */
-    private static void executeScriptFromReader(FileInfo fileInfo, SQLCommandLineReader reader)
+
```
akhanzode: oh the empty lines

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -91,6 +91,24 @@ public void tearDown() {
...
+                     "-- This command cannot be part of a DDL batch.\n" +
+                     "LOAD CLASSES voter-procs.jar\n";
+        VoltCompiler compiler = new VoltCompiler(false);
+        boolean success = compileInitDDL(true, ddl, compiler);
+        assertTrue(success);
+
+        success = compileInitDDL(false, ddl, compiler);
+        assertFalse(success);
+    }
+
```
akhanzode: You should add more test cases.
1. One which does not have ending batch.
2. case insensitivity of filter keywords.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2227,7 +2227,7 @@ private void stageSchemaFiles(Configuration config) {
         final boolean standalone = false;
         final boolean isXCDR = false;
         VoltCompiler compiler = new VoltCompiler(standalone, isXCDR);
-        if (!compiler.compileFromDDL(stagedCatalogFH.getAbsolutePath(), config.m_userSchema.getAbsolutePath())) {
+        if (!compiler.compileFromDDL(true, stagedCatalogFH.getAbsolutePath(), config.m_userSchema.getAbsolutePath())) {
```
jeffreycrump: Sure, if we're ok with adding state.  I figured these compiler objects would be reusable but if that's not the case, it would certainly make the code cleaner.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -292,6 +299,74 @@ protected final String checkIdentifierStart(
...
+
+        try {
+            SQLCommand.executeScriptFromReader(fileInfo, lineReader, callback);
+        }
+        catch (Exception e) {
+            throw m_compiler.new VoltCompilerException(e);
+        }
+
+        try {
+            reader.close();
```
jeffreycrump: Copy-n-paste from the original implementation (lines 377-ff).  The readers are all created up front (VoltCompiler.createDDLFileReader) but closed when they are consumed.   We could refactor this, I suppose, but the asymmetry already exists.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -306,45 +381,7 @@ void loadSchema(Reader reader, Database db, DdlProceduresToLoad whichProcs)
...
-                    if (ddlStmtInfo.creatStream) {
-                       processCreateStreamStatement(stmt, db, whichProcs);
-                    }
-                } catch (HSQLParseException e) {
-                    String msg = "DDL Error: \"" + e.getMessage() + "\" in statement starting on lineno: " + stmt.lineNo;
-                    throw m_compiler.new VoltCompilerException(msg, stmt.lineNo);
-                }
-
-            }
+            executeStatement(db, whichProcs, stmt);
```
jeffreycrump: Sure, no problem.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1972,4 +2009,47 @@ private static AbstractExpression buildPartialIndexPredicate(
     public void processMaterializedViewWarnings(Database db) throws VoltCompilerException {
             m_mvProcessor.processMaterializedViewWarnings(db, m_matViewMap);
     }
+
+    private void executeStatement(final Database db, final DdlProceduresToLoad whichProcs, DDLStatement stmt) throws VoltCompilerException {
```
jeffreycrump: Got it.

src/frontend/org/voltdb/compiler/VoltCompiler.java
jeffreycrump: VoltCompiler.compileFromDDL is the public method, it calls this internal one with the default of false.

src/frontend/org/voltdb/utils/LineReaderAdapter.java
jeffreycrump: Same as above -- it was working this way originally.  A refactor could be done, I suppose.

src/frontend/org/voltdb/utils/SQLCommand.java
jeffreycrump: Removed.

src/frontend/org/voltdb/utils/SQLCommand.java
jeffreycrump: I actually think it makes the code more readable, but I'll remove it.

src/frontend/org/voltdb/utils/SQLCommand.java
jeffreycrump: Without it, the extra line of whitespace on System.out causes problems with the code generator that builds some of the unit tests, and the build will fail.  If there's a callback specified, we're not truly "interactive" at this point.   We should also consider refactoring/cleaning up the use of System.out and replacing it with a generic output stream and adding real logging to these classes.

src/frontend/org/voltdb/compiler/VoltCompiler.java
jeffreycrump: OK.  I'll make sure to document it accordingly if we decide to take out the parameter and replace it with a member field.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -292,6 +299,74 @@ protected final String checkIdentifierStart(
...
+                DDLStatement stmt = new DDLStatement(statement, lineNum);
+                processVoltDBStatements(db, whichProcs, stmt);
+            }
+        };
+
+        try {
+            SQLCommand.executeScriptFromReader(fileInfo, lineReader, callback);
+        }
+        catch (Exception e) {
+            throw m_compiler.new VoltCompilerException(e);
```
hhakimi53: lineReader.close() before throwing exception?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1972,4 +2009,47 @@ private static AbstractExpression buildPartialIndexPredicate(
     public void processMaterializedViewWarnings(Database db) throws VoltCompilerException {
             m_mvProcessor.processMaterializedViewWarnings(db, m_matViewMap);
     }
+
+    private void processVoltDBStatements(final Database db, final DdlProceduresToLoad whichProcs, DDLStatement stmt) throws VoltCompilerException {
+
+        boolean processed = false;
+
+        try {
+            processed = m_voltStatementProcessor.process(stmt, db, whichProcs);
```
hhakimi53: is the original comment "Process a VoltDB-specific DDL statement, like PARTITION, REPLICATE, CREATE PROCEDURE, CREATE FUNCTION, and CREATE ROLE." still valid. If so, we should leave the original comment in there

src/frontend/org/voltdb/compiler/DDLCompiler.java
hhakimi53: lineReader.close() ?

src/frontend/org/voltdb/compiler/DDLCompiler.java
jeffreycrump: OK

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
jeffreycrump: Done. https://github.com/VoltDB/voltdb/commit/f611ad45028d184bef8dd9327f13fb127c6b3c63

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -157,6 +158,9 @@
 
     private final boolean m_isXDCR;
 
+    // Whether or not to use SQLCommand as a pre-processor for DDL (in voltdb init --classes). Default is false.
+    private boolean m_initClasses = false;
```
brians127: The comment is very helpful, but consider using a different name for this variable. Somebody other than 'init --classes' may want to use it in the future.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -680,7 +668,7 @@ private static void executeScriptFromReader(FileInfo fileInfo, SQLCommandLineRea
                         // For now, treat the final semicolon as optional and
                         // assume that we are not just adding a partial statement to the batch.
                         batch.append(statement);
-                        executeDDLBatch(fileInfo.getFilePath(), batch.toString());
+                        executeDDLBatch(fileInfo.getFilePath(), batch.toString(), reader.getLineNumber(), callback);
```
brians127: Nitpick / not review blocker: you're passing in the line number before the callback here, but after the callback in executeStatements().

src/frontend/org/voltdb/compiler/DDLCompiler.java
brians127: I think a 'finally' block would be helpful here.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -1024,7 +1024,13 @@ private void compileDatabase(
                 // add the file object's path to the list of files for the jar
                 m_ddlFilePaths.put(schemaReader.getName(), schemaReader.getPath());
 
-                ddlcompiler.loadSchema(schemaReader, db, whichProcs);
+                SQLParser.FileInfo fi = new SQLParser.FileInfo(schemaReader.getPath());
```
brians127: If this is only needed when m_initClasses is true, why define the variable here? If there is an straightforward and safe way to add proper line number tracking to ddlcompiler.loadSchema() that would be even better.

src/frontend/org/voltdb/utils/SQLCommand.java
brians127: If you can't eliminate the check without substantial changes to the build infrastructure, please leave a code comment to that effect.

src/frontend/org/voltdb/compiler/DDLCompiler.java
jeffreycrump: I've added a comment to indicate we should straighten that all out, but it's a pretty big effort that I don't want to get into for this ticket.  The exception handling is pretty messy and too generic right now.

src/frontend/org/voltdb/utils/LineReaderAdapter.java
jeffreycrump: I've added a comment to indicate the issue and that we should fix it later.

src/frontend/org/voltdb/compiler/VoltCompiler.java
jeffreycrump: Any suggestions?  Plain "init" doesn't really capture the spirit.

src/frontend/org/voltdb/compiler/VoltCompiler.java
jeffreycrump: Goot catch. I'll move it inside.

src/frontend/org/voltdb/compiler/VoltCompiler.java
brians127: m_filterWithSQLCMD?

src/frontend/org/voltdb/utils/SQLCommand.java
jeffreycrump: Good catch, my OCD is satisfied!

jeffreycrump: Another good catch, I'll put the close inside a finally here.

src/frontend/org/voltdb/compiler/VoltCompiler.java
jeffreycrump: Sounds good.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -292,6 +299,73 @@ protected final String checkIdentifierStart(
...
+        try {
+            SQLCommand.executeScriptFromReader(fileInfo, lineReader, callback);
+        }
+        catch (Exception e) {
+            throw m_compiler.new VoltCompilerException(e);
+        }
+        finally {
+            // NEEDSWORK (ENG-12255): This really should be done outside the loop, by the caller. The exception propagation, at this point, is pretty messy, and
+            // changing it would be a pretty big effort that's really not relevant to this ticket. Fix this sometime.
+            lineReader.close();
```
hhakimi53: Passed in reader also needs to be closed along in here if it's not getting closed elsewhere.

brians127: The lineReader is created from the passed in reader, and is responsible for closing it.

hhakimi53: Oh yes, the linereader's is created through the reader so the linereader close will take care of it.

PR #4545, opened 2017-05-22
importer module management logic update

hhakimi53:
- Update module management logic - the OSGI module load/unload logic has been tweaked so that on importer restart request on UAC does not load ->(start->stop)+ every importer bundle. Instead the bundle is loaded once the importer type is configured to ingest data is defined and remains in memory and in started mode. 
- Tweak Kinesis importer logic to create worker upfront to so that during shutdown , Kenesis worker is signalled to stop. This helps preventing scenario where the shutdown may have get triggered before the worker is created (in accept thread), which results in importer shutdown to hang. Also the cloudwatch metric reporting for Kenesis has been turned off (due to issues around graceful/clean shutdown - https://github.com/awslabs/amazon-kinesis-client/issues/79 and https://github.com/awslabs/amazon-kinesis-client/issues/108)


src/frontend/org/voltdb/importclient/kinesis/KinesisStreamImporter.java
```diff
@@ -134,7 +135,7 @@ public IRecordProcessor createProcessor() {
 
     private class StreamConsumer implements IRecordProcessor {
 
-        private String m_shardId;
+        private String m_shardId = new String("");
```
brians127: Not blocker: you don't need "new String()".

src/frontend/org/voltdb/importclient/kinesis/KinesisStreamImporter.java
```diff
@@ -210,6 +211,11 @@ public void processRecords(ProcessRecordsInput records) {
         @Override
         public void shutdown(ShutdownInput shutDownInput) {
 
+            if (isDebugEnabled()) {
+                debug(null, "shard ID: " + (m_shardId.trim().isEmpty()? "unknown" : m_shardId)
```
brians127: Not blocker: please put a space before the ?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -146,13 +159,102 @@ private synchronized void create(int myHostId, CatalogContext catalogContext) {
...
+     */
+    private void discoverConfigsAndLoadBundles(CatalogContext catalogContext) {
+        m_configsForProcessor.clear();
+        for (String configName : m_processorConfig.keySet()) {
+            ImportConfiguration importConfig = m_processorConfig.get(configName);
+            Properties properties = importConfig.getmoduleProperties();
+
+            String importBundleJar = properties.getProperty(ImportDataProcessor.IMPORT_MODULE);
+            Preconditions.checkNotNull(importBundleJar,
+                    "Import source is undefined or custom export plugin class missing.");
```
brians127: Custom import plugin?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -57,10 +62,18 @@
 
     private final int m_myHostId;
     private ChannelDistributer m_distributer;
-    private boolean m_serverStarted;
+    private volatile boolean m_serverStarted;
     private final ImporterStatsCollector m_statsCollector;
     private final ModuleManager m_moduleManager;
 
+    // Maintains the record of the importer bundles that have been loaded into the memory based on importer config.
+    // These loaded loaded bundles remains in memory. ready for use and don't get loaded/unloaded
```
brians127: Is it necessary to unload unused bundles when the importer config changes? For example, if someone wants to switch from Kafka 8 to Kafka 10 without stopping VoltDB?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -146,13 +159,102 @@ private synchronized void create(int myHostId, CatalogContext catalogContext) {
...
+    /**
+     * Checks if the module for importer has been loaded in the memory. If bundle doesn't exists, it loades one and
+     * updates the mapping records of the bundles.
+     *
+     * @param moduleProperties
+     * @return true, if the bundle corresponding to the importer is in memory (uploaded or was already present)
+     */
+    private boolean loadImporterBundle(Properties moduleProperties){
+        String importModuleName = moduleProperties.getProperty(ImportDataProcessor.IMPORT_MODULE);
+        String attrs[] = importModuleName.split("\\|");
```
brians127: Can this ever be in an unexpected or invalid format?

src/frontend/org/voltdb/importer/ImportProcessor.java
```diff
@@ -221,30 +157,40 @@ public void run() {
...
-            String procedure = properties.getProperty(IMPORT_PROCEDURE);
-            //TODO: If processors is a list dont start till all procedures exists.
-            Procedure catProc = catalogContext.procedures.get(procedure);
-            if (catProc == null) {
-                catProc = catalogContext.m_defaultProcs.checkForDefaultProcedure(procedure);
-            }
+    private void addProcessorConfig(ImportConfiguration config, Map<String, AbstractImporterFactory> importerModules) {
+        Properties properties = config.getmoduleProperties();
+
+        String module = properties.getProperty(ImportDataProcessor.IMPORT_MODULE);
```
brians127: Can this ever be in an unexpected format?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -146,13 +159,102 @@ private synchronized void create(int myHostId, CatalogContext catalogContext) {
                 }
             }
 
-            newProcessor.setProcessorConfig(catalogContext, m_processorConfig);
-            m_processor.set(newProcessor);
+            discoverConfigsAndLoadBundles(catalogContext);
+            if (!m_configsForProcessor.isEmpty()) {
```
brians127: Why this conditional? If we're switching from having importers to not having any, don't we want to remove the import processor?

src/frontend/org/voltdb/importclient/kinesis/KinesisStreamImporter.java
```diff
@@ -59,9 +60,22 @@
     private AtomicLong m_cbcnt = new AtomicLong(0);
 
     private Worker m_worker;
+    private String m_workerId;
```
brians127: Is this used anywhere outside the constructor, or are we just keeping it for debugging purposes?

hhakimi53: It's helpful for diagnosing which threads belongs to which workers

src/frontend/org/voltdb/importclient/kinesis/KinesisStreamImporter.java
hhakimi53: it's needed for debug msg in shutdown. will tweak it so no conditional evaluation is needed there

src/frontend/org/voltdb/importer/ImportManager.java
hhakimi53: No, currently unloading of bundles is not done. That has potential to pollute next bundle restart if the importer did not shutdown properly.

src/frontend/org/voltdb/importer/ImportManager.java
hhakimi53: let me think about with no importer case or importers with not complete config case. thanks

src/frontend/org/voltdb/importer/ImportProcessor.java
hhakimi53: it we are this point the module should be there.

src/frontend/org/voltdb/importer/ImportManager.java
hhakimi53: yes, for that case the process needs to be removed. thnxs

src/frontend/org/voltdb/importer/ImportManager.java
brians127: It was verbally explained to me that this should never be the case. This is an internal construct.

PR #4546, opened 2017-05-22
ENG-9326 Don't restart the importers for every UpdateApplicationCatalog - only those which affect the importer configuration or classes.

brians127:
Importers will only be restarted if the importer configuration changes or importer-relevant procedures are added or dropped. Previous behavior was to restart importers for every UAC. 

Restarting importers is still all or none; there's no fine grained control yet.


src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -192,16 +197,37 @@ private synchronized void restart(CatalogContext catalogContext, HostMessenger m
...
+        Map<String, ImportConfiguration> oldImportConfig = m_processorConfig;
+        Map<String, ImportConfiguration> newImportConfig = CatalogUtil.getImportProcessorConfig(catalogContext.getDeployment().getImport());
+        setupFormatterFactoryForConfig(newImportConfig);
+        final boolean importConfigChanged = !oldImportConfig.equals(newImportConfig);
+
+        boolean procedureAvailabilityChanged = false;
+        ImportDataProcessor currentProcessor = m_processor.get();
+        if (currentProcessor != null) {
+            for (Map.Entry<String, Boolean> procedureStatus : currentProcessor.getUtilizedProcedures().entrySet()) {
+                // NOTE: default procedures are not tracked since it's assumed they cannot be added or removed.
```
akhanzode: This is not true CRUD procedures are created when table is partitioned.

src/frontend/org/voltdb/importclient/junit/JUnitImporter.java
```diff
@@ -0,0 +1,80 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.voltdb.importclient.junit;
```
jeffreycrump: If this is only used in test scenarios, it should live in the 'test' subdirectory instead of 'src'.  You can keep the package names the same, though.

src/frontend/org/voltdb/importer/ImportManagerWithMocks.java
```diff
@@ -0,0 +1,34 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.voltdb.importer;
```
jeffreycrump: Similarly, if this is only test code it should live in the test subdirectory. You might need to change package/field/class visibility but that's a small price to pay.

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -276,10 +285,7 @@ public void testImportSimpleData() throws Exception {
     public void testImportMultipleTimes() throws Exception {
         System.out.println("testImportUpdateApplicationCatalog");
         Client client = getClient();
```
jeffreycrump: You can move stuff like this into the @Setup method, since that method gets called by JUnit prior to invocation of each test case.

src/frontend/org/voltdb/importer/ImportManagerWithMocks.java
brians127: Sadly this and MockChannelDistributer have to stay here. The build fails with "class not found" because src and tests aren't compiled together. Since these are tiny, low impact files I don't think it's worth switching to a "load at runtime" approach. I will move the test importer though.

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -260,70 +274,82 @@ private void verifyData(Client client, int count, int min) throws Exception {
...
+        verifyData(m_client, 7000, 3001);
+    }
+
+    /** Verify that importer can withstand unrelated UACs without restarting.
+     * We don't expect the importers to be restarted during this test;
+     * the pushers will get IOExceptions if the importers restart.
+     * It's possible to get a false pass if the data finishes before the UACs and importers restart anyway,
+     * but manual runs indicate that the data outlasts the UACs by 2-3 seconds.
+     */
+    public void testImportUnrelatedUACWhilePushing() throws Exception {
```
brians127: This test is arguably a duplicate of some of the TestImportManager cases. That said it should be more resilient to false negatives created by broken JUnit infrastructure, since it's and end to end test with a real Volt cluster.

tests/frontend/org/voltdb/importclient/junit/JUnitImporterConfig.java
```diff
@@ -0,0 +1,95 @@
...
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
```
brians127: One curious consequence of moving this to the 'test' subdirectory is that it's no longer under the AGPL (licensecheck won't let me use it in /test). This means Community users could use it as a template on how to write an importer.

jhugg: So I'm totally fine with using this license and putting this here (business-wise). That said, if there are questions about this kind of thing while you're coding, it would be better to raise them before committing to GitHub. At this point we're mostly talking about obfuscation.

brians127: OK. There was no question in my mind that this is test code, but something to keep in mind I suppose. For anyone else looking at this, the above license is our standard test code license.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -1304,7 +1304,7 @@ private String writeDeploymentFile(
         }
 
         // Have some yummy boilerplate!
-        File file = File.createTempFile("myAppDeployment", ".tmp");
+        File file = File.createTempFile("vdbTestDeployment", ".xml");
```
akhanzode: Why change unrelated things?

src/frontend/org/voltdb/importer/ImportManagerWithMocks.java
akhanzode: Why mock is not in test?

src/frontend/org/voltdb/importer/MockChannelDistributer.java
```diff
@@ -0,0 +1,67 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.voltdb.importer;
```
akhanzode: This should be in mock.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1308,6 +1310,21 @@ public void setFormatterFactory(AbstractFormatterFactory formatterFactory) {
...
         }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(m_moduleProps, m_formatterBuilder);
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (!(o instanceof ImportConfiguration)) {
```
akhanzode: add if (o == this) return true;

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -178,12 +187,17 @@ protected void close() {
...
+    private CountDownLatch asyncPushDataToImporters(int count, int loops) throws Exception {
         CountDownLatch latch = new CountDownLatch(2*loops);
         for (int i=0; i<loops; i++) {
             (new SocketDataPusher("localhost", 7001, count, latch, '\t')).start();
             (new Log4jDataPusher(count, latch, ',')).start();
         }
+        return latch;
+    }
+
+    private void pushDataToImporters(int count, int loops) throws Exception {
```
akhanzode: Whats the purpose of creating this wrapper? Looks like no logic change. Is it just for change of name of method you didnt like?

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -260,70 +274,82 @@ private void verifyData(Client client, int count, int min) throws Exception {
...
         System.out.println("testImportUpdateApplicationCatalog");
-        Client client = getClient();
-        while (!((ClientImpl) client).isHashinatorInitialized()) {
-            Thread.sleep(1000);
-            System.out.println("Waiting for hashinator to be initialized...");
-        }
 
         pushDataToImporters(100, 1);
-        verifyData(client, 100);
+        verifyData(m_client, 100);
```
akhanzode: You don't have to refactor everything you see every time. Not when there are no benefits and especially with test code. e.g. in this case you created a member m_client as opposed to client locally in test.

tests/frontend/org/voltdb/importclient/junit/JUnitImporter.java
```diff
@@ -0,0 +1,152 @@
...
+
+package org.voltdb.importclient.junit;
+
+import com.google_voltpatches.common.base.Preconditions;
+import org.voltdb.importer.AbstractImporter;
+
+import java.net.URI;
+import java.util.List;
+
+import static org.voltdb.importclient.junit.JUnitImporter.Event.*;
```
akhanzode: Isnt this inside the same class?

tests/frontend/org/voltdb/importclient/junit/JUnitImporter.java
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Created by bshaw on 5/31/17.
+ */
+public class JUnitImporter extends AbstractImporter {
```
akhanzode: I would have liked this outside of src. You could have used SocketImporter for this right?

tests/frontend/org/voltdb/importclient/junit/JUnitImporterMessenger.java
```diff
@@ -0,0 +1,165 @@
...
+ */
+public class JUnitImporterMessenger {
+
+    private static JUnitImporterMessenger m_self = null;
+    private EventBus m_eventBus = new EventBus("JUnitImporterMessenger");
+    private Map<URI, List<JUnitImporter.Event>> m_eventTracker = new HashMap<>();
+    private Throwable m_firstReportedException = null;
+
+    private static class Event {
+        private URI                 m_importerID;
```
akhanzode: tabs/indent. This whole thing can be used in socketimporter.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
brians127: I was debugging using the deployment and VoltDB requires a .xml file extension. I agree that this change is unrelated.

tests/frontend/org/voltdb/TestImportSuite.java
brians127: It exposes the latch. My new test waits on at the end. Existing tests want to wait for the latch immediately.

tests/frontend/org/voltdb/importclient/junit/JUnitImporter.java
brians127: This importer is in /tests not /src. Let's discuss what to do about SocketImporter - there are two of them, and my test does not import any data.

tests/frontend/org/voltdb/importclient/junit/JUnitImporterMessenger.java
brians127: I don't understand what you're asking me to do here with respect to spacing.

src/frontend/org/voltdb/importer/ImportManagerWithMocks.java
jeffreycrump: I don't understand. What is failing to compile?

tests/frontend/org/voltdb/importclient/junit/JUnitImporter.java
brians127: It is. IDE did the wrong thing. If we keep this code in JUnitImporter I'll fix this.

src/frontend/org/voltdb/importer/ChannelDistributer.java
```diff
@@ -101,7 +101,7 @@
  *  channels in their respective /import/host/[host-name] nodes. When a node leaves the mesh
  *  its assigned channels are redistributed among the surviving nodes
  */
-public class ChannelDistributer implements ChannelChangeCallback {
```
brians127: These changes will be reverted

src/frontend/org/voltdb/importer/formatter/FormatterBuilder.java
```diff
@@ -28,7 +29,7 @@
 
     private final String m_formatterName;
     private final Properties m_formatterProps;
-    private AbstractFormatterFactory  m_formatterFcatory;
+    private AbstractFormatterFactory m_formatterFactory;
```
jeffreycrump: Thanks for fixing this.  I made the same change on my branch so I'll take your changes before merging mine up.

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -262,68 +314,124 @@ private void verifyData(Client client, int count, int min) throws Exception {
...
+        updateDeploymentFile(false, false);
+        Thread.sleep(1000);
+        try {
+            testConnector.tryPush(5);
+            fail("Importer is still running even though it is no longer configured");
+        } catch (IOException expected) {
+        }
+
+        updateDeploymentFile(true, false);
+        Thread.sleep(1000);
```
akhanzode: You should not have to sleep here.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -195,6 +196,7 @@
     private static boolean m_exportEnabled = false;
     public static final String CATALOG_FILE_NAME = "catalog.jar";
     public static final String STAGED_CATALOG_PATH = Constants.CONFIG_DIR + File.separator + "staged-catalog.jar";
+    public static final java.lang.String VOLTDB_BUNDLE_LOCATION_PROPERTY_NAME = "voltdbbundlelocation";
```
akhanzode: final java.lang.String wonder why this needs to be fully qualified is there another String in here?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -117,94 +118,122 @@ public static synchronized void initialize(int myHostId, CatalogContext catalogC
...
             //TODO: If processors is a list dont start till all procedures exists.
             Procedure catProc = catalogContext.procedures.get(procedure);
             if (catProc == null) {
                 catProc = catalogContext.m_defaultProcs.checkForDefaultProcedure(procedure);
             }
-
             if (catProc == null) {
                 importLog.info("Importer " + configName + " Procedure " + procedure +
                         " is missing will disable this importer until the procedure becomes available.");
+                iter.remove();
```
akhanzode: Removing inside the iterator. Should collect keys and remove at end.

src/frontend/org/voltdb/importer/formatter/FormatterBuilder.java
```diff
@@ -59,4 +60,23 @@ public Properties getFormatterProperties()
...
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        }
+        if (!(o instanceof FormatterBuilder)) {
+            return false;
+        }
+        FormatterBuilder<?> other = (FormatterBuilder<?>) o;
+        return ((m_formatterName == null && other.m_formatterName == null) || m_formatterName.equals(other.m_formatterName))
```
akhanzode: does case sensitivity of name matters here?

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -375,15 +493,33 @@ public TestImportSuite(final String name) {
...
+    }
+
+    static public junit.framework.Test suite() throws Exception
+    {
+        LocalCluster config;
+        Map<String, String> additionalEnv = new HashMap<>();
+        //Specify bundle location
+        String bundleLocation = System.getProperty("user.dir") + "/bundles";
+        System.out.println("Bundle location is: " + bundleLocation);
+        additionalEnv.put("voltdbbundlelocation", bundleLocation);
```
akhanzode: If you are making a public static constant for voltdbbundlelocation please USE it.

tests/frontend/org/voltdb/TestImportSuite.java
jeffreycrump: On the other hand, "best practice" for JUnit says that fixtures which are created in every test should be instead created in the setUp() method.   It's clean to create the client in setUp and then you can remove it from the parameters to verifyData()

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -262,68 +314,124 @@ private void verifyData(Client client, int count, int min) throws Exception {
...
 
         pushDataToImporters(100, 1);
-        verifyData(client, 200);
-
-        client.close();
+        verifyData(m_client, 200);
     }
 
     public void testImportMultipleClientsInParallel() throws Exception {
         System.out.println("testImportMultipleClientsInParallel");
```
jeffreycrump: No need for console noise.

tests/frontend/org/voltdb/TestImportSuite.java
jeffreycrump: Does this mean everyone (and the test runners) would have to have this directory/file on their machine?

src/frontend/org/voltdb/importer/ImportManager.java
brians127: Why is that better?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -117,94 +118,122 @@ public static synchronized void initialize(int myHostId, CatalogContext catalogC
...
                 }
-            }
 
-            discoverConfigsAndLoadBundles(catalogContext);
-            if (!m_configsForProcessor.isEmpty()) {
-                newProcessor.setProcessorConfig(m_configsForProcessor, m_loadedBundles);
-                m_processor.set(newProcessor);
-            } else {
-                m_processor.set(null);
+                if (start) {
```
hhakimi53: May be this can be moved outside - function returns true indicating that the importers config and processors that are present and processors are needs to enabled to receive data.

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -117,94 +118,122 @@ public static synchronized void initialize(int myHostId, CatalogContext catalogC
...
+            boolean bundlePresent = loadImporterBundle(properties);
+            if (!bundlePresent) {
+                iter.remove();
             }
         }
-        if (!m_configsForProcessor.isEmpty()) {
-            importLog.info("Loaded importer modules: " + m_loadedBundles.keySet() +", types: " + m_importersByType.keySet());
-        }
+
+        setupFormatterFactoryForConfig(newProcessorConfig);
```
hhakimi53: Does m_formatterFactories needs to be cleared before calling setup formatter config to clear the formatter config that were there before the UAC event?

tests/frontend/org/voltdb/TestImportSuite.java
brians127: Since I'm adding four new tests which create a client on start-up, the amount of boilerplate I'd have to add to my new tests is more than the amount I removed with this change.

tests/frontend/org/voltdb/TestImportSuite.java
brians127: This code isn't new - I just moved it. I can replace the bundle path tweak with a different implementation I wrote for the previous generation of JUnit tests if you want. Yes you do need a bundles directory to load the CSVFormatter.

src/frontend/org/voltdb/importer/ImportManager.java
brians127: We discussed. I made this change believing that formatters were supposed to behave the same way as bundles, but this is apparently not true. I'll restore the clear().

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -72,7 +73,7 @@
     private Map<String, AbstractImporterFactory> m_importersByType = new HashMap<String, AbstractImporterFactory>();
     // maintains mapping of active importer config to the bundle bundle path used for communicating import processor
     // about the jars to use
-    private Map<String, ImportConfiguration> m_configsForProcessor= new HashMap<String, ImportConfiguration>();
+    private Map<String, ImportConfiguration> m_processorConfig = new HashMap<>();
```
brians127: This is only used inside synchronized methods.

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -85,9 +92,49 @@ private void setupLog4jSocketHandler() {
...
+                    Socket pushSocket = new Socket(m_server, m_port);
+                    OutputStream socketStream = pushSocket.getOutputStream();
+                    System.out.printf("Connected to VoltDB socket importer at: %s.\n", m_server + ":" + m_port);
+                    String s = String.valueOf(m_counter) + m_separator + System.currentTimeMillis() + "\n";
+                    socketStream.write(s.getBytes());
+                    pushSocket.close();
+                    return;
+                } catch (IOException e) {
+                    numConnectAttempts++;
+                    if (numConnectAttempts < maxAttempts) {
```
hhakimi53: Should it be "if (numConnectAttempts >= maxAttempts) { ... }" ?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -117,94 +118,122 @@ public static synchronized void initialize(int myHostId, CatalogContext catalogC
                 statsCollector);
 
         m_self = em;
-        em.create(myHostId, catalogContext);
+        em.initializeChannelDistributer();
```
brians127: I'll look into this

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -117,94 +118,122 @@ public static synchronized void initialize(int myHostId, CatalogContext catalogC
...
+        em.applyCatalogToImporters(catalogContext, false);
```
brians127: OK. I'll see if I can get rid of "boolean start" as well.

src/frontend/org/voltdb/importer/ImportManager.java
brians127: We discussed. It's a stylistic issue which may have performance implications depending on the map implementation.

src/frontend/org/voltdb/utils/CatalogUtil.java
brians127: It does not. I'll fix this.

src/frontend/org/voltdb/importer/formatter/FormatterBuilder.java
brians127: No. Good catch!

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -85,9 +92,49 @@ private void setupLog4jSocketHandler() {
...
+        m_client.close();
         super.tearDown();
     }
 
+    /** Allows the test thread to check if the importer is alive by writing data to it. */
+    static class ImporterConnector {
+        private final String m_server;
+        private final int m_port;
+        private final char m_separator;
+        private final int m_counter = 0;
```
brians127: It's used when generating the data being submitted to Volt.

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -284,20 +306,22 @@ public synchronized void close() {
...
+    public synchronized void updateCatalog(CatalogContext catalogContext, HostMessenger messenger) {
+        try {
+            Map<String, ImportConfiguration> newProcessorConfig = loadNewConfigAndBundles(catalogContext);
+            if (m_processorConfig == null || !m_processorConfig.equals(newProcessorConfig)) {
+                close();
+                startImporters(newProcessorConfig);
+                readyForDataInternal(catalogContext, messenger);
+            }
+        } catch (final Exception e) {
+            // TODO do I need Andrew to review error message changes?
```
brians127: TODO I am addressing

tests/frontend/org/voltdb/TestImportSuite.java
```diff
@@ -262,68 +314,124 @@ private void verifyData(Client client, int count, int min) throws Exception {
...
+
+    /** Verify that importer can withstand unrelated UACs without restarting.
+     * The data pushers will get IOExceptions if the importers restart.
+     * It's possible but unlikely to get a false pass if the data finishes before the deployment changes and importers restart anyway.
+     */
+    public void testImportUnrelatedDeploymentChangesWhilePushing() throws Exception {
+        System.out.println("Deployment changes that don't affect importers should not result in them restarting.");
+
+        // Count was selected to keep importers busy during all UACs without making the test take much longer than necessary.
+        CountDownLatch dataGenerationAwaiter = asyncPushDataToImporters(4000, 3);
```
brians127: In any implementation where UAC causes importer restart, there is always going to be a race between writing the data and restarting the importers. It's possible the write takes longer than the restart. What I can do is make sure the data does not run out until all the UACs are complete.

akhanzode:
If jenkins is good squash merge.

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -117,94 +117,104 @@ public static synchronized void initialize(int myHostId, CatalogContext catalogC
...
-            if (loadImporterBundle(properties)) {
-                m_configsForProcessor.put(configName, importConfig);
+            // NOTE: if bundle is already loaded, loadImporterBundle does nothing and returns true
+            boolean bundlePresent = loadImporterBundle(properties);
+            if (!bundlePresent) {
+                iter.remove();
             }
         }
-        if (!m_configsForProcessor.isEmpty()) {
-            importLog.info("Loaded importer modules: " + m_loadedBundles.keySet() +", types: " + m_importersByType.keySet());
```
hhakimi53: This log message is helpful in debugging. See if we can retain this one.

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -284,20 +294,21 @@ public synchronized void close() {
...
-    public void updateCatalog(CatalogContext catalogContext, HostMessenger messenger) {
-        restart(catalogContext, messenger);
+    public synchronized void updateCatalog(CatalogContext catalogContext, HostMessenger messenger) {
+        try {
+            Map<String, ImportConfiguration> newProcessorConfig = loadNewConfigAndBundles(catalogContext);
+            if (m_processorConfig == null || !m_processorConfig.equals(newProcessorConfig)) {
+                close();
+                startImporters(newProcessorConfig);
+                readyForDataInternal(catalogContext, messenger);
+            }
```
hhakimi53: May be add conditional assert before exiting routine to make sure the import processor is set to null if no valid import config was discovered.

src/frontend/org/voltdb/importer/ImportManager.java
brians127: It moved to line 139 in ImportManager. There was a small clarification made; it now says "Currently loaded importer modules".

PR #4547, opened 2017-05-23
ENG-12547: Revert to old index naming scheme

wolffcm:
I broke a system test: DR needs unique indexes and PK indexes to have the same name across versions, so these commits partly undo some of work I did to fix ENG-12384 (while still fixing that bug).

Most of the changed files are plannertester baseline files.


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -1563,12 +1563,8 @@ static Table addTableConstraintDefinitions(Session session, Table table,
                         index = table.createAndAddIndexStructure(indexName,
                             c.core.mainCols, null, null, true, true, false).setAssumeUnique(c.assumeUnique);
                     }
-                    /* disable 2 lines ...
-                    Index index = table.createAndAddIndexStructure(indexName,
-                        c.core.mainCols, null, null, true, true, false);
-                    ... disabled 2 lines */
-                    // End of VoltDB extension
```
BillWhite: Hooray.  (Removed VoltDB extension comments.)

PR #4549, opened 2017-05-23
ENG-11091:

manjujames:
- Send catalog update events even if DR tables are deleted.
- In XDCR mode, DR schema compatibility check should fail if the producer side deleted DR'ed tables.

### DIFF
--- a/src/catgen/in/javasrc/DRCatalogDiffEngine.java
+++ b/src/catgen/in/javasrc/DRCatalogDiffEngine.java
@@ -54,6 +54,8 @@ public class DRCatalogDiffEngine extends CatalogDiffEngine {
             "sqltext", "querytype", "readonly", "singlepartition", "replicatedtabledml", "iscontentdeterministic", "isorderdeterministic", "nondeterminismdetail",
             "cost", "seqscancount", "explainplan", "tablesread", "tablesupdated", "indexesused", "cachekeyprefix"
             );
+
+
     public DRCatalogDiffEngine(Catalog localCatalog, Catalog remoteCatalog) {
         super(localCatalog, remoteCatalog);
     }
@@ -103,10 +105,18 @@ public class DRCatalogDiffEngine extends CatalogDiffEngine {
 
     @Override
     protected String checkAddDropWhitelist(final CatalogType suspect, final ChangeType changeType) {
+        // Only on remote
         if (ChangeType.ADDITION == changeType && suspect instanceof Table) {
             assert ((Boolean)suspect.getField("isDRed"));
             return "Missing DR table " + suspect.getTypeName() + " on replica cluster";
         }
+
+        // Only on local. We care only if it is XDCR.
+        if (ChangeType.DELETION == changeType && suspect instanceof Table && isXDCR()) {
+            assert ((Boolean)suspect.getField("isDRed"));
+            return "Missing DR table " + suspect.getTypeName() + " on master cluster";
+        }
+
         if (suspect instanceof Column || isUniqueIndex(suspect) || isUniqueIndexColumn(suspect)) {
             return "Missing " + suspect + " from " + suspect.getParent() + " on " + (ChangeType.ADDITION == changeType ? "replica" : "master");
         }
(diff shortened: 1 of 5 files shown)


src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -103,10 +105,18 @@ public static Catalog deserializeCatalogCommandsForDr(String encodedCatalogComma
...
+        // Only on remote
         if (ChangeType.ADDITION == changeType && suspect instanceof Table) {
             assert ((Boolean)suspect.getField("isDRed"));
             return "Missing DR table " + suspect.getTypeName() + " on replica cluster";
         }
+
+        // Only on local. We care only if it is XDCR.
+        if (ChangeType.DELETION == changeType && suspect instanceof Table && isXDCR()) {
+            assert ((Boolean)suspect.getField("isDRed"));
+            return "Missing DR table " + suspect.getTypeName() + " on master cluster";
```
wweiss-voltdb: do you really mean "master cluster"

manjujames: Are you asking about the term 'master' or are you asking if it is a typo? It's not a typo. It gets in here if a table is deleted on master and still exists in replica. So the table is missing on master.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -2154,6 +2154,7 @@ public static LocalCluster compileBuilder(String schemaDDL, int siteCount, int h
         if (remoteReplicationPort != 0) {
             builder.setDRMasterHost("localhost:" + remoteReplicationPort);
         }
+        builder.setUseDDLSchema(true);
```
slowsoul: Do we support DDL schema and Catalog schema at the same time? Just wondering because this method has called addLiteralSchema() which is for Catalog schema mode. Seems no problem as Jenkins is happy. Maybe in this case, catalog schema will be the base schema and it still allows you to change schema using AdHoc DDL?

slowsoul:
Not fully understand "Send catalog update events even if DR tables are modified." in the description.

manjujames:
Meant to say 'if DR tables are deleted'. I will update the description.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
manjujames: I didn't notice the catalog schema node being setting explicitly. I will figure it out.

manjujames: Yes, that looks fine. What we send in addLiteralSchema is just used at initial compile time looks like.

src/catgen/in/javasrc/DRCatalogDiffEngine.java
wweiss-voltdb: The point is that isXDCR() is true so there is no master (or more precisely master is ambiguous.

slowsoul: I see, that's true. Maybe the message should be “missing table DR on remote/peer cluster”? It would be clear if this message will be prefixed by the remote cluster id? Maybe we should prepend the remote cluster id in DRNormalPartitionBufferReceiver.java, Line 360, the WARN message?

manjujames: I think I like 'local' and 'remote' better than 'producer' and 'consumer' because there are producers and consumers on both sides. And it works in both xdcr and non-xdcr modes. Otherwise, we will have to use 'replica' for non-xdcr and 'consumer' for non-xdcr.

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -54,8 +54,18 @@
             "sqltext", "querytype", "readonly", "singlepartition", "replicatedtabledml", "iscontentdeterministic", "isorderdeterministic", "nondeterminismdetail",
             "cost", "seqscancount", "explainplan", "tablesread", "tablesupdated", "indexesused", "cachekeyprefix"
             );
+
+    private boolean m_isXDCR;
+    private byte m_remoteClusterId;
+
     public DRCatalogDiffEngine(Catalog localCatalog, Catalog remoteCatalog) {
         super(localCatalog, remoteCatalog);
+        localCatalog.getClusters().get("cluster").getDrclusterid();
```
slowsoul: This seems not needed.

PR #4550, opened 2017-05-23
ENG-12354, inject random sleep into message handling loop to simulate network delay.

luyangco:
Change-Id: I1eac15480b301d892fb5b16f358e240f238f49e8

### DIFF
--- a/tests/frontend/org/voltcore/agreement/MiniSite.java
+++ b/tests/frontend/org/voltcore/agreement/MiniSite.java
@@ -25,6 +25,7 @@ package org.voltcore.agreement;
 
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -50,9 +51,11 @@ class MiniSite extends Thread implements MeshAide
     Set<Long> m_initialHSIds = new HashSet<Long>();
     Set<Long> m_currentHSIds = new HashSet<Long>();
     Set<Long> m_failedHSIds = new HashSet<Long>();
+    Random m_rand;
+    final private static int MAX_INJECTED_DELAY = 10; // in milliseconds, for error injection only
 
     MiniSite(Mailbox mbox, Set<Long> HSIds, DisconnectFailedHostsCallback callback,
-            VoltLogger logger)
+            VoltLogger logger, long givenSeed)
     {
         m_siteLog = logger;
         m_initialHSIds.addAll(HSIds);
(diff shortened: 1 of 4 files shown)


tests/frontend/org/voltcore/agreement/MiniSite.java
```diff
@@ -50,16 +51,24 @@
...
         m_siteLog = logger;
         m_initialHSIds.addAll(HSIds);
         m_currentHSIds.addAll(HSIds);
         m_mailbox = mbox;
         m_arbiter = new MeshArbiter(mbox.getHSId(), mbox, this);
         m_failedHosts = callback;
+        long seed = System.currentTimeMillis() + mbox.getHSId();
+        if (givenSeed != 0L) {
+            seed = givenSeed;
+        }
```
zeemanhe: All the seeds in the tests are zero, is there a need to pass int the seed?  can just use System.currentTimeMillis() as seed to reduce all the changes?

tests/frontend/org/voltcore/agreement/TestFuzzMeshArbiter.java
```diff
@@ -72,18 +72,20 @@ MiniNode getNode(int i)
...
             long HSId = getHSId(i);
             m_nodes.put(HSId, null);
+            seedMap.put(HSId, seeds[i]);
         }
         Set<Long> HSIds = m_nodes.keySet();
         for (long HSId : HSIds) {
-            m_nodes.put(HSId, new MiniNode(HSId, HSIds, m_fakeMesh));
+            m_nodes.put(HSId, new MiniNode(HSId, HSIds, m_fakeMesh, seedMap.get(HSId)));
             m_nodes.get(HSId).start();
         }
```
zeemanhe: This method should be simplified.  One loop should be enough. local variable seedMap is not needed.

tests/frontend/org/voltcore/agreement/MiniSite.java
luyangco: Like the comments say, the seed array in these tests is a convenient way to reproduce any issue. If the unit test failed in Jenkins, you can grab the seeds printed in log files, fill in the array and re-run the test. It makes random test easier to reproduce.

tests/frontend/org/voltcore/agreement/TestFuzzMeshArbiter.java
luyangco: No I can't. MiniNode needs to know all the peers, which the list can only be get after 1st for loop.

zeemanhe: I got it.  One way to simplify it a little bit is to get an array of HSIDs on the first loop.  Then loop through the array of HSIDs.  This way could avoid updating m_nodes twice and seedMap.

luyangco: OK. I'll make the change.

luyangco:
Just use `System.currentTimeMillis()` is not enough, as all the MiniSites are started almost in the same time and every MiniSite needs to its own unique seed.

PR #4554, opened 2017-05-24
ENG-12385: print tuple schema information when values exceeding varch…

xinjiacs:
…ar length limit


xinjiacs:
ENG-12561 is created to track the ultimate goals of this error message. So close this PR because the final change may revert it eventually.

PR #4555, opened 2017-05-24
Eng 12112

vittvolt:
Added print out message when the properties file is empty or does not contain necessary field. This resolves the print out message problem when the path.properties is empty in some case.


xinjiacs:
I've reviewed Tianhe's code and talked with him in person.

PR #4559, opened 2017-05-24
ENG-12516 Modified CSVLoader to be topology-aware

sycai:


src/frontend/org/voltdb/utils/CSVLoader.java
```diff
@@ -574,10 +574,18 @@ private static void configuration() {
...
+        config.setTopologyChangeAware(true); // Set client to be topology-aware
         final Client client = ClientFactory.createClient(config);
-
         for (String server : servers) {
-            client.createConnection(server.trim(), port);
+            // Try connecting servers one by one until we have a success
+            try {
+                client.createConnection(server.trim(), port);
+                break;
+            } catch(Exception e) { /* Do Nothing */ }
```
brians127: Right now you're swallowing all exceptions, which may hide problems that have nothing to do with servers being unavailable. Perhaps check for IOException and UnknownHostException, since createConnection() declares that it can throw them?

src/frontend/org/voltdb/utils/CSVLoader.java
```diff
@@ -574,10 +574,18 @@ private static void configuration() {
...
+        }
+        
+        if (client.getConnectedHostList().isEmpty()) {
+            throw new Exception("Fail to connect any server.");
```
brians127: Wording could be improved. Perhaps use: "Unable to connect to any servers."

PR #4570, opened 2017-05-30
ENG-11971: Large Temp Table Classes and Unit Test

wolffcm:
This work implements a LargeTempTable class that uses a cache for storing tuple blocks (and associated variable-length data) to disk via a all to the Topend.  This new class isn't yet used via the VoltDB server, just via a unit test.


src/ee/storage/LargeTempTableBlock.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+LargeTempTableBlock::LargeTempTableBlock(int64_t id, LargeTempTable *ltt)
+    : m_id(id)
+    , m_pool(new Pool(ltt->getTableAllocationSize() / 4, 1))
+    , m_tupleBlockPointer(new TupleBlock(ltt, TBBucketPtr()))
+    , m_isPinned(false)
+{
+    LargeTempTableBlockCache* lttBlockCache = ExecutorContext::getExecutorContext()->lttBlockCache();
+    lttBlockCache->increaseAllocatedMemory(getAllocatedMemory());
+}
```
BillWhite: This seemed hard to understand.  Is allocated memory the memory actually in use or the memory which would be in use if all the blocks were resident?  I thought it was the first until I read the code in some detail.  Maybe the terminology could be changed, or at least you could add some commentary describing exactly what you mean.

src/ee/storage/LargeTempTableIterator.h
```diff
@@ -0,0 +1,130 @@
...
+
+    std::vector<int64_t>::iterator m_blockIdIterator;
+    LargeTempTableBlock* m_currBlock;
+    int64_t m_currBlockId;
+    char * m_dataPtr;
+    uint32_t m_blockOffset;
+};
+
+
+bool LargeTempTableIterator::next(TableTuple& out) {
```
BillWhite: If there was a hasNext operation for iterators it would make programming these much easier.

There's nothing to be done here in this ticket.  I was just thinking about this as I read through the code.

src/ee/storage/LargeTempTableIterator.h
```diff
@@ -0,0 +1,130 @@
...
+    if (m_foundTuples < m_activeTuples) {
+        if (m_currBlock == NULL ||
+            m_blockOffset >= m_currBlock->unusedTupleBoundary()) {
+            LargeTempTableBlockCache* lttCache = ExecutorContext::getExecutorContext()->lttBlockCache();
```
BillWhite: Does the second of these two disjuncts mean the block offset is past the used data in the current block?  Given that m_foundTuples < m_activeTuples, does that mean there is no useful data in m_currBlock?

I'm just asking to see if I understand it.

src/ee/storage/TupleBlock.h
```diff
@@ -235,6 +235,14 @@ class TupleBlock {
     inline TBBucketPtr currentBucket() {
         return m_bucket;
     }
+
+    /**
+     * Return the number of bytes in the storage for this
+     * tuple block.
+     */
```
BillWhite: Maybe better would be "Return the maximum number of bytes in this tuple block which may actually used for tuples."
The phrase "number of bytes in the storage for this tuple block" seems odd to me.

src/ee/common/LargeTempTableBlockCache.h
```diff
@@ -0,0 +1,145 @@
...
+
+    /** The large temp table for this block is being destroyed, so
+        release all resources associated with this block. */
+    void releaseBlock(int64_t blockId);
+
+    /** Called from LargeTempTableBlock.  Increase the amount of
+        memory in use by the cache, and store a block to disk if
+        necessary to make more room. */
+    void increaseAllocatedMemory(int64_t numBytes);
+
```
wolffcm: It is a delta.  I'll update the comment.

src/ee/common/LargeTempTableBlockCache.cpp
```diff
@@ -0,0 +1,127 @@
...
+    m_idToBlockMap[id] = it;
+    (*it)->pin();
+
+    return std::make_pair(id, m_blockList.front().get());
+}
+
+LargeTempTableBlock* LargeTempTableBlockCache::fetchBlock(int64_t blockId) {
+    auto mapIt = m_idToBlockMap.find(blockId);
+    assert (mapIt != m_idToBlockMap.end());
+    auto listIt = mapIt->second;
```
wolffcm: That's right.  Even blocks that are stored on disk are tracked by the cache and it will load them if requested.  I'll update the comment.

wolffcm: I updated this method and related methods to throw an exception if the block ID does not exist.

src/ee/common/LargeTempTableBlockCache.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (! (*listIt)->isResident()) {
+        Topend* topend = ExecutorContext::getExecutorContext()->getTopend();
+        bool rc = topend->loadLargeTempTableBlock((*listIt)->id(), listIt->get());
+        assert(rc);
```
wolffcm: I used `FixUnusedAssertHack.h` which addresses this (casts the `assert` argument to void).

src/ee/storage/LargeTempTableBlock.cpp
wolffcm: The allocated memory is the memory that is actually in use.  Tuple blocks allocates memory for the block upon construction, so it doesn't actually grow as tuples are inserted. 

The pool does grow as tuples containing out-of-line values are inserted, as new chunks are allocated.  But it also pre-allocates a chunk.

I'll add some comments that clarify all this.

src/ee/common/LargeTempTableBlockCache.cpp
```diff
@@ -0,0 +1,127 @@
...
+void LargeTempTableBlockCache::unpinBlock(int64_t blockId) {
+    (*m_idToBlockMap[blockId])->unpin();
+}
+
+void LargeTempTableBlockCache::releaseBlock(int64_t blockId) {
+    auto it = m_idToBlockMap[blockId];
+    if (! (*it)->isResident()) {
+        Topend* topend = ExecutorContext::getExecutorContext()->getTopend();
+        bool rc = topend->releaseLargeTempTableBlock(blockId);
+        assert(rc);
```
wolffcm: No, it won't see above.

src/ee/storage/LargeTempTableIterator.h
wolffcm: I agree.  `TableIterator` actually has a `hasNext` but I don't think it's really used even by its own `next` method.

src/ee/storage/LargeTempTableIterator.h
wolffcm: Not quite.  `m_foundTuples` and `m_activeTuples` are counts of tuples across all blocks.  `m_blockOffset` is the ordinal position of the current tuple in the current block.  Once we've returned all the tuples in the block, `m_blockOffset >= m_currBlock->unusedTupleBoundary()` will be true, so we need to transition to the next current block.

src/ee/storage/TupleBlock.h
wolffcm: Ok, I fixed it.

PR #4574, opened 2017-05-31
ENG-12391: Allow init to initialize with no schema.

akhanzode:


tests/frontend/org/voltdb/regressionsuites/TestStartWithClassesOnly.java
```diff
@@ -0,0 +1,154 @@
...
+                assertTrue(nameForDebugging + " does not contain " + artifact, files.contains(artifact));
+            }
+        }
+    }
+
+    /** Counts how many nodes contain staged catalogs (user-initialized schemas)
+     * @param cluster
+     * @return number of nodes in cluster that have staged catalogs
+     */
+    static int countNodesWithStagedCatalog(LocalCluster cluster) throws IOException {
```
brians127: This should be shared code with TestStartWithSchema

tests/frontend/org/voltdb/regressionsuites/TestStartWithClassesOnly.java
```diff
@@ -0,0 +1,154 @@
...
+ * 'init' tests can be found in TestInitStartAction
+ */
+final public class TestStartWithClassesOnly {
+
+    static final int siteCount = 1;
+    static final int hostCount = 3;
+    static final int kfactor   = 0;
+    static final int clusterID = 0;
+
+    static void ensureAllCatalogDefaultArtifactsExists(InMemoryJarfile jarFile, String nameForDebugging) {
```
brians127: This should be shared code with TestStartWithSchema

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -301,10 +301,7 @@ public LocalCluster(String schemaToStage,
                         boolean isRejoinTest,
                         Map<String, String> env)
     {
-        if (schemaToStage == null) {
-            assert catalogJarFileName != null : "Catalog jar file name is null";
-            setNewCli(isNewCli);
-        } else {
+        if (schemaToStage != null) {
```
brians127: How does LocalCluster know that it's an 'init with schema and/or classes' test if schemaToStage is not provided? I don't think we want those running with OldCLI and there are other tweaks as well.

tests/frontend/org/voltdb/regressionsuites/TestStartWithClassesOnly.java
```diff
@@ -0,0 +1,154 @@
...
+        // Catalog compilation is taken care of by VoltDB itself - no need to do so explicitly.
+        cluster = new LocalCluster(
+                null,
+                null,
+                siteCount,
+                hostCount,
+                kfactor,
+                clusterID,
+                BackendTarget.NATIVE_EE_JNI,
+                FailureState.ALL_RUNNING, false, false, null);
```
brians127: I don't see how this loads the classes. You may be getting a false test pass if SelectStarHelloWorld is on the server's CLASSPATH. Can you explain or fix this?

tests/frontend/org/voltdb/TestInitStartAction.java
```diff
@@ -447,6 +472,42 @@ public void testInitWithClassesAndArtifacts() throws Exception {
...
+        int classesFound = 0;
+        for (ClassInfo myclass : classpath.getTopLevelClassesRecursive(packageName)) {
+            compiler.addClassToJar(originalInMemoryJar, myclass.load());
+            classesFound++;
+        }
+        // check that classes were found and loaded. If another test modifies "fakeusecase.greetings" it should modify this assert also.
+        assertEquals(5, classesFound);
+
+        System.out.println("Writing " + classesFound + " classes to jar file");
+        File classesJarfile = File.createTempFile("TestInitStartWithClasses-procedures", ".jar");
```
brians127: Please update the temp file name to reflect the test name

tests/frontend/org/voltdb/regressionsuites/TestStartWithClassesOnly.java
akhanzode: This is copied from your tests how did it work for you?

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
akhanzode: Again this is your old code. I will remove the assert as thats not useful but then it didnt look right why set setNewCli to isNewCli? Can you give more context so I could do it in a way that's understandable.

brians127: The explicit setting of newCli is safe to remove. I did that to make it more obvious that the behavior is different when staged schema is not null.

brians127: On further investigation: classpath works correctly, but m_usesStagedSchema is still not set when "init with classes" is used. This affects several asserts, enables newCli, and stops clearLocalDataDirectories from being forced to true by the ServerThread. Only the NewCLI problem impacts your test directly, but the way the code is now is confusing. 

The correct location for staged classes in the configuration is m_stagedClassesPath. LocalCluster's catalogJarFileName routes to CommandLine.jarFileName() and m_pathToCatalog. I failed to update the code in LocalCluster to allow for this when I moved 'init --classes' from the CLASSPATH to the command line because none of the tests I added use LocalCluster.

tests/frontend/org/voltdb/regressionsuites/TestStartWithClassesOnly.java
brians127: I did. Disregard.

tests/frontend/org/voltdb/TestInitStartAction.java
```diff
@@ -447,6 +472,42 @@ public void testInitWithClassesAndArtifacts() throws Exception {
         validateStagedCatalog(schema, originalInMemoryJar);
     }
 
+    /** Tests that when there are base classes and non-class files in the stored procedures,
+     * that these also exist in the staged catalog.
+     * @throws Exception upon failure or error
+     */
+    @Test
+    public void testInitWithClassesAndNoSchema() throws Exception {
```
brians127: This is good.

tests/frontend/org/voltdb/regressionsuites/TestStartWithClassesOnly.java
brians127: I misspoke. LocalCluster is not getting passed a jar file with classes, so it is using the VoltDB classpath.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -315,6 +312,9 @@ public LocalCluster(String schemaToStage,
                 throw new RuntimeException(e);
             }
         }
+        if (catalogJarFileName != null) {
+            m_usesStagedSchema = true;
+        }
```
brians127: I think this changes the behavior of existing tests which pass a full catalog to LocalCluster. They will instead be passing a classes jar file, which should mean schemas are missing in affected tests. They also are now forced to use NewCLI.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -1075,6 +1079,9 @@ void startOne(int hostId, boolean clearLocalDataDirectories,
             cmdln.pathToDeployment(null);
             cmdln.setForceVoltdbCreate(clearLocalDataDirectories);
         }
+        if (m_usesStagedSchema) {
+            cmdln.m_pathToCatalog = null;
```
brians127: You also need to set cmdln.m_stagedClassesPath to the catalog and update the CommandLine copy constructor if it hasn't been already. Once this is done I think I'll be ready to approve. Thanks!

PR #4580, opened 2017-06-01
Eng 5532 read-only procedure non-deterministic no warning

vittvolt:
The problem has already been addressed in various previous commits. The non-deterministic warning is checked in summarizeSuccess() method and checks if the query is read-only or not.

This PR adds a test method for checking the output of summarizeSuccess() and see if any unwanted warning is given for any non-deterministic read-only queries.

### DIFF
--- a/tests/frontend/org/voltdb/planner/TestDeterminism.java
+++ b/tests/frontend/org/voltdb/planner/TestDeterminism.java
@@ -1101,8 +1101,10 @@ public class TestDeterminism extends PlannerTestCase {
         System.out.println(outputStream.toString());
         // There is no warnings generated for read-only ND queries
         String msg = outputStream.toString();
-        assert(!msg.contains("ND"));
-        assert(!msg.contains("WARN"));
+        assertFalse(stringContains(msg, "ND"));
+        assertFalse(stringContains(msg, "WARN"));
+        assertTrue(stringContains(msg, "Successfully created"));
+        assertTrue(stringContains(msg, "Catalog contains"));
         outputStream.close();
     }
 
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/planner/testplans-determinism-read-only.sql
```diff
@@ -0,0 +1,9 @@
+create table test (a int not null);
+partition table test on column a;
+
+create procedure tp as select top 3 * from test;
+
+create table test2 (a int, b int not null, primary key (a, b));
+partition table test2 on column b;
+
+create procedure tp2 as select top 1 * from test2;
```
xinjiacs: You may not want to use the same query pattern in this file.

Maybe:
select b, count(*) from test2 group by b limit 3;

tests/frontend/org/voltdb/planner/TestDeterminism.java
```diff
@@ -1074,6 +1082,30 @@ public void testDeterminismForUniqueConstraintEquivalence() {
...
+        URL path = TestDeterminism.class.getResource("testplans-determinism-read-only.sql");
+        String pathStr = URLDecoder.decode(path.getPath(), "UTF-8");
+        compiler.loadSchema(hsql, all_procs, pathStr);
+
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        compiler.summarizeSuccess(new PrintStream(outputStream), null, "");
+        System.out.println(outputStream.toString());
+        // There is no warnings generated for read-only ND queries
+        String msg = outputStream.toString();
+        assert(!msg.contains("ND"));
```
xinjiacs: Please use the assert related functions from Junit instead.

Also, can you also verify some messages exists in the output stream?

PR #4583, opened 2017-06-02
ENG-12542

sycai:
File "voltdbroot/config/path.properties" now stores relative paths if the deployment does not set an absolute path.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -2550,14 +2550,16 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
             case GET:
                 // once a voltdbroot is inited, the path properties contain the true path values
                 Settings.initialize(config.m_voltdbRoot);
-                // only override the local sites count
-                nodeSettings = NodeSettings.create(config.asNodeSettingsMap());
+                // override the local sites count and paths
+                nodeSettings = NodeSettings.create(config.asNodeSettingsMap(),
+                        config.asRelativePathSettingsMap());
                 break;
             case PROBE:
                 // once a voltdbroot is inited, the path properties contain the true path values
                 Settings.initialize(config.m_voltdbRoot);
-                // only override the local sites count
-                nodeSettings = NodeSettings.create(config.asNodeSettingsMap());
+                // override the local sites count and paths
+                nodeSettings = NodeSettings.create(config.asNodeSettingsMap(),
+                        config.asRelativePathSettingsMap());
                 File nodeSettingsFH = new File(getConfigDirectory(config), "path.properties");
                 consoleLog.info("Loaded node-specific settings from " + nodeSettingsFH.getPath());
                 hostLog.info("Loaded node-specific settings from " + nodeSettingsFH.getPath());
--- a/src/frontend/org/voltdb/settings/NodeSettings.java
+++ b/src/frontend/org/voltdb/settings/NodeSettings.java
@@ -193,9 +193,15 @@ public interface NodeSettings extends Settings {
                 throw new SettingsException("Missing VoltDB root " +
                                             "information in path.properties file.");
             }
-            mb.put(VOLTDBROOT_PATH_KEY, getVoltDBRoot().getCanonicalPath());
+            File voltdbroot = getVoltDBRoot().getCanonicalFile();
+            mb.put(VOLTDBROOT_PATH_KEY, voltdbroot.getCanonicalPath());
             for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
-                mb.put(e.getKey(), e.getValue().getCanonicalPath());
+                File path = e.getValue();
+                if (path.isAbsolute()) {
+                    mb.put(e.getKey(), path.getCanonicalPath());
+                } else {
+                    mb.put(e.getKey(), voltdbroot.toPath().relativize(path.getCanonicalFile().toPath()).toString());
+                }
             }
             mb.put(LOCAL_SITES_COUNT_KEY, Integer.toString(getLocalSitesCount()));
         } catch (IOException e) {
--- a/src/frontend/org/voltdb/utils/CatalogUtil.java
+++ b/src/frontend/org/voltdb/utils/CatalogUtil.java
@@ -141,6 +141,7 @@ import org.voltdb.plannodes.AbstractPlanNode;
 import org.voltdb.settings.ClusterSettings;
 import org.voltdb.settings.DbSettings;
 import org.voltdb.settings.NodeSettings;
+import org.voltdb.settings.SettingsException;
 import org.voltdb.snmp.DummySnmpTrapSender;
 import org.voltdb.types.ConstraintType;
 import org.xml.sax.SAXException;
@@ -1695,6 +1696,7 @@ public abstract class CatalogUtil {
         File voltDbRoot;
         // Handles default voltdbroot (and completely missing "paths" element).
         voltDbRoot = getVoltDbRoot(paths);
+
         //Snapshot
         setupSnapshotPaths(paths.getSnapshots(), voltDbRoot);
         //export overflow
@@ -1741,7 +1743,8 @@ public abstract class CatalogUtil {
         snapshotPath = new File(VoltDB.instance().getSnapshotPath(paths));
         if (!snapshotPath.isAbsolute())
         {
-            snapshotPath = new VoltFile(voltDbRoot, VoltDB.instance().getSnapshotPath(paths));
+            String sanitizedPath = sanitizePath(VoltDB.instance().getSnapshotPath(paths));
+            snapshotPath = new VoltFile(voltDbRoot, sanitizedPath);
         }
 
         if (!snapshotPath.exists()) {
@@ -1764,7 +1767,8 @@ public abstract class CatalogUtil {
         commandlogPath = new File(VoltDB.instance().getCommandLogPath(paths));
         if (!commandlogPath.isAbsolute())
         {
-            commandlogPath = new VoltFile(voltDbRoot, VoltDB.instance().getCommandLogPath(paths));
+            String sanitizedPath = sanitizePath(VoltDB.instance().getCommandLogPath(paths));
+            commandlogPath = new VoltFile(voltDbRoot, sanitizedPath);
         }
 
         if (!commandlogPath.exists()) {
@@ -1789,7 +1793,8 @@ public abstract class CatalogUtil {
         commandlogSnapshotPath = new File(VoltDB.instance().getCommandLogSnapshotPath(paths));
         if (!commandlogSnapshotPath.isAbsolute())
         {
-            commandlogSnapshotPath = new VoltFile(voltDbRoot, VoltDB.instance().getCommandLogSnapshotPath(paths));
+            String sanitizedPath = sanitizePath(VoltDB.instance().getCommandLogSnapshotPath(paths));
+            commandlogSnapshotPath = new VoltFile(voltDbRoot, sanitizedPath);
         }
 
         if (!commandlogSnapshotPath.exists()) {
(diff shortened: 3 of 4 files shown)


src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -193,9 +192,17 @@ default Properties asProperties() {
...
+            File voltDBRoot = getVoltDBRoot();
+            if (voltDBRoot.isAbsolute()) { // Store absolute paths only if the given voltdbroot path is absolute
+                mb.put(VOLTDBROOT_PATH_KEY, voltDBRoot.getCanonicalPath());
+                for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
+                    mb.put(e.getKey(), e.getValue().getCanonicalPath());
+                }
+            } else { // Otherwise, store relative paths
+                mb.put(VOLTDBROOT_PATH_KEY, voltDBRoot.getPath());
+                for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
+                    mb.put(e.getKey(), e.getValue().getName());
```
brians127: I don't think this will work for all cases. Have you tried a deeply/nested/directory/like/this? getName() only returns the outer most subdirectory, which is only adequate when you control the paths (voltdbroot/log for example).

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -193,9 +192,17 @@ default Properties asProperties() {
                 throw new SettingsException("Missing VoltDB root " +
                                             "information in path.properties file.");
             }
-            mb.put(VOLTDBROOT_PATH_KEY, getVoltDBRoot().getCanonicalPath());
-            for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
-                mb.put(e.getKey(), e.getValue().getCanonicalPath());
+            File voltDBRoot = getVoltDBRoot();
+            if (voltDBRoot.isAbsolute()) { // Store absolute paths only if the given voltdbroot path is absolute
```
brians127: I think these comments are helpful. You should consider moving them to the next line if that makes them easier to see

src/frontend/org/voltdb/settings/NodeSettings.java
sycai: I've just manually tested your deeply-nested-directory scenario and it worked. It seems that VoltDB always translates relative paths to absolute ones while loading the deployment.

src/frontend/org/voltdb/settings/NodeSettings.java
sycai: Sounds good. I will move these comments.

src/frontend/org/voltdb/settings/NodeSettings.java
brians127: I don't think I understand. You've tested this and it works because Volt always uses absolute paths? How did you exercise your new code which uses URI.relativeize()?

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -29,13 +29,12 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.aeonbits.owner.Config.Sources;
+import org.aeonbits.owner.ConfigFactory;
```
brians127: What is this package?! It's not clear that you're using it, but all new dependencies need to be approved for copyright and maintainability reasons. I always run a "git difftool HEAD" prior to committing to catch slip ups like this.

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -193,9 +192,19 @@ default Properties asProperties() {
                 throw new SettingsException("Missing VoltDB root " +
                                             "information in path.properties file.");
             }
-            mb.put(VOLTDBROOT_PATH_KEY, getVoltDBRoot().getCanonicalPath());
-            for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
-                mb.put(e.getKey(), e.getValue().getCanonicalPath());
+            File voltDBRoot = getVoltDBRoot();
+            if (voltDBRoot.isAbsolute()) {
+             // Store absolute paths only if the given voltdbroot path is absolute
```
brians127: Not a blocker: please fix indentation

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -193,9 +192,19 @@ default Properties asProperties() {
...
+                mb.put(VOLTDBROOT_PATH_KEY, voltDBRoot.getCanonicalPath());
+                for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
+                    mb.put(e.getKey(), e.getValue().getCanonicalPath());
+                }
+            } else {
+             // Otherwise, store relative paths to current working directory
```
brians127: Not a blocker: please fix indentation

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1711,10 +1711,10 @@ public static File getVoltDbRoot(PathsType paths) {
...
         File snapshotPath;
         snapshotPath = new File(VoltDB.instance().getSnapshotPath(paths));
-        if (!snapshotPath.isAbsolute())
-        {
-            snapshotPath = new VoltFile(voltDbRoot, VoltDB.instance().getSnapshotPath(paths));
-        }
+//        if (!snapshotPath.isAbsolute())
+//        {
+//            snapshotPath = new VoltFile(voltDbRoot, VoltDB.instance().getSnapshotPath(paths));
+//        }
```
brians127: It's OK to do this while debugging but please do not commit commented out code to master.

src/frontend/org/voltdb/settings/NodeSettings.java
brians127: ALWAYS know what changes you're making! Don't trust the IDE. New dependencies need to get approved by management. It is my belief this could get approved - BSD is a friendly license - but I'm not management.
http://owner.aeonbits.org/docs/license/

akhanzode: aeonbits is already vetted and listed in third-party dependency.

sycai: These two lines were in the original code. The Eclipse moved their positions while I was debugging.

akhanzode:
Can you test following scenarios.

1. voltdb  get deployment
2. update deployment from VMC and the do get of deployment.
3. Can user use this deployment to init at different location just by giving different -D. whats the impact?
4. What happens when voltdbroot is moved and someone does start after moving? Does it work if its relative it should right?
Put all your findings in the jira tickets.

sycai:
Sure, I will test them right away.

brians127:
I didn't realize Eclipse just moved existing dependencies and that no new ones were added. My apologies. Thanks for cleaning up the comments and such. I have no objections to merging once testing is done and the results are reviewed.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2603,51 +2608,61 @@ public static DeploymentType updateRuntimeDeploymentPaths(DeploymentType deploym
         PathsType paths = deployment.getPaths();
         if (paths.getVoltdbroot() == null) {
             PathsType.Voltdbroot root = new PathsType.Voltdbroot();
-            root.setPath(VoltDB.instance().getVoltDBRootPath());
+            root.setPath(new File(VoltDB.instance().getVoltDBRootPath()).getAbsolutePath());
```
brians127: Please use getCanonicalPath() instead of getAbsolutePath() for anything you plan to store in a file.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2603,51 +2608,61 @@ public static DeploymentType updateRuntimeDeploymentPaths(DeploymentType deploym
...
+            droverflow.setPath(relativizeToVoltDBRoot(VoltDB.instance().getDROverflowPath()));
             paths.setDroverflow(droverflow);
         } else {
-            paths.getDroverflow().setPath(VoltDB.instance().getDROverflowPath());
+            paths.getDroverflow().setPath(relativizeToVoltDBRoot(VoltDB.instance().getDROverflowPath()));
         }
         return deployment;
     }
+
+    private static String relativizeToVoltDBRoot(String s) {
```
brians127: A method named "relativize" should never return an absolute path. Please change the name to more accurately reflect the behavior. Perhaps "normalizePath"?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2540,14 +2540,16 @@ ReadDeploymentResults readPrimedDeployment(Configuration config) {
             case GET:
                 // once a voltdbroot is inited, the path properties contain the true path values
                 Settings.initialize(config.m_voltdbRoot);
-                // only override the local sites count
-                nodeSettings = NodeSettings.create(config.asNodeSettingsMap());
+                // override the local sites count and paths
+                nodeSettings = NodeSettings.create(config.asNodeSettingsMap(),
+                        config.asPathSettingsMap());
```
brians127: What do these changes do?

sycai: They guarantee that relative paths in the file path.properties do not change.

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -195,7 +195,12 @@ default Properties asProperties() {
             }
             mb.put(VOLTDBROOT_PATH_KEY, getVoltDBRoot().getCanonicalPath());
             for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
-                mb.put(e.getKey(), e.getValue().getCanonicalPath());
+                File path = e.getValue();
+                if (path.isAbsolute()) {
+                    mb.put(e.getKey(), path.getCanonicalPath());
+                } else {
+                    mb.put(e.getKey(), getVoltDBRoot().getCanonicalFile().toPath().relativize(path.getCanonicalFile().toPath()).toString());
```
nshi: You can cache the voltroot path in a local variable so that it doesn't have to get it in every iteration of the loop.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2601,53 +2608,83 @@ public static DeploymentType shallowClusterAndPathsClone(DeploymentType o) {
...
-        if (paths.getVoltdbroot() == null) {
-            PathsType.Voltdbroot root = new PathsType.Voltdbroot();
-            root.setPath(VoltDB.instance().getVoltDBRootPath());
-            paths.setVoltdbroot(root);
-        } else {
-            paths.getVoltdbroot().setPath(VoltDB.instance().getVoltDBRootPath());
+        try {
+            if (paths.getVoltdbroot() == null) {
+                PathsType.Voltdbroot root = new PathsType.Voltdbroot();
+                root.setPath(new File(VoltDB.instance().getVoltDBRootPath()).getCanonicalPath());
```
nshi: Shouldn't the root path returned by `VoltDB.instance().getVoltDBRootPath()` be already absolute? Why do you have to call `getCanonicalPath()` again?

sycai: Due to the modifications from this branch, the returned voltdbroot path is not guaranteed to be absolute.

nshi: I thought that the root path is always saved in the file by calling `mb.put(VOLTDBROOT_PATH_KEY, voltdbroot.getCanonicalPath());`, which means that it's always absolute. Did I miss something?

sycai:
- Init with absolute path: Relative paths from the deployment file become absolute. Default path values are absolute too.

- Init with relative path: Paths from the deployment file are preserved. Default path values are set to be relative to the voltdbroot. The voltdbroot path itself is absolute.

- Start with absolute path: Path properties stay the same as they are in the "path.properties" file.

- Start with relative path: Path properties stay the same as they are in the "path.properties" file.

- Get with absolute path: Paths in the deployment file are the same as those in the voltdbroot.

- Get with relative path: Paths in the deployment file are the same as those in the voltdbroot.

PR #4588, opened 2017-06-05
ENG-12522: return the decimal when it is expected

xinjiacs:


src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -349,6 +349,9 @@ else if (param == VoltType.NULL_TIMESTAMP ||
         else if (inputClz == BigDecimal.class) {
             // Only conversion to primitive numeric types are considered
             BigDecimal pBigDecimal = (BigDecimal) param;
+            // Set the scale bd = VoltDecimalHelper.setDefaultScale(bd); ?
+            if (expectedClz == BigDecimal.class) return pBigDecimal;
```
vittvolt: Should this be converted to VoltDecimalHelper.setDefaultScale(pBigDecimal) like it's done in the bottom to set the scale?

PR #4590, opened 2017-06-05
Eng 12434 too many projects

BillWhite:
Don't add so many projection nodes.

We sometimes add projection nodes when we don't need them.  This is a performance problem.  This commit adds a microoptimization to avoid this.

The extra projection nodes happen mostly for this reason.  
1. We create a subplan, which is essentially the join tree or scan.
1. We put a receive/send pair on the subplan, in order to denote the distributed fragment.
1. We put things like orderby/limit/window and so forth on top.
1. We sometimes put a projection node on the top, to project out the select list.  Sometimes we can tell that the projection node is not needed, but not always, since we haven't computed output schemas yet.  So we put it on the top more often than is necessary.
1. We decide the plan is only a single fragment plan.  So, we remove the send/receive pair.  But we still can't remove the select projection node.
1. Sometimes we add an insert node to the top, if the plan is for insert into select.  We could inline this insert node if the subplan is a scan node.  But we often can't because there is an extra projection node in the way.

In this commit we do everything that we do above.  But after the best plan is selected and after the output schemas are computed, we look for extra projection nodes.  We then look again for insert nodes which we can make inline.

There is room for more optimization here.  It should be possible to make all projection nodes inline.

https://issues.voltdb.com/browse/ENG-12434


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -1157,10 +1157,11 @@ else if (m_parsedSelect.hasDistinctWithGroupBy() &&
...
         plan.statementGuaranteesDeterminism(hasLimitOrOffset, orderIsDeterministic, contentDeterminismMessage);
 
-        // Apply the micro-optimization:
+        // Apply the select construction phase micro-optimizations:
         // LIMIT push down, Table count / Counting Index, Optimized Min/Max
-        MicroOptimizationRunner.applyAll(plan, m_parsedSelect);
-
+        MicroOptimizationRunner.applyAll(plan,
+                                         m_parsedSelect,
+                                         MicroOptimizationRunner.Phases.SELECT_CONSTRUCTION_PHASE);
```
BillWhite: We have to apply different microoptimizations at different phases.  By default they are applies when creating select statements.  But we can also apply them after the best plan has been selected and the output schemas have been calculated.

src/frontend/org/voltdb/planner/QueryPlanner.java
```diff
@@ -436,7 +437,11 @@ private CompiledPlan compileFromXML(VoltXMLElement xmlSQL, String[] paramValues)
         // Execute the generateOutputSchema and resolveColumnIndexes once for the best plan
         bestPlan.rootPlanGraph.generateOutputSchema(m_db);
         bestPlan.rootPlanGraph.resolveColumnIndexes();
-
+        // Now that the plan is all together we
+        // can compute the best selection microoptimizations.
+        MicroOptimizationRunner.applyAll(bestPlan,
+                                         parsedStmt,
+                                         MicroOptimizationRunner.Phases.AFTER_BEST_SELECTION);
```
BillWhite: Again, these are the later microoptimizations.

tests/frontend/org/voltdb/planner/TestIndexReverseScan.java
```diff
@@ -484,8 +484,6 @@ private void checkReverseScan(String indexName, IndexLookupType lookupType,
         assertTrue(pn instanceof SendPlanNode);
         pn = pn.getChild(0);
         if (needOrderby) {
```
BillWhite: Note that a lot of these these planner tests, below this one, are changed by making projection nodes optional.

tests/scripts/plannertester/ENG-2549/baseline/ENG-2549.plan0
```diff
@@ -1,2 +1,2 @@
...
...]},{"ID":7,"PLAN_NODE_TYPE":"LIMIT","OFFSET":0,"LIMIT":-1,"OFFSET_PARAM_IDX":-1,"LIMIT_PARAM_IDX":-1,"LIMIT_EXPRESSION":{"TYPE":1,"VALUE_TYPE":5,"LEFT":{"TYPE":30,"VALUE_TYPE":6,"ISNULL":false,"VALUE":0},"RIGHT":{"TYPE":31,"VALUE_TYPE":6,"PARAM_IDX":1}}}],"TARGET_TABLE_NAME":"L","TARGET_TABLE_ALIAS":"L","LOOKUP_TYPE":"GTE","SORT_DIRECTION":"ASC","TARGET_INDEX_NAME":"IDX_B","SEARCHKEY_EXPRESSIONS":[{"TYPE":31,"VALUE_TYPE":9,"VALUE_SIZE":0,"PARAM_IDX":0},{"TYPE":30,"VALUE_TYPE":5,"ISNULL":false,"VALUE":0}],"COMPARE_NOTDISTINCT":[false,false],"END_EXPRESSION":{"TYPE":20,"VALUE_TYPE":23,"LEFT":{"TYPE":10,"VALUE_TYPE":23,"LEFT":{"TYPE":32,"VALUE_TYPE":3,"COLUMN_IDX":3},"RIGHT":{"TYPE":30,"VALUE_TYPE":5,"ISNULL":false,"VALUE":0}},"RIGHT":{"TYPE":10,"VALUE_TYPE":23,"LEFT":{"TYPE":32,"VALUE_TYPE":9,"VALUE_SIZE":32,"COLUMN_IDX":1},"RIGHT":{"TYPE":31,"VALUE_TYPE":9,"VALUE_SIZE":0,"PARAM_IDX":0}}},"SKIP_NULL_PREDICATE":{"TYPE":20,"VALUE_TYPE":23,"LEFT":{"TYPE":9,"VALUE_TYPE":23,"LEFT":{"TYPE":32,"VALUE_TYPE":3,"COLUMN_IDX":3}},"RIGHT":{"TYPE":10,"VALUE_TYPE":23,"LEFT":{"TYPE":32,"VALUE_TYPE":9,"VALUE_SIZE":32,"COLUMN_IDX":1},"RIGHT":{"TYPE":31,"VALUE_TYPE":9,"VALUE_SIZE":0,"PARAM_IDX":0}}}}]}
```
BillWhite: All of these plans differ by removing PROJECTION plan nodes.  It's hard to see that.

src/frontend/org/voltdb/planner/microoptimizations/RemoveUnnecessaryProjectNodes.java
```diff
@@ -0,0 +1,98 @@
...
+                    parent.setAndLinkChild(parentIndex, child);
+                }
+                plan = child;
+            } else {
+                break;
+            }
+        }
+        for (int idx = 0; idx < plan.getChildCount(); idx += 1) {
+            AbstractPlanNode child = plan.getChild(idx);
+            AbstractPlanNode newChild = recursivelyApply(child, idx);
```
wolffcm: I guess this loop will remove redundant project nodes that may be in subqueries (meaning derived tables)?  Would be good to have tests for that, I'll keep an eye out.

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -476,7 +476,13 @@ public void addAndLinkChild(AbstractPlanNode child) {
         child.m_parents.add(this);
     }
 
-    // called by PushDownLimit, re-link the child without changing the order
+    /**
+     * Used to re-link the child without changing the order.
```
wolffcm: The order of what?  The order of the plan node's children I guess?

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -439,12 +444,28 @@ static protected void assertClassesMatchNodeChain(
...
         assertEquals(startType, result.getPlanNodeType());
-        for (PlanNodeType type : nodeTypes) {
+        for (int nodeNumber = 0; nodeNumber < nodeTypes.length; nodeNumber += 1) {
+            PlanNodeType type = nodeTypes[nodeNumber];
             assertTrue(result.getChildCount() > 0);
             result = result.getChild(0);
-            assertEquals(type, result.getPlanNodeType());
+            if (hasOptionalProjection
+                    && (nodeNumber == 0)
+                    && (result.getPlanNodeType() == PlanNodeType.PROJECTION)) {
```
wolffcm: So the effect of setting the hasOptionalProjection flag is that the projection node may be there or it may not be?  Wouldn't it be better to lock it down as not there if that is in fact the correct and more optimal plan?

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -532,6 +554,9 @@ protected void printPlanNodes(AbstractPlanNode root, int fragmentNumber, int num
             for (Entry<PlanNodeType, AbstractPlanNode> entry : root.getInlinePlanNodes().entrySet()) {
                 System.out.printf("        Inline %s\n", entry.getKey());
             }
+            for (Entry<PlanNodeType, AbstractPlanNode> entry : root.getInlinePlanNodes().entrySet()) {
```
wolffcm: This for loop seems to do exactly the same thing as the loop immediately above.  Cut and paste error?

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -618,6 +656,21 @@ protected static void assertTopDownTree(AbstractPlanNode start,
...
                 continue;
             }
+            // If we have specified that this has an optional
+            // projection node, then we may find a projection
+            // node at node number 1.  This is from the select
+            // list projection, but we may have optimized it away.
+            if (nodeNumber == 1 && hasOptionalProjection) {
+                if (PlanNodeType.PROJECTION == parent.getPlanNodeType()) {
+                    assertEquals(1, parent.getChildCount());
+                    // We don't actually want to use up this
```
wolffcm: Again, it seems strange that this test is fuzzy on whether the projection needs to actually be there. If you have derived tables in the FROM clause in the table, you might also have missing projections somewhere in the middle of the plan tree?

tests/frontend/org/voltdb/planner/TestIndexSelection.java
```diff
@@ -574,7 +586,11 @@ private void checkDualIndexedJoin(AbstractPlanNode pn,
             String leftIndexName, String rightIndexName, int nJoinKeys) {
         IndexScanPlanNode ispn;
         // Skip the Send and Projection plan nodes.
-        pn = pn.getChild(0).getChild(0);
+        assertEquals(PlanNodeType.SEND, pn.getPlanNodeType());
+        pn = pn.getChild(0);
+        if (PlanNodeType.PROJECTION == pn.getPlanNodeType()) {
+            pn = pn.getChild(0);
```
wolffcm: I continue to be disconcerted by test logic that basically says, "if there's a projection node, just skip it."  Either it should be there, because it's necessary, or it shouldn't.  Right?

BillWhite:
We are rethinking this.

PR #4599, opened 2017-06-08
Eng 11957 fix hashmismatch backportv5.8

luyangco:
When HSQLDB parse the sql expression, it uses a hash map to cache expression ids it has already computed. But it uses the hash code of object instead of the object itself as the key, which is not guaranteed to be unique for different instances. In some rare cases, this will cause hash key collision which let the SQL parser returns incorrect columns for values to be inserted into the table.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/SubQuery.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/SubQuery.java
@@ -277,4 +277,43 @@ class SubQuery implements ObjectComparator {
                                           : -1;
         }
     }
+
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + level;
+        result = prime * result + ((queryExpression == null) ? 0 : queryExpression.hashCode());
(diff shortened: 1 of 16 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/SubQuery.java
```diff
@@ -277,4 +277,43 @@ public int compare(Object a, Object b) {
...
     }
+
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + level;
+        result = prime * result + ((queryExpression == null) ? 0 : queryExpression.hashCode());
+        result = prime * result + ((database == null) ? 0 : database.hashCode());
+        result = prime * result + ((view == null) ? 0 : view.hashCode());
+        result = prime * result + (isExistsPredicate ? 1231 : 1237);
```
xinjiacs: You probably want to use a different prime number, maybe 7, 17? 

This algorithm will easily overflow the MAX integer.

PR #4601, opened 2017-06-08
ENG-12450 fixed command line naming convention conflicts

GuaGuaDean:
None of our command line arguments are mixed case, except the short names that can be a single uppercase character. Otherwise, the options are all lowercase. So I changed "--characterSet" to "--charset"


luyangco:
Looks good to me.

PR #4603, opened 2017-06-09
ENG-12572:

wweiss-voltdb:
InMemoryJarfile has been extended to allow for the inclusion of non-class resources to be included in the catalog or the jar file specified through loadclasses. The restriction is that files added at the root of the jar file will be ignored. Class files will continue to be loaded from any package in the jar file including the root. Also note that .sql files will be included in the catalog but will no longer be loaded if they are not in the root directory of the jar file.

Assuming that the resource file is named "info.txt" under the "Resources" package/directory the following Java code will correctly find the file from a stored procedure:
this.getClass().getResource("/Resources/info.txt")
this.getClass().getClassLoader().getResource("Resources/info.txt")
this.getClass().getResourceAsStream("/Resources/info.txt")
this.getClass().getClassLoader().getResourceAsStream("Resources/info.txt")

Note that in order for class.getResource to correctly find the file using a relative path the resource has to be defined inside the package containing the stored procedure that is accessing the resource. Otherwise the absolute path within the jar file must be specified.

Also verified that on recover the resource file is still loaded from a snapshot.


src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -306,6 +308,32 @@ public void removeClassFromJar(String classname)
...
     }
 
+    class InMemoryJarHandler extends URLStreamHandler {
+        @Override
+        protected URLConnection openConnection(URL u) throws IOException {
+            return new InMemoryJarUrlConnection(u);
+        }
+    }
+
+    class InMemoryJarUrlConnection extends URLConnection {
```
jhugg: Comment for what this does?

src/frontend/org/voltdb/utils/InMemoryJarfile.java
```diff
@@ -386,6 +414,15 @@ public InMemoryJarfile getInMemoryJarfile() {
         public void initFrom(JarLoader loader) {
             m_classNames.addAll(loader.getClassNames());
         }
+
+        @Override
+        protected URL findResource(String name) {
```
jhugg: Again, a brief comment explaining what this does would be neat.

jhugg:
So this can probably merge for 7.4 if walter vouches for it, but it needs tests to be a documented feature. Can we put this in 7.4, but not announce it until 7.4 to anyone but the specific customer who needs it?

xinjiacs:
Test with update source files?

PR #4604, opened 2017-06-09
TestUserDefinedFunctions.java: added a new class containing many

johnTcrawford:
different UDF's (user-defined functions) to be used for testing.


tests/testfuncs/org/voltdb_testfuncs/TestUserDefinedFunctions.java
```diff
@@ -0,0 +1,1127 @@
...
+        case -102:  return VoltType.NULL_TINYINT;   // the TINYINT (byte) null value (-128)
+        case -103:  return VoltType.NULL_SMALLINT;  // the SMALLINT (short) null value (-32768)
+        case -104:  return VoltType.NULL_INTEGER;   // the INTEGER (int) null value (-2147483648)
+        case -105:  return VoltType.NULL_BIGINT;    // the BIGINT (long) null value (-9223372036854775808L)
+        case -106:  return VoltType.NULL_FLOAT;     // the FLOAT (double) null value (-1.7E+308D)
+        case -107:  return new BigDecimal("-99999999999999999999999999.999999999999");  // the DECIMAL (BigDecimal) minimum value
+        case -108:  return new BigDecimal("-100000000000000000000000000");  // a DECIMAL (BigDecimal) null value (??)
+        case -109:  return Math.log(value.doubleValue());   // Return NaN
+        case -110:  return uninitialized.intValue();        // Throw a NullPointerException (NPE)
+        case -111:  Character.toChars(intValue);            // Throw an IllegalArgumentException
```
yzhang1991: I think this function introduces many magic numbers, making the test case hard to read. Like:
>What is testExceptionsByValue(-111) testing?

Can we add some kind of enumerations? Just to improve readability.

tests/testfuncs/org/voltdb_testfuncs/TestUserDefinedFunctions.java
```diff
@@ -0,0 +1,1127 @@
...
+import org.voltdb.types.GeographyValue;
+import org.voltdb.types.TimestampType;
+
+public class TestUserDefinedFunctions {
+
+    // We start with private methods that are called by (some of) the test
+    // UDF's (user-defined functions) below:
+
+    /**
+     * Returns the VoltDB <b>null</n> value that corresponds to the data type
```
johnTcrawford: Oops, yes; fixed in new check-in.

tests/testfuncs/org/voltdb_testfuncs/TestUserDefinedFunctions.java
johnTcrawford: Yes, good idea, though I used simple (public static final) int constants (defined in a UDF_TEST static class), since enum.getValue() does not work in a case statement (values must be not only constant, but obviously so, to the compiler).

PR #4608, opened 2017-06-12
ENG-11855-unary-minus

dm89:
add support for unary minus


tests/ee/expressions/expression_test.cpp
```diff
@@ -329,6 +329,38 @@ TEST_F(ExpressionTest, SimpleAddition) {
...
+TEST_F(ExpressionTest, SimpleUnaryMinus) {
+    queue<AE*> e;
+    TableTuple junk;
+
+    // -5
+    e.push(new CV(EXPRESSION_TYPE_VALUE_CONSTANT, VALUE_TYPE_TINYINT, 1, (int64_t)5));
+    e.push(new AE(EXPRESSION_TYPE_OPERATOR_UNARY_MINUS, VALUE_TYPE_TINYINT, 1));
+    // dummy left expression to prevent segmentation fault
+    // since unary minus is the only arithmetic operator with one operand
+    e.push(new CV(EXPRESSION_TYPE_VALUE_CONSTANT, VALUE_TYPE_TINYINT, 1, (int64_t)1));
```
BillWhite: Maybe you should create enough infrastructure to not need the dummy left expression.  I'm not sure how this would be done, but it seems like it should be possible and people would be less likely to make a mistake.

Also, look at the function "bin".  I think there's an easier way to create these unit tests.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -95,6 +95,143 @@ public void testMod() throws Exception {
         verifyStmtFails(client, "select MOD(-25.32, ratio) from R1", "incompatible data type in operation");
     }
 
+    public void testUnaryMinus() throws Exception {
+        System.out.println("STARTING testUnaryMinus");
+        Client client = getClient();
+
```
BillWhite: Do we have a test for something like "-(-(a))"?  I don't see it here.  Also, "- -(a)".  The two minus signs together are a SQL comment, but putting a separator between them makes it two operators.

I suspect it will pass just fine, but making the test is simple and fast, and will lock down the behavior.

It's probably not worth it to optimize "-(-(a))" into just "a".

tests/frontend/org/voltdb/types/TestExpressionType.java
```diff
@@ -53,6 +53,8 @@ public void testGet() {
                      ExpressionType.get("sub"));
         assertEquals(ExpressionType.OPERATOR_MINUS,
                      ExpressionType.get("subtract"));
+        assertEquals(ExpressionType.OPERATOR_UNARY_MINUS,
+                ExpressionType.get("negate"));
```
BillWhite: I didn't know we had these tests.  This is pretty good.

src/frontend/org/voltdb/types/ExpressionType.java
```diff
@@ -68,6 +68,8 @@
     // unary null evaluation
     OPERATOR_EXISTS                (OperatorExpression.class, 18, "EXISTS", true),
     // unary exists evaluation
+    OPERATOR_UNARY_MINUS           (OperatorExpression.class, 22, "UNARY MINUS", true),
+    // unary exists evaluation
```
dm89: the operators have been grouped in different categories - but the numbers 19-21 also exist

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
```diff
@@ -95,6 +95,143 @@ public void testMod() throws Exception {
...
+        r.advanceRow();
+        // check if unary minus equals 0-value
+        // EDGE CASE -0 integer
+        assertEquals( 0, r.get("C1", VoltType.INTEGER));
+
+        // invalid data type for unary minus
+        verifyStmtFails(client, "select -desc from P1", "incompatible data type in operation");
+
+        // check -(-var) = var
+        cr = client.callProcedure("@AdHoc", "select num, -(-num) from P1");
```
dm89: here is the test for -(-a)

tests/frontend/org/voltdb/regressionsuites/TestFunctionsSuite.java
dm89: the tests for -(-a) are there (see comment below)
and - -a gives an error (by compiler) as its not a valid syntax

PR #4615, opened 2017-06-13
ENG-12663: Add authentication timeout to the Java client.

nshi:
Time out authentication on the client side if it takes longer than the
connection timeout setting in case the server dies or the connection is broken
during the authentication process. It is governed by the same connection timeout
setting specified in the ClientConfig.

If authentication times out, it throws an IOException.

### DIFF
--- a/src/frontend/org/voltdb/client/ConnectionUtil.java
+++ b/src/frontend/org/voltdb/client/ConnectionUtil.java
@@ -186,7 +186,7 @@ public class ConnectionUtil {
         if (addr.isUnresolved()) {
             throw new java.net.UnknownHostException(addr.getHostName());
         }
-        SocketChannel aChannel = SocketChannel.open(addr);
+        final SocketChannel aChannel = SocketChannel.open(addr);
         returnArray[0] = aChannel;
         assert(aChannel.isConnected());
         if (!aChannel.isConnected()) {
--- a/src/frontend/org/voltdb/client/ConnectionUtil.java
+++ b/src/frontend/org/voltdb/client/ConnectionUtil.java
@@ -21,6 +21,7 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/client/ConnectionUtil.java
```diff
@@ -94,6 +98,10 @@ private void shutdown() throws InterruptedException {
 
     private static final GSSManager m_gssManager = GSSManager.getInstance();
 
+    // Thread pool for checking authentication timeout
+    private static final ScheduledThreadPoolExecutor m_periodicWorkThread =
+        CoreUtils.getScheduledThreadPoolExecutor("Periodic Work", 0, CoreUtils.SMALL_STACK_SIZE);
```
akhanzode: Can you call this "Authentication Timer" or something unless we intend to use this for other tasks.

src/frontend/org/voltdb/client/ConnectionUtil.java
```diff
@@ -171,20 +180,38 @@ public DelegatePrincipal apply(Principal input) {
...
+        // Setup a timer that times out the authentication if it is stuck (server dies, connection drops, etc.)
+        final ScheduledFuture<?> timeoutFuture;
+        if (timeoutMillis > 0) {
+            timeoutFuture = m_periodicWorkThread.schedule(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        aChannel.close();
+                    } catch (IOException e) {
+                        // Ignore
```
akhanzode: For client this may be useful to log otherwise we wont know if client is timing out.

nshi: We don't log in the client library.

It throws an IOException if authentication timed out. User would know.

src/frontend/org/voltdb/client/ConnectionUtil.java
nshi: Will do.

PR #4616, opened 2017-06-14
ENG-12140 - Guard against site mastership kicking in while rolling.

akhanzode:
- Generations are rolling
- Next generation becomes active because of site mastership
- truncate comes in for active generation and gets scheduled after PBDs are opened.


akhanzode:
Review only: Need to decide wether to remove debug messages or leave as useful info.

PR #4618, opened 2017-06-14
ENG-12134-Removing-JSON-Warning

sycai:
Interface exception won't be thrown when sending HTTP requests to a rejoining node.

### DIFF
--- a/src/frontend/org/voltdb/HTTPClientInterface.java
+++ b/src/frontend/org/voltdb/HTTPClientInterface.java
@@ -80,7 +80,7 @@ public class HTTPClientInterface {
         @Override
         public InternalConnectionHandler get() {
             ClientInterface ci = VoltDB.instance().getClientInterface();
-            if (ci == null || !ci.isAcceptingConnections()) {
+            if (ci == null || (!VoltDB.instance().rejoining() && !ci.isAcceptingConnections())) {
                 throw new IllegalStateException("Client interface is not ready to be used or has been closed.");
             }
             return ci.getInternalConnectionHandler();
--- a/src/frontend/org/voltdb/HTTPClientInterface.java
+++ b/src/frontend/org/voltdb/HTTPClientInterface.java
@@ -80,7 +80,7 @@ public class HTTPClientInterface {
         @Override
         public InternalConnectionHandler get() {
             ClientInterface ci = VoltDB.instance().getClientInterface();
-            if (ci == null || (!VoltDB.instance().rejoining() && !ci.isAcceptingConnections())) {
+            if (ci == null || !ci.isAcceptingConnections()) {
                 throw new IllegalStateException("Client interface is not ready to be used or has been closed.");
             }
             return ci.getInternalConnectionHandler();
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -3966,17 +3966,16 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
 
             //Tell import processors that they can start ingesting data.
             ImportManager.instance().readyForData(m_catalogContext, m_messenger);
-        }
 
-        try {
-            if (m_adminListener != null && !m_rejoining) {
-                m_adminListener.start();
+            try {
+                if (m_adminListener != null) {
+                    m_adminListener.start();
+                }
+            }catch (Exception e) {
+                hostLog.l7dlog(Level.FATAL, LogKeys.host_VoltDB_ErrorStartHTTPListener.name(), e);
+                VoltDB.crashLocalVoltDB("HTTP service unable to bind to port.", true, e);
             }
-        } catch (Exception e) {
-            hostLog.l7dlog(Level.FATAL, LogKeys.host_VoltDB_ErrorStartHTTPListener.name(), e);
-            VoltDB.crashLocalVoltDB("HTTP service unable to bind to port.", true, e);
-        }
-        if (!m_rejoining && !m_joining) {
+
             Object args[] = { (m_mode == OperationMode.PAUSED) ? "PAUSED" : "NORMAL"};
             consoleLog.l7dlog( Level.INFO, LogKeys.host_VoltDB_ServerOpMode.name(), args, null);
             consoleLog.l7dlog( Level.INFO, LogKeys.host_VoltDB_ServerCompletedInitialization.name(), null, null);
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -3971,7 +3971,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
                 if (m_adminListener != null) {
                     m_adminListener.start();
                 }
-            }catch (Exception e) {
+            } catch (Exception e) {
                 hostLog.l7dlog(Level.FATAL, LogKeys.host_VoltDB_ErrorStartHTTPListener.name(), e);
                 VoltDB.crashLocalVoltDB("HTTP service unable to bind to port.", true, e);
             }
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -3969,7 +3969,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
         }
 
         try {
-            if (m_adminListener != null) {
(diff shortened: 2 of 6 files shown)


src/frontend/org/voltdb/HTTPClientInterface.java
```diff
@@ -80,7 +80,7 @@
         @Override
         public InternalConnectionHandler get() {
             ClientInterface ci = VoltDB.instance().getClientInterface();
-            if (ci == null || !ci.isAcceptingConnections()) {
+            if ((!VoltDB.instance().rejoining()) && ( ci == null || !ci.isAcceptingConnections())) {
                 throw new IllegalStateException("Client interface is not ready to be used or has been closed.");
             }
             return ci.getInternalConnectionHandler();
```
brians127: I don't think this is safe. If "ci" is null, you'll get a NullPointerException on line 86. Also, what is the behavior of getInternalConnectionHandler() when Volt is rejoining?

brians127: Consider catching the exception and ignoring it if Volt is rejoining. I'm not sure where that code belongs, you will have to look.

sycai: Thank you Brian, these are good suggestions

sycai: "Catching the exception and ignoring it if Volt is rejoining" seems to be logically the same as the "if" condition written here. How about "if (ci == null || (!VoltDB.instance().rejoining() && !ci.isAcceptingConnections())) ..."? By writing this way we can avoid the NullPointerException.

brians127: It depends on where you handle the error. If you catch the exception in this method, you are right. The problem is, what do you return? null? An empty object?

If you catch the exception where this method is being called, you don't have to worry about the return value. That would work well if this is used in one place or very few places.

sycai: I don't think we need to consider what is returned from "ci.getInternalConnectionHandler()". The original code only cares about whether "ci" itself is null or is able to accept connections. Should this "get()" function return null or an empty object, I assume it will be handled by the callers.

brians127: Please check who uses this method. The callers could NullPointerException if you return null and they assume you don't. I don't know this code well enough to say without looking.

sycai: Okay I've checked the two callers. They do not catch the NullPointerException. However, the connection handler object is initialized in all the constructors of the "clientInterface" class.

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -959,7 +959,7 @@ public void handle(String target, Request baseRequest,
                 // http://www.ietf.org/rfc/rfc4627.txt dictates this mime type
                 response.setContentType(JSON_CONTENT_TYPE);
                 if (m_jsonEnabled) {
-                    if (target.equals("/")) {
+                    if (target.equals("/") && !VoltDB.instance().rejoining()) {
```
brians127: Is "Resource not found" the correct response during a rejoin? If so I am OK with these changes so long as they work and pass tests.

nshi: I agree with @brians127 . If the node is still rejoining, it should either not accept any HTTP requests in the first place or return 503 Service Unavailable. As written, it will fall through to the else block and return Resource Not Found.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3958,7 +3958,7 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
         }
 
         try {
-            if (m_adminListener != null) {
+            if (m_adminListener != null && !m_rejoining) {
```
nshi: This approach looks good to me. I think you can move this whole try/catch block into the if block above, which already protects the rejoin case.

sycai: Sounds good. Also, the block below has the same if condition as the one above. Maybe we could merge all these blocks into one?

nshi: Excellent idea.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3955,17 +3955,16 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
...
-        }
 
-        try {
-            if (m_adminListener != null) {
-                m_adminListener.start();
+            try {
+                if (m_adminListener != null) {
+                    m_adminListener.start();
+                }
+            }catch (Exception e) {
```
brians127: Not a blocker: please fix the spacing; should be "} catch (".

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3955,17 +3955,16 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
...
             //Tell import processors that they can start ingesting data.
             ImportManager.instance().readyForData(m_catalogContext, m_messenger);
-        }
 
-        try {
-            if (m_adminListener != null) {
-                m_adminListener.start();
+            try {
+                if (m_adminListener != null) {
+                    m_adminListener.start();
```
brians127: Is the admin listener guaranteed to start at the end of a join or rejoin? If so this is fine.

sycai: Volt will call "onExecutionSiteRejoinCompletion" while joining or rejoining. That function will start the admin listener. So I think admin listener is guaranteed to start.

tests/frontend/org/voltdb/TestJSONInterface.java
```diff
@@ -978,6 +980,54 @@ public void testSimple() throws Exception {
...
+            cluster.recoverOne(0, 0, "", true);
+            Thread.sleep(1000);
+            url = "http://127.0.0.1:" + cluster.httpPort(2) + "/deployment";
+            try {
+                getUrlOverJSON(url, null, null, null, 200,  "application/json");
+            } catch (HttpHostConnectException e) {
+                fail("HTTP Listener is not open after the node has rejoined.");
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
brians127: This isn't correct. If fail() throws AssertionError, you'll get a stack trace but the test will pass. It's OK to declare the test method "throws Exception" - this will result in any uncaught exception propagating back to JUnit and failing the test.

tests/frontend/org/voltdb/TestJSONInterface.java
```diff
@@ -978,6 +980,52 @@ public void testSimple() throws Exception {
...
+            url = "http://127.0.0.1:" + cluster.httpPort(0) + "/deployment";
+            try {
+                getUrlOverJSON(url, null, null, null, 200,  "application/json");
+                fail("HTTP Listener is not closed when the node is down.");
+            } catch (HttpHostConnectException e) {
+                assertTrue(e.toString().contains("Connection refused"));
+            }
+
+            cluster.recoverOne(0, 0, "", true);
+            Thread.sleep(1000);
```
brians127: Somewhere in this window, the interesting state where one node is rejoining and you're trying to access it via HTTP occurs. I don't know if there's a way to consistently test this, but maybe you shouldn't wait? Perhaps retry getUrlOverJSON() until it succeeds or a timeout is hit?

tests/frontend/org/voltdb/TestJSONInterface.java
```diff
@@ -978,6 +980,52 @@ public void testSimple() throws Exception {
...
+
+            String url = "http://127.0.0.1:" + cluster.httpPort(0) + "/deployment";
+            try {
+                getUrlOverJSON(url, null, null, null, 200,  "application/json");
+            } catch (HttpHostConnectException e) {
+                fail("HTTP Listener is not open when the node is on");
+            }
+
+            cluster.killSingleHost(0);
+            Thread.sleep(1000);
```
brians127: if killSingleHost() waits until the node is dead, you won't need to sleep. Looking at a different test I believe this is the case.

tests/frontend/org/voltdb/TestJSONInterface.java
brians127: After discussion we decided this scenario is much better suited to system test.

brians127:
For testing, please bump the JSON warning to error level and run the system tests to see if it hits. It will be interesting to see if there are failures.

PR #4619, opened 2017-06-14
ENG-12683, fix some SQL junit tests failures introduced by ENG-11957.

luyangco:
I believe the expected StackOverflow will no longer happen because of ENG-11957, which same column-ref will be reused in the JSON plan. This patch will change these tests to be positive test.

But I don't know why TestMultipleOuterJoinPlans.testOuterSimplificationJoin() fails. This patch doesn't fix that.


xinjiacs:
I think TestMultipleOuterJoinPlans is just a plan change, it could be a valid or better change. Will ask SQL team to verify as well.

PR #4621, opened 2017-06-14
Fix up some junit test failures.

BillWhite:
There are two categories of errors.  The first is TestAdHocPlans,
TestAdHocQueries and TestProcCompiler.  In each of these cases we make
progressively larger SQL strings, with more complicated expressions
until we make one that's big enough that it breaks the parser.  We then
declare victory.  The second class is TestMultipleOuterJoinPlans.  It
seems that the hash collision problem fixed in voltdb commit
4a7020630c047eb5a16c1005f61a99e6aab83ba7 caused an expression to be
hashed differently, which caused an expression to be corrected.

https://issues.voltdb.com/browse/ENG-12683

### DIFF
--- a/tests/frontend/org/voltdb/compiler/TestProcCompiler.java
+++ b/tests/frontend/org/voltdb/compiler/TestProcCompiler.java
@@ -98,23 +98,27 @@ public class TestProcCompiler extends TestCase {
     }
 
     public void testStmtForStackOverflowCondition() throws Exception {
-        String schema =  "Create Table foo ( " +
-                         "id BIGINT DEFAULT 0 NOT NULL, " +
-                         "name VARCHAR(255) NOT NULL, " +
-                         "PRIMARY KEY(id));";
+        boolean success = true;
+        ByteArrayOutputStream capturer = null;
+        for (int npreds = 2000; success && npreds < 100000; npreds += 1000) {
+            String schema =  "Create Table foo ( " +
+                             "id BIGINT DEFAULT 0 NOT NULL, " +
+                             "name VARCHAR(255) NOT NULL, " +
+                             "PRIMARY KEY(id));";
 
-        VoltProjectBuilder builder = new VoltProjectBuilder();
-        ByteArrayOutputStream capturer = new ByteArrayOutputStream();
-        PrintStream capturing = new PrintStream(capturer);
-        builder.setCompilerDebugPrintStream(capturing);
-        builder.addLiteralSchema(schema);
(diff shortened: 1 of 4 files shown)


tests/frontend/org/voltdb/planner/TestMultipleOuterJoinPlans.java
```diff
@@ -263,7 +263,7 @@ public void testOuterSimplificationJoin() {
         n = n.getChild(1);
         // HSQL doubles the join expression for the first join. Once it's corrected the join expression type
         // should be ExpressionType.COMPARE_EQUAL
-        verifyJoinNode(n, PlanNodeType.NESTLOOP, JoinType.INNER, null, ExpressionType.CONJUNCTION_AND, null, PlanNodeType.SEQSCAN, PlanNodeType.SEQSCAN);
+        verifyJoinNode(n, PlanNodeType.NESTLOOP, JoinType.INNER, null, ExpressionType.COMPARE_EQUAL, null, PlanNodeType.SEQSCAN, PlanNodeType.SEQSCAN);
```
wolffcm: I'd still like to know how the hashCode/equals change fixed this, but I think you're explanation of why the new plan is correct (and more performant) is spot-on.

tests/frontend/org/voltdb/compiler/TestProcCompiler.java
```diff
@@ -98,23 +98,27 @@ public void testStmtWithLotsOfPredicates() throws Exception {
...
-        builder.addLiteralSchema(schema);
-
-        // Test that a stored procedure with more than the max allowable predicates
-        // results in an error and does not crash or hang the system.
-        String sql = getQueryForFoo(2000);
-        builder.addStmtProcedure("StmtForStackOverFlow", sql, null);
+        boolean success = true;
+        ByteArrayOutputStream capturer = null;
+        for (int npreds = 2000; success && npreds < 100000; npreds += 1000) {
+            String schema =  "Create Table foo ( " +
```
BillWhite: I don't know.  It doesn't seem to be onerous, and it is a test.

tests/frontend/org/voltdb/planner/TestMultipleOuterJoinPlans.java
BillWhite: I'll change the comment, but I don't want to create a new branch and do another Jenkins run just for the comment.  I'll merge the comment change in with inline insert.

PR #4623, opened 2017-06-15
sqlcmdtest: fixed the stackoverflow, exec, and versioncheck tests, in 2

johnTcrawford:
ways: first, quadrupled the number of AND predicates, to make a stack
overflow much more likely to occur, as these tests expect; second, added
a "drop procedure max_plus_predicates if exists;" SQL statement, near
the end (of the latter two tests), so that even if these tests fail to
work (i.e., fail to generate a stack overflow), this failure will not
cause other, later tests (which also use table T) to fail. Each test's
.in, .outbaseline, and (for the latter two tests) .errbaseline files
were modified. (Also tweaked one comment.)


johnTcrawford:
Thanks, Bill. I did not realize that you had already reviewed this, before our meeting.

PR #4625, opened 2017-06-15
ENG-12682: catch and log NoSuchElementException if stats change collide with stats query

JadenW:


JadenW:
Improved by ENG-13325 (https://github.com/VoltDB/voltdb/pull/4983)

PR #4629, opened 2017-06-16
Pass Kafka SSL client properties from command line.

jeffreycrump:


src/frontend/org/voltdb/utils/KafkaLoader10.java
```diff
@@ -286,6 +289,12 @@ private Properties kafkaConfigProperties() throws IOException {
             }
         }
 
+        if (!m_cliOptions.kafkassl.trim().isEmpty()) {
+            try (FileInputStream fis = new FileInputStream(new File(m_cliOptions.kafkassl))) {
+                props.load(fis);
```
brians127: Is there any risk that Kafka SSL properties will collide with any other properties used by Volt? Also has this been UX reviewed?

jeffreycrump: It's a separate properties file and we're loading the Kafak properties object here. So no chance of collision. Andrew is a reviewer.

jeffreycrump:
After talking with Huzefa, we think it's best to just include the sample properties in the existing config file (the one specified by --config). We do some filtering on the properties and we don't want to risk customers putting non-SSL related properties in the SSL config file and having them undo our filtering.

So no code changes will be required to support SSL after all. We just need to test it.

brians127:
There are other UX options here. We could throw an error if SSL.X is in the non-SSL file or non-SSL options are in the SSL file. Also we could optionally take multiple properties files using the existing parameter. Have you discussed with Andrew?

PR #4631, opened 2017-06-16
ENG-11657 do not resume replication for DR consumer if durability is off

dhan111:
If you use the new CLI without command logging, the database will pick the recover start action internally when you "start" after the first time even though there is nothing to recover from. This tricks the DR system into thinking that it can resume replication. It eventually fails of course, but it may confuse the remote cluster and generate confusing error messages.

Now before resuming replication, we check if durability is turned on and if there is a completed snapshot. If durability is off or no snapshot is found, we start fresh.

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -353,6 +353,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
      */
     String m_terminusNonce = null;
 
+//  m_durable means commandlogging is enabled.
     boolean m_durable = false;
 
     private int m_maxThreadsCount;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1039,14 +1041,15 @@ public void initialize(Configuration config) {
             validateStartAction();
 
             // durable means commandlogging is enabled.
-            boolean durable = readDeploymentAndCreateStarterCatalogContext(config);
+            m_durable = readDeploymentAndCreateStarterCatalogContext(config);
+
```
JadenW: So m_durable is only set in the initialize time and don't need to be updated when commandlogging setting changes (i.e.  change commandlogging from disable to enable via catalog update)?  
I think this is fine as m_durable is used for determining resumeReplication of consumer gateway (which happens before any catalog update).

JadenW: nitpick: change durable to m_durable in the comment and maybe move this comment to its definition.

manjujames: I don't think we allow enabling command logging after server is up.

manjujames:
It looks good to me. We will look at jenkins today before merging.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1571,6 +1574,18 @@ public Boolean get() {
     }
 
     /**
+     * Check if replication needs to be resumed
+     * Return true replication needs to be resumed,
+     * false if we need to starting new or command log was disabled or
+     * there was no complete snapshot
+     */
+   private boolean willResumeReplication()
```
slowsoul: Nitpicking: renaming this method to willDoActualRecover() may be more accurate.

PR #4636, opened 2017-06-19
Eng 12642 count index botch backportv6.6

BillWhite:
The IndexCountExecutor did not do the right thing when a filter tested
for a string which was too long for the column it was testing. For
example, "T.COL < 'somelongstring'" when T.COL has type varchar(10), for
example.

As it turns out, there are two problems. One is that the
indexcountexecutor did not have the right logic for this case. The
other is that a constructor of SQLException was called that did not
propagate the internal flags which noted that the exception was for
overflow, underflow or type size mismatch. Consequently we didn't know
why the error happened, and we just threw the error exception without
trying to patch things up.

I'm a little astonished that this does not happen for indexscanexecutor.
Perhaps it does.

https://issues.voltdb.com/browse/ENG-12642


build.py
```diff
@@ -53,7 +53,7 @@
     CTX.CPPFLAGS += " -pthread"
     CTX.LDFLAGS += " -rdynamic"
     if (CTX.compilerMajorVersion() >= 4):
-        CTX.CPPFLAGS += " -Wno-deprecated-declarations  -Wno-unknown-pragmas"
+        CTX.CPPFLAGS += " -Wno-deprecated-declarations  -Wno-unknown-pragmas -Wno-unused-local-typedefs -Wno-ignored-qualifiers -Wno-strict-aliasing"
```
yzhang1991: Why is this necessary? It seems we are suppressing some warnings here?

src/ee/common/tabletuple.h
```diff
@@ -588,8 +588,17 @@ class TableTuple {
         bool isInBytes = columnInfo->inBytes;
         char *dataPtr = getWritableDataPtr(columnInfo);
         int32_t columnLength = columnInfo->length;
-        value.serializeToTupleStorage(dataPtr, isInlined, columnLength, isInBytes,
-                                      allocateObjects, tempPool);
+
+        try {
```
yzhang1991: This debug print is from a Xin's debug build branch? Do we want to get it in? I do not think so.

tests/sqlcmd/build_jar.sh
```diff
@@ -4,9 +4,24 @@
 # (which is used to test the LOAD CLASSES and REMOVE CLASSES commands).
 # Also adds other files to the jar file that are used to test handling of
 # class loader errors.
-
+BUILD=release
```
yzhang1991: Are those changes for this ticket? Or from some places else?

BillWhite: I needed this to be able to fix sqlcmdtest on my debug build.
This is also on master.

src/ee/common/tabletuple.h
BillWhite: I didn't take anything from Xin's branch.  This was from master.  I thought there was a test looking for it.

build.py
BillWhite: We do this for other gcc versions.  This is only to build on u16.10, which is all I have.

PR #4639, opened 2017-06-20
ENG- 12281 fix a test case

zeemanhe:
simplified the logic to check missing partitions.


src/frontend/org/voltdb/iv2/LeaderAppointer.java
```diff
@@ -604,14 +603,16 @@ public boolean isClusterKSafe(Set<Integer> failedHosts)
                     isInitialized = partitionState[0] == LeaderElector.INITIALIZED;
                 }
 
+                List<String> replicas = childrenCallbacks.poll().getChildren();
+                if (pid == MpInitiator.MP_INIT_PID) continue;
```
xinjiacs: Why we need to pull the replicas anyway when pid is MPI? @zeemanhe

PR #4641, opened 2017-06-20
ENG-12605: add optional elasticallyModified field in query response

JadenW:
pro side: https://github.com/VoltDB/pro/pull/1936
DR elastic-add support for Jun
1. elastically add node(s) on producer cluster before enable DR producer ( i.e. listen == false);
2. elastically add node(s) on consumer cluster before enable DR consumer ( i.e connection source == "" or enabled == false);

changes including:
1. add one optional elasticallyModified field in query response (protoBuffer change), which use Hashinator's helper function to find out local has ever done elastic add;
2. producer will update cached local Partition Count when answering query before snapshot; (assume elastic add happens before DR producer enable, so it will also before snapshot);
3. consumer will update local partitionKeys each time when catalog updates (assume there will a @uac after elastic add for enabling DR consumer);
4. in consumer dispatcher attachToMaster, cache both local and producer elasticallyModified state for deciding UpstreamHeterogeneous;
5. lift a restriction in LeaderAppointer for allowing assignLeader for newly added partition before snapshotComplete.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -4072,7 +4074,7 @@ private void prepareReplication() {
                     m_consumerDRGateway.setInitialConversationMembership(expectedClusterMembers.getFirst(),
                             expectedClusterMembers.getSecond());
                 }
-                m_consumerDRGateway.initialize(m_config.m_startAction != StartAction.CREATE);
+                m_consumerDRGateway.initialize(m_config.m_startAction != StartAction.CREATE && m_config.m_startAction != StartAction.JOIN);
```
slowsoul: This means a joining node will always initialize its consumer side with resumeReplication set to false, which is not correct if the original cluster was started as "recover". I think join should be handled the same as rejoin, which means resumeReplication should be true, thus this change is not needed. We should also rebase to master after this is merged to take in Dong's change.

src/frontend/org/voltdb/iv2/LeaderAppointer.java
```diff
@@ -208,11 +208,13 @@ public void run(List<String> children)
...
                 // If we are still a DR replica (not promoted) and starting from a snapshot,
-                // check if that has completed
+                // check if that has completed if this is not a new replica
                 if (VoltDB.instance().getReplicationRole() == ReplicationRole.REPLICA &&
-                    m_expectingDrSnapshot && m_snapshotSyncComplete.get() == false) {
+                    m_expectingDrSnapshot && m_snapshotSyncComplete.get() == false
+                        && !(m_currentLeader == Long.MAX_VALUE && !updatedHSIds.isEmpty()))  {
                     VoltDB.crashGlobalVoltDB("Detected node failure before DR sync snapshot completes. Cluster will shut down.",
                                              false, null);
+
```
slowsoul: Extra line.

src/frontend/org/voltdb/iv2/LeaderAppointer.java
```diff
@@ -208,11 +208,13 @@ public void run(List<String> children)
                                              false, null);
                 }
                 // If we are still a DR replica (not promoted) and starting from a snapshot,
-                // check if that has completed
+                // check if that has completed if this is not a new replica
                 if (VoltDB.instance().getReplicationRole() == ReplicationRole.REPLICA &&
-                    m_expectingDrSnapshot && m_snapshotSyncComplete.get() == false) {
+                    m_expectingDrSnapshot && m_snapshotSyncComplete.get() == false
+                        && !(m_currentLeader == Long.MAX_VALUE && !updatedHSIds.isEmpty()))  {
```
slowsoul: Could you explain a little about this change? Not so familiar with currentLeader and updatedHSIds. Might also be helpful to include them in the comment

JadenW: This if for changes #5 lift a restriction in LeaderAppointer.
It seems previously leaderAppointer won't allow a leader assignment if DR role is Replica and hasn't finished snapshot. As for new replica introduced by join, this should be excluded.

PR #4642, opened 2017-06-20
ENG-12663: Add authentication timeout to the Java client.

nshi:
Time out authentication on the client side if it takes longer than the
connection timeout setting in case the server dies or the connection is broken
during the authentication process. It is governed by the same connection timeout
setting specified in the ClientConfig.

If authentication times out, it throws an IOException.


nshi:
r=akhanzode

PR #4643, opened 2017-06-20
ENG-12677: Fix race condition between BulkLoader accounting and callback

nshi:
BulkLoader keeps track of the outstanding row count and uses it in the drain()
method to make sure everything is processed. Users can also use the counter
directly to track how many rows are in-flight.

If a row fails to be inserted, it will be retried automatically. If the retry
fails, too, a user-specified callback will be invoked to notify the user of the
failed row.

However, the callback is invoked after the outstanding row count has already
been decremented. In other words, a user watching the outstanding row count
could see it go to zero, meaning all rows are processed, but not see the failure
callback to be invoked even if there are failed rows. When the callback will be
invoked depends on thread scheduling, which is nondeterministic.

The outstanding row count is now updated AFTER the failure callback has been
invoked.


nshi:
r=akhanzode

PR #4645, opened 2017-06-20
Fix race condition in config/test/cluster startup for logfile Collector

jeffreycrump:
In the test environment, the configuration files are created in a separate process launched by the database creation logic.  Every now and then, this would not finish before the Collector looked for it.  So, this fix will retry a few times, waiting for the files to appear, before giving up.

I'm not entirely sure this is the best place for this logic -- it feels like it should belong in the TestCollector code -- but the race comes in to play in Collector.populateVoltDBCollectionPaths, which is product code and that expects these files to exist.

I'm open to suggestions for a better place for this code.

The tests have been running for a few days, constantly, with this problem not reappearing. Jenkins job is here: http://ci.voltdb.lan/view/Branch-jobs/view/branch-ENG-12684-collector/job/branch-2-community-junit-ENG-12684-collector


hhakimi53:
Collector utility will be run in field with all files in place. Retrying for the config/deployment for voltdb will not help as it will be either pass or fail either ways. Alternate approach will be to update the test logic for old cli environment to make sure LocalCluster is fully initialized with all needed voltdb files in place or by adding delay for old cli environment after initializing cluster.

jeffreycrump:
@hhakimi53 Thanks for the context on Collector. I've moved that logic into TestCollector.startup: https://github.com/VoltDB/voltdb/commit/328552f54cb95485eb0098fe716ada4e22d75cea

jeffreycrump:
Sure, no problem.

src/frontend/org/voltdb/utils/Collector.java
```diff
@@ -144,10 +144,10 @@ private static void populateVoltDBCollectionPaths() {
         // Validate voltdbroot path is valid or not - check if deployment and config info json exists
         File deploymentFile = new File(m_deploymentPath);
         File configInfoFile = new File(m_configInfoPath);
+
         if (!deploymentFile.exists() || !configInfoFile.exists()) {
             System.err.println("ERROR: Invalid database directory " + m_voltdbRoot.getParentFile().getAbsolutePath()
-                    + ". Specify valid database directory using --dir option.");
-            VoltDB.exit(-1);
+                               + ". Specify valid database directory using --dir option.");
```
akhanzode: Why these changes? I also see VoltDB.exit removed accidental change?

hhakimi53: Exit is needed in there. I missed that one during review!

PR #4646, opened 2017-06-20
Eng-11955 Enable same topics for different Kafka importer configurations

sycai:


src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1557,17 +1559,54 @@ private static void setImportInfo(Catalog catalog, ImportType importType) {
...
+                    groupid = pt.getValue();
+                }
+            }
+            if (groupidToTopics.containsKey(groupid)) {
+                HashSet<String> union = new HashSet<>(groupidToTopics.get(groupid));
+                union.addAll(topics);
+                if (union.size() == topics.size() + groupidToTopics.get(groupid).size()) {
+                    groupidToTopics.put(groupid, union);
+                } else {
+                    VoltDB.crashLocalVoltDB("Two kafka configurations cannot have the same topic and groupid");
```
akhanzode: Dont crash... Users can do update deployment with bad groupid configuration and that should not crash.

sycai: How about logging this issue with a warning message instead of crushing the Volt?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1557,17 +1560,54 @@ private static void setImportInfo(Catalog catalog, ImportType importType) {
...
+                    groupid = pt.getValue();
+                }
+            }
+            if (groupidToTopics.containsKey(groupid)) {
+                HashSet<String> union = new HashSet<>(groupidToTopics.get(groupid));
+                union.addAll(topics);
+                if (union.size() == topics.size() + groupidToTopics.get(groupid).size()) {
+                    groupidToTopics.put(groupid, union);
+                } else {
+                    throw new RuntimeException("Import failed to configure, two Kafka configurations have the same groupid and topic.");
```
sycai: @ajgent Hi Andrew, could you help me improve this error message? Users are not supposed to give multiple kafka configurations with the same topic+group id. So if Volt finds that two configurations share a same topic while having the same group id, it will throw out this exception.

nshi:
I would suggest including the description of the issue and the solution in the commit message, especially when the ticket doesn't contain much information. This helps any reviewers understand the problem and your approach to solve it. It also associates enough context with the code changes. Brian's comment on the ticket is a good starting point.

sycai:
Each Kafka importer has a unique URI based on the partition number and the topic. Therefore if there are two importer configurations having the same topic, the URI is not guaranteed to be unique. The solution is to let user assign unique group ids to configurations, and let Volt generate URIs based on partition number, topic as well as the group id.

A validation method is added in the CatalogUtil.java file. It checks whether two Kafka importer configurations share the same topic and group id. 

If user gives invalid Kafka configurations, Volt will not start or update.

hhakimi53:
What happens when the procedure name are the same for different importer configs with same topic and different group-id?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1570,17 +1573,54 @@ private static void setImportInfo(Catalog catalog, ImportType importType) {
...
+                if (pt.getName().equals("topics")) {
+                    topics.addAll(Arrays.asList(pt.getValue().split("\\s*,\\s*")));
+                } else if (pt.getName().equals("groupid")) {
+                    groupid = pt.getValue();
+                }
+            }
+            if (groupidToTopics.containsKey(groupid)) {
+                HashSet<String> union = new HashSet<>(groupidToTopics.get(groupid));
+                union.addAll(topics);
+                if (union.size() == topics.size() + groupidToTopics.get(groupid).size()) {
```
akhanzode: Put parenthesis around add part the == looks confusing otherwise.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1570,17 +1573,54 @@ private static void setImportInfo(Catalog catalog, ImportType importType) {
...
             if (!streamList.contains(importConfiguration.getModule())) {
                 streamList.add(importConfiguration.getModule());
             }
 
             checkImportProcessorConfiguration(importConfiguration);
         }
+        validateKafkaConfig(kafkaConfigs);
+    }
+
+    private static void validateKafkaConfig(List<ImportConfigurationType> configs) {
```
akhanzode: put a detailed comment on this method what it is trying to do.

akhanzode:
If jenkins is blue please squash merge.

PR #4652, opened 2017-06-22
Eng 12298

GuaGuaDean:
I re-order the upgrade plan steps so the load schema step can be replaced with init using schema adn procs.


GuaGuaDean:
Project branch-system-test-online-upgrade test is suggested pass on Jenkins

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -245,13 +245,21 @@ def generateDeploymentFile(runner, hosts, surviveSet, killSet, clusterIds, post_
                 writeCommands(file,
                               'Step %d: copy deployment file' % step,
                               '%s#instruction# copy %s to %s' % (warningForDeploy, new_cluster_deploy, runner.opts.newRoot))
+
+    # originally it is the 5th step - get all the ddl and scp to all machines
+    file.write('#instruction# get ddl file by using: voltdb get --output={path to dir} schema')
```
luyangco: I think you misunderstood the ticket description here, the idea of this script is to generate commands that can be copy&paste to a console as much as possible.  Here are a few things you need to improve:
1. `voltdb get ` command needs a `--dir` option to specify where the voltdbroot is, you can't assume current directory is always voltdbroot.  `--output` can be anywhere, but it's better to give a specific path (maybe voltdbroot or the current directory) instead of asking user to fill it.
2. You miss another command： `voltdb get classes`, similarly it also needs a `--dir` option to specify voltdbroot.
3. A minor issue. Please follow the same pattern to create a new step, instead putting them between step 1 and step 2.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -174,26 +174,26 @@ def generateCommands(runner, hosts, kfactor, clusterIds):
     leadersString = generateStartNewClusterCommand(runner.opts, killSet, hostcount, files, newNodeF, step)
 
     # 5 load schema into the new cluster
-    step += 1
-    writeCommands(files[getKey(killSet[0])], 'Step %d: load schema' % step, '#instruction# load schema into the new-version cluster')
+    #step += 1
```
luyangco: If you no longer need this piece of code, delete it please.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -268,25 +276,32 @@ def generateInitNewClusterCommand(opts, killSet, files, new_cluster_deploy, newN
...
+
     if opts.newNode is not None:
         writeCommands(newNodeF,
                       'Step %d: initialize new cluster' % step,
-                      '%s init --dir=%s --config=%s --force' % (os.path.join(opts.newKit, 'bin/voltdb'),
+                      '%s init --dir=%s --config=%s --schema=%s --classes%s --force' % (os.path.join(opts.newKit, 'bin/voltdb'),
                                                                 opts.newRoot,
-                                                                os.path.join(opts.newRoot, new_cluster_deploy)))
+                                                                os.path.join(opts.newRoot, new_cluster_deploy),
+                                                                os.path.join(opts.newRoot, 'ddl.sql'),
```
luyangco: Please use the default name of those artifacts. I believe one is `schema.sql` and another is `procedures.jar`, correct me if I'm wrong.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -300,13 +315,13 @@ def generateStartNewClusterCommand(opts, killSet, hostcount, files, newNodeF, st
     for hostInfo in killSet:
         writeCommands(files[getKey(hostInfo)],
                       'Step %d: start new cluster' % step,
-                      "%s start --dir=%s -H %s -c %d --missing=%d" % (os.path.join(opts.newKit, 'bin/voltdb'),
+                      "%s start --dir=%s -H %s -c %d --missing=%d\nPlease use proper port to avoid error"
```
luyangco: Leader port is generated from `@SystemInformation`, so the extra warning is unnecessary.

GuaGuaDean: I am talking about the replication port here. I think users need to specify that in order to avoid the confusion of IP address interfaces. I checked the SystemInformation file which don't have any replication port attributes.

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -245,13 +245,21 @@ def generateDeploymentFile(runner, hosts, surviveSet, killSet, clusterIds, post_
                 writeCommands(file,
                               'Step %d: copy deployment file' % step,
                               '%s#instruction# copy %s to %s' % (warningForDeploy, new_cluster_deploy, runner.opts.newRoot))
+
+    # originally it is the 5th step - get all the ddl and scp to all machines
+    file.write('#instruction# get ddl file by using: voltdb get --output={path to dir} schema')
+    file.write('\n')
+    file.write('#instruction# scp the ddl to other machines')
```
GuaGuaDean: Because users need to init the voltroot directory at each node in new cluster. If we only init with schema once, will that cause any collision?

lib/python/voltcli/voltadmin.d/plan_upgrade.py
```diff
@@ -245,13 +241,22 @@ def generateDeploymentFile(runner, hosts, surviveSet, killSet, clusterIds, post_
                 writeCommands(file,
                               'Step %d: copy deployment file' % step,
                               '%s#instruction# copy %s to %s' % (warningForDeploy, new_cluster_deploy, runner.opts.newRoot))
+        # originally it is the 5th step - get all the ddl and scp to all machines
```
luyangco: Now it looks much better, good job. 

But I still see some places to improve. Here are my suggestions, 
a) wrap the following lines into a function, and
b) move the new function to be the new step 2, use the `writeCommands()` API, all steps after it move off by one, and
c) since `voltdb get` command is executable, remove the `#instruction` tag.

GuaGuaDean:
Should be fine now.

PR #4654, opened 2017-06-22
ENG-11529, ENG-12751, ENG-12752: catalog jar async write and other work moved to asynchronous thread

vittvolt:
## Overview
Now writing an updated catalog jar is moved to a new NT system proc (@WriteCatalog).

- @UpdateClasses, @UAC and @AdHoc now first call the new NT proc on all hosts, then call @UpdateCore to rename the jars synchronously after all asynchronous writing has successfully completed.
- A new test method has been added in pro to test loaded procedures after recover.
- The lock (including the zookeeper node) has been moved to the asynchronous thread
- The catalog classes verification has also been moved to NT proc.

## Pro
- Check https://github.com/VoltDB/pro/pull/1941 for the updated test on the pro side.


src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -312,11 +312,57 @@ public CatalogContext update(
...
+            throw new RuntimeException("Error with current catalog jar file status, \"catalog.jar\" existence"
+                    + ": " + catalog_file.exists() + " catalog.jar.tmp existence: " + catalog_tmp_file.exists() +
+                    "\nPlease make such changes synchronously from a single " + "connection to the cluster.");
+        }
+    }
+
+    // This function is supposed to be in the @WriteCatalog system call.
+    // Therefore we expect to see an existing catalog jar, and no other
+    // temporary jar file exists
+    public void writeCatalogJarToTempFile(String path, String name) throws IOException
```
xinjiacs: I am thinking to move these two catalog jar write functions into a public static function in "InMemoryJarfile" out of the CatalogContext. 

The benefits:
1. Static function by pass in the InMemoryJarfile from the catalog context.
2. Separation of jar writes inside of catalog context.
3. Closer with InMemoryJarfile.writeToFile.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3311,6 +3327,57 @@ else if (currentTxnId < m_lastLogUpdateTxnId) {
...
+                throw new RuntimeException("Trying to update main catalog context with diff " +
+                        "commands generated for an out-of date catalog. Expected catalog version: " +
+                        expectedCatalogVersion + " does not match actual version: " + m_catalogContext.catalogVersion);
+            }
+
+            // A temporary context to be created for writing the updated jar
+            // The context will be updated later in @UpdateCore call
+            // Note that the txn id and unique id should not matter for writing
+            // the catalog jar to disk
+            CatalogContext temp_catalogContext =
```
xinjiacs: After the refactoring work that I mentioned about the catalog jar writes, you do not need to construct a new catalog context for jar writes and pass in most of these parameters.

src/frontend/org/voltdb/VoltDBInterface.java
```diff
@@ -167,6 +167,36 @@
...
+     * @param expectedCatalogVersion
+     * @param currentTxnId
+     * @param currentTxnTimestamp
+     * @param deploymentBytes
+     * @param deploymentHash
+     * @param requireCatalogDiffCmdsApplyToEE
+     * @param hasSchemaChange
+     * @param requiresNewExportGeneration
+     */
+    default public void writeCatalogJar(
```
xinjiacs: rename: writeCatalogTemporaryJar

Also update the comments.

Once after the refactoring, you do not need to pass in these unused parameters.

src/frontend/org/voltdb/VoltNonTransactionalProcedure.java
```diff
@@ -32,6 +32,11 @@
 
     ProcedureRunnerNT m_runner = null;
 
+    // Is this a good replacement for the transactional id?
```
xinjiacs: not useful or necessary, removed?

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -174,20 +178,56 @@ else if (ddlToken.equals("create")) {
             return makeQuickResponse(ClientResponseImpl.SUCCESS, "Catalog update with no changes was skipped.");
         }
 
-        // initiate the transaction.
+        // Write the new catalog to a temporary jar file
+        CompletableFuture<Map<Integer,ClientResponse>> cf =
```
xinjiacs: Can you refactor these catalog jar disk writes into a static function into UpdateApplicationBase? It has been repeated a couple times.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -174,20 +178,56 @@ else if (ddlToken.equals("create")) {
...
-                             ccr.deploymentString,
-                             ccr.tablesThatMustBeEmpty,
-                             ccr.reasonsForEmptyTables,
-                             ccr.requiresSnapshotIsolation ? 1 : 0,
-                             ccr.worksWithElastic ? 1 : 0,
-                             ccr.deploymentHash,
-                             ccr.requireCatalogDiffCmdsApplyToEE ? 1 : 0,
-                             ccr.hasSchemaChange ?  1 : 0,
-                             ccr.requiresNewExportGeneration ? 1 : 0);
+                              ccr.encodedDiffCommands,
```
xinjiacs: reason?

src/frontend/org/voltdb/sysprocs/UpdateApplicationCatalog.java
```diff
@@ -98,20 +102,56 @@
             noteRestoreCompleted();
         }
 
-        // initiate the transaction.
+        // Write the new catalog to a temporary jar file
+        CompletableFuture<Map<Integer,ClientResponse>> cf =
```
xinjiacs: Same refactoring here.

src/frontend/org/voltdb/sysprocs/UpdateCore.java
```diff
@@ -276,56 +271,56 @@ public DependencyPair executePlanFragment(
...
-                        log.warn(msg);
-                        throw new VoltAbortException(e);
-                    }
-                }
-            } catch (Exception e) {
-                Throwables.propagate(e);
-            }
-
+//
+//            // We know the ZK bytes are okay because the run() method wrote them before sending
```
xinjiacs: Remove the precheck logic.

src/frontend/org/voltdb/sysprocs/WriteCatalog.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+import org.voltcore.logging.VoltLogger;
+import org.voltdb.ClientResponseImpl;
+import org.voltdb.VoltDB;
+import org.voltdb.client.ClientResponse;
+
+public class WriteCatalog extends UpdateApplicationBase {
```
xinjiacs: I would rename it with "UpdateCatalogHelper" and add comments.

src/frontend/org/voltdb/sysprocs/WriteCatalog.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // TODO: maybe we can add another option to clean up the
+    // temporary jar files on all hosts, in case some operations
+    // failed in the middle of the way ?
+
+    public static final byte WRITE = 0;
+    public static final byte CLEAN_UP = 1;
+    public static final byte VERIFY = 2;
```
xinjiacs: I will not add a separate "VERIFY" here and merge it with "WRITE" by "CHECK_AND_WRITE".

The benefit is that you do not need to invoke run everywhere NT procedure twice increasing the total run time of UpdateClasses, and these two job share the same input catalog bytes.

src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -304,19 +304,38 @@ public CatalogContext update(
...
+            return m_jarfile.writeToFile(catalog_file);
+        } else if (catalog_file.exists() && !catalog_tmp_file.exists()) {
+            // This may happen during cluster recover step, in this case
+            // we must overwrite the file (the file may have been changed)
+            catalog_file.delete();
+            return m_jarfile.writeToFile(catalog_file);
+        } else {
+            throw new RuntimeException("Error with current catalog jar file status, \"catalog.jar\" existence"
+                    + ": " + catalog_file.exists() + " catalog.jar.tmp existence: " + catalog_tmp_file.exists() +
+                    "\nPlease make such changes synchronously from a single " + "connection to the cluster.");
```
xinjiacs: When can we hit this case? How should us survive from it?

vittvolt:
Currently I have another branch **ENG-11529-catalog-jar-async-write-merging** which merges with the latest master and added some miscellaneous changes. I will merge in those changes after the unit tests become stable. @xinjiacs

src/frontend/org/voltdb/CatalogContext.java
vittvolt: @xinjiacs I don't think there is any way to survive from it if any exception is thrown. The current implementation just ignores any IOException and continue working. 

I think the best defensive way is to shutdown the local instance upon any IO failure , in order to avoid any future pain.

src/frontend/org/voltdb/VoltDBInterface.java
```diff
@@ -167,6 +167,24 @@
...
 
     /**
+     * Given the information, write the new catalog jar file only
+     */
+    default public void writeCatalogJar(byte[] newCatalogBytes) throws IOException
+    {
+        return;
+    }
+
+    default public String verifyZKCatalog()
```
vittvolt: @xinjiacs it was intended to run on every host: pull from the ZooKeeper, and check its validity. Or is checking the bytes without involving ZooKeeper enough ?

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -127,59 +129,41 @@ else if (ddlToken.equals("create")) {
...
-                             null,
-                             sqlStatements.toArray(new String[0]),
-                             null,
-                             false,
-                             true,
-                             false,
-                             getHostname(),
-                             getUsername(),
-                             false);
+        return updateApplication(ccr);
```
vittvolt: Cleanup work to delete any remaining temporary jar files is still needed. I remember in cases like 

"Alter table" to add a constraint on a non-empy table, the AdHoc query will fail during performVerification() check. And the UpdateCore will return failure.

In this case the temporary jar file is already written but never deleted, and later UAC calls will crash the localdatabases.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -131,64 +129,41 @@ else if (ddlToken.equals("create")) {
...
-                             ccr.tablesThatMustBeEmpty,
-                             ccr.reasonsForEmptyTables,
-                             ccr.requiresSnapshotIsolation ? 1 : 0,
-                             ccr.worksWithElastic ? 1 : 0,
-                             ccr.deploymentHash,
-                             ccr.requireCatalogDiffCmdsApplyToEE ? 1 : 0,
-                             ccr.hasSchemaChange ?  1 : 0,
-                             ccr.requiresNewExportGeneration ? 1 : 0,
-                             ccr.m_ccrTime);
+        return updateApplication(ccr);
```
vittvolt: Cleanup work to delete any remaining temporary jar files is still needed. I remember in cases like

"Alter table" to add a constraint on a non-empy table, the AdHoc query will fail during performVerification() check. And the UpdateCore will return failure.

In this case the temporary jar file is already written but never deleted, and later UAC calls will crash the local databases.

src/frontend/org/voltdb/sysprocs/UpdateCore.java
```diff
@@ -518,63 +450,37 @@ private final void performCatalogVerifyWork(
...
-                    log.info("To: catalog hash: " + Encoder.hexEncode(catalogHash).substring(0, 10) +
-                            ", deployment hash: " + Encoder.hexEncode(deploymentHash).substring(0, 10));
-                }
+            // log the start of UpdateCore
+            log.info("New catalog update from: " + VoltDB.instance().getCatalogContext().getCatalogLogString());
+            log.info("To: catalog hash: " + Encoder.hexEncode(catalogHash).substring(0, 10) +
+                    ", deployment hash: " + Encoder.hexEncode(deploymentHash).substring(0, 10));
+
+            try {
+                performCatalogVerifyWork(
```
vittvolt: If it fails then probably we should cleanup the temporary jar files ?

src/frontend/org/voltdb/sysprocs/UpdateCore.java
```diff
@@ -518,63 +467,38 @@ private final void performCatalogVerifyWork(
...
-            CatalogAndIds catalogStuff = CatalogUtil.getCatalogFromZK(zk);
-            // New update?
-            if (catalogStuff.version == expectedCatalogVersion) {
-                if (log.isInfoEnabled()) {
-                    log.info("New catalog update from: " + catalogStuff.toString());
-                    log.info("To: catalog hash: " + Encoder.hexEncode(catalogHash).substring(0, 10) +
-                            ", deployment hash: " + Encoder.hexEncode(deploymentHash).substring(0, 10));
-                }
+            String failureErrorMsg = null;
+            if (worksWithElastic == 0 && zk.exists(VoltZK.elasticJoinActiveBlocker, false) != null) {
```
vittvolt: I'm thinking if it should be moved to updateApplication() in UpdateApplicationBase ?

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -388,4 +377,130 @@ static protected boolean allowPausedModeWork(boolean internalCall, boolean admin
...
+                     "no longer live when attempting to apply the change.  This is likely " +
+                     "the result of multiple concurrent attempts to change the cluster " +
+                     "configuration.  Please make such changes synchronously from a single " +
+                     "connection to the cluster.";
+            return cleanupAndMakeResponse(ClientResponseImpl.GRACEFUL_FAILURE, errMsg);
+        }
+
+        // write the new catalog to a temporary jar file
+        errMsg = verifyAndWriteCatalogJar(ccr.catalogBytes);
+        if (errMsg != null) {
```
vittvolt: What if writing fails on one of the hosts, but completes successfully on all the other servers ? Should we also do a cleanup of jars ?

This is optional since current code simply overwrites the temporary jars in the future, so there should not be any issue with the current implementation.

tests/frontend/org/voltdb/regressionsuites/TestUpdateDeployment.java
```diff
@@ -269,14 +269,30 @@ public void testConsecutiveCatalogDeploymentRace() throws Exception
...
                 new File(newCatalogURL), null);
         // Then, update the users in the deployment
         SyncCallback cb2 = new SyncCallback();
         client.updateApplicationCatalog(cb2, null, new File(deploymentURL));
+        cb1.waitForResponse();
         cb2.waitForResponse();
-        assertEquals(ClientResponse.USER_ABORT, cb2.getResponse().getStatus());
-        assertTrue(cb2.getResponse().getStatusString().contains("Invalid catalog update"));
+
+        // System.err.println("cb1: " + Byte.toString(cb1.getResponse().getStatus()) + " " + cb1.getResponse().getStatusString());
```
xinjiacs: Clean up

tests/frontend/org/voltdb/regressionsuites/TestConcurrentUpdateCatalog.java
```diff
@@ -0,0 +1,267 @@
...
+
+    /*
+     * Helper function to check the results of 2 concurrent calls, assuming
+     * only one of them should become the winner
+     */
+    private void checkResults(SyncCallback cb1, SyncCallback cb2) throws Exception {
+        cb1.waitForResponse();
+        cb2.waitForResponse();
+
+//        System.err.println(cb1.getResponse().getStatus());
```
xinjiacs: Clean up

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1894,16 +1898,25 @@ private void logConfigInfo() {
...
                 e.printStackTrace();
+
+                // Defensive way to avoid any future problem. The local instance should shutdown.
+                hostLog.fatal("Error when writing catalog jar to disk: " + e.getMessage());
+                // Only 1 catalog update could happen at the same time. Before shutting down the
+                // local voltdb instance, we have to remove the global lock.
+                VoltZK.removeCatalogUpdateBlocker(VoltDB.instance().getHostMessenger().getZK(),
+                                                  VoltZK.uacActiveBlocker,
+                                                  hostLog);
+                VoltDB.crashLocalVoltDB("Fatal error when writing the catalog jar to disk.", true, e);
```
vittvolt: Since now the writing of jars in writeCatalogJarToFile() only depends on the mode, is crashing the local database still desirable ? Does VoltDB tolerate the case when a single host in the cluster does not have a proper catalog jar file ?

At least in the old implementation any IOException is simply logged only.

vittvolt:
There are currently some class loading issues with the TestMockUpdateApplication, I'm trying to solve it now.

xinjiacs:
@vittvolt Why the class loading issues are not shown on Jenkins tests?

vittvolt:
@xinjiacs There was a mistake I made in TestMockUpdateApplicationCatalog.java, I did not add @Test annotation to the test method.

tests/frontend/org/voltdb/sysprocs/TestMockUpdateApplicationCatalog.java
```diff
@@ -103,15 +97,56 @@ public void tearDown() throws Exception {
...
+        new File(newCatalogURL).delete();
+        new File(tempjarURL).renameTo(new File(newCatalogURL));
+
         try {
             m_client.updateApplicationCatalog(new File(newCatalogURL), new File(deploymentURL));
             fail("Update catalog should fail with version error.");
         } catch (ProcCallException e) {
-            assertTrue(e.getMessage().contains("Cannot load classes compiled with a higher version"));
-            assertTrue(e.getMessage().contains("Java 8"));
+            assertTrue(e.getMessage().contains("NoClassDefFoundError"));
```
vittvolt: Here I'm checking for NoClassDef error only instead of version error.

Reproducing the version error on Junit is not easy, I tried 2 ways:

1. Explicit add a test class in CatalogContext (mixing test code with production code, which I think is inappropriate)
2. Use PowerMockito to spy the CatalogContext class (which has nasty class duplicate name issue)

None of them tend to work out well.

xinjiacs:
Will merge it into an integration branch first

PR #4657, opened 2017-06-23
ENG-12034 When an error is seen distributing channel assignments, reacquire the channels before resending the assignments.

brians127:
This addresses a multiple node kill scenario where the Zookeeper data went stale midway through sending the assignments.

When the leader is distributing a new set of importer channel assignments, it will retry sending ALL assignments if ANY fail. This is fine if the first assignment fails, but in this scenario there is a race condition between the distribution of channel assignments and a mesh monitor responding to the node failure. In the original problem report, the leader has sent roughly 2/3 of the assignments out before Zookeeper throws a "bad version" error. When the entire re-send gets retried, the very first assignment gets flagged as a duplicate resulting in the reported error. The retry is aborted, meaning the remaining channel assignments don't get distributed.

Anyone who sees this in production should first be sure the cluster didn't initiate a global crash (if it did, all nodes will be gone soon and there are bigger issues). Then pause the cluster, verify the pause has completed (due to an unrelated issue), and then resume the cluster. This will restart the importers, refreshing their channel assignments. The data which was unable to import will remain in Kafka - data won't be loss unless Kafka deletes it.

The primary change is in ChannelDistributer. I also add some test logic in an attempt to make the issue more easily caught in JUnit, and I fix a rare unrelated NPE.

{TEST SCHEME STILL A WORK IN PROGRESS}

### DIFF
--- a/src/frontend/org/voltdb/importer/ChannelDistributer.java
+++ b/src/frontend/org/voltdb/importer/ChannelDistributer.java
@@ -722,7 +722,7 @@ public class ChannelDistributer implements ChannelChangeCallback {
                                 + ") Retrying channel assignment because write attempt to "
                                 + setter.path + " failed with " + setter.getCallbackCode()
                                );
-                        m_es.submit(new AssignChannels(seed));
+                        m_es.submit(new GetChannels(MASTER_DN));
                         return;
                     }
                 }
--- a/src/frontend/org/voltdb/importer/ChannelDistributer.java
+++ b/src/frontend/org/voltdb/importer/ChannelDistributer.java
@@ -641,16 +641,13 @@ public class ChannelDistributer implements ChannelChangeCallback {
         final NavigableMap<String,AtomicInteger> hosts = m_hosts.getReference();
 
         final int seed;
-        final boolean firstTime;
 
         AssignChannels(final int seed) {
             this.seed = seed;
-            firstTime = false;
         }
 
         AssignChannels() {
             seed = System.identityHashCode(this);
-            firstTime = true;
         }
 
         @Override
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/importer/ChannelDistributer.java
```diff
@@ -722,7 +722,7 @@ public void susceptibleRun() throws Exception {
                                 + ") Retrying channel assignment because write attempt to "
                                 + setter.path + " failed with " + setter.getCallbackCode()
                                );
-                        m_es.submit(new AssignChannels(seed));
+                        m_es.submit(new GetChannels(MASTER_DN));
```
nshi: I'm assuming that this is the magic line that fixed the original issue in the ticket. It seems to me that instead of restarting the current channel assignment task, you are falling back to let `GetChannels` trigger it again. Is the issue that restarting channel assignment immediately may result in a new assignment based on out-of-date knowledge or something else? It'd be great if you could elaborate on the original problem and your solution in the commit message. Thanks!

akhanzode:
Can you add a test that mimicks large number of channels and Zk failure due to host dead?
Also enable system tests on your branch to sun importer tests.

src/frontend/org/voltdb/importer/ChannelDistributer.java
brians127: Your understanding is the same as mine. There's a description in the JIRA ticket comments, but I'll add it to the pull request also. That will help me be sure it ends up in the final squash merge description as well.

brians127:
I don't plan on merging this pull request. The bug is not unit testable; we have to rely on existing system tests not picking up the original error. I filed a new pull request.

PR #4660, opened 2017-06-27
Linear algebra feature

diwakarj:
Innovation week project adding Linear algebra feature. parsing matrix .


src/ee/expressions/functionexpression.cpp
```diff
@@ -434,6 +434,9 @@ ExpressionUtil::functionFactory(int functionId, const std::vector<AbstractExpres
         case FUNC_INET6_ATON:
             ret = new UnaryFunctionExpression<FUNC_INET6_ATON>((*arguments)[0]);
             break;
+        case FUNC_T_ADD:
+            ret = new UnaryFunctionExpression<FUNC_T_ADD>((*arguments)[0]);
```
BillWhite: Isn't FUNC_T_ADD a binary function?

src/ee/expressions/functionexpression.h
```diff
@@ -139,6 +139,7 @@ static const int FUNC_DATEDIFF         = 140;
 static const int FUNC_SECONDS_MIDNIGHT = 141;
 static const int FUNC_CSC              = 142;
 static const int FUNC_SEC              = 143;
+static const int FUNC_T_ADD            = 144;
```
BillWhite: We'll have to sort these numbers out, but that's a detail.

src/ee/expressions/numericfunctions.h
```diff
@@ -314,6 +314,49 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+
+    // starting add for the col value
+    addr = addr+sizeof(int)*3;
+
+    // double ** mat = new double[r*c];
+    double **mat = (double **)malloc(r * sizeof(double *));
+    for (int i=0; i<r; i++)
+         mat[i] = (double *)malloc(c * sizeof(double));
+
+    for(int ii =0; ii <r; ii++)
```
BillWhite: Usually we use curly braces for everything.  This is just a style point.

src/ee/expressions/numericfunctions.h
```diff
@@ -314,6 +314,49 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+
+    int r = *row;
+    int c = *col;
+
+    // starting add for the col value
+    addr = addr+sizeof(int)*3;
+
+    // double ** mat = new double[r*c];
+    double **mat = (double **)malloc(r * sizeof(double *));
+    for (int i=0; i<r; i++)
```
BillWhite: I don't think you need to mallo anything here.   We don't use malloc at this level.  I think you need somehow to write a wrapper class which takes a const char *& or maybe const char * and some extent information (row size and col size) and exports indexing operations.  I'll look into roughing that out in a minute.

src/ee/expressions/numericfunctions.h
```diff
@@ -314,6 +314,49 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+    if (getValueType() != VALUE_TYPE_VARBINARY) {
+        throw SQLException(SQLException::dynamic_sql_error, "Unsupported non-VARBINARY type for Matrix function");
+    }
+    if (isNull()) {
+        return getNullValue(VALUE_TYPE_VARBINARY);
+    }
+
+    int32_t addr_len;
+    const int32_t *addr = (const int32_t *)getObject_withoutNull(&addr_len);
+
```
BillWhite: It looks like this is the right way to get at the data.  getObject_withoutNull will know whether the data is inline or out-of-line, so you don't need to worry about it.

src/ee/expressions/numericfunctions.h
```diff
@@ -292,6 +293,38 @@ template<> inline NValue NValue::callUnary<FUNC_SEC>() const {
...
+    double *addrp = (double *)getObject_withoutNull(&addr_lenp);
+
+    TensorWrapper P((char*)addrp,addr_lenp);
+    // create a transpose of Tensor
+
+    int32_t addrr_lenr = addr_lenp;
+    double *addrr = nullptr;
+
+    ValueFactory::getTempBinaryValue((const char *)addrr, addrr_lenr);
+
```
BillWhite: You never use addrr as a double * anywhere that I can see.  And it's not really a double * anyway.  It points to a tensor, which has some dope at the front.  So I would just make this a char *.

src/ee/expressions/numericfunctions.h
```diff
@@ -313,6 +346,163 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+    ValueFactory::getTempBinaryValue((const char *)addrr, addr_lenr);
+
+    TensorWrapper R((char*)addrr,addr_lenr);
+
+    for (int i = 0; i < P.numRows(); i += 1) {
+      for (int j = 0; j < Q.numCols(); j += 1) {
+          double sum = 0.0;
+          for (int k = 0; k < P.numCols(); k += 1) {
+              sum += P.get(i,k) + Q.get(k,j);
```
BillWhite: Is this right?  Why the extra loop?  Isn't is just
```
    for (i = 0 to P.numRows()) {
      for (j = 0 to Q.numCols) {
        R.set(i, j, P.get(i,j) + Q.get(i,j));
      }
    }
```

src/ee/expressions/numericfunctions.h
```diff
@@ -313,6 +346,163 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+
+/** implement the sql FUNC_T_TENSOR_MUL function for all numeric values */
+template<> inline NValue NValue::call<FUNC_T_TENSOR_MUL>(const std::vector<NValue>& arguments) {
+
+      assert(arguments.size() == 2);
+    const NValue& tensor1 = arguments[0];
+    const NValue& tensor2 = arguments[1];
+
+    const ValueType tesnsor1Type = tensor1.getValueType();
+    const ValueType tesnsor2Type = tensor2.getValueType();
```
BillWhite: This seems like a misspelling tesnsor2 -> tensor2.

Eclipse will let you change all occurrences of a misspelling at once.  Select the misspelled word and right click on it.  Then in the context menu go to Refactor->Rename.  Correct the spelling and push enter to correct all of them.

src/ee/expressions/numericfunctions.h
```diff
@@ -313,6 +346,163 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+}
+
+/** implement the sql FUNC_T_TENSOR_MUL function for all numeric values */
+template<> inline NValue NValue::call<FUNC_T_SCALAR_MUL>(const std::vector<NValue>& arguments) {
+
+      assert(arguments.size() == 2);
+    const NValue& tensor = arguments[0];
+    const NValue& scalar = arguments[1];
+
+    const ValueType tesnsorType = tensor.getValueType();
```
BillWhite: Another misspelling.

src/ee/expressions/numericfunctions.h
```diff
@@ -313,6 +346,163 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+    const ValueType scalarType = scalar.getValueType();
+
+    if (tesnsorType != VALUE_TYPE_VARBINARY && isNumeric(scalarType)) {
+        throw SQLException(SQLException::dynamic_sql_error, "Unsupported non-VARBINARY type for Matrix function");
```
BillWhite: Should there be a not before isNumeric?
Should && be ||?
Should this be two tests?  The error message will be misleading if the tensor is tensor but the scalar value has the wrong type.

src/ee/expressions/numericfunctions.h
```diff
@@ -313,6 +346,163 @@ template<> inline NValue NValue::call<FUNC_POWER>(const std::vector<NValue>& arg
...
+    }
+
+    if (tensor.isNull() || scalar.isNull()) {
+        return getNullValue(VALUE_TYPE_VARBINARY);
+    }
+
+    int32_t base = scalar.castAsIntegerAndGetValue();
```
BillWhite: I would think it would be better to make base a double.

src/ee/common/TensorWrapper.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef TENSOR_WRAPPER_H
+#define TENSOR_WRAPPER_H
+#include <iostream>
+#include <cassert>
+#include <cstring>
+#include "common/SQLException.h"
+//#include ... Some file to provide a declaration for SQLException.
```
diwakarj: copy paste error, changed it.

src/ee/expressions/functionexpression.h
```diff
@@ -139,6 +139,16 @@ static const int FUNC_DATEDIFF         = 140;
 static const int FUNC_SECONDS_MIDNIGHT = 141;
 static const int FUNC_CSC              = 142;
 static const int FUNC_SEC              = 143;
+/*
```
diwakarj: Matched the values of four implemented function.
static const int FUNC_T_TENSOR_MUL     = 21027; // (T1, T2)
static const int FUNC_T_TR             = 21028; // (T)
static const int FUNC_T_ADD            = 21029; // (T1, T2)
static const int FUNC_T_SCALAR_MUL     = 21030; // (S, T)

src/ee/expressions/numericfunctions.h
diwakarj: changed to char *, used casting to match (const char *) at other places. but TensorWrapper expect char *. can modify it also later.

src/ee/expressions/numericfunctions.h
diwakarj: Yes it was wrong, changed it.

src/ee/expressions/numericfunctions.h
diwakarj: yes i did find all and changed it in sublime text.

src/ee/expressions/numericfunctions.h
diwakarj: corrected.

src/ee/expressions/numericfunctions.h
diwakarj: i changed && to || (thats the right way). the isNull test is similar to other API and it returns the NULL value. But if its not Null and types don't match then it throws error. Not sure if i need to remove one. For now kept it.

src/ee/expressions/numericfunctions.h
diwakarj: Changed it to Double. Also did few changes int he code which were wrong.

BillWhite:
This is an innovation week project.  It's been overtaken by events.

PR #4661, opened 2017-06-28
ENG-9309: Tolerate missing shortCircuitRead CI handle in CIHM findHandle() and removeHandle()

slowsoul:


src/frontend/org/voltdb/ClientInterfaceHandleManager.java
```diff
@@ -287,6 +291,22 @@ Iv2InFlight findHandle(long ciHandle)
             m_outstandingTxns--;
             return inflight;
         }
+        else if (isShortCircuitReadHandle(ciHandle)) {
```
xinjiacs: You do not need the ELSE IF, but just be able to use IF. 

Also you can add a comment here:
// Normal short circuit read has been handled above. 
// If this handle is for short circuit read, it must be a NT transaction.
// NT transactions are handled as short circuit read for simplicity.

src/frontend/org/voltdb/ClientInterfaceHandleManager.java
```diff
@@ -357,6 +377,10 @@ Iv2InFlight removeHandle(long ciHandle)
...
```
xinjiacs: Similar to use IF simply.

src/frontend/org/voltdb/ClientInterfaceHandleManager.java
slowsoul: Good point, thanks. New commit has been pushed to address these.

PR #4663, opened 2017-06-29
Eng-12542-VoltdbInit-Relative-Path

sycai:
nit with absolute path: Relative paths from the deployment file become absolute. Default path values are absolute too.

Init with relative path: Paths from the deployment file are preserved. Default path values are set to be relative to the voltdbroot. The voltdbroot path itself is absolute.

Start with absolute path: Path properties stay the same as they are in the "path.properties" file.

Start with relative path: Path properties stay the same as they are in the "path.properties" file.

Get with absolute path: Paths in the deployment file are the same as those in the voltdbroot.

Get with relative path: Paths in the deployment file are the same as those in the voltdbroot.


src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -193,9 +193,15 @@ default Properties asProperties() {
                 throw new SettingsException("Missing VoltDB root " +
                                             "information in path.properties file.");
             }
-            mb.put(VOLTDBROOT_PATH_KEY, getVoltDBRoot().getCanonicalPath());
+            File voltdbroot = getVoltDBRoot().getCanonicalFile();
+            mb.put(VOLTDBROOT_PATH_KEY, voltdbroot.getCanonicalPath());
             for (Map.Entry<String, File> e: getManagedArtifactPaths().entrySet()) {
-                mb.put(e.getKey(), e.getValue().getCanonicalPath());
+                File path = e.getValue();
+                if (path.isAbsolute()) {
```
hhakimi53: Can u pls add comment stating case for which path will be absolute vs not?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -470,77 +470,125 @@ public String getVoltDBRootPath(PathsType.Voltdbroot path) {
         if (isRunningWithOldVerbs()) {
            return path.getPath();
         }
-        return m_nodeSettings.getVoltDBRoot().getPath();
+        try {
```
hhakimi53: Can include the path for which it failed to canonicalize. Something like

```Java
       File root = m_nodeSettings.getVoltDBRoot();
       String path = voltdbRoot.getPath();
        try {
            return root.getCanonicalPath();
        } catch (IOException e) {
            throw new SettingsException("failed to canonicalize voltdbroot " + path + ". Reason: " + e.getMessage());
        }
```

Similar recommendation for all other paths below - command log, snapshot, dr/export overflow and so forth.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2708,4 +2711,4 @@ public static String printProcedureDetail(Procedure proc, String sqlText) {
         }
         return sb.toString();
     }
-}
+}
```
hhakimi53: Add new line at eof and this will go away

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -828,6 +829,20 @@ private void inspectGetCommand() {
...
+            try {
+            Settings.initialize(m_voltdbRoot);
+            File currDir = new File("").getCanonicalFile();
+            File voltdbroot = m_voltdbRoot.getCanonicalFile();
+            String relativePath = currDir.toPath().relativize(voltdbroot.toPath()).toString();
+            return ImmutableMap.<String, String>builder()
+                    .put(NodeSettings.VOLTDBROOT_PATH_KEY, relativePath)
+                    .build();
+            } catch (IOException e) {
+                throw new SettingsException("Failed to relativize voltdbroot");
```
hhakimi53: If feasible, can we pass the path information in the exception along with the exception message?

tests/frontend/org/voltdb/utils/TestCatalogUtil.java
```diff
@@ -568,6 +568,39 @@ public void testRelativePathsToVoltDBRoot() throws Exception
...
+
+        final File tmpDeploy = VoltProjectBuilder.writeStringToTempFile(deploy);
+        CatalogUtil.compileDeployment(catalog, tmpDeploy.getPath(), false);
+
+        File fileExportDir = new File(voltdbroot, "file_export");
+        assertTrue("file export directory: " + fileExportDir.getAbsolutePath() + " does not exist",
+                   fileExportDir.exists());
+        assertTrue("file export directory: " + fileExportDir.getAbsolutePath() + " is not a directory",
+                   fileExportDir.isDirectory());
+    }
```
hhakimi53: Can add or enhance existing test case to use explicit path for file export too. that will cover both default as well explicit path.

```
            "<deployment>" +
            "    <cluster kfactor=\"0\" schema=\"ddl\" sitesperhost=\"6\" />" +
            "    <export>" +
            "        <configuration enabled=\"true\" target=\"log\" type=\"file\">" +
...
...
            "            <property name="outdir"> path xyz </property>
            "        </configuration>" +
            "    </export>" +
            "</deployment>";
```

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -828,6 +829,24 @@ private void inspectGetCommand() {
                     .build();
         }
 
+        public Map<String,String> asRelativePathSettingsMap() {
+            try {
```
brians127: Indentation looks wrong here

src/frontend/org/voltdb/RealVoltDB.java
sycai: Hi Huzefa, I added "absResolve()"  in the NodeSettings.java for resolving absolute path. Now most path error messages would be collectively handled there.

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -828,6 +829,24 @@ private void inspectGetCommand() {
...
+                Settings.initialize(m_voltdbRoot);
+                File currDir = new File("").getCanonicalFile();
+                File voltdbroot = m_voltdbRoot.getCanonicalFile();
+                String relativePath = currDir.toPath().relativize(voltdbroot.toPath()).toString();
```
hhakimi53: Suggestion: Can relativize() use the absolute path and provide the expected behavior in that scenario? 
If so, in case if getCanonicalFile() generates an IOException, logic can fall back absolute file (or even use absolute file to get relativize path) and use that  to form the relativized path.

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -82,6 +82,18 @@ default File resolve(File path) {
         return path.isAbsolute() ? path : new File(getVoltDBRoot(), path.getPath());
     }
 
+    default File absResolve(File path) {
```
hhakimi53: Suggestion: May be can rename it to resolveToAbsolutePath() or something on that lines.

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -82,6 +82,18 @@ default File resolve(File path) {
         return path.isAbsolute() ? path : new File(getVoltDBRoot(), path.getPath());
```
hhakimi53: Is this functionality needed as there is absResolve() now?

sycai: Yes we still need this method as it is called in "getManagedArtifactPaths()" for storing path information.

hhakimi53: ok

src/frontend/org/voltdb/VoltDB.java
sycai: I just confirmed that "relativize()" does not work properly on absolute paths containing ".." So It would probably be better to use canonical paths here.

hhakimi53: ok, thanks for trying it out. Minor: May be can move the relativize out of try-catch block as toPath() and relativize() does not generate IOException.

PR #4665, opened 2017-07-03
Eng-539-boxed-sp

dm89:
support boxed types in stored procedure java classes


src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -276,39 +285,44 @@ public static Object tryToMakeCompatible(final Class<?> expectedClz, final Objec
         if (inputClz == Long.class) {
             if (expectedClz == long.class) return param;
             if ((Long) param == VoltType.NULL_BIGINT) return nullValueForType(expectedClz);
+            if (expectedClz == Long.class) return param;
```
wolffcm: It's a common pattern in these changes that you're looking for both boxed and unboxed classes, so your changes could be simplified if you defined some private helper functions,
```Java
private static boolean isLongClass(Class<?> clz) { return clz == Long.class || clz == long.class; }
```
(And so on for the other primitive types.)

src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -200,7 +202,8 @@ else if (inputLength == 0) {
             return Array.newInstance(expectedComponentClz, 0);
         }
         // hack to make strings work with input as bytes
-        else if ((inputComponentClz == byte[].class) && (expectedComponentClz == String.class)) {
+        else if ((inputComponentClz == byte[].class || inputComponentClz == Byte[].class)
```
wolffcm: Most of your changes allow for boxed types as the expected type, but in this case you're adding logic for boxed types as the input component type.  Is this what you intended?  If so, would be good to have tests for that.  (Maybe you've added them but I haven't started reviewing tests yet.)

src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -214,12 +217,18 @@ else if ((inputComponentClz == byte[].class) && (expectedComponentClz == String.
...
         }
         // hack to make varbinary work with input as hex string
-        else if ((inputComponentClz == String.class) && (expectedComponentClz == byte[].class)) {
+        else if ((inputComponentClz == String.class) &&
+                (expectedComponentClz == byte[].class || expectedComponentClz == Byte[].class )) {
             byte[][] values = new byte[inputLength][];
+            Byte[][] boxvalues = new Byte[inputLength][];
             for (int i = 0; i < inputLength; i++) {
                 values[i] = Encoder.hexDecode((String) Array.get(param, i));
+                boxvalues[i] = ArrayUtils.toObject( values[i] );
```
wolffcm: You're populating both `values` and `boxvalues` even though one of those will be needed.  Can you break refactor this to be more efficient?

tests/frontend/org/voltdb/TestParameterConverter.java
```diff
@@ -101,20 +156,28 @@ public void testCommasStringIntegerToInt() throws Exception
...
     {
         Object r = ParameterConverter.
             tryToMakeCompatible(double.class, "2,301,100.23");
         assertTrue("expect integer", r.getClass() == Double.class);
-        assertEquals(new Double(2301100.23), ((Double)r).doubleValue());
+        assertEquals(2301100.23, ((Double)r).doubleValue());
     }
 
-    public void testNULLToInt() throws Exception
+    public void testCommasStringIntegerToBoxedDouble() throws Exception
```
wolffcm: Did you get rid of the `testNULLToInt` test that used to be here?  Seems like that was unintentional?

wolffcm: Or maybe the method that was here is now tested below in `testNulls`.

tests/frontend/org/voltdb/TestParameterConverter.java
```diff
@@ -196,6 +282,27 @@ public void testNulls()
...
+        assertEquals(VoltType.NULL_BIGINT, ParameterConverter.tryToMakeCompatible(long.class, null));
+        assertEquals(VoltType.NULL_FLOAT, ParameterConverter.tryToMakeCompatible(double.class, null));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Byte.class, null));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Short.class, null));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Integer.class, null));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Long.class, null));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Double.class, null));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(byte[].class, VoltType.NULL_TINYINT));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Byte[].class, VoltType.NULL_TINYINT));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(short[].class, VoltType.NULL_SMALLINT));
```
wolffcm: Seems like your missing cases for expected classes `Short[].class`, `Integer[].class`, `Long[].class`, `Double[].class` here?

tests/frontend/org/voltdb/TestParameterConverter.java
```diff
@@ -281,4 +500,34 @@ public static void testBigDecimalFailWithInvalidConversion(Class<?> expectedClz,
...
+            * parameters */
+        } catch (Exception ex) {
+            assertTrue(ex.getMessage().contains(
+                    "Array / Scalar parameter mismatch ([I to java.lang.Integer)"));
+        }
+    }
+
+    public void testIntArrayToIntegerArray() throws Exception
+    {
+        int[] t = {1, 2, 3};
```
wolffcm: This looks like it's a copy of the method above.  I think you meant to have a test case that actually succeeded?

tests/frontend/org/voltdb/TestParameterConverter.java
```diff
@@ -281,4 +500,34 @@ public static void testBigDecimalFailWithInvalidConversion(Class<?> expectedClz,
...
+            ParameterConverter.
+                tryToMakeCompatible(Integer[].class, t);
+            /* Arrays can be quite large so it doesn't make sense to silently do the conversion
+            * and incur the performance hit. The client should serialize the correct invocation
+            * parameters */
+        } catch (Exception ex) {
+            assertTrue(ex.getMessage().contains(
+                    "tryScalarMakeCompatible: Unable to match parameter array:java.lang.Integer to provided int"));
+        }
+    }
```
wolffcm: Some nice testing here, including test of existing code that was missing originally.  Thanks!

Before you merge, I'd like to see more testing of arrays, especially arrays of byte and arrays of byte[].  That is, in SQL array of VARBINARY would have an expected class of `Byte[][]`.

We have some open bugs in this area, so you might hit some issues, but it would be good to at least have tests that characterize current behavior.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -302,6 +303,49 @@ else if (results[0].getLong(1) == 500)
...
+            client.callProcedure("InsertBoxed", table, new Long(1), "desc", new Long(100), new Double(14.5));
+            client.callProcedure("InsertBoxed", table, new Long(2), "desc", new Long(100), new Double(14.5));
+            client.callProcedure("InsertBoxed", table, new Long(3), "desc", new Long(100), new Double(14.5));
+            client.callProcedure("InsertBoxed", table, new Long(6), "desc", new Long(300), new Double(14.5));
+            client.callProcedure("InsertBoxed", table, new Long(7), "desc", new Long(300), new Double(14.5));
+            client.callProcedure("InsertBoxed", table, new Long(8), "desc", new Long(500), new Double(14.5));
+
+            String query =
+                String.format("select count(*), %s.NUM from %s group by %s.NUM",
+                              table, table, table);
```
wolffcm: Would be good to add an ORDER BY clause here so the query is deterministic.

wolffcm: Actually I see below that your test accounts for variation in row order, but a more concise way to write this test would be to use an ORDER BY clause here and then just use the method `RegressionSuite.assertContentOfTable` to verify the result is as expected.

src/frontend/org/voltdb/ParameterConverter.java
wolffcm: Based on this change I would expect to see some tests where provided input parameters are of type `Integer[]` as well as `int[]`, and similarly for the other types, etc.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: Would be good also to have some procedures that accept arrays of boxed types.  SQL statements with predicates of the form `WHERE V IN ?` expect arrays.  Let me know if you have questions about this.

src/frontend/org/voltdb/ParameterConverter.java
dm89: Actually this method takes the parameter as Object - so all input types will be boxed - was not sure why there was `byte[]`

```
private static Object tryToMakeCompatibleArray(
            final Class<?> expectedComponentClz,
            final Class<?> inputComponentClz,
            Object param)
```

tests/frontend/org/voltdb/TestParameterConverter.java
```diff
@@ -196,6 +282,27 @@ public void testNulls()
...
         assertEquals(VoltType.NULL_BIGINT, ParameterConverter.tryToMakeCompatible(long.class, VoltType.NULL_BIGINT));
         assertEquals(VoltType.NULL_FLOAT, ParameterConverter.tryToMakeCompatible(double.class, VoltType.NULL_FLOAT));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Byte.class, VoltType.NULL_TINYINT));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Short.class, VoltType.NULL_SMALLINT));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Integer.class, VoltType.NULL_INTEGER));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Long.class, VoltType.NULL_BIGINT));
+        assertEquals(null, ParameterConverter.tryToMakeCompatible(Double.class, VoltType.NULL_FLOAT));
+        assertEquals(VoltType.NULL_TINYINT, ParameterConverter.tryToMakeCompatible(byte.class, null));
+        assertEquals(VoltType.NULL_SMALLINT, ParameterConverter.tryToMakeCompatible(short.class, null));
+        assertEquals(VoltType.NULL_INTEGER, ParameterConverter.tryToMakeCompatible(int.class, null));
```
dm89: Here is the `testNULLToInt` test

tests/frontend/org/voltdb/TestParameterConverter.java
dm89: add the tests

tests/frontend/org/voltdb/TestParameterConverter.java
dm89: This is for the `Array / Scalar parameter mismatch ([I to java.lang.Integer)` error from original code which has not been tested

src/frontend/org/voltdb/ParameterConverter.java
```diff
@@ -121,7 +129,7 @@ else if (value != null) {
      * Corresponding code for handling hex literals appears in HSQL's ExpressionValue class
      * and in voltdb.expressions.ConstantValueExpression.
      */
-    private static Object convertStringToPrimitive(String value, final Class<?> expectedClz)
+    private static Object convertStringToPrimitiveOrPrimitiveWrapper(String value, final Class<?> expectedClz)
```
yzhang1991: Please update the comment above accordingly.

src/frontend/org/voltdb/ParameterConverter.java
yzhang1991: Should be using `isLongClass()` here.
Same for the other types below.

src/frontend/org/voltdb/compiler/ProcedureCompiler.java
```diff
@@ -436,21 +436,11 @@ else if (partitioning.getCountOfPartitionedTables() == 0) {
             else
                 param.setIsarray(false);
 
-            // boxed types are not supported parameters at this time
-            if ((cls == Long.class) || (cls == Integer.class) || (cls == Short.class) ||
-                (cls == Byte.class) || (cls == Double.class) ||
-                (cls == Character.class) || (cls == Boolean.class))
```
yzhang1991: Is there a mapping type for `Character`?

src/frontend/org/voltdb/ParameterConverter.java
dm89: Actually no, because if the input value is NULL for that type (say  VoltType.NULL_BIGINT), it gets converted to `null` for boxed types

src/frontend/org/voltdb/compiler/ProcedureCompiler.java
dm89: did not find `Character` class mapping

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+                            + "Incompatible parameter type: can not convert type 'double[]' to 'INLIST_OF_BIGINT' "
+                            + "for arg 0 for SQL stmt: SELECT * FROM ENG_12105 WHERE NUM IN ?;. "
+                            + "Try explicitly using a long[] parameter"));
+        }
+
+        try {
+            client.callProcedure("InPrimitiveArrays", "BIGDS", null, null, null, null, null,
+                    new BigDecimal[]{new BigDecimal(1), new BigDecimal(2), new BigDecimal(3)}, null, null).getResults();
+        } catch (ProcCallException e) {
+            assertTrue(e.getMessage().contains("VOLTDB ERROR: USER ABORT\n"
```
wolffcm: This error might be an issue with how Java handles vararg methods...  Java is interpreting the array as three separate arguments (because `voltQueueuSQL` is a vararg method).  If you cast the array arguments to `Object` then Java should treat the array as a single argument, which is what you want.  If you do this, does the query then succeed?

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+                            + "  Number of arguments provided was 3 where 1 was expected "
+                            + "for statement SELECT * FROM ENG_12105 WHERE DEC IN ?;"));
+        }
+
+        try {
+            client.callProcedure("InPrimitiveArrays", "STRS", null, null, null, null, null, null,
+                    new String[]{"1", "2", "3"}, null).getResults();
+        } catch (ProcCallException e) {
+            assertTrue(e.getMessage().contains("VOLTDB ERROR: USER ABORT\n"
+                            + "  Number of arguments provided was 3 where 1 was expected "
```
wolffcm: Likewise for this error.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+            byte[] insByteArr = Encoder.hexDecode("0A");
+            results = client.callProcedure("InPrimitiveArrays", "INSBYTES", null, null,
+                    null, null, null, null, null, insByteArr).getResults();
+            assertEquals(1, results[0].getRowCount());
+        }
+
+        String query =
+                String.format("select * from ENG_12105");
+        results = client.callProcedure("@AdHoc", query).getResults();
+        System.out.println(results);
```
wolffcm: Did you mean to assert that the table contains something here?  If not, would be good to make the test less noisy and remove the println.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+            assertTrue(e.getMessage().contains("VOLTDB ERROR: UNEXPECTED FAILURE:\n"
+                            + "  org.voltdb.VoltTypeException: Unsupported type: VoltType.INLIST_OF_BIGINT"));
+        }
+
+        if( !isHSQL() ) {
+            // HSQL does not convert convert null to null value for TIMESTAMP
+            byte[] insByteArr = Encoder.hexDecode("0A");
+            results = client.callProcedure("InPrimitiveArrays", "INSBYTES", null, null,
+                    null, null, null, null, null, insByteArr).getResults();
+            assertEquals(1, results[0].getRowCount());
```
wolffcm: You can use `RegressionSuite.validateDMLTupleCount` here.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+            client.callProcedure("BoxedByteArrays", "STR", new Integer(2), null,
+                    null, null, null, "x'3A'");
+        } catch (ProcCallException e) {
+            assertTrue(e.getMessage().contains("Incompatible parameter type: "
+                        + "can not convert type 'String' to 'VARBINARY' for arg 1 for SQL stmt"));
+        }
+
+        if( !isHSQL() ) {
+            // HSQL does not convert Strings to BigInt
+            client.callProcedure("BoxedByteArrays", "DSTR", new Integer(2), null,
```
wolffcm: If this works it would be good to lock down the result.

wolffcm: Actually for any query that is succeeding.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
wolffcm: I mean cast to `Object` inside the stored procedure in the call to `voltQueueSQL`.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
dm89: `new byte[][]{ Encoder.hexDecode("0A"), Encoder.hexDecode("1E") }` needs `@AdHoc` query and does not seem to work with `exec ProcName params..`

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+        try {
+            byte[][] byteArr = new byte[][]{ Encoder.hexDecode("0A"), Encoder.hexDecode("1E") };
+            client.callProcedure("InPrimitiveArrays", "BYTES", byteArr, null,
+                    null, null, null, null, null, null).getResults();
+        } catch (ProcCallException e) {
+            assertTrue(e.getMessage().contains("VOLTDB ERROR: USER ABORT\n"
+                          + "  Unknown type VoltType.INLIST_OF_BIGINT can not be converted "
+                          + "to NULL representation for arg 0 for SQL stmt: "
+                          + "SELECT * FROM ENG_12105 WHERE VARBIN IN ?"));
+        }
```
dm89: updated with the function call

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -346,6 +350,176 @@ else if (results[0].getLong(1) == 500)
...
+
+        try {
+            client.callProcedure("InPrimitiveArrays", "SHORTS", null, new short[]{1, 2, 3},
```
dm89: I overlooked it. updated to correct parameters now

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
dm89: With cast to `Object`, some statements work like `short, int, long`

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
dm89: used to for debugging - removed it

PR #4667, opened 2017-07-05
ENG-12779 Queue depth

dhan111:
Currently we have the following performance statistics:

- Initiator stats or latency stats (intra-cluster latency)
- Procedure stats (partition execution time)
- Command log stats

The closest to client-side latency is the initiator stats. It doesn't include the network round-trip time between the client and the server and any delay incurred on the client. The procedure stats only show the duration of the run() methods of procedures. It only accounts for the computation portion of SPs.

Looking at these stats typically gives a good idea of the cluster performance, but they are not enough to diagnose any latency issues. There is a big gap between the intra-cluster latency and the procedure execution time. They are usually magnitudes apart.

A big portion of the difference between the intra-cluster latency and the procedure execution time is the time the procedure request sitting in the partition queue. If command logging is used, back pressure from the command log and/or fsyncing time can also account for the difference.

Exposing the partition queue depth can help us answer some latency-related questions. The stats include the following:

- Queue depth
- Average wait time in the past 5s
- Max wait time in the past 5s

### DIFF
--- a/src/frontend/org/voltdb/QueueDepthTracker.java
+++ b/src/frontend/org/voltdb/QueueDepthTracker.java
@@ -28,7 +28,8 @@ import org.voltdb.iv2.SiteTasker;
 /**
  * A class to track and generate statistics regarding queue depth.
  * Generate information on instantaneous queue depth and number of tasks
- * pulled from queue, average wait time and max wait time within a 5-second window
+ * pulled from queue, average wait time and max wait time (in microseconds)
+ * within a 5-second window
  */
 public class QueueDepthTracker extends SiteStatsSource {
 
@@ -40,12 +41,12 @@ public class QueueDepthTracker extends SiteStatsSource {
     private long m_lastWaitTime;
     private Deque<QueueStatus> m_historicalData;
     private LinkedTransferQueue<SiteTasker> m_tasks;
-    private long m_maxWaitTimeWindowSize = 5000000000L; // window size set to 5 seconds
+    private long m_maxWaitTimeWindowSize = 5_000_000_000L; // window size set to 5 seconds
     private long m_maxWaitLastLogTime;
     private long m_recentMaxWaitTime;
     private long m_recentTotalWaitTime;
     private long m_recentPollCount;
-    private long m_recentWindowSize = m_maxWaitTimeWindowSize / 500;
+    private long m_recentWindowSize = m_maxWaitTimeWindowSize / 10; // recent window size set to 0.5 second
 
     public class QueueStatus {
         public long timestamp;
@@ -124,7 +125,7 @@ public class QueueDepthTracker extends SiteStatsSource {
         if (nextTask == null) {
             currentWaitTime = 0;
         } else {
-            currentWaitTime = currentTime - nextTask.getTimestamp();
+            currentWaitTime = currentTime - nextTask.getQueueOfferTime();
         }
         // check historicalMaxWaitTime, report max wait time and mean wait time in window
         long maxWaitTimeInWindow = Math.max(currentWaitTime, m_recentMaxWaitTime);
@@ -143,8 +144,9 @@ public class QueueDepthTracker extends SiteStatsSource {
         }
         rowValues[columnNameToIndex.get("CURRENT_DEPTH")] = m_depth;
         rowValues[columnNameToIndex.get("POLL_COUNT")] = totalPollCountInWindow;
-        rowValues[columnNameToIndex.get("AVG_WAIT")] = totalWaitTimeInWindow / Math.max(1, totalPollCountInWindow);
-        rowValues[columnNameToIndex.get("MAX_WAIT")] = maxWaitTimeInWindow;
+        // wait times are in microseconds
+        rowValues[columnNameToIndex.get("AVG_WAIT")] = (totalWaitTimeInWindow / Math.max(1, totalPollCountInWindow)) / 1000;
+        rowValues[columnNameToIndex.get("MAX_WAIT")] = maxWaitTimeInWindow / 1000;
 
         super.updateStatsRow(rowKey, rowValues);
     }
--- a/src/frontend/org/voltdb/QueueDepthTracker.java
+++ b/src/frontend/org/voltdb/QueueDepthTracker.java
@@ -111,7 +111,7 @@ public class QueueDepthTracker extends SiteStatsSource {
         super.populateColumnSchema(columns);
         columns.add(new ColumnInfo("CURRENT_DEPTH", VoltType.INTEGER));
         columns.add(new ColumnInfo("POLL_COUNT", VoltType.BIGINT));
-        columns.add(new ColumnInfo("MEAN_WAIT", VoltType.BIGINT));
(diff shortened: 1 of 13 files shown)


src/frontend/org/voltdb/iv2/SiteTaskerQueue.java
```diff
@@ -20,30 +20,40 @@
...
         if (task == null) {
             m_starvationTracker.beginStarvation();
         } else {
+            m_queueDepthTracker.pollUpdate(task.getTimestamp());
             return task;
         }
         try {
-            return CoreUtils.queueSpinTake(m_tasks);
+            task = CoreUtils.queueSpinTake(m_tasks);
+            if (task != null) {
```
JadenW: Will queueSpinTake() ever return a null?

src/frontend/org/voltdb/QueueDepthTracker.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public QueueDepthTracker(long siteId) {
+        super(siteId, false);
+        m_historicalData = new ArrayDeque<QueueStatus>();
+    }
+    private int m_depth;
+    private long m_lastWaitTime;
+    private Deque<QueueStatus> m_historicalData;
+    private LinkedTransferQueue<SiteTasker> m_tasks;
+    private long m_maxWaitTimeWindowSize = 5000000000L; // window size set to 5 seconds
```
JadenW: nitpick: add underscore for long value. http://docs.oracle.com/javase//8/docs/technotes/guides/language/underscores-literals.html

tests/frontend/org/voltdb/regressionsuites/statistics/TestStatisticsSuite.java
```diff
@@ -664,7 +699,7 @@ public void testManagementStats() throws Exception {
         results = client.callProcedure("@Statistics", "MANAGEMENT", 0).getResults();
         // eight aggregate tables returned.  Assume that we have selected the right
         // subset of stats internally, just check that we get stuff.
-        assertEquals(8, results.length);
+        assertEquals(9, results.length);
     }
```
JadenW: nitpick: update the comment above "eight aggregate" too.

src/frontend/org/voltdb/iv2/ElasticJoinProducer.java
```diff
@@ -125,6 +126,7 @@ private void doInitiation(RejoinMessage message)
         m_mailbox.send(m_coordinatorHsId, msg);
 
         m_taskQueue.offer(this);
+        m_taskQueue.setQueueDepthTracker(new QueueDepthTracker(m_partitionId));
```
JadenW: Do we have to do this for RejoinProducer too?

src/frontend/org/voltdb/QueueDepthTracker.java
```diff
@@ -0,0 +1,179 @@
...
+                    maxWaitTimeInWindow = Math.max(maxWaitTimeInWindow, status.maxWait);
+                    totalWaitTimeInWindow += status.totalWait;
+                    totalPollCountInWindow += status.pollCount;
+                }
+            }
+        }
+        rowValues[columnNameToIndex.get("CURRENT_DEPTH")] = m_depth;
+        rowValues[columnNameToIndex.get("POLL_COUNT")] = totalPollCountInWindow;
+        rowValues[columnNameToIndex.get("AVG_WAIT")] = totalWaitTimeInWindow / Math.max(1, totalPollCountInWindow);
+        rowValues[columnNameToIndex.get("MAX_WAIT")] = maxWaitTimeInWindow;
```
nshi: Maybe microseconds instead of nanoseconds are better here. I don't we need nanosecond precision for wait time, and it'll be consistent with the procedure execution time statistics.

src/frontend/org/voltdb/StatsAgent.java
```diff
@@ -526,6 +526,9 @@ private void collectPartitionCount(PendingOpsRequest psr)
         case STARVATION:
             stats = collectStats(StatsSelector.STARVATION, interval);
             break;
+        case QUEUEDEPTH:
+            stats = collectStats(StatsSelector.QUEUEDEPTH, interval);
```
nshi: Since the statistics include wait times as well, maybe a name like QUEUE is better. What do you think, @ajgent?

src/frontend/org/voltdb/iv2/ElasticJoinProducer.java
```diff
@@ -123,7 +124,7 @@ private void doInitiation(RejoinMessage message)
         // respond to the coordinator with the sink HSID
         RejoinMessage msg = new RejoinMessage(m_mailbox.getHSId(), -1, sinkHSId);
         m_mailbox.send(m_coordinatorHsId, msg);
-
+        m_taskQueue.setQueueDepthTracker(new QueueDepthTracker(m_partitionId));
```
nshi: The `m_taskQueue` should already been initialized with a tracker by `BaseInitiator` at this point. If you are adding this because of some unit test failures, that might be because the unit tests construct `ElasticJoinProducer` without a real `BaseInitiator`. In that case, the proper place to set the tracker is in the unit tests.

Maybe you can follow the starvation tracker initialization pattern because it is also used in the `SiteTaskerQueue`.

src/frontend/org/voltdb/iv2/SiteTasker.java
```diff
@@ -19,19 +19,30 @@
...
 import java.io.IOException;
 
-import org.voltdb.rejoin.TaskLog;
-
 import org.voltdb.SiteProcedureConnection;
+import org.voltdb.rejoin.TaskLog;
 
 public abstract class SiteTasker {
 
+    private long timestamp = -1L;
```
nshi: Maybe consider renaming this variable so that it tells what specific timestamp it stores, e.g. queueOfferTime.

src/frontend/org/voltdb/StatsAgent.java
nshi: Andrew suggested naming it QUEUE.

src/frontend/org/voltdb/QueueDepthTracker.java
```diff
@@ -0,0 +1,181 @@
...
+        m_lastWaitTime = currentTime - offerTime;
+        // if max wait time was last logged less than m_recentWindowSize ago
+        // keep the max wait time in m_recentMaxWaitTime
+        // or log and reset the recentMaxWaitTime, update last log time
+        if (currentTime - m_maxWaitLastLogTime < m_recentWindowSize) {
+            m_recentMaxWaitTime = Math.max(m_recentMaxWaitTime, m_lastWaitTime);
+            m_recentTotalWaitTime += m_lastWaitTime;
+            m_recentPollCount++;
+        } else {
+            m_historicalData.addLast(new QueueStatus(currentTime,
```
manjujames: If the first pollUpdate that comes in is after recentWindowSize time, we will add an unnecessary entry into historical data. If we check recentPollCount>0 before adding it, we can avoid that. However, I think this is harmless because this doesn't affect the calculation in updateStatsRow, so this may be ok.

PR #4668, opened 2017-07-05
Fix ENG-12574

yzhang1991:


src/ee/expressions/functionexpression.cpp
```diff
@@ -197,6 +197,13 @@ using namespace functionexpression;
 AbstractExpression*
 ExpressionUtil::functionFactory(int functionId, const std::vector<AbstractExpression*>* arguments) {
     if (IS_USER_DEFINED_ID(functionId)) {
+        // This part is temporary to avoid the memory leak.
+        // It will be removed after the UDF code is checked in.
+        size_t i = arguments->size();
+        while (i--) {
```
wolffcm: I had to think about this.  You are just deleting all the arguments, starting with the last one.  Is the order of deletion important?  Why not just use a for loop or BOOST_FOREACH, which would be more idiomatic and readable?

PR #4669, opened 2017-07-05
Eng 12696 degree radians

diwakarj:
Implement Numeric Functions Degree and Radian.


src/ee/expressions/numericfunctions.h
```diff
@@ -18,6 +18,10 @@
 #include "common/NValue.hpp"
 #include "boost/math/constants/constants.hpp"
 
+#ifndef M_PI
+#define M_PI 3.1415926535
+#endif
```
BillWhite: 1. If you add -D_USE_MATH_DEFINES to the compilation line I think you get this already.
1. The value to a few more decimal places seems to be 3.14159265358979323846.  I think you want to round this up the 3.1415926536.

This is a tiny gripe.

src/ee/expressions/numericfunctions.h
```diff
@@ -292,6 +296,32 @@ template<> inline NValue NValue::callUnary<FUNC_SEC>() const {
     return retval;
 }
 
+/** implement the SQL SEC function for all numeric values */
+template<> inline NValue NValue::callUnary<FUNC_DEGREES>() const {
```
BillWhite: Is this comment right?

src/ee/expressions/numericfunctions.h
```diff
@@ -292,6 +296,32 @@ template<> inline NValue NValue::callUnary<FUNC_SEC>() const {
...
+    NValue retval(VALUE_TYPE_DOUBLE);
+    double inputValue = castAsDoubleAndGetValue();
+    double resultDouble = inputValue*(180.0 / M_PI);
+    throwDataExceptionIfInfiniteOrNaN(resultDouble, "function DEGREES");
+    retval.getDouble() = resultDouble;
+    return retval;
+}
+
+/** implement the SQL SEC function for all numeric values */
+template<> inline NValue NValue::callUnary<FUNC_RADIANS>() const {
```
BillWhite: Is this comment right?

src/ee/expressions/numericfunctions.h
diwakarj: Changes PI define value.  will -D_USE_MATH_DEFINES be a CPP flag in build.py?

src/ee/expressions/numericfunctions.h
diwakarj: updated comment.

src/ee/expressions/numericfunctions.h
diwakarj: updated comment

src/ee/expressions/numericfunctions.h
```diff
@@ -18,6 +18,10 @@
 #include "common/NValue.hpp"
 #include "boost/math/constants/constants.hpp"
 
+#ifndef M_PI
+#define M_PI 3.1415926536
```
johnTcrawford: Is this enough digits for a double value? Does C++ not have a built-in value or function for Pi?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
```diff
@@ -541,7 +541,13 @@ private FunctionCustom(int id) {
             case FUNC_ATAN :
             case FUNC_ATAN2 :
             case FUNC_DEGREES :
+                name = Tokens.T_DEGREES;
+                parseList = singleParamList;
+                break;
             case FUNC_RADIANS :
+                name = Tokens.T_RADIANS;
+                parseList = singleParamList;
+                break;
```
johnTcrawford: I am unfamiliar with this code, but it appears to interfere with the previous fall-through behavior of the functions above. I don't know whether that matters or not.

tests/frontend/org/voltdb/fullddlfeatures/fullDDL.sql
```diff
@@ -1135,6 +1136,7 @@ CREATE INDEX ENG_8168_INDEX_USES_REGEXP_POSITION ON T55 (REGEXP_POSITION(str, st
 CREATE INDEX ENG_8168_INDEX_USES_REGEXP_POSITION_C ON T55 (REGEXP_POSITION(str, str, 'c'));
 CREATE INDEX ENG_8168_INDEX_USES_REGEXP_POSITION_I ON T55 (REGEXP_POSITION(str, str, 'i'));
 CREATE INDEX ENG_8168_INDEX_USES_QUARTER ON T55 (QUARTER(TS));
+CREATE INDEX ENG_8168_INDEX_USES_RADIANS ON T55 (RADIANS(BIG));
```
johnTcrawford: BIG seems like an odd choice here (and for DEGREES, above), why not use one of the FLOAT or DECIMAL columns from T22 (as Bill did for his "INET..." functions).

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionCustom.java
BillWhite: You are right.  This will not work.  If someone calls the inverse trig functions they will actually call the DEGREES function.

You just need to move this up.  Good catch, John.

diwakarj: moved function up to avoid breaking the flow.

tests/frontend/org/voltdb/fullddlfeatures/fullDDL.sql
diwakarj: changed index to decimal. i think both float and decimal can cover the range. let me know if needs float only.

src/ee/expressions/numericfunctions.h
diwakarj: Using CPP flag -D_USE_MATH_DEFINES in compilation instead of #define.

tests/frontend/org/voltdb/planner/TestFunctions.java
```diff
@@ -122,6 +122,30 @@ public void testBitwise() {
...
+                "Function 'roundmagic' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  SOUNDEX(VARCHAR_TYPE) FROM NUMERICFUN",
+                "Function 'soundex' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  ASCII(VARCHAR_TYPE) FROM NUMERICFUN",
+                "Function 'ascii' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  RAWTOHEX(VARBINARY_TYPE) FROM NUMERICFUN",
+                "Function 'rawtohex' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  HEXTORAW(VARCHAR_TYPE) FROM NUMERICFUN",
+                "Function 'hextoraw' is not supported in VoltDB: Custom Function");
+        //TODO : add (ATAN2)
```
BillWhite: Why not just add atan2?  It takes two float arguments, I think.  For us float is just the same as double.

This test should be exactly like the previous tests.

diwakarj: I wanted to check its (atan2) error message with you but forgot in the end, my bad. I don't get the expected error text as not supported for Atan2, rather its object not found. please see the below query with two float. I may be wrong in using atan2, could you please check it once. 

voltdb diwakar$ sqlcmd
SQL Command :: localhost:21212
1> create table t(a float, b float);
Command succeeded.
2> insert into t values(5,3);
(Returned 1 rows in 0.06s)
3> select atan2(a,b) from t;
Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: Error compiling query: org.voltdb.planner.PlanningErrorException: Error in "select atan2(a,b) from t" user lacks privilege or object not found: ATAN2

tests/frontend/org/voltdb/planner/TestFunctions.java
```diff
@@ -122,6 +122,31 @@ public void testBitwise() {
...
+        failToCompile("SELECT  ASCII(VARCHAR_TYPE) FROM NUMERICFUN",
+                "Function 'ascii' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  RAWTOHEX(VARBINARY_TYPE) FROM NUMERICFUN",
+                "Function 'rawtohex' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  HEXTORAW(VARCHAR_TYPE) FROM NUMERICFUN",
+                "Function 'hextoraw' is not supported in VoltDB: Custom Function");
+        failToCompile("SELECT  ATAN2(FLOAT_TYPE,FLOAT_TYPE) FROM NUMERICFUN",
+                "user lacks privilege or object not found: ATAN2");
+
+    }
```
BillWhite: ATAN2 seems to be partially defined.  It would be interesting to puzzle out why it's not actually recognized by HSQL.  But I think we've beaten this horse enough.

PR #4670, opened 2017-07-06
ENG-12710: put the exception message into the compiler exception, the…

xinjiacs:
…n the error can be seen on server logs


luyangco:
Looks good to me.

PR #4671, opened 2017-07-06
ENG-12756: Change offset fetching log to WARN as we retry.

akhanzode:


nshi:
In what cases will it get this error, when losing connection to Kafka?

akhanzode:
This could happen kafka is not ready with offset information as its kept in topic or zookeeper or issues with kafka to fetch any offset information because of network etc. If we dont get offset information we continue to get it with backoff sleep.

PR #4672, opened 2017-07-06
Eng 12391 resubmit

brians127:
Resubmit of work that allows "init --classes" without schema.

LocalCluster handles CLASSPATH differently from the previous submission. Test procs are always removed from the CLASSPATH - there is no exemption for "init with schema" tests.


tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -320,10 +321,12 @@ public LocalCluster(String jarFileName,
             boolean debug,
             boolean isRejoinTest,
             Map<String, String> env) {
-        this(null, jarFileName, siteCount, hostCount, kfactor, clusterId, target, failureState, debug, isRejoinTest, env);
+        // NOTE: isRejoinTest is unused
+        this(null, null, jarFileName, siteCount, hostCount, kfactor, clusterId, target, failureState, debug, env);
```
akhanzode: Why remove unused at only one level below if its not used at all remove all together.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -376,14 +367,15 @@ public LocalCluster(String schemaToStage,
...
-            if (schemaToStage == null) {
+            if (m_usesStagedSchema == false) {
                 log.info("Instantiating empty LocalCluster with class.method: " +
                         m_callingClassName + "." + m_callingMethodName);
             } else {
                 log.info("Instantiating LocalCluster with schema and class.method: " +
                         m_callingClassName + "." + m_callingMethodName);
             }
         } else {
+            Preconditions.checkArgument(m_usesStagedSchema == false, "Cannot use OldCLI catalog with staged schema and/or classes");
```
akhanzode: Why not use existing error handling mechanism? Adding more ways to fail will complicate how LocalCluster fails.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -412,6 +404,18 @@ public LocalCluster(String schemaToStage,
         m_failureState = m_kfactor < 1 ? FailureState.ALL_RUNNING : failureState;
         m_pipes = new ArrayList<>();
         m_cmdLines = new ArrayList<>();
+        if (schemaToStage != null) {
+            try {
+                templateCmdLine.m_userSchema = VoltProjectBuilder.createFileForSchema(schemaToStage);
```
akhanzode: I am guessing this should be only valid for 'init' and should get reset for 'start'. I don't see code where these gets reset.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
brians127: Fully removing this changes 20+ unrelated tests. Normally you don't want me to do that.

PR #4673, opened 2017-07-06
Eng 10782 split junit other

jwiebema1:
this build's off of work done previously to split up the tests to run in docker.  The former junit_other is now available to run in 4 parts, junit_other_p1 ... other_p4 . each part takes about 35 minutes to run.  the tests can also be run as one large group as junit_other.


xinjiacs:
Make sure your total Junit test run count equals to junit_other.

PR #4678, opened 2017-07-10
ENG-9932: Add ccrTime as generation.

akhanzode:
For review only.
- Use a new unique id when replaying UAC to generate generation higher than current generation.

### DIFF
--- a/src/frontend/org/voltdb/CatalogContext.java
+++ b/src/frontend/org/voltdb/CatalogContext.java
@@ -45,6 +45,7 @@ import org.voltdb.settings.NodeSettings;
 import org.voltdb.utils.CatalogUtil;
 import org.voltdb.utils.InMemoryJarfile;
 import org.voltdb.utils.VoltFile;
+import static org.voltdb.compiler.CatalogChangeResult.CATALOG_CHANGE_NOREPLAY;
 
 public class CatalogContext {
     private static final VoltLogger hostLog = new VoltLogger("HOST");
@@ -135,7 +136,7 @@ public class CatalogContext {
         m_transactionId = transactionId;
         m_uniqueId = uniqueId;
         //This is only set to something other than m_uniqueId when we are replaying a UAC.
-        m_ccrTime = ((ccrTime == -1L) ? uniqueId : ccrTime);
+        m_ccrTime = ((ccrTime == CATALOG_CHANGE_NOREPLAY) ? uniqueId : ccrTime);
         // check the heck out of the given params in this immutable class
         if (catalog == null) {
             throw new IllegalArgumentException("Can't create CatalogContext with null catalog.");
--- a/src/frontend/org/voltdb/CatalogContext.java
+++ b/src/frontend/org/voltdb/CatalogContext.java
@@ -98,6 +98,7 @@ public class CatalogContext {
     // database settings. contains both cluster and path settings
     private final DbSettings m_dbSettings;
 
+    public final long m_ccrTime;
     /**
      * Constructor especially used during @CatalogContext update when @param hasSchemaChange is false.
      * When @param hasSchemaChange is true, @param defaultProcManager and @param plannerTool will be created as new.
@@ -127,10 +128,11 @@ public class CatalogContext {
             HostMessenger messenger,
             boolean hasSchemaChange,
             DefaultProcedureManager defaultProcManager,
-            PlannerTool plannerTool)
+            PlannerTool plannerTool, long ccrTime)
     {
         m_transactionId = transactionId;
         m_uniqueId = uniqueId;
+        m_ccrTime = (ccrTime == -1L ? m_uniqueId : ccrTime);
         // check the heck out of the given params in this immutable class
         if (catalog == null) {
             throw new IllegalArgumentException("Can't create CatalogContext with null catalog.");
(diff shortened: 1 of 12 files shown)


src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -231,7 +233,7 @@ public CatalogContext(
             HostMessenger messenger)
     {
         this(transactionId, uniqueId, catalog, settings, catalogBytes, catalogBytesHash, deploymentBytes,
-                version, messenger, true, null, null);
+                version, messenger, true, null, null, -1L);
```
nshi: Why is this -1?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3411,7 +3412,7 @@ else if (m_catalogContext.catalogVersion != expectedCatalogVersion) {
 
 
                 // 1. update the export manager.
-                ExportManager.instance().updateCatalog(m_catalogContext, requireCatalogDiffCmdsApplyToEE, requiresNewExportGeneration, partitions);
+                ExportManager.instance().updateCatalog(m_catalogContext, ccrTime, requireCatalogDiffCmdsApplyToEE, requiresNewExportGeneration, partitions);
```
nshi: If it's already part of the catalog context, do you still need to pass it in as a separate parameter?

src/frontend/org/voltdb/export/ExportManager.java
```diff
@@ -623,23 +623,23 @@ public int getConnCount() {
         return m_connCount;
     }
 
-    public synchronized void updateCatalog(CatalogContext catalogContext, boolean requireCatalogDiffCmdsApplyToEE, boolean requiresNewExportGeneration, List<Integer> partitions)
+    public synchronized void updateCatalog(CatalogContext catalogContext, long time, boolean requireCatalogDiffCmdsApplyToEE, boolean requiresNewExportGeneration, List<Integer> partitions)
```
nshi: I don't think the new parameter `time` is used in the method.

akhanzode:
The timestamp is generated by cl replay. The change is in pro. I will do cleanup.

nshi:
So UAC during replay uses different number than UAC during normal operation? The uniqueIDs are generated using Volt epoch-based timestamps, which is much smaller than regular unix timestamps. It's error prone to mix numbers with completely different compositions for the same purpose, especially when they are used for ordering.

akhanzode:
It generates a new unique id and not a timestamp so composition should be same.

src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -132,7 +134,8 @@ public CatalogContext(
     {
         m_transactionId = transactionId;
         m_uniqueId = uniqueId;
-        m_ccrTime = (ccrTime == -1L ? m_uniqueId : ccrTime);
+        //This is only set to something other than m_uniqueId when we are replaying a UAC.
+        m_ccrTime = ((ccrTime == -1L) ? uniqueId : ccrTime);
```
nshi: The -1 check is unnecessary here since you always pass a real value in.

akhanzode: @Promote calls with -1 otherwise I will have to push -1 check in UpdateCore where it gets the param.

src/frontend/org/voltdb/compiler/CatalogChangeResult.java
```diff
@@ -52,4 +52,5 @@ public PrepareDiffFailureException(byte statusCode, String msg) {
     public int expectedCatalogVersion = -1;
     // This is set to true if schema change involves stream or connector changes or a view on stream is created or dropped.
     public boolean requiresNewExportGeneration;
+    public long m_ccrTime = -1L;
```
xinjiacs: Can you used a public static final long variable to assign -1L to "m_ccrTime" so that reader can easily find references with this default value in our code base?

Also add a comment here.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -188,6 +188,6 @@ else if (ddlToken.equals("create")) {
                              ccr.deploymentHash,
                              ccr.requireCatalogDiffCmdsApplyToEE ? 1 : 0,
                              ccr.hasSchemaChange ?  1 : 0,
-                             ccr.requiresNewExportGeneration ? 1 : 0);
+                             ccr.requiresNewExportGeneration ? 1 : 0, ccr.m_ccrTime);
```
xinjiacs: Code format: use a new line for the extra parameter? 

Same apply to other similar procedure.

src/frontend/org/voltdb/sysprocs/UpdateCore.java
```diff
@@ -375,7 +376,7 @@ else if (fragmentId == SysProcFragmentId.PF_updateCatalog) {
                         catalogStuff.getDeploymentHash(),
                         requireCatalogDiffCmdsApplyToEE,
                         hasSchemaChange,
-                        requiresNewExportGeneration);
+                        requiresNewExportGeneration, ccrTime);
```
xinjiacs: @akhanzode I have a "crazy" thought, why you need to add "ccrTime" to make it so complex?

In the invocation of " VoltDB.instance().catalogUpdate(", it can pass in different unique id based on whether it's reply or not.  Will the problem be solved in this way?

Am I missing something else? @nshi

PR #4680, opened 2017-07-10
ENG-12034 fix only

brians127:
This addresses a multiple node kill scenario where the Zookeeper data went stale midway through sending the assignments.

When the leader is distributing a new set of importer channel assignments, it will retry sending ALL assignments if ANY fail. This is fine if the first assignment fails, but in this scenario there is a race condition between the distribution of channel assignments and a mesh monitor responding to the node failure. In the original problem report, the leader has sent roughly 2/3 of the assignments out before Zookeeper throws a "bad version" error. When the entire re-send gets retried, the very first assignment gets flagged as a duplicate resulting in the reported error. The retry is aborted, meaning the remaining channel assignments don't get distributed.

Anyone who sees this in production should first be sure the cluster didn't initiate a global crash (if it did, all nodes will be gone soon and there are bigger issues). Then pause the cluster, verify the pause has completed (due to an unrelated issue), and then resume the cluster. This will restart the importers, refreshing their channel assignments. The data which was unable to import will remain in Kafka - data won't be loss unless Kafka deletes it.

This will be system tested using existing endurance tests, since they originally found the issue and attempts to reproduce in a JUnit style environment have been unsuccessful.

### DIFF
--- a/src/frontend/org/voltdb/importer/ChannelDistributer.java
+++ b/src/frontend/org/voltdb/importer/ChannelDistributer.java
@@ -642,10 +642,6 @@ public class ChannelDistributer implements ChannelChangeCallback {
 
         final int seed;
 
-        AssignChannels(final int seed) {
-            this.seed = seed;
-        }
-
         AssignChannels() {
             seed = System.identityHashCode(this);
         }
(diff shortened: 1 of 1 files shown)


nshi:
I really like your description of the problem followed by the workaround.

If I understand correctly, your solution is to not rerun `AssignChannels` immediately, which may not know what's already there, but to let `GetChannels` kick it off with the ones that are missing. Is that correct?

brians127:
@nshi Yes, that's exactly it.

PR #4681, opened 2017-07-10
ENG-12786:

manjujames:
Generate transactions and events into drstream only after receiving DR_STREAM_START event.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -563,10 +563,10 @@ void VoltDBEngine::releaseUndoToken(int64_t undoToken) {
     }
     m_undoLog.release(undoToken);
 
-    if (m_executorContext->drStream() && m_executorContext->drStream()->drStreamStarted()) {
+    if (m_executorContext->drStream()) {
         m_executorContext->drStream()->endTransaction(m_executorContext->currentUniqueId());
     }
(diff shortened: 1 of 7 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -562,10 +563,10 @@ void VoltDBEngine::releaseUndoToken(int64_t undoToken) {
     }
     m_undoLog.release(undoToken);
 
-    if (m_executorContext->drStream()) {
+    if (m_executorContext->drStream() && m_executorContext->drStream()->drStreamStarted()) {
         m_executorContext->drStream()->endTransaction(m_executorContext->currentUniqueId());
```
wweiss-voltdb: I don't think you need this. If the transaction is never begun (m_opened is false),  endTransaction does nothing.

src/ee/storage/persistenttable.cpp
```diff
@@ -321,6 +321,10 @@ void PersistentTable::deleteAllTuples(bool, bool fallible) {
     }
 }
 
+bool PersistentTable::doDRActions(AbstractDRTupleStream* drStream) {
+    return (drStream && drStream->drStreamStarted() && !m_isMaterialized && m_drEnabled);
```
wweiss-voltdb: This is done on every table update (frequently multiple times in one transaction), so it would be allot more efficient to collapse this into a single boolean, that gets re-evaluated when any of these values change.

PR #4682, opened 2017-07-10
ENG-12070 don't throw IllegalStateException when importers are shut down

brians127:
If an importer has not fully initialized, a callback will not be registered for it. There are a few scenarios where this can happen - see ImportManager.readyForData for them. This results in a benign error being thrown on importer shutdown, such as when the cluster is paused.

In the benign scenario, log a NOTICE and do nothing. An exception will still get thrown if the importer assignments are ever non-empty.

Unit test added to catch the issue reliably, since the issue is easiest to reproduce if Kafka doesn't actually exist. System tests see this once in awhile as well. They don't explicitly try to shut down clusters with importers before data gets processed, but due to the stressful conditions of those tests it happens anyway.

Minor related changes in Pro: https://github.com/VoltDB/pro/pull/1957

### DIFF
--- a/src/frontend/org/voltdb/importer/ChannelDistributer.java
+++ b/src/frontend/org/voltdb/importer/ChannelDistributer.java
@@ -29,18 +29,8 @@ import static org.voltcore.zk.ZKUtil.joinZKPath;
 import java.io.File;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Deque;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.NavigableMap;
-import java.util.NavigableSet;
-import java.util.Random;
-import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Semaphore;
@@ -352,12 +342,14 @@ public class ChannelDistributer implements ChannelChangeCallback {
                 !FluentIterable.from(uris).anyMatch(isNull()),
                 "uris set %s contains null elements", uris
                 );
-        if (!uris.isEmpty()) {
-            // during shutdown, registerChannels(empty) can get called for importers that were not fully initialized.
-            Preconditions.checkState(
-                registered.contains(importer),
-                "no callbacks registered for %s", importer
-                );
+        if (!registered.contains(importer)) {
+            if (uris.isEmpty()) {
+                LOG.info("Skipping channel un-registration for " + importer + " since it did not finish initialization");
+                return;
+            } else {
+                throw new IllegalStateException("no callbacks registered for " + importer
+                        + " - unable to register channels " + Arrays.toString(uris.toArray()));
+            }
         }
 
         Predicate<ChannelSpec> forImporter = ChannelSpec.importerIs(importer);
--- a/src/frontend/org/voltdb/importer/ChannelDistributer.java
+++ b/src/frontend/org/voltdb/importer/ChannelDistributer.java
@@ -344,6 +344,9 @@ public class ChannelDistributer implements ChannelChangeCallback {
                 );
         if (!registered.contains(importer)) {
             if (uris.isEmpty()) {
+                // ImporterLifeCycleManager.stop() calls registerChannels() is called with an empty set of URIs.
+                // If the importer never finished starting, we hit this condition.
+                // This log message is used by the TestImporterStopAfterIncompleteStart JUnit.
                 LOG.info("Skipping channel un-registration for " + importer + " since it did not finish initialization");
                 return;
             } else {
--- a/src/frontend/org/voltdb/importer/ChannelDistributer.java
+++ b/src/frontend/org/voltdb/importer/ChannelDistributer.java
@@ -352,10 +352,13 @@ public class ChannelDistributer implements ChannelChangeCallback {
                 !FluentIterable.from(uris).anyMatch(isNull()),
                 "uris set %s contains null elements", uris
                 );
-        Preconditions.checkState(
+        if (!uris.isEmpty()) {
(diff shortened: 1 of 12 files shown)


src/frontend/org/voltdb/importer/ChannelDistributer.java
```diff
@@ -352,10 +352,13 @@ public void registerChannels(String importer, Set<URI> uris) {
                 !FluentIterable.from(uris).anyMatch(isNull()),
                 "uris set %s contains null elements", uris
                 );
-        Preconditions.checkState(
+        if (!uris.isEmpty()) {
```
akhanzode: Why not add a log if this happens and return and dont do precondition check.

brians127:
@vittvolt You should primarily look at my changes to LocalCluster - they are small but affect your log scraper API. I made a few small usability improvements.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
```diff
@@ -1681,7 +1676,7 @@ synchronized public void shutDown() throws InterruptedException {
         }
         shutDownExternal();
 
-        if (m_logMessageMatchPatterns != null) {
+        if (m_logMessageMatchPatterns != null && clearLogMatches) {
```
vittvolt: Is cleaning up after shutdown necessary ? I mean the log search results will not be useful after shutdown anyway, or the user could choose to reset them by themselves.

vittvolt:
@brians127 That looks good ! I made a comment over there for the cleanup of log search results after 
cluster shutdown.

tests/frontend/org/voltdb/regressionsuites/LocalCluster.java
brians127: I don't think it is. Where this matters is if someone tries to reuse a LocalCluster instance - for example, to test recovery. I think the right behavior is to have startUp() explicitly clears the logs if and only if clearLocalDataDirectories is true, and rely on the user the rest of the time. Since my test does not care what startUp() does with the logs, we should probably change that later if it's needed.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1358,8 +1359,15 @@ private static String buildBundleURL(String bundle, boolean alwaysBundle) {
                 String bundlelocation = System.getProperty(VOLTDB_BUNDLE_LOCATION_PROPERTY_NAME);
                 if (bundlelocation == null || bundlelocation.trim().length() == 0) {
                     String rpath = CatalogUtil.class.getProtectionDomain().getCodeSource().getLocation().toURI().getPath();
-                    hostLog.info("Module base is: " + rpath + "/../bundles/");
-                    String bpath = (new File(rpath)).getParent() + "/../bundles/" + bundleUrl;
+                    if (CoreUtils.isJunitTest()) {
```
jeffreycrump: Ugh, this is gross. Can the unit test just set the property accordingly so you don't have to couple the product code to the tests?

src/frontend/org/voltdb/importer/ChannelDistributer.java
```diff
@@ -352,10 +342,15 @@ public void registerChannels(String importer, Set<URI> uris) {
                 !FluentIterable.from(uris).anyMatch(isNull()),
                 "uris set %s contains null elements", uris
                 );
-        Preconditions.checkState(
-                registered.contains(importer),
-                "no callbacks registered for %s", importer
-                );
+        if (!registered.contains(importer)) {
+            if (uris.isEmpty()) {
+                LOG.info("Skipping channel un-registration for " + importer + " since it did not finish initialization");
```
hhakimi53: The log message is little confusing - "skipping unregistration of channel ... " in register channel function. Can u pls add comment about it if the message is needed in here. thnxs

src/frontend/org/voltdb/utils/CatalogUtil.java
hhakimi53: Yes, that will be cleaner approach if can use VOLTDB_BUNDLE_LOCATION_PROPERTY_NAME to point to correct bundle location.

brians127: I don't think that forcing all test developers to remember that the environment needs to be set is a good approach. This has cost me a lot of time so far and will continue to hurt new developers. 

A better approach would be to stop relying on the current working directory Volt is running in or to make it the same both in production and in JUnit. I took the "isJunitTest" approach instead because it doesn't risk breaking production behavior.

tests/frontend/org/voltdb/regressionsuites/TestImporterStopAfterIncompleteStart.java
```diff
@@ -0,0 +1,110 @@
...
+        project.addImport(true, "kafka", "csv", "kafkastream.jar", props);
+
+        m_cluster.compileDeploymentOnly(project);
+        new File(project.getPathToDeployment()).deleteOnExit();
+        m_cluster.setExpectedToCrash(false);
+        m_cluster.startUp();
+    }
+
+    @Test
+    public void test() throws Exception {
```
akhanzode: Name the test not just test()

tests/frontend/org/voltdb/regressionsuites/TestImporterStopAfterIncompleteStart.java
```diff
@@ -0,0 +1,110 @@
...
+
+/** Tests a specific scenario where the importer never finishes starting up before the cluster is paused and shut down.
+ * An error was being thrown to the log for a benign condition.
+ * This scenario isn't Kafka specific, but is easiest to reproduce with a Kafka importer but no Kafka.
+ */
+public class TestImporterStopAfterIncompleteStart {
+    private LocalCluster m_cluster = null;
+
+    // This is how we know the appropriate code path was executed
+    private static final String CHANNEL_UNREGISTRATION_PATTERN = "Skipping channel un-registration for KafkaStreamImporter";
```
akhanzode: You should not search with importer name in it. just skipping channel should do.

brians127:
Debug build looks fine.

PR #4684, opened 2017-07-11
Eng 12769 queue depth with log

dhan111:
If a stored procedure takes a long, long time in Java code, there was nothing in the logfile to indicate that a procedure is doing too much Java. There are only warnings for long-running MP (possible mp deadlock) or long SQL execution times.

This is a watch on the sites to show that there is a long-running txn. A warning is generated if no task was polled from the partition queue in the last {X} seconds, where {X} is configurable from deployment file with default value 10 seconds.

### DIFF
--- a/src/frontend/org/voltdb/iv2/TickProducer.java
+++ b/src/frontend/org/voltdb/iv2/TickProducer.java
@@ -32,7 +32,7 @@ import org.voltdb.rejoin.TaskLog;
 public class TickProducer extends SiteTasker implements Runnable
 {
     private final SiteTaskerQueue m_taskQueue;
-    private final long m_taskTimeThreshold;
+    private final long m_procedureLogThreshold;
     private final long SUPPRESS_INTERVAL = 60; // 60 seconds
     private VoltLogger m_logger;
     private int m_partitionId;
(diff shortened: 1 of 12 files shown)


manjujames:
The changes look good to me. But, we also want to change the 5 seconds default to 10 seconds (similar to query timeout) and make it configurable through deployment file. We should include those changes before we merge.

src/frontend/org/voltdb/iv2/TickProducer.java
```diff
@@ -32,18 +32,27 @@
...
     {
         m_taskQueue = taskQueue;
         m_logger = new VoltLogger("HOST");
         m_partitionId = taskQueue.getPartitionId();
+        // get warning threshold from deployment
+        // convert to nano seconds (default 10s)
+        m_taskTimeThreshold = 1_000_000L * VoltDB.instance()
+                                .getCatalogContext()
+                                .getDeployment()
+                                .getSystemsettings()
```
manjujames: You will need null checks here. SystemSettings or Procedure can be null since minOccurs is 0 for those. If those are null, we need to use the default value of 10s.

manjujames: Oh never mind. I see that you are populating the default deployment.

PR #4686, opened 2017-07-12
ENG-12742 micro opt refactor

yzhang1991:


src/frontend/org/voltdb/expressions/AbstractSubqueryExpression.java
```diff
@@ -89,7 +89,7 @@ public void resetSubqueryNodeId() {
     // Create a matching PVE for this expression to be used on the EE side
     // to get the original expression value
     protected void addCorrelationParameterValueExpression(AbstractExpression expr, List<AbstractExpression> pves) {
-        int paramIdx = AbstractParsedStmt.NEXT_PARAMETER_ID++;
+        int paramIdx = AbstractParsedStmt.NEXT_PARAMETER_ID.getAndIncrement();
```
wolffcm: Hmm... parameter IDs should all be adjacent for the same statement, right?  If there were concurrent planners that might not be the case and havoc could result.

I think `NEXT_PARAMETER_ID` should be a normal (non-static) member of some object that is not shared by more than one thread.  `QueryPlanner` seems to fit this description?

wolffcm: Seems like this applies to all of the static variables that you've made Atomic.

src/frontend/org/voltdb/planner/microoptimizations/ReplaceWithIndexLimit.java
```diff
@@ -58,7 +59,7 @@ protected void recursivelyPrint(AbstractPlanNode node, StringBuilder sb)
     }
```
wolffcm: This class has a member variable called `indent` which will incur a data race in `recursivelyPrint`.  This is easily fixable though by just making `indent` a parameter to `recursivelyPrint`.  I realize it's debug-only code.

src/frontend/org/voltdb/expressions/AbstractSubqueryExpression.java
yzhang1991: I don't think the parameter ID has to be adjacent to work, but it makes a lot of sense to see them be adjacent for each statement.

PR #4687, opened 2017-07-12
Eng 12676 Add inline insert.

BillWhite:
Add the ability to process inline insert nodes.

An insert node and executor can be made inline if it's source executor is a scan, indexed or sequential.
This saves one temp table.

This was pulled from 7.4 because of some bugs.  We needed a small redesign and to repair some bad code.

Note that most of the changed files are plannertester plans which have changed.  The changes are due to removed unnecessary projection nodes.  The plannertester has no insert statements, so there are no plans with insert nodes being made inline.  This seems like a testing failure, though there are more precise planner tests.


tests/ee/test_utils/plan_testing_baseclass.h
```diff
@@ -234,18 +234,14 @@ class PlanTestingBaseClass : public Test {
                          int                             num_strings) {
```
BillWhite: Changes in the file are to cause exceptions to be thrown rather than silly error messages.

BillWhite:
Closing.  This will be back soon.

src/frontend/org/voltdb/plannodes/AbstractPlanNode.java
```diff
@@ -467,6 +468,76 @@ public NodeSchema getOutputSchema()
...
+            if (childProj != null) {
+                AbstractPlanNode inlineInsertNode = childProj.getInlinePlanNode(PlanNodeType.INSERT);
+                if (inlineInsertNode != null) {
+                    child = inlineInsertNode;
+                } else {
+                    child = childProj;
+                }
+                childSchema = child.getOutputSchema();
+            } else if (child.getChildCount() > 0) {
+                child = child.getChild(0);
```
wolffcm: Joins can have more than one child.  I guess in that case, the schema will be pre-defined and m_hasSignificantOutputSchema will be true?  If that's the case, then this must be called after generateOutputSchema has been completed?

wolffcm: If it only makes sense to call this method at certain points in the planning flow, would be good to note that in the header comment.

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -430,12 +444,28 @@ static protected void assertClassesMatchNodeChain(
     protected static AbstractPlanNode followAssertedLeftChain(
             AbstractPlanNode start,
             PlanNodeType startType, PlanNodeType... nodeTypes) {
+        return followAssertedLeftChain(start, false, startType, nodeTypes);
+    }
+
+    protected static AbstractPlanNode followAssertedLeftChain(
+            AbstractPlanNode start,
+            boolean hasOptionalProjection,
```
wolffcm: I said this last time.  I'm not a big fan of the "optional projection."  If you're making plans better by eliminating project nodes... can't someone come in later and re-introduce them?  And the way the tests are now, that won't be flagged as a regressions?

Maybe there's other alternatives here.

wolffcm:
Also, these changes are a big improvement.  Thanks especially for making the micro-optimizer improvements.

PR #4688, opened 2017-07-13
ENG-12528-Fix-voltTupleWrapper-bug

sycai:


brians127:
Can you put a description of what you are fixing in the pull request?
I am not the right person to review these. You should find someone who's more familiar with Python and/or our Python API.

PR #4690, opened 2017-07-13
ENG-12804: Add null check for stream table insert.

akhanzode:


nshi:
I agree with @wolffcm. An EE test is best suited for these types of changes.

src/ee/storage/streamedtable.cpp
```diff
@@ -134,8 +152,24 @@ void StreamedTable::nextFreeTuple(TableTuple *) {
                                   "May not use nextFreeTuple with streamed tables.");
 }
 
+bool StreamedTable::checkNulls(TableTuple& tuple) const {
```
nshi: This looks like something that could go into the base `Table` class. It's exactly the same as the one in `PersistentTable`.

PR #4694, opened 2017-07-14
Allow the use of sub-queries within DML (SQL) statements,

johnTcrawford:
i.e., within INSERT, UPSERT, UPDATE, and DELETE statements, including the possibility of "correlated" sub-queries, referencing a table (or view or alias) used in the outer SQL statement; and other improvements.

### DIFF
--- a/tests/sqlgrammar/sql-grammar.txt
+++ b/tests/sqlgrammar/sql-grammar.txt
@@ -21,13 +21,13 @@
 #      must be preceded and followed by a space (otherwise the || concatenation
 #      operator would be impossible without some sort of escape symbol, which
 #      I did not want to get into).
-#   4. Adding an integer before a | symbol makes the preceding option more
-#      likely; the "n|" still needs to be preceded and followed by a space.
+#   4. Adding a (positive) integer before a | symbol makes the preceding option
+#      more likely; the "n|" still needs to be preceded and followed by a space.
 #      For example, in this definition:
-#          {definition} :== foo 2| bar
+#          definition :== foo 2| bar
 #      the "foo" option is twice as likely as the "bar" option. You may use
 #      more than one such integer; for example, in this definition:
-#          {definition} :== foo 3| bar 2| null
+#          definition :== foo 3| bar 2| null
 #      the "foo" option will be chosen half (3/6) the time, the "bar" option
 #      will be chosen a third (2/6) of the time, and the "null" option will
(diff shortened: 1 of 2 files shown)


tests/sqlgrammar/sql-grammar.txt
```diff
@@ -131,44 +140,126 @@ in-or-up-sert-select    ::= {insert-select-multiple} 8| {insert-select-all}    2
...
+                            SERT INTO {table-name} ({timestamp-column-name}, ID)   SELECT {timestamp-col-or-subq}, {int-column-name} FROM {table-reference}
+insert-select-id-varbin ::= SERT INTO {table-name} (ID, {varbinary-column-name})   SELECT {int-column-name}, {varbinary-col-or-subq} FROM {table-reference} | \
+                            SERT INTO {table-name} ({varbinary-column-name}, ID)   SELECT {varbinary-col-or-subq}, {int-column-name} FROM {table-reference}
+insert-select-id-point  ::= SERT INTO {table-name} (ID, {point-column-name})       SELECT {int-column-name}, {point-col-or-subq}     FROM {table-reference} | \
+                            SERT INTO {table-name} ({point-column-name}, ID)       SELECT {point-col-or-subq}, {int-column-name}     FROM {table-reference}
+insert-select-id-poly   ::= SERT INTO {table-name} (ID, {polygon-column-name})     SELECT {int-column-name}, {polygon-col-or-subq}   FROM {table-reference} | \
+                            SERT INTO {table-name} ({polygon-column-name}, ID)     SELECT {polygon-col-or-subq}, {int-column-name}   FROM {table-reference}
+
+numeric-col-or-subq     ::= {numeric-column-name}   9| {numeric-scalar-sub-q}
```
wolffcm: The `9|` part means that one option is 9 times more likely than the other, right?  But which one is it?  I forget.

johnTcrawford: The one to the left (in this case, {numeric-column-name}) is 9 times more likely than the one to the right (in this case, {numeric-scalar-sub-q}). So, using a column-name is the the "normal" thing here; sub-queries are unusual (though still reasonably common).

johnTcrawford:
There are still more invalid SQL statements than I would like, but I have not made that significantly worse, and actually improved it for some types of SQL statements. (I am giving some thought to how I might improve that, but that is for some other ticket.)

PR #4573, opened 2017-05-31
ENG-12528

sycai:
File exporter delimiters now support character '\n'.

PR #4698, opened 2017-07-17
Eng 12528 file export delimiter patch

sycai:
Character Ctrl-A can be used as a file export delimiter

### DIFF
--- a/src/frontend/org/voltdb/utils/CatalogUtil.java
+++ b/src/frontend/org/voltdb/utils/CatalogUtil.java
@@ -1259,7 +1259,7 @@ public abstract class CatalogUtil {
                     } else if (key.toLowerCase().contains("delim")){
                         // Don't trim \n in delimiters
                         String trimmedDelimiters = value.replaceAll("^(\r|\f|\t| )+", "").replaceAll("(\r|\f|\t| )+$", "");
-                        processorProperties.setProperty(key, StringEscapeUtils.escapeJava(trimmedDelimiters));
+                        processorProperties.setProperty(key, trimmedDelimiters);
                     } else {
                         processorProperties.setProperty(key, value.trim());
                     }
--- a/src/frontend/org/voltdb/utils/CatalogUtil.java
+++ b/src/frontend/org/voltdb/utils/CatalogUtil.java
@@ -1256,10 +1256,6 @@ public abstract class CatalogUtil {
                     if (key.toLowerCase().contains("passw")) {
                         // Don't trim password
                         processorProperties.setProperty(key, value);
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -203,7 +203,7 @@
 
     /** Default deployment file contents if path to deployment is null */
     private static final String[] defaultDeploymentXML = {
-        "<?xml version=\"1.0\"?>",
+        "<?xml version=\"1.1\"?>",
```
akhanzode: Does this mean that customers will have to use 1.1 in their deployment?

sycai:
Will add Junit test and useful error message when user put &#01; in a xml file with unsupported version

nshi:
My concern is that this patch only addresses the the `\1` case, other control characters may still have the same problem. As Ben pointed out in the ticket, the parsing of the delimiters property need to take HTML escaping into account.

sycai:
A previous merge has addressed part of the HTML escaping issue: [https://github.com/VoltDB/voltdb/pull/4573](url) . So for characters between \9 and \20, only \r (\13), \f (\12), \t (\9), and space character are not supported. 

On the other hand, do we need full support for characters from \1 to \8?

nshi:
Is there any reason that the individual control characters are handled differently? Would a generic solution help?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -845,15 +847,49 @@ public static DeploymentType parseDeploymentFromString(String deploymentString)
...
             }
+            // This is a makeshift solution for parsing unsupported character '&#01;' out of a XML1.0 file
+            StringBuilder sb = new StringBuilder();
+            BufferedReader br = new BufferedReader(new InputStreamReader(deployIS));
+            String line;
+            boolean inDelimiter = false;
+            // We substitute delimiter '&#01' with another helper string so that JAXB can parse it
+            try {
+                while ((line = br.readLine()) != null) {
+                    if (line.contains("property name=\"delimiters\"")) {
```
nshi: If the XML is formatted slightly differently, such as an extra space in between the words, this will not work. This looks like a hacky way to workaround the problem. I wonder if there is a less error-prone approach to sanitizing the file for parsing.

Is `\1` the only character that has this problem or is `&#01;` also bad?

sycai:
Character <, >, & should be encoded as XML entities. 

Control characters should be encoded as Java literals. For example, user should should put \1 for Ctrl-A character and \n for new line character.

akhanzode:
Changes look good if ctrl-a works merge. changes look generic.

PR #4700, opened 2017-07-17
Eng 11789 mesh arbitor

luyangco:
This patch contains three improvements:
1) inject random delays to simulate a slow network in TestFuzzMeshArbiter.
2) When node enters the final decision stage, don't re-broadcasts final update messages that don't carry with new information (new failures, new survivors, etc.), see description below.
3) dump internal state of mesh arbiter into debug log.

Description of #2
When one host starts to send out its final decision and waits for other hosts to agree with it, it means this host have the enough information to decide which host is dead. So if it receives a final update message but not carry with new information (no new failed node or new survivor), there is no need to interrupt current loop of arbitration and to broadcast the message to other hosts.  Otherwise, other hosts need extra round of information exchange to reach consensus.

This patch addresses the above problem, makes arbitration process shorter and easier to reach an agreement.


src/frontend/org/voltcore/agreement/AgreementSeeker.java
```diff
@@ -537,6 +558,21 @@ public long bestKillCandidateAmong(Set<Long> candidates) {
         return pick;
     }
 
+    public boolean alreadyKnow(FaultMessage fm) {
+        for (Long survivor : fm.survivors) {
+            // Do we already know all the survivors?
+            if (!m_alive.get(survivor).contains(fm.reportingSite)) {
```
xinjiacs: Why check the reporting site other than 
`if (m_alive.get(survivor).size() == 0) return false;`

src/frontend/org/voltcore/agreement/AgreementSeeker.java
```diff
@@ -537,6 +558,21 @@ public long bestKillCandidateAmong(Set<Long> candidates) {
...
+                return false;
+            }
+        }
+        // Do we already know the dead?
+        if (!m_dead.get(fm.failedSite).contains(fm.reportingSite)) {
```
xinjiacs: Similar question: Why check the reporting site other than
`if (m_dead.get(fm.failedSite).size() == 0) return false;`

src/frontend/org/voltcore/agreement/MeshArbiter.java
```diff
@@ -453,9 +459,18 @@ protected boolean notifyOnKill(Set<Long> hsIds, Map<Long, Long> decision) {
                     // In case of concurrent fault, handle it
                     m_mailbox.deliverFront(msg);
                     return false;
-                } else if (mayIgnore(hsIds, fm) == Discard.DoNot) {
-                    m_mailbox.deliverFront(msg);
-                    return false;
+                } else {
```
xinjiacs: You can remove the ELSE block with 
`if (!m_seeker.alreadyKnow(fm) && mayIgnore(hsIds, fm) == Discard.DoNot) {`
    `m_mailbox.deliverFront(msg);
    return false;
}`

tests/frontend/org/voltcore/agreement/TestFuzzMeshArbiter.java
```diff
@@ -149,7 +158,11 @@ public void testNodeFail() throws InterruptedException
 
     public void testLinkFail() throws InterruptedException
     {
-        constructCluster(5);
+        m_rejoinLog.info("testLinkFail");
+        // Fill the array if you want to use specific per-site random seed to reproduce an issue.
+        // 0L means don't use predefined random seed for the site
+        long[] seeds = new long[] {0L, 0L, 0L};
```
luyangco: The purpose of those predefined`seeds` arrays is that when the test failed, you can fill those arrays with seeds printed in the test output to reproduce.

tests/frontend/org/voltcore/agreement/TestFuzzMeshArbiter.java
```diff
@@ -526,15 +543,17 @@ private void dumpNodeState()
...
         final int killSize = 2;
-        long seed = System.currentTimeMillis();
-        System.out.println("SEED: " + seed);
-        constructCluster(clusterSize);
+        constructCluster(clusterSize, seeds);
         while (!getNodesInState(NodeState.START).isEmpty()) {
             Thread.sleep(50);
         }
-        FuzzTestState state = new FuzzTestState(seed, m_nodes.keySet());
+        FuzzTestState state = new FuzzTestState(0L, m_nodes.keySet());
```
luyangco: Good catch. `FuzzTestState` seed needs to be random too.

tests/frontend/org/voltcore/agreement/TestFuzzMeshArbiter.java
```diff
@@ -595,14 +624,27 @@ public void testFuzz() throws InterruptedException
     }
 
     public void thereBeDragonsHeretestNastyFuzz() throws InterruptedException {
-        long seed = System.currentTimeMillis();
+        m_rejoinLog.info("thereBeDragonsHeretestNastyFuzz");
```
luyangco: For this particular test case, it's because the failure rate is too high. It was disabled for a long while.

tests/frontend/org/voltcore/agreement/TestFuzzMeshArbiter.java
```diff
@@ -645,12 +688,18 @@ private void partitionGraph(Set<Long> nodes, Set<Long> subset)
 
     public void needsWorkTestPartition() throws InterruptedException
     {
-        long seed = System.currentTimeMillis();
-        System.out.println("SEED: " + seed);
-        constructCluster(10);
+        // Fill the array if you want to use specific per-site random seed to reproduce an issue.
```
luyangco: I think this was disabled since 2013, maybe it's time to remove it.

src/frontend/org/voltcore/agreement/AgreementSeeker.java
luyangco: `m_alive` is a multimap of <aliveSiteId, reportingSiteId>. So In 
```Java
if (!m_alive.get(survivor).contains(fm.reportingSite)) {
    return false;
}
```
It means for every survivor `fm.reportingSite` reports, did we see that `fm.reportingSite` reported the same thing before? If yes, continue. Otherwise, return false.

But in the following statement, it means something different.
```Java
if (m_alive.get(survivor).size() == 0) return false;
```
This means for every survivor `fm.reportingSite` reports, did anyone report it before. If yes, continue. Otherwise, return false.

It's possible that `fm.reportingSite` reports the different survivors other than its last report, but maybe other reporters have already reported that, the second if statement returns true in this condition.

src/frontend/org/voltcore/agreement/AgreementSeeker.java
luyangco: See comment above. `m_dead` likes `m_alive`, is a multimap of <deadSiteId, reportingSiteId>.

src/frontend/org/voltcore/agreement/MeshArbiter.java
luyangco: I'll follow your advice.

PR #4702, opened 2017-07-17
Eng 12784 multi stmt proc

dm89:
for java multi stmt proc

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLLexer.java
+++ b/src/frontend/org/voltdb/parser/SQLLexer.java
@@ -391,17 +391,17 @@ public class SQLLexer extends SQLPatternFactory
                 }
             } else {
                 // Outside of a quoted string - watch for the next separator, quote or comment.
-                if ( (iCur == 0 || Character.isWhitespace(buf[iCur-1]) || buf[iCur-1] == ';')
+                if ( (iCur == 0 || !Character.isLetterOrDigit(buf[iCur-1]) || buf[iCur-1] == ';')
                         && (iCur <= buf.length - 4)
                         && String.copyValueOf(buf, iCur, 4).equalsIgnoreCase("CASE")
-                        && (iCur+4 == buf.length || Character.isWhitespace(buf[iCur+4]) )) {
+                        && (iCur+4 == buf.length || !Character.isLetterOrDigit(buf[iCur+4]) )) {
                     inCase++;
                     iCur += 4;
-                } else if ( (iCur == 0 || Character.isWhitespace(buf[iCur-1]) || buf[iCur-1] == ';') // if beginning of string or whitespace character before "begin"
+                } else if ( (iCur == 0 || !Character.isLetterOrDigit(buf[iCur-1]) || buf[iCur-1] == ';') // if beginning of string or whitespace character before "begin"
                         && (buf[iCur] == 'B' || buf[iCur] == 'b')
                         && (iCur <= buf.length - 5)
                         && String.copyValueOf(buf, iCur, 5).equalsIgnoreCase("BEGIN")
-                        && (iCur+5 == buf.length || Character.isWhitespace(buf[iCur+5])) ) {
+                        && (iCur+5 == buf.length || !Character.isLetterOrDigit(buf[iCur+5])) ) {
                     inBegin = true;
                     iCur += 5;
                 } else if ( !inBegin && buf[iCur] == ';') {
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/utils/TestSplitSQLStatements.java
```diff
@@ -70,4 +70,168 @@ public void testSQLSplitter() {
...
+                + "select * from r where f = 'foo';"
+                + "select * from r where f = 'begin' or f = 'END';"
+                + "select a, "
+                + "case when a > 100.00 then "
+                + "case when a > 1000.00 then 'Super Expensive' else 'Pricy' end "
+                + "'Expensive' else 'Cheap' end "
+                + "from t;"
+                + "end;";
+        checkSplitter(sql, sql.substring(0, sql.length() - 1));
+    }
```
BillWhite: It's possible to create a table named BEGIN or a column named BEGIN.  It doesn't cause us trouble here, because BEGIN can't start a SQL statement, and it can only occur at the start of a procedure body here.  So it's usable syntax.  But we don't want to get confused with the perfectly legal statement "select begin.begin from begin;" .  I don't think you will have trouble with this, but it might be a good thing to test.

You might also test that you aren't matching the string 'CASE' inside a longer string.  For example 
```
create procedure p as begin 
    select emptycase from R;
    select caseofbeer from R;
    select suitcaseofbeer from R;
end'.
```
Here emptycase,  caseofbeer and suitcaseofbeer all contain 'case'.  I can't tell from your code if these will be properly handled.  It may be different when 'case' starts or ends a token or is in the middle of one.

dm89: good catch! I updated the code to handle these cases

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -387,8 +391,22 @@ else if (sCommentEnd != null) {
...
             } else {
                 // Outside of a quoted string - watch for the next separator, quote or comment.
-                if (buf[iCur] == ';') {
+                if ( (iCur == 0 || Character.isWhitespace(buf[iCur-1]) || buf[iCur-1] == ';')
+                        && (iCur <= buf.length - 4)
+                        && String.copyValueOf(buf, iCur, 4).equalsIgnoreCase("CASE")
+                        && (iCur+4 == buf.length || Character.isWhitespace(buf[iCur+4]) )) {
+                    inCase++;
+                    iCur += 4;
+                } else if ( (iCur == 0 || Character.isWhitespace(buf[iCur-1]) || buf[iCur-1] == ';') // if beginning of string or whitespace character before "begin"
```
BillWhite: I'm not 100% sure isWhiteSpace is what you want here.  What about

```
select 100+case when a < 0 then 'cheap else 'expensive',  from T;
```
Here the CASE in the case expression is not preceeded by white space.

I think you need to add a function which tells you whether a character can be in an identifier, say isAlphabetic, and ask if buf[iCur+4] is alphabetic.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -387,8 +391,22 @@ else if (sCommentEnd != null) {
...
+                        && (buf[iCur] == 'B' || buf[iCur] == 'b')
+                        && (iCur <= buf.length - 5)
+                        && String.copyValueOf(buf, iCur, 5).equalsIgnoreCase("BEGIN")
```
BillWhite: I just found out about this.  The function String.regionMatches seems to do substring matching inside of strings and can match ignoring case.  You might find that useful.  I don't know if it makes a copy, but you are copying strings a lot here, and that seems expensive.

https://docs.oracle.com/javase/tutorial/java/data/comparestrings.html

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -400,6 +418,19 @@ else if (sCommentEnd != null) {
                     // Start of quoted string.
                     cQuote = buf[iCur];
                     iCur++;
+                } else if ( (iCur == 0 || Character.isWhitespace(buf[iCur-1]) || buf[iCur-1] == ';')
+                        && inBegin && (buf[iCur] == 'E' || buf[iCur] == 'e')
```
BillWhite: Again, this sequence has an end which is not followed by or preceeded by whitespace.
```
select case when id < 0 then (id+0)end+100 from aaa;
```
And this is legal sql.  The CASE expression can be used like any other value-producing expression.

This stuff is all really fiddly.  This is why people produce lexers with tools like flex or antlr.

tests/frontend/org/voltdb/utils/TestSplitSQLStatements.java
BillWhite: You should add test cases like the ones I specified above.

I think you are on your way, but this is really fiddly work.  You are also trying to make changes in a very badly written function, which makes your job even more difficult.

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: I think of another case when the identifiers can have numbers in them.

`select cas1end3ofbeer from R;`

so it should be alpha numeric instead of just alphabetic I suppose.

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: Yes, thanks for the information. It seems to be the most efficient way to match portion of strings in Java 

https://stackoverflow.com/a/25379180/1821324

PR #4703, opened 2017-07-17
Eng 12834 inline insert crash

BillWhite:
Get tuple schemas right in the scan executors.

The tuple schema handling was incorrect in the scan executors for inline insert into select.
We have three interesing schemas.  One is the schema of the scanned input table, one is the
schema of the insert target table and one is the schema of the derived table of the select
statement.  The latter comes from the expressions in the select list.  We were using the
schema from the target table in the insert statement where the last was wanted.  This
caused wrong answers for sequential scan and a server crash for index scan.

I don't know why this was not caught by the unit tests. If the schema of the scanned
table is not exactly like the schema of the insert's target table something will go wrong.
I think this only affects inline insert.  Without inline insert there is an extra temp table to
hold the select column values, and we know to use the schema of that table.
With inline insert there is no extra temp table, so we had to create a tuple from the
schema of the select list.
 
There is a unit test for this now.

### DIFF
--- a/src/ee/executors/indexscanexecutor.cpp
+++ b/src/ee/executors/indexscanexecutor.cpp
@@ -97,12 +97,16 @@ bool IndexScanExecutor::p_init(AbstractPlanNode *abstractNode,
     //
     // INLINE PROJECTION
     //
-    // We can set the projection node here.  But
-    // we can't optimize it until p_execute.  See
-    // the comment there for some insight.
-    //
     m_projectionNode = static_cast<ProjectionPlanNode*>
             (m_node->getInlinePlanNode(PLAN_NODE_TYPE_PROJECTION));
+    //
+    // Optimize the projection if we can.
+    //
+    if (m_projectionNode != NULL) {
+        m_projector = OptimizedProjector(m_projectionNode->getOutputColumnExpressions());
+        m_projector.optimize(m_projectionNode->getOutputTable()->schema(),
+                             m_node->getTargetTable()->schema());
+    }
 
     // For the moment we will not produce a plan with both an
     // inline aggregate and an inline insert node.  This just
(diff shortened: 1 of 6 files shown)


src/ee/executors/indexscanexecutor.cpp
```diff
@@ -213,16 +226,46 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
...
+            // This is in the input schema.  We don't
+            // actually have a tuple with this schema
+            // yet, because we don't have an output
+            // table for the projection node.  That's
+            // the reason for the inline insert node,
+            // after all.  So we have to construct a
+            // tuple which the inline insert will be
+            // happy with.  The p_execute_init knows
+            // how to do this.  Note that temp_tuple will
+            // be initialized if this returns false.
```
wolffcm: Do you mean *not* initialized?

src/ee/executors/indexscanexecutor.cpp
```diff
@@ -201,8 +200,22 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
...
+    // 1.) We can look up the inline projection node any time.
+    // 2.) We need the projection node to get the input schema
+    //     correctly.
+    // 3.) We need the input schema to pass it to p_execute_init
+    //     to initialize the inline aggregate or insert node.
+    // 4.) We need to get the temp_tuple from the inline
+    //     aggregate or insert node, which must be initialized.
+    // 5.) We need the temp_tuple to be computed properly before
+    //     we can optimize the projector.
+    // This could all be done in p_init above, but it's not.
```
wolffcm: If you look at how `p_init` is called, it's called when we retrieve the plan from the cache.  That's where we should be putting any initialization that has to be done only once (and not every time the fragment is executed).  Moving the initialization of the OptimizedProjector to here will degrade performance.

Can you move it back, even if you have to do something kind of fiddly?

src/ee/executors/indexscanexecutor.cpp
```diff
@@ -213,16 +226,46 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
...
         }
     } else {
         temp_tuple = m_outputTable->tempTuple();
     }
 
+    //
+    // Optimize the projection if we can.
+    //
+    if (m_projectionNode != NULL) {
+        m_projector = OptimizedProjector(m_projectionNode->getOutputColumnExpressions());
```
wolffcm: Again, why do this here?  It seems like all the arguments to `optimize` are available in `p_init`.

src/ee/executors/indexscanexecutor.cpp
BillWhite: Ok.  I thought it would be hard, but I think I was remembering a previous attempt at getting the schemas right.

src/ee/executors/indexscanexecutor.cpp
BillWhite: You are right.  I've moved it.

src/ee/executors/indexscanexecutor.cpp
BillWhite: Yes.

BillWhite:
I don't see an OptimizedProjector in the sequential scan executor anywhere.  I think it needs one.

PR #4705, opened 2017-07-17
Eng 12846 sqlcmd multi stmt sp

dm89:


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -234,12 +236,48 @@ public static void getInteractiveQueries(SQLConsoleReader interactiveReader) thr
                 // mistake. That case works differently now, so this code path
                 // MAY be dead. If not, cut our losses by rigging a quick exit.
                 statement.setLength(0);
+                multiStmtLog.setLength(0);
                 line = "EXIT;";
             }
+            // if it is a multi statement procedure
+            if(!inMultiStmtProc) {
```
BillWhite: I personally think this is in the wrong place.  This seems quite fragile to me.  I'm not sure I believe it's going to work.  I'd leave all this out and make changes in parseQuery.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -341,10 +342,17 @@ private static boolean matchToken(String buffer, int position, String token) {
...
         List<String> statements = new ArrayList<>();
+        SplitStmtResults results = new SplitStmtResults();
+
+        // strip out comments
+        String sqlNoComments = SQLParser.AnyWholeLineComments.matcher(sql).replaceAll("");
+        sqlNoComments = SQLParser.EndOfLineComment.matcher(sqlNoComments).replaceAll("");
+
         // Use a character array for efficient character-at-a-time scanning.
-        char[] buf = sql.toCharArray();
+        char[] buf = sqlNoComments.toCharArray();
```
BillWhite: I know this is not your change, but how does this affect non-ascii UTF-8 identifiers?  It's probably not an issue, but it's still something to think about.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -422,10 +428,10 @@ else if (sCommentEnd != null) {
                 }
             } else {
                 // Outside of a quoted string - watch for the next separator, quote or comment.
-                if( matchToken(bufStr, iCur, "case") ) {
+                if( matchToken(sqlNoComments, iCur, "case") ) {
                     inCase++;
                     iCur += 4;
-                } else if ( matchToken(bufStr, iCur, "begin") ) {
+                } else if ( matchToken(sqlNoComments, iCur, "begin") ) {
```
BillWhite: What if the procedure is 'create procedure f as select begin.begin from begin'  You are never inBegin in this case.  I think you can only match the begin at the beginning, so to speak.

It might help to write a grammar.  Maybe something as simple as:
```
STMT ::= MULTIPLE_STMT | SINGLE_STMT;
SINGLE_STMT ::= (^('begin' | 'case'))* (( 'case' SINGLE_STMT 'end' ) ('^'case')*)* '(^'case')*;
MULTIPLE_STMT ::= 'begin' SINGLE_STMT (';' SINGLE_STMT )* 'end' ';'
```
By this I mean something like:

1. A STMT is SINGLE_STMT or a MULTIPLE_STMT.
2. A SINGLE_STMT is a sequence of tokens not equal to 'begin' or 'case', followed by a balanced 'case'/'end' sequence followed by some tokens which are not 'case'.  Maybe semicolon needs to be put in there somehow.
3. A MULTIPLE_STMT is 'begin' followed by a sequence of semicolon separated SINGLE_STMTs followed by 'end.

This kind of a grammar might help organize our thinking.  It doesn't have to be 100% accurate to be useful.

BillWhite: Around line 406 the code checks for cQuote, to see if you are in a single quoted string.  It seems to me this could be a nested loop, and it would make it easier to understand and probably faster.

This is not your change, but it might help, since you are in this area of the code anyway.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -442,7 +448,7 @@ else if (sCommentEnd != null) {
                     // Start of quoted string.
                     cQuote = buf[iCur];
                     iCur++;
-                } else if ( matchToken(bufStr, iCur, "end") ) {
+                } else if ( matchToken(sqlNoComments, iCur, "end") ) {
```
BillWhite: Doesn't it seem strange to copy sqlNoComments out of the String into a char[], because it's somehow more efficient, and then use the heavy-weight sqlNoComments string in matchToken?  Since matchToken is only called from here, maybe changing it's signature would be in order?

This is not a bug, just a suggestion.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -453,14 +459,15 @@ else if (sCommentEnd != null) {
...
                     // Comment (double-dash or C-style)?
-                    if (buf[iCur] == '-' && buf[iCur+1] == '-') {
-                        // One line double-dash comment start.
-                        sCommentEnd = "\n"; // Works for *IX (\n) and Windows (\r\n).
-                        if (iCur == iStart) {
-                            statementIsComment = true;
-                        }
-                        iCur += 2;
-                    } else if (buf[iCur] == '/' && buf[iCur+1] == '*') {
+//                    if (buf[iCur] == '-' && buf[iCur+1] == '-') {
```
BillWhite: If this is dead code it could really just be removed.  I think it's confusing things here.

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -453,14 +459,15 @@ else if (sCommentEnd != null) {
...
+//                        // One line double-dash comment start.
+//                        sCommentEnd = "\n"; // Works for *IX (\n) and Windows (\r\n).
+//                        if (iCur == iStart) {
+//                            statementIsComment = true;
+//                        }
+//                        iCur += 2;
+//                    } else
+                    if (buf[iCur] == '/' && buf[iCur+1] == '*') {
```
BillWhite: What if iCur is the last character.  Doesn't buf[iCur+1] get an exception?

src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -1028,94 +1029,127 @@ private static String parseOptionallyQuotedString(String sIn) throws SQLParser.E
...
-            }
-            fragment = fragment.replace("#(SQL_PARSER_ESCAPE_SINGLE_QUOTE)", "''");
-            queries.add(fragment);
-        }
-        return queries;
-    }
+    // TODO: replace calls to this redundant function with call to SQLLexer.splitStatements
+//    public static List<String> parseQuery(String query)
+//    public static SplitStmtResults parseQuery(String query)
+//    {
```
BillWhite: I would just remove this code rather than comment it out.  You can look at it using git if you want to see it, and I think it clutters the code visually.

But that's your call.

src/frontend/org/voltdb/utils/SplitStmtResults.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.ArrayList;
+
+/**
+ * To store the statements which are parsed/split by the SQLLexer.splitStatements()
+ * currently the incompleteStmt is used to store the incomplete multi statement procedure
+ * and is used to return back to SQLCMD so that more user input can be appended at the end of it
+ */
+public class SplitStmtResults {
+    public List<String> completelyParsedStmts;
```
BillWhite: Personally I would put splitSqlStmts into this class, along with matchToken and maybe some other stuff.  I would have this class hold the list of already parsed statements, and I would let splitSqlStmts be a boolean which tells whether it's a partial or not.

That way all the parsing state, including the list of partial statements, is encapsulated where it's used and not cluttering up the other code.

But that's just me.

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: It is actually my change :)

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: I was thinking to remove these comments with regular expressions in the beginning and `sqlNoComments` will have no comments at all then - and we can avoid this loop and condition checks

dm89: few lines ahead there is a check condition 

```
} else if (iCur <= buf.length - 2) {
                    if (buf[iCur] == '/' && buf[iCur+1] == '*') {
                        // Multi-line C-style comment start.
                        sCommentEnd = "*/";
                        if (iCur == iStart) {
                            statementIsComment = true;
                        }
                        iCur += 2;
```

src/frontend/org/voltdb/parser/SQLParser.java
dm89: removed it

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: removed it

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: its actually a string due to using the `regionMatches` of string function to search for the token

src/frontend/org/voltdb/parser/SQLLexer.java
dm89: its nested `if` conditions - doesn't seem like a loop ... its checking for double escaped quote end of quoted string I guess..

tests/frontend/org/voltdb/utils/TestSplitSQLStatements.java
```diff
@@ -242,6 +250,11 @@ public void testProcSQLSplit() {
         sql = "create procedure p as "
                 + "select begin.begin from begin;";
         checkSplitter(sql+sql1, sql.substring(0, sql.length() - 1), sql1);
+
```
dm89: @BillWhite here I added some tests

BillWhite: But what about a begin in a begin...end block?  Where is that tested?

tests/frontend/org/voltdb/utils/TestSplitSQLStatements.java
```diff
@@ -227,6 +245,16 @@ public void testProcSQLSplit() {
                 + "select emptycase from R; "
```
dm89: @BillWhite  here and above this line are more tests from before

src/frontend/org/voltdb/parser/SQLLexer.java
```diff
@@ -422,12 +428,25 @@ else if (sCommentEnd != null) {
                 }
             } else {
                 // Outside of a quoted string - watch for the next separator, quote or comment.
-                if( matchToken(bufStr, iCur, "case") ) {
+
+                // check if the next token is BEGIN if the previous one is AS
+                // if not, do not look for BEGIN
+                if (checkForNextBegin
+                        && !(Character.isWhitespace(buf[iCur]) || Character.isSpaceChar(buf[iCur])) ) {
```
dm89: it doesn't since it only checks for whitespaces outside of statements

PR #4710, opened 2017-07-18
ENG-12721: fixes to address coverity security report thread-safety is…

xinjiacs:
…sues

Coverity scan find some thread-safety issues

### DIFF
--- a/src/frontend/org/voltdb/sysprocs/saverestore/TableSaveFile.java
+++ b/src/frontend/org/voltdb/sysprocs/saverestore/TableSaveFile.java
@@ -30,6 +30,7 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.zip.Checksum;
 
 import org.apache.hadoop_voltpatches.util.PureJavaCrc32;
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/sysprocs/saverestore/TableSaveFile.java
```diff
@@ -487,7 +486,7 @@ public synchronized boolean hasMoreChunks() throws IOException
     private final int m_totalPartitions;
     private final long m_txnId;
     private final long m_timestamp;
-    private boolean m_hasMoreChunks = true;
+    private volatile boolean m_hasMoreChunks = true;
```
luyangco: A better way is to make it an `AtomicBoolean`, then you don't need those `synchronized` statements.

src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -86,15 +86,13 @@ public PlannerTool(final Database database, byte[] catalogHash)
...
-        if (m_plannerStats == null) {
-            synchronized (this.getClass()) {
-                if (m_plannerStats == null) {
-                    final StatsAgent statsAgent = VoltDB.instance().getStatsAgent();
-                    // In mock test environments there may be no stats agent.
-                    if (statsAgent != null) {
-                        m_plannerStats = new PlannerStatsCollector(-1);
-                        statsAgent.registerStatsSource(StatsSelector.PLANNER, -1, m_plannerStats);
-                    }
+        synchronized (PlannerTool.class) {
```
luyangco: I think the double check of `PlannerTool.class` still makes sense. For example, let's say `m_plannerStats` can only be null before the first instance of `PlannerTool` is created, so in 99.999% of the case, `new PlannerTool()` can be called in parallel and no synchronization is needed. But after this change, multiple `new PlannerTool()` calls must wait for each other to finish the synchronized block, which could lead to some performance degradation.

PR #4711, opened 2017-07-18
ENG-12852: increase the initial NT procedure pool size

xinjiacs:
According to documentation of ThreadPoolExecutor, unfortunately, we used corePoolSize one which basically means it's single thread execution for NT-Procedure until the number of NT-invocations exceeds the queue size 10k. 

This PR tries to increase the corePoolSize and reduce the queue size before starting a new thread. This may start up with more idle threads than before, but still maxed out at 20 or less because of back-pressure. 

John W will run his concurrent UAC/AdHoc system tests against this branch as well.


xinjiacs:
This PR needs more changes: handling the back-pressure tracking correctly for each NT-Procedure invocation.
Performance degression on uncached AdHoc investigation, etc.

PR #4715, opened 2017-07-19
Fix dependency checking for TABLE COUNT plans

wolffcm:
We weren't reporting the set of tables that are used correctly for plans that have a TABLE COUNT node.  As a result it was possible to drop a table that has a plan with a TABLE COUNT node on that table.

### DIFF
--- a/src/frontend/org/voltdb/plannodes/TableCountPlanNode.java
+++ b/src/frontend/org/voltdb/plannodes/TableCountPlanNode.java
@@ -28,17 +28,15 @@ public class TableCountPlanNode extends AbstractScanPlanNode {
         super();
     }
 
-    public TableCountPlanNode(String tableName, String tableAlias) {
-        super(tableName, tableAlias);
-        assert(tableName != null && tableAlias != null);
-    }
-
     public TableCountPlanNode(AbstractScanPlanNode child, AggregatePlanNode apn) {
         super(child.getTargetTableName(), child.getTargetTableAlias());
         m_outputSchema = apn.getOutputSchema().clone();
         m_hasSignificantOutputSchema = true;
         m_estimatedOutputTupleCount = 1;
         m_tableSchema = child.getTableSchema();
+
+        m_tableScan = child.m_tableScan;
+        m_tableScanSchema = child.m_tableScanSchema;
 
         m_isSubQuery = child.isSubQuery();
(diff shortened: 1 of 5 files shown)


tests/frontend/org/voltdb/TestAdhocDropTable.java
```diff
@@ -37,6 +38,11 @@
 
 public class TestAdhocDropTable extends AdhocDDLTestBase {
 
+    @After
+    public void tearItDown() throws Exception {
```
wolffcm: I made this test use an `@After` method to shut down the DB because it's more concise.  Except for the test I added, the other changes are whitespace changes that pull test method bodies out of try/catch/finally blocks.

tests/frontend/org/voltdb/TestAdhocDropTable.java
```diff
@@ -192,73 +193,109 @@ public void testDropTableWithIndexesAndProcedures() throws Exception {
...
+            fail("Should be able to drop table and view with cascade");
+        }
+        resp = m_client.callProcedure("@SystemCatalog", "TABLES");
+        System.out.println(resp.getResults()[0]);
+        assertFalse(findTableInSystemCatalogResults("VIEWBASE"));
+        assertFalse(findTableInSystemCatalogResults("BLAT"));
+    }
+
+    @Test
+    public void testEng12816() throws Exception {
```
wolffcm: This test method is the only truly new code here.

src/frontend/org/voltdb/plannodes/TableCountPlanNode.java
```diff
@@ -28,18 +28,16 @@ public TableCountPlanNode() {
...
-    }
-
     public TableCountPlanNode(AbstractScanPlanNode child, AggregatePlanNode apn) {
         super(child.getTargetTableName(), child.getTargetTableAlias());
         m_outputSchema = apn.getOutputSchema().clone();
         m_hasSignificantOutputSchema = true;
         m_estimatedOutputTupleCount = 1;
         m_tableSchema = child.getTableSchema();
 
+        m_tableScan = child.m_tableScan;
```
wolffcm: `getTablesAndIndexes` needs a table scan instance in order to report tables and indexes used.

PR #4716, opened 2017-07-19
ENG-12702, remove deprecated feature "IMPORT CLASS".

luyangco:
Change-Id: I4ca2cad236b273f5cd40b4badd98615ec97d6361


tests/test_apps/genqa/ddl2.sql
```diff
@@ -1,4 +1,4 @@
-IMPORT CLASS genqa2.procedures.SampleRecord;
+--IMPORT CLASS genqa2.procedures.SampleRecord;
```
xinjiacs: why not remove statements, but comment them out?

luyangco: `ddl2.sql ddl3.sql ddl4.sql` are not being used by the genqa test. But since this is apprunner test, I'd rather ask someone in qa team to remove it.

PR #4718, opened 2017-07-20
Eng-6790-Auto-Reconnect-CSV/Kafka/JDBCLoader

sycai:
Add auto-reconnect feature to CSVLoader, KafkaLoader 8 & 10, and JDBCLoader. They pause and wait for the connection when it is lost.

### DIFF
--- a/src/frontend/org/voltdb/client/ClientImpl.java
+++ b/src/frontend/org/voltdb/client/ClientImpl.java
@@ -80,7 +80,6 @@ public final class ClientImpl implements Client {
     private String m_createConnectionUsername = null;
     private byte[] m_hashedPassword = null;
     private int m_passwordHashCode = 0;
-    private boolean m_autoReconnect;
     final InternalClientStatusListener m_listener = new InternalClientStatusListener();
     ClientStatusListenerExt m_clientStatusListener = null;
 
@@ -159,7 +158,6 @@ public final class ClientImpl implements Client {
             m_reconnectStatusListener = null;
         }
 
-        m_autoReconnect = config.m_autoReconnect;
         m_hashScheme = config.m_hashScheme;
         if (config.m_cleartext) {
             m_passwordHash = ConnectionUtil.getHashedPassword(m_hashScheme, config.m_password);
@@ -1033,7 +1031,7 @@ public final class ClientImpl implements Client {
 
     @Override
     public boolean isAutoReconnectEnabled() {
-        return m_autoReconnect;
+        return (m_reconnectStatusListener != null);
     }
 
     @Override
--- a/src/frontend/org/voltdb/client/ClientImpl.java
+++ b/src/frontend/org/voltdb/client/ClientImpl.java
@@ -22,7 +22,6 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
@@ -81,6 +80,7 @@ public final class ClientImpl implements Client {
     private String m_createConnectionUsername = null;
     private byte[] m_hashedPassword = null;
     private int m_passwordHashCode = 0;
+    private boolean m_autoReconnect;
     final InternalClientStatusListener m_listener = new InternalClientStatusListener();
     ClientStatusListenerExt m_clientStatusListener = null;
 
@@ -159,6 +159,7 @@ public final class ClientImpl implements Client {
             m_reconnectStatusListener = null;
         }
 
+        m_autoReconnect = config.m_autoReconnect;
         m_hashScheme = config.m_hashScheme;
         if (config.m_cleartext) {
             m_passwordHash = ConnectionUtil.getHashedPassword(m_hashScheme, config.m_password);
@@ -881,8 +882,6 @@ public final class ClientImpl implements Client {
 
     private final ReconnectStatusListener m_reconnectStatusListener;
 
-    private final List<VoltBulkLoader> m_bulkLoaders = new ArrayList<>();
-
     @Override
     public void configureBlocking(boolean blocking) {
         m_blockingQueue = blocking;
@@ -1013,9 +1012,7 @@ public final class ClientImpl implements Client {
     public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, boolean upsertMode, BulkLoaderFailureCallBack failureCallback) throws Exception
     {
         synchronized(m_vblGlobals) {
-            VoltBulkLoader loader = new VoltBulkLoader(m_vblGlobals, tableName, maxBatchSize, upsertMode, failureCallback, null);
-            m_bulkLoaders.add(loader);
-            return loader;
+            return new VoltBulkLoader(m_vblGlobals, tableName, maxBatchSize, upsertMode, failureCallback, null);
         }
     }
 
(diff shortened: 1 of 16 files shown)


src/frontend/org/voltdb/utils/CSVTupleDataLoader.java
```diff
@@ -159,25 +160,28 @@ public CSVTupleDataLoader(ClientImpl client, String procName, BulkLoaderErrorHan
...
+                        new PartitionSingleExecuteProcedureCallback(metaData);
+                if (!m_client.callProcedure(cbmt, m_insertProcedure, values)) {
+                    m_log.fatal("Failed to send CSV insert to VoltDB cluster.");
+                    ClientResponse response = new ClientResponseImpl(ClientResponseImpl.SERVER_UNAVAILABLE,
+                            new VoltTable[0], "Failed to call procedure.", 0);
+                    m_errHandler.handleError(metaData, response, "Failed to call procedure.");
+                }
+                // Row inserted successfully, so move on
+                break;
+            } catch (NoConnectionsException ex) {
```
nshi: This is a type of IOException, you don't need a separate handler.

src/frontend/org/voltdb/utils/CSVTupleDataLoader.java
```diff
@@ -96,7 +96,8 @@ public void run() {
                     });
                 }
             }
-            else {
+            // Connection lost response is ignored because this data loader has auto-reconnect feature
+            else if (status != ClientResponse.CONNECTION_LOST){
```
sycai: `insertRow()` is responsible for connection retry

src/frontend/org/voltdb/client/VoltBulkLoader/PerPartitionTable.java
```diff
@@ -299,4 +299,28 @@ private void loadTable(ProcedureCallback callback, VoltTable toSend) throws Exce
...
+
+        while (true) {
+            try {
+                if (m_isMP) {
+                    m_clientImpl.callProcedure(callback, m_procName, m_tableName, m_upsert, toSend);
+                } else {
+                    Object rpartitionParam = VoltType.valueToBytes(toSend.fetchRow(0).get(
+                            m_partitionedColumnIndex, m_partitionColumnType));
+                    m_clientImpl.callProcedure(callback, m_procName, rpartitionParam, m_tableName, m_upsert, toSend);
+                }
```
sycai: Sounds good. I am going to remove `loadTableWithRetry()` and put that retry loop in `loadTable()` instead. It will simplify our code.

src/frontend/org/voltdb/utils/CSVTupleDataLoader.java
sycai: Yes this is definitely a bug. Will fix it.

src/frontend/org/voltdb/client/VoltBulkLoader/PerPartitionTable.java
```diff
@@ -128,6 +128,39 @@ public void run() {
...
+        final VoltBulkLoaderRow m_reinsertRow;
+
+        ReinsertCallback(VoltBulkLoaderRow row) {
+            m_reinsertRow = row;
+        }
+
+        @Override
+        public void clientCallback(ClientResponse response) throws Exception {
+            if (response.getStatus() == ClientResponse.CONNECTION_LOST) {
+                m_es.execute(new Runnable() {
```
akhanzode: PerPartitionTable is under BulkLoader API so wether to resubmit on connection loss depends on client and if its enabled for auto reconnect so per partition table should pick this flag from client that created the bulkloader.
I am debating if we should provide a CLI switch for this going to ask some users. But those who use their own loader using bulkloader should def not get this behavior.

src/frontend/org/voltdb/utils/CSVTupleDataLoader.java
```diff
@@ -159,32 +179,41 @@ public CSVTupleDataLoader(ClientImpl client, String procName, BulkLoaderErrorHan
...
         }
     }
 
     @Override
     public void close() throws InterruptedException, NoConnectionsException
     {
-        m_client.drain();
+        while (m_processedCount.get() != m_totalRowCount.get()) {
+            m_client.drain();
+            Thread.sleep(500);
```
sycai: I copied this while loop from VoltBulkLoader.drain() and changed Thread.yield() to Thread.sleep() because I'm more familiar with sleep(). We can always switch to yield() if it is better suited for this scenario.

src/frontend/org/voltdb/client/ClientImpl.java
```diff
@@ -1029,6 +1029,10 @@ public VoltBulkLoader getNewBulkLoader(String tableName, int maxBatchSize, boole
         }
     }
 
+    public boolean isAutoReconnectEnabled() {
```
akhanzode: I think this should be also on Client interface.

src/frontend/org/voltdb/client/ReconnectStatusListener.java
```diff
@@ -101,6 +101,10 @@ private void connectToOneServerWithRetry(String hostname, int port) {
             try {
                 m_client.createConnection(hostname, port);
                 LOG.info(String.format("Connected to VoltDB node at %s:%d.", hostname, port));
+                synchronized (m_client) {
```
akhanzode: Why do you need this? Is this the case when cluster is down and cant get connection? Also should this be only when auto connect is enabled?

src/frontend/org/voltdb/client/VoltBulkLoader/PerPartitionTable.java
```diff
@@ -128,8 +132,41 @@ public void run() {
...
+        final VoltBulkLoaderRow m_reinsertRow;
+
+        ReinsertCallback(VoltBulkLoaderRow row) {
+            m_reinsertRow = row;
+        }
+
+        @Override
+        public void clientCallback(ClientResponse response) throws Exception {
+            if (response.getStatus() == ClientResponse.CONNECTION_LOST && m_autoReconnect) {
+                m_reinsertExecutor.execute(new Runnable() {
```
akhanzode: Why not use single executor?

src/frontend/org/voltdb/client/VoltBulkLoader/PerPartitionTable.java
```diff
@@ -128,8 +132,41 @@ public void run() {
...
+            m_reinsertRow.m_loader.m_loaderCompletedCnt.incrementAndGet();
+            m_reinsertRow.m_loader.m_outstandingRowCount.decrementAndGet();
+        }
+
+    }
+
     PerPartitionTable(ClientImpl clientImpl, String tableName, int partitionId, boolean isMP,
-            VoltBulkLoader firstLoader, int minBatchTriggerSize, BulkLoaderSuccessCallback successCallback) {
+            VoltBulkLoader firstLoader, int minBatchTriggerSize, BulkLoaderSuccessCallback successCallback,
+            boolean autoReconnect) {
```
akhanzode: You dont need to pass this down you can get it from client

src/frontend/org/voltdb/client/VoltBulkLoader/PerPartitionTable.java
```diff
@@ -226,20 +267,8 @@ private void reinsertFailed(List<VoltBulkLoaderRow> rows) throws Exception {
...
-                        row.m_loader.m_notificationCallBack.failureCallback(row.m_rowHandle, row.m_rowData, response);
-                    }
-
-                    row.m_loader.m_loaderCompletedCnt.incrementAndGet();
-                    row.m_loader.m_outstandingRowCount.decrementAndGet();
-                }
-            };
-
+            //one insert at a time callback
+            ProcedureCallback callback = new ReinsertCallback(row);
```
akhanzode: This can be done in single ProcedureCallback.

src/frontend/org/voltdb/utils/CSVTupleDataLoader.java
```diff
@@ -127,6 +147,8 @@ public CSVTupleDataLoader(ClientImpl client, String procName, BulkLoaderErrorHan
         m_errHandler = errHandler;
         m_callbackExecutor = callbackExecutor;
         m_successCallback = successCallback;
+        m_reinsertExecutor = CoreUtils.getSingleThreadExecutor(procName);
```
akhanzode: You should create this executor only if auto connect is on.

src/frontend/org/voltdb/utils/CSVTupleDataLoader.java
```diff
@@ -159,32 +181,63 @@ public CSVTupleDataLoader(ClientImpl client, String procName, BulkLoaderErrorHan
...
 
     @Override
     public void close() throws InterruptedException, NoConnectionsException
     {
-        m_client.drain();
+        while (m_processedCount.get() != m_totalRowCount.get()) {
+            m_client.drain();
+            Thread.yield();
+        }
+        m_reinsertExecutor.shutdown();
```
akhanzode: only shutdown if its present.

PR #4720, opened 2017-07-20
Eng 11529 catalog jar async write

vittvolt:
## Overview
Now writing an updated catalog jar is moved to a new NT system proc (@VerifyCatalogAndWriteJar).

- ZK lock for one catalog update at a time per cluster (including the async planner thread work, now called NT-Procedure)
- Move class loading verification work to async threads
- Move catalog jar disk writes to async threads

## Pro
- Check https://github.com/VoltDB/pro/pull/1970


xinjiacs:
When you squash, make sure the commit message includes the details from this PR.

PR #4721, opened 2017-07-20
Avoid synchronized blocks in QueueDepthTracker.

manjujames:


src/frontend/org/voltdb/QueueDepthTracker.java
```diff
@@ -62,19 +59,31 @@ public QueueStatus(long timestamp, long max, long total, long count) {
         }
     }
 
-    public synchronized void offerUpdate() {
-        m_depth++;
+    public QueueDepthTracker(long siteId, LinkedTransferQueue<SiteTasker> tasks) {
+        super(siteId, false);
+        m_historicalData = new ConcurrentLinkedDeque<>();
```
nshi: An `ArrayBlockingQueue` will probably do here. It has very good performance and we know the max size of this queue. There's only one writer and one reader.

manjujames: Yes, we do know the size ahead of time! I have made the change.

PR #4723, opened 2017-07-21
Eng 12881 cleanup catalogcontext

xinjiacs:
This PR tries to remove or refactor unused members from CatalogContext and UpdateCore. Ultimately, We want to write the zookeeper catalog bytes asynchronously by removing the transaction id and unique id from procedure runner.

First commit: remove the transaction id in catalog context
Second commit: remove the special ccrTime and refactor unique id with stream table generation id

Background:
We already add a couple of UAC related changes in ENG-12888 branch, these include:
1. ZK lock for one catalog update at a time per cluster (including the async planner thread work, now called NT-Procedure)
2. Move class loading verification work to async threads
3. Move catalog jar disk writes to async threads

Check Pro PR: VoltDB/pro#1974


src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -77,8 +74,7 @@ public ProcedurePartitionInfo(VoltType type, int index) {
     private final byte[] deploymentBytes;
     public final byte[] deploymentHash;
     public final UUID deploymentHashForConfig;
-    public final long m_transactionId;
-    public long m_uniqueId;
+    public long m_genId; // export generation id
```
akhanzode: Can it be final?

xinjiacs: Yes, good catch.

src/frontend/org/voltdb/Inits.java
```diff
@@ -419,7 +415,6 @@ public void run() {
 
             try {
                 m_rvdb.m_catalogContext = new CatalogContext(
-                        catalogStuff.txnId,
                         catalogStuff.uniqueId,
```
vittvolt: should the uniqueId in CatalogAndIds also be renamed genId ?

xinjiacs: I guess we should just delete it as it's not used anywhere.

PR #4726, opened 2017-07-24
ENG-12900: move the catalog zk writes into non-transactional path; cl…

xinjiacs:
…eanup the unnecessary deployment hash parameters

### DIFF
--- a/src/frontend/org/voltdb/utils/CatalogUtil.java
+++ b/src/frontend/org/voltdb/utils/CatalogUtil.java
@@ -24,6 +24,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringWriter;
+import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
@@ -2129,13 +2130,12 @@ public abstract class CatalogUtil {
     {
         ByteBuffer versionAndBytes =
             ByteBuffer.allocate(
+                    8 +  // generation Id
+                    20 + // catalog SHA-1 hash
                     4 +  // catalog bytes length
                     catalogBytes.length +
                     4 +  // deployment bytes length
-                    deploymentBytes.length +
-                    8 +  // generation ID
-                    20 + // catalog SHA-1 hash
-                    20   // deployment SHA-1 hash
+                    deploymentBytes.length
                     );
 
         if (catalogHash == null) {
@@ -2150,7 +2150,6 @@ public abstract class CatalogUtil {
 
         versionAndBytes.putLong(genId);
         versionAndBytes.put(catalogHash);
-        versionAndBytes.put(makeDeploymentHash(deploymentBytes));
         versionAndBytes.putInt(catalogBytes.length);
         versionAndBytes.put(catalogBytes);
         versionAndBytes.putInt(deploymentBytes.length);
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2256,7 +2263,35 @@ public static CatalogAndDeployment getCatalogFromZK(ZooKeeper zk)
         catalogDeploymentBytes.get(deploymentBytes);
         catalogDeploymentBytes = null;
 
-        return new CatalogAndDeployment(genId, catalogHash, deploymentHash, catalogBytes, deploymentBytes);
+        return new CatalogAndDeployment(genId, catalogHash, catalogBytes, deploymentBytes);
+    }
+
+    public static void copyCurrentCatalogToPreviousZK(ZooKeeper zk, long genId,
+            byte[] catalogBytes, byte[] catalogHash, String deployment) {
```
vittvolt: can catalogBytes, catalogHash and deployment be deleted ?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2256,7 +2263,35 @@ public static CatalogAndDeployment getCatalogFromZK(ZooKeeper zk)
...
+    public static void copyPreviousCatalogToCurrentZK(ZooKeeper zk)
+    {
+        try {
+            // read the current catalog bytes
+            byte[] data = zk.getData(VoltZK.catalogbytesPrevious, false, null);
+            assert(data != null);
+            // write to the previous catalog bytes place holder
+            zk.setData(VoltZK.catalogbytes, data, -1);
+        } catch (KeeperException | InterruptedException e) {
+            VoltDB.crashLocalVoltDB("error read catalog bytes from zookeeper", true, e);
```
vittvolt: One thing I'm worried about is that crashLocalVoltDB will call System.exit() in the end, and any wrapping finally block will not be executed. In this case crashing a local database without removing global uac lock might cause future problem ?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2192,23 +2195,36 @@ public static void updateCatalogToZK(ZooKeeper zk,
...
+                                        byte[] catalogHash,
+                                        String deploymentString)
+    {
+        try {
+            byte[] deploymentBytes = deploymentString.getBytes("UTF-8");
+            CatalogUtil.updateCatalogToZK(zk, genId, catalogBytes, catalogHash, deploymentBytes);
+        } catch (UnsupportedEncodingException e) {
+            VoltDB.crashLocalVoltDB("invalid encoding deployment file: " + deploymentString, true, e);
+        } catch (KeeperException | InterruptedException e) {
+            VoltDB.crashLocalVoltDB("error updating catalog bytes to zookeeper", true, e);
```
vittvolt: the same issue in other places, maybe adding removeUACblocker before crashing ?

src/frontend/org/voltdb/utils/CatalogUtil.java
vittvolt: that was my mistake. In crashLocalVoltDB it calls System.exit(-1), so it should throw an exception first that can be caught by finally block.

vittvolt: I found the finally block will not be executed after system.exit(), so maybe this is a valid point to consider.

PR #4728, opened 2017-07-24
ENG-11019  BalanceSPI and integration

zeemanhe:
The pull request covers BalanceSPI and integration:
1.  BalanceSPI system procedure which takes partition id, partition key and new host id.  The procedure will move a partition leader to other host.  
2.  Transaction handling during BalanceSPI.  The current partition leader will be marked as non-leader as soon as it receives BalanceSPI request and update ZooKeeper to trigger new leader promotion process on the new leader site.  There will be a window within which the partition does not have a leader.  The old partition leader will  continuously  receive requests from ClientInterface (Sp transactions) or MpTransactionState (fragment)  and sends these requests back to the senders  until new leader is ready to process requests.   To ensure transactions are in order,  after the new leader is promoted, it will not process requests immediately. It will wait to be notified by the old leader which has cleaned all the transactions before BalanceSPI request and then processes transactions as a leader.  Thus there may be a window within which both old and new leaders will send requests back to senders (transaction restart).  For batched mp transactions, if the first write fragment is processed  on the old leader before BalanceSPI request, the followup fragments will be processed on the same site (old leader will still act as coordinator).
3. All the hosts in the cluster have a periodic task which checks the balance of partition leaders among the cluster.  The task will be executed only when the cluster has full count of live hosts--all the missing hosts join back to the cluster.   BalanceSPI will be initiated on the host with the old leader.   The leader balance starts with the host which has the most partition leaders, one at a time, till the leaders are balanced.


zeemanhe:
Test case on pro: https://github.com/VoltDB/pro/pull/1762

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2110,4 +2120,126 @@ public AuthUser getInternalUser() {
...
+
+        final int partitionId = target.getFirst();
+        final int targetHostId = target.getSecond();
+        int partitionKey = -1;
+
+        VoltTable partitionKeys = TheHashinator.getPartitionKeys(VoltType.INTEGER);
+        ByteBuffer buf = ByteBuffer.allocate(partitionKeys.getSerializedSize());
+        partitionKeys.flattenToBuffer(buf);
+        buf.flip();
+        VoltTable keyCopy = PrivateVoltTableFactory.createVoltTableFromSharedBuffer(buf);
```
zeemanhe: Others may also iterate through the partition keys.    A copy of the partition keys helps avoiding  a race condition.

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2110,4 +2120,126 @@ public AuthUser getInternalUser() {
...
+                //not necessary a failure.
+                tmLog.warn(String.format("Fail to move the leader of partition %d to host %d. %s",
+                        partitionId, targetHostId, resp.getStatusString()));
+            }
+        } catch (IOException | InterruptedException e) {
+            tmLog.warn(String.format("errors in leader change for partition %d: %s", partitionId, e.getMessage()));
+        } finally {
+            long remainingWaitTime = TimeUnit.MINUTES.toSeconds(5);
+            final int waitingInterval = 1;
+            while (remainingWaitTime > 0) {
```
zeemanhe: If the new leader does not show up in 5 min, the host may be down before it gets chance to promote it self.  Otherwise, the cluster may experience serious problems because the the leader promotion process should happen instantly. I think it is safe to remove the blocker on ZooKeeper anyway.

src/frontend/org/voltdb/iv2/Cartographer.java
```diff
@@ -656,4 +711,188 @@ private String doPartitionsHaveReplicas(int hid, Set<Integer> nodesBeingStopped)
...
+
+        //Sort the hosts by partition leader count, descending
+        LinkedList<Host> hostList = new LinkedList<Host>(hostsMap.values());
+        Collections.sort(hostList);
+
+        //only move SPI from the one with most partition leaders
+        //The local ClientInterface will pick it up and start @BalanceSPI
+        Iterator<Host> it = hostList.iterator();
+        Host srcHost = it.next();
+        if (srcHost.m_hostId != localHostId) {
```
zeemanhe: will delay acquiring the lock

src/frontend/org/voltdb/iv2/InitiatorMailbox.java
```diff
@@ -282,6 +312,7 @@ void run() {
     protected void deliverInternal(VoltMessage message) {
         assert(lockingVows());
         logRxMessage(message);
+        notifyNewLeaderOfTxnDone();
```
zeemanhe: will move the notification to Scheduler.handleInitiateResponseMessage().

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2110,4 +2120,131 @@ public AuthUser getInternalUser() {
...
+
+            //wait for the Cartographer to see the new partition leader. The leader promotion process should happens instantly.
+            //If the new partition leader does not show up in 5 min, the host with new partition leader does not get chance
+            //to promote itself before it is down or stale--the cluster ma experience problems.
+            //remove the blocker on ZooKeeper anyway
+            long remainingWaitTime = TimeUnit.MINUTES.toMillis(5);
+            final long waitingInterval = TimeUnit.SECONDS.toMillis(1);
+            while (remainingWaitTime > 0) {
+                try {
+                    Thread.sleep(waitingInterval);
```
luyangco: Another problem is that sleep() here stops the `periodicThread` for up to 5 min, other tasks on the task queue will be blocked until @BalanceSPI is done, this may cause other unknown problems. Why not use a separate ExecutorService thread to do it? 

`CoreUtils.getCachedSingleThreadExecutor()` can create such a thread only when needed.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2119,6 +2167,38 @@ public void run() {
...
+
+    // The latest joined host will take over partition leader migration.
+    private void scheduleBalanceSpiTask() {
+        final boolean disableSpiTask = "true".equals(System.getProperty("DISABLE_SPI_BALANCE", "false"));
+        if (disableSpiTask) {
+            hostLog.info("Balance SPI is not scheduled.");
+            return;
+        }
+        //cleanup when a node start
+        CoreZK.removeSPIBalanceInfo(m_messenger.getZK());
```
luyangco: Another `CoreZK.removeSPIBalanceInfo(m_messenger.getZK())` is called when one node finally grab `SPIBalanceIndicator` lock, is there any reason of removing the SPIBalanceInfo here?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1687,6 +1691,50 @@ public void run()
...
+        //the new leader has not got a chance to be promoted yet. then re-install the old leader
+        if (failedHosts.contains(spiInfo.getNewLeaderHostId()) &&
+                spiInfo.getOldLeaderHostId() == m_messenger.getHostId() &&
+                spiInfo.getOldLeaderHostId() == currentLeaderHostId) {
+            SpInitiator initiator = (SpInitiator)m_iv2Initiators.get(spiInfo.getPartitionId());
+
+            //leader on this host but the initiator is marked as none-leader
+            //re-elect the older site as leader
+            if (!initiator.isLeader()) {
+                LeaderCache leaderAppointee = new LeaderCache(m_messenger.getZK(), VoltZK.iv2appointees);
```
luyangco: Same comment also applies here, what if the old leader dies before setting up the new leader, then there may be  two leader elections run in parallel (this one plus the leader election caused by node failure).

src/frontend/org/voltdb/ClientInterface.java
zeemanhe: update: executor service will be started as needed upon rejoin.

src/frontend/org/voltdb/RealVoltDB.java
zeemanhe: this task will be started only if the cluster is full (last rejoining node) . So removing any previously persisted info.

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2102,4 +2123,165 @@ public AuthUser getInternalUser() {
...
+        if (target == null) {
+            return;
+        }
+
+        final int partitionId = target.getFirst();
+        final int targetHostId = target.getSecond();
+        int partitionKey = -1;
+
+        //Balance spi is completed or there are hosts down. Stop BalanceSPI service on this host
+        if (targetHostId == -1 || !voltDB.isClusterCompelte()) {
```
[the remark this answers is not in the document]
zeemanhe: spi balance is triggered only if a cluster is complete after rejoining.  When any nodes are down,  the leader appointer will handle the new leader selection.  spi balance shouldn't be needed in this case.

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2102,4 +2123,165 @@ public AuthUser getInternalUser() {
...
+                //not necessary a failure.
+                tmLog.warn(String.format("Fail to move the leader of partition %d to host %d. %s",
+                        partitionId, targetHostId, resp.getStatusString()));
+            }
+        } catch (IOException | InterruptedException e) {
+            tmLog.warn(String.format("errors in leader change for partition %d: %s", partitionId, e.getMessage()));
+        } finally {
+
+            //wait for the Cartographer to see the new partition leader. The leader promotion process should happen instantly.
+            //If the new leader does not show up in 5 min, the cluster may have experienced host-down events.
```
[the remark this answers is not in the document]
zeemanhe: This is a to check if Cartographer has updated its master list every second.  As soon as  it has seen the master list,  the spi balance blocker will be removed. So it will not be 5 min if the host which gets new spi assigned completes the spi promotion.  Otherwise, if the host which gets the new spi is down and fails to promote,  the loop will also be broken and  the spi balance blocker will be removed.

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
```diff
@@ -78,6 +81,8 @@ public CompleteTransactionMessage(long initiatorHSId, long coordinatorHSId,
         setBit(ISROLLBACK, isRollback);
         setBit(REQUIRESACK, requiresAck);
         setBit(ISRESTART, isRestart);
+        m_toLeader = false;
```
xinjiacs: Can you use the inherited "m_isLeaderToReplica" instead of adding a new "m_toLeader"? The message is either from leader to replica or from MPI to leader?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1174,35 +1202,40 @@ else if (result == DuplicateCounter.MISMATCH) {
...
     {
         CompleteTransactionMessage msg = message;
-        if (m_isLeader) {
+        TransactionState txn = m_outstandingTxns.get(msg.getTxnId());
+
+        // 1) The site is not a leader any more, thanks to spi migration but the message is intended for leader.
+        //    action: advance TxnEgo, send it to all original replicas (before spi migration)
+        // 2) The site is the new leader but the message is intended for replica
+        //    action: no TxnEgo advance
+        if ((m_isLeader && message.isToLeader()) || message.isToLeader()) {
```
xinjiacs: Maybe we do not care flag of "m_isLeader" and simply check " message.isToLeader()"?

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
```diff
@@ -30,6 +30,9 @@
     boolean m_requiresAck;
     boolean m_rollbackForFault;
 
+    //indicate if this message is sent to partition leader
+    boolean m_toLeader;
+    boolean m_ackRequestedFromSender;
```
xinjiacs: Is it always "TRUE"? not used?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -247,6 +259,11 @@ public void updateReplicas(List<Long> replicas, Map<Integer, Long> partitionMast
             }
         }
 
+        //notify the new partition leader that the old leader has completed the Txns if needed.
+        if (!m_isLeader && m_mailbox instanceof InitiatorMailbox) {
```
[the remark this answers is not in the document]
zeemanhe: could be a MockMailbox from unit test.

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
zeemanhe: it is a bug. should be initialized to be false

src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2199,6 +2199,10 @@ void startMigratePartitionLeader() {
         }
 
         //grab a lock
+        if (VoltZK.isSnapshotInProgress(m_zk)) {
+            tmLog.rateLimitedLog(60, Level.INFO, null, "Snapshot in progress");
+            return;
+        }
```
xinjiacs: Even you check the snapshot here, it does not prevent the race condition. The snapshot may still start immediately after this check, but before the "VoltZK.migratePartitionLeaderBlocker" getting created.

Do we have a ZK node for snapshot when it gets started? Could it get triggered while MigratePartitionLeader is running?

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
```diff
@@ -30,6 +30,8 @@
     boolean m_requiresAck;
     boolean m_rollbackForFault;
 
+    //indicate if this message is sent to partition leader
+    boolean m_ackRequestedFromSender;
```
xinjiacs: It seems that we do not need this extra flag as well. 

"m_isForReplica" can replace it:
If true, it needs ack requestes.
If false, it does not need ack requests. 
Later this information is used to advance truncation handler on SpScheduler.

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
```diff
@@ -78,6 +80,8 @@ public CompleteTransactionMessage(long initiatorHSId, long coordinatorHSId,
         setBit(ISROLLBACK, isRollback);
         setBit(REQUIRESACK, requiresAck);
         setBit(ISRESTART, isRestart);
+        m_isForReplica = true;
```
xinjiacs: Why change the default value?

src/frontend/org/voltdb/messaging/CompleteTransactionResponseMessage.java
```diff
@@ -42,6 +43,7 @@ public CompleteTransactionResponseMessage(CompleteTransactionMessage msg)
         m_spHandle = msg.getSpHandle();
         m_isRestart = msg.isRestart();
         m_spiHSId = msg.getCoordinatorHSId();
+        m_ackRequestedFromSender = msg.isAckRequestedFromSender();
```
xinjiacs: We can get the flag status from CompleteTransactionMessage's "m_isForReplica".

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1503,14 +1543,27 @@ void run() {
...
             existingDC.logWithCollidingDuplicateCounters(counter);
             VoltDB.crashGlobalVoltDB("DUPLICATE COUNTER MISMATCH: two duplicate counter keys collided.", true, null);
         }
-        else {
-            m_duplicateCounters.put(dpKey, counter);
+    }
+
+    //Both former and current partition leader may send CompleteTransactionMessage
+    //over, which may introduce duplicate counter collision.
+    boolean skipCollisionFromMigratePartitionLeader(DuplicateCounter counter1, DuplicateCounter counter2) {
```
xinjiacs: I do not understand why we need to treat CompleteTransactionMessage special here. Is it possible for the MPI issuing the CompleteTransactionMessage to the new leader which should be handled by the old leader instead?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1547,31 +1600,35 @@ private long getRepairLogTruncationHandleForReplicas()
...
             // These include node failures (promotion phase) and node rejoin (early rejoin phase).
             if (tmLog.isDebugEnabled()) {
                 tmLog.debug("Updating truncation point from " + TxnEgo.txnIdToString(m_repairLogTruncationHandle) +
-                        "to" + TxnEgo.txnIdToString(newHandle));
+                        "to" + TxnEgo.txnIdToString(newHandle)+ " isLeader:" + m_isLeader + " isForLeader:" + isForLeader);
             }
         }
     }
 
+    private void setRepairLogTruncationHandle(long newHandle)
```
xinjiacs: There is only one caller. Better to remove this API and use the full API always?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1612,11 +1669,44 @@ void run()
         });
     }
 
-    private void logRepair(VoltMessage message) {
+    public TransactionState getTransaction(long txnId) {
```
xinjiacs: rename to "getTransactionState()"?

src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
```diff
@@ -219,9 +235,10 @@ public int getSerializedSize()
             + 1 // read only
             + 1 // node recovering indication
             + 1 // mispartitioned invocation
+            + 1 // createdFromLeader
             + m_response.getSerializedSize();
 
-        if (m_mispartitioned) {
+        if (m_mispartitioned || isMisrouted()) {
```
xinjiacs: what's the relation between "m_mispartitioned" and "isMisrouted()"? Can them be combined?

tests/frontend/org/voltdb/iv2/RandomMsgGenerator.java
```diff
@@ -96,6 +96,7 @@ private CompleteTransactionMessage makeCompleteTxnMsg(boolean readOnly, boolean
         CompleteTransactionMessage msg =
             new CompleteTransactionMessage(0l, 0l, m_mpiTxnEgo.getTxnId(), readOnly, 0, isRollback,
                     false, isRestart, false);
+        msg.setForReplica(false);
```
xinjiacs: If by default it's false, then you can remove it.

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
zeemanhe: may use requiresAck bit instead. This bit is not used to my knowledge in IV2.

src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
zeemanhe: m_mispartitioned uses ClientResponse.TXN_RESTART.  It looks like this member shouldn't be needed because m_response has it.  isMisrouted() just checks the response status.

src/frontend/org/voltdb/iv2/SpScheduler.java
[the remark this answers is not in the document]
xinjiacs: Can you add comments to this IF check?

src/frontend/org/voltdb/VoltZK.java
```diff
@@ -372,6 +426,15 @@ public static String createActionBlocker(ZooKeeper zk, String node, CreateMode m
...
                             "does not support remote elastic join during DR. " +
                             "DR needs to be reset before elastic join is allowed again.";
+                } else if ( blockers.contains(migrate_partition_leader)) {
+                    errorMsg = "while leader migration is active. Please retry elastic join later.";
+                }
+                break;
+            case migratePartitionLeaderBlocker:
+                //MigratePartitionLeader can not happen when join, rejoin, catalog update is in progress.
+                blockers.remove(leafNodeBanElasticJoin);
+                if (blockers.size() > 1) {
```
xinjiacs: @zeemanhe If I understand correct, there is a new action blocker "banElasticJoin" which can be a persistent zk node. "size() > 1" will not be valid check here.

zeemanhe: Action blocker "banElasticJoin" does not block leader migration. So it is counted out by removing it in the children list. The check should be valid.

PR #4731, opened 2017-07-25
ENG-12778-CSVLoader-Hang-On-Error

sycai:
The errors occurs when VoltDecimalHelper fails to deserialize decimal from a string and throws out a RuntimeException. This exception is not caught in PerPartitionTable.buildTable(), so it will eventually propagate to the console. Also, m_outstandingRowCount is not properly updated, which is why the loader doesn't stop running.


akhanzode:
Looks good to me.

PR #4732, opened 2017-07-25
ENG-12902 resolve importer IllegalStateException by forcing importer into a valid state.

brians127:
This is intended as a low risk fix for a crash scenario when the importer sees a UAC during pause and then resumes. It does not eliminate the root cause scenario or resolve importers starting when the database is paused, but the database can now continue into a valid state without crashing.


brians127:
Superceded by https://github.com/VoltDB/voltdb/pull/4735. If that branch looks OK there is no need to merge this one.

PR #4733, opened 2017-07-25
Eng 12901 valgrind botch

BillWhite:
Report valgrind error messages.

Don't depend on the return code from valgrind only to determine valgrind errors.


tests/ee/expressions/expression_test.cpp
```diff
@@ -338,9 +338,9 @@ TEST_F(ExpressionTest, SimpleUnaryMinus) {
     // -5
     e.push(new CV(EXPRESSION_TYPE_VALUE_CONSTANT, VALUE_TYPE_TINYINT, 1, (int64_t)5));
     e.push(new AE(EXPRESSION_TYPE_OPERATOR_UNARY_MINUS, VALUE_TYPE_TINYINT, 1));
-    // dummy left expression to prevent segmentation fault
+    // dummy right expression to prevent segmentation fault
```
dm89: maybe we can remove this comment now since there is no dummy expression..

BillWhite: I was thinking we had to add the NULL, so that is kind of a dummy expression.

If you don't add an expression here the tree gets mangled.  We pop the operation and try to interpret it as a tree.  No good will come of that.

PR #4734, opened 2017-07-25
Eng 12850 exec multi stmt sp

dm89:
backend for multi statement procedures


src/frontend/org/voltdb/compiler/statements/CreateProcedureAsSQL.java
```diff
@@ -37,10 +37,19 @@ public CreateProcedureAsSQL(DDLCompiler ddlCompiler) {
...
             throws VoltCompilerException {
-        // Matches if it is CREATE PROCEDURE <proc-name> [ALLOW <role> ...] [PARTITION ON ...] AS <select-or-dml-statement>
-        Matcher statementMatcher = SQLParser.matchCreateProcedureAsSQL(ddlStatement.statement);
+
+        /* first check if it matches a multi statement procedure
+         * if not, then check if it is a single statement procedure
+         */
+        // Matches if it is CREATE PROCEDURE <proc-name> [ALLOW <role> ...] [PARTITION ON ...] AS BEGIN <select-or-dml-statement>
+        Matcher statementMatcher = SQLParser.matchCreateMultiStmtProcedureAsSQL(ddlStatement.statement);
+
```
BillWhite: Why do you have to match the 'end', which is possibly a red herring anyway?  What if matchCreateMultiStmtProcedureAsSQL matched 
```
'CREATE' 'PROCEDURE' ident unparsedProcedureModifierClauses() 'as' 'begin'
```
without messing with the stuff after the ```begin```.  Then you can tell where the ```begin``` started, since you matched it, and you know you will be parsing at the start of the hard part.

src/frontend/org/voltdb/sysprocs/LoadSinglepartitionTable.java
```diff
@@ -164,8 +164,8 @@ public long run(SystemProcedureExecutionContext ctx,
                             action,tableName));
         }
 
-        // statements of all single-statement procs are named "sql"
-        Statement catStmt = p.getStatements().get(VoltDB.ANON_STMT_NAME);
+        // statements of all single-statement procs are named "sql0"
+        Statement catStmt = p.getStatements().get(VoltDB.ANON_STMT_NAME + "0");
```
BillWhite: You could put this comment above.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -1171,533 +1152,482 @@ else if (type == VoltType.NUMERIC) {
...
-               paramBuf = ByteBuffer.allocate(queuedSQL.params.getSerializedSize());
-               queuedSQL.params.flattenToBuffer(paramBuf);
-           } catch (IOException e) {
-               throw new RuntimeException("Error serializing parameters for SQL statement: " +
-                                          queuedSQL.stmt.getText() + " with params: " +
-                                          queuedSQL.params.toJSONString(), e);
-           }
-           assert(paramBuf != null);
-           paramBuf.flip();
+    /**
```
dm89: code below from here till the end of the file is just formatted (accidentally using eclipse)

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -222,38 +226,37 @@
...
-     * This function returns the ExecutionEngine for this site.
-     * ProcedureRunner needs the access to the ExecutionEngine for two purposes:
-     *     - Telling the C++ side to turn on/off the per-fragment time measurement.
-     *     - Collecting the time measurements and the failure indicator from
-     *       the per-fragment stats shared buffer.
-     * The ExecutionEngine may not have been initialized when a ProcedureRunner is created,
-     *     (See BaseInitiator.java, the configureCommon() function)
-     * so we do not find this engine in the constructor.
+     * This function returns the ExecutionEngine for this site. ProcedureRunner
+     * needs the access to the ExecutionEngine for two purposes: - Telling the
```
dm89: updated it to old format

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -552,10 +578,14 @@ public boolean checkPartition(TransactionState txnState, TheHashinator hashinato
                 parameterAtIndex = invocation.getParameterAtIndex(m_partitionColumn);
             }
 
-            // Note that @LoadSinglepartitionTable has problems if the parititoning param
-            // uses integers as bytes and isn't padded to 8b or using the right byte order.
-            // Since this is not exposed to users, we're ok for now. The right fix is to probably
-            // accept the right partitioning type from the user, then rewrite the params internally
+            // Note that @LoadSinglepartitionTable has problems if the
+            // parititoning param
```
dm89: yea it was accidental. I updated some parts to old format

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -393,108 +396,125 @@ private ClientResponseImpl coreCall(Object... paramListIn) {
...
-                        VoltTable table =
-                            getNonVoltDBBackendIfExists().runSQLWithSubstitutions(
-                                m_cachedSingleStmt.stmt,
-                                m_cachedSingleStmt.params,
-                                m_cachedSingleStmt.stmt.statementParamTypes);
-                        table.convertToHeapBuffer();
-                        results = new VoltTable[] { table };
+
+                    for (int i = 0; i < m_catProc.getStatements().size(); i++) {
+                        QueuedSQL curStmt = m_sqlStmts.get(i);
```
dm89: this part is extension of previous work with a loop - the loop is the new code - it is to get each statement and the parameters corresponding to that statement for execution

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -393,108 +396,125 @@ private ClientResponseImpl coreCall(Object... paramListIn) {
...
+                        int numStmtParams = curStmt.stmt.statementParamTypes.length;
+                        Object[] stmtParamList = Arrays.copyOfRange(paramList, curParamOffset,
```
dm89: `import java.util.Arrays;` was imported for this copying of statement parameters

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -26,7 +26,9 @@
 import java.lang.reflect.Modifier;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
```
dm89: I added the Arrays because I used `Arrays.copyOfRange` to copy parameters for each statement in a multi statement procedure.

I accidentally formatted and committed due some keyboard shortcut for eclipse and it is mixed up with other behavioral code changes - so could not revert only the formatting- will be more careful about it from now on.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -864,8 +864,46 @@ private void fillTrackerFromXML()
     private static int kStateReadingEndCodeBlockDelim = 10 ;      // dealing with ending code block delimiter ###
     private static int kStateReadingEndCodeBlockNextDelim = 11;   // dealing with ending code block delimiter ###
 
+    // To indicate if inside multi statement procedure
+    private static boolean inBegin = false;
+    // To indicate if inside CASE .. WHEN .. END
```
dm89: its not a k-state value since we cannot have a begin within a begin and also when we check for next begin only after `as`

Added comments in the code

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -864,8 +864,46 @@ private void fillTrackerFromXML()
...
+    // BEGIN should follow AS for create procedure
+    // added this case since 'begin' can be table or column name
+    private static boolean checkForNextBegin = false;
 
     private static int readingState(char[] nchar, DDLStatement retval) {
+
+        if (!Character.isLetterOrDigit(nchar[0])) {
+            char prev = prevChar(retval.statement);
+            if (checkForNextBegin) {
+                if (prev == 'n' || prev == 'N') {
```
dm89: I agree its confusing since the code is already this way - I found this to be the simplest what I could think of since we only have access to the current character and the characters we have seen so far

src/frontend/org/voltdb/compiler/StatementCompiler.java
```diff
@@ -407,7 +407,7 @@ public static Procedure compileDefaultProcedure(PlannerTool plannerTool, Procedu
         CatalogMap<Statement> statements = newCatProc.getStatements();
         assert(statements != null);
 
-        Statement stmt = statements.add(VoltDB.ANON_STMT_NAME);
+        Statement stmt = statements.add(VoltDB.ANON_STMT_NAME + "0");
```
dm89: Since the statements are now annotated as `sql0` even for single statement procedures

src/frontend/org/voltdb/sysprocs/LoadSinglepartitionTable.java
dm89: added comments

src/frontend/org/voltdb/compiler/DDLCompiler.java
BillWhite: Ah, I tried to argue with this, and in arguing figured out why this is so.  The kState values are states in a FSM.  The inBegin, inCase and checkForNextBegin elements are the left context, which would be on a pushdown automata if we were doing this in a more theory-informed way.

I think you are right.

PR #4735, opened 2017-07-25
ENG-12929 importers no longer start up if the cluster is paused.

brians127:
All setup work to (re)start an importer is still done regardless of the paused state, but if the cluster is paused, the importers do not run. The only reason setup work is done at all is for error checking - otherwise, errors from schema and deployment updates would be deferred until 'voltdb resume'.

RealVoltDB's m_startAction should now be a reliable indicator for whether or not the node is starting in the paused state. ImportManager / ChannelDistributer need it to initialize their state so they can accurately detect when the cluster transitions to running. There is still room for cleanup in this code; I didn't attempt to reduce the number of places where this can get set.

ENG-12929 was split off from ENG-12902 and will be used for the squash merge commit.

Tests in Pro: https://github.com/VoltDB/pro/pull/1999


src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -74,6 +75,9 @@
     // about the jars to use
     private Map<String, ImportConfiguration> m_processorConfig = new HashMap<>();
 
+    // If this is set, importers can be restarted even if the catalog has not changed
+    private boolean m_importersDisabled = false;
```
akhanzode: So there is a m_serverStarted flag can it be used or replaced?

src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -328,8 +332,14 @@ public synchronized void readyForDataInternal(CatalogContext catalogContext, Hos
         if (m_processor.get() == null) {
             return;
         }
-        //Tell import processors and in turn ImportHandlers that we are ready to take in data.
-        m_processor.get().readyForData(catalogContext, messenger);
+
+        if (VoltDB.instance().getMode() != OperationMode.PAUSED) {
+            //Tell import processors and in turn ImportHandlers that we are ready to take in data.
+            m_processor.get().readyForData(catalogContext, messenger);
```
akhanzode: If that flag is gone there is no need for readyForData() or readyForDataInternal() becomes readyForData()

src/frontend/org/voltdb/importer/ImportManager.java
nshi: Why do we want to restart them even if nothing changed?

brians127: I'll look into this.

brians127:
Progress so far: 
- Code is not done but review is welcome
- Rejoin manual test case NPEs since startMode is null.
- In a 3 node cluster that starts paused, importers now reliably come up during resume but not while the cluster is paused!
- I plan to incorporate the JUnit I finished in a different branch. More if time permits.

sycai:
Branch update: After user elastically joins a node into a paused cluster, the importer of the newly-added node does not pause

PR #4739, opened 2017-07-26
Eng 12924

luyangco:
When the cluster has >3 hosts and K-factor set to be >= 1, nodes within the cluster will automatically be assigned into groups ( a.k.a partition group, each group has K+1 members). To fully utilize CPU cores, we used to create a few secondary PicoNetwork thread to evenly distribute the internal traffic across those threads. 

However there was a bug which will prevent some members within a partition group to use secondary PicoNetwork threads, which in turn leads to a traffic skew between PicoNetwork threads.

Profiler result also shows performance hotspot in primary PicoNetwork thread, which slows down Site threads.

This patch fixes the problem by monitoring how many secondary PicoNetwork threads are created on every host, once there are enough threads then HostMessenger mark itself as ready to use secondary PicoNetwork threads.

This patch also gives secondary PicoNetwork a different name by adding a _p suffix, this helps identifying thread in @statistics IOStats.

### DIFF
--- a/src/frontend/org/voltcore/messaging/HostMessenger.java
+++ b/src/frontend/org/voltcore/messaging/HostMessenger.java
@@ -1089,7 +1089,6 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
         boolean ready = true;
         for (int peer : m_peers) {
             ready = ready && m_foreignHosts.get(peer).size() == (m_secondaryConnections + 1);
-            m_hostLog.warn("NotifyOfConnection For hostId " + hostId + " ready=" + ready);
         }
         if (ready) {
             // Now it's time to use secondary pico network, see comments in presend() to know why we can't
@@ -1304,7 +1303,10 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
                 if (fhost == null) {
                     int index = Math.abs(m_nextForeignHost.getAndIncrement() % fhosts.size());
                     fhost = (ForeignHost) fhosts.toArray()[index];
-                    m_hostLog.warn("bind " + CoreUtils.getHostIdFromHSId(hsId) + ":" + CoreUtils.getSiteIdFromHSId(hsId) + " to " + fhost);
+                    if (m_hostLog.isDebugEnabled()) {
+                        m_hostLog.debug("bind " + CoreUtils.getHostIdFromHSId(hsId) + ":" + CoreUtils.getSiteIdFromHSId(hsId) +
+                                " to " + fhost.hostnameAndIPAndPort());
+                    }
                     bindForeignHost(hsId, fhost);
                 }
             } else {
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1076,9 +1081,18 @@ public void notifyOfConnection(
         prepSocketChannel(socket);
         // Auxiliary connection never time out
         ForeignHost fhost = new ForeignHost(this, hostId, socket, Integer.MAX_VALUE,
-                listeningAddress, new PicoNetwork(socket));
+                listeningAddress, new PicoNetwork(socket, true));
         putForeignHost(hostId, fhost);
         fhost.enableRead(VERBOTEN_THREADS);
+        boolean ready = true;
+        for (int peer : m_peers) {
+            ready = ready && m_foreignHosts.get(peer).size() == (m_secondaryConnections + 1);
```
zeemanhe: can break out the loop earlier, as soon as ready is false.

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1287,6 +1301,10 @@ ForeignHost presend(long hsId, VoltMessage message)
                 if (fhost == null) {
                     int index = Math.abs(m_nextForeignHost.getAndIncrement() % fhosts.size());
                     fhost = (ForeignHost) fhosts.toArray()[index];
+                    if (m_hostLog.isDebugEnabled()) {
```
zeemanhe: can you rename the static loggers , remove "m_"?

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1700,18 +1718,80 @@ public void cutLink(int hostIdA, int hostIdB) {
...
+        m_peers = partitionGroupPeers;
+        m_secondaryConnections = computeSecondaryConnections(hostCount);
+    }
+
+    private int computeSecondaryConnections(int hostCount) {
+        /**
+         *  Basic goal is each host should has the same number of connections compare to the number
+         *  without partition group layout.
+         *
+         * (targetConnectionsWithinPG - existingConnectionsWithinPG) is the the total number of secondary
```
zeemanhe: move the java doc out of the method?

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1722,7 +1802,15 @@ public void createAuxiliaryConnections(Set<Integer> peers, int secondaryConnecti
                 }
             }
         }
-        m_hasAllSecondaryConnectionCreated = true;
+        boolean ready = true;
+        for (int hostId : m_peers) {
+            ready = ready && m_foreignHosts.get(hostId).size() == (m_secondaryConnections + 1);
+        }
```
zeemanhe: same as the previous comment. can break out the loop earlier, as soon as ready is false.

src/frontend/org/voltcore/network/PicoNetwork.java
```diff
@@ -127,17 +127,25 @@ public void start(InputHandler ih, Set<Long> verbotenThreads) {
...
         m_remoteSocketAddressString = remoteAddress.getAddress().getHostAddress();
         m_remoteHostAndAddressAndPort = "/" + m_remoteSocketAddressString + ":" + m_remoteSocketAddress.getPort();
-        m_toString = super.toString() + ":" + m_remoteHostAndAddressAndPort;
         String remoteHost = ReverseDNSCache.hostnameOrAddress(m_remoteSocketAddress.getAddress());
         if (!remoteHost.equals(m_remoteSocketAddress.getAddress().getHostAddress())) {
             m_remoteHostname = remoteHost;
             m_remoteHostAndAddressAndPort = remoteHost + m_remoteHostAndAddressAndPort;
+        }
+        if (isSecondary) {
+            m_toString = super.toString() + ":" + m_remoteHostAndAddressAndPort + "(s)";
```
zeemanhe: can you change m_toString to a name more meaningful?

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -1700,29 +1710,104 @@ public void cutLink(int hostIdA, int hostIdB) {
...
+        if (hostsToConnect.isEmpty()) return;
+
+        for (int hostId : hostsToConnect) {
+            Iterator<ForeignHost> it = m_foreignHosts.get(hostId).iterator();
+            if (it.hasNext()) {
+                InetSocketAddress listeningAddress = it.next().m_listeningAddress;
+                for (int ii = 0; ii < m_secondaryConnections; ii++) {
                     try {
-                        SocketChannel socket = m_joiner.requestForConnection(fh.m_listeningAddress);
+                        SocketChannel socket = m_joiner.requestForConnection(listeningAddress);
```
xinjiacs: You only need "listeningAddress" and it's not changed with the number of secondary connections. So you can pull the above iterator out side of the for loop.

PR #4740, opened 2017-07-26
ENG-12921: add the catalog version back to zk, because rejoin node ne…

xinjiacs:
…eds the latest version to pass the deterministic hash mismatch check

### DIFF
--- a/src/frontend/org/voltdb/utils/CatalogUtil.java
+++ b/src/frontend/org/voltdb/utils/CatalogUtil.java
@@ -2122,6 +2122,7 @@ public abstract class CatalogUtil {
     }
 
     private static ByteBuffer makeCatalogAndDeploymentBytes(
+                int version,
                 long genId,
                 byte[] catalogBytes,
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2122,13 +2122,15 @@ private static String extractPassword(String password, ClientAuthScheme scheme)
...
     private static ByteBuffer makeCatalogAndDeploymentBytes(
+                int version,
                 long genId,
                 byte[] catalogBytes,
                 byte[] catalogHash,
                 byte[] deploymentBytes)
     {
         ByteBuffer versionAndBytes =
             ByteBuffer.allocate(
+                    4 +  // version number
```
vittvolt: Btw can we use short ?

xinjiacs: Short is too small and will have overflow issue. And we use version to compare as well.

PR #4759, opened 2017-07-31
[ENG-10944] phase 1: Supporting 2p syntax, but still treats 2p txn as mp when scheduling

vittvolt:
## More
- Concurrent 2p has been partially done in https://github.com/VoltDB/voltdb/compare/ENG-10944-np-txn?expand=1 will need more test in the following days.
- This should still increase tps when SP rate is high.


src/frontend/org/voltdb/compiler/ProcedureCompiler.java
```diff
@@ -493,7 +493,7 @@ public static void addPartitioningInfo(VoltCompiler compiler, Procedure procedur
                     throws VoltCompilerException {
         // parse the procedureInfo
         procedure.setSinglepartition(info.singlePartition);
-        if (!info.singlePartition) return;
+        if (!info.singlePartition && info.partitionInfo.length() == 0) return;
```
xinjiacs: You can hide this logic by adding a function in ProcInfoData.java, like `isAllPartitionProcedure()`.

// some where we need to add this comment
MP: 
   |-> AP: all partition 
   |-> NP: 2P as a special case
SP: single partition

src/frontend/org/voltdb/utils/CatalogSchemaTools.java
```diff
@@ -456,7 +456,7 @@ public static void toSchema(StringBuilder sb, Procedure proc)
         // Build the optional PARTITION clause.
         StringBuilder partitionClause = new StringBuilder();
         ProcedureAnnotation annot = (ProcedureAnnotation) proc.getAnnotation();
-        if (proc.getSinglepartition()) {
+        if (proc.getSinglepartition() || proc.getPartitionparameter2() != 0) {
```
xinjiacs: Short term fix: Add a static function in CatalogUtil.java like `isProcedurePartitioned(Procedure proc)`

I think there maybe other places that we need similar check. Adding this util function will help us refactor in future when supporting NP TXN.

tests/test_apps/nptxn-perf-benchmark/schema.sql
```diff
@@ -37,6 +37,6 @@ PARTITION PROCEDURE Authorize ON TABLE card_activity COLUMN pan PARAMETER 0;
 CREATE PROCEDURE FROM CLASS np.Redeem;
 PARTITION PROCEDURE Redeem ON TABLE card_activity COLUMN pan PARAMETER 0;
 
-CREATE PROCEDURE FROM CLASS np.Transfer;
+CREATE PROCEDURE PARTITION ON TABLE card_activity COLUMN pan AND ON TABLE card_activity COLUMN pan FROM CLASS np.Transfer;
```
vittvolt: The java benchmark on the master does not call AdHoc for initialization.

PR #4760, opened 2017-08-01
ENG-12553: Merge large temp table iterator code

wolffcm:
This moves logic for large temp table iterators into our regular TableIterator class, so that TableIterator can now iterator over three different kinds of temp tables (normal temp, large temp, and persistent).


src/ee/storage/tableiterator.h
```diff
@@ -97,99 +107,242 @@ class TableIterator : public TupleIterator {
...
+
+        TypeSpecificState()
+        : m_persBlockIterator()
+        , m_tempBlockIterator()
+        , m_largeTempBlockIterator()
+            , m_tempTableDeleteAsGo(false)
+        {
+        }
+
+        TypeSpecificState(TBMapI it)
```
BillWhite: Does it make sense to document the differences between these different constructors?  I can guess, but I might guess wrong.

src/ee/storage/tableiterator.h
```diff
@@ -97,99 +107,242 @@ class TableIterator : public TupleIterator {
...
+        , m_tempTableDeleteAsGo(deleteAsGo)
+        {
+        }
+
+        ~TypeSpecificState()
+        {
+        }
+
+        /** Table block iterator for persistent tables */
+        TBMapI m_persBlockIterator;
```
BillWhite: It probably doesn't matter but why can't these be in a union again?  I would think
```
union {
    TBMapI     m_persBlockIterator;
    std::vector<TBPtr>::iterator m_tempBlockIterator;
    std::vector<int64_t>::iterator m_largeTempBlockIterator;
};
```
Does exactly the same as listing these three.

Since this is an anonymous union, the field names are hoisted to the outer scope, which in this case is scope of the TableIterator class.  This has been a part of C since before C++.  I'm pretty sure it works reliably.  I could be wrong, though.

src/ee/storage/tableiterator.h
```diff
@@ -223,10 +389,16 @@ inline bool TableIterator::hasNext() {
 // This function should be replaced by specific iteration functions
 // when the caller knows the table type.
 inline bool TableIterator::next(TableTuple &out) {
-    if (m_tempTableIterator) {
+    switch (m_iteratorType) {
+    case TEMP:
```
BillWhite: This is probably more obscure than is necessary.  But TEMP, PERSISTENT and LARGE_TEMP could have small values, and there could be a static array of length 3 with pointer to member functions TableIterator::tempNext, TableIterator::persistenNext and TableIterator::largeTempNext.  That would avoid the switch at the cost of a table lookup.  That seems right to me.  Other than that, I think it's the same call.

src/ee/storage/tableutil.cpp
```diff
@@ -57,24 +57,60 @@
 
 namespace voltdb {
 
+class JumpingTableIterator : public TableIterator {
```
wolffcm: It's used for selecting random tuples from a table, for EE unit testing.

wolffcm: I'll add a comment.

src/ee/storage/tableiterator.h
wolffcm: Yes, that is exactly what I did originally.  The issue was that the iterator classes have non-trivial constructors.  This would work if we have more complete support for C++11, but some machines in the Jenkins pool did not like it.  I had this working on my OS X machine, but I had to settle for what you see here now.

src/ee/storage/tableiterator.h
wolffcm: Do you think it would make a big difference?  I would guess the compiler would make this transformation for me if it did?

wolffcm: I'm going to forgo an array of pointers to member functions for now, since performance looks pretty much unchanged.  If we ever need to try to make improvements, this is something we can try.

wolffcm:
Thanks for the review Bill.  I've added the comments that you requested.

PR #4762, opened 2017-08-01
ENG-12379-Log-Addr-Match-With-Stats

sycai:
If one calls system procedure from VMC, the INITIATOR stats will show correct ip address of the caller, rather than "+!_InternalAdapter_!+"

### DIFF
--- a/src/frontend/org/voltdb/InternalConnectionHandler.java
+++ b/src/frontend/org/voltdb/InternalConnectionHandler.java
@@ -102,6 +102,20 @@ public class InternalConnectionHandler {
             String procName,
             Object...args)
     {
+        return callProcedure(user, isAdmin, timeout, cb, ntPriority, backPressurePredicate, procName, null, args);
+    }
+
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/InternalConnectionHandler.java
```diff
@@ -102,6 +102,20 @@ public boolean callProcedure(
...
+
+    public boolean callProcedure(
+            AuthUser user,
+            boolean isAdmin,
+            int timeout,
+            ProcedureCallback cb,
+            boolean ntPriority,
+            Function<Integer, Boolean> backPressurePredicate,
+            String procName,
+            String hostname,
```
nshi: Nitpick: it probably makes more sense to put this parameter before `user`, instead of in the middle of transaction parameters.

sycai: Sounds good!

PR #4770, opened 2017-08-03
Eng 11877 rootuser

ndenggg:
Add a host log when root user logged in

### DIFF
--- a/lib/python/voltcli/checkconfig.py
+++ b/lib/python/voltcli/checkconfig.py
@@ -170,6 +170,12 @@ def test_python_version(output):
                 return
     output['Python'] = ["PASS", pythonVersion]
 
+def test_username(output):
+    if os.geteuid()==0:
+        output['RootUser'] = ["WARN", "Running the VoltDB server software from the system root account is not recommended"]
+    else:
+        output['RootUser'] = ["PASS", "Not running as root"]
+
 def test_thp_config(output):
     thpError = _check_thp_config()
     if thpError is not None:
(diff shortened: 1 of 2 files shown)


luyangco:
Have you tested on Linux/Mac OS?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -807,6 +807,10 @@ public void initialize(Configuration config) {
                             null);
                 }
 
+                if (System.getProperty("user.name").equals("root")) {
```
akhanzode: Please check if this returns effective user id for root #id if someone sets USERNAME what happens?
may be best done in python wrapper??

GuaGuaDean: Hi Anish, I have tried export USER=root in the command line. And yes it will trigger this log. I guess I need to look into this issue. Thanks for pointing it out.

GuaGuaDean: I tried export USER= NEWUSER in the command line. That won't change the property name in JVM.

lib/python/voltcli/checkconfig.py
```diff
@@ -223,6 +229,7 @@ def test_full_config(output):
     test_ntp(output)
     test_java_version(output)
     test_python_version(output)
+    test_username(output)
```
akhanzode: You should call this from init and start also so users know at init time and runtime.

PR #4772, opened 2017-08-07
ENG-12968 NT-Procedure memory leak fix

xinjiacs:
Use internal NT-Procedure connection handler to invoke NT-Procedures and track the call backs. Otherwise the NT-Procedure callAllNodeNTProcedure API ends up keeping all the reference of NT-Procedure internal invocations and eventually hit OOM.

### DIFF
--- a/src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java
+++ b/src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java
@@ -59,12 +59,12 @@ public class LightweightNTClientResponseAdapter implements Connection, WriteStre
 
     private final long m_connectionId;
     private final AtomicLong m_handles = new AtomicLong();
-    private final ConcurrentMap<Long, NTNestedProcedureCallback> m_callbacks = new ConcurrentHashMap<>(2048, .75f, 128);
+    private final ConcurrentMap<Long, ProcedureCallback> m_callbacks = new ConcurrentHashMap<>(2048, .75f, 128);
 
     private final InvocationDispatcher m_dispatcher;
 
     private void createTransaction(final InternalAdapterTaskAttributes kattrs,
-            final NTNestedProcedureCallback cb,
+            final ProcedureCallback cb,
             final StoredProcedureInvocation task,
             final AuthSystem.AuthUser user)
     {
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java
```diff
@@ -59,12 +59,12 @@
 
     private final long m_connectionId;
     private final AtomicLong m_handles = new AtomicLong();
-    private final ConcurrentMap<Long, NTNestedProcedureCallback> m_callbacks = new ConcurrentHashMap<>(2048, .75f, 128);
```
jhugg: Is there ever a reason the more generic, "ProcedureCallback" would be stored here?

xinjiacs: The "NTNestedProcedureCallback" works for NT-Procedure Runner invoking procedure locally. Any NT-Proc invocation of `callAllNodeNTProcedure`  will lose the Procedure Runner information on other nodes. Thus we need generic call back to handle this case.

akhanzode: Whats special about NT ClientResponseAdapter? Whats Lightweight about it?

src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java
```diff
@@ -242,14 +241,18 @@ public long connectionId() {
...
-        if (callback==null) {
+        ProcedureCallback callback = m_callbacks.get(clientHandle);
+        if (callback == null) {
             m_logger.rateLimitedLog(SUPPRESS_INTERVAL, Level.WARN, null,
                     "Could not find caller details for client handle %d. Using internal adapter level connection id", clientHandle);
             return connectionId();
         }
 
-        return callback.getConnectionId(clientHandle);
+        if (callback instanceof NTNestedProcedureCallback == false) {
```
akhanzode: Instanceof is expensive if all are going to be NTNestedProcedureCall why the check?

src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java
xinjiacs: You can read the comments:
https://github.com/VoltDB/voltdb/blob/master/src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java#L42

akhanzode: I read it after I commented...this is good.

src/frontend/org/voltdb/LightweightNTClientResponseAdapter.java
xinjiacs: The NT-Procedure called from existing NT-Procedure does not have the NTNestedProcedureCallback. You can check at: 
https://github.com/VoltDB/voltdb/blob/abf3b52342122af088bd9b5c48be62df532a9427/src/frontend/org/voltdb/ClientInterface.java#L1208

PR #4774, opened 2017-08-07
Eng partition multisp tests

dm89:


src/frontend/org/voltdb/compiler/ProcedureCompiler.java
```diff
@@ -851,53 +851,59 @@ static void compileSingleStmtProcedure(VoltCompiler compiler,
...
-                        String msg = "This procedure " + shortName + " would benefit from being partitioned, by ";
-                        String tableName = "tableName", partitionColumnName = "partitionColumnName";
-                        try {
-                            assert(partitioning.getFullColumnName() != null);
-                            String array[] = partitioning.getFullColumnName().split("\\.");
-                            tableName = array[0];
-                            partitionColumnName = array[1];
-                        } catch(Exception ex) {
-                        }
+        if (stmtNum == 0) {
```
dm89: I added lines 856-869 for error checking

src/frontend/org/voltdb/compiler/ProcedureCompiler.java
```diff
@@ -851,53 +851,59 @@ static void compileSingleStmtProcedure(VoltCompiler compiler,
...
+            throw compiler.new VoltCompilerException("Cannot create a stored procedure with no statements "
+                    + "for procedure: " + procedure.getClassname());
+        }
+
+        int paramCount = procedure.getParameters().size();
+        // parse the procinfo
+        procedure.setSinglepartition(info.singlePartition);
```
dm89: all the code from here to below is moved out of the loop after going through all the statements in the procedure (done above) - Earlier the partition info was checked once for every statement which is not correct (and not necessary)
But github shows all the code below as modified just because of the indentation.

PR #4781, opened 2017-08-09
ENG-12976.  Better SSL usability in for python client

rmorgenstein:
- Allow ssl_config_file to have environment variables or ~
  - Better error message when pyjks module is missing

### DIFF
--- a/lib/python/voltdbclient.py
+++ b/lib/python/voltdbclient.py
@@ -25,6 +25,7 @@ import struct
 import datetime
 import decimal
 import hashlib
+import os
 try:
     import ssl
     ssl_available = True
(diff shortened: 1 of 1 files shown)


lib/python/voltdbclient.py
```diff
@@ -194,6 +201,7 @@ def __init__(self, host = None, port = 21212, usessl = False,
             self.usekerberos = kerberos
         self.kerberosprinciple = None
         self.ssl_config = ssl_config
+        print ssl_config_file
```
JadenW: only for debug?

JadenW: rest look good to me.

PR #4789, opened 2017-08-10
ENG-12995: downgrade UAC log message to debug level. ENG-12954: Bette…

xinjiacs:
…r error message for class not found in the UpdateClasses jar. ENG-12845: Better error message for SQLStmt is not declared or initialized properly.

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -644,6 +644,11 @@ public class ProcedureRunner {
         if (stmt == null) {
             throw new IllegalArgumentException("SQLStmt parameter to voltQueueSQL(..) was null.");
         }
+        if (stmt.statementParamTypes == null) {
+            // reflected SQLStmt is not initialized, this means the SQLStmt is not declared as public final StmtSQL
+            throw new IllegalArgumentException("SQLStmt is not declared as final or initialized at compile time.");
+        }
+
         QueuedSQL queuedSQL = new QueuedSQL();
         queuedSQL.expectation = expectation;
         queuedSQL.params = getCleanParams(stmt, true, args);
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -111,6 +111,7 @@ public static CatalogChangeResult prepareApplicationCatalogDiff(String invocatio
             InMemoryJarfile oldJar = context.getCatalogJar().deepCopy();
             String deploymentString = operationString;
             if ("@UpdateApplicationCatalog".equals(invocationName)) {
+                compilerLog.info("@UpdateApplicationCatalog is invoked, current catalog version: " + context.catalogVersion);
```
luyangco: Should this line be DEBUG level?

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -358,7 +359,8 @@ private static InMemoryJarfile modifyCatalogClasses(Catalog catalog, InMemoryJar
             return null;
         }
 
-        compilerLog.info("Updating java classes available to stored procedures");
+        compilerLog.info("@UpdateClasses is invoked, updating java classes available to stored procedures");
```
luyangco: Should this line be DEBUG level?

xinjiacs: I think the info level is OK, as this can not be called frequent enough to flush the log like normal SP transactions.  And it can trace the invocations or calculate the UAC execution time. It's only printed on the node that receives this invocation.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -644,6 +644,11 @@ public void voltQueueSQL(final SQLStmt stmt, Expectation expectation, Object...
         if (stmt == null) {
             throw new IllegalArgumentException("SQLStmt parameter to voltQueueSQL(..) was null.");
         }
+        if (stmt.statementParamTypes == null) {
```
vittvolt: Does it affect any existing tests ? I think there are a few existing unit tests that declare without final keyword.

PR #4791, opened 2017-08-10
ENG-12732: DR consumer must add new partition receivers and update partition count on the consumer side

slowsoul:
This PR includes work for both ENG-12512 and ENG-12732. It enables elastic add to happen on DR producer cluster while DR is active. In this first phase, it cannot safely replicate producer cluster DR table changes before elastic add finishes yet, therefore it's not really useful yet as elastic add can take a long time. For now, you have to issue new transactions on DR table on producer cluster after elastic add has finished.

It handles DR_ELASTIC_CHANGE event on DR consumer side and adds partition buffer receivers for new producer partitions, it also makes sure consumer will recover partition buffer receivers and proceed correctly in case that any consumer cluster node fails during @ApplyBinaryLogMP for the DR_ELASTIC_CHANGE event.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -2188,7 +2188,9 @@ void VoltDBEngine::executeTask(TaskType taskType, ReferenceSerializeInputBE &tas
         assert(uq);
         uq->registerUndoAction(
                 new (*uq) ExecuteTaskUndoGenerateDREventAction(
-                        m_executorContext->drStream(), m_executorContext->drReplicatedStream(), type, lastCommittedSpHandle,
+                        m_executorContext->drStream(), m_executorContext->drReplicatedStream(),
+                        m_executorContext->m_partitionId,
+                        type, lastCommittedSpHandle,
                         spHandle, uniqueId, payloads));
         break;
     }
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -322,7 +322,7 @@ void PersistentTable::deleteAllTuples(bool, bool fallible) {
 }
 
 bool PersistentTable::doDRActions(AbstractDRTupleStream* drStream) {
-    return (m_drEnabled && drStream && drStream->drStreamStarted());
+    return m_drEnabled && !m_isMaterialized && drStream && drStream->drStreamStarted();
 }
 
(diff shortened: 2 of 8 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -322,7 +322,7 @@ void PersistentTable::deleteAllTuples(bool, bool fallible) {
 }
 
 bool PersistentTable::doDRActions(AbstractDRTupleStream* drStream) {
-    return (m_drEnabled && drStream && drStream->drStreamStarted());
+    return (drStream && drStream->drStreamStarted() && m_drEnabled && !m_isMaterialized);
```
JadenW: Why change the checking order?
Previous we didn't check m_isMaterialized, so we create unnecessary undo actions for materialized views?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -2069,15 +2069,25 @@ void VoltDBEngine::executeTask(TaskType taskType, ReferenceSerializeInputBE &tas
...
-        if (type == DR_STREAM_START || m_executorContext->drStream()->drStreamStarted()) {
-            m_executorContext->drStream()->generateDREvent(type, lastCommittedSpHandle,
-                                                           spHandle, uniqueId, payloads);
-        }
         if (m_executorContext->drReplicatedStream() &&
             (type == DR_STREAM_START || m_executorContext->drReplicatedStream()->drStreamStarted())) {
             m_executorContext->drReplicatedStream()->generateDREvent(type, lastCommittedSpHandle,
                                                                      spHandle, uniqueId, payloads);
         }
+        if (type == DR_ELASTIC_CHANGE) {
```
JadenW: nitpick: put a TODO so we can simplify this when your remove replication stream got merged.

src/ee/storage/persistenttable.cpp
```diff
@@ -322,7 +322,7 @@ void PersistentTable::deleteAllTuples(bool, bool fallible) {
 }
 
 bool PersistentTable::doDRActions(AbstractDRTupleStream* drStream) {
-    return (m_drEnabled && drStream && drStream->drStreamStarted());
+    return m_drEnabled && !m_isMaterialized && drStream && drStream->drStreamStarted();
```
manjujames: drEnabled is initialized by checking m_isMaterialized as well. It was intentional to remove one condition in the 'if' that happens over and over again.

PR #4793, opened 2017-08-14
ENG-12985: Delta tables used for joins in materialized views should never generate DR records

slowsoul:
The delta table used in joins in materialized views used to inherit the "DR" catalog attribute from its source table. If the source table is a DR table, so will the delta table, which essentially records all operations on delta table into the DR stream, which is undesirable because 1) it breaks DR binary log structural constraint in case of snapshot restore and 2) it causes the consumer cluster to diverge from the producer cluster.

To fix this, delta tables should always be set as non-DR tables.

### DIFF
--- a/src/ee/storage/TableCatalogDelegate.cpp
+++ b/src/ee/storage/TableCatalogDelegate.cpp
@@ -73,13 +73,12 @@ Table* TableCatalogDelegate::getTable() const {
 }
 
 TupleSchema* TableCatalogDelegate::createTupleSchema(catalog::Table const& catalogTable,
-                                                     bool isXDCR,
-                                                     bool forceNoDR) {
+                                                     bool isXDCR) {
     // Columns:
     // Column is stored as map<std::string, Column*> in Catalog. We have to
     // sort it by Column index to preserve column order.
     auto numColumns = catalogTable.columns().size();
-    bool needsDRTimestamp = isXDCR && !forceNoDR && catalogTable.isDRed();
+    bool needsDRTimestamp = isXDCR && catalogTable.isDRed();
     TupleSchemaBuilder schemaBuilder(numColumns,
                                      needsDRTimestamp ? 1 : 0); // number of hidden columns
 
@@ -281,7 +280,7 @@ Table* TableCatalogDelegate::constructTableFromCatalog(catalog::Database const& 
     }
 
     // get the schema for the table
-    TupleSchema* schema = createTupleSchema(catalogTable, isXDCR, forceNoDR);
+    TupleSchema* schema = createTupleSchema(catalogTable, isXDCR);
 
     // Indexes
     std::map<std::string, TableIndexScheme> index_map;
@@ -448,7 +447,7 @@ PersistentTable* TableCatalogDelegate::createDeltaTable(catalog::Database const&
     // Delta table will only have one row (currently).
     // Set the table block size to 64KB to achieve better space efficiency.
     // FYI: maximum column count = 1024, largest fixed length data type is short varchars (64 bytes)
-    Table* deltaTable = constructTableFromCatalog(catalogDatabase, catalogTable, false, 1024 * 64, true);
+    Table* deltaTable = constructTableFromCatalog(catalogDatabase, catalogTable, false, true, 1024 * 64);
     deltaTable->incrementRefcount();
     // We have the restriction that view on joined table cannot have non-persistent table source.
     // So here we could use static_cast. But if we in the future want to lift this limitation,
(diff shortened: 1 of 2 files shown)


src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -73,12 +73,13 @@ Table* TableCatalogDelegate::getTable() const {
...
 TupleSchema* TableCatalogDelegate::createTupleSchema(catalog::Table const& catalogTable,
-                                                     bool isXDCR) {
+                                                     bool isXDCR,
+                                                     bool forceNoDR) {
     // Columns:
     // Column is stored as map<std::string, Column*> in Catalog. We have to
     // sort it by Column index to preserve column order.
     auto numColumns = catalogTable.columns().size();
-    bool needsDRTimestamp = isXDCR && catalogTable.isDRed();
+    bool needsDRTimestamp = isXDCR && !forceNoDR && catalogTable.isDRed();
```
yzhang1991: When `createDeltaTable()` called `createTupleSchema()`, the `isXDCR` flag is set to false:
```c++
Table* deltaTable = constructTableFromCatalog(catalogDatabase, catalogTable, false, 1024 * 64);
```
The `needsDRTimestamp` is always false for the delta table case.
So `forceNoDR` is not needed here, I think you can remove it from this function.

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -279,7 +281,7 @@ Table* TableCatalogDelegate::constructTableFromCatalog(catalog::Database const&
     }
 
     // get the schema for the table
-    TupleSchema* schema = createTupleSchema(catalogTable, isXDCR);
+    TupleSchema* schema = createTupleSchema(catalogTable, isXDCR, forceNoDR);
```
yzhang1991: Remove the `forceNoDR` flag here.

src/ee/storage/TableCatalogDelegate.hpp
```diff
@@ -144,6 +145,7 @@ class TableCatalogDelegate {
     Table *constructTableFromCatalog(catalog::Database const &catalogDatabase,
                                      catalog::Table const &catalogTable,
                                      bool isXDCR,
+                                     bool forceNoDR = false,
```
yzhang1991: Remove forceNoDR flag here

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -369,7 +371,7 @@ Table* TableCatalogDelegate::constructTableFromCatalog(catalog::Database const&
 
     bool exportEnabled = isExportEnabledForTable(catalogDatabase, tableId);
     bool tableIsExportOnly = isTableExportOnly(catalogDatabase, tableId);
-    bool drEnabled = catalogTable.isDRed();
+    bool drEnabled = !forceNoDR && catalogTable.isDRed();
```
yzhang1991: This is the only place that we need the `forceNoDR` flag.

src/ee/storage/TableCatalogDelegate.cpp
slowsoul: I see, I will remove it.

src/ee/storage/TableCatalogDelegate.hpp
slowsoul: I think `forceNoDR` is still needed in `constructTableFromCatalog()`, right? Did you mean to add this comment on the declaration of `createTupleSchema()`

src/ee/storage/TableCatalogDelegate.hpp
```diff
@@ -144,7 +144,8 @@ class TableCatalogDelegate {
     Table *constructTableFromCatalog(catalog::Database const &catalogDatabase,
```
yzhang1991: Please add a comment here about what `forceNoDR` is and where we use it (Now only the delta table creation will use it).

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -446,7 +447,7 @@ PersistentTable* TableCatalogDelegate::createDeltaTable(catalog::Database const&
     // Delta table will only have one row (currently).
     // Set the table block size to 64KB to achieve better space efficiency.
     // FYI: maximum column count = 1024, largest fixed length data type is short varchars (64 bytes)
-    Table* deltaTable = constructTableFromCatalog(catalogDatabase, catalogTable, false, 1024 * 64);
```
yzhang1991: Add a comment here that we will force the construction of the table to have no DR.

PR #4796, opened 2017-08-14
ENG-13005

akhanzode:
Add security on /catalog and remove /ddl as VMC grabs ddl from DOM.
Add test.


src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -326,32 +325,20 @@ public void handle(String target,
 
             super.handle(target, baseRequest, request, response);
             if (baseRequest.isHandled()) return;
+            //jsonp is specified when response is expected to go to javascript function.
+            String jsonp = request.getParameter(HTTPClientInterface.JSONP);
+            AuthenticationResult authResult = authenticate(baseRequest);
+            if (!authResult.isAuthenticated()) {
+                response.getWriter().print(buildClientResponse(jsonp, ClientResponse.UNEXPECTED_FAILURE, authResult.m_message));
+                response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
```
zeemanhe: Should the status be set before the response prints? Does the order depends on the implementation of App servers?

akhanzode: It does not matter until request is set as Handled jetty wont write back. We use this all over the places.

zeemanhe: great

PR #4797, opened 2017-08-15
ENG-13006: Correctly handle statement splitting for sqlcmd batches

wolffcm:
For `voltdb init --schema=ddl.sql` we use a sqlcmd inline batch under the hood.  The changes that support multi-statement procedures tweaked sqlcmd's logic in a subtle way and made it so we weren't trimming comments appearing in between DDL statements in batches.

The test doesn't actually reproduce the issue, since sqlcmd will use a different codepath when invoked from the command line.  I wasn't able to reproduce it outside of `voltdb init`.  But at least it gets us more coverage.  

I did some refactoring of SplitStmtResults to make it an immutable class.  The functional changes are all in SQLCommand.java.

### DIFF
--- a/src/frontend/org/voltdb/utils/SQLCommand.java
+++ b/src/frontend/org/voltdb/utils/SQLCommand.java
@@ -811,7 +811,7 @@ public class SQLCommand
                         statementStarted = false;
                     }
                 }
-                else { // not in a batch:
+                else { // when in a batch:
                     SplitStmtResults splitResults = SQLLexer.splitStatements(statementString);
                     if (splitResults.getIncompleteStmt() == null) {
                         // not in the middle of a statement.
--- a/src/frontend/org/voltdb/utils/SQLCommand.java
+++ b/src/frontend/org/voltdb/utils/SQLCommand.java
@@ -30,7 +30,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
-import java.io.StringReader;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.UnknownHostException;
@@ -793,8 +792,8 @@ public class SQLCommand
             // Check if the current statement ends here and now.
             // if it is an incomplete multi statement procedure, it is returned back
             if (SQLParser.isSemiColonTerminated(line)) {
+                String statementString = statement.toString();
                 if (batch == null) {
-                    String statementString = statement.toString();
                     // Trim here avoids a "missing statement" error from adhoc in an edge case
                     // like a blank line from stdin.
                     if ( ! statementString.trim().isEmpty()) {
@@ -809,6 +808,13 @@ public class SQLCommand
                         }
                     } else {
                         statement.setLength(0);
+                        statementStarted = false;
+                    }
+                }
+                else { // not in a batch:
+                    SplitStmtResults splitResults = SQLLexer.splitStatements(statementString);
+                    if (splitResults.getIncompleteStmt() == null) {
+                        // not in the middle of a statement.
                         statementStarted = false;
                     }
                 }
@@ -830,11 +836,11 @@ public class SQLCommand
     private static String executeStatements(String statements, DDLParserCallback callback, int lineNum)
     {
         SplitStmtResults parsedOutput = SQLLexer.splitStatements(statements);
-        List<String> parsedStatements = parsedOutput.completelyParsedStmts;
+        List<String> parsedStatements = parsedOutput.getCompletelyParsedStmts();
         for (String statement: parsedStatements) {
             executeStatement(statement, callback, lineNum);
         }
-        return parsedOutput.incompleteMuliStmtProc;
+        return parsedOutput.getIncompleteStmt();
     }
 
     private static void executeStatement(String statement, DDLParserCallback callback, int lineNum)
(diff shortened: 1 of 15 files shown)


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -708,24 +707,21 @@ public static void executeScriptFromReader(FileInfo fileInfo, SQLCommandLineRead
...
-                        // For now, treat the final semicolon as optional and
-                        // assume that we are not just adding a partial statement to the batch.
-                        batch.append(statement);
+                }
+                else {
+                    batch.append(statement);
+                    if (batch.length() > 0) {
                         executeDDLBatch(fileInfo.getFilePath(), batch.toString(), callback, reader.getLineNumber());
                     }
                 }
```
yzhang1991: Line 732, it seems that we now use
```
statementString.trim().isEmpty()
```
instead of
```
line.trim().equals("")
```

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -793,8 +790,8 @@ public static void executeScriptFromReader(FileInfo fileInfo, SQLCommandLineRead
             // Check if the current statement ends here and now.
             // if it is an incomplete multi statement procedure, it is returned back
             if (SQLParser.isSemiColonTerminated(line)) {
+                String statementString = statement.toString();
                 if (batch == null) {
-                    String statementString = statement.toString();
                     // Trim here avoids a "missing statement" error from adhoc in an edge case
                     // like a blank line from stdin.
                     if ( ! statementString.trim().isEmpty()) {
```
yzhang1991: Line 799, can we assign a better name to `incompleteSt`? Maybe `incompleteStmt`?

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -812,6 +809,16 @@ public static void executeScriptFromReader(FileInfo fileInfo, SQLCommandLineRead
...
                     }
                 }
+                else { // when in a batch:
+                    SplitStmtResults splitResults = SQLLexer.splitStatements(statementString);
+                    if (splitResults.getIncompleteStmt() == null) {
+                        // not in the middle of a statement.
+                        statementStarted = false;
+                    }
+
+                    batch.append(statement);
```
yzhang1991: So this is the actual fix for the logic. If we do not call `splitStatements()` here, the comments won't be removed and we will keep accumulating statements in the batch string builder until reaching the end of the file/batch. Am I right?

src/frontend/org/voltdb/utils/SQLCommand.java
wolffcm: I'm unclear on what you mean.  Are you just saying that `isEmpty()` is clearer and preferable to `equals("")`?  I will change it to `isEmpty()`.

yzhang1991: Yes, I personally prefer the `isEmpty()` way.

src/frontend/org/voltdb/utils/SQLCommand.java
wolffcm: You are right.  I should have explained in the pull request header comment.  We were not setting `statementStarted` back to false after reading a complete statement, and this flag is what controls removing comments.  Actually it seems we could remove whole line comments everywhere, but I was trying to be conservative with my changes.

But looking over these changes now I think that we would incorrectly detect the end of a statement in a multi-statement procedure.  I will rework the logic a little more, and request another look from you.

tests/frontend/org/voltdb/utils/TestSplitSQLStatements.java
```diff
@@ -135,14 +135,14 @@ public void testProcSQLSplit() {
         // there is no END statement for BEGIN, so the ; is included as the parsing of BEGIN is not complete
         sql = "CREATE PROCEDURE foo AS BEGIN SELECT * from t; SELECT * from t;";
         parsedOut = SQLLexer.splitStatements(sql);
-        assertEquals(0, parsedOut.completelyParsedStmts.size());
-        assertEquals(sql, parsedOut.incompleteMuliStmtProc);
+        assertEquals(0, parsedOut.getCompletelyParsedStmts().size());
```
akhanzode: You should check and add test if
------- PROCEDURES -------- CREATE PROCEDURE 
works or not looks like those fail too (any voltdb specific syntax with comment before.)

akhanzode:
Also add more tests in TestInitStartAction

akhanzode:
Looks good if jenkins is happy.

src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -796,9 +796,9 @@ public static void executeScriptFromReader(FileInfo fileInfo, SQLCommandLineRead
                     // like a blank line from stdin.
```
yzhang1991: Here, the `line` is appended to `statement`(line 788), and `line` is terminated by semicolons (the regex was defined as ";+"). How statementString.trim().isEmpty() can equal to `true`(line 807)?

wolffcm: Yes, you are right.  The string will never be empty there.  It was like that before too.  I will fix it.

wolffcm:
Closing this pull request because work was merged on a different branch.

PR #4798, opened 2017-08-15
ENG-13015: add log message to report exact block time for @UpdateCore…

xinjiacs:
…. The block time starts recording when issuing first MP fragment to other SP sites, ends on receving the last fragment aggregated response. Also update the blocking time calculation for the UAC benchmark app.


xinjiacs:
Reviewed by Anish, test fixed in PRO.

PR #4799, opened 2017-08-15
ENG-12980: This change to VoltCompiler is to allow auto gen ddl as th…

xinjiacs:
…e single source of schema during catalog jar upgrade path, other .sql files will be ignored. We used to search all the .sql files in the jar and compile them all. This behavior is fine if VoltDB has full control what will be inclueded in the jar. However, in v7.4, UpdateClasses allow user to load other source files not limited to .class file only. The user loaded .sql files as resource file for their procedure when doing UpdateClasses, and our compiler compiled these additional sql files during jar upgrade path, throwing exception if loading bad duplicated ddl source file.


src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -2012,14 +2012,20 @@ public String upgradeCatalogAsNeeded(InMemoryJarfile outputJar)
             Entry<String, byte[]> entry = outputJar.firstEntry();
             while (entry != null) {
                 String path = entry.getKey();
-                //TODO: It would be better to have a manifest that explicitly lists
-                // ddl files instead of using a brute force *.sql glob.
-                if (path.toLowerCase().endsWith(".sql")) {
+                // ENG-12980: only look for auto-gen.ddl on root directory
+                if (AUTOGEN_DDL_FILE_NAME.equals(path.toLowerCase())) {
```
zeemanhe: may use equalsIgnoreCase() instead of equals() without converting path to lower case.  equalsIgnoreCase() is generally more efficient than updating cases.

PR #4800, opened 2017-08-15
ENG-12973 UDF buffer resizing

yzhang1991:
This pull request enables the UDF buffer to resize when needed.
The resizing could be the result of having larger-than-capacity parameters or return values.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -441,6 +441,15 @@ int VoltDBEngine::executePlanFragments(int32_t numFragments,
 
     m_currentIndexInBatch = -1;
 
+    // If we were expanding the UDF buffer too much, shrink it back a little bit.
+    // We check this at the end of every batch execution. So we won't resize the buffer
+    // too frequently if most of the workload in the same batch requires a much larger buffer.
+    // We initiate the resizing work in EE because this is a common place where
+    // both single-partition and multi-partition transactions can get.
+    if (m_udfBufferCapacity > MAX_UDF_BUFFER_SIZE) {
+        m_topend->resizeUDFBuffer(MAX_UDF_BUFFER_SIZE);
+    }
+
     return failures;
 }
 
@@ -564,8 +573,6 @@ UniqueTempTableResult VoltDBEngine::executePlanFragment(ExecutorVector* executor
 }
 
 NValue VoltDBEngine::callJavaUserDefinedFunction(int32_t functionId, std::vector<NValue>& arguments) {
-    resetUDFOutputBuffer();
-
     UserDefinedFunctionInfo *info = findInMapOrNull(functionId, m_functionInfo);
     if (info == NULL) {
         // There must be serious inconsistency in the catalog if this could happen.
@@ -578,6 +585,10 @@ NValue VoltDBEngine::callJavaUserDefinedFunction(int32_t functionId, std::vector
     //   * parameters.
     size_t bufferSizeNeeded = sizeof(int32_t); // size of the function id.
     for (int i = 0; i < arguments.size(); i++) {
+        // It is very common that the argument we are going to pass is in
+        // a compatible data type which does not exactly match the type that
+        // is defined in the function.
+        // We need to cast it to the target data type before the serialization.
         arguments[i] = arguments[i].castAs(info->paramTypes[i]);
         bufferSizeNeeded += arguments[i].serializedSize();
     }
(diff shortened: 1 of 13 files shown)


src/frontend/org/voltdb/jni/ExecutionEngineJNI.java
```diff
@@ -260,17 +262,6 @@ final void clearPerFragmentStatsAndEnsureCapacity(int batchSize) {
         }
     }
 
-    final void clearUDFBufferAndEnsureCapacity(int size) {
```
yzhang1991: This is not used.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -584,22 +584,31 @@ NValue VoltDBEngine::callJavaUserDefinedFunction(int32_t functionId, std::vector
 
     // Check buffer size here.
     // Adjust the buffer size when needed.
+    // Note that bufferSizeNeeded does not include its own size.
+    // So we are testing bufferSizeNeeded + sizeof(int32_t) here.
+    if (bufferSizeNeeded + sizeof(int32_t) > m_udfBufferCapacity) {
+        m_topend->resizeUDFBuffer(bufferSizeNeeded + sizeof(int32_t));
```
wolffcm: So we never shrink the buffer?  Grow it once and it stays grown?

tests/frontend/org/voltdb/regressionsuites/TestUserDefinedFunctions.java
```diff
@@ -391,6 +391,36 @@ public void testAdd2GeographyPoint() throws IOException, ProcCallException {
...
+        // Call the callAdd2Varbinary function with two 256K byte[] parameters.
+        // This will greatly enlarge the size of the UDF buffer to more than 512K.
+        byte[] largeByteArray = new byte[256 * 1024];
+        cr = client.callProcedure("callAdd2Varbinary", largeByteArray, largeByteArray);
+        assertEquals("callAdd2Varbinary failed", ClientResponse.SUCCESS, cr.getStatus());
+
+        // Truncate the table afterwards.
+        cr = client.callProcedure("@AdHoc", truncateStatement);
+        assertEquals("Truncation of table R1 failed", ClientResponse.SUCCESS, cr.getStatus());
+    }
```
wolffcm: We should also have a test that will grow the UDF buffer to its largest possible size.  Also a test that will attempt to return a value that is too large entirely, and verify that a reasonable error is returned and that the server does not crash.  What happens if a user tries to return a VARCHAR value that is 900MB large?

You might run afoul of the network message size limit, but you could work around this because the value produced by a UDF doesn't need to actually be returned to the client (for example, use a UDF in a WHERE clause).

src/frontend/org/voltdb/jni/ExecutionEngineJNI.java
```diff
@@ -789,9 +797,9 @@ else if (returnValue instanceof Byte[]) {
                     break;
                 default:
                 }
-                // resizeUDFBuffer() function will compare the required size and the actual size,
-                // then take action if needed.
-                resizeUDFBuffer(sizeRequired);
+                if (sizeRequired > m_udfBuffer.capacity()) {
```
yzhang1991: `resizeUDFBuffer()` does not do the check anymore.
This is because VoltDBEngine can reset it at the end of every batch.
Enforcing the desired size is greater than the capacity in the `resizeUDFBuffer()` will complicate the problem.

src/frontend/org/voltdb/jni/ExecutionEngineJNI.java
```diff
@@ -809,15 +817,15 @@ else if (returnValue instanceof Byte[]) {
         }
         // Getting here means the execution was not successful.
         try {
-            if (throwable != null) {
```
yzhang1991: `throwable` cannot be null if we can get here.

src/frontend/org/voltdb/jni/ExecutionEngineJNI.java
```diff
@@ -90,7 +92,8 @@
 
     /** Create a ByteBuffer (in a container) for serializing arguments to C++. Use a direct
     ByteBuffer as it will be passed directly to the C++ code. */
-    private static final int MAX_PSETBUFFER_SIZE = 50 * 1024 * 1024; // 50MB
+    // This matches MAX_UDF_BUFFER_SIZE in VoltDBEngine.h
+    private static final int MAX_BUFFER_SIZE = 50 * 1024 * 1024; // 50MB
```
wolffcm: This seems like something of a misnomer because the buffer can actually be bigger than this, if only temporarily.

PR #4802, opened 2017-08-15
Eng 12420

GuaGuaDean:
Making voltadmin status command topology aware under continuous mode. This period command works during an online upgrade when any node loses connection.

### DIFF
--- a/lib/python/voltcli/runner.py
+++ b/lib/python/voltcli/runner.py
@@ -392,7 +392,7 @@ class VerbRunner(object):
             args2 = [verb_name] + list(args[1:])
             self._run_command(verbspace, *args2, **kwargs)
 
-    def call_proc(self, sysproc_name, types, args, check_status=True, timeout=None):
+    def call_proc(self, sysproc_name, types, args, check_status=True, timeout=None, exception_on_failure=False):
         if self.client is None:
             utility.abort('Command is not set up as a client.',
                           'Add an appropriate admin or client bundle to @VOLT.Command().')
@@ -400,7 +400,10 @@ class VerbRunner(object):
         proc = voltdbclient.VoltProcedure(self.client, sysproc_name, types)
         response = proc.call(params=args, timeout=timeout)
         if check_status and response.status != 1:
-            utility.abort('"%s" procedure call failed.' % sysproc_name, (response,))
+            if exception_on_failure:
+                raise Exception('"%s" procedure call failed.' % sysproc_name)
+            else:
+                utility.abort('"%s" procedure call failed.' % sysproc_name, (response,))
         utility.verbose_info(response)
         return utility.VoltResponseWrapper(response)
 
--- a/lib/python/voltcli/voltadmin.d/status.py
+++ b/lib/python/voltcli/voltadmin.d/status.py
@@ -28,7 +28,6 @@ from voltcli.checkstats import StatisticsProcedureException
 
 RELEASE_MAJOR_VERSION = 7
 RELEASE_MINOR_VERSION = 2
-available_hosts = []
 
 @VOLT.Command(
     bundles=VOLT.AdminBundle(),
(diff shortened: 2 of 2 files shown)


lib/python/voltcli/voltadmin.d/status.py
```diff
@@ -28,6 +28,8 @@
 
 RELEASE_MAJOR_VERSION = 7
 RELEASE_MINOR_VERSION = 2
+available_hosts = []
+remote_clusters = {}
```
luyangco: Looks like `remote_clusters` is unused.

lib/python/voltcli/voltadmin.d/status.py
```diff
@@ -46,22 +48,39 @@ def status(runner):
...
-    if runner.client.host != runner.opts.host:
-        runner.voltdb_connect(runner.opts.host.host,
-                              runner.opts.host.port,
-                              runner.opts.username,
-                              runner.opts.password,
-                              runner.opts.ssl_config)
-
-    clusterInfo = getClusterInfo(runner)
+    # the cluster(host) which voltadmin is running on always comes first
+    global available_hosts
```
luyangco: When do you clean the elements in `available_hosts`? If you don't do that, every 2s the same number of hosts will be added to `available_hosts`.

lib/python/voltcli/voltadmin.d/status.py
```diff
@@ -88,15 +107,22 @@ def doStatus(runner):
                     pass  # ignore it
 
 def getClusterInfo(runner):
+    global available_hosts
```
luyangco: You don't need those `global` declarations, definition in line 31 is enough, you can use the variable globally in this file.

lib/python/voltcli/voltadmin.d/status.py
```diff
@@ -46,22 +48,39 @@ def status(runner):
...
+                runner.__voltdb_connect__(hostname,
+                                  runner.opts.host.port,
+                                  runner.opts.username,
+                                  runner.opts.password,
+                                  runner.opts.ssl_config)
+                clusterInfo = getClusterInfo(runner)
+                if clusterInfo != None:
+                    break
+            except:
+                pass
```
luyangco: Every failure means you need to remove that host from `available_hosts`

lib/python/voltcli/voltadmin.d/status.py
```diff
@@ -46,22 +48,39 @@ def status(runner):
...
+    global available_hosts
+    if len(available_hosts) == 0:
+        if runner.client.host != runner.opts.host.host:
+            runner.__voltdb_connect__(runner.opts.host.host,
+                                  runner.opts.host.port,
+                                  runner.opts.username,
+                                  runner.opts.password,
+                                  runner.opts.ssl_config)
+        clusterInfo = getClusterInfo(runner)
+        available_hosts.append(runner.opts.host.host)
```
luyangco: You don't need to treat first connected machine specially, every following `getClusterInfo()` will fill it up.

lib/python/voltcli/voltadmin.d/status.py
GuaGuaDean: I have a conditional if statement to avoid existing item being added again, in Line 122

lib/python/voltcli/voltadmin.d/status.py
```diff
@@ -46,22 +48,39 @@ def status(runner):
...
+    else:
+        for hostname in available_hosts:
```
GuaGuaDean: It will try every node in the map until the end, nothing will be printed in this case

lib/python/voltcli/runner.py
```diff
@@ -404,6 +404,18 @@ def call_proc(self, sysproc_name, types, args, check_status=True, timeout=None):
         utility.verbose_info(response)
         return utility.VoltResponseWrapper(response)
 
+    def __call_proc__(self, sysproc_name, types, args, check_status=True, timeout=None):
```
xinjiacs: Reuse the "call_proc" by adding another default parameter?

PR #4742, opened 2017-07-27
Eng 12040 proc runner upd cls

xinjiacs:
More operations removed from UpdateCore/UpdateClasses blocking path for performance improvement, they include:
1. Move ZK catalog reads out of transactional path
2. Move CatalogContext update preparation out of transactional path
3. Move user procedure runner creation, classes reflection, etc, out of transactional path
4. Remove catalog context change debug logging
5. Skip NT procedure classes instance creation check

PR #4804, opened 2017-08-16
ENG-13018: Giving a better error message running update classes when …

xinjiacs:
…having node failures. Also use info level to log the run time error during this proces. Fix a bug introduced in v7.6 #4742 with commit (9a7caafd0d4625072c4f374b2d80642acaafe163) to compare the right deployment hash array.


luyangco:
Looks good to me.

PR #4808, opened 2017-08-17
Added several new sqlcmdtest files (basic_create_drop_udf.in,

johnTcrawford:
simple_update_classes_udf.in, update_classes_udf.in, dml_udf.in, and
associated files, in a new /udf/ subdirectory, including changes to
build_jar.sh).

UserDefinedTestFunctions.java: tweaked to use a new boolean variable,
USE_ALTERNATIVE_VERSION, which, when true (it is normally false), causes
a bunch of the UDF's to return different results (e.g. subtract rather
than add two numbers); also deleted the main method; and added a
testfuncs_alternative.jar file containing the alternative version of
UserDefinedTestFunctions (with USE_ALTERNATIVE_VERSION = true).

TestUserDefinedFunctions.java: just corrected a typo in a JUnit test
name.


tests/sqlcmd/functions/org/voltdb_testfuncs/UserDefinedTestFunctions.java
```diff
@@ -0,0 +1,1116 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package org.voltdb_testfuncs;
```
wolffcm: I guess this file is exactly the same as the other file with the same name in `tests/testfuncs`?  It would be good if there was a single copy in source control.  It could be confusing if they got out of sync.

wolffcm:
Another comment on this pull request: it would be good to automate the steps that generate the alternative jar somehow.  If we ever want to extend the test, or modify the contents of the jar, it will be a bit of a hassle (not that bad, I realize) to regenerate it.  If creating the jar was part of the build, or part of the test set up for sqlcmdtest, that would be better.

johnTcrawford:
Thanks, Chris. Since this stuff has already merged to master, I've created a new branch, ENG-12856-udf-test-review, to address your (very good) suggestions.

PR #4809, opened 2017-08-17
ENG-13032: Add timeout setting for run everywhere NT-Procedure @Verif…

xinjiacs:
…yCatalogAndWriteJar and abort the catalog update before invoking the transactional procedure @UpdateCore. Thus it will remove the UAC blocker on ZK node and other rejoin node will be resumed to proceed. Without this change, the run everywhere NT-Procedure API may send invocation to rejoining node whose client interface has not been setted up yet, thus the message is lost and no response will be returned to the primary NT thread.

### DIFF
--- a/src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
+++ b/src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
@@ -23,6 +23,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.zookeeper_voltpatches.KeeperException;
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/sysprocs/VerifyCatalogAndWriteJar.java
```diff
@@ -45,7 +46,41 @@
...
+
+        if (timeoutEnvString == null) {
+            return timeoutValue;
+        }
+
+        try {
+            timeoutValue = Long.parseLong(timeoutEnvString);
+        } catch (NumberFormatException ex) {
+            VoltDB.crashLocalVoltDB("Invalid system environment setting for "
+                        + NTProcedureService.NTPROCEDURE_RUN_EVERYWHERE_TIMEOUT
```
zeemanhe: Would it be better to use default 5 min if the input is invalid?

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -411,10 +413,16 @@ protected String verifyAndWriteCatalogJar(CatalogChangeResult ccr)
...
+        long timeoutSetting = VerifyCatalogAndWriteJar.TIMEOUT;
         try {
-            resultMapByHost = cf.get();
+            resultMapByHost = cf.get(timeoutSetting, TimeUnit.SECONDS);
         } catch (InterruptedException | ExecutionException e) {
-            err = "An invocation of procedure " + procedureName + " on all hosts failed: " + e.getMessage();
+            err = procedureName + " run everywhere call failed: " + e.getMessage();
+            hostLog.info(err);
+            return err;
+        } catch (TimeoutException e) {
```
zeemanhe: Could aggregate the exception with the above catch.

src/frontend/org/voltdb/sysprocs/VerifyCatalogAndWriteJar.java
xinjiacs: I do not think so. If the user intentionally set the environment variable, he probably knows this hidden option and want to use it. This guards fails on VoltDB startup, so crashing is not that bad.

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
xinjiacs: Yes, but the error message is not the same.

PR #4810, opened 2017-08-17
Eng 12852 support concurrent NT procedures

luyangco:
Pro Side: https://github.com/VoltDB/pro/pull/2018


luyangco:
When node failures transaction could be lost, but on each node a callback `removeHandlesForPartitionAndInitiator()` will be called to clean corresponding bucket.

PR #4813, opened 2017-08-18
Eng 12472 Take dependences into account for UDFs.

BillWhite:
When a UDF is dropped, it may invalidate stored procedures which use it.  We also want to preclude the use of UDFs in indexes and materialized views.  This commit fixes both of these.

Jenkins is mostly blue, but not completely.  There is a memory leak in VoltDBEngine.cpp.  I wanted to make this reviewable early, though.  Perhaps Ethan can look at the memory leak, since it looks like it predates this change.

The memory leak is in the memcheck job, not the memcheck-lite job.  This is not enabled by default.


src/catgen/catalog.py
```diff
@@ -542,18 +542,24 @@ def gencpp( classes, javaOnlyClasses, prepath, postpath ):
                 write ( interp( '$ftype $clsname::$pubname() const {\n    return $privname;\n}\n', locals() ) )
 
 #
+# Set this to true to start debugging.
```
BillWhite: This change, and the change to parser.py below, fix comments in spec.txt.

I wanted to make a change in the order in which catalog commands were output, but I never got around to it.  We will need to output functions before indexes and MVs if we allow those to use UDFs.  But for now it is not crucial.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1173,6 +1180,13 @@ bool VoltDBEngine::processCatalogAdditions(bool isStreamUpdate, int64_t timestam
             int key = std::stoi(iter->first);
             info->paramTypes[key] = (ValueType)iter->second->parameterType();
         }
+        VOLT_DEBUG("UDFCAT: Adding function id %d", catalogFunction->functionId());
```
BillWhite: Maybe the VOLT_DEBUG above could be moved into the conditional compilation on the next line.  It doesn't matter all that much, as they will either both be included or neither be included at compile time.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -2191,14 +2356,14 @@ Statement getCachedStatement(String keyPrefix, String sql) {
         // check that no underlying tables have been modified since the proc had been compiled
         String[] tablesTouched = candidate.getTablesread().split(",");
         for (String tableName : tablesTouched) {
-            if (m_dirtyTables.contains(tableName.toLowerCase())) {
+            if (isDirtyTable(tableName)) {
```
BillWhite: If we drop a UDF we set a flag that makes all tables dirty, so that all procedures are recompiled.  This function tests both the tableName and this new flag.

src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
```diff
@@ -198,24 +206,38 @@ protected boolean processStatement(DDLStatement ddlStatement, Database db, DdlPr
...
-        }
-
-        Function func = db.getFunctions().add(functionName);
-        func.setFunctionid(functionId);
-        func.setFunctionname(functionName);
-        func.setClassname(className);
-        func.setMethodname(methodName);
-        CatalogMap<FunctionParameter> parameters = func.getParameters();
+        // Add the description of the function to the VoltXMLElement
+        // in m_schema.  Don't assign the function id now.  We'll do that
```
BillWhite: Actually, this comment is not correct.  We do assign the function id now, since we define this in FunctionForVoltDB.

I'll fix this.

src/frontend/org/voltdb/compiler/statements/DropFunction.java
```diff
@@ -49,19 +74,14 @@ protected boolean processStatement(DDLStatement ddlStatement, Database db, DdlPr
...
-        else {
-            if (! ifExists) {
+        if ( ! removeUDFInSchema(functionName)) {
+            if ( ! ifExists ) {
                 throw m_compiler.new VoltCompilerException(String.format(
                         "Function name \"%s\" in DROP FUNCTION statement does not exist.",
                         functionName));
             }
         }
+        FunctionForVoltDB.deregisterUserDefinedFunction(functionName);
```
BillWhite: If the function doesn't exist, say if we had "if exists" supplied, this will do nothing.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionSQL.java
```diff
@@ -2312,24 +2312,24 @@ VoltXMLElement voltAnnotateFunctionXML(VoltXMLElement exp)
             exp.children.remove(0);
             return exp;
 
-        case FunctionForVoltDB.FunctionId.FUNC_VOLT_SINCE_EPOCH :
+        case FunctionForVoltDB.FunctionDescriptor.FUNC_VOLT_SINCE_EPOCH :
```
BillWhite: All changes in this file are just renaming FunctionId --> FunctionDescriptor.

tests/frontend/org/voltdb/regressionsuites/TestAddDropUDF.java
```diff
@@ -95,6 +104,198 @@ public void testAddRemoveUDF() throws IOException, ProcCallException {
         }
     }
 
+    String catalogMatchesCompilerFunctionSet(Client client) throws NoConnectionsException, IOException, ProcCallException {
+        StringBuffer success = new StringBuffer();
```
BillWhite: This is just for debugging.  There are times when we expect catalog's notion of user defined functions and the compiler's notion of user defined functions to be pretty much identical.  This allows us to test this.  The tests are not a strong as one might want.

src/frontend/org/voltdb/UserDefinedFunctionManager.java
```diff
@@ -151,8 +155,12 @@ public UserDefinedFunctionRunner(Function catalogFunction, Object funcInstance)
             }
             m_returnType = VoltType.typeFromClass(m_functionMethod.getReturnType());
 
-            FunctionForVoltDB.registerTokenForUDF(m_functionName, m_functionId,
-                                                  m_functionMethod.getReturnType(), paramTypeClasses);
+            m_logger.debug(String.format("User Defined Function Manager is defining function %s (%s)", m_functionName, m_functionId));
+
+            FunctionForVoltDB.registerTokenForUDF(m_functionName,
```
BillWhite: Note the type change from Class<?> to VoltType for return types and parameter types.  This is because we need to construct catalog function elements from the VoltXML.  VoltXML only has the VoltType byte value.  We could look up the Class<?> from that, but this seemed simpler.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -992,73 +1104,126 @@ private void compileDatabase(
...
+            addExtraClasses(jarOutput);
 
-        compileRowLimitDeleteStmts(db, hsql, ddlcompiler.getLimitDeleteStmtToXmlEntries());
+            compileRowLimitDeleteStmts(db, hsql, ddlcompiler.getLimitDeleteStmtToXmlEntries());
+        }
+        catch (VoltCompilerException ex) {
+            if (m_logger.isDebugEnabled()) {
+                FunctionForVoltDB.logTableState("Compilation failed: " + ex.getMessage());
+            }
+            ddlcompiler.restoreSavedFunctions();
```
wolffcm: We only restore the saved functions if there is a `VoltCompilerException` but is it possible that other exceptions might be thrown?  Seems like we would want to restore the saved functions if any exception is thrown---maybe it belongs in a `finally` block?

src/frontend/org/voltdb/compiler/VoltDDLElementTracker.java
```diff
@@ -21,6 +21,7 @@
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
```
wolffcm: How is it that we added two new imports to this file, but no new code?  I guess the imports aren't needed?

src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
```diff
@@ -76,6 +76,20 @@ public CreateFunctionFromMethod(DDLCompiler ddlCompiler) {
...
+     *
+     * @param functionName
+     * @return
+     */
+    private boolean isDefinedFunctionName(String functionName) {
+        return FunctionForVoltDB.isFunctionNameDefined(functionName)
+                || FunctionSQL.isFunction(functionName)
+                || FunctionCustom.getFunctionId(functionName) != ID_NOT_DEFINED
+                || (null != m_schema.findChild("ud_function", functionName));
+    }
```
wolffcm: I'm curious about the contexts that each of these checks would return true...  when would the function be in `FunctionForVoltDB` but not the XML schema, and vice versa?  My guess is that the XML schema reflects the state of the DB before we started processing the current DDL batch, and that `FunctionForVoltDB` contains any functions in the current batch (but not pre-existing ones?)... Do I have this right?

Would be good to annotate this with some details.

wolffcm: Looking at the logic below now I'm not so sure of my guess...

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -897,4 +1047,93 @@ public int hashCode() {
         return val;
     }
 
+    private static Map<String, FunctionDescriptor> m_savedFunctionDfns = new HashMap<>();
```
wolffcm: This should get a comment that describes how its used.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -897,4 +1047,93 @@ public int hashCode() {
...
+
+    /**
+     * Delete all the user defined functions, but cache them if that is wanted.
+     *
+     * @param saved  Where to cache the old functions.
+     */
+    public static void deleteOldFunctions(Map<String, FunctionDescriptor> saved) {
```
wolffcm: The second argument to this method seems never to be used?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -196,250 +201,261 @@ private FunctionId(String name, Type type, int id, int typeParameter, Type[] par
...
-            new FunctionId("inet6_ntoa", Type.SQL_VARCHAR, FUNC_INET6_NTOA, -1,
+            new FunctionDescriptor("inet6_ntoa", Type.SQL_VARCHAR, FUNC_INET6_NTOA, -1,
                     new Type[] { Type.SQL_VARBINARY },
                     singleParamList),
 
         };
 
-        private static Map<String, FunctionId> by_LC_name = new HashMap<>();
-
+        private static Map<String, FunctionDescriptor> by_LC_name = new HashMap<>();
```
wolffcm: What does LC stand for here?

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -897,4 +1047,93 @@ public int hashCode() {
...
+     * Delete the defined functions, but leave the saved ones alone.
+     */
+    public static void deleteDefinedFunctions() {
+        for (String name : FunctionDescriptor.defined_functions) {
+            FunctionDescriptor.by_LC_name.remove(name);
+        }
+        FunctionDescriptor.defined_functions.clear();
+    }
+
+    public static void saveDefinedFunctions() {
```
wolffcm: This method seems equivalent to `deleteOldFunctions`?

tests/frontend/org/voltdb/regressionsuites/TestAddDropUDF.java
```diff
@@ -95,6 +104,198 @@ public void testAddRemoveUDF() throws IOException, ProcCallException {
...
+
+        catalogError = catalogMatchesCompilerFunctionSet(client);
+        assertEquals("", catalogError);
+
+        //
+        // The procedure should still exist, because the drop function failed.  We can call it in
+        // proc and in adhoc sql.
+        //
+        cr = client.callProcedure("proc");
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
```
wolffcm: I'm not sure it's necessary but it would be good to check the actual result here.

src/frontend/org/voltdb/compiler/VoltCompiler.java
yzhang1991: I changed it to `throwable`
RE: finally block
No, if the compilation succeeded, we cannot go into the finally block and restore the old ones.

src/frontend/org/voltdb/compiler/VoltDDLElementTracker.java
yzhang1991: I asked for Eclipse's help to remove them.

src/frontend/org/voltdb/compiler/statements/CreateFunctionFromMethod.java
yzhang1991: Some built-in functions are registered in `FunctionForVoltDB`. Those functions are not in the XML schema.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
yzhang1991: Don't know...

PR #4815, opened 2017-08-18
ENG-13031, fix NPE found during extensive UAC test.

luyangco:
CatalogTypeReference is not thread safe, during UAC both Site thread and Ad Hoc Planner thread may access the same object.

Change-Id: I1f4c8ffa559a24f40ffd0644daa20d4de62f3ebe

### DIFF
--- a/src/catgen/in/javasrc/CatalogType.java
+++ b/src/catgen/in/javasrc/CatalogType.java
@@ -29,6 +29,7 @@ import java.lang.reflect.Field;
  * have a name and a path (from the root). They have fields and children.
  * All fields are simple types. All children are CatalogType instances.
  *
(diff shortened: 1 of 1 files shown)


src/catgen/in/javasrc/CatalogType.java
```diff
@@ -29,32 +29,40 @@
  * have a name and a path (from the root). They have fields and children.
  * All fields are simple types. All children are CatalogType instances.
  *
+ * MUST BE THREAD SAFE, MUITIPLE THREADS ACCESS IT!
```
xinjiacs: Can you elaborate the "MULTIPLE THREADS"? Like site thread and planner thread. It gives better understanding of this thread safe issue.

PR #4818, opened 2017-08-18
ENG_13043: using site count to initialize user procedure map other th…

xinjiacs:
…an depending on site id. The join node does not have sequential site id, so it will voilate the current usage

### DIFF
--- a/src/frontend/org/voltdb/CatalogContext.java
+++ b/src/frontend/org/voltdb/CatalogContext.java
@@ -24,13 +24,12 @@ import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -72,7 +71,7 @@ public ProcedurePartitionInfo(VoltType type, int index) {
         public final byte[] m_deploymentHash;
         public final UUID m_deploymentHashForConfig;
         public Catalog m_catalog;
-        public ConcurrentHashMap<Long, ImmutableMap<String, ProcedureRunner>> m_userProcsMap;
+        public ConcurrentLinkedQueue<ImmutableMap<String, ProcedureRunner>> m_userProcsMap;
```
luyangco: Rename it?

PR #4821, opened 2017-08-21
ENG-13048: DR catalog exchange should still use gzip compression for backward compatibility

slowsoul:


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -1543,12 +1544,24 @@ else if (prevValue != null) {
...
+                                    System.out.println("DEBUG VERBOSE where new plannodetree expands to: " +
                                             new String(CompressionService.decodeBase64AndDecompressToBytes((String)newValue), "UTF-8"));
                                 }
-                                catch (UnsupportedEncodingException e) {}
+                                catch (Exception e) {
+                                    try {
+                                        System.out.println("DEBUG VERBOSE where new plannodetree expands to: " +
+                                                new String(Encoder.decodeBase64AndDecompressToBytes((String)newValue), "UTF-8"));
+                                    }
+                                    catch (UnsupportedEncodingException e2) {}
```
nshi: Why did you revert these changes, too?

slowsoul: I actually the part that tries to use Encoder if the CompressionService decompress fails. Since DRCatalogDiffEngine is a subclass of CatalogDiffEngine and calls this function as well, if verbose is turned on, this may throw Exception other than UnsupportedEncodingException.
I believe catalog from DR side will need to use Encoder to decompress, while catalog from other places such as local UAC will need to use CompressionService. So I added this so that both will be attempted.

PR #4822, opened 2017-08-21
ENG-12925, fix a bug that hash mismatch detection doesn't check the c…

luyangco:
…atalog version on user procedures.

Pro PR reference: https://github.com/VoltDB/pro/pull/2026

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -123,7 +123,6 @@ public class ProcedureRunner {
     //
     protected SiteProcedureConnection m_site;
     protected ExecutionEngine m_ee;
-    protected final SystemProcedureExecutionContext m_systemProcedureContext;
 
     // per procedure state and catalog info
     //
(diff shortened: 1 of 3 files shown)


akhanzode:
Add a comment about why this reset is needed so that next time someone looks at this code can decide what to do. Otherwise looks good.

luyangco:
Per procedure CRC includes number of statements, catalog version and parameters of each statement, that's the reason every procedure call needs to reset the value.

I'll add it to the comment.

src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -336,11 +336,7 @@ private ClientResponseImpl coreCall(Object... paramListIn) {
         Object[] paramList = paramListIn;
 
         // reset the hash of results for a new call
-        if (m_systemProcedureContext != null) {
-            m_determinismHash.reset(m_systemProcedureContext.getCatalogVersion());
-        } else {
-            m_determinismHash.reset(0);
-        }
+        m_determinismHash.reset(m_site.getSystemProcedureExecutionContext().getCatalogVersion());
```
xinjiacs: I would completely remove `m_systemProcedureContext ` from ProcedureRunner, then we will not make mistakes accessing it for user procedures. 

Other than that, it looks good.

luyangco: Do you mean separate `SystemProcedureRunner` and `UserProcedureRunner` from `ProcedureRunner`? So far both type of procedure runners instantiate from `ProcedureRunner`, user procedure runner has a null `m_systemProcedureContext`.

luyangco: Oh I see what do you mean. I'll do it.

PR #4826, opened 2017-08-22
ENG-12941: Crash in set op statement with subquery

wolffcm:
These changes have a lot of refactoring: removing dead code and making members private or final where it's possible.

The functional changes here are to use the same parameter counter for every phase of planning.  
- In HSQL, where parameters are first identified in the XML
- During ad hoc parameterization, where we convert constants into parameters
- During VoltDB planning when we find outer references in subqueries

I got rid of the NEXT_PARAMETER_ID and created a couple static methods in ParameterizationInfo reset and to get the current parameter offset.  I also created an interface used to pass this state into HSQL so it can be used there too.


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -132,7 +133,7 @@ public synchronized CompiledPlan planSqlCore(String sql, StatementPartitioning p
         QueryPlanner planner = new QueryPlanner(
             sql, "PlannerTool", "PlannerToolProc", m_database,
             partitioning, m_hsql, estimates, !VoltCompiler.DEBUG_MODE,
-            AD_HOC_JOINED_TABLE_LIMIT, costModel, null, null, DeterminismMode.FASTER);
+            costModel, null, null, DeterminismMode.FASTER);
```
wolffcm: AD_HOC_JOINED_TABLE_LIMIT was unused in QueryPlanner.

src/frontend/org/voltdb/planner/ParameterizationInfo.java
```diff
@@ -40,39 +41,78 @@
...
+        this.m_paramLiteralValues = paramLiteralValues;
     }
 
     public static ParameterizationInfo parameterize(VoltXMLElement xmlSQL) {
         assert(xmlSQL != null);
 
         VoltXMLElement parameterizedXmlSQL = xmlSQL.duplicate();
 
-        Map<String, Integer> idToParamIndexMap = new HashMap<String, Integer>();
+        Set<String> visitedParamSet = new HashSet<String>();
```
wolffcm: The map that was here didn't actually use the values associated with the keys, it was just used to ensure that we visited each param exactly once.

src/frontend/org/voltdb/planner/QueryPlanner.java
```diff
@@ -35,47 +35,44 @@
...
     // generated by parameterize(...)
-    int m_adhocUserParamsCount = 0;
+    private int m_adhocUserParamsCount = 0;
 
     // generated by plan(...)
-    boolean m_hasExceptionWhenParameterized = false;
+    private boolean m_hasExceptionWhenParameterized = false;
 
-    static boolean m_debuggingStaticModeToRetryOnError = true;
+    private static final boolean DEBUGGING_STATIC_MODE_TO_RETRY_ON_ERROR = false;
```
wolffcm: This debug flag was set to true for a long time!

src/hsqldb19b3/org/hsqldb_voltpatches/StatementDMQL.java
```diff
@@ -1332,41 +1323,46 @@ static protected void voltAppendParameters(Session session, VoltXMLElement paren
...
-            org.hsqldb_voltpatches.types.Type paramType = expr.getDataType();
-            if (paramType == null) {
-                // Parameters used with " IN ?" use a different method of recording their paramType
-                // to avoid confusing the HSQL front end. Account for that here.
-                if (expr.nodeDataTypes != null &&
-                        expr.nodeDataTypes.length == 1 &&
-                        expr.nodeDataTypes[0] != null) {
-                    paramType = expr.nodeDataTypes[0];
+
+        if (parameters.length > 0) {
```
wolffcm: It looks like more changed here because of indenting.  The main change here is that we call `resetCurrentParamIndex()` if there are parameters, and increment the current param index for each one.  That way coming out of HSQL any additional parameters we generate will start at the right index.

src/hsqldb19b3/org/hsqldb_voltpatches/HSQLInterface.java
```diff
@@ -123,13 +123,18 @@ public void finalize() {
...
     }
 
+    public static abstract interface ParameterStateManager {
+        public int getNextParamIndex();
+        public void resetCurrentParamIndex();
+    }
+
     /**
      * Load up an HSQLDB in-memory instance.
      *
```
yzhang1991: Add a comment here explaining what is `ParameterStateManager psMgr` used for?

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -957,7 +958,7 @@ private void compileDatabaseNode(
         Database db = initCatalogDatabase(m_catalog);
 
         // shutdown and make a new hsqldb
-        HSQLInterface hsql = HSQLInterface.loadHsqldb();
+        HSQLInterface hsql = HSQLInterface.loadHsqldb(ParameterizationInfo.getParamStateManager());
```
yzhang1991: Could you explain why do you want to have an extra interface on top of ParameterizationInfo?
It seems like it is just calling the static methods in ParameterizationInfo.

src/frontend/org/voltdb/planner/AbstractParsedStmt.java
```diff
@@ -426,16 +424,17 @@ private AbstractExpression parseVectorExpression(VoltXMLElement exprNode) {
      */
     private AbstractExpression parseValueExpression(VoltXMLElement exprNode) {
         String isParam = exprNode.attributes.get("isparam");
-        String isPlannerGenerated = exprNode.attributes.get("isplannergenerated");
+        String isPlannerGeneratedAttr = exprNode.attributes.get("isplannergenerated");
+        boolean isPlannerGenerated = isPlannerGeneratedAttr == null ? false :
+            (isPlannerGeneratedAttr.equalsIgnoreCase("true") ? true : false);
```
yzhang1991: I think this is OK but I am generally not very in favor of this kind of nested logical expressions.

src/frontend/org/voltdb/planner/QueryPlanner.java
yzhang1991: Nice catch

src/frontend/org/voltdb/compiler/VoltCompiler.java
wolffcm: The reason is that we currently have no dependencies on voltdb packages in HSQL.  It seems a nice design principle to continue to enforce.  If I refer to voltdb packages in HSQL, the build also breaks.  Rather than deal with that, I created an interface.

yzhang1991: That's good.
However, I guess our UDF code that directly interacts with the HSQL function map is a counter-example.

wolffcm: Currently there is just one import from voltdb in HSQL, it is the class `VoltType`.  Because that's a class that's in the client, maybe that's why the build isn't failing?

PR #4827, opened 2017-08-23
Eng 13032 timeout ntprocedure

xinjiacs:


src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -417,25 +415,31 @@ protected String verifyAndWriteCatalogJar(CatalogChangeResult ccr)
...
+            while ((elapsed = sw.elapsed(TimeUnit.SECONDS)) < (timeoutSeconds)) {
+                resultMapByHost = cf.getNow(null);
+                if (resultMapByHost != null) {
+                    sw.stop();
+                    break;
+                }
+
+                if (elapsed < 5) {
+                    // do not log under 5 seconds and sleep for 100 milliseconds
+                    Thread.sleep(TimeUnit.MILLISECONDS.toMillis(100));
```
zeemanhe: could use 100 directly.

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -417,25 +415,31 @@ protected String verifyAndWriteCatalogJar(CatalogChangeResult ccr)
...
+                    continue;
                 }
+                hostLog.info(elapsed + " seconds has elapsed but " + procedureName + " is still wait for remote response."
+                        + "The max timeout value is " + timeoutSeconds + " seconds.");
+                Thread.sleep(TimeUnit.SECONDS.toMillis(5));
             }
-        } catch (InterruptedException | ExecutionException e) {
+        } catch (Exception e) {
+            e.printStackTrace();
```
zeemanhe: remove stack trace?

xinjiacs:
ENG-13032: There is an edge case that a MPI node may fail while running UpdateClasses and the ZK UAC blocker is not removed. This is possible if this failed MPI node has not received any invocation of @UpdateCore. 

The fix is to get the completable future with a 10 seconds time out. If the future is not complete, we will remove the UAC blocker anyway and return a time out error message to client. The UpdateCore may eventually succeed even client receiving a timeout response. This will guarantee the removal of UAC blocker.

PR #4828, opened 2017-08-23
be less specific when ignoring UC shutdown errors.

jwiebema1:


jwiebema1:
this fixes a problem in the txnid2 client failing on the DR-Recovery-ap-het test.

PR #4836, opened 2017-08-25
ENG-10378: UAC can be broken into two stage: NT + TXN. We see two con…

xinjiacs:
…current UAC running in unexpected order, like NT1, NT2, TXN1, which TXN1 crashed in the middle of update catalog using corrupted prepared catalog information from NT2. We should guard against this case to happen and abort the UpdateCore TXN by checking the prepared catalog hash and deployment hash. Addiontally, we should check UAC ZK NT lock and TXN lock before starting to run rejoin.


xinjiacs:
The commit message is using a wrong ticket number instead of ENG-13078.

PR #4845, opened 2017-08-31
ENG-13090: Fix high-water undo memory issue when putting 1M batches in a txn.

jhugg:
Keep a max of 64 unused undo pools per site.
Also null out a per-proc stats object at the end of a call so it can be GCed.

This may have some impact on performance when the number of batches is super high. Unclear if 64 is the best number?


src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -290,6 +290,8 @@ public ClientResponseImpl call(Object... paramListIn) {
                                       (result.getStatus() != ClientResponse.USER_ABORT) &&
                                       (result.getStatus() != ClientResponse.SUCCESS),
                                       m_perCallStats);
+        // allow the GC to collect per-call stats if this proc isn't called for a while
+        m_perCallStats = null;
```
luyangco: That also means for every NT proc call a new `ProcedureStatsCollector` is created. So far we reuse `ProcedureStatsCollector` for all concurrent calls of the same procedure.

jhugg: Uhm. I don't read the code that way.

luyangco: My eyes fooled me, I thought you're setting `m_statsCollector = null` which will cause `ProcedureStatsCollector` to be created every time, but I am wrong. I'm ok with this pull request.

PR #4846, opened 2017-08-31
Fix interactive inline batch bug.

BillWhite:
The command "file -inlinebatch EOB" failed interactively.  This is
because we create context to remember the name of the input file.  But
an interactive command execution has no input file, so this context is
null.  The NPE is not handled properly.

I don't actually know how to test this other than manually.  The problem
does not show up in sqlcmdtest, since this input is not interactive.

### DIFF
--- a/src/frontend/org/voltdb/parser/SQLParser.java
+++ b/src/frontend/org/voltdb/parser/SQLParser.java
@@ -1365,9 +1365,10 @@ public class SQLParser extends SQLPatternFactory
                 return m_file.getPath();
             case INLINEBATCH:
             default:
+                String filePath = (m_context == null) ? "no file" : m_context.getFilePath();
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/parser/SQLParser.java
```diff
@@ -1365,9 +1365,10 @@ public String getFilePath() {
                 return m_file.getPath();
             case INLINEBATCH:
             default:
+                String filePath = (m_context == null) ? "no file" : m_context.getFilePath();
```
yzhang1991: Do we say "no file" in other places?
I remembered we say something like "AdHoc DDL Input" when the DDL is not from a file?

BillWhite: Ok.

PR #4850, opened 2017-09-05
Eng 12622 postgres test

johnTcrawford:


tests/sqlcoverage/SQLCoverageReport.py
```diff
@@ -475,18 +475,20 @@ def getTopSummaryLines(cmpdb, includeAll=True):
...
+<td colspan=3 align=center>Crashes</td>
 <td colspan=5 align=center>SQL Statements per Pattern</td>
 <td colspan=5 align=center>Time (min:sec)</td>
 </tr><tr>
 <td>Valid</td><td>Valid %%</td>
 <td>Invalid</td><td>Invalid %%</td>
 <td>Total</td>
 <td>Mismatched</td><td>Mismatched %%</td>
-<td>NPE's(V)</td><td>NPE's(%s)</td><td>Crashes</td>
+<td>NPE's(V)</td><td>NPE's(%s)</td>
```
wolffcm: Minor nit, but most style guides say not to use an apostrophe when pluralizing abbreviations without periods.  So this might be changed to `N.P.E.'s` or `NPEs`---I prefer the latter.

tests/sqlcoverage/start_postgresql.sh
```diff
@@ -20,26 +20,28 @@ if [[ "$OSTYPE" == darwin* ]]; then
...
 # Prepare to start the PostgreSQL server, in a new temp dir
 export PG_TMP_DIR=$(mktemp -d $MKTEMP_TEMPLATE)
-echo "PG_TMP_DIR:" $PG_TMP_DIR
-sudo chown postgres $PG_TMP_DIR
+echo  "PG_TMP_DIR:" $PG_TMP_DIR
 export PG_PATH=$(locate pg_restore | grep /bin | grep -v /usr/bin | tail -1 | xargs dirname)
-echo "PG_PATH:" $PG_PATH
+echo  "PG_PATH:" $PG_PATH
+export PG_PORT=5432
+echo  "PG_PORT:" $PG_PORT
```
prosegay: it would be "nice" to stay off the standard port number(s) in case someone else is running an adhoc instance

tests/sqlcoverage/stop_postgresql.sh
```diff
@@ -7,13 +7,18 @@
 # and PG_TMP_DIR must be set before running this, which they will be, if
 # start_postgresql.sh was run first, using 'source' or '.'.
 
-# Stop the PostgreSQL server & delete the temp directory
+# Stop the PostgreSQL server & delete the temp directory, but save the log file(s)
+export ORIG_DIR=`pwd`
+echo  "ORIG_DIR:" $ORIG_DIR
 ps -ef | grep -i postgres
-sudo su - postgres -c "$PG_PATH/pg_ctl stop   -D $PG_TMP_DIR/data"
+$PG_PATH/pg_ctl stop -D $PG_TMP_DIR/data
```
prosegay: I think you meant to include -w on the stop command to make it synchronous

tests/sqlcoverage/stop_postgresql.sh
```diff
@@ -7,13 +7,18 @@
...
 code1=$?
-sudo su - postgres -c "ls -l $PG_TMP_DIR/postgres.log"
-sudo su - postgres -c "cat   $PG_TMP_DIR/postgres.log"
-sudo su - postgres -c "rm -r $PG_TMP_DIR"
+ls -l $PG_TMP_DIR/postgres.log
+cat   $PG_TMP_DIR/postgres.log
+mv    $PG_TMP_DIR/postgres.log $ORIG_DIR
+# On some machines (CentOS) the log is in a different directory
+mv    $PG_TMP_DIR/data/pg_log/postgres*.log $ORIG_DIR
```
prosegay: you could also just write the log to the directory where you want it to end up by changing the postgres configuration file.

tests/sqlcoverage/SQLCoverageReport.py
johnTcrawford: Done, and while I was at it, I also changed NPEs to be major heading, with "V" and "P" (or "H", when using HSQL) sub-headings, just as I did for "Crashes"; makes the HTML Summary more compact and easier to read, I think.

tests/sqlcoverage/start_postgresql.sh
johnTcrawford: OK, changed to 5431.

tests/sqlcoverage/stop_postgresql.sh
johnTcrawford: Done

tests/sqlcoverage/stop_postgresql.sh
johnTcrawford: I'll look into this.

johnTcrawford: FYI, I already put the log where I want it; the problem is that, on certain machines, I see this:
>LOG:  redirecting log output to logging collector process
>HINT:  Future log output will appear in directory "pg_log".
After Googling this a bit, it really does not seem worth it to figure out the maze of PostgreSQL log settings so that it always goes where I want, when 'mv' will work just fine.

tests/sqlcoverage/start_postgresql.sh
johnTcrawford: Nope, undid this. I don't want to take the time right now to change the JDBC settings in PostgreSQLBackend.java.

PR #4855, opened 2017-09-07
Eng 13045 fail host on truncation snapshot failure

wweiss-voltdb:
Be more careful about incorrectly reporting success after a snapshot. Specifically, the last reporting host could overwrite a failure with success for snapshot completion listeners such as CommandLogging. In CommandLogging's case, this could cause a trucation even though the snapshot did not complete successfully.

Make sure that any IO errors generated for a Truncation Snapshot brings down the host. Detect IO errors on table, digest, completion, and catalog files.

Add Unit test that uses hacks in table file construction to force an IO exception. Verify that this generates a crash for truncation snapshots only.

Another case was SnapshotStatus Statistics, which could report a successful snapshot if a failure occured during file open or the writing of the initial table header. This is because a header write failure caused the NativeSnapshotTarget to be replaced with the DevNullSnapshotTarget, which did not preserve the earlier exception. Now, DevNullSnapshotTarget can optionally be constructed to include and propagate earlier exceptions to snapshot statistics so that snapshot stats correctly mark the result for that file as "FAILURE".

### DIFF
--- a/src/frontend/org/voltdb/SnapshotSaveAPI.java
+++ b/src/frontend/org/voltdb/SnapshotSaveAPI.java
@@ -90,6 +90,9 @@ public class SnapshotSaveAPI
             new HashMap<Integer, JSONObject>();
 
     private static ExtensibleSnapshotDigestData m_allLocalSiteSnapshotDigestData;
+
+    private static boolean m_isTruncation;
+
     /**
      * The only public method: do all the work to start a snapshot.
      * Assumes that a snapshot is feasible, that the caller has validated it can
@@ -203,6 +206,7 @@ public class SnapshotSaveAPI
             SNAP_LOG.debug("Registering transaction id " + partitionTxnId + " for " + TxnEgo.getPartitionId(partitionTxnId));
             m_partitionLastSeenTransactionIds.put(partitionId, partitionTxnId);
             m_remoteDataCenterLastIds.put(partitionId, perSiteRemoteDataCenterDrIds);
+            m_isTruncation = finalJsData != null && finalJsData.has("truncReqId");
         }
 
         boolean runPostTasks = false;
@@ -256,6 +260,7 @@ public class SnapshotSaveAPI
                                 format,
                                 taskList,
                                 multiPartTxnId,
+                                m_isTruncation,
                                 m_allLocalSiteSnapshotDigestData);
                     }
 
@@ -274,6 +279,10 @@ public class SnapshotSaveAPI
                                 }
 
                                 assert deferredSnapshotSetup != null;
+                                if (m_isTruncation && deferredSnapshotSetup.getError() != null) {
+                                    VoltDB.crashLocalVoltDB("Unexpected exception while attempting to create truncation snapshot headers",
+                                            true, deferredSnapshotSetup.getError());
+                                }
                                 context.getSiteSnapshotConnection().startSnapshotWithTargets(
                                         deferredSnapshotSetup.getPlan().getSnapshotDataTargets());
                             }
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/SnapshotSaveAPI.java
```diff
@@ -274,6 +279,10 @@ public void run()
                                 }
 
                                 assert deferredSnapshotSetup != null;
+                                if (m_isTruncation && deferredSnapshotSetup.getError() != null) {
+                                    VoltDB.crashLocalVoltDB("Unexpected exception while attempting to create truncation snapshot headers",
+                                            true, deferredSnapshotSetup.getError());
```
nshi: I'm still a bit concerned about crashing the node when snapshot fails to complete. We've done so much work earlier to make sure that running out of disk doesn't crash the database. This brings that back.

If we don't truncate the command log on incomplete snapshot, it should be fine.

src/frontend/org/voltdb/SnapshotSaveAPI.java
```diff
@@ -396,7 +405,7 @@ public void run()
             stringer.object();
             stringer.keySymbolValuePair("txnId", txnId);
             stringer.keySymbolValuePair("isTruncation", isTruncation);
-            stringer.keySymbolValuePair("didSucceed", false);
+            stringer.keySymbolValuePair("didSucceed", true);
```
nshi: Changing the default from `false` to `true` looks a bit dangerous. What if all nodes fail to write `false` to it after failures?

src/frontend/org/voltdb/SnapshotSaveAPI.java
nshi: One other point is that we have made any snapshot that ends up in the truncation snapshot directory potentially be eligible for truncating the command log. So crashing the node for system issued truncation snapshots may not be sufficient.

src/frontend/org/voltdb/SnapshotSaveAPI.java
wweiss-voltdb: The consumers of this json will only use didSucceed if hostCount is 0, and the hostCount transitions from -1 to hosts participating in snapshot. After that it behaves like a countdown latch by all the hosts racing to update this string.

src/frontend/org/voltdb/SnapshotSaveAPI.java
wweiss-voltdb: I discussed this with Ruth. The concern is that the command log keeps growing to the point where a recovery becomes an outage. Her position was that any durability failure is a system failure. I would say that if the cluster goes down because of a command log that grows too big, it may well be on a different host from the snapshot failure. At least this way the host with the disk problem will be the host that is taken offline.

To your second point, the commandlogging filters out snapshot completion notifications that don't have the truncation flag set. The flag is only set when the path is commandlog snapshot path and a truncationRequestId was passed into the snapshot request. I suppose that someone could hack this up but right now commandlogging does not verify partial snapshots either. Also, the crash is not on the node issuing the truncation snapshot but on the node that had the IOException.

nshi: There's been a regression in how truncation snapshot is determined. We used to treat all snapshots in the truncation snapshot directory as eligible truncation snapshots (https://github.com/VoltDB/pro/blob/9deb00f9ed0b8051a23daa3dff3df6e404a683ad/src/frontend/org/voltdb/sysprocs/SnapshotSave.java#L236-L248), but the following commit changed the behavior unexpectedly.

src/frontend/org/voltdb/SnapshotSaveAPI.java
nshi: Got it. Thanks.

PR #4858, opened 2017-09-08
ENG-12827 + 12736 + 12913

slowsoul:
For ENG-12827, Added a constant NO_REPLICATED_STREAM_PROTOCOL_VERSION in DRProtocol to control if the replicated stream is created or not. Currently, it is set to 9 while the latest protocol version is 8, so replicated stream is still there now. Any if checks that use this constant if part of the work to make DR work regardless of the existence of the replicated stream.

For ENG-12736, Changed DRProducer to record and persist the negotiated version with the first consumer cluster. Any later consumer clusters with protocol version lower than this version will be denied for connecting.

For ENG-12913, create a ZK blocker for elastic join if the first negotiated version is a version that does not support elastic join during DR (currently only 7), and remove the ZK blocker when all conversations with this negotiated version are reset. Also, disallow consumers with version 7 to connect if the producer cluster has been elastically modified.


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -2163,11 +2158,20 @@ void VoltDBEngine::executeTask(TaskType taskType, ReferenceSerializeInputBE &tas
...
-        m_drVersion = taskInfo.readInt();
-        m_executorContext->setDrStream(m_drStream);
-        if (m_drReplicatedStream) {
-            m_executorContext->setDrReplicatedStream(m_drReplicatedStream);
+        uint8_t drProtocolVersion = static_cast<uint8_t >(taskInfo.readInt());
+        // create or delete dr replicated stream as needed
+        if (drProtocolVersion >= DRTupleStream::NO_REPLICATED_STREAM_PROTOCOL_VERSION &&
+                m_drReplicatedStream != NULL) {
+            delete m_drReplicatedStream;
+            m_drReplicatedStream = NULL;
```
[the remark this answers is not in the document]
slowsoul: setProtocolVersion (and therefore this EE execute task) will be called only once in a DR lifecycle (starting from initial listening state and ending in a global DR reset), the version is determined by the first connected cluster. So it should be impossible for the replicated stream to be deleted if it is being used in current DR lifecycle.

src/ee/storage/DRTupleStream.h
```diff
@@ -40,10 +40,13 @@ class DRTupleStream : public voltdb::AbstractDRTupleStream {
     // Also update DRProducerProtocol.java if version changes
     // whenever PROTOCOL_VERSION changes, check if DRBufferParser needs to be updated,
     // check if unit tests that use MockPartitionQueue and getTestDRBuffer() need to be updated
-    static const uint8_t PROTOCOL_VERSION = 7;
+    static const uint8_t PROTOCOL_VERSION = 8;
     static const uint8_t COMPATIBLE_PROTOCOL_VERSION = 7;
 
-    DRTupleStream(int partitionId, int defaultBufferSize);
+    static const uint8_t ELASTICADD_PROTOCOL_VERSION = 8;
+    static const uint8_t NO_REPLICATED_STREAM_PROTOCOL_VERSION = 9;
```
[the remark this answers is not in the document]
slowsoul: Yes, we can set NO_REPLICATED_STREAM_PROTOCOL_VERSION to 8 to enable it in version 8. That was the intention, I think I should change it to 8 as early as possible to test DR across version now that the code reviews seems not having any major concern on the implementation. Hopefully, we can include it in the release.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -187,6 +187,7 @@ void VoltDBEngine::initialize(int32_t clusterIndex,
     m_partitionId = partitionId;
     m_tempTableMemoryLimit = tempTableMemoryLimit;
     m_compactionThreshold = compactionThreshold;
+    m_createDrReplicatedStream = createDrReplicatedStream;
```
wweiss-voltdb: Ultimately this boolean will go away as it is currently named but it is generally useful for other things like Shared Replicated. If you are touching this anyway, I would rename this to lowestLocalPartitionId.

src/frontend/org/voltdb/VoltZK.java
```diff
@@ -351,7 +352,8 @@ public static String createCatalogUpdateBlocker(ZooKeeper zk, String node, VoltL
             case elasticJoinActiveBlocker:
                 // elastic join can not happen during node rejoin
                 if (zk.getChildren(VoltZK.catalogUpdateBlockers, false).size() > 1) {
-                    errorMsg = "while another elastic join, rejoin or catalog update is active";
+                    errorMsg = "while another elastic join, rejoin or catalog update is active" +
+                        " or while elastic join is disallowed";
```
wweiss-voltdb: This message is confusing to me

src/ee/execution/VoltDBEngine.cpp
slowsoul: Do you mean something renaming this boolean to "isLowestLocalPartiiton"?

src/frontend/org/voltdb/VoltZK.java
slowsoul: I wanted to express that if elastic join is blocked by DR because of using version 7, this active blocker cannot be created, which means you can't start an elastic join. Maybe the wording is a bit confusing? Do you have any suggestions?

PR #4859, opened 2017-09-11
ENG-12812 voltQueueExperimental was moved, so fix txnid2 to use refle…

prosegay:
…ction to find it in its old and new home


jhugg:
I think this looks fine. It's simple enough. It really could use some comments just in case there comes a time you or I aren't here and someone asks WTF. The helper file should explain what problem it's solving at the very least.

PR #4861, opened 2017-09-12
ENG-13046: Add WARN log for HOST logger in case of MP routed to SPI

slowsoul:
After this is merged to master, the system tests can turn on DEBUG level logging for this one-time MISROUTE logger so that we may have more information if this happens again.

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunner.java
+++ b/src/frontend/org/voltdb/ProcedureRunner.java
@@ -61,6 +61,7 @@ import org.voltdb.iv2.UniqueIdGenerator;
 import org.voltdb.jni.ExecutionEngine;
 import org.voltdb.messaging.FastDeserializer;
 import org.voltdb.messaging.FragmentTaskMessage;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -567,6 +569,19 @@ public boolean checkPartition(TransactionState txnState, TheHashinator hashinato
             return false;
         } else {
             if (!m_catProc.getEverysite() && m_site.getCorrespondingPartitionId() != MpInitiator.MP_INIT_PID) {
+                if (misrouteLog.isDebugEnabled()) {
```
nshi: This will never show up in production or test unless this new logger is enabled to the DEBUG level. How about making it a warning in the HOST log?

slowsoul: Yes, we can also do that. Made the changes.

slowsoul:
@nshi Adjusted log message per Andrew's suggestion, ready to merge.

PR #4867, opened 2017-09-18
ENG-13108:

manjujames:
- Throw mispartitioned exception when SP apply binary logs see mispartitioned hashes.
- Change to make ClientInterface not restart txn on mispartitioned, if explicitly told not to restart.
- New internal status in ClientResponse to indicate mispartitioned txn.

Tested manually for the 2 mispartitioning cases:
- Dummy key is mispartitioned
- Hashes in txn are mispartitioned


src/ee/storage/BinaryLogSink.cpp
```diff
@@ -509,26 +509,19 @@ int64_t BinaryLogSink::applyTxn(ReferenceSerializeInputLE *taskInfo,
...
-    }
-    else {
-        // Check MP single hash txn to see if it is for local site.
-        // This also handles TXN_PAR_HASH_REPLICATED case, where nothing ever needs to be skipped.
-        skipWrongHashRows = hashFlag == TXN_PAR_HASH_SINGLE &&
-            UniqueId::isMpUniqueId(uniqueId) &&
-            !engine->isLocalSite(partitionHash);
-    }
+    isLocal = engine->isLocalSite(partitionHash);
+    bool isLocalMpTxn = UniqueId::isMpUniqueId(uniqueId);
```
wweiss-voltdb: You could call skipOrThrow here and assign the result to skipWrongHashRows.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -544,16 +537,37 @@ int64_t BinaryLogSink::applyTxn(ReferenceSerializeInputLE *taskInfo,
...
-                             const char *txnStart, int64_t sequenceNumber, int64_t uniqueId, bool skipRow) {
+                             const char *txnStart, int64_t sequenceNumber, int64_t uniqueId,
+                             bool isLocalMpTxn, bool isForLocalPartition) {
     switch (type) {
     case DR_RECORD_INSERT: {
         int64_t tableHandle = taskInfo->readLong();
         int32_t rowLength = taskInfo->readInt();
         const char *rowData = reinterpret_cast<const char *>(taskInfo->getRawPointer(rowLength));
-        if (skipRow) {
+        if (skipOrThrow(isLocalMpTxn, isForLocalPartition)) {
```
nshi: You can actually call this once at the beginning and cache it for each type.

src/ee/storage/BinaryLogSink.h
```diff
@@ -41,10 +41,12 @@ class BinaryLogSink {
                      const char *txnStart);
 
 private:
+    bool skipOrThrow(bool isLocalMpTxn, bool isForLocalPartition);
+
     int64_t apply(ReferenceSerializeInputLE *taskInfo, const DRRecordType type,
                   boost::unordered_map<int64_t, PersistentTable*> &tables,
                   Pool *pool, VoltDBEngine *engine, int32_t remoteClusterId,
-                  const char *txnStart, int64_t sequenceNumber, int64_t uniqueId, bool skipRow);
+                  const char *txnStart, int64_t sequenceNumber, int64_t uniqueId, bool skipRow, bool isLocal);
```
nshi: Should also rename the second to last parameter to `isLocalMp`

src/frontend/org/voltdb/client/ClientResponse.java
```diff
@@ -83,6 +83,12 @@
     public static final byte OPERATIONAL_FAILURE = -9;
 
     /**
+     * Status code indicating that the transaction was sent to the wrong partition.  These are used internally to Volt
+     * and shouldn't leak out to actual clients.
+     */
+    public static final byte TXN_MISPARTITIONED = -10;
```
nshi: Maybe the current mispartitioned txn response can also use this instead of having two different ones.

src/ee/storage/BinaryLogSink.cpp
manjujames: Yea, I will see how I can change it around. I can throw outside of apply() and skip inside. Skip has to be inside the apply because we read there.

src/ee/storage/BinaryLogSink.cpp
manjujames: The hash value changes, so we have to check this every time. But, I am going to rearrange the code to make it more easily understandable based on Walter's comment.

src/ee/storage/BinaryLogSink.h
manjujames: OK.

src/frontend/org/voltdb/client/ClientResponse.java
manjujames: The current mispartitioned handling is in ClientInterface and does it before a ClientResponse is generated. It bases it on the setting on InitateResponseMessage.

PR #4870, opened 2017-09-19
Eng 12474 udf benchmark

yzhang1991:


tests/frontend/org/voltdb/client/MockVoltClient.java
```diff
@@ -289,6 +289,11 @@ public ClientStatsContext createStatsContext() {
     }
 
     @Override
+    public void writeSummaryCSV(String statsRowName, ClientStats stats, String path) throws IOException {
+        // TODO Auto-generated method stub
```
BillWhite: I wonder.  Maybe there should be a comment here explaining why this is empty.
I'm not sure the TODO line is helpful.  It looks kind of like you just forgot.

tests/test_apps/udfbenchmark/run.sh
```diff
@@ -0,0 +1,100 @@
...
+    mkdir -p obj
+    javac -classpath $APPCLASSPATH -d obj \
+        $APPNAME/*.java
+    # stop if compilation fails
+    if [ $? != 0 ]; then exit; fi
+    pushd obj > /dev/null
+    jar cvf ../UDFLib.jar $APPNAME/UDFLib*.class
+    popd > /dev/null
+}
+
```
jwiebema1: Apprunner needs a function called jars() that creates the jar file, you can either rename this function or add a second function that points to this.

PR #4884, opened 2017-09-22
ENG-13132 alter table drop unnamed constraint

yzhang1991:


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -1563,9 +1566,33 @@ static Table addTableConstraintDefinitions(Session session, Table table,
                         index = table.createAndAddIndexStructure(indexName,
                             c.core.mainCols, null, null, true, true, false).setAssumeUnique(c.assumeUnique);
                     }
+                    Constraint newconstraint;
+                    if (c.getIsAutogeneratedName()) {
+                        namePart = HSQLInterface.AUTO_GEN_CONSTRAINT_PREFIX + table.getName().name + "_" + StringUtils.join(((IndexAVL)index).getColumnNameList().toArray(), "_");
```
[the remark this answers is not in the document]
wolffcm: In any case, you need the constraint names in VoltDB and in HSQL to match exactly don't you?  The best way to enforce them would be to name the VoltDB and HSQL constraints using the same code.

src/hsqldb19b3/org/hsqldb_voltpatches/index/IndexAVL.java
```diff
@@ -1763,10 +1763,10 @@ public VoltXMLElement voltGetIndexXML(Session session, String tableName,
                 // Also name the constraint here based on our naming scheme.
                 String voltdbConstraintName = null;
                 if (isPrimaryKey) {
-                    voltdbConstraintName = HSQLInterface.AUTO_GEN_CONSTRAINT_PREFIX + "_PK_" + namePart;
+                    voltdbConstraintName = HSQLInterface.AUTO_GEN_CONSTRAINT_PREFIX + "PK_" + namePart;
```
wolffcm: I don't see where the `PK_` part gets inserted in the corresponding HSQL's `addTableConstraintDefinitions`.  Won't the constraint names be different?

tests/frontend/org/voltdb/TestAdhocAlterTable.java
```diff
@@ -835,13 +835,27 @@ public void testAlterConstraintAssumeUnique() throws Exception
...
-            // Should be no indexes in the system (no constraints)
-            VoltTable indexes = m_client.callProcedure("@Statistics", "INDEX", 0).getResults()[0];
-            assertEquals(0, indexes.getRowCount());
+                    "alter table FOO drop constraint VOLTDB_AUTOGEN_CT_FOO_VAL;");
+            indexes = m_client.callProcedure("@Statistics", "INDEX", 0).getResults()[0];
+            do {
+                indexes = m_client.callProcedure("@Statistics", "INDEX", 0).getResults()[0];
+            }
+            while (indexes.getRowCount() != 0);
+
```
wolffcm: Nice test.  I think there's more to add though: do the same with dropping a PK constraint with an auto-generated name.  Also add the constraint (before doing the test that ensures you can drop it) with an `ALTER TABLE ADD CONSTRAINT` command.  I think that the name is optional for `ALTER TABLE ADD CONSTRAINT`... so auto-generated names can be created that way too, and I believe that it uses a different codepath in HSQL.

wolffcm: How does the user discover the auto-generated name of a constraint?  I believe that there is no way short of using a debugger.  There is `@SystemCatalog PRIMARYKEYS`, but that will not work for [ASSUME]UNIQUE constraints.  Maybe we should file a ticket for this.

If you do add a similar test for PRIMARY KEY constraint with an auto-generated name, then you could verify that the name produced by `@Statistics PRIMARYKEY` looks as its expected and that it can be used to drop the constraint (meaning that the name must be the same in HSQL and in VoltDB.

wolffcm: Maybe the additional testing that I mentioned is already in there, I'm not sure.

PR #4886, opened 2017-09-25
ENG-12931: Ensure DR is started before verifying poison pill path. Al…

wweiss-voltdb:
…so update C++ UniqueId object to provide equivalent toString as Java and use it to print out better UniqueIds in PoisonPill messages. Fix minor IDE complaint in btree_map.h where we try to return a value in a void method.

### DIFF
--- a/third_party/cpp/stx/btree_map.h
+++ b/third_party/cpp/stx/btree_map.h
@@ -533,7 +533,7 @@ public:
     /// Erase the key/data pair referenced by the iterator.
     void erase(iterator iter)
     {
-	return tree.erase(iter);
+        tree.erase(iter);
     }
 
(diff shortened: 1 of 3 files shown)


third_party/cpp/stx/btree_map.h
```diff
@@ -533,7 +533,7 @@ class btree_map
     /// Erase the key/data pair referenced by the iterator.
     void erase(iterator iter)
     {
-	return tree.erase(iter);
```
nshi: This looks like a rather random change. Was it failing compilation?

wweiss-voltdb: It generates a IDE error so I fixed it up (and mentioned it in the patch comment).

PR #4892, opened 2017-09-27
Eng 13130 fix issue1 rebase

luyangco:
This patch addresses two different causes of recent hash mismatch exception.

Stored procedure generates inconsistent query plans or catalog hashes on different nodes,
Stored procedure succeed on one side but failed because wrong number of parameters on the other side.
Also just for precaution, this patch reverts a previous commit that allow two NT-threads running in a single node.


luyangco:
reviewed by nshi

PR #4898, opened 2017-09-28
Eng 13180 fix issue2

luyangco:
Serialize all planning/HSQL access. Multi-threaded planning is not supported, so before we can do it all accesses to the planner need to be serialized.

### DIFF
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -38,7 +38,6 @@ import org.voltdb.planner.PlanningErrorException;
 import org.voltdb.planner.QueryPlanner;
 import org.voltdb.planner.StatementPartitioning;
 import org.voltdb.planner.TrivialCostModel;
-import org.voltdb.plannodes.AbstractPlanNode;
 import org.voltdb.utils.CompressionService;
 import org.voltdb.utils.Encoder;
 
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -197,9 +194,6 @@ public synchronized AdHocPlannedStatement planSql(String sqlIn, StatementPartiti
                 }
             }
 
-            // Reset plan node id counter
-            AbstractPlanNode.resetPlanNodeIds();
```
nshi: Hmm, this was definitely not thread-safe if another place other the current instance of PlannerTool accesses it.

src/frontend/org/voltdb/planner/ParameterizationInfo.java
```diff
@@ -50,10 +51,10 @@
      * Stores the current count of parameters that we've
      * assigned to the parameter vector in the EE.
      *
-     * TODO: this will need to become thread safe
-     * when parallel planning is possible.
+     * Make it atomic because it's possible to have either both Site thread and
+     * NTService Thread or multiple NTService threads planning in parallel.
      */
-    static private int curParamIndex = 0;
+    static private AtomicInteger curParamIndex = new AtomicInteger(0);
```
nshi: This only prevents it from generating the same ID in different threads, but it doesn't guarantee that the numbers will be monotonically increasing for a given thread. Does that matter? I feel like it does.

wolffcm: Yes, it matters.  We really want the adjacent and starting with zero here.  Also, the numbers we generate need to be deterministic.

src/frontend/org/voltdb/compiler/PlannerTool.java
yzhang1991: Actually, it is also necessary to reset it here, nice catch.

PR #4906, opened 2017-10-03
Fix Pro check.

akhanzode:


src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -442,7 +442,7 @@ else if ("@UpdateLogging".equals(procName)) {
 
             // ERROR MESSAGE FOR PRO SYSPROC USE IN COMMUNITY
 
-            if (!MiscUtils.isPro()) {
```
luyangco: Wait, I think the check is right, it means "If you're running community, and you try to run pro-only sysproc, fails the transaction."

PR #4907, opened 2017-10-03
ENG-13137, drop enterprise only tag on some system procedures.

luyangco:
Change-Id: Ic5179ba55846029d147bb0036b5b487d1a318d36


akhanzode:
Looks good.

PR #4911, opened 2017-10-03
Eng 10092 Fix misoriented polygons from polygonfromtext.

BillWhite:
When one uses polygonfromtext to create a polygon, we now try to fix up the orientation of the rings.

We also have a benchmark app.  It turns out to be quite expensive to do this fixing up.  For
polygons with 750 vertices, it takes about twice as long to insert invalid polygons than to
insert valid polygons.

Note that inserting GEOGRAPHY objects, and not polygonfromtext, allows inserting invalid
polygons.


src/ee/expressions/geofunctions.cpp
```diff
@@ -32,6 +32,9 @@ namespace voltdb {
 
 static const int POINT = FUNC_VOLT_POINTFROMTEXT;
 static const int POLY = FUNC_VOLT_POLYGONFROMTEXT;
+// Set this to false if we don't want polygonfromtext
+// to automatically repair badly configured polygons.
```
BillWhite: I left this in to make it easier to make baselines for the benchmark program.

src/ee/expressions/geofunctions.cpp
```diff
@@ -292,7 +295,7 @@ static void readLoop(bool is_shell,
     loop->Init(points);
 }
 
-static NValue polygonFromText(const std::string &wkt, bool doValidation)
+static NValue polygonFromText(const std::string &wkt, bool doValidation, bool doRepairs)
```
BillWhite: This is more complexity.  But sometimes we want to validate but not do repairs and sometimes we want to do both or neither.

wolffcm: You can't do repairs without validation, so there are only three valid states?  You could use an enum type to combine these two parameters, which would make code in the callers easier to read.

src/ee/expressions/geofunctions.cpp
```diff
@@ -339,7 +358,7 @@ static NValue polygonFromText(const std::string &wkt, bool doValidation)
     char* storage = const_cast<char*>(ValuePeeker::peekObjectValue(nval));
 
     Polygon poly;
-    poly.init(&loops); // polygon takes ownership of loops here.
+    poly.init(&loops, doRepairs); // polygon takes ownership of loops here.
     if (doValidation) {
```
wolffcm: So I guess `doValidation` means "throw if invalid"?  Seems like this parameter could have a better name.

src/ee/expressions/geofunctions.cpp
```diff
@@ -362,7 +381,7 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POLYGONFROMTEXT>() const
     const char* textData = getObject_withoutNull(&textLength);
     const std::string wkt(textData, textLength);
 
-    return polygonFromText(wkt, false);
+    return polygonFromText(wkt, false, false);
```
wolffcm: I only see two callers to `polygonFromText` one with "false, false" and one with "true, true".  Maybe the parameters could be combined, or am I missing a caller?

src/ee/expressions/geofunctions.cpp
```diff
@@ -584,6 +607,41 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_INVALID_REASON>() const {
...
+    // Extract the polygon and check its validity.
+    std::stringstream msg;
+    Polygon poly;
+    poly.initFromGeography(getGeographyValue(), true);
+    if ( ! poly.IsValid(&msg)) {
+        std::string res (msg.str());
+        assert(res.size() > 0);
+        throwInvalidMakeValidPoly(res);
+        // No return from here.
+    } else if ( isMultiPolygon(poly, &msg)) {
```
wolffcm: There should be a newline after the closing brace.

src/ee/expressions/geofunctions.cpp
```diff
@@ -584,6 +607,41 @@ template<> NValue NValue::callUnary<FUNC_VOLT_POLYGON_INVALID_REASON>() const {
...
+        assert(res.size() > 0);
+        throwInvalidMakeValidPoly(res);
+        // No return from here.
+    }
+    // Ok, so the polygon either was valid before or else
+    // we repaired it, and it is not a multi polygon.
+    // So, msg will not be the empty string, and we can package
+    // this polygon up.
+    //
+    std::string res (msg.str());
```
wolffcm: What is `res`?  You can remove it and just assert `msg.str().size() == 0`.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -188,10 +188,12 @@ private FunctionDescriptor(String name, Type type, int id, int typeParameter, Ty
                                                                             // certain distance of each other
         static final int FUNC_VOLT_DWITHIN_POINT_POINT          = 21018;    // if two points are within certain distance of each other
         static final int FUNC_VOLT_DWITHIN_POLYGON_POINT        = 21019;    // if a polygon and a point are within certain distance of each other
-        static final int FUNC_VOLT_VALIDPOLYGONFROMTEXT         = 21020;    // list polygonFromText, but validates after construction
+        static final int FUNC_VOLT_VALIDPOLYGONFROMTEXT         = 21020;    // like polygonFromText, but validates after construction
```
wolffcm: Now, not just validates, but also fixes, and will throw if unfixable.

tests/frontend/org/voltdb/planner/eegentests/GenerateEETests.java
```diff
@@ -750,6 +779,7 @@ public static void main(String args[]) {
         try {
             tg.setUp();
             tg.generatedPlannerTest();
+            tg.generatedInsertPolygonTest();
```
wolffcm: If you are only adding this so you can debug your code in GDB, do you still want to generate the test for every build?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -665,8 +688,19 @@ public void testPolygonPointDistance() throws Exception {
      * These are two nested CCW (Widdershins) rectangles.
      */
     private static String TWO_NESTED_WIDDERSHINS
```
wolffcm: I guess all these private static variables can also be made final.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -769,6 +840,48 @@ public void testValidPolygonFromText() throws Exception {
         }
     }
 
+    /**
+     * Insert the polygons in the borders table, using polygonfromtext.  Some
```
wolffcm: You mean `VALIDPOLYGONFROMTEXT` here?

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
```diff
@@ -208,7 +215,10 @@ void S2Polygon::InsertLoop(S2Loop* new_loop, S2Loop* parent,
 void S2Polygon::InitLoop(S2Loop* loop, int depth, LoopMap* loop_map) {
   if (loop) {
     loop->set_depth(depth);
-    loops_.push_back(loop);
+    /// If we had cleared loops_ this is where
+    /// we would have put them back.  But for now
+    /// we just set the depth.
+    /// loops_.push_back(loop);
```
wolffcm: If we had cleared them?  Why would we have done that?  So many questions.

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
```diff
@@ -241,14 +252,37 @@ void S2Polygon::Init(vector<S2Loop*>* loops) {
...
-  /// Reorder the loops in depth-first traversal order.
-  loops_.clear();
+  /// The loop_map helps us calculate the depth.
+  /// Stock S2 reorders the loops so that all holes
+  /// follow their shells.  That's the normal order
+  /// for us, and users will complain if we reorder
+  /// their loops, so we will leave them alone.
+  ///
+  /// This line clears the loops, as S2 originally did.
+  /// We don't want that here.
```
wolffcm: I find it really confusing to have comments about what the code used to do, and commented out lines of code.  I guess we own this fork now.  My suggestion is to just describe the code as it works now, with your changes.  If we ever need to understand what it used to do, we can use version control for that.

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
```diff
@@ -241,14 +252,37 @@ void S2Polygon::Init(vector<S2Loop*>* loops) {
...
+  ///
+  /// loops_.clear();
   InitLoop(NULL, -1, &loop_map);
 
+#if  0
+  /// This is no longer necessarily valid, since we don't reorder
```
wolffcm: I'm not a fan of `#if 0`... This is a pretty straightforward check of the invariants in the loop map, right?

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
```diff
@@ -259,17 +293,30 @@ void S2Polygon::Init(vector<S2Loop*>* loops) {
...
-
-  /// Compute the bounding rectangle of the entire polygon.
+#endif
+  /// Compute the bounding rectangle of the entire polygon,
+  /// and whether the polygon has holes.
+#if 0
   has_holes_ = false;
   bound_ = S2LatLngRect::Empty();
+  ///
+  /// This is the original code.  We actually know
```
wolffcm: I think you can just remove this.

src/ee/expressions/geofunctions.cpp
BillWhite: Ok.  I eliminated the doValidation parameter.  It seems like the SQL interface is clumsy, and we will be called to do this over again.  Then validation and repair will want to be separated.  But for this version of the SQL interface I guess this will do.

src/ee/expressions/geofunctions.cpp
BillWhite: Fixed.

src/ee/expressions/geofunctions.cpp
BillWhite: I've seen that elsewhere.  Why do we do that?  It seems unmotivated and weird.  But I changed it.

src/ee/expressions/geofunctions.cpp
BillWhite: What's the difference?  I've changed it, but it seems like not a significant difference.

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
BillWhite: Fixed.

tests/frontend/org/voltdb/planner/eegentests/GenerateEETests.java
BillWhite: You know, none of these really need to be generated at build time.  They are all there as patterns for
GDB debuggable EE unit tests.  Maybe we should eliminate all of them.  Nobody likes them but me, anyway.
I suppose I could go back to debugging with printf.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
BillWhite: Does it really matter?  The only advantage final has over const in C++ is that it's not as much of a pain in the neck.  Otherwise it's equally useless.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
BillWhite: Done.

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
BillWhite: Ok, I've removed this.  It seems like we will forget that this is here, though, and sometime in the future we will reinvent it.

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
BillWhite: Ok, it's gone.

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
BillWhite: Ok, it's gone.

third_party/cpp/google-s2-geometry/geometry/s2polygon.cc
BillWhite: All gone, except for the bounds setting, of course.

src/ee/expressions/geofunctions.cpp
wolffcm: I thought it was in our coding standard that opening braces are at the end of lines, and that closing braces are on their own line.  Most of our code seems to follow this convention.  I don't see it in the our repo wiki now though.

src/ee/expressions/geofunctions.cpp
wolffcm: If you've got suggestions to how the SQL needs to improved please file tickets so we can discuss...

src/ee/expressions/geofunctions.cpp
wolffcm: I wasn't sure what `res` was supposed to stand for, so I started looking around for other uses... but found none.  This variable is unused in release builds.  If you only want to assert on the length, I think it's better to avoid creating a local variable if you can.  Even in the other instances of `res` above, we may be copying the string an extra time when we don't need to.

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
wolffcm: It's not useless---it tells the reader of the code that the variable will always refer to the same object, and the compiler will enforce that for you.  It's also more useful than in C++ because finality cannot be casted away.  You took the time to use call caps for these variables, why not follow through with `final`?

src/ee/expressions/geofunctions.cpp
```diff
@@ -561,7 +561,7 @@ static bool isMultiPolygon(const Polygon &poly, std::stringstream *msg) {
         }
     }
     if (nouters != 1) {
-        VMLOG(2, msg) << "Polygons can have only one shell";
+        VMLOG(2, msg) << "Polygons can have only one shell, not " << nouters;
```
wolffcm: What is VMLOG?  Seems like it's only used in S2 code.  I think it would be preferable to use a more Volt-like mechanism, like VOLT_TRACE.

src/ee/expressions/geofunctions.cpp
```diff
@@ -616,6 +616,12 @@ template<> NValue NValue::callUnary<FUNC_VOLT_MAKE_VALID_POLYGON>() const {
     std::stringstream msg;
     Polygon poly;
     poly.initFromGeography(getGeographyValue(), true);
+    for (int idx = 0; idx < poly.num_loops(); idx += 1) {
+        S2Loop *loop = poly.loop(idx);
+        if ( ! loop->IsNormalized() ) {
+            std::cout << "Not normalized loop in make_valid_polygon: " << idx << "\n";
```
wolffcm: You'll want to take out this writing to `std::cout` before you merge.  Should this be an assertion?

tests/frontend/org/voltdb/regressionsuites/TestGeospatialFunctions.java
```diff
@@ -74,16 +76,23 @@ static private void setUpSchema(VoltProjectBuilder project) throws IOException {
...
-
-        Border(long pk, String name, String message, GeographyValue region) {
-            this(pk, name, message, region, false);
+            m_valid = isValid;
+            m_fixable = isFixable;
+            // m_message must not be null, unless m_region is also null.
+            // and if m_region is null then m_message must be null.
+            // The message may be "Valid Polygon", or it may be a
+            // regexp pattern for an error message from isinvalidreason().
+            assertTrue((message == null) == (region == null));
```
wolffcm: You could just add another constructor rather that enforces this.

third_party/cpp/google-s2-geometry/geometry/s2loop.cc
```diff
@@ -288,9 +288,11 @@ bool S2Loop::IsNormalized(std::stringstream *msg) const {
   return turningAngle >= -1e-14;
 }
 
-void S2Loop::Normalize() {
+void S2Loop::Normalize(bool fixedFirstVertex) {
   CHECK(owns_vertices_);
-  if (!IsNormalized()) Invert();
+  if (!IsNormalized()) {
+      Invert(fixedFirstVertex);
```
wolffcm: Do you ever want to invert and not have a fixed first vertex?

src/ee/common/GeographyValue.hpp
```diff
@@ -506,23 +508,54 @@ inline void Loop::pointArrayFromBuffer(Deserializer& input, std::vector<S2Point>
...
+        set_rect_bound(bound);
+        std::vector<S2Point> vertices(num_vertices());
+        for (int idx = 0; idx < num_vertices(); idx += 1) {
+            // Note that this makes a copy of the vertices
+            // in the vector of vertices.  The variable vertices
+            // is not a vector of pointers but a vector of
+            // S2Point objects.
+            vertices[idx] = src[idx];
+        }
+        Init(vertices);
```
wolffcm: I guess you don't need to allocate the temporary `vertices` here---but you would need to change `S2Loop.Init` to accept an array to avoid the extra copy.

BillWhite: Well, we need a copy someplace.  Otherwise we would be inverting the argument loop.

src/ee/expressions/geofunctions.cpp
BillWhite: I've thought about this.  But it seemed like more work than was necessary.  It's used all over S2, and it's style is very different from VOLT_TRACE.  It seems like a lot of work to ferret out all the occurrences and convert them, and there would be no functional improvement.

third_party/cpp/google-s2-geometry/geometry/s2loop.cc
BillWhite: I don't know.  The original didn't invert the first vertex, and I didn't want to have to find subtle bugs caused by this.  I just left the existing behavior and added our behavior for what we do.

src/ee/expressions/geofunctions.cpp
wolffcm: I'm not advocating for using VOLT_TRACE in S2 code, just for avoiding use of VMLOG in Volt code, and using VOLT_TRACE instead.

BillWhite: Hmm.  VMLOG writes to msg.  This is not really a trace, it's an output to an ostringstream.

PR #4929, opened 2017-10-11
ENG-10247 Deprecate GeographyPointValue algebra methods

yzhang1991:
I have a question:
Does the `mul()` function also need to be deprecated?
```java
/**
 * Return a point scaled by the given alpha value.
 *
 * @param alpha  The amount by which to scale this.
 * @return The scaled point.
 */
 public GeographyPointValue mul(double alpha) {
    return GeographyPointValue.normalizeLngLat(getLongitude() * alpha + 0.0,
                                               getLatitude() * alpha  + 0.0);
}
```

### DIFF
--- a/src/frontend/org/voltdb/types/GeographyPointValue.java
+++ b/src/frontend/org/voltdb/types/GeographyPointValue.java
@@ -419,6 +419,7 @@ public class GeographyPointValue {
      * @param alpha  The amount by which to scale this.
      * @return The scaled point.
      */
+    @Deprecated
     public GeographyPointValue mul(double alpha) {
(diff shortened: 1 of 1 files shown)


wolffcm:
Yes, `mul` should be deprecated, and also `rotate` too.

wolffcm:
If you were feeling ambitious you could also see to it that `javadoc` is invoked with the `-nodeprecated` option in `build-client.xml`.  (Maybe we could do this for all our client classes?)

yzhang1991:
I doubt Andrew would like to see all the deprecated APIs to disappear from javadoc. Let me ask him.

PR #4936, opened 2017-10-16
Eng 13168 monatomic last received

wweiss-voltdb:


src/frontend/org/voltdb/dr2/DRIDTrackerHelper.java
```diff
@@ -107,15 +108,15 @@ public static String jsonifyClusterTrackers(Pair<Long, Long> lastConsumerUniqueI
      * @param base The base map to merge trackers into.
      * @param add  The additional trackers to merge.
      */
-    public static void mergeTrackers(Map<Integer, Map<Integer, DRConsumerDrIdTracker>> base,
-                                     Map<Integer, Map<Integer, DRConsumerDrIdTracker>> add)
+    public static void mergeTrackers(Map<Integer, Map<Integer, DRSiteDrIdTracker>> base,
```
slowsoul: This function is now only used in restarting consumer in case of elastic join, we cannot really directly merge DRSiteDrIdTracker in this way. Before the change, we do a gather->merge->scatter so that trackers are merged and all consumer partitions' lastReceivedDRId will be reset to the some point (the lastDrId of the merged tracker). After the change, lastReceivedLogId don't rely on tracker any more, ，they are independent to each consumer partition, but using the old approach, only the first trackers' lastReceivedLogId will be used, others will be dropped. I think what the restartConsumerNT needs to do under the new scheme is to issue a MP to populate empty tracker and 0 lastReceivedLogId on the new consumer partitions, old consumer partitions' trackers can stay as is.

PR #4938, opened 2017-10-17
ENG-13132 alter-table-constraint

yzhang1991:
This is a large refactor and fix for our auto-generated names for constraints and indexes.
Previously, HSQL was using its own internal names for indexes and constraints that were not given a name. Those names are different from what VoltDB used. So sometimes although you see those internal names in some @SystemCatalog selectors, you have no way to drop them.

This pull request united all the system internal names. As a benefit, some constraints that we were not able to drop in the past are deletable now.


yzhang1991:
Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/hsqldb19b3/org/hsqldb_voltpatches/HSQLInterface.java, line 57 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/4938#-KwkVJI6w92KHKx4BYfD:-KwkVJI6w92KHKx4BYfE:b-jkgmnb) ([raw file](https://github.com/voltdb/voltdb/blob/cc86e8820902b1e644fd98c9bdad12fde81e6b6a/src/hsqldb19b3/org/hsqldb_voltpatches/HSQLInterface.java#L57)):*
> ```Java
>     public static final String AUTO_GEN_PRIMARY_KEY_PREFIX = AUTO_GEN_PREFIX + "PK_";
>     public static final String AUTO_GEN_UNIQUE_IDX_PREFIX = AUTO_GEN_PREFIX + "UQ_";
>     public static final String AUTO_GEN_NAMED_CONSTRAINT_IDX = AUTO_GEN_PREFIX + "CONSTRAINT_IDX_";
> ```

Need to take care of this `AUTO_GEN_NAMED_CONSTRAINT_IDX`

---

*[src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java, line 299 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/4938#-KwkVjFz4ciTMqh4K7rA:-KwkVjFz4ciTMqh4K7rB:b-kn0vtd) ([raw file](https://github.com/voltdb/voltdb/blob/cc86e8820902b1e644fd98c9bdad12fde81e6b6a/src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java#L299)):*
> ```Java
>         }
> 
>         // Get a hash number if there are expressions based on simple columns.
> ```

I am actually thinking to remove this hash number completely.
It seems like that we can only have one unique constraint on one column.
You cannot have constraints on both `a+1` and `a+2` where they both have the same prefix for the auto-generated name but different hash numbers.

I need to test more to decide whether I should remove this hash number completely.

---

*[src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java, line 308 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/4938#-KwkWIAUrzadm3zzDe1h:-KwkWIAUrzadm3zzDe1i:bl1h9ps) ([raw file](https://github.com/voltdb/voltdb/blob/cc86e8820902b1e644fd98c9bdad12fde81e6b6a/src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java#L308)):*
> ```Java
>                 hashStringBuffer.setLength(hashStringBuffer.length() - 1);
>             }
>             byte[] bytes = hashStringBuffer.toString().getBytes();
> ```

This is the hash number computation code I grabbed from HSQL, It is moved here from `IndexAVL`

---

*[src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java, line 69 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/4938#-KwkWVcqKmSrjhw6qoE4:-KwkWVcqKmSrjhw6qoE5:b19zggv) ([raw file](https://github.com/voltdb/voltdb/blob/cc86e8820902b1e644fd98c9bdad12fde81e6b6a/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java#L69)):*
> ```Java
>     };
> 
>     static final String PRIMARY_KEY = "PK";
> ```

Maybe I will move them to another file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/voltdb/voltdb/4938)*

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1282,28 +1283,30 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
             }
 
             if (subNode.name.equals("indexes")) {
-
-                // Do the system indexes first, since we don't want to
-                // drop them: there are constraint objects in the catalog
+                // Add system-generated indexes first so they get priority when the
```
yzhang1991: We used to check every index twice here, now we only check once.

src/frontend/org/voltdb/JdbcDatabaseMetaDataGenerator.java
```diff
@@ -574,7 +574,7 @@ VoltTable getPrimaryKeys()
 
                         // if the index name is "MATVIEW_PK_CONSTRAINT", the column name for the materialized view is obtained
                         // from its column information - ENG-6927
-                        if (c.getTypeName().equals(HSQLInterface.AUTO_GEN_MATVIEW_CONST) ) {
```
yzhang1991: All the name prefixes are now moved to HsqlNameManager. I think this makes more sense.

src/frontend/org/voltdb/plannodes/IndexCountPlanNode.java
```diff
@@ -494,9 +494,8 @@ protected String explainPlanForNode(String indent) {
         String retval = "INDEX COUNT of \"" + m_targetTableName + "\"";
         String indexDescription = " using \"" + m_targetIndexName + "\"";
         // Replace ugly system-generated index name with a description of its user-specified role.
-        if (m_targetIndexName.startsWith(HSQLInterface.AUTO_GEN_PRIMARY_KEY_PREFIX) ||
-                m_targetIndexName.startsWith(HSQLInterface.AUTO_GEN_NAMED_CONSTRAINT_IDX) ||
-                m_targetIndexName.equals(HSQLInterface.AUTO_GEN_MATVIEW_IDX) ) {
+        if (m_targetIndexName.startsWith(HsqlNameManager.AUTO_GEN_PRIMARY_KEY_PREFIX)
+            ||  m_targetIndexName.equals(HsqlNameManager.AUTO_GEN_MATVIEW_IDX)) {
```
yzhang1991: I did not touch the auto-generated indexes for materialized views in this pull request. They can be left as they are now.

src/hsqldb19b3/org/hsqldb_voltpatches/HSQLInterface.java
```diff
@@ -46,24 +46,6 @@
     private static final VoltLogger m_logger = new VoltLogger("HSQLDB_COMPILER");
 
     public static final String XML_SCHEMA_NAME = "databaseschema";
-    /**
-     * Naming conventions for unnamed indexes and constraints
```
yzhang1991: Everything here is moved to HsqlNameManager, some of them are removed.

src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java
```diff
@@ -75,14 +99,15 @@
...
+            autoColumnNames[i] = new HsqlName(staticManager,
+                                              makeAutoColumnName("C", i),
+                                              0,
                                               false);
             autoNoNameColumnNames[i] = String.valueOf(i);
         }
     }
 
     private int      serialNumber = 1;        // 0 is reserved in lookups
-    private int      sysNumber    = 10000;    // avoid name clash in older scripts
```
yzhang1991: This was a counter that HSQL used to generate system-reserved names. Now it is completely removed, we will use the same VoltDB generated names in HSQL as well.

src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java
```diff
@@ -253,43 +265,84 @@ static public String getAutoSavepointNameString(long i, int j) {
...
-                sb.append(prefix);
-                sb.append('_');
-
-                if (namepart != null) {
-                    sb.append(namepart);
-                    sb.append('_');
-                }
-
-                sb.append(++sysNumber);
+    public HsqlName newAutoName(String objectTypePrefix,
```
yzhang1991: This is the completely new naming function that replaces the old HSQL one.
The name generated here will be used in both HSQL and VoltDB.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -1535,12 +1536,16 @@ static Table addTableConstraintDefinitions(Session session, Table table,
...
+                    boolean hasComplexExpr = false;
+                    List<Expression> indexExprList = null;
                     // A VoltDB extension to support indexed expressions and the assume unique attribute
                     if (c.indexExprs != null) {
                         // Special case handling for VoltDB indexed expressions
                         if (table.getUniqueConstraintForExprs(c.indexExprs) != null) {
                             throw Error.error(ErrorCode.X_42522);
                         }
+                        indexExprList = Arrays.asList(c.indexExprs);
+                        hasComplexExpr = getSimpleColumnNames(indexExprList) == null;
```
yzhang1991: If the constraint is defined on some expressions more general than simple column expression, we need to append the generated name with a hash number.

src/hsqldb19b3/org/hsqldb_voltpatches/Table.java
```diff
@@ -2704,20 +2709,16 @@ VoltXMLElement voltGetTableXML(Session session)
...
-                indexConstraintMap.put(constraintChild.attributes.get("index"), constraintChild);
-            }
         }
 
         // read all the indexes
         VoltXMLElement indexes = new VoltXMLElement("indexes");
         indexes.attributes.put("name", "indexes");
         for (Index index : indexList) {
-            VoltXMLElement indexChild = index.voltGetIndexXML(session, tableName, indexConstraintMap);
+            VoltXMLElement indexChild = index.voltGetIndexXML(session, tableName);
```
yzhang1991: HSQL and VoltDB are now using the same name, so we do not need a map now.

src/hsqldb19b3/org/hsqldb_voltpatches/TableWorks.java
```diff
@@ -338,19 +339,16 @@ void addColumn(ColumnSchema column, int colIndex,
                     }
 
                     HsqlName indexName =
-                        database.nameManager.newAutoName("IDX",
-                                                         c.getName().name,
-                                                         table.getSchemaName(),
-                                                         table.getName(),
+                        database.nameManager.newAutoName("IDX", table,
```
yzhang1991: Didn't touch here because it's for foreign key...We do not have that.

src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java
```diff
@@ -253,43 +265,84 @@ static public String getAutoSavepointNameString(long i, int j) {
...
+        if (parentSchemaObject != null) {
+            name.schema = parentSchemaObject.getSchemaName();
+            name.parent = parentSchemaObject.getName();
+        }
+        return name;
+    }
 
+    /**
+     * Get an auto-generated name for auto-generated indexes or anonymous constraints.</br>
+     * This method omits <i>associatedSchemaObject</i> and <i>detailedDescriptor</i> for simplicity.
```
wolffcm: Is this just a convenience method?  I guess you mean `parentSchemaObject`?  And detailedDescriptor seems disappeared now?

src/hsqldb19b3/org/hsqldb_voltpatches/Table.java
wolffcm: Nice!!

tests/frontend/org/voltdb/TestAdhocAlterTable.java
```diff
@@ -250,7 +250,7 @@ public void testAlterDropColumn() throws Exception
             assertFalse(doesColumnExist("FOO", "VIEWCOL"));
             assertFalse(findTableInSystemCatalogResults("FOOVIEW"));
 
-            // single-column indexes get cascaded automagically
+            // single-column indexes get cascaded automatically
```
wolffcm: Just so you know, this was not a typo, I'm sure the author meant it as a whimsical portmanteau of "automatic" and "magically".

tests/frontend/org/voltdb/catalog/TestDRCatalogDiffs.java
```diff
@@ -529,7 +529,7 @@ public void testExtraPrimaryKeyOnReplica() throws Exception {
 
         CatalogDiffEngine diff = runCatalogDiff(masterSchema, false, replicaSchema, false);
         assertFalse(diff.supported());
-        assertTrue(diff.errors().contains("Missing Index{VOLTDB_AUTOGEN_IDX_PK_T1_C1} from Table{T1} on remote cluster"));
+        assertTrue(diff.errors().contains("Missing Index{VOLTDB_AUTOGEN_PK_T1_C1} from Table{T1} on remote cluster"));
```
wolffcm: So you are changing the names of primary key and/or unique indexes?  Does this have implications for DR... do they require the index names to match?  I seem to recall that maybe they do need to match.  I'm not sure though.

wolffcm: Keeping in mind that the remote cluster could be running a different version of VoltDB.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
```diff
@@ -1530,82 +1528,6 @@ public void testDDLCompilerCreateAndDropIndexesOnMatView() {
         checkDDLErrorMessage(ddl, null);
     }
 
-    public void testColumnNameIndexHash() {
```
wolffcm: You are deleting this test?  It's still possible to create has indexes, isn't it?

src/hsqldb19b3/org/hsqldb_voltpatches/HsqlNameManager.java
yzhang1991: Oh you are right, I forgot to update this part of the comment.

tests/frontend/org/voltdb/TestAdhocAlterTable.java
yzhang1991: Oh right, I did not look closely enough.

tests/frontend/org/voltdb/catalog/TestDRCatalogDiffs.java
yzhang1991: Let me run this through Walter tomorrow.

tests/frontend/org/voltdb/compiler/TestVoltCompiler.java
yzhang1991: Users still can, and we give deprecation warnings. But I changed the code so that auto-generated indexes will no longer be created using the hash index. That was what this test for.

manjujames:
With index and constraint name changes, it could cause problems when DR'ing with older versions because the schema won't match.

yzhang1991:
Can we add code to somewhere to make the change compatible with older versions?
I don't like to be restricted by this that once we add something to the catalog and it is released, then we will never have a chance to make changes in the future.

manjujames:
We probably can. But, that is something we have to think about carefully and do.

If we do something like translating to older version names, then we need to handle multiple versions. It might be better if we can compare the indexes and constraints after ignoring the names and comparing other attributes like columns. I am not sure what the reasoning behind the original strategy is, so we need to talk/think about it.

yzhang1991:
We will resolve the DR issues in ENG-13585 then merge it.

PR #4941, opened 2017-10-18
ENG-13107: Add hashinator signature to DR query response

nshi:


src/frontend/org/voltdb/TheHashinator.java
```diff
@@ -194,14 +194,6 @@ public static TheHashinator getHashinator(Class<? extends TheHashinator> hashina
     abstract protected boolean pIsPristine();
     abstract public int getPartitionFromHashedToken(int hashedToken);
 
-    /**
-     * Helper method to check of the Hashinator has ever been modified
-     * @return  true if modified (due to an elastic join)
-     */
-    static public boolean clusterHasBeenElasticallyJoined() {
```
slowsoul: Producer side may still need to use this function to figure out whether it can accept a protocol version that doesn't support elastic join during DR. So I think this may need to be preserved.

nshi: I didn't see this gets used other than sending it to the consumer, so I removed it.

I also did some code reading and it turned out that the hashinator after elastic rebalance is actually the same hashinator as if it had started with the new partition count. So theoretically, it should be fine to support older versions.

PR #4946, opened 2017-10-19
ENG-13201:

manjujames:
Now allows a single extendBufferChain for defaultSize < minLength <= maxSize.

DR should allow changes where a single tuple size exceeds the default size but is less than the max size. Before this change, extendBufferChain in TupleStreamBase checked if the requested extension is greater than the default size and failed if it was, which is incorrect behavior for DR. This PR changes TupleStreamBase to support maxSize also and changed extendBufferChain accordingly.

### DIFF
--- a/src/ee/storage/TupleStreamBase.cpp
+++ b/src/ee/storage/TupleStreamBase.cpp
@@ -39,9 +39,10 @@ using namespace voltdb;
 
 const int MAX_BUFFER_AGE = 4000;
 
-TupleStreamBase::TupleStreamBase(size_t defaultBufferSize, size_t extraHeaderSpace /*= 0*/)
+TupleStreamBase::TupleStreamBase(size_t defaultBufferSize, size_t extraHeaderSpace /*= 0*/, int maxBufferSize /*= -1*/)
     : m_flushInterval(MAX_BUFFER_AGE),
       m_lastFlush(0), m_defaultCapacity(defaultBufferSize),
+      m_maxCapacity( (maxBufferSize < defaultBufferSize) ? defaultBufferSize : maxBufferSize),
       m_uso(0), m_currBlock(NULL),
       // snapshot restores will call load table which in turn
       // calls appendTupple with LONG_MIN transaction ids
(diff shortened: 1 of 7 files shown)


src/ee/storage/TupleStreamBase.cpp
```diff
@@ -67,6 +68,9 @@ TupleStreamBase::setDefaultCapacity(size_t capacity)
                             "TupleStreamBase is used");
     }
     cleanupManagedBuffers();
+    if (m_maxCapacity < capacity || m_maxCapacity == m_defaultCapacity) {
```
nshi: Why `m_maxCapacity == m_defaultCapacity`? Wouldn't this lower the max capacity by accident?

tests/ee/storage/DRTupleStream_test.cpp
```diff
@@ -466,12 +502,23 @@ TEST_F(DRTupleStreamTest, TxnSpanBufferThrowException)
...
 /**
+ * Single tuple (single extendBufferChain call) larger than default size,
+ * but less than secondary size should work correctly.
+ */
+TEST_F(DRTupleStreamTest, TupleLargerThanDefaultSize)
+{
+    appendLargeTuple(0, 1);
+    m_wrapper.endTransaction(addPartitionId(1));
+    m_wrapper.periodicFlush(-1, addPartitionId(1));
+    ASSERT_TRUE(m_topend.receivedDRBuffer);
```
nshi: Maybe also assert that the buffer size is between default and max.

src/ee/storage/TupleStreamBase.cpp
manjujames: This method is only used by tests. They create the tuple stream with a default capacity (no max specified) and then update the defaultCapacity to use a new value. In these scenarios, the intention is to use the same value for defaultCapacity and maxCapacity. There is really no way to know at this point if maxCapacity was specified or if we just made it the same as defaultCapacity. If I change the implementation to leave maxCapacity as -1 to indicate that it was unspecified, we can know, I suppose. At the time I didn't think it was worth the code because this is only a test method. Let me know if you think otherwise.

tests/ee/storage/DRTupleStream_test.cpp
manjujames: OK.

src/ee/storage/TupleStreamBase.cpp
nshi: I didn't realize that it was only for tests. Maybe rename the method to something like `setFixedCapacityForTest()` ?

manjujames: OK

nshi:
There's a test failure. http://ci.voltdb.lan/view/Branch-jobs/view/branch-ENG-13201/job/branch-2-memcheck-lite-ENG-13201/4/console

PR #4947, opened 2017-10-19
Eng 13178 snapshot restore error msg

xinjiacs:
1. Fix to have better error messages.
2. Hack to use the same API to report error messages, instead of database crashing by non-exists path for restore.


src/frontend/org/voltdb/sysprocs/SnapshotRestore.java
```diff
@@ -555,6 +555,22 @@ else if (fragmentId == SysProcFragmentId.PF_restoreScan)
...
                     dupPath = (SnapshotPathType.valueOf(m_filePathType) == SnapshotPathType.SNAP_PATH ?
                             dupPath : m_filePath);
+
+                    VoltFile outputPath = new VoltFile(dupPath);
+                    String errorMsg = null;
+                    if (!outputPath.exists()) {
+                        errorMsg = "Output path for Json duplicatesPath \"" + outputPath + "\" does not exist";
+                    }
+                    if (!outputPath.canExecute()) {
+                        errorMsg = "Output path for Json duplicatesPath \"" + outputPath + "\" is not executable";
```
zeemanhe: change "Json duplicatesPath"  in these messages  to "duplicate rows"?

xinjiacs: I left this error message unchanged. If there is a need, we can add a new commit to address this message. 

In terms of my perspective, the SnapshotRestore system procedure uses Json string as an input parameter, thus it returns this "Json duplicatsPath" message. Check this line:
 https://github.com/VoltDB/voltdb/blob/master/src/frontend/org/voltdb/sysprocs/saverestore/SnapshotUtil.java#L1673

PR #4949, opened 2017-10-19
ENG-12555: large block manager

wolffcm:
This class adds a LargeBlockManager class which gets a singleton instance created when the database starts up.  It also adds a new directory `large_query_swap` under voltdbroot.

In the next ticket that I'll work on, this class will be called from the EE to store large temp table blocks that do not fit into memory.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -959,6 +974,20 @@ public void initialize(Configuration config) {
...
+            }
+
+            try {
+                LargeBlockManager.getInstance().clearSwapDir();
+            }
+            catch (IOException ioExc) {
+                // Failure to clear out files from large_query_swap doesn't seem fatal,
+                // perhaps the user created a file in there and the server doesn't have
+                // permission to delete it?
+                hostLog.warn("Could not clear large query swap directory: " + ioExc.getMessage());
```
akhanzode: This should crash right?

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -959,6 +974,20 @@ public void initialize(Configuration config) {
             m_configLogger = null;
             ActivePlanRepository.clear();
 
+            if (LargeBlockManager.getInstance() == null) {
+                LargeBlockManager.initializeInstance(Paths.get(getLargeQuerySwapPath()));
```
akhanzode: getInstance() can create initialize right? If this is not required for init it should be done only for StartAction.PROBE

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3190,6 +3219,17 @@ public boolean shutdown(Thread mainSiteThread) throws InterruptedException {
                         init.shutdown();
                 }
 
+                LargeBlockManager lbm = LargeBlockManager.getInstance();
```
akhanzode: will be good if it looks like LargeBlockManager.getInstance().shutdown()

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -140,6 +140,11 @@
               <xs:attribute name="path" type="xs:string" default="command_log_snapshot"/>
             </xs:complexType>
         </xs:element>
+        <xs:element name="largequeryswap" minOccurs="0" maxOccurs="1">
```
akhanzode: Why cant it be just queryswap?

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
```diff
@@ -140,6 +140,11 @@
...
+            <xs:complexType>
+              <xs:attribute name="path" type="xs:string" default="large_query_swap"/>
```
akhanzode: We dont have _ in deployment maybe use camelCase?

src/frontend/org/voltdb/largequery/LargeBlockManager.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Cleans out the large_query_swap directory of all files.
+     * Large query intermediate storage and results do not persist
+     * across shutdown/startup.
+     * This method is to clean up any spurious files that may not have
+     * been deleted due to an unexpected shutdown.
+     * @throws IOException
+     */
+    public void clearSwapDir() throws IOException {
```
akhanzode: If its needed to be cleared in init --force please do cleanup there too.

src/frontend/org/voltdb/largequery/LargeBlockManager.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * Store the given block with the given ID to disk.
+     * @param id     the ID of the block
+     * @param block  the bytes for the block
+     * @throws IOException
+     */
+    public void storeBlock(long id, ByteBuffer block) throws IOException {
+        synchronized (m_accessLock) {
```
akhanzode: Do you intend to block all partitions? Is it good to make these blocks per partition?
If this is going to be called from EE you will need static entry points.

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
wolffcm: We've had discussions about the name.  Consensus was elusive.  This was what we ended up with.

src/frontend/org/voltdb/compiler/DeploymentFileSchema.xsd
wolffcm: This is consistent with the other names in the file.

src/frontend/org/voltdb/largequery/LargeBlockManager.java
wolffcm: For now, access to the block manager is single-threaded.  For (not supported yet) MP plans we'll need to use the block manager to pass large temp tables between the coordinator and the distributed fragment.  Perhaps this design will need to be revisited, but for now this will suffice.

wolffcm:
@akhanzode Hey Anish, I pushed a commit that I hope addresses your concerns, can you have another look?

As far as making access to the block manager via static method, there are a few other examples of ExecutionEngineJNI instance methods being called from the JNI Topend.  I've been thinking it over and I like this way for a few reasons:
1. It helps to limit the number of classes that the EE topend talks to.  I think that this makes the code a bit easier to understand.
2. In the future there might be processing that needs to happen between the EE and the LargeBlockManager, and the Java ExecutionEngine classes could be a good place to put this.
3. I've already got a branch based on this one that uses instance methods...
4. Done this way, IPC support will be symmetrical to JNI

wolffcm:
Oh, I also added static methods to LargeBlockManager for start up and shut down, rather than exposing the lower level things that need to happen.

yzhang1991:
Reviewed 8 of 12 files at r1, 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/frontend/org/voltdb/largequery/LargeBlockManager.java, line 74 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/4949#-KxxSR9D40XzGmU3ZSHY:-KxxSR9D40XzGmU3ZSHZ:b-fza3zm) ([raw file](https://github.com/voltdb/voltdb/blob/c69e09cdfc0c5134be2065fbfbf773b3a43f501a/src/frontend/org/voltdb/largequery/LargeBlockManager.java#L74)):*
> ```Java
>     public static void startup(Path largeQuerySwapPath) throws IOException {
>         if (INSTANCE == null) {
>             INSTANCE = new LargeBlockManager(largeQuerySwapPath);
> ```

So I saw Anish had this comment that he likes the idea of `LargeBlockManager.getInstance().startUp()`. What is your opinion about that?

---

*[tests/frontend/org/voltdb/regressionsuites/LocalCluster.java, line 744 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/4949#-Kxx_8YGq-V3NSXu8GE3:-Kxx_8YGq-V3NSXu8GE4:b-u2ykd7) ([raw file](https://github.com/voltdb/voltdb/blob/c69e09cdfc0c5134be2065fbfbf773b3a43f501a/tests/frontend/org/voltdb/regressionsuites/LocalCluster.java#L744)):*
> ```Java
>             return m_hostRoots.get(hostId) + File.separator + Constants.DBROOT;
>         }
>         else {
> ```

Curious why we did not have this code for handling old CLI

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/voltdb/voltdb/4949)*

wolffcm:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/frontend/org/voltdb/largequery/LargeBlockManager.java, line 74 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/4949#-KxxSR9D40XzGmU3ZSHY:-KxysoItWm5Lpec_DMSw:b-nrn7rg) ([raw file](https://github.com/voltdb/voltdb/blob/c69e09cdfc0c5134be2065fbfbf773b3a43f501a/src/frontend/org/voltdb/largequery/LargeBlockManager.java#L74)):*
<details><summary><i>Previously, yzhang1991 (Yiqun Zhang) wrote…</i></summary><blockquote>

So I saw Anish had this comment that he likes the idea of `LargeBlockManager.getInstance().startUp()`. What is your opinion about that?
</blockquote></details>

I took Anish's comment to say that he wants a general method that's called on start up, and I thought it was a good idea too, so I added a method called `startUp`.  I wanted the interface with `RealVoltDB.initialize()` to be abstract and stable also, so I opted for a static method rather than requiring the client to get a singleton instance.  At some point we might want to have multiple threads, etc, and it was easy to hide these details from callers.

---

*[tests/frontend/org/voltdb/regressionsuites/LocalCluster.java, line 744 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/4949#-Kxx_8YGq-V3NSXu8GE3:-Kxyu7V8WwRcRKGs-Wh-:bt08vqt) ([raw file](https://github.com/voltdb/voltdb/blob/c69e09cdfc0c5134be2065fbfbf773b3a43f501a/tests/frontend/org/voltdb/regressionsuites/LocalCluster.java#L744)):*
<details><summary><i>Previously, yzhang1991 (Yiqun Zhang) wrote…</i></summary><blockquote>

Curious why we did not have this code for handling old CLI
</blockquote></details>

I'm guessing that it was just never needed.  Not many tests want to know the root directories if a running cluster it seems.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/voltdb/voltdb/4949)*

wolffcm:
Merged on a separate branch.

PR #4951, opened 2017-10-19
ENG-13194, fix a race condition in @Shutdown or @StopNode, that MP Si…

luyangco:
…te waits for borrowed task response but

the SP site runs that borrowed task was halted by @Shutdown or @StopNode, it causes MP Site to wait forever.

Change-Id: I15688d4d0bc4c60e285acbfce6146aaa5d7f56c2

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -3136,11 +3136,7 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
                 // no longer available
                 m_snmp.hostDown(FaultLevel.INFO, m_messenger.getHostId(), "Host is shutting down");
 
-                // tell the iv2 sites to stop their runloop
-                if (m_iv2Initiators != null) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3137,9 +3137,10 @@ public boolean shutdown(Thread mainSiteThread) throws InterruptedException {
                 m_snmp.hostDown(FaultLevel.INFO, m_messenger.getHostId(), "Host is shutting down");
 
                 // tell the iv2 sites to stop their runloop
+                // The reason to halt MP sites first is that it may wait for some fragment dependencies
+                // to be done on SP sites, kill SP sites first may risk MP site to wait forever.
                 if (m_iv2Initiators != null) {
-                    for (Initiator init : m_iv2Initiators.values())
-                        init.shutdown();
+                    m_iv2Initiators.descendingMap().values().stream().forEach(p->p.shutdown());
```
akhanzode: Why not consolidate in single place shutdownInitiators()

PR #4954, opened 2017-10-20
ENG-13265, delay setting mode to RUNNING until server is ready for ac…

luyangco:
…cepting clients.

Change-Id: Idd8db2e119bbb0c51a97f6cd39e45e0a4813d3cf


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -4003,6 +3996,13 @@ public void onReplayCompletion(long txnId, Map<Integer, Long> perPartitionTxnIds
             m_statusTracker.setNodeState(NodeState.UP);
```
wweiss-voltdb: So if m_startMode == PAUSED, what will be logged, and what will the clientInterface do when it is initialized?

PR #4955, opened 2017-10-20
Fix ENG-13271-check-as-begin-syntax

yzhang1991:
When splitting statements, we checked for keywords like "as", "begin", "case", "end" because they could change our semantics for a semicolon that we usually use to determine the end of a statement.

We used `Character.isLetterOrDigit` and `isLetterOrDigitFast` to check whether the token we checked was indeed a standalone token or a part of another token. But those two functions will return wrong answers when the tokens have "_" before or after them.

This pull request updates the functions we use so we can give right answers when underscores are presented.

### DIFF
--- a/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
+++ b/src/frontend/org/voltdb/planner/ParsedSelectStmt.java
@@ -829,9 +829,6 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
         // Parse the expression.  We may substitute for this later
         // on, but it's a place to start.
         AbstractExpression colExpr = parseExpressionTree(child);
-        if (colExpr instanceof ConstantValueExpression) {
-            assert(colExpr.getValueType() != VoltType.NUMERIC);
-        }
         assert(colExpr != null);
 
         if (isDistributed) {
@@ -839,6 +836,9 @@ public class ParsedSelectStmt extends AbstractParsedStmt {
             updateAvgExpressions();
         }
         ExpressionUtil.finalizeValueTypes(colExpr);
+        if (colExpr instanceof ConstantValueExpression) {
+            assert(colExpr.getValueType() != VoltType.NUMERIC);
+        }
 
         if (colExpr.getValueType() == VoltType.BOOLEAN) {
             throw new PlanningErrorException(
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/planner/ParsedSelectStmt.java
```diff
@@ -829,16 +829,16 @@ private void parseDisplayColumn(int index, VoltXMLElement child,
...
-        }
         assert(colExpr != null);
 
         if (isDistributed) {
             colExpr = colExpr.replaceAVG();
             updateAvgExpressions();
         }
         ExpressionUtil.finalizeValueTypes(colExpr);
+        if (colExpr instanceof ConstantValueExpression) {
+            assert(colExpr.getValueType() != VoltType.NUMERIC);
```
wolffcm: What's with this change?  Seems unrelated to everything else.

tests/frontend/org/voltdb/TestAdhocCreateStatementProc.java
```diff
@@ -36,6 +36,28 @@
...
+    public void testENG13271() throws Exception {
+        String pathToCatalog = Configuration.getPathToCatalogForTest("adhocddl.jar");
+        VoltProjectBuilder builder = new VoltProjectBuilder();
+        builder.addLiteralSchema(
+            "CREATE TABLE T (\n" +
+            "   COLUMN_CASE tinyint\n" +
+            ");\n" +
+            "CREATE PROCEDURE PROC1\n" +
+            "AS BEGIN\n" +
+            "   INSERT INTO T (COLUMN_CASE) VALUES (?);\n" +
```
wolffcm: So the issue was that they had `_case` in an identifier?  Should you be testing for `_` afterwards as well?

src/frontend/org/voltdb/planner/ParsedSelectStmt.java
yzhang1991: The user's query hit this with an assertion error because the `colExpr.getValueType() == VoltType.NUMERIC`. But this value type will be addressed by `ExpressionUtil.finalizeValueTypes`. So I am moving this assertion back a little bit.
It is really an issue if the expression is NUMERIC type even after being processed by the `finalizeValueTypes` function.

tests/frontend/org/voltdb/TestAdhocCreateStatementProc.java
yzhang1991: Adding a test case for it.

yzhang1991:
The matchToToken function is only used for parsing multi-statement procedure DDL.

PR #4957, opened 2017-10-21
ENG-12827 No Replicated Stream

slowsoul:
Fix DR binary log generation and handling of replicated table changes in non-16383 stream, the hash flag in BEGIN_TXN and HASH_DELIMITER is ANDed with a mask to indicate the next records are for replicated tables until the next HASH_DELIMITER or the END_TXN.

On the receiving side, also avoid calling isLocalSite() if we know the record is for a replicated table.

Backward compatible with old format where there will be no mask and replicated table changes are replicated in a specialized replicated stream, in that case, the hash flag will be TXN_PAR_HASH_REPLICATED and all the records in the txn will be replicated table change and handled accordingly.

### DIFF
--- a/src/ee/storage/DRTupleStream.cpp
+++ b/src/ee/storage/DRTupleStream.cpp
@@ -747,7 +747,11 @@ int32_t DRTupleStream::getTestDRBuffer(uint8_t drProtocolVersion,
                                        long startSequenceNumber,
                                        char *outBytes)
 {
-    DRTupleStream stream(partitionId,
+    int tupleStreamPartitionId = partitionId;
+    if (partitionId == 16383 && drProtocolVersion >= NO_REPLICATED_STREAM_PROTOCOL_VERSION) {
+        tupleStreamPartitionId = 0;
+    }
+    DRTupleStream stream(tupleStreamPartitionId,
                          2 * 1024 * 1024 + MAGIC_HEADER_SPACE_FOR_JAVA + MAGIC_DR_TRANSACTION_PADDING, // 2MB
                          drProtocolVersion);
 
--- a/src/ee/storage/DRTupleStream.cpp
+++ b/src/ee/storage/DRTupleStream.cpp
@@ -48,6 +48,9 @@ DRTupleStream::DRTupleStream(int partitionId, size_t defaultBufferSize, uint8_t 
       m_hashFlag(m_initialHashFlag),
       m_firstParHash(LONG_MAX),
       m_lastParHash(LONG_MAX),
+      m_hasReplicatedStream(drProtocolVersion < NO_REPLICATED_STREAM_PROTOCOL_VERSION),
+      m_wasFirstChangeReplicatedTable(false),
+      m_wasLastChangeReplicatedTable(false),
       m_beginTxnUso(0),
       m_lastCommittedSpUniqueId(0),
       m_lastCommittedMpUniqueId(0)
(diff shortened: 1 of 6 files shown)


src/ee/storage/DRTupleStream.cpp
```diff
@@ -712,7 +747,11 @@ int32_t DRTupleStream::getTestDRBuffer(uint8_t drProtocolVersion,
                                        long startSequenceNumber,
                                        char *outBytes)
 {
-    DRTupleStream stream(partitionId,
+    int tupleStreamPartitionId = partitionId;
+    if (partitionId == 16383 && drProtocolVersion >= NO_REPLICATED_STREAM_PROTOCOL_VERSION) {
+        tupleStreamPartitionId = 0;
```
nshi: Does this assume that the replicated table stream is always going to be on partition 0? What if the current local partition is not 0?

src/ee/storage/DRTupleStream.h
```diff
@@ -40,12 +40,14 @@ class DRTupleStream : public voltdb::AbstractDRTupleStream {
...
     // whenever PROTOCOL_VERSION changes, check if DRBufferParser needs to be updated,
     // check if unit tests that use MockPartitionQueue and getTestDRBuffer() need to be updated
-    static const uint8_t PROTOCOL_VERSION = 8;
+    static const uint8_t PROTOCOL_VERSION = 9;
     static const uint8_t COMPATIBLE_PROTOCOL_VERSION = 7;
 
     static const uint8_t ELASTICADD_PROTOCOL_VERSION = 8;
     static const uint8_t NO_REPLICATED_STREAM_PROTOCOL_VERSION = 9;
 
+    static const int8_t REPLICATED_TABLE_MASK = INT8_MIN;
```
nshi: This sets the first bit to 1, right?

src/ee/storage/DRTupleStream.cpp
slowsoul: This is only used for test, so in this case, it will create a DRTupleStream with partition 0 and generate the requested buffer on this stream.

src/ee/storage/DRTupleStream.h
nshi: I wonder if we really need to put this in the hash flag byte. Can we tell if it's for replicated table by either the hash flag or the table name?

If we really do need this extra bit, I suggest putting it closer to the record type it affects, which is the hash flag type.

slowsoul: Basically, the bit is used to make the hash flag and the hash delimiter to act as an indicator that the next records until the next hash delimiter or the END_TXN is for replicated tables or not. The bit in the hash flag is specially used as the first indicator since we don't have hash delimiter at the beginning. The intention was to put a layer above the actual hash flag semantics so that the determination and population of the hash flag need very small change. I could have added a _REPLICATED variation for each of the existing hash flag but that may be too verbose and is not as efficient as bit operations can't be used. Considering that there are still 128 values after the first bit is used and we now only have 5 values used, I decided to use the first bit. For the table name approach, I think it's doable but not very efficient and doesn't go well with hash delimiter because then we would have to check the table name for every single record before the first hash delimiter, which also requires looking ahead into the actual record data. Does it make sense?

I agree that we can put it closer to the definition of hash flag types and record types, which is in types.h, instead of putting it in DRTupleStream.h

nshi: It makes sense. I'm fine if you just put this closer to the type definition and make it obvious that this only sets the first bit.

src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -50,6 +50,11 @@
         SPECIAL
     }
 
+    // Keep sync with EE REPLICATED_TABLE_MASK at types.h
+    // This mask uses -128 which corresponds to 0b10000000
```
wweiss-voltdb: Binary conventions are seldom used (particularly for masks). Use 0x80 instead.

PR #4969, opened 2017-10-26
ENG-13162 Guard window function partition/order by subquery

yzhang1991:
We now do not allow ORDER BY subquery. We add similar guards for ORDER BY and PARTITION BY subqueries for the window functions.

Can Andrew confirm if the message is OK?


BillWhite:
Looks right to me.

PR #4971, opened 2017-10-26
ENG-13290, add missing piece that when VoltZK.banElasticJoin is set, …

luyangco:
…no elastic join is allowed. (#4970)

Change-Id: I7ba252378e71d793d1a198c2f1394e0d2e1f18e0


luyangco:
r/=Rui

PR #4973, opened 2017-10-26
Add indicator message for retry

slowsoul:


src/frontend/org/voltdb/VoltZK.java
```diff
@@ -332,7 +332,8 @@ public static String createActionBlocker(ZooKeeper zk, String node, CreateMode m
                 VoltDB.crashLocalVoltDB("Unable to create action blocker " + node, true, e);
             }
             // node exists
-            return "Invalid " + request + " request: Can't run " + request + " when another one is in progress";
+            return "Invalid " + request + " request: Can't do " + request +
+                    " while another one is in progress. Please retry " + request;
```
JadenW: This would be translated to "Please retry update(@UpdateClasses)" which wouldn't be catched by pdateClassesThread.java that you modified?

src/frontend/org/voltdb/sysprocs/UpdateCore.java
```diff
@@ -438,8 +438,8 @@ private final void performCatalogVerifyWork(
         ZooKeeper zk = VoltDB.instance().getHostMessenger().getZK();
         long start, duration = 0;
 
-        if (worksWithElastic == 0 && VoltZK.zkNodeExists(zk, VoltZK.rejoinInProgress)) {
-            throw new VoltAbortException("Can't do a catalog update while an elastic join is active");
+        if (worksWithElastic == 0 && VoltZK.zkNodeExists(zk, VoltZK.elasticJoinInProgress)) {
+            throw new VoltAbortException("Can't do a catalog update while an elastic join is active. Please retry catalog update.");
```
JadenW: Good catch, so we are checking wrong blocker before?

slowsoul: Yes, I think so, even before Yang's patch.

PR #4974, opened 2017-10-26
ENG-13235 dynamic sizing of view columns

yzhang1991:
Calculate the column size limit based on the number of columns.
So we can relax the size limit of view columns as much as possible.


src/frontend/org/voltdb/compiler/MaterializedViewProcessor.java
```diff
@@ -178,6 +178,7 @@ public void startProcessing(Database db, HashMap<Table, String> matViewMap, Tree
             // to transfer the message through the catalog, but
             // we can transmit the existence of the message.
             boolean isSafeForDDL = (stmt.getUnsafeMVMessage() == null);
+            int maximumDefaultColumnSize = DDLCompiler.MAX_ROW_SIZE / stmt.m_displayColumns.size();
```
wolffcm: This isn't quite the maximum, right?  Since some of the display columns could well have a fixed size.  I guess you want to take the max row size, subtract the widths of any fixed-length columns, and then divide that result by the number of variable length columns.

tests/frontend/org/voltdb/regressionsuites/TestGroupByComplexMaterializedViewSuite.java
```diff
@@ -1447,91 +1453,6 @@ public void testENG5386() throws Exception {
...
-
-        Client client = getClient();
-
-        // Inserts a large number of rows so that when truncate is performed on source table,
-        // we optimize by creating an empty table.
-        int numSourceRows = insertRowsForENG11941(client, true);
-
-        // Rows 0 is current max, row 1 has a value that exceeds view column width.
-
-        verifyProcFails(client, "exceeds the size", "R6.delete", 0);
```
wolffcm: So this whole test is being removed... but the behavior that it's characterizing isn't really gone, right?  The view column just has a different, larger size.  We should still be testing that when a delete fails because the underlying view can't contain the new MIN or MAX, that we produce expected behavior and don't crash or anything.  Your fix makes this harder to test, but we still need a test for it.

wolffcm:
Looks good.  Thanks for updating the test.

PR #4975, opened 2017-10-26
ENG-13263 printout transaction debug-level message for long running processes

zeemanhe:
Add transaction debug message to the rate limited logging for long running processes.

### DIFF
--- a/src/frontend/org/voltdb/iv2/InitiatorMailbox.java
+++ b/src/frontend/org/voltdb/iv2/InitiatorMailbox.java
@@ -265,7 +265,6 @@ public class InitiatorMailbox implements Mailbox
     {
         if (SCHEDULE_IN_SITE_THREAD) {
             this.m_scheduler.getQueue().offer(new SiteTasker.SiteTaskerRunnable() {
-                private String taskInfo = "";
                 @Override
                 void run() {
                     synchronized (InitiatorMailbox.this) {
@@ -275,10 +274,6 @@ public class InitiatorMailbox implements Mailbox
                 private SiteTasker.SiteTaskerRunnable init(VoltMessage message){
                     taskInfo = message.getMessageInfo();
                     return this;
-                }
-                @Override
-                public String getTaskInfo() {
-                    return taskInfo;
                 }
             }.init(message));
         } else {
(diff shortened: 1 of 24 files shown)


src/frontend/org/voltdb/iv2/InitiatorMailbox.java
```diff
@@ -265,13 +265,22 @@ public void deliver(final VoltMessage message)
...
             this.m_scheduler.getQueue().offer(new SiteTasker.SiteTaskerRunnable() {
+                private String taskInfo = "";
                 @Override
                 void run() {
                     synchronized (InitiatorMailbox.this) {
                         deliverInternal(message);
                     }
                 }
-            });
+                private SiteTasker.SiteTaskerRunnable init(VoltMessage message){
```
luyangco: Why not move `init()` and `getTaskInfo()` into base class `SiteTaskerRunnable`? If you doing so we'll a wider picture of what blocks site task queue, e.g. beyond Iv2 volt messages, durability checks, repair log truncate messages and some DR messages are running on site task queue.

src/frontend/org/voltdb/iv2/TickProducer.java
```diff
@@ -74,20 +77,23 @@ public void run()
...
         if (headOfQueueOfferTime != m_previousTaskTimestamp) {
             m_previousTaskTimestamp = headOfQueueOfferTime;
             m_previousTaskPeekTime = currentTime;
         } else if (currentTime - m_previousTaskPeekTime >= m_procedureLogThreshold) {
-            String fmt = " A process (procedure, fragment, or operational task) is taking a long time "
-                    + "-- over %d seconds -- and blocking the queue for site %d. "
-                    + "No other jobs will be executed until that process completes.";
             long waitTime = (currentTime - m_previousTaskPeekTime)/1_000_000_000L; // in seconds
-            m_logger.rateLimitedLog(SUPPRESS_INTERVAL, Level.INFO, null, fmt, waitTime, m_partitionId);
+            if (m_logger.isDebugEnabled()) {
```
luyangco: You can move line 83 down to here, so that in most case you don't have to create new strings.

src/frontend/org/voltdb/iv2/TickProducer.java
```diff
@@ -90,6 +88,7 @@ public void run()
         } else if (currentTime - m_previousTaskPeekTime >= m_procedureLogThreshold) {
             long waitTime = (currentTime - m_previousTaskPeekTime)/1_000_000_000L; // in seconds
             if (m_logger.isDebugEnabled()) {
```
luyangco: I am think make it a info-level message, because when something goes wrong this message help both customer and us to identify the cause of problem. Turning it into debug-level means in the most time we can only see it in our system test.

luyangco:
I think the log message should be info-level

PR #4983, opened 2017-10-31
ENG-13325: Fix possible StatsAgent hang when a new StatsSource set is registered in place of an old set

slowsoul:


src/frontend/org/voltdb/StatsAgent.java
```diff
@@ -760,13 +782,26 @@ private VoltTable getStatsAggregateInternal(
...
 
-        for (Entry<Long, NonBlockingHashSet<StatsSource>> entry : siteIdToStatsSources.entrySet()) {
-            NonBlockingHashSet<StatsSource> statsSources = entry.getValue();
-            //The window where it is empty exists here to
+        // make a copy of the site Ids
+        Set<Long> siteIds = new HashSet<>(siteIdToStatsSources.keySet());
+        for (long siteId : siteIds) {
+            NonBlockingHashSet<StatsSource> statsSources = siteIdToStatsSources.get(siteId);
+            // entries of this site id to sources set map may be removed in another thread
+            if (statsSources == null) {
```
manjujames: Better if we change the loop condition in the while loop in the next block as below?
while (statsSources==null || statsSources.isEmpty())

That will make sure that we do the yield if it is empty or null. The way this is written now, we will not yield if it is null.

PR #4984, opened 2017-11-01
ENG-13194  Fix failure to shutdown a node with@StopNode

zeemanhe:
When MP sites are in the middle of transaction processing, @StopNode  may cause MP deadlock and fail to stope a node. The pull request tries to terminate these transactions with a poison fragment response and then shutdown  MP Ro site pool.  and also offer a fake task to unblock sites to be shutdown.

### DIFF
--- a/src/frontend/org/voltdb/iv2/BaseInitiator.java
+++ b/src/frontend/org/voltdb/iv2/BaseInitiator.java
@@ -199,8 +199,6 @@ public abstract class BaseInitiator implements Initiator
 
         if (m_siteThread != null) {
             try {
-                //add a null task to unblock the queue
-                m_scheduler.getQueue().offer(Scheduler.m_nullTask);
                 m_siteThread.join();
             } catch (InterruptedException e) {
                 tmLog.info("Interrupted during shutdown", e);
--- a/src/frontend/org/voltdb/iv2/BaseInitiator.java
+++ b/src/frontend/org/voltdb/iv2/BaseInitiator.java
@@ -199,6 +199,8 @@ public abstract class BaseInitiator implements Initiator
 
         if (m_siteThread != null) {
             try {
+                //add a null task to unblock the queue
+                m_scheduler.getQueue().offer(Scheduler.m_nullTask);
                 m_siteThread.join();
             } catch (InterruptedException e) {
                 tmLog.info("Interrupted during shutdown", e);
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/iv2/BaseInitiator.java
```diff
@@ -199,6 +199,8 @@ public void shutdown()
 
         if (m_siteThread != null) {
             try {
+                //add a null task to unblock the queue
+                m_scheduler.getQueue().offer(Scheduler.m_nullTask);
```
luyangco: SpScheduler already does that in `SpScheduler.shutdown()`,  you just need to offer a null task to MpScheduler.

MpScheduler owns `m_tasks` which can be used to offer null task.

src/frontend/org/voltdb/iv2/MpRoSite.java
```diff
@@ -95,6 +95,8 @@
     // Current topology
     int m_partitionId;
 
+    private TransactionState m_txnState = null;
```
luyangco: Maybe add a comment to mention about what does it mean when `m_txnState` is not null

src/frontend/org/voltdb/iv2/MpRoSitePool.java
```diff
@@ -221,7 +227,8 @@ boolean doWork(long txnId, TransactionTask task)
                             m_poolThreadFactory));
             }
             site = m_idleSites.pop();
-            m_busySites.put(txnId, site);
+            busySites.put(txnId, site);
+            m_busySites.set(busySites);
```
luyangco: Before we merge this, we should measure the read-only MP queries performance.

http://ci/job/performance-graphs/lastSuccessfulBuild/artifact/performance-graphs-90-days/stats-index.html#ProjectionMP-cached%202%20nodes

luyangco: What if while the MpScheduler is shutting down, a task reaches to MpRoSite, pop a site out from `m_idleSites` but before it put into `m_busySites`. Shutdown path could miss this one in both idle and busy sites, causing an incomplete shutdown.

I think the right way to do is to synchronize both here and shutdown path.

zeemanhe: MpRoSitePool will be locked down and accept no more work upon shutting down. So this should not be an issue.

luyangco:
Please remember to run the read-only Mp performance test before merge. Otherwise it looks good to me.

PR #4990, opened 2017-11-02
ENG-13319: Populate DR information for new partitions if recovering from a join snapshot

slowsoul:


src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -208,6 +208,8 @@ public void run() {
                         jsObj.put(SnapshotUtil.JSON_PATH_TYPE, m_snapshotToRestore.pathType);
                         jsObj.put(SnapshotUtil.JSON_NONCE, m_snapshotToRestore.nonce);
                         jsObj.put(SnapshotUtil.JSON_IS_RECOVER, true);
+                        jsObj.put(SnapshotUtil.JSON_PARTITION_COUNT, m_snapshotToRestore.partitionCount);
+                        jsObj.put(SnapshotUtil.JSON_NEW_PARTITION_COUNT, m_snapshotToRestore.newPartitionCount);
```
nshi: Aren't these information in the snapshot to restore itself? Why do you need to pass it into the @SnapshotRestore?

slowsoul: My thinking was that newPartitionCount is in the digest, which is available in the digest scan fragment, but partition count is only in table save file which is not read until the table restore fragment. Since we need to populate these placeholder dr sequence numbers for new partitions in digest distribute fragment, which is before table restore fragment. And considering that restore agent has all these information before it invokes @SnapshotRestore in SnapshotInfo Object, and the cost to have 2 more fields in the JSON string parameter to @SnapshotRestore is quite small, I chose to pass them here so that the logic of @SnapshotRestore don't need to change much.

slowsoul: Overall, I just think this is much cleaner than needle through the fragments to collect information and possibly need to adjust some order. Since these information is already retrieved and verified before @SnapshotRestore is invoked, also maybe doing some redundant work to figure them out again in the sysproc.

PR #4993, opened 2017-11-02
Add "volatile" status bit to NValue and TableTuple

wolffcm:
I added a "volatile" status bits to `NValue` and `TableTuple` that is set depending on context:
- For regular tuples in persistent or regular temp tables, nothing is marked as volatile
- For stand-alone tuples (`table->tempTuple()`, `StandAloneTupleStorage` and `PoolBackedTupleStorage`) variable-length inlined data is marked as volatile
- For tuples in large temp tables, both inlined and non-inlined variable-length data is marked as volatile

This change also strengthen's `NValue`'s abstraction since clients don't really care how/where data is allocated, but they do care if a value might unexpectedly change or become invalid, i.e., if it's volatile.  For large temp tables especially, it makes sense to separate these two concerns.

Failing to make copies of inlined values has historically been a source of bugs, and these changes affect normal execution, so I'm qualifying this work on a separate branch.


src/ee/executors/windowfunctionexecutor.cpp
```diff
@@ -140,7 +140,7 @@ struct WindowAggregate {
     {
         m_value.castAs(type);
         if (m_inlineCopiedToNonInline) {
-            m_value.allocateObjectFromNonInlinedValue();
+            m_value.allocateObjectFromPool();
```
BillWhite: There are some other pool backed tuples in windowfunctionexecutor, used to scan tables.  I wouldn't think those need this treatment, but maybe they do.  Should you look at them?  They are all the objects of type PoolBackedTupleStorage in the windowfunctionexecutor.h file.

wolffcm: They look fine.  Inlined values extracted from them will be marked as volatile which is exactly what should happen.  There are tests of windowed MIN/MAX with inlined variable-length columns, right?

BillWhite: Yes, I think so.

PR #5001, opened 2017-11-07
ENG-13341: Add more comment to elastic join blockers

JadenW:


src/frontend/org/voltdb/VoltZK.java
```diff
@@ -110,10 +110,16 @@
      */
     public static final String actionBlockers = "/db/action_blockers";
     // being able to use as constant string
+
+    // two elastic join blockers
+    // elasticJoinInProgress blocks the rejoin (create in init state, release before data migration start)
+    // banElasticJoin blocks elastic join (currently create in DRProducer connect to a cluster has version before v7.8,
```
slowsoul: Maybe the more precise explanation for our current implementation is that "currently created by DRProducer if the agreed protocol version for the mesh does not support elastic join during DR (i.e. version <= 7), it is now only released after a DR global reset"
Currently if we have version 8 and version 7 in the mesh and all version 7 clusters are reset, the remaining version 8 clusters will still talk with version 7, they do not upgrade protocol version automatically for now.

JadenW: Thanks, good catch

PR #5003, opened 2017-11-07
Eng 12926 Don't reverse hole loops.

BillWhite:
In the OGC polygon format, WKT, we represent a polygon as a counter clockwise shell followed by zero or more clockwise holes.  In S2 these are all counter clockwise, and S2 maintains the nesting relationship in a different way.  This means we need to reorder vertices in polygons to send them to S2.  In the Java GeographyValue class we represent polygons using the S2 conventions, but the constructor uses the OGC conventions.  This works alright except for the GeographyValue.add function.  This function is internal, and operates on the loops directly from the internal arrays.   Consequently, the add function does not need to reverse the order of holes.  In this commit we make sure this happens.

### DIFF
--- a/src/frontend/org/voltdb/types/GeographyValue.java
+++ b/src/frontend/org/voltdb/types/GeographyValue.java
@@ -88,6 +88,24 @@ public class GeographyValue {
      * @param rings  A list of lists of points that will form a polygon.
      */
      public GeographyValue(List<List<GeographyPointValue>> rings) {
+         this(rings, false);
+     }
+
+     /**
+      * Create a polygon from a list of rings.  This is internal.  If the
+      * argument ringsAreInS2Order is true then we will not reorder any holes.
+      * Since the OGC text format, WKT, used CCW shells and CW holes, and S2
+      * uses all CCW loops, we need to reverse holes for S2, and this is the
+      * vertex order we use internally.  But if this is an operation on rings
+      * which are already in S2 order, say for the polynomial creation or
+      * algebra operations, we don't want to reorder the loops again.  The
+      * argument holesAreInS2Order is true in this case.
+      *
+      * @param rings
+      * @param holesAreInS2Order
+      */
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/types/GeographyValue.java
```diff
@@ -773,15 +786,15 @@ else if (token != ',') {
      * @return  The resulting GeographyValue.
      */
     public GeographyValue add(GeographyPointValue offset) {
-        List<List<GeographyPointValue>> newLoops = new ArrayList<List<GeographyPointValue>>();
+        List<List<GeographyPointValue>> newLoops = new ArrayList<>();
```
wolffcm: I would have thought that this method is deprecated.  Should it be?

tests/frontend/org/voltdb/types/TestGeographyValue.java
```diff
@@ -172,6 +172,7 @@ public void testGeographyValuePositive() {
         String rtStr = "POLYGON ((0.0 20.0, -17.320508076 -10.0, 17.320508076 -10.0, 0.0 20.0))";
         rtGeog = new GeographyValue(rtStr);
         assertEquals(rtStr, rtGeog.toString());
+        assertEquals(rtGeog, new GeographyValue(rtGeog.toWKT()));
```
wolffcm: The original ticket made use of the `add` method.  But I don't see that you added any call to `add`.  Does this extra check actually confirm the fix?

src/frontend/org/voltdb/types/GeographyValue.java
wolffcm: If there are other methods that should be deprecated, piggybacking on your fix here would be a good way to do it.

tests/frontend/org/voltdb/types/TestGeographyValue.java
BillWhite: I added another test.

PR #5012, opened 2017-11-09
ENG-13369 make fragment response timeout configurable

zeemanhe:
add a system property to make fragment response timeout configurable

### DIFF
--- a/src/frontend/org/voltdb/iv2/MpTransactionState.java
+++ b/src/frontend/org/voltdb/iv2/MpTransactionState.java
@@ -77,6 +77,9 @@ public class MpTransactionState extends TransactionState
     boolean m_haveDistributedInitTask = false;
     boolean m_isRestart = false;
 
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -88,6 +91,7 @@
         m_masterHSIds.putAll(partitionMasters);
         m_buddyHSId = buddyHSId;
         m_isRestart = isRestart;
+        m_pullTimeout = Long.valueOf(System.getProperty("MP_TXN_RESPONSE_TIMEOUT", "5")) * 60L;
```
luyangco: Please also update https://wiki.voltdb.com/display/Partner/Hidden+Options+for+VoltDB after you merge.

akhanzode: This should be a static final as opposed to parsing and reading system property every time.

PR #5014, opened 2017-11-09
ENG-13363 Add property for kerberos authentication to JDBC driver

benjaminballard:
To enable kerberos authentication for the VoltDB JDBC driver, set the JDBC property "kerberos=VoltDBClient". This will use the "VoltDBClient" configuration entry id in your JAAS file, similar to sqlcmd and java clients.


akhanzode:
Review with Andrew on property name. Also check how other jdbc driver does it. If its consistent easy to understand ....good.

PR #5015, opened 2017-11-09
Eng 12889 nibble delete sysproc

luyangco:
The sysproc copies pattern from windowing app, it has three statements:
1) select numbers of rows that meet the delete criteria,
2) find the value sits on boundary of delete chunk size, if rows to be deleted are more than chunk size,
3) delete rows based on the value given in parameter, or the boundary value.

It's mostly ad-hoc style, compiles a plan based on replicated/partitioned table and column every time.
Maybe I should also cache the plan in LoadedProcedureSet, like CRUD procedure.


xinjiacs:
Maybe I missed the NibbleDelete discussion, but ticket ENG-12889 does not mention any NibbleDelete. How different it is with ENG-12877? 

I thought run-everywhere API for NT procedure is like adding a new API inside `VoltNonTransactionalProcedure` like `callNTProcedureOnAllHosts`. Instead of calling the procedure on every host, this new API should call a procedure or AdHoc query on every partition.

luyangco:
Pull request moved to https://github.com/VoltDB/voltdb/pull/5024

PR #5018, opened 2017-11-10
ENG-13373. Make licensecheck verify existence of jars.

rmorgenstein:
Added check for hardcoded library versions referenced in template.classpath and build-importers.xml.


rmorgenstein:
Never mind. I think I found a better solution.

PR #5020, opened 2017-11-10
Added exec Stored Procedures to SQL-grammar-generator tests

johnTcrawford:


wolffcm:
I like what you've done here with the system stored procedures, especially swap tables, and the explain variants.  It's good to have calls to the other system stored procedures as well, with random arguments.

For the user stored procedures, the randomization is really coming from passing parameters, right?  We have JUnit tests that do this (I don't think they randomize, but they do test "interesting" values like min or max for the data type, etc).  If you're not randomly generated the user stored procedures themselves (i.e., doing random generation of Java code), then I think this kind of testing is best done in JUnits.  (Not saying you should be randomly generating Java code for this ticket!  That would be a big challenge.)

johnTcrawford:
Chris, your comment (especially the second paragraph) is similar to what Ruth said to me verbally. As you can see, I made several additional changes after the original commit, most of which were about getting run.sh or other features to work; but one of them addresses your concerns to a degree. I did not delete any of the pre-canned (Java or SQL user-defined, or default) Stored Procedures, but I did significantly decrease their probability, in favor of increasing the probability of the random (only SQL, for now) user-defined Stored Procs (which were Ruth's suggestion, BTW). As you said, random Java Stored Procs are beyond the scope of this task, but might be worth doing at some point (perhaps Innovation Week?), and the canned ones in this task could provide a starting point for those.

PR #5024, opened 2017-11-13
Eng 12877 nibbledelete sysproc

luyangco:
The sysproc copies pattern from windowing app, it has three statements:

select numbers of rows that meet the delete criteria, find the value sits on boundary of delete chunk size, if rows to be deleted are more than chunk size, delete rows based on the value given in parameter, or the boundary value.

### DIFF
--- a/src/frontend/org/voltdb/compiler/StatementCompiler.java
+++ b/src/frontend/org/voltdb/compiler/StatementCompiler.java
@@ -25,11 +25,13 @@ import java.util.TreeSet;
 import org.hsqldb_voltpatches.HSQLInterface;
 import org.hsqldb_voltpatches.VoltXMLElement;
 import org.voltcore.logging.VoltLogger;
+import org.voltdb.CatalogContext;
 import org.voltdb.CatalogContext.ProcedurePartitionInfo;
 import org.voltdb.VoltDB;
 import org.voltdb.VoltType;
 import org.voltdb.catalog.Catalog;
 import org.voltdb.catalog.CatalogMap;
+import org.voltdb.catalog.Column;
 import org.voltdb.catalog.Database;
 import org.voltdb.catalog.Function;
 import org.voltdb.catalog.PlanFragment;
@@ -50,6 +52,7 @@ import org.voltdb.plannodes.DeletePlanNode;
 import org.voltdb.plannodes.InsertPlanNode;
 import org.voltdb.plannodes.PlanNodeList;
 import org.voltdb.plannodes.UpdatePlanNode;
+import org.voltdb.sysprocs.NibbleDeleteBase.ComparisonConstant;
 import org.voltdb.types.QueryType;
 import org.voltdb.utils.BuildDirectoryUtils;
 import org.voltdb.utils.CatalogUtil;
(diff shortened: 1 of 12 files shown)


src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
```diff
@@ -0,0 +1,203 @@
...
+        Procedure newCatProc = StatementCompiler.compileNibbleDeleteProcedure(catTable,
+                this.getClass().getName() + "-" + tableName, column, comparisonConstant);
+
+
+        Statement countStmt = newCatProc.getStatements().get(VoltDB.ANON_STMT_NAME + "0");
+        Statement deleteStmt = newCatProc.getStatements().get(VoltDB.ANON_STMT_NAME + "1");
+        Statement valueAtStmt = newCatProc.getStatements().get(VoltDB.ANON_STMT_NAME + "2");
+        if (countStmt == null || deleteStmt == null || valueAtStmt == null) {
+            throw new VoltAbortException(
+                    String.format("Unable to find SQL statement for found table %s: BAD",
```
xinjiacs: This is not possible to happen, right?

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Execute a set of queued inserts. Ensure each insert successfully
+     * inserts one row. Throw exception if not.
+     *
+     * @return Count of rows inserted or upserted.
+     * @throws VoltAbortException if any failure at all.
+     */
+    VoltTable executeSQLOnTheFly(Statement catStmt, Object[] params) throws VoltAbortException {
+        // Create a SQLStmt instance on the fly
+        // This unusual to do, as they are typically required to be final instance variables.
```
xinjiacs: This is unusual to do?

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
```diff
@@ -0,0 +1,203 @@
...
+                            actualType.toString(), expectedType.toString()));
+        }
+
+        if (comparison < ComparisonConstant.GREATER_THAN.ordinal() ||
+                comparison > ComparisonConstant.EQUAL.ordinal()) {
+            throw new VoltAbortException("Invalid comparison constant: " + comparison);
+        }
+        ComparisonConstant comparisonConstant = ComparisonConstant.values()[comparison];
+
+        // TODO: should cache the plan in somewhere
```
xinjiacs: Yes, every invocation will trigger the planning and the planning function is synchronized. 

Also, does it mean we create a Procedure for every invocation as well? Anything we need to worry about?

src/frontend/org/voltdb/compiler/StatementCompiler.java
```diff
@@ -622,4 +625,155 @@ public static Procedure compileDefaultProcedure(PlannerTool plannerTool, Procedu
...
+            Column col, ComparisonConstant comp) {
+        Procedure newCatProc = addProcedure(catTable, procName);
+
+        String countingQuery = genSelectSql(catTable, col, comp);
+        addStatement(catTable, newCatProc, countingQuery, "0");
+
+        String deleteQuery = genDeleteSql(catTable, col, comp);
+        addStatement(catTable, newCatProc, deleteQuery, "1");
+
+        String valueAtQuery = genValueAtOffsetSql(catTable, col);
```
xinjiacs: So this is not the ben's midpoint guessing strategy. I am ok as the initial version.

src/frontend/org/voltdb/sysprocs/NibbleDeleteMP.java
```diff
@@ -0,0 +1,191 @@
...
+
+    public DependencyPair executePlanFragment(
+            Map<Integer, List<VoltTable>> dependencies, long fragmentId,
+            ParameterSet params, SystemProcedureExecutionContext context) {
+        // Never called, we do all the work in run()
+        return null;
+    }
+
+    /**
+     * Execute a set of queued inserts. Ensure each insert successfully
```
xinjiacs: Similar comments as the NibbleDeleteSp.

xinjiacs:
Although these two sysprocs are not intended for external usage, we may put these logics directly into the Hugg's NT procedure @NibbleDelete, and use @AdHoc system proc directly. In this way, it handles the query planning and cache issue, making it a simplified and flexible solution. 

Importantly, we are trying to compute the next batch timestamp point for delete query, it does not need to be accurate for the middle value. For each partition, the NT procedure will do 1) select min(timestamp), count(*)... , 2) guessing a point value 3) run delete query with that point value. 

Maybe we should wait for @jhugg , he may have more thoughts on the design.

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
```diff
@@ -0,0 +1,203 @@
...
+
+        VoltType expectedType = VoltType.values()[column.getType()];
+        VoltType actualType = VoltType.typeFromObject(params[0]);
+        if (actualType != expectedType) {
+            throw new VoltAbortException(
+                    String.format("Parameter type %s doesn't match column type %s",
+                            actualType.toString(), expectedType.toString()));
+        }
+
+        if (comparison < ComparisonConstant.GREATER_THAN.ordinal() ||
```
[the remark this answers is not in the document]
luyangco: Sure.

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
luyangco: NibbleDelete can be treated like default procedure. Procedure is cached at defaultProcCache in each site.

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
luyangco: I don't think it's possible to fail here, but just a bullet-proof vest.

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
```diff
@@ -0,0 +1,203 @@
...
+
+    public DependencyPair executePlanFragment(
+            Map<Integer, List<VoltTable>> dependencies, long fragmentId,
+            ParameterSet params, SystemProcedureExecutionContext context) {
+        // Never called, we do all the work in run()
+        return null;
+    }
+
+    /**
+     * Execute a set of queued inserts. Ensure each insert successfully
```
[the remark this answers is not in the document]
luyangco: Will change it.

src/frontend/org/voltdb/sysprocs/NibbleDeleteMP.java
luyangco: Will change it.

src/frontend/org/voltdb/compiler/StatementCompiler.java
luyangco: If we solve the performance issue in query
```sql 
select * from t order by ts offset N limit 1
```
Then we don't need to use midpoint guessing strategy.

src/frontend/org/voltdb/sysprocs/NibbleDeleteSP.java
```diff
@@ -0,0 +1,203 @@
...
+     * @param table       value to compare
+     * @param chunksize   maximum number of rows allow to be deleted
+     * @return how many rows are deleted and how many rows left to be deleted (if any)
+     */
+    public VoltTable run(SystemProcedureExecutionContext ctx, int partitionParam, String tableName, String columnName,
+            int comparison, VoltTable table, long chunksize) {
+        // Some basic checks
+        Table catTable = ctx.getDatabase().getTables().getIgnoreCase(tableName);
+        if (catTable == null) {
+            throw new VoltAbortException("Table not present in catalog");
```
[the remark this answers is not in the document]
luyangco: Will do.

src/frontend/org/voltdb/sysprocs/NibbleDeleteBase.java
```diff
@@ -0,0 +1,222 @@
...
+
+        voltQueueSQL(stmt, params);
+        return voltExecuteSQL()[0];
+    }
+
+    VoltTable nibbleDeleteCommon(SystemProcedureExecutionContext ctx,
+                                 String tableName,
+                                 String columnName,
+                                 int comparison,
+                                 VoltTable table,
```
xinjiacs: Rename "table" to "paramTable"?

src/frontend/org/voltdb/sysprocs/NibbleDeleteBase.java
```diff
@@ -0,0 +1,222 @@
...
+        // so far should only be single column, single row table
+        int columnCount = table.getColumnCount();
+        if (columnCount > 1) {
+            throw new VoltAbortException("More than one input parameter is not supported right now.");
+        }
+        assert(columnCount == 1);
+        table.resetRowPosition();
+        table.advanceRow();
+        Object[] params = new Object[columnCount];
+        for (int i = 0; i < columnCount; i++) {
```
xinjiacs: I am not sure how much value left to keep "columnCount" other than assuming 1. The other places does assume that already. 

This API already works for the simple use case and TTL later, maybe changing the parameter voltdb to be an java object as well for simplicity. The harder AdHoc Delete case will definitely need other solution.

src/frontend/org/voltdb/sysprocs/NibbleDeleteBase.java
```diff
@@ -0,0 +1,222 @@
...
+        result = executePrecompiledSQL(countStmt, params, replicated);
+        long rowCount = result.asScalarLong();
+        // If number of rows meet the criteria is more than chunk size, pick the column value
+        // which offset equals to chunk size as new predicate.
+        // TODO:how to delete rows match exactly to the chunk size while not causing the
+        // non-deterministic issue?
+        // Please be noted that it means rows be deleted can be more than chunk size, normally
+        // data has higher cardinality won't exceed the limit a lot,  but low cardinality data
+        // might cause this procedure to delete a large number of rows than the limit chunk size.
+        if (op != ComparisonConstant.EQUAL) {
```
xinjiacs: Also, if we can assume the length of "params" is 1, we can simply the following delete stmt as:
```java
        Object paramValue = params[0];
        if (op == ComparisonConstant.EQUAL || rowCount <= chunksize) {
            cutoffValue = paramValue;
        } else {
            Object[] newParams = new Object[]{paramValue, chunksize - 1};

            result = executePrecompiledSQL(valueAtStmt, newParams, replicated);
            cutoffValue = result.fetchRow(0).get(0, actualType);
        }
        result = executePrecompiledSQL(deleteStmt, new Object[] {cutoffValue}, replicated);
```

src/frontend/org/voltdb/compiler/StatementCompiler.java
```diff
@@ -622,4 +625,169 @@ public static Procedure compileDefaultProcedure(PlannerTool plannerTool, Procedu
         fragment.setPlannodetree(bin64String);
         return jsonBytes;
     }
+
+    private static String genSelectSql(Table table, Column column,
```
xinjiacs: As these functions are not intended for general use, maybe add "NibbleDelete" to these function names for clarification?

tests/frontend/org/voltdb/sysprocs/TestNibbleDelete.java
```diff
@@ -0,0 +1,430 @@
...
+        }
+
+        pair = new Pair<>(0l, Long.MAX_VALUE); // <deleted, toBeDeleted>
+        loop = 0;
+        existingRows = numberOfItems;
+        while (pair.getSecond() > 0) {
+            pair = nibbleDeleteReplicated(op, ts, existingRows);
+            existingRows -= pair.getFirst();
+            if (++loop > 100) {
+                fail("Make no progress on delete, something wrong happens in @NibbleDeleteSP");
```
xinjiacs: NibbleDeleteMP?

tests/frontend/org/voltdb/sysprocs/TestNibbleDelete.java
```diff
@@ -0,0 +1,430 @@
...
+                    table,
+                    500);
+            VoltTable result = response.getResults()[0];
+            assertEquals(1, result.getRowCount());
+            result.advanceRow();
+            deleted = result.getLong("DELETED_ROWS");
+            long leftoverRows = result.getLong("LEFTOVER_ROWS");
+            toBeDeleted += leftoverRows;
+        } catch (ProcCallException e) {
+            fail("Fail to run NibbleDeleteSP: " + e.getMessage());
```
xinjiacs: NibbleDeleteMP?

src/frontend/org/voltdb/compiler/StatementCompiler.java
```diff
@@ -622,4 +625,169 @@ public static Procedure compileDefaultProcedure(PlannerTool plannerTool, Procedu
...
+        sb.append(" WHERE " + column.getName() + " " + comparison.toString() + " ?;");
+        return sb.toString();
+    }
+
+    private static String genValueAtOffsetSql(Table table, Column column,
+            ComparisonConstant comparison) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("SELECT " + column.getName() + " FROM " + table.getTypeName());
+        sb.append(" WHERE " + column.getName() + " " + comparison.toString() + " ?");
+        sb.append(" ORDER BY " + column.getName());
```
xinjiacs: Does this still mean doing an index scan anyway?

tests/frontend/org/voltdb/sysprocs/TestNibbleDelete.java
```diff
@@ -0,0 +1,430 @@
...
+    {
+        System.out.println("testSelectByOffsetOnIndexColumn");
+
+        int numberOfItems = 100000;
+        VoltTable part_table = createTable(numberOfItems, 0, 0, 0, 0);
+        loadTable(m_client, "part", false, part_table);
+
+        String query = "select * from part order by ts offset 1000 limit 1";
+        CatalogContext context = VoltDB.instance().getCatalogContext();
+        PlannerTool plannerTool = context.m_ptool;
```
xinjiacs: query and plannerTool not used?

PR #5029, opened 2017-11-14
ENG-13386: Synchronize partition leader zk names map.

akhanzode:


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -157,7 +157,7 @@ public void processResult(int rc,
     private static final ListeningExecutorService m_childUpdatingThread =
             CoreUtils.getListeningExecutorService("Export ZK Watcher", 1);
 
-    private final Map<Integer, String> m_partitionLeaderZKName = new HashMap<Integer, String>();
+    private final Map<Integer, String> m_partitionLeaderZKName = Collections.synchronizedMap(new HashMap<Integer, String>());
```
zeemanhe: From line 120, should the iteration block be synchronized with m_partitionLeaderZKName?

xinjiacs:
@akhanzode Can you explain the thread model here? What threads are competing to access collection `m_partitionLeaderZKName `?

src/frontend/org/voltdb/export/ExportGeneration.java
akhanzode: @xinjiacs The leader election happen at start and at UAC time. Other node can cause drain to start on a generation which can kick off m_onSourceDrained on its own executor so not serialized. Looking at NPE from failed system test it looks like race between put and get.
@zeemanhe I am guarding the deletes as well.

xinjiacs: I am so lucky that these generations will be gone soon.

PR #5031, opened 2017-11-14
ENG-13286: EE foundation for recursive CTE

wolffcm:
This set of changes just lays the foundation for handling recursive CTEs in the EE.  The EE unit test has some JSON that I've mocked up that should be pretty close to what the planner will produce.

The product changes here won't affect any user-visible behavior, and so are safe to merge to master.  They are:
- Added a new bit to the JSON for SeqScanPlanNode that says if the node scans a CTE.  This seems much easier than creating a new kind of scan plan node that needs to duplicate logic for filtering, projection, inlined insert or aggregation, etc.
- Changed logic in SeqScanPlanNode that allows for either subquery, persistent table, or CTE scan, using an enum to represent which it is.
- Added a new kind of plan node and executor, `RecursiveCtePlanNode` and `RecursiveCteExecutor`.  These don't do anything yet and cannot be produced by the planner.


tests/ee/execution/RecursiveCommonTableExpressionTest.cpp
```diff
@@ -0,0 +1,637 @@
...
+    "set $PREV elasticthroughput 2\n"
+    "set $PREV querytimeout 10000\n"
+    "add /clusters#cluster logconfig log\n"
+    "set /clusters#cluster/logconfig#log enabled false\n"
+    "set $PREV synchronous false\n"
+    "set $PREV fsyncInterval 200\n"
+    "set $PREV maxTxns 2147483647\n"
+    "set $PREV logSize 1024\n";
+
+const std::string jsonPlan =
```
BillWhite: What's the SQL for this plan?

tests/ee/execution/RecursiveCommonTableExpressionTest.cpp
```diff
@@ -0,0 +1,637 @@
...
+    "        {\"EXECUTE_LIST\": [\n"
+    "            10,\n"
+    "            12,\n"
+    "            9,\n"
+    "            8\n"
+    "        ]}\n"
+    "    ],\n"
+    "    \"IS_LARGE_QUERY\": false,\n"
+    "    \"PLAN_NODES_LISTS\": [\n"
+    "        {\n"
```
BillWhite: Is there ever a case when the execute lists is not the list of IDs of an element of the PLAN_NODE_LISTS in the order they are listed?  Could we just eliminate the EXECUTE_LISTS and execute the executors of the PLAN_NODE_LISTS in the order given?  Or maybe the reverse of the order given?  If the PLAN_NODE_LISTS are sorted so that each parent comes before all it's children then we could just execute them in reverse order.

Or, you could just have one list of plan nodes in one plan node list and keep the execute lists.  The tree structure of the plan nodes is given by ID anyway.  So, a fragment is a forest of plan trees.  There is one execute list for each plan tree, and one executor for each plan node, with some special care for inline plan nodes.  The execute lists tell what order we execute the executors for the plan node trees, taking care to execute children before parents.

src/ee/common/SerializableEEException.h
```diff
@@ -20,7 +20,7 @@
 
 #include <string>
 
-#define throwSerializableEEException(...) { char msg[8192]; snprintf(msg, 8192, __VA_ARGS__); throw voltdb::SerializableEEException(VOLT_EE_EXCEPTION_TYPE_EEEXCEPTION, msg); }
+#define throwSerializableEEException(...) do { char msg[8192]; snprintf(msg, 8192, __VA_ARGS__); throw voltdb::SerializableEEException(VOLT_EE_EXCEPTION_TYPE_EEEXCEPTION, msg); } while (false)
```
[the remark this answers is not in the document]
wolffcm: It could be.  It's been a while since I've written a printf-style varargs function... I took a stab at this and it turns out that just using macro with `...` is a lot easier.  I believe this is a gcc extension that clang also supports.  Nothing is broken here, so I think I'll leave it.

src/ee/common/types.cpp
```diff
@@ -460,6 +460,9 @@ string planNodeToString(PlanNodeType type)
     case PLAN_NODE_TYPE_WINDOWFUNCTION: {
         return "WINDOWFUNCTION";
     }
+    case PLAN_NODE_TYPE_RECURSIVECTE: {
+        return "RECURSIVECTE";
```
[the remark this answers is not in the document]
wolffcm: I can live with COMMONTABLE, as in `CommonTablePlanNode` and `CommonTableExecutor`.

src/ee/executors/abstractexecutor.cpp
```diff
@@ -110,8 +110,8 @@ bool AbstractExecutor::init(VoltDBEngine* engine,
         // If the target_table is NULL, then we need to ask the engine
         // for a reference to what we need
         // Really, we can't enforce this when we load the plan? --izzy 7/3/2010
-        bool is_subquery = (scan_node != NULL && scan_node->isSubQuery());
-        if (target_table == NULL && !is_subquery) {
+        bool isPersistentTableScan = (scan_node != NULL && scan_node->isPersistentTableScan());
+        if (target_table == NULL && isPersistentTableScan) {
```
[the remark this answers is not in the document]
wolffcm: If by recursive table you mean the output of CommonTableExecutor, I think that should be a temp table.  Persistent tables support indexes, compaction, etc in addition to actually persisting, whereas as the output of CommonTableExecutor only needs to live as long as the fragment is executing.

I'm still hashing out the details of set up and execution, so this is all subject to change.

tests/ee/execution/RecursiveCommonTableExpressionTest.cpp
wolffcm: It was in the ticket, I've added it to a comment in the test.

tests/ee/execution/RecursiveCommonTableExpressionTest.cpp
wolffcm: It's not that simple, because some of the execute lists are executed multiple times, in the case of subqueries with correlated parameters.  So they really need to be distinct.

We could just list the plan nodes in the order that they need to be executed though.  That would require some changes to the planner to ensure that it does this.

tests/ee/execution/CommonTableExpressionTest.cpp
```diff
@@ -0,0 +1,649 @@
...
+    "    ]\n"
+    "}\n";
+
+TEST_F(CommonTableExpressionTest, Basic) {
+    UniqueEngine engine = UniqueEngineBuilder().build();
+    bool success = engine->loadCatalog(0, catalogPayload);
+    ASSERT_TRUE(success);
+
+    auto ev = ExecutorVector::fromJsonPlan(engine.get(), jsonPlan, 0);
+    ASSERT_NE(NULL, ev.get());
```
xinjiacs: Other than validating non-null vector, is it better to validate the CTE type in the json plan?

wolffcm: Yes, that would definitely be better.  I will add some checks on the content of the vector.

wolffcm:
Merged on another branch.

PR #5050, opened 2017-11-21
Eng 13016 kafka apprunner enhancments

jwiebema1:
make sure export has drained before counting importer rows.
add options to enable topologyawareness and security(username/password)


tests/test_apps/kafkaimporter/client/kafkaimporter/KafkaImportBenchmark.java
```diff
@@ -461,15 +476,22 @@ public static void main(String[] args) throws Exception {
         long exportRowCount = 0;
         if (config.useexport) {
             exportRowCount = MatchChecks.getExportRowCount(client);
+            // make sure the export table has drained
+            while ( ! MatchChecks.isExportDrained(client) ) {
+                log.info("Waiting for export table to drain");
+                Thread.sleep(2000);
+            }
```
prosegay: AFAIK there are bugs in export that result in not draining that we see in txnid2 tests (when nodes are killed/rejoined?) so this could end up looping forever if this test hits that condition. Put in a timeout, fail the client.

PR #5052, opened 2017-11-27
ENG-13401 clean up  http sessions after catalog update

zeemanhe:


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3470,7 +3470,7 @@ public CatalogContext catalogUpdate(
                 // 3. update HTTPClientInterface (asynchronously)
                 // This purges cached connection state so that access with
                 // stale auth info is prevented.
-                if (m_adminListener != null)
+                if (m_adminListener != null && m_catalogContext.authSystem.isSecurityEnabled())
```
akhanzode: You need to notify regardless of security enabled or not. If security was enabled and then flipped to disabled old session cookies will give you old authResult thus failing those calls.

src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -367,4 +360,13 @@ public void notifyOfCatalogUpdate() {
             m_log.error("Failed to update HTTP interface after catalog update", ex);
         }
     }
+
+    private static class HttpSessionIdManager extends HashSessionIdManager {
+        @Override
+        public void doStop() throws Exception {
+            synchronized (this) {
```
akhanzode: This does not block other jetty threads from getting old session information as they dont block. How does this class help?

src/frontend/org/voltdb/RealVoltDB.java
zeemanhe: Will revert back the change. Are there any use cases that the security is on and off?

src/frontend/org/voltdb/utils/HTTPAdminListener.java
zeemanhe: Correct. It does not block getting old sessions.  The intention is to block  the session addition and removal from the session cache.  Should getting session be blocked too? The old sessions seem still valid.

PR #5058, opened 2017-11-30
ENG-13396: Sporadic failure in TestJSONInterface

wolffcm:
These changes fix sporadic failures in the test TestJSONInterface which were the result of planning of multiple statements happening at the same time.  This test runs a bunch of ad hoc queries in parallel with catalog updates, and so it caused a failed assertion on the expected value of a static variable used in planning, `ParameterizationInfo.curParamIndex`.

The `QueryPlanner` method `plan` was only invoked serially, but the method `parameterize`, which is called for ad hoc planning, was not guarded with a `synchronize` block.  The calls to `parameterize` and `plan` must happen with no other planning happening in between.  This caused a data race on the value of `curParamIndex`.

My fix was to force creation of an instance of `QueryPlanner` to take out a global lock to ensure that planning happens serially.

### DIFF
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -297,6 +297,8 @@ public class PlannerTool {
                 if (plan.getStatementPartitioning() != null) {
                     partitioning = plan.getStatementPartitioning();
                 }
+
+                planHasExceptionsWhenParameterized = planner.wasBadPameterized();
             }
             catch (Exception e) {
                 /*
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -59,6 +59,8 @@ public class PlannerTool {
 
     private static PlannerStatsCollector m_plannerStats;
 
+    private static final Object PLANNER_LOCK = new Object();
+
     public PlannerTool(final Database database, byte[] catalogHash)
     {
         assert(database != null);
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -129,20 +129,19 @@ private void logException(Exception e, String fmtLabel) {
...
-        QueryPlanner planner = new QueryPlanner(
-            sql, "PlannerTool", "PlannerToolProc", m_database,
-            partitioning, m_hsql, estimates, !VoltCompiler.DEBUG_MODE,
-            costModel, null, null, DeterminismMode.FASTER, false);
 
         CompiledPlan plan = null;
-        try {
+        // This try-with-resources block acquires a global lock on all planning
+        // This is required until we figure out how to do parallel planning.
+        try (QueryPlanner planner = new QueryPlanner(
```
BillWhite: Well, you learn something new every day, or at least I do.  I didn't know one could do this.

It's a little bit like Java and C++ convergence, but not through a fun house mirror somehow.

Do we have tests of subqueries?  Are these locks recursive locks?  That is, if a thread owns a
lock, L, can it lock L a second time?  Are the locks counting locks?  I think Java locking is mostly
recursive, so we are probably safe.

wolffcm: Instances of QueryPlanner are only created in the places you see here in this pull request, so subqueries will not acquire the PLANNER_LOCK an additional time.  Subqueries use lower-level classes to do their planning.  QueryPlanner is the top-level class invoked from `@AdHoc` and the store procedure compiler.

Even if they did, the PLANNER_LOCK is an instance of ReentrantLock, which allows the same thread to acquire a lock multiple times since it maintains a hold count:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html

PR #5059, opened 2017-12-01
ENG-13394 skip importers which have no procedures configured

zeemanhe:


src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1410,7 +1410,10 @@ public void mergeProperties(Properties props) {
         @SuppressWarnings("unchecked")
         public boolean checkProcedures(CatalogContext catalogContext, VoltLogger importLog, String configName) {
             String procedure = m_moduleProps.getProperty(ImportDataProcessor.IMPORT_PROCEDURE);
-            assert procedure != null;
+            if (procedure == null) {
```
akhanzode: Why not push this check into catalogContext and then rest will get logged and ignored.

zeemanhe: added testImporterWithMissingProcedure in TestImportSuite

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -1410,7 +1410,10 @@ public void mergeProperties(Properties props) {
...
+                importLog.info("Importer " + configName + " has no procedures. The importer will be disabled.");
```
akhanzode: Can you review message with Andrew. Rest looks good.

PR #5061, opened 2017-12-04
sql_grammar_generator.py: fixed the previous not-quite-right fix of

johnTcrawford:
handling 'show classes' commands; improved the handling of the various
types of errors (Java Exception, ERROR, WARN, etc.) grep-ed in the log
(and/or console output); added a new 'suffix' option, which adds a
suffix to the name of each output file (useful when running locally
repeatedly, so they don't clobber the previous run's output).

run.sh: minor tweaks: added a new 'prepare' option, which calls all the
options needed before running the tests (and 'all' now starts by calling
that instead of those options directly); updated & improved the 'help'
option; and added comments.


tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -266,51 +281,63 @@ def print_file_tail_and_errors(from_file, to_file, number_of_lines=50):
...
-                   "Attribute 'enabled' is not allowed to appear in element 'export'",
-                   'PartitionInfo specifies invalid parameter index for procedure',
-                   'unexpected token: EXEC', 'Failed to plan for statement',
-                   'Unexpected error in the SQL parser for statement', 'Could not parse reader',
-                   'No index found to support UPDATE and DELETE on some of the min() / max() columns in the materialized view',
-                   'WARN', 'ERROR']
+    # Various types of errors we're interested in finding (in the log or
+    # console output), divided into several groups, i.e., (Java) Exceptions
+    # vs. generic errors vs. warnings
+    error_types = [{'type':'Exception', 'title':"Java 'Exception'",
```
wolffcm: This would be a bit easier to read if there was just one subtype per line.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -266,51 +281,63 @@ def print_file_tail_and_errors(from_file, to_file, number_of_lines=50):
...
+                                "Attribute 'enabled' is not allowed to appear in element 'export'",
+                                'PartitionInfo specifies invalid parameter index for procedure',
+                                'Unexpected error in the SQL parser for statement']},
+                     {'type':'WARN', 'title':"'WARN'",
+                    'subtypes':['No index found to support UPDATE and DELETE on some of the min() / max() columns in the materialized view']}
+                   ]
+    # Avoid double-counting: ArrayIndexOutOfBoundsException is also counted as
+    # IndexOutOfBoundsException; 'unexpected token: EXEC' is also counted as
+    # 'Failed to plan for statement'
+    error_subtypes_contained_in_others = ['ArrayIndexOutOfBoundsException', 'unexpected token: EXEC']
```
wolffcm: Seems like you could avoid needing to have this by prepending `[^a-zA-z]` to the pattern when invoking `grep`.

tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -703,27 +734,21 @@ def print_sql_statement(sql, num_chars_in_sql_type=6):
...
-                    previous_output = None
-                    current_output  = None
-                    if sqlLen > 1:
-                        previous_sql    = last_n_sql_statements[sqlLen-2]['sql']
-                        previous_output = last_n_sql_statements[sqlLen-2]['output']
-                        current_output  = last_n_sql_statements[sqlLen-1]['output']
-                    if ('show' in previous_sql and 'classes' in previous_sql
-                            and 'Procedure Classes' in previous_output
-                            and 'Procedure Classes' in current_output
+                    # Check for a special case: avoid double-counting 'show classes'
```
wolffcm: Do you need to do this because we don't echo "show classes" to the output?

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: Done.

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: I think that would solve the first case (ArrayIndexOutOfBoundsException vs. IndexOutOfBoundsException), but not the second ('unexpected token: EXEC', which appears on the same line as a more generic 'Failed to plan for statement' message).

tests/sqlgrammar/sql_grammar_generator.py
johnTcrawford: Yes, the "special case" handling here is needed due to the combination of the fact that "show classes" (like other "show" commands) do not get echoed, plus, as the comment says, the output of "show classes" is more unpredictable than others, returning anywhere from 1 to 3 lists of different types of classes (i.e., 'Potential Procedure Classes', 'Active Procedure Classes', 'Non-Procedure Classes').

PR #5073, opened 2017-12-13
ENG-13390 fix catalog diff engine to evaluate the security user change with catalog update

zeemanhe:
HTTP sessions will be cleaned up only if security users are updated in catalog update.

### DIFF
--- a/src/catgen/in/javasrc/CatalogDiffEngine.java
+++ b/src/catgen/in/javasrc/CatalogDiffEngine.java
@@ -241,6 +241,11 @@ public class CatalogDiffEngine {
         return true;
     }
 
+    public boolean hasSecurityUserChanges() {
+        CatalogChangeGroup ccg = m_changes.get(DiffClass.USER);
+        return ccg.hasChanges();
+    }
+    
     public String[][] tablesThatMustBeEmpty() {
         ArrayList<String> tableSetNames = new ArrayList<>();
         ArrayList<String> errorMessages = new ArrayList<>();
@@ -834,6 +839,10 @@ public class CatalogDiffEngine {
                                             final CatalogType prevType,
                                             final String field)
     {
+        if (suspect instanceof User) {
+            return ("shadowPassword".equals(field) || "sha256ShadowPassword".equals(field));
+        }
+        
         return false;
     }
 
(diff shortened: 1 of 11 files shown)


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -241,6 +241,11 @@ public boolean requiresNewExportGeneration() {
         return true;
     }
 
+    public boolean hasSecurityUserChanges() {
+        CatalogChangeGroup ccg = m_changes.get(DiffClass.USER);
```
akhanzode: What about security enabled="true|false" boolean that should be checked too.

src/catgen/spec.txt
```diff
@@ -91,6 +91,7 @@ end
 
 begin User javaonly
   GroupRef* groups
+  string password "unencrypted password"
```
akhanzode: Why do you need to add cleartext password in catalog this is bad as customers who will send us catalog will end up sending their passwords too.

akhanzode:
Add tests in TestCatalogDiff see verifyDiff()

src/catgen/in/javasrc/CatalogDiffEngine.java
zeemanhe: a check has been added

src/catgen/spec.txt
zeemanhe: passwords are encoded with random seeds. added a encrypted  version with fixed salt for easy comparison.

src/catgen/spec.txt
```diff
@@ -91,6 +91,7 @@ end
 
 begin User javaonly
   GroupRef* groups
+  string password "encrypted version"
```
xinjiacs: what's the "encrypted version"?

“encrypted password version for catalog comparison usage"?

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2172,7 +2172,8 @@ private static void setUsersInfo(Catalog catalog, UsersType users) throws Runtim
                             saltGen);
             catUser.setShadowpassword(hashedPW);
             catUser.setSha256shadowpassword(hashedPW256);
-
+            //use fixed seed for comparison
+            catUser.setPassword( BCrypt.hashpw(sha1hex, "$2a$10$pWO/a/OQkFyQWQDpchZdEe"));
```
xinjiacs: Use sha256 password as we are going to remove sha1 soon.

PR #5080, opened 2017-12-19
Make function/statement dependency lists deterministically ordered in c…

wolffcm:
…atalog


yzhang1991:
You can use StringUtil.join

wolffcm:
I agree that removing the leading and trailing `,` would be nice, but if we change the catalog format, will it have effects on migrating between versions?

wolffcm:
@yzhang1991 Ning made a diagnostic commit that is probably too verbose for users, I will attempt to clean it up, so I'll be adding another commit to this branch.

PR #5086, opened 2017-12-22
ENG-12889 NT All-Partition API

jhugg:


src/frontend/org/voltdb/VoltTable.java
```diff
@@ -1744,6 +1744,37 @@ public int hashCode() {
...
+     */
+    VoltTable semiDeepCopy() {
+        assert(verifyTableInvariants());
+        // share the immutable metadata if it's present for tests
+        final VoltTable cloned = new VoltTable(m_extraMetadata);
+        cloned.m_colCount = m_colCount;
+        cloned.m_rowCount = m_rowCount;
+        cloned.m_rowStart = m_rowStart;
+
+        cloned.m_buffer = m_buffer.duplicate();
```
luyangco: So the underlying buffer is still shared by all copies, is this the reason that it is called "semi"?

tests/frontend/org/voltdb/TestNTProcs.java
```diff
@@ -618,6 +712,30 @@ public void testRunOnAllHostsSerialness() throws Exception {
         localServer.join();
     }
 
+    /*public void testRunOnAllHostsAPI() throws Exception {
```
luyangco: Can be removed if not needed anymore.

PR #5101, opened 2018-01-04
Make EE unit test deterministic

wolffcm:
The CTE EE unit test made use of a seq scan of a persistent table, which is not deterministic when the EE is compiled in memcheck mode.  These changes add an ORDER BY node to the JSON plan, so that the result is always deterministic.


wolffcm:
Yes, that's exactly what I did to produce the plan.

PR #5104, opened 2018-01-05
sql-grammar.txt: added a new section defining grammar rules for Common

johnTcrawford:
Table Expressions (CTEs) - for now, just Recursive CTEs; and added CTE
statements to the list of possible types of SQL statements; added a new
VARCHAR column (see below for details), and simplified future adding of
columns; added a small section defining JSON values (while I was messing
with VARCHAR columns; needs more work, though); and a couple of tweaks
of IPV4/6 values (VARCHAR and VARBINARY); improved some of the
definitions related to joins (since they were messing up CTEs that use
joins), particularly around using the same table, view, or alias name
more than once in the same query; added 'rarely-null' versions of each
data type (e.g. 'str-rarely-null-value'), for cases (like CTEs) where we
want NULL values to be tested occasionally, but not normally used, since
they are likely to produce invalid SQL statements; various other minor
changes / improvements; and updated / improved some comments.

DDL.sql, JSP*.java, GetOrInsertBase.java: just added one column to each
table (and view): the old VCHAR column (16 chars or 64 BYTES) is renamed
VCHAR_OUTLINE_MIN; the new column has the old name of VCHAR, but is of
unspecified size (i.e., the default size) - which is what PostgreSQL
requires for RECURSIVE CTE queries.

sql_grammar_generator.py: increased the time to wait for sqlcmd to
"hang", from 10 to 15 seconds, so that it exceeds the standard query
timeout of 10 seconds; added the ability to escape curly braces ('{' and
'}') and backslashes ('\'), by adding a backslash in front of it (this
is in addition to the existing escaping of brackets, '[' and ']'); and
moved the echo / grammar symbols code outside of the if sqlcmd, so that
it works even when SQL statements are being generated without being
passed to sqlcmd.


tests/sqlgrammar/sql-grammar.txt
```diff
@@ -195,15 +211,15 @@ table-alias-or-name     ::= {table-alias} 3| {table-or-view-name}
 # be SELECT statements; but only 1% will be "delete-statement", since we don't
 # want to delete all the rows in a table too often.
 #
-sql-statement           ::= {standard-sql-statement} 19| {special-sql-statement}
-
-standard-sql-statement  ::= {select-statement} 50| {insert-statement} 18| {upsert-statement} 18| \
-                            {update-statement}  8| {delete-statement}
+sql-statement           ::= {select-statement} 50| {cte-statement} 5| \
```
wolffcm: Is it possible to run grammar generator using a particular non-terminal as a starting point?  Like if I only cared about CTEs, just generate a whole bunch of statements from the rule for `cte-statement` so I wouldn't have any other kind of statement in there.

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -263,38 +293,40 @@ varb-ipv6-value-or-subq ::= {varbin-ipv6-value} 9| {varb-ipv6-scalar-sub-q}
...
+insert-values-all       ::= SERT INTO {table-name} [({every-column-name})]     VALUES ( {insert-values-all-val} )
+insert-values-backward  ::= SERT INTO {table-name} ({every-col-name-backward}) VALUES ( {insert-values-back-val} )
 
 insert-values-mult-val  ::= {insert-values-mult-val1} 4| {insert-values-mult-val2}
 insert-values-mult-val1 ::= {integer-non-null-value}, {numeric-value}, {string-value}, {timestamp-value}, \
                             {varbinary-value}, {point-value}, {polygon-value}
 insert-values-mult-val2 ::= {integer-non-null-value}, {numeric-value-or-subq}, {string-value-or-subq}, {timestamp-value-or-subq}, \
                             {varbinary-value-or-subq}, {point-value-or-subq}, {polygon-value-or-subq}
 
+# Note: if you add a new column to the DDL file, add it here
```
wolffcm: One could imagine creating a grammar-generator-generator that would generate both the DDL and the grammar rules based on a spec that contains the set of column types and other stuff.  That way you would only need to update one place when you want to add a column to your tables.  You could also use such a technique to randomly generate DDL?

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -448,6 +493,230 @@ basic-delete-statement  ::= DELETE FROM {table-name} [{where-clause}] [{sort-cla
 
 truncate-statement      ::= TRUNCATE TABLE {table-name}
 
+################################################################################
+# Grammar rules for a Common Table Expression (CTE) statement, i.e. using a
+# WITH clause, with or without RECURSIVE:
+################################################################################
+#
+# TODO: switch back to something reasonable (after non-recursive CTE's added):
```
wolffcm: Non-recursive CTEs should be working now in the ENG-13398-CTEPlanner branch.

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -448,6 +493,230 @@ basic-delete-statement  ::= DELETE FROM {table-name} [{where-clause}] [{sort-cla
...
+
+########################################
+#
+recursive-cte-statement ::= WITH RECURSIVE {recursive-cte-query}
+recursive-cte-query     ::= {rec-cte-int} 4| {rec-cte-num} 2| {rec-cte-str} 4| \
+                            {rec-cte-time} | {rec-cte-varbin} | {rec-cte-point} | {rec-cte-polygon} | \
+                            {rec-cte-num-str} | {rec-cte-str-num} | rec-cte-3-int | rec-cte-3-str
+
+rec-cte-int             ::= rcte{optional-1-rec-cte-col}  AS (\
+                            {basic-query-int} \
```
wolffcm: You could add testing for non-recursive CTEs by just using `basic-query-*` rules that you've already defined here (but you could also get more sophisticated too).

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -448,6 +493,230 @@ basic-delete-statement  ::= DELETE FROM {table-name} [{where-clause}] [{sort-cla
...
+                            FROM {table-references} [{where-clause}]  49| \
+                            VALUES ({str-rarely-null-value}, {str-rarely-null-value}, {str-rarely-null-value})
+
+rcte-string-expression  ::= {simple-rcte-string-expr} 49| {string-scalar-sub-q}
+simple-rcte-string-expr ::= {string-value} | [[[[{table-alias-or-name}.]]]]{rcte-string-column-name} 8| \
+                            {string-function-expr}
+rcte-string-column-name ::= VCHAR 9| {any-string-column-name}
+
+# TODO: lots more options here (for the WHERE clause)
+recursive-query-int     ::= SELECT {recursive-clause-int1} FROM rcte WHERE {rec-cte-col1:rctec1} {comparison-operator} {int-rarely-null-value}
```
wolffcm: I think you can have GROUP BY on the base and/or recursive queries, which would be nice to test.

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: Yes. If you use the relatively new <voltdb>/tests/sqlgrammar/run.sh script, the default option is 'help', which describes all of the script's options, one of which is 'tests-help', which describes all of the test arguments, one of which is '--type=TYPE', which defaults to '--type=sql-statement'; but you can instead specify '--type=cte-statement', which will do what you want. (This argument has existed much longer than the run.sh script.)

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: Yeah, sorry, that's an old comment - should now say once they're implemented in the SQL grammar.

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: Thanks, I'll try that.

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: I'll try that.

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: Getting it down to 4 clearly identified places seemed like great progress; but yes, more could be done, though those 4 spots are all slightly different, one requiring values or sub-queries of each type; one columns or sub-queries, etc. As for generating DDL (beyond the existing CREATE and DROP PROCEDURE statements): if we're talking about DDL only, it's not too hard (I think); if we're talking about generating both DDL and (non-DDL) SQL statements, intermingled, then it will take some careful work to make sure that a reasonable percentage of the (non-DDL) SQL statements remain valid.

prosegay: I suggested this concept quite a while ago, it was discussed at QA meetings.

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -591,93 +612,255 @@ rec-cte-col-alias7      ::= CA7
...
+                            FROM {table-references} [{where-clause}] 150| \
+                            SELECT {rcte-string-expression} {rec-cte-col1-alias}, {rcte-string-expression} {rec-cte-col2-alias}, {rcte-string-expression} {rec-cte-col3-alias}, \
+                                   {rcte-string-expression} {rec-cte-col4-alias}, {rcte-string-expression} {rec-cte-col5-alias}, {rcte-string-expression} {rec-cte-col6-alias}, \
+                                   {rcte-string-expression} {rec-cte-col7-alias}, {rcte-string-expression} {rec-cte-col8-alias}, {rcte-string-expression} {rec-cte-col9-alias}  \
+                            FROM {table-references} [{where-clause}]  49| \
+                            VALUES ({str-rarely-null-value}, {str-rarely-null-value}, {str-rarely-null-value}, \
+                                    {str-rarely-null-value}, {str-rarely-null-value}, {str-rarely-null-value}, \
+                                    {str-rarely-null-value}, {str-rarely-null-value}, {str-rarely-null-value})
+basic-query-9-mixed     ::= SELECT {int-expression}, {rcte-string-expression}, {numeric-expression}, \
```
wolffcm: So you are not mixing types here between the base and recursive query.  You could mismatch the types of corresponding columns, say, have an integer as the first column of the base query and a string as the first column of the recursive query.  It may not be a sensible thing to do, but it would be a good test.

johnTcrawford:
Closing because I've merged this code to ENG-13398-CTEPlanner, rather than to master.

PR #5112, opened 2018-01-09
ENG-13535: Fix broken upgrade to V7.9

wolffcm:


src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -71,6 +71,7 @@
     public File getDROverflow();
 
     @Key(LARGE_QUERY_SWAP_PATH_KEY)
+    @DefaultValue("large_query_swap") // must match value in voltdb/compiler/DeploymentFileSchema.xsd
```
luyangco: I think this change would fix the upgrade issue, but I have some concerns about the default value. After upgrade from a lower version, the default value LARGE_QUERY_SWAP_PATH would prevail even though deployment file sets to a different value.

wolffcm: I don't think that will happen.  What happens now is we take the values from config/path.properties and use that to populate the node settings.  So whatever value is in there (from the older version of VoltDB) will be used.  I'll verify this manually.

wolffcm: I have verified that if you change the large query swap path to something else, and then do `voltadmin shutdown --save` and restart the database, the custom path will be retained as a node setting.  So the default value will only be used if there is nothing in path.properties (which is the case for older versions of VoltDB).

luyangco: I mean if we upgrade from 7.8 to 7.9 using terminal snapshot but not re-initializing the voltdbroot, will path.properties pick up the LARGE_QUERY_SWAP_PATH specified in deployment file or use default value?

wolffcm: Hmm, I'll check.

PR #5113, opened 2018-01-09
ENG-12441: remove python cli voltdb old cli API

xinjiacs:


lib/python/voltcli/voltdb.d/create.py
```diff
@@ -34,4 +34,4 @@
     hideverb=True
 )
 def create(runner):
-    runner.go()
+    runner.abort('voltdb create is no longer supported, please use init to initialize and start to start the database.')
```
luyangco: Missing the single quote for the words `init` and `start`.

PR #5115, opened 2018-01-09
Eng 13524 Stop CTE Leaks

BillWhite:
This fixes ENG-13524, which complains about CTE tables leaking out of their queries.

We also fix a bug in which we did not complain when there are too many or too few column names in a CTE query.


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -715,6 +715,16 @@ private void XreadWithList(WithList withList) {
...
             Type[] colTypes = baseQueryExpression.getColumnTypes();
+            String cmp = null;
+            if (colNames.length > colTypes.length) {
+                cmp = "many";
+            }
+            else if (colNames.length < colTypes.length) {
+                cmp = "few";
+            }
+            if (cmp != null) {
+                throw Error.error("Too " + cmp + " many column names for schema in query " + queryName.name);
```
wolffcm: You'll want to remove the "many" from this so the message reads right, otherwise you'll get " too many many column names ".

wolffcm: Also don't forget to get the message reviewed by Andrew.

BillWhite: Done to both.

PR #5116, opened 2018-01-09
ENG-13132-explain-CTE

yzhang1991:
Basically, this pull request added explain strings to the CTE plan node and the sequential scan plan node. It also linked the base and the recurse query in a CTE to the main query. This is done in two places:
* **When we first plan a CTE query**: The explain strings for stored procedures are generated during this time.
* **When a plan node tree is deserialized from the JSON plan string**: AdHoc explain strings are built from the reconstructed plan node tree.

The original work had some methods that were depending on the SmtCommonTableScan, but his object is lost when the plan was deserialized from the JSON plan string. I corrected those APIs.

I remembered that Chris said some CTE plan elements are not visible from `explainjson` system stored procedure. I think this is also because the plan trees were disconnected. I would expect this pull request can also fix that issue.


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -1029,17 +1030,31 @@ private AbstractPlanNode connectChildrenBestPlans(AbstractPlanNode parentPlan) {
...
+                assert(parentPlan instanceof SeqScanPlanNode);
+                SeqScanPlanNode scanPlanNode = (SeqScanPlanNode)parentPlan;
+                StmtCommonTableScan cteScan = (StmtCommonTableScan)tableScan;
+                CompiledPlan bestCostBasePlan = cteScan.getBestCostBasePlan();
+                CompiledPlan bestCostRecursePlan = cteScan.getBestCostRecursePlan();
+                assert(bestCostBasePlan != null);
+                AbstractPlanNode basePlanRoot = bestCostBasePlan.rootPlanGraph;
+                scanPlanNode.setCTEBaseNode(basePlanRoot);
+                if (bestCostRecursePlan != null) {
+                    // Either the CTE is not recursive, or this is a recursive CTE but we
```
BillWhite: I'm not sure this comment is right.  When we get here all the best case plans have been calculated.  So if the recursive plan is null then it is not recursive.  I could be wrong, though.

src/frontend/org/voltdb/planner/parseinfo/StmtCommonTableScan.java
```diff
@@ -99,15 +99,15 @@ public final void setBestCostBasePlan(CompiledPlan plan, int stmtId) {
...
+        m_sharedScan.setBestCostRecursePlan(plan, stmtId);
     }
 
     public final CompiledPlan getBestCostBasePlan() {
         return m_sharedScan.getBestCostBasePlan();
     }
 
-    public final CompiledPlan getBestCostRecursivePlan() {
-        return m_sharedScan.getBestCostRecursivePlan();
+    public final CompiledPlan getBestCostRecursePlan() {
```
BillWhite: Why did you change these names?  It seems like the bulk of your changes are these name changes.  I don't think this is idiomatic.  Recursive is an adjective, plausibly modifying the noun plan.  Recurse is a verb, and this case seems like a command to recurse the plan in the best cost way.  I would change it back.

src/frontend/org/voltdb/planner/parseinfo/StmtCommonTableScan.java
```diff
@@ -159,7 +159,7 @@ public Integer getBaseStmtId() {
     }
 
     public Integer getRecursiveStmtId() {
-        return m_sharedScan.getBestCostRecursiveStmtId();
+        return m_sharedScan.getBestCostRecurseStmtId();
```
BillWhite: Again, I would change Recurse to Recursive.

src/frontend/org/voltdb/plannodes/CommonTablePlanNode.java
```diff
@@ -26,7 +26,12 @@
 public class CommonTablePlanNode extends AbstractPlanNode {
     private String m_commonTableName;
     private StmtCommonTableScan m_tableScan;
-    private Integer m_recursiveStatementId;
+    private Integer m_recurseNodeId;
```
BillWhite: Again, these should be renamed from m_recurseNodeId to m_recursiveNodeId.  But I think these are statement ids and not node ids.  Calling them node ids will be confusing.  The node ids are ids of plan nodes.  The statement ids are ids of executor sequences.

src/frontend/org/voltdb/plannodes/PlanNodeTree.java
```diff
@@ -150,15 +150,37 @@ public void loadFromJSONPlan(JSONObject jobj, Database db)  throws JSONException
...
         }
     }
 
+    /**
+     * Connect the base and the recursive (if there is one) query plan nodes
+     * to the scan plan node.
+     * @param node The common table scan plan node.
+     */
+    private void connectCTENodesIfAny(SeqScanPlanNode node) {
+        if (! node.isCommonTableScan()) {
```
BillWhite: What does this change have to do with the explain plans?

src/frontend/org/voltdb/plannodes/SeqScanPlanNode.java
```diff
@@ -36,8 +36,8 @@
 import org.voltdb.types.SortDirectionType;
 
 public class SeqScanPlanNode extends AbstractScanPlanNode implements ScanPlanNodeWhichCanHaveInlineInsert {
-    int m_CTEBaseStmtId;
-    boolean m_isCTEScan;
+    private Integer m_CTEBaseNodeId;
```
BillWhite: Again, a statement id is not a node id.  It seems confusing the confuse them.

src/frontend/org/voltdb/planner/PlanAssembler.java
yzhang1991: No, we can still get here when planning the recursive query itself.

src/frontend/org/voltdb/plannodes/PlanNodeTree.java
yzhang1991: We just need to have the plan nodes of the base query and recursive query linked to the main query. So that the explain method in the main query can have access to those separate plan node trees to generate proper explaination text.

PR #5123, opened 2018-01-10
Fix HSQL so that it parses subqueries for the base case.

BillWhite:
Add a minimal test for this as well.

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestCommonTableExpressionsSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestCommonTableExpressionsSuite.java
@@ -190,7 +190,8 @@ public class TestCommonTableExpressionsSuite extends RegressionSuite {
                 + "    select cte_data.* from cte_data join rt on cte_data.id = rt.l "
                 + ") "
                 + "select * from rt order by rt.id";
-        ClientResponse cr = client.callProcedure("@AdHoc", SQL);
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/regressionsuites/TestCommonTableExpressionsSuite.java
```diff
@@ -412,6 +428,11 @@ public void testEng13500Crash() throws Exception {
                 vt);
     }
 
+    public void testSubqueries() throws Exception {
+        Client client = getClient();
+        // %%%
```
wolffcm: Whoa, looks like you meant to put something here.  I've merged my "more testing" ticket to the topic branch, so you could just put it there, using a subquery with an ORDER BY in non-recursive and the base query of recursive CTEs.  A natural place would be the `testGroupByAndOrderBy` method that you reviewed.

wolffcm:
Other than the test issue, the fix looks good!

tests/frontend/org/voltdb/regressionsuites/TestCommonTableExpressionsSuite.java
BillWhite: No.  I forgot to delete TestCommonTableExpressionSuite.testSubqueries.  The test is the second test in TestCommonTableExpressionSuite.testCTE.

PR #5131, opened 2018-01-17
ENG-13560 immediate shutdown of Kafka importer to avoid blocking other processes

zeemanhe:
When cluster state changes, such as @pause, importers will  grab a lock on ImporterManager and will be shutdown. In the meantime, catalog update needs the same lock to proceed. Shutting down importers immediately to avoid blocking catalog update.

### DIFF
--- a/src/frontend/org/voltdb/importclient/kafka10/KafkaStreamImporter.java
+++ b/src/frontend/org/voltdb/importclient/kafka10/KafkaStreamImporter.java
@@ -143,8 +143,10 @@ public class KafkaStreamImporter extends AbstractImporter {
 
     @Override
     public void stop() {
-        for (KafkaInternalConsumerRunner consumer : m_consumers) {
-            consumer.shutdown();
+        if (m_consumers != null) {
+            for (KafkaInternalConsumerRunner consumer : m_consumers) {
+                consumer.shutdown();
+            }
         }
         if (m_executorService != null) {
             try {
--- a/src/frontend/org/voltdb/importclient/kafka10/KafkaStreamImporter.java
+++ b/src/frontend/org/voltdb/importclient/kafka10/KafkaStreamImporter.java
@@ -153,7 +153,7 @@ public class KafkaStreamImporter extends AbstractImporter {
 
         if (m_executorService != null) {
             try {
-                //shutdown immediately. Consumers will pick up where they left
+                //ENG-13560 shutdown immediately. Consumers will pick up where they left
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/importer/ImportManager.java
```diff
@@ -158,7 +158,7 @@ private synchronized void restartImporters(Map<String, ImportConfiguration> newP
      * @param catalogContext new catalog context
      * @return new importer configuration
      */
-    private synchronized Map<String, ImportConfiguration> loadNewConfigAndBundles(CatalogContext catalogContext) {
+    private Map<String, ImportConfiguration> loadNewConfigAndBundles(CatalogContext catalogContext) {
```
xinjiacs: @zeemanhe I think we need to remove all the synchronized under updateCatalog() method. As the caller already guarantee that there is only one caller.

xinjiacs:
@zeemanhe Does this race deadlock issue start to happen when we upgrade the importer with kafka10?

zeemanhe:
I have not traced down. It is possible since Kafka10 has different shutdown mechanim.

src/frontend/org/voltdb/importclient/kafka10/KafkaStreamImporter.java
```diff
@@ -153,8 +153,9 @@ public void stop() {
 
         if (m_executorService != null) {
             try {
+                //shutdown immediately. Consumers will pick up where they left
```
xinjiacs: @zeemanhe Can you add the ENG-13560 in this comment as well?

src/frontend/org/voltdb/importer/ImporterLifeCycleManager.java
```diff
@@ -268,12 +268,12 @@ public final void stop()
         }
 
         if (m_executorService != null) {
-            m_executorService.shutdown();
+            //shutdown right away to avoid delay. Importers should pick up where they left.
```
xinjiacs: @zeemanhe Can you add the ENG-13560 in this comment as well?

zeemanhe: sure

xinjiacs:
@zeemanhe I verified that this issue starts from kafka10 upgrade. 
The ImporterLifeCycleManager awaitTermination was there since 2015. It's only the kafka10 client not getting shutting down immediately that blocks the process.

PR #5132, opened 2018-01-17
Eng 13591 explain json fix

BillWhite:
Print all the plans in explainjson, including subqueries and CTE plans.


src/frontend/org/voltdb/plannodes/CommonTablePlanNode.java
```diff
@@ -123,7 +123,9 @@ protected String explainPlanForNode(String indent) {
     public void toJSONString(JSONStringer stringer) throws JSONException {
         super.toJSONString(stringer);
         stringer.key(Members.COMMON_TABLE_NAME.name()).value(m_commonTableName);
-        m_recursiveStatementId = m_tableScan.getRecursiveStmtId();
+        if (m_recursiveStatementId == null && m_tableScan != null) {
+            m_recursiveStatementId = m_tableScan.getRecursiveStmtId();
```
BillWhite: m_tableScan can be null if we have reconstructed this from JSON.

src/frontend/org/voltdb/plannodes/PlanNodeList.java
```diff
@@ -43,8 +43,8 @@ public PlanNodeList() {
         super();
     }
 
-    public PlanNodeList(AbstractPlanNode root_node, boolean isLargeQuery) {
-        m_tree = new PlanNodeTree(root_node);
+    public PlanNodeList(PlanNodeTree tree, boolean isLargeQuery) {
```
wolffcm: What is the difference if you create a PlanNodeList from a root node vs. a PlanNodeTree?

wolffcm: This seems like a simple change, I would just like to understand why it was broken and how you fixed it.

BillWhite: The PlanNodeTree knows how to dismember a plan into statements, which are PlanNodes.  The statements are stored in subquery expression nodes and in CTE sequential scan nodes.  The PlanNodeList creates a PlanNodeTree, then creates execute lists from the statements, taking into account the parent/child relations.  I don't know why we can't just do a depth first search of the plan node trees, but apparently we can't.  The result is a list of lists of plan node ids, where in each inner list the children are listed before their parents.

The problem was that we took a two fragment plan, reconnected it to be a single plan tree, then created a PlanNodeTree from it. This did the dismembering part but not the topological sort part to create the execute lists.  We then took the first plan in the PlanNodeTree, did the topological sort of the nodes and printed the execute list.  But there are other plans in the PlanNodeTree which we did not get to.  I changed this to let one create a PlanNodeList from a PlanNodeTree directly, bypassing the dismembering part.  I could have taken the single plan tree and created a PlanNodeList from it directly.  But that would have duplicated the effort we already made in creating the PlanNodeTree.

You are right that this is confusing, and could be done more clearly.

PR #5133, opened 2018-01-17
Eng 13536 union sort order

BillWhite:
Fix union sort order.

We process the order by list when we want to process the display list.


tests/frontend/org/voltdb/planner/TestUnion.java
```diff
@@ -622,6 +624,36 @@ public void testUnionOrderByLimitParams() {
...
+                                                        PlanNodeType.LIMIT),
+                                      // This is an order by node.  We know this
+                                      // already from the test above.  This is an
+                                      // example of using a lambda to test a node.
+                                      // One could add any computation here.  Of
+                                      // course, a tastier way to do this would be
+                                      // to have the lambda be statically defined
+                                      // in PlannerTestCase.  But this works better
+                                      // as an example.
+                                    (node) -> {
```
wolffcm: Seems like this test could be a bit more strict.  You could verify the number of order by columns, their expressions as well as their number.  Also you could verify the top-level order by as well.

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -736,72 +741,232 @@ public String match(AbstractPlanNode node) {
...
+            for (PlanMatcher pm : m_nodeSpecs) {
+                if (node == null) {
+                    return "Expected more nodes in plan.";
+                }
+                String err = pm.match(node);
+                if (err != null) {
+                    return err;
+                }
+                if (node.getChildCount() > 0) {
+                    node = node.getChild(0);
```
wolffcm: You've had to change a bunch of existing tests to work with this way of testing plans.  I wonder if it makes sense to take it all the way and start to handle both left and right sides of plan nodes?

Overall I like the schema you've come up with here.  Does this duplicate functionality you can get from Hamcrest though?

BillWhite: Probably.  But I have never understood how to create Hamcrest tests.  It's probably easy, but I find it very confusing, and I evolved this myself by following my nose.  Convergent evolution is, I think what they call it.

Maybe we should revisit this.

I do think we need something to traverse the left and right sides of a plan node.  I have an idea but I didn't need it for this ticket.

tests/frontend/org/voltdb/planner/TestUnion.java
BillWhite: True.  Almost anything is possible.  This is more an example of a lambda to match nodes than anything else.

PR #5139, opened 2018-01-23
ENG-13439: Sort large temp tables

wolffcm:
These two commits add support for sorting large temp tables.
The first commit is just sorting-related with no LIMIT or OFFSET
- I changed the LTT block cache to evict MRU (not LRU) blocks.  It turns out this is the best policy for FIFO workloads.
- I added a LargeTempTableBlock::iterator class that behaves like a std random access iterator.  I had hoped to use std::sort to sort blocks, but this proved difficult since tuple size is not known at compile time.

The second commit adds LIMIT/OFFSET support.  You might find it easier to review the commits one at a time.


yzhang1991:
Reviewed 18 of 20 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/ee/common/LargeTempTableBlockCache.cpp, line 50 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4DSrlKdjgjHvt5NE43:-L4DSrlKdjgjHvt5NE44:b-ro9g98) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/common/LargeTempTableBlockCache.cpp#L50)):*
> ```cpp
>     int64_t id = getNextId();
> 
>     m_blockList.emplace_back(new LargeTempTableBlock(id, schema));
> ```

I really admire your changes here to adopt the MRU method.

---

*[src/ee/common/LargeTempTableBlockCache.h, line 80 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4DZMusf9jZTWY9_GY1:-L4DZMusf9jZTWY9_GY2:b-bceajp) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/common/LargeTempTableBlockCache.h#L80)):*
> ```objc
> 
>     /** The large temp table for this block is being destroyed, so
>         release all resources associated with this block. */
> ```

The releaseBlock method is not only used when a LTT is being destroyed. For example, you also use that in sorting?

---

*[src/ee/common/ValueFactory.hpp, line 69 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4C2PULmfzDZIK7ICPK:-L4C2PULmfzDZIK7ICPL:b-lawmc) ([raw file](https://github.com/voltdb/voltdb/blob/e0f251e7dbc77cb751eddb7883c6aa6a6c071bec/src/ee/common/ValueFactory.hpp#L69)):*
> ```cpp
>     /// that will require an explicit NValue::free.
>     static NValue getStringValue(const std::string& value, Pool* pool = NULL) {
>         return NValue::getAllocatedValue(VALUE_TYPE_VARCHAR, value.c_str(), value.length(), pool);
> ```

Oh. good catch. This could have resulted in some bugs?

---

*[src/ee/storage/LargeTempTable.cpp, line 256 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4C6hMhqurRbMgQUJfZ:-L4C6hMhqurRbMgQUJf_:b-rwfvsp) ([raw file](https://github.com/voltdb/voltdb/blob/e0f251e7dbc77cb751eddb7883c6aa6a6c071bec/src/ee/storage/LargeTempTable.cpp#L256)):*
> ```cpp
> 
>             // Copy all the non-inlined data at once
>             outputBlock->copyNonInlinedData(*block);
> ```

Being able to get here means there is no non-inlined data in this block, so why still copy?

---

*[src/ee/storage/LargeTempTable.cpp, line 273 at r1](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4C7nuTPjkjfu-PaKab:-L4C7nuTPjkjfu-PaKac:b-htnzd6) ([raw file](https://github.com/voltdb/voltdb/blob/e0f251e7dbc77cb751eddb7883c6aa6a6c071bec/src/ee/storage/LargeTempTable.cpp#L273)):*
> ```cpp
> 
>             // Swap the blocks so that the caller sees the input block as sorted.
>             block->swap(outputBlock);
> ```

Wait, you just swapped it? So the rest of the tuples (after limit+offset position) will disappear. Sounds OK if it is and will still be impossible to scan the temp table more than once. But if that's the case, all large temp table iterators should be delete-as-go iterators?

---

*[src/ee/storage/LargeTempTable.cpp, line 138 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4DQ1zQ1Yp6ZjN1_ndc:-L4DQ1zQ1Yp6ZjN1_ndd:b1ze9tm) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/storage/LargeTempTable.cpp#L138)):*
> ```cpp
>     m_blockIds.swap(otherLargeTable->m_blockIds);
>     std::swap(m_tupleCount, otherLargeTable->m_tupleCount);
>     if (m_blockForWriting || otherLargeTable->m_blockForWriting) {
> ```

Should this check happen first? Throwing an exception after the actual swapping work is done will put the two tables in a wrong state, I think.

---

*[src/ee/storage/LargeTempTable.cpp, line 309 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4Df0v_4umqpIYh20hh:-L4Df0v_4umqpIYh20hi:b-kpi9yt) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/storage/LargeTempTable.cpp#L309)):*
> ```cpp
>             pivot = endIt - 1;
> 
>             difference_type i = -1; // index of last less-than-pivot element
> ```

Do you think you can use an iterator for the last less than pivot element? It just made me think about this seeing you use `beginIt[i]` repeatedly in this code.

---

*[src/ee/storage/LargeTempTable.cpp, line 312 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4DbvLMjNZdI0OqhXrG:-L4DbvLNloXaSQZoFBxI:bscj8fv) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/storage/LargeTempTable.cpp#L312)):*
> ```cpp
>             for (difference_type j = 0; j < numElems - 1; ++j) {
>                 iterator it = beginIt + j;
>                 if (m_compare(it->toTableTuple(m_schema), pivot->toTableTuple(m_schema))) {
> ```

Kind of confused here. If the comparer returns true, the current element is smaller than the pivot, right? Can you add a comment?

---

*[src/ee/storage/LargeTempTable.cpp, line 408 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4ZTBkyvRfm9pYCkUdH:-L4ZTBkyvRfm9pYCkUdI:bexe6bm) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/storage/LargeTempTable.cpp#L408)):*
> ```cpp
>     ~SortRun() {
>         if (m_table) {
>             m_iterator = m_table->iteratorDeletingAsWeGo(); // unpins block if scan in progress
> ```

I am not sure what this means

---

*[src/ee/storage/LargeTempTable.cpp, line 562 at r2](https://reviewable.io:443/reviews/voltdb/voltdb/5139#-L4ZbYLC_sD9K8F0z0t_:-L4ZbYLC_sD9K8F0z0ta:bz2d3wt) ([raw file](https://github.com/voltdb/voltdb/blob/ec50ad839cb6659b8d18bfcd5d68d9f42e4fdeb2/src/ee/storage/LargeTempTable.cpp#L562)):*
> ```cpp
>                 // into output sort run.
>                 if (run->advance()) {
>                     mergeHeap.push(run);
> ```

Ah... I see... You are completely leveraging priority queue to merge..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/voltdb/voltdb/5139)*

PR #5140, opened 2018-01-24
ENG-13582 delay catalog read on join/rejoin

zeemanhe:
When a node is rejoining, the node will read the  catalog from ZooKeeper immediately and creates all the sites with the catalog.  Then the rejoining node proceeds to acquire a rejoining blocker on ZooKeeper.   There is a window between the site's creation and acquiring the rejoining blocker. If the catalog is updated within the window, the loaded catalog on the rejoining node is not updated.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1143,6 +1141,9 @@ public void initialize(Configuration config) {
                 }
             }
 
+            //delay catalog read after join/rejoin lock is acquired.
+            m_durable = readDeploymentAndCreateStarterCatalogContext(config);
```
xinjiacs: "readDeploymentAndCreateStarterCatalogContext" creates a m_catalogContext. Is there any other places that used this instance before? If so, we need to make sure it picked a valid instance value. @zeemanhe

zeemanhe: moved the places where m_catalogContext is used after readDeploymentAndCreateStarterCatalogContext.

PR #5145, opened 2018-01-26
ENG-13627: Don't release the compositebytebuf if it was already done.…

manjujames:
… Thus avoid unnecessary error.


src/frontend/org/voltcore/network/TLSEncryptionAdapter.java
```diff
@@ -317,7 +317,7 @@ void shutdown() {
             }
             m_partialMessages.clear();
 
-            m_encryptedMessages.release();
+            if (m_encryptedMessages.refCnt() > 0) m_encryptedMessages.release();
```
JadenW: Could the refCnt more than 1. Do we have to release all of them?

PR #5150, opened 2018-01-29
sql_grammar_generator.py: when sqlcmd output includes 'Server is paused

johnTcrawford:
and is available in read-only mode', end execution immediately with a
fatal error; moved (in the summary file and to STDOUT) the detailed
"hanging SQL" messages to before the last-n and summary messages, with
only a single line about "hanging SQL" after them (when such messages
appear); added 'Resource limit exceeded' to the list of ERROR messages
to be searched for, in the log and in the console output; in
get_last_n_sql_statements method, added optional times (that SQL was
sent and output received) to be printed out; and added code to save
these times, when we have "hanging SQL"; and therefore moved the
formatted_time method earlier, to be used in print_sql_statement.

### DIFF
--- a/tests/sqlgrammar/sql_grammar_generator.py
+++ b/tests/sqlgrammar/sql_grammar_generator.py
@@ -800,12 +800,13 @@ def print_sql_statement(sql, num_chars_in_sql_type=6):
                     sqlcmd_proc.communicate('exit')
                     exit(99)
 
-                # Check if sqlcmd command not found, or if sqlcmd cannot, or
-                # can no longer, reach the VoltDB server
+                # Check if the VoltDB server has been put into paused / read-only
+                # mode, usually due using too much memory (RSS), typically caused
+                # by a Recursive CTE causing an infinite loop
                 elif ('Server is paused and is available in read-only mode' in output):
                     error_message = '\n\n\nFATAL ERROR: sqlcmd responded:\n    "' + output + \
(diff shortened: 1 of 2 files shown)


tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -781,6 +800,16 @@ def print_sql_statement(sql, num_chars_in_sql_type=6):
...
 
+                # Check if sqlcmd command not found, or if sqlcmd cannot, or
+                # can no longer, reach the VoltDB server
+                elif ('Server is paused and is available in read-only mode' in output):
+                    error_message = '\n\n\nFATAL ERROR: sqlcmd responded:\n    "' + output + \
+                                    '"\npossibly due to a Recursive CTE (WITH) statement that caused ' + \
+                                    'an infinite loop, consuming too much memory:\n    "' + sql + '"\n'
+                    print_summary(error_message)
+                    sqlcmd_proc.communicate('exit')
+                    exit(98)
```
wolffcm: Is there any significance to these numbers?  Do you check the return code in the invoking script?

johnTcrawford: There is no significance, except that they are different from each other, so if there is such a failure, I know where to look. (Also, I avoid values commonly used by other issues, such as 1, 2, -1, 255.)

PR #5157, opened 2018-02-01
Fix CPP DRBinaryLog test by tracking and switch EngineLocals for two …

JadenW:
…MockVoltDBEngine


JadenW:
already merged into integration branch

PR #5158, opened 2018-02-01
Eng 13558:  Fix DRBinaryLog apply truncate table calling wrong function signature

JadenW:


src/ee/common/SynchronizedThreadLock.h
```diff
@@ -158,9 +160,9 @@ class SynchronizedThreadLock {
     static pthread_cond_t s_wakeLowestEngineCondition;
     static int32_t s_globalTxnStartCountdownLatch;
     static int32_t s_SITES_PER_HOST;
-    static EngineLocals s_mpEngine;
 
 public:
+    static EngineLocals s_mpEngine;
```
wweiss-voltdb: Don't make this public just because of a test.

src/ee/common/SynchronizedThreadLock.cpp
```diff
@@ -386,6 +386,16 @@ long int SynchronizedThreadLock::getThreadId() {
 #endif
 }
 
+void SynchronizedThreadLock::resetEngineLocalsForTest() {
+    s_mpEngine = new EngineLocals(true);
+    s_enginesByPartitionId = SharedEngineLocalsType();
```
wweiss-voltdb: clear() would be the more common way to do this.

PR #5159, opened 2018-02-01
Eng 13224 Large Temp Table Block Ids

BillWhite:
We change the block ids for Large Temp Tables from single int16_t values to structured values.

Objects of this class still have by-value semantics.  But they should be globally unique across
the entire cluster.

### DIFF
--- a/src/ee/common/LargeTempTableBlockCache.cpp
+++ b/src/ee/common/LargeTempTableBlockCache.cpp
@@ -31,7 +31,7 @@ LargeTempTableBlockCache::LargeTempTableBlockCache(Topend *topend, int64_t maxCa
     , m_maxCacheSizeInBytes(maxCacheSizeInBytes)
     , m_blockList()
     , m_idToBlockMap()
-    , m_nextId(0)
+    , m_nextId(int32_t(ExecutorContext::getExecutorContext()->m_siteId), 0)
     , m_totalAllocatedBytes(0)
 {
 }
--- a/src/ee/common/LargeTempTableBlockCache.cpp
+++ b/src/ee/common/LargeTempTableBlockCache.cpp
@@ -39,7 +39,7 @@ LargeTempTableBlockCache::LargeTempTableBlockCache(Topend *topend, int64_t maxCa
     , m_totalAllocatedBytes(0)
 {
     ExecutorContext *ec = ExecutorContext::getExecutorContext();
-    int32_t siteId = ec ? ec->m_siteId : 0;
+    int64_t siteId = ec ? ec->m_siteId : 0;
     m_nextId = LargeTempTableBlockId(siteId, 0);
 }
 
--- a/src/ee/common/LargeTempTableBlockCache.cpp
+++ b/src/ee/common/LargeTempTableBlockCache.cpp
@@ -31,9 +31,16 @@ LargeTempTableBlockCache::LargeTempTableBlockCache(Topend *topend, int64_t maxCa
     , m_maxCacheSizeInBytes(maxCacheSizeInBytes)
     , m_blockList()
     , m_idToBlockMap()
-    , m_nextId(int32_t(ExecutorContext::getExecutorContext()->m_siteId), 0)
+      // Sometimes for tests we don't have a site id.
+    , m_nextId(ExecutorContext::getExecutorContext()
+                 ? (ExecutorContext::getExecutorContext()->m_siteId)
+                 : 0,
+               0)
     , m_totalAllocatedBytes(0)
 {
+    ExecutorContext *ec = ExecutorContext::getExecutorContext();
+    int32_t siteId = ec ? ec->m_siteId : 0;
+    m_nextId = LargeTempTableBlockId(siteId, 0);
 }
 
 LargeTempTableBlockCache::~LargeTempTableBlockCache() {
--- a/src/ee/common/LargeTempTableBlockCache.cpp
+++ b/src/ee/common/LargeTempTableBlockCache.cpp
@@ -43,7 +43,7 @@ LargeTempTableBlockCache::~LargeTempTableBlockCache() {
 LargeTempTableBlock* LargeTempTableBlockCache::getEmptyBlock(TupleSchema* schema) {
     ensureSpaceForNewBlock();
 
-    int64_t id = getNextId();
+    LargeTempTableBlockId id = getNextId();
 
     m_blockList.emplace_front(new LargeTempTableBlock(id, schema));
     auto it = m_blockList.begin();
@@ -55,7 +55,7 @@ LargeTempTableBlock* LargeTempTableBlockCache::getEmptyBlock(TupleSchema* schema
     return m_blockList.front().get();
 }
 
-LargeTempTableBlock* LargeTempTableBlockCache::fetchBlock(int64_t blockId) {
+LargeTempTableBlock* LargeTempTableBlockCache::fetchBlock(LargeTempTableBlockId blockId) {
     auto mapIt = m_idToBlockMap.find(blockId);
     if (mapIt == m_idToBlockMap.end()) {
         throwSerializableEEException("Request for unknown block ID in LargeTempTableBlockCache (fetch)");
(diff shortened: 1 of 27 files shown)


src/ee/common/lttblockid.h
```diff
@@ -0,0 +1,69 @@
...
+
+
+#ifndef VOLTDB_LTTBLOCKID_H
+#define VOLTDB_LTTBLOCKID_H
+#include <inttypes.h>
+#include <ostream>
+#include "common/iosflagsaver.h"
+
+namespace voltdb {
+class LargeTempTableBlockId {
```
wolffcm: Newer code seems to use the convention of a camel case file name that is the name of the primary class in the file.  Inconsistencies in file naming is a source of frustration for me--I can never remember the file name of a class I'm interested in.  

We don't really have any formal conventions, but I'm proposing this one... Can you change the name of this file to `LargeTempTableBlockId.hpp`?  The `hpp` extension is useful as well since some editors use it to infer syntax highlighting.

src/ee/common/lttblockid.h
```diff
@@ -0,0 +1,69 @@
...
+public:
+    LargeTempTableBlockId(int32_t siteId, int32_t blockId) : m_siteId(siteId), m_bid(blockId) {}
+    // Preincrement.
+    LargeTempTableBlockId operator++() {
+      m_bid++;
```
wolffcm: Maybe name this `m_blockId` or `m_blockCounter`?

src/ee/common/lttblockid.h
```diff
@@ -0,0 +1,69 @@
...
+    bool operator<(const LargeTempTableBlockId &other) const {
+      return (m_siteId < other.m_siteId)
+              || ((m_siteId == other.m_siteId) && (m_bid < other.m_bid));
+    }
+
+    bool operator==(const LargeTempTableBlockId &other) const {
+        return m_id64 == other.m_id64;
+    }
+
+    int64_t getLongId() const {
```
wolffcm: Seems like you would want to use an unsigned integer types throughout this class?

src/ee/common/lttblockid.h
wolffcm: Would be good to have a header comment for this class, declaring it's purpose.

src/ee/common/lttblockid.h
wolffcm: The only counter argument I can think of is that there's no unsigned types in Java.

src/ee/common/lttblockid.h
wolffcm: I noticed that I've been good about doing camel case but not so good with the `hpp` extension for the classes I've recently added.

tests/ee/common/lttblockid_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include "harness.h"
+#include "common/lttblockid.h"
+
+using namespace voltdb;
+
+class LargeTempTableBlockIdTest : public Test {
```
wolffcm: I've been following the same convention for unit tests as for other C++ classes.  I think most of our newer EE unit tests do use camel case with the name of the test class as the file name.

src/ee/common/LargeTempTableBlockCache.cpp
```diff
@@ -31,9 +31,16 @@ LargeTempTableBlockCache::LargeTempTableBlockCache(Topend *topend, int64_t maxCa
     , m_maxCacheSizeInBytes(maxCacheSizeInBytes)
     , m_blockList()
     , m_idToBlockMap()
-    , m_nextId(0)
+      // Sometimes for tests we don't have a site id.
+    , m_nextId(ExecutorContext::getExecutorContext()
+                 ? (ExecutorContext::getExecutorContext()->m_siteId)
```
wolffcm: The value of `m_siteId` is typed as `int64_t`... In Walter's email he said it includes both a host ID and a thread ID...  I don't know how many bits are used by the host ID and thread ID, but how what happens if the magnitude of the site ID is larger than 2^31 -1?  Seems like this could very well be the case?

wolffcm: Really curious what kind of values will be in m_siteId for an actual running server.

BillWhite: I'm thinking the value of m_siteId is 16 bits of hostid and 16 bits of partition id.  If we had 32k hosts or 32k partitions we might overflow it.  It seems not really that problematic.

We could also make all three of them 32 or 64 bits, or something else.  I don't think it matters all that much.

src/ee/common/lttblockid.h
BillWhite: Ok.  I'm changing this.

src/ee/common/lttblockid.h
BillWhite: I thought about unsigned, but since there's no unsigned in Java, and I don't really think we'll overflow these fields, it wasn't clear that it mattered.  I think unsigned is problematic.  It leads to bugs when you compare with signed or do arithmetic with signed, and I don't really believe we will ever get numbers with a set sign bit.

But I'll change them all to unsigned.  I don't have a strong opinion.

tests/ee/common/lttblockid_test.cpp
BillWhite: Fixed.

src/ee/common/lttblockid.h
BillWhite: m_blockCounter;

tests/ee/common/lttblockid_test.cpp
BillWhite: Fixed.

src/frontend/org/voltdb/largequery/LargeBlockManager.java
```diff
@@ -47,7 +47,53 @@
...
+
+        private final long m_siteId;
+        private final long m_blockId;
+        public final long getSiteId() {
+            return m_siteId;
+        }
+        public final long getBlockId() {
+            return m_blockId;
+        }
+        @Override
```
wolffcm: It looks weird for the annotation to be before the method Javadoc comment...

src/frontend/org/voltdb/largequery/LargeBlockManager.java
```diff
@@ -47,7 +47,53 @@
...
+
+            if(bigSiteId.signum() < 0) {
+                bigSiteId.add(BIT_64);
+            }
+            if(bigBlockCounter.signum() < 0) {
+                bigBlockCounter.add(BIT_64);
+            }
+            return bigSiteId.toString() + "___" + bigBlockCounter.toString();
+        }
+     }
```
wolffcm: If you are going to use this class as the key type in a HashMap, then you should override the default `equals` and `hashCode` methods for this class.

src/frontend/org/voltdb/largequery/LargeBlockManager.java
```diff
@@ -163,10 +209,11 @@ private void clearSwapDir() throws IOException {
      * @param block        the bytes for the block
      * @throws IOException
      */
-    public void storeBlock(long id, long origAddress, ByteBuffer block) throws IOException {
+    public void storeBlock(long siteId, long blockCounter, long origAddress, ByteBuffer block) throws IOException {
```
wolffcm: Would it be better to have the client pass an instance of LargeTempTableBlockId?  I think maybe it would.

src/frontend/org/voltdb/largequery/BlockId.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * This class serves mostly as keys to the map from
+ * siteId/blockId pairs to file names.  It's also used
+ * to transmit these values to the LargeBlockManager
+ * API operations.
+ *
+ * Each BlockId object is essentially a Pair<long, long>.
+ */
+public class BlockId {
+    public BlockId(long siteId, long blockId) {
```
wolffcm: Usually all the methods appear after the member variables.

PR #5168, opened 2018-02-08
Eng 12256 Change the EE build to use CMake.

BillWhite:
The EE build used to use a python script to create a Makefile.  This changeset
uses CMake to create the makefile. See the file CMakeLists.txt for more details.

https://issues.voltdb.com/browse/ENG-12256


tests/ee/CMakeLists.txt
```diff
@@ -0,0 +1,310 @@
...
+#
+SET (VOLTDB_MANUAL_FAILING_TEST_PROGRAMS
+  )
+#
+# These are tests which are not run under valgrind
+# but are run when not in a valgrind build.  There
+# are none of them now, but there used to be and
+# may be in the future.
+SET (VOLTDB_NOVALGRIND_TEST_PROGRAMS
+  )
```
wolffcm: I would remove this.  It should be easier to just make EE unit tests pass valgrind than to avoid testing them in valgrind.

tests/frontend/org/voltdb/TestSnapshotWithViews.java
```diff
@@ -84,16 +84,14 @@ public void testExportViewSnapshotRestore() throws Exception {
         for (String ddl : ddls) {
             client.callProcedure("@AdHoc", ddl);
         }
-        StringBuilder insertSql;
-        for (int i=0;i<5000;i++) {
-            insertSql = new StringBuilder();
-            insertSql.append("insert into ex values(" + i + ");");
-            client.callProcedure("@AdHoc", insertSql.toString());
+        int numValues=5000;
```
wolffcm: I'm curious why you changed this?  The change looks fine, I'm just curious.

tests/sqlcoverage/version.txt
```diff
@@ -0,0 +1 @@
+7.5
```
wolffcm: What's this?

tools/VoltDB.cmake
```diff
@@ -0,0 +1,227 @@
...
+  SET(SUPPRESSIONS ${${PROJECT_NAME}_SOURCE_DIR}/test_utils/vdbsuppressions.supp)
+  IF ( WILL_FAIL )
+    SET(FAIL_ARG --expect-fail=true)
+  ELSE()
+    SET(FAIL_ARG --expect-fail=false)
+  ENDIF()
+  IF ( IS_VALGRIND_BUILD )
+    VALGRIND_FILE_NAME(${TARGET_DIR} ${TARGET_NAME} VGFN)
+    SET(CMD_LIST
+      ${CMAKE_SOURCE_DIR}/tools/runvalgrindtest.py
```
wolffcm: So you are sure that if there are leaks or other memory errors that tests will actually fail and Jenkins will be red?  I just want to be sure because we've had problems in the past.

wolffcm: And I mean for both EE unit tests and for JUnit tests run in Memcheck mode.

tools/build_cmake.py
```diff
@@ -0,0 +1,454 @@
...
+import argparse
+import re
+import multiprocessing
+
+########################################################################
+#
+# Make the command line parser.
+#
+########################################################################
+def makeParser():
```
wolffcm: Is this mostly stuff that used to be in buildtools.py?

build.xml
```diff
@@ -1435,75 +1442,91 @@ NATIVE EE STUFF
...
+             See the comment in that file for complete details.
+         -->
+        <arg line="--generator='${cmake.generator}'" />
+        <arg line="--build-type=${build}" />
+        <arg line="--log-level='${VOLT_LOG_LEVEL}'" />
+        <arg line="--max-processors=${cmake.build.maxcores}" />
+        <arg line="--verbose-build='${cmake.verbose.build}'" />
+    	<arg line="--verbose-config='${cmake.verbose.config}'" />
+        <arg line="--coverage" />
+        <arg line="--install" />
```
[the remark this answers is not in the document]
BillWhite: They were not used or functioning on master.  This needs to be revisited.  I don't think
it's hard, but it will require some work.

tests/frontend/org/voltdb/TestSnapshotWithViews.java
BillWhite: It was failing for me and I wondered why.  It turns out that there is a timeout which is triggered when this number is smaller, say 1000, but not 5000.  I decided it wasn't my fault after all, but I left this in, hoping that it might lead to some understanding of the real problem.

tools/build_cmake.py
BillWhite: Kind of.  The file buildtools.py is more like cmake itself, in that it makes the makefile.  The ant script called make.  Since cmake is invoked pretty much every time make is invoked, and it's really fast, it seemed more useful to make a python interface for the whole thing.

tools/VoltDB.cmake
BillWhite: I think I've tested this.  We should have tests which fail and which are expected to fail.  But then the tests ultimately pass, since we expect them to pass.  I don't know how to find expected unexpected failures.

I don't think I ever tested jUnit tests run in Memcheck mode.

tests/sqlcoverage/version.txt
BillWhite: I have no idea.

PR #5169, opened 2018-02-08
ENG-11578-pause-view-on-snaprestore

yzhang1991:
The enable/disable switch is now only checked on insertion, which is the case which our snapshot restore process hits.


yzhang1991:
Reviewed 15 of 20 files at r1, 4 of 7 files at r2, 3 of 4 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/voltdb/voltdb/5169)*

PR #5173, opened 2018-02-20
Eng 13658 gram cte nonrec

johnTcrawford:
Added "basic" (non-recursive) CTE statements to grammar; and tweaked some Recursive CTE definitions to make them usable for non-recursive CTEs too; plus some small python code changes to facilitate those changes.

### DIFF
--- a/tests/sqlgrammar/sql-grammar.txt
+++ b/tests/sqlgrammar/sql-grammar.txt
@@ -24,10 +24,10 @@
 #   4. Adding a (positive) integer before a | symbol makes the preceding option
 #      more likely; the "n|" still needs to be preceded and followed by a space.
 #      For example, in this definition:
-#          definition :== foo 2| bar
+#          definition ::= foo 2| bar
 #      the "foo" option is twice as likely as the "bar" option. You may use
 #      more than one such integer; for example, in this definition:
-#          definition :== foo 3| bar 2| null
+#          definition ::= foo 3| bar 2| null
 #      the "foo" option will be chosen half (3/6) the time, the "bar" option
 #      will be chosen a third (2/6) of the time, and the "null" option will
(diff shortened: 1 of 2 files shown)


tests/sqlgrammar/sql-grammar.txt
```diff
@@ -498,87 +498,333 @@ truncate-statement      ::= TRUNCATE TABLE {table-name}
...
+cte-col2                ::= CTE_C2
+cte-col3                ::= CTE_C3
+cte-col4                ::= CTE_C4
+cte-col5                ::= CTE_C5
+cte-col6                ::= CTE_C6
+cte-col7                ::= CTE_C7
+cte-col8                ::= CTE_C8
+cte-col9                ::= CTE_C9
+
+ctec1-9                 ::= {cte-col1:ctec1} | {cte-col2:ctec2} | {cte-col3:ctec3} | \
```
wolffcm: I have a question about this syntax... what does `{cte-col1:ctec1}` mean?  I see that `cte-col1` is defined, but I don't see `cte-col1` or `ctec1` defined anywhere in this file.

johnTcrawford: The {foo:bar} syntax is a way of using the same thing (typically a table or column name or alias) more than once in the same SQL statement. Basically, it says call this thing "bar": if "bar" has already been defined (in this statement) then use the existing value; if not, then define it as "foo" - or rather, the end result of what "foo" produces in this instance, possibly using randomization. I've used this syntax before, but there is quite a bit of it in this CTE stuff (recursive & otherwise). So, in this case, the use of {cte-col1:ctec1} means to use a previously existing value of "ctec1", if there is one; if not, define it as whatever "cte-col1" evaluates to, and use that value for "ctec1" wherever else it appears in this statement.

PR #5178, opened 2018-02-22
Added IntelliJ files.

BillWhite:
Add IntelliJ module files.  These are location independent.  That is to say, they don't depend on the file system path.


xinjiacs:
Can you add how-to-use description about the IntelliJ file? maybe like the wiki page to set up VoltDB project using Eclipse.

BillWhite:
I think this is too naive.  It doesn't work well  when switching from one branch to another.

I think we need to do some research and thinking before committing to a strategy for committing
IntelliJ files to source control.

PR #5181, opened 2018-02-22
Eng 13407 index offset log n optimization

xinjiacs:
This PR is to speed up offset query by using counting index, getting O(logN) look up other than sequential index scan O(N). It works for both ascending and descending order when applying offset. 

For example:
select * from t order by col offset N limit 1

If col is an indexed column of table t, EE should utilize the rank of value to find the target tuple directly. Now although the explained plan says this is index scan, EE actually performs N index scan to locate the right tuple.

### DIFF
--- a/src/ee/executors/indexscanexecutor.cpp
+++ b/src/ee/executors/indexscanexecutor.cpp
@@ -154,6 +154,8 @@ bool IndexScanExecutor::p_init(AbstractPlanNode *abstractNode,
     m_lookupType = m_node->getLookupType();
     m_sortDirection = m_node->getSortDirection();
 
+    m_offsetRank = m_node->hasOffsetRankOptimization();
+
     VOLT_DEBUG("IndexScan: %s.%s\n", targetTable->name().c_str(), tableIndex->getName().c_str());
 
     return true;
@@ -201,8 +203,13 @@ bool IndexScanExecutor::p_execute(const NValueArray &params)
     }
 
     // Initialize the postfilter
-    CountingPostfilter postfilter(m_outputTable, post_expression, limit, offset);
+    int postfilterOffset = offset;
+    if (m_offsetRank) {
+        postfilterOffset = CountingPostfilter::NO_OFFSET;
+    }
+    CountingPostfilter postfilter(m_outputTable, post_expression, limit, postfilterOffset);
 
+    // Progress monitor
     ProgressMonitorProxy pmp(m_engine->getExecutorContext(), this);
 
     //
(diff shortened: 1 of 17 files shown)


src/ee/executors/indexscanexecutor.cpp
```diff
@@ -154,6 +154,8 @@ bool IndexScanExecutor::p_init(AbstractPlanNode *abstractNode,
     m_lookupType = m_node->getLookupType();
     m_sortDirection = m_node->getSortDirection();
 
+    m_offsetRank = m_node->hasOffsetRankOptimization();
```
wolffcm: If `m_offsetRank` is a boolean, then maybe it should have a better name... I would kind of expect it to be numeric based on this name.

src/ee/indexes/tableindex.h
```diff
@@ -449,6 +450,11 @@ class TableIndex
         throwFatalException("Invoked non-countable TableIndex virtual method getCounterLET which has no implementation");
     }
 
+    // dense rank value tuple look up
+    virtual bool moveToRankTuple(int64_t iRank, bool forward, IndexCursor& cursor) const
```
wolffcm: What does the `i` in `iRank` stand for?

Most of the methods in this base class have nicely formatted comments that explain what they do, it would be good to  also follow this convention for your new method.  For example, is the rank 0- or 1-based, what does the return value mean, etc.

src/frontend/org/voltdb/planner/microoptimizations/OffsetQueryUsingCountingIndex.java
```diff
@@ -0,0 +1,89 @@
...
+                indexscan.getEndExpression() != null ||
+                indexscan.getSkipNullPredicate() != null ||
+                indexscan.getInitialExpression() != null) {
+            return;
+        }
+
+        if (indexscan.getInlinePlanNodes().isEmpty()) {
+            return;
+        }
+
```
wolffcm: Are there other kinds of plan nodes that could be inlined into the index scan that should prohibit this optimization?  What about aggregation?

xinjiacs:
@wolffcm Thanks for the review. SqlCoverage caught some errors at my earlier commits when I do not deal well with the rank calculation, so I believe it definitely has coverage for replicated table. For partition table, it has less coverage. Most of the offset queries on partition tables are MP queries, which will often use offset 0 on local partitions and apply real offset on the coordinator.

src/frontend/org/voltdb/planner/microoptimizations/OffsetQueryUsingCountingIndex.java
xinjiacs: @wolffcm I thought this before. Inline aggregation can be with Index Scan node, but not at the same time with limit plan node. I did not find other cases. 

For example:
SELECT val, sum(id) FROM R1 A1 group by val ORDER BY val OFFSET 2
 INDEX SCAN of "R1 (A1)" using "IDX_R1" (for sort order only)
  inline Serial AGGREGATION ops: SUM(column#0)
   inline LIMIT with parameter

Anyway, maybe we should explicitly validate the number of inline plan nodes to be 1 as well.

src/ee/indexes/tableindex.h
xinjiacs: I though "ith_rank" at the beginning, maybe it should be "denseRank" instead with more formatted comments.

PR #5183, opened 2018-02-23
Eng 13677 slow memcheck

johnTcrawford:


tests/ee/storage/LargeTempTableSortTest.cpp
```diff
@@ -481,7 +481,9 @@ int main(int argc, char* argv[]) {
     using namespace std::chrono;
     auto startTime = high_resolution_clock::now();
 
-    int rc = TestSuite::globalInstance()->runAll();
+    // Temporarily disable this test; see ENG-13702
```
BillWhite: I think this is not the right thing to do.

I think the right thing to do is either (1) comment out the test in tests/ee/CMakeLists.txt or else
add LargeTempTableSortTest to the list of tests which are not run in valgrind.

PR #5185, opened 2018-02-24
Eng-12821 Optimzed SynchronizedUndoAction and  Change LMPT use voltLoadTable path

JadenW:
rebased.

### DIFF
--- a/src/ee/common/UndoReleaseAction.cpp
+++ b/src/ee/common/UndoReleaseAction.cpp
@@ -1,0 +1,48 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2018 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "UndoReleaseAction.h"
+#include "UndoQuantum.h"
+
+namespace voltdb {
+
+
+UndoReleaseAction* UndoReleaseAction::getSynchronizeUndoAction(UndoQuantum* currUQ) {
+    return (new (*currUQ) SynchronizedUndoReleaseAction(this));
+}
+
+UndoReleaseAction* UndoReleaseAction::getDummySynchronizeUndoAction(UndoQuantum* currUQ) {
+    return (new (*currUQ) SynchronizedDummyUndoReleaseAction());
+}
+
+UndoReleaseAction* UndoOnlyAction::getSynchronizeUndoAction(UndoQuantum* currUQ) {
+    return (new (*currUQ) SynchronizedUndoOnlyAction(this));
+}
+
+UndoReleaseAction* UndoOnlyAction::getDummySynchronizeUndoAction(UndoQuantum* currUQ) {
+    return (new (*currUQ) SynchronizedDummyUndoOnlyAction());
+}
+
+UndoReleaseAction* ReleaseOnlyAction::getSynchronizeUndoAction(UndoQuantum* currUQ) {
+    return (new (*currUQ) SynchronizedReleaseOnlyAction(this));
+}
+
+UndoReleaseAction* ReleaseOnlyAction::getDummySynchronizeUndoAction(UndoQuantum* currUQ) {
+    return (new (*currUQ) SynchronizedDummyReleaseOnlyAction());
+}
+
+}
--- a/src/ee/common/UndoReleaseAction.cpp
+++ b/src/ee/common/UndoReleaseAction.cpp
@@ -17,9 +17,71 @@
 
 #include "UndoReleaseAction.h"
 #include "UndoQuantum.h"
+#include "SynchronizedThreadLock.h"
 
 namespace voltdb {
 
+void SynchronizedUndoReleaseAction::undo() {
+    assert(!SynchronizedThreadLock::isInSingleThreadMode());
+    SynchronizedThreadLock::countDownGlobalTxnStartCount(true);
+    {
+        ExecuteWithMpMemory usingMpMemory;
+        m_realAction->undo();
+    }
+    SynchronizedThreadLock::signalLowestSiteFinished();
+}
+
+void SynchronizedUndoReleaseAction::release() {
+    assert(!SynchronizedThreadLock::isInSingleThreadMode());
+    SynchronizedThreadLock::countDownGlobalTxnStartCount(true);
+    {
+        ExecuteWithMpMemory usingMpMemory;
+        m_realAction->release();
+    }
+    SynchronizedThreadLock::signalLowestSiteFinished();
+}
+
+void SynchronizedUndoOnlyAction::undo() {
+    assert (!SynchronizedThreadLock::isInSingleThreadMode());
+    SynchronizedThreadLock::countDownGlobalTxnStartCount(true);
+    {
+        ExecuteWithMpMemory usingMpMemory;
+        m_realAction->undo();
+    }
+    SynchronizedThreadLock::signalLowestSiteFinished();
+
+}
+
+void SynchronizedReleaseOnlyAction::release() {
+    assert (!SynchronizedThreadLock::isInSingleThreadMode());
+    SynchronizedThreadLock::countDownGlobalTxnStartCount(true);
+    {
+        ExecuteWithMpMemory usingMpMemory;
+        m_realAction->release();
+    }
+    SynchronizedThreadLock::signalLowestSiteFinished();
+}
+
+void SynchronizedDummyUndoReleaseAction::undo() {
+    assert(!SynchronizedThreadLock::isInSingleThreadMode());
+    SynchronizedThreadLock::countDownGlobalTxnStartCount(false);
+
+}
+
+void SynchronizedDummyUndoReleaseAction::release() {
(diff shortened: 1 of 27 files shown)


src/ee/execution/VoltDBEngine.h
```diff
@@ -792,6 +793,9 @@ class __attribute__((visibility("default"))) VoltDBEngine {
         // This stateless member acts as a counted reference to keep the ThreadLocalPool alive
         // just while this VoltDBEngine is alive. That simplifies valgrind-compliant process shutdown.
         ThreadLocalPool m_tlPool;
+
+        // static variable for loadTable Result
```
wweiss-voltdb: for sharing loadTable result (and exception) across VoltDBEngines

src/ee/storage/persistenttable.cpp
```diff
@@ -687,6 +687,7 @@ void PersistentTable::swapTable(PersistentTable* otherTable,
     assert(hasNameIntegrity(name(), otherIndexNames));
     assert(hasNameIntegrity(otherTable->name(), theIndexNames));
 
+    ConditionalExecuteOutsideMpMemory getOutOfMpMemory(m_isReplicated);
```
wweiss-voltdb: If you are going to fix swaptable in a separate ticket then don't change it here.

wolffcm: Is there a problem with swap tables?  What is it?

src/ee/storage/persistenttable.cpp
```diff
@@ -1564,25 +1565,50 @@ std::string PersistentTable::debug(const std::string& spacer) const {
...
+            target.setPendingDeleteOnUndoReleaseFalse();
+            // Then copy the source into the target
+            target.copyForPersistentInsert(source);
+        } else {
+            target = source;
+        }
+
+        insertTupleCommon(source, target, true, shouldDRStreamRows);
     }
     catch (ConstraintFailureException& e) {
```
wweiss-voltdb: I am sceptical about this check. You should minimally check the path if a load table exceeds the DR Buffer size because I don't know if it is a ConstraintFailure.

src/ee/storage/persistenttable.cpp
JadenW: When we swap two replicated tables, right now in swapExecutor, we will also swap to MP memory Context. In the end of finish swapping, we need to rebuild TableCollections from lowest site. So we need to switch to lowest Site context. But we never switch back to MP again. So the deconstruct of ExecuteWithMpMemory will fail a flag check. This should be a minor issue and only fail testSwapTables in TestSystemProcedureSuite in debug build. Also we properly don't need to swap to MP context for Swap Table. I will try to address in another PR.

src/ee/storage/table.cpp
```diff
@@ -475,16 +475,27 @@ void Table::loadTuplesFromNoHeader(SerializeInputBE &serialInput,
...
     for (int i = 0; i < tupleCount; ++i) {
-        nextFreeTuple(&target);
-        target.setActiveTrue();
-        target.setDirtyFalse();
-        target.setPendingDeleteFalse();
-        target.setPendingDeleteOnUndoReleaseFalse();
+        VOLT_DEBUG("AAA loadTuplesFromNoHeader start");
+        if (forLoadTable) {
+            // using executorContext tempStringPool for loadTable path
+            char *storage = static_cast<char *>(stringPool->allocateZeroes(m_schema->tupleLength()));
```
wolffcm: You are allocating tuple storage for each tuple being loaded?  Why can't you allocate a buffer once and reuse it?

tests/frontend/org/voltdb/jni/TestExecutionEngine.java
```diff
@@ -41,11 +41,14 @@
 import org.voltdb.TableStreamType;
 import org.voltdb.TheHashinator.HashinatorConfig;
 import org.voltdb.VoltDB;
+import org.voltdb.VoltProcedure;
```
wolffcm: You added imports but made no code changes?   I guess you don't actually need those new imports?

src/ee/storage/table.cpp
```diff
@@ -475,16 +475,25 @@ void Table::loadTuplesFromNoHeader(SerializeInputBE &serialInput,
...
+    TableTuple source;
     for (int i = 0; i < tupleCount; ++i) {
-        nextFreeTuple(&target);
-        target.setActiveTrue();
-        target.setDirtyFalse();
-        target.setPendingDeleteFalse();
-        target.setPendingDeleteOnUndoReleaseFalse();
+        if (forLoadTable) {
+            // using executorContext tempStringPool for loadTable path
+            char *storage = static_cast<char *>(stringPool->allocateZeroes(m_schema->tupleLength()));
```
wolffcm: I don't think you need to allocate new tuple storage for every tuple being inserted.  I'm also unclear on why you can't just call `nextFreeTuple` to avoid doing an extra copy.  Is there a good reason for this?

src/ee/storage/persistenttable.cpp
```diff
@@ -1564,26 +1564,54 @@ std::string PersistentTable::debug(const std::string& spacer) const {
...
-    catch (ConstraintFailureException& e) {
+        if (!ignoreTupleLimit && visibleTupleCount() >= m_tupleLimit) {
+                    char buffer [256];
+                    snprintf (buffer, 256, "Table %s exceeds table maximum row count %d",
+                            m_name.c_str(), m_tupleLimit);
+                    throw ConstraintFailureException(this, source, buffer);
+        }
+
+        if (forLoadTable) {
+            target = TableTuple(m_schema);
```
wolffcm: I'm still really unclear on why the `loadTable` path is different from others.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1593,23 +1600,33 @@ VoltDBEngine::loadTable(int32_t tableId,
...
-            }
+        ConditionalSynchronizedExecuteWithMpMemory possiblySynchronizedUseMpMemory(
+                table->isCatalogTableReplicated(), isLowestSite(), s_loadTableResult);
+        if (possiblySynchronizedUseMpMemory.okToExecute()) {
+            table->loadTuplesFrom(serializeIn, &m_stringPool, returnUniqueViolations ? &m_resultOutput : NULL, shouldDRStream, ExecutorContext::currentUndoQuantum() == NULL, true);
+            s_loadTableResult = 0;
         }
         else {
-            table->loadTuplesFrom(serializeIn, NULL, returnUniqueViolations ? &m_resultOutput : NULL, shouldDRStream);
+            if (s_loadTableResult == -1) {
```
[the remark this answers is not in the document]
JadenW: I am using an auxiliary coordination pattern ConditionalSynchronizedExecuteWithMpMemory which also used in InsertExecutor and deleteExecutor. 

If table is partitioned, no coordination is needed and every site will execute on their own to do the LoadTuplesFrom(). The static variable "s_loadTableResult" is redundant but should be no harm.

If table is replicated, we want to do the loadTuples only on loweset site and also report result or exeception on other sites. So we start countDownLatch on the lowest site, set the static variable to -1 as an indication of exception. If loadTuplesFrom() runs successfully, we switch "s_loadTableResult"  to 0 to indicate no exception. Other site should be blocked in this whole time until the deconstructor of ConditionalSynchronizedExecuteWithMpMemory release the latch from the lowest site.

Does this make sense? what race condition are you afraid of?

src/ee/common/UndoReleaseAction.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "UndoReleaseAction.h"
+#include "UndoQuantum.h"
+
+namespace voltdb {
+
+
+UndoReleaseAction* UndoReleaseAction::getSynchronizeUndoAction(UndoQuantum* currUQ) {
```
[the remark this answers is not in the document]
JadenW: fixed, thanks.

tests/frontend/org/voltdb/jni/TestExecutionEngine.java
JadenW: removed, thanks.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1593,23 +1600,33 @@ VoltDBEngine::loadTable(int32_t tableId,
...
+                char msg[1024];
+                snprintf(msg, 1024, "Replicated load table threw an unknown exception on other thread for table %s",
+                           table->name().c_str());
+                VOLT_DEBUG("%s", msg);
+                throw SerializableEEException(VOLT_EE_EXCEPTION_TYPE_REPLICATED_TABLE, msg);
+            }
         }
     }
     catch (const SerializableEEException &e) {
         if (SynchronizedThreadLock::isInSingleThreadMode()) {
```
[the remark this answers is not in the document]
JadenW: Right, will change that to assert.

src/ee/execution/VoltDBEngine.cpp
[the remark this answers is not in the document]
wolffcm: Yes, it makes sense now.  I was confused as to where the other threads would block waiting for the lowest site.

src/ee/storage/table.cpp
JadenW: The main change I add to LoadTable path is to enable it to throw exception when insert failed (eg. hit a constraint violation or DR buffer over limit).
The difference with previous path is following. 
It used to serialize the tuple to output buffer immediate after exception (https://github.com/VoltDB/voltdb/blob/master/src/ee/storage/persistenttable.cpp#L1561). 
For loadTable path, it has to propagate the exception which contains the tuple and serialize it at jni call (https://github.com/VoltDB/voltdb/blob/ENG-12821-bulkLoad-rebase/src/ee/voltdbjni.cpp#L445).
So for loadTable path I have to create both source tuple (for exception, reside in executorContext tempStringPool) and target Tuple (for table,  reside in threadLocal Pool). I was trying to mimic the insertExecutor path, maybe there is a better approach?

JadenW: sure. anytime after 2 o'clock is good for me.

wolffcm:
Looks good.

PR #5188, opened 2018-02-26
ENG-11578-pause-view-on-snaprestore-rebase

yzhang1991:


src/ee/storage/MaterializedViewTriggerForInsert.cpp
```diff
@@ -67,9 +76,95 @@ MaterializedViewTriggerForInsert::~MaterializedViewTriggerForInsert() {
     BOOST_FOREACH (auto aggExpr, m_aggExprs) {
         delete aggExpr;
     }
+    m_dest->setMaterializedViewTrigger(NULL);
     m_dest->decrementRefcount();
 }
 
+void MaterializedViewTriggerForInsert::setEnabled(bool enabled) {
+    if (m_supportSnapshot) {
+        // Only views that can be snapshoted are allowed to be disabled.
```
BillWhite: What if m_supportSnapshot is false?  Do we just not do anything?  Shouldn't we mark m_enabled false as well?  What if m_supportSnapshot is false but enabled is true?  Maybe that case is impossible.

src/ee/storage/MaterializedViewTriggerForInsert.cpp
```diff
@@ -67,9 +76,95 @@ MaterializedViewTriggerForInsert::~MaterializedViewTriggerForInsert() {
...
+    int aggOffset = m_groupByColumnCount;
+    for (int aggIndex = 0, columnIndex = aggOffset; aggIndex < m_aggColumnCount; aggIndex++, columnIndex++) {
+        NValue existingValue = m_existingTuple.getNValue(columnIndex);
+        NValue newValue = deltaTuple.getNValue(columnIndex);
+        if (newValue.isNull()) {
+            newValue = existingValue;
+        }
+        else {
+            switch(m_aggTypes[aggIndex]) {
+                case EXPRESSION_TYPE_AGGREGATE_SUM:
```
BillWhite: Do we support AVG here?

src/ee/storage/MaterializedViewTriggerForInsert.cpp
yzhang1991: The m_enabled default is always true. If m_supportSnapshot is false, m_enabled can never be set to false.

src/ee/storage/MaterializedViewTriggerForInsert.cpp
yzhang1991: No we do not support AVG in views

src/ee/storage/persistenttable.cpp
```diff
@@ -760,7 +761,9 @@ void PersistentTable::setDRTimestampForTuple(ExecutorContext* ec, TableTuple& tu
 
 void PersistentTable::insertTupleIntoDeltaTable(TableTuple& source, bool fallible) {
     // If the current table does not have a delta table, return.
-    if (! m_deltaTable) {
+    // If the current table has a delta table, but it is used by
+    // a single table view during snapshot restore process, return.
+    if (! m_deltaTable || m_mvTrigger) {
```
wolffcm: Hmm, it seems very subtle to make the delta table used for join views, and for single-table views but only for snapshot restore.  Would be great if someday they used the same mechanism.

src/ee/storage/persistenttable.cpp
```diff
@@ -1261,18 +1262,16 @@ void PersistentTable::deleteTuple(TableTuple& target, bool fallible) {
     // Note that this is guaranteed to succeed, since we are inserting an existing tuple
     // (soon to be deleted) into the delta table.
     insertTupleIntoDeltaTable(target, fallible);
-    {
-        SetAndRestorePendingDeleteFlag setPending(target);
+    SetAndRestorePendingDeleteFlag setPending(target);
```
wolffcm: I guess this variable `setPending` is only needed to make the newly deleted row appear temporarily not deleted when updating the views?  If that's the case, I prefer the old revision where `setPending` was more explicitly scoped.

src/ee/storage/persistenttable.cpp
```diff
@@ -2194,20 +2193,43 @@ void PersistentTable::configureIndexStats() {
...
+    // When adding view handlers from partitioned tables to replicated source tables, all partitions race to
+    // add the delta table for the replicated table. Therefore, it is likely that the first to add the delta
+    // table is not the lowest site. All add Views are done holding a global mutex so structure management is
+    // safe. However when the replicated table is deallocated it also deallocates the delta table so the memory
+    // allocation of the delta table needs to be done in the lowest site thread's context.
+    assert(m_deltaTable == NULL);
+    VOLT_TRACE("%s to check the memory context to use.\n", needToCheckMemoryContext?"Need":"No need");
+    ConditionalExecuteWithMpMemory usingMpMemoryIfReplicated(m_isReplicated && needToCheckMemoryContext);
+    TableCatalogDelegate* tcd = engine->getTableDelegate(m_name);
+    m_deltaTable = tcd->createDeltaTable(*engine->getDatabase(), *engine->getCatalogTable(m_name));
```
wolffcm: Is there really no need to synchronize here?  Even if we are modifying a partitioned table, if we are allocating it in MP pools that are not thread-safe.

tests/frontend/org/voltdb/TestSnapshotWithViews.java
```diff
@@ -187,16 +187,78 @@ public void testUpdateInExportViewSnapshotRestore() throws Exception {
...
+            "create view v_ex_p (i, counti) AS select i, count(*) from ex1 group by i",
+            // randomly partitioned view
+            "create view v_ex_np (j, countj) AS select j, count(*) from ex1 group by j"
+        };
+
+        for (String ddl : ddls) {
+            client.callProcedure("@AdHoc", ddl);
+        }
+        StringBuilder insertSql;
+        for (int i=0;i<5000;i++) {
```
wolffcm: Seems like a lot of rows for a JUnit.  Do you need that many?

wolffcm: I mean I guess that `5000` is used throughout this test, but... maybe the test would run faster if this number were smaller.  Does it time out when run in valgrind?

src/ee/storage/MaterializedViewTriggerForInsert.cpp
```diff
@@ -67,9 +76,116 @@ MaterializedViewTriggerForInsert::~MaterializedViewTriggerForInsert() {
...
+    // We already ensured this in its parent call.
+    const bool noNeedToCheckMemoryContext = false;
+    if (! m_enabled && ! m_dest->isPersistentTableEmpty()) {
+        // If the view maintenance is disabled, and the view is not empty,
+        // we need to use a delta table to hold the view content restored from
+        // the snapshot and do a manual merge afterwards.
+        m_dest->instantiateDeltaTable(noNeedToCheckMemoryContext);
+    }
+    else if (m_enabled && m_dest->deltaTable()) {
+        // When we turn on the maintenance, if a delta table exists, it means that the view table was
```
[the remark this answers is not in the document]
yzhang1991: For joined views, the source tables have to be empty, but not for single table views.
The link you attached here showed a change in the view handler, which is only used for joined views.
I wish we could merge the code path someday, but currently it has to be the way it is.
I will add a comment to explain.

PR #5191, opened 2018-02-28
ENG-13493: performance greatly improved by balancing replicate table AdHoc read queries

xinjiacs:
AdHoc single partition replicate table read only queries currently do not have partition parameter. NULL partition value gets routed to partition 0 always. This means that all AdHoc_SP_RO queries without partition key are only executed on one partition per cluster, which can lead to performance bottleneck. This commit is to address the above performance issue by round robin routing those queries to each partition in the cluster.

### DIFF
--- a/src/frontend/org/voltdb/InvocationDispatcher.java
+++ b/src/frontend/org/voltdb/InvocationDispatcher.java
@@ -120,7 +120,7 @@ public final class InvocationDispatcher {
 
     // Next partition to service adhoc replicated table reads
     private static int m_nextPartition = -1;
-    // Number of partitions, will change when new node joins cluster
+    // Number of partitions, will NOT change when new node joins cluster
     private static int m_partitionCount;
     // the partition id list, which does not assume starting from 0
     private static ArrayList<Integer> m_partitionIds;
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/ProcedureRunner.java
```diff
@@ -533,6 +534,13 @@ public boolean checkPartition(TransactionState txnState, TheHashinator hashinato
                 // ClientInterface should pre-validate this param is valid
                 parameterAtIndex = invocation.getParameterAtIndex(0);
                 parameterType = VoltType.get((Byte) invocation.getParameterAtIndex(1));
+
+                if (parameterAtIndex == null && m_isReadOnly) {
```
luyangco: I remember you mentioned to me that for the AdHoc readonly query on replicated table, you refill the partition parameter with a new value from the partition key table, and use this trick to bypass the EE partition check. But code doesn't looks like doing this way.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -117,6 +118,13 @@
 
     private final NTProcedureService m_NTProcedureService;
 
+    // Next partition to service adhoc replicated table reads
+    private static int m_nextPartition = -1;
+    // Number of partitions, will change when new node joins cluster
```
luyangco: Be sure to update the comment with "elastic join won't change partition count here, because it is created during initialization".

src/frontend/org/voltdb/ProcedureRunner.java
xinjiacs: I did that first but found out the the EE hashinator does not check the misrouted transactions. So it should be simpler in this way.

PR #5197, opened 2018-03-03
Fix issue with truncate table and switching memory contexts

wolffcm:
The test TestSQLFeaturesSuite was failing in memcheck and debug builds with failed assertions.  The issue was that the automatic guards (like ExecuteWithMpMemory and such) assume that they start from the local site context, which is not necessarily the case.  

I added a class called ExecuteWithAllSitesMemory that returns to the previous memory context, whatever it was, when it goes out of scope.  Maybe all such classes should do this.  This ensures that we can continue to pass existing assertions in the code enforcing how we switch memory contexts.

The issue here was with the truncate path, which calls `rebuildTableCollections` from the storage layer, which needs to do a lot of switching to the memory context of each site.

Summary of changes:
- Moved ExecuteWithMpMemory and related classes to separate file
- Added new class ExecuteWithAllSitesMemory
- Made static variable s_enginesByPartitionId private
- Removed references to DEBUG, as NDEBUG is standard


src/ee/common/SynchronizedThreadLock.h
```diff
@@ -133,6 +142,7 @@ class SynchronizedThreadLock {
     static bool isInLocalEngineContext();
 #ifndef  NDEBUG
```
wweiss-voltdb: I am not clear on the NDEBUG vs DEBUG flags but I saw that you changed  the code from a ifndef NDEBUG to ifdef DEBUG. Does that also apply here?

src/ee/common/ExecuteWithMpMemory.cpp
```diff
@@ -0,0 +1,111 @@
...
+#endif
+{
+    assert(SynchronizedThreadLock::isLowestSiteContext());
+}
+
+ExecuteWithAllSitesMemory::~ExecuteWithAllSitesMemory()
+{
+    ExecutorContext::assignThreadLocals(m_engineLocals);
+#ifndef NDEBUG
+    SynchronizedThreadLock::setUsingMpMemory(m_wasUsingMpMemory);
```
wweiss-voltdb: Is an assert(getThreadPartitionId() == getEnginePartitionId()) appropriate here?

src/ee/common/SynchronizedThreadLock.h
wolffcm: I changed something *to* `#ifdef DEBUG`?  I meant to get rid of all of those and replace with `#ifndef NDEBUG`.  The symbol `NDEBUG` is more standard and controls whether asserts are enabled at compile time.  It's confusing to have both.

src/ee/common/ExecuteWithMpMemory.cpp
```diff
@@ -0,0 +1,111 @@
...
+#ifndef NDEBUG
+        , m_wasUsingMpMemory(SynchronizedThreadLock::usingMpMemory())
+#endif
+{
+    assert(SynchronizedThreadLock::isLowestSiteContext());
+}
+
+ExecuteWithAllSitesMemory::~ExecuteWithAllSitesMemory()
+{
+    ExecutorContext::assignThreadLocals(m_engineLocals);
```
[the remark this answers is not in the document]
wolffcm: Hmm, the name assumeLocalSiteContext suggests that the context is going back to the lowest site ID's context.  But in this case the caller may have started in some other context (like the MP context) and we're returning to that here.  So I think we still need both.

src/ee/common/ExecuteWithMpMemory.cpp
wolffcm: No, I don't think so, we may have started in some other context.

PR #5199, opened 2018-03-05
ENG-13701:

wweiss-voltdb:
There are two maps which track the set of blocks that are non-empty. These maps are used to determine if a block is in a bucket. However, an optimization to keep a cache block around when the table's tuple count goes to zero does not cleanly track this buffer either when the last tuple is deallocated nor subsequently when the first tuple is allocated. Therefore, this block may not participate in subsequent compactions.

This branch adds special checks to ensure that the last empty block is correctly tracked when the last tuple is deleted and when this block is used by the first tuple that is subsequently inserted.

### DIFF
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -237,6 +237,10 @@ void PersistentTable::nextFreeTuple(TableTuple* tuple) {
         VOLT_TRACE("GRABBED FREE TUPLE!\n");
         stx::btree_set<TBPtr >::iterator begin = m_blocksWithSpace.begin();
         TBPtr block = (*begin);
+        if (m_tupleCount == 0) {
+            assert(m_blocksNotPendingSnapshot.find(block) == m_blocksNotPendingSnapshot.end());
+            m_blocksNotPendingSnapshot.insert(block);
+        }
         std::pair<char*, int> retval = block->nextFreeTuple();
 
         /**
--- a/src/ee/storage/persistenttable.h
+++ b/src/ee/storage/persistenttable.h
@@ -568,7 +568,8 @@ private:
     }
 
     bool blockCountConsistent() const {
-        return m_blocksNotPendingSnapshot.size() == m_data.size();
+        // if the table is empty, the empty cache block will not be present in m_blocksNotPendingSnapshot
+        return isPersistentTableEmpty() || m_blocksNotPendingSnapshot.size() == m_data.size();
     }
 
     void snapshotFinishedScanningBlock(TBPtr finishedBlock, TBPtr nextBlock) {
@@ -652,6 +653,8 @@ private:
     };
 
     TableTuple lookupTuple(TableTuple tuple, LookupType lookupType);
+
+    TBPtr allocateFirstBlock();
 
     TBPtr allocateNextBlock();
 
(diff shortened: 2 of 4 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -237,6 +237,10 @@ void PersistentTable::nextFreeTuple(TableTuple* tuple) {
         VOLT_TRACE("GRABBED FREE TUPLE!\n");
         stx::btree_set<TBPtr >::iterator begin = m_blocksWithSpace.begin();
         TBPtr block = (*begin);
+        if (m_tupleCount == 0) {
+            assert(m_blocksNotPendingSnapshot.find(block) == m_blocksNotPendingSnapshot.end());
```
BillWhite: So, if `block` has space, and `m_tupleCount` == 0, then presumable `block` must be
_all_ space.  So it must be cached.  So it can't be pending a snapshot.  This seems to imply
that `m_blocksNotPendingSnapshot`  really should be `m_blocksNotPendingSnapshotAndNotCached'
or something like that.

src/ee/storage/persistenttable.h
```diff
@@ -1032,11 +1035,19 @@ inline void PersistentTable::deleteTupleStorage(TableTuple& tuple, TBPtr block,
...
     }
 
-    if (block->isEmpty() && (m_data.size() > 1 || deleteLastEmptyBlock)) {
-        // Release the empty block unless it's the only remaining block and caller has requested not to do so.
-        // The intent of doing so is to avoid block allocation cost at time tuple insertion into the table
-        m_data.erase(block->address());
-        m_blocksWithSpace.erase(block);
+    if (block->isEmpty()) {
+        if (m_data.size() > 1 || deleteLastEmptyBlock) {
+            // Release the empty block unless it's the only remaining block and caller has requested not to do so.
```
BillWhite: What if m_data.size() == 0?  Is that possible?  Do we need an assert?

src/ee/storage/persistenttable.h
```diff
@@ -1069,6 +1080,12 @@ inline TBPtr PersistentTable::findBlock(char* tuple, TBMap& blocks, int blockSiz
     return TBPtr(NULL);
 }
 
+inline TBPtr PersistentTable::allocateFirstBlock() {
+    TBPtr block(new TupleBlock(this, TBBucketPtr()));
```
BillWhite: I think this is the real fix.  The first block is allocated from it's own bucket. `TBBucketPtr()q and subsequent blocks are allocated from a shared bucket.  Is that right?

src/ee/storage/persistenttable.h
```diff
@@ -568,7 +568,8 @@ class PersistentTable : public Table, public UndoQuantumReleaseInterest,
     }
 
     bool blockCountConsistent() const {
-        return m_blocksNotPendingSnapshot.size() == m_data.size();
+        // if the table is empty, the empty cache block will not be present in m_blocksNotPendingSnapshot
```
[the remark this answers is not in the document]
wweiss-voltdb: It is more accurate to say that m_blocksNotPendingSnapshot is the set of blocks that are in one of the m_blocksNotPendingSnapshotLoad buckets. In point of fact, assuming we keep compaction, we could completely remove m_blocksWithSpace, and just use the m_blocks[Not]PendingSnapshot sets instead. The only catch is that then we would never use the free space of a block currently used by the COWIter (which IMHO adds complexity to the free row calculation process anyway).

The lifecycle for these blocks is that when a block is created, it is placed in the NotPending Set. When a snapshot is started the Pending and NotPending sets are swapped. As each block is streamed (order is determined by the order of the blocks in m_data), it is removed from Pending. When the streaming of the block is completed, the block is moved to NotPending. If compaction could tolerate 95% empty blocks that were in fact completely empty, we could completely remove the notion of a cache block.

src/ee/storage/persistenttable.cpp
wweiss-voltdb: And sadly m_blocksPendingSnapshotAndNotCached.

src/ee/storage/persistenttable.h
wweiss-voltdb: Not quite. Since other blocks are guaranteed to have a tuple assigned to them (because this is only called from the insert path), the block will automatically be placed in the 95% empty bucket. The assert was actually when a few tuples were added (the block just stayed in the NotPendingSnapshot set and the bucket), and then they were all deleted, which caused us to remove the last block from the set and bucket. At that point, it was never re-added to the set and bucket after the first new Tuple was added.

src/ee/storage/persistenttable.h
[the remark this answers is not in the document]
BillWhite: That's great for a pull request.  It would be better if that very paragraph was someplace in the code.  It may be already, I don't know.  You would know better than I do.  Can you add this somewhere?  This seems really important.

src/ee/storage/persistenttable.h
BillWhite: Yes.  So the first block is owned by a table-specific bucket, and all other blocks are owned by some other bucket.  When the table is done with the block it releases its reference and the block is returned to the specific bucket which owns it.  In this way the first tuple block is never really released until the table is destroyed.

This may not be right, but it's my understanding, or misunderstanding.

PR #5200, opened 2018-03-05
Add JetBrains metadata to the git repository.

BillWhite:
Add new .gitignore lines.

Commit the .idea and *.iml files, as recommended by JetBrains.

This also makes voltdb/lib/python a pycharm project and txnid2 an IntelliJ module.  These are included as examples of how to make projects and modules from subtrees of the VoltDB source tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,18 +8,14 @@
 /*.sublime-project
 /*.sublime-workspace
 /.tags
-/.idea
 /.DS_Store
 .DS_Store
 .cache/
 .pydevproject
 *.pyc
 *.swp
-*~
 *runapp*.py
 *.log
-*.eml
-*.iml
 /build.local
 buildstring.txt
 host*+0000.txt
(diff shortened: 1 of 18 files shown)


.gitignore
```diff
@@ -69,17 +65,143 @@ bundles
...
+
+# directory configuration
+.dir-locals.el
+
+# These are Vim crufty stuff file names.  The list comes from github.
+
+# Swap
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-v][a-z]
```
wolffcm: These patterns seem a little bit like overkill... Any three letter file name that begins with an `s` will be ignored?

.gitignore
```diff
@@ -69,17 +65,143 @@ bundles
...
+# KDE directory preferences
+.directory
+
+# Linux trash folder which might appear on any partition or disk
+.Trash-*
+
+# .nfs files are created when an open file is removed but is still being accessed
+.nfs*
+
+# These are emacs cruft files.  The list comes from Github.
```
wolffcm: Would be good to include a URL where on can find the list...

BillWhite:
I think this is not well thought out.  We need to organize this better.

PR #5202, opened 2018-03-06
ENG-13723: shared replicated: valgrind reports invalid reads during shutdown

wolffcm:
Fix was to explicitly shutdown the IPC EEs from Java class ExecutionEngineIPC.  This forces synchronous shutdown of all the engines (lowest site last).

### DIFF
--- a/tests/frontend/org/voltdb/regressionsuites/TestIndexOffsetSuite.java
+++ b/tests/frontend/org/voltdb/regressionsuites/TestIndexOffsetSuite.java
@@ -124,7 +124,7 @@ public class TestIndexOffsetSuite extends RegressionSuite {
         // check expression index
 
         // non-deterministic query, but our index is compacting index without holes
-        if (!isHSQL()) {
+        if (!isHSQL() && !isValgrind()) {
             callWithExpectedTupleId(client, 10, "TU1_ABS_POINTS", 0, 0);
             callWithExpectedTupleId(client, 1, "TU1_ABS_POINTS", 0, 1);
             callWithExpectedTupleId(client, 8, "TU1_ABS_POINTS", 0, 5);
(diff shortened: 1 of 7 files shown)


tests/frontend/org/voltdb/regressionsuites/TestIndexOffsetSuite.java
```diff
@@ -137,31 +137,35 @@ public void testSingleColumnIndex() throws Exception {
...
-        callWithExpectedTupleId(client, 1, "TM1_POINTS", 0, 2);
-        callWithExpectedTupleId(client, 2, "TM1_POINTS", 0, 3);
-        callWithExpectedTupleId(client, 3, "TM1_POINTS", 0, 4);
-        callWithExpectedTupleId(client, 4, "TM1_POINTS", 0, 5);
-        callWithExpectedTupleId(client, 5, "TM1_POINTS", 0, 6);
-        callWithExpectedTupleId(client, 6, "TM1_POINTS", 0, 7);
-        callWithExpectedTupleId(client, 7, "TM1_POINTS", 0, 8);
-        callWithExpectedTupleId(client, 8, "TM1_POINTS", 0, 9);
-        callWithExpectedTupleId(client, Integer.MIN_VALUE, "TM1_POINTS", 0, 10);
+        if (!isValgrind()) {
```
wweiss-voltdb: Most other places have: if (isValgrind()) return;

wolffcm: Yes, that may be true.  If I put in an early return though, some future developer could add a test to the end of the test method and it would never get test in memcheck, even if it was possible to reliably do so.  I would like to err on the side of testing more stuff in valgrind.

PR #5206, opened 2018-03-07
ENG-13314: report mp pool meory consumption on lowest site; only acco…

JadenW:
…unt table/index stats once on lowest site

### DIFF
--- a/src/ee/common/SynchronizedThreadLock.cpp
+++ b/src/ee/common/SynchronizedThreadLock.cpp
@@ -114,7 +114,7 @@ void SynchronizedThreadLock::init(int32_t sitesPerHost, EngineLocals& newEngineL
             s_mpEngine.stringData = new CompactingStringStorage();
 
             delete s_mpEngine.allocated;
-            s_mpEngine.allocated = new std::size_t;
+            s_mpEngine.allocated = new std::size_t(0);
         }
     }
 }
@@ -349,7 +349,7 @@ long int SynchronizedThreadLock::getThreadId() {
 #endif
 }
 
-EngineLocals SynchronizedThreadLock::getMpEngineForTest() {return s_mpEngine;}
+EngineLocals SynchronizedThreadLock::getMpEngine() {return s_mpEngine;}
 
 void SynchronizedThreadLock::resetEngineLocalsForTest() {
     s_mpEngine = EngineLocals(true);
--- a/src/ee/common/SynchronizedThreadLock.h
+++ b/src/ee/common/SynchronizedThreadLock.h
@@ -112,7 +112,7 @@ public:
     static long int getThreadId();
     static void resetEngineLocalsForTest();
     static void setEngineLocalsForTest(int32_t partitionId, EngineLocals mpEngine, SharedEngineLocalsType enginesByPartitionId);
-    static EngineLocals getMpEngineForTest();
+    static EngineLocals getMpEngine();
 
 private:
     static bool s_inSingleThreadMode;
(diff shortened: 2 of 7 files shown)


src/ee/common/SynchronizedThreadLock.h
```diff
@@ -112,7 +112,7 @@ class SynchronizedThreadLock {
     static long int getThreadId();
     static void resetEngineLocalsForTest();
     static void setEngineLocalsForTest(int32_t partitionId, EngineLocals mpEngine, SharedEngineLocalsType enginesByPartitionId);
-    static EngineLocals getMpEngineForTest();
+    static EngineLocals getMpEngine();
```
wolffcm: I think this can be made a private method now that DRBinaryLogTest is a friend.

src/ee/common/ThreadLocalPool.cpp
```diff
@@ -531,22 +532,30 @@ void ThreadLocalPool::freeExactSizedObject(std::size_t sz, void* object)
     pool->free(object);
 }
 
-std::size_t ThreadLocalPool::getPoolAllocationSize() {
-    size_t bytes_allocated =
-        *static_cast< std::size_t* >(pthread_getspecific(m_allocatedKey));
+std::size_t ThreadLocalPool::getPoolAllocationSize_internal(size_t *bytes, CompactingStringStorage *poolMap){
```
wolffcm: This method uses no member variables, so it can be a non-member function private to this file (either by declaring it static or putting it in an anonymous namespace).

src/ee/common/ThreadLocalPool.h
```diff
@@ -195,6 +195,8 @@ class ThreadLocalPool {
 #endif
 
     static void resetStateForDebug();
+private:
+    static std::size_t getPoolAllocationSize_internal(size_t *bytes_allocated, CompactingStringStorage *poolMap);
```
wolffcm: I think you can remove this function as a member method as I said above.  But ideally this class would have all private member variables, and they would all be at the end of the class declaration, with member methods appearing first.  If it's not difficult (i.e., just compiles) could you remedy this?  (This is partly my fault since I added `resetStateForDebug` after the member variables).

tests/ee/storage/DRBinaryLog_test.cpp
```diff
@@ -252,13 +252,13 @@ class DRBinaryLogTest : public Test {
     {
         m_engine = new MockVoltDBEngine(CLUSTER_ID, &m_topend, &m_enginesPool, &m_drStream, &m_drReplicatedStream);
         s_clusterMap[CLUSTER_ID] = ClusterCtx(m_engine,
-                                              SynchronizedThreadLock::getMpEngineForTest(),
+                                              SynchronizedThreadLock::s_mpEngine,
```
wolffcm: You could use the accessor here...

src/ee/common/SynchronizedThreadLock.cpp
```diff
@@ -349,7 +349,7 @@ long int SynchronizedThreadLock::getThreadId() {
 #endif
 }
 
-EngineLocals SynchronizedThreadLock::getMpEngineForTest() {return s_mpEngine;}
+EngineLocals SynchronizedThreadLock::getMpEngine() {return s_mpEngine;}
```
wolffcm: I think there are no other uses of this method other than DRBinaryLogTest, so maybe this accessor can go away entirely.

JadenW: This is no longer need for DRBinaryLog_test, but still need for the ThreadLocalPool. (For access MP from lowest site when reporting mp pool allocation size)

PR #5210, opened 2018-03-13
Eng 13694 Shared Replicated Table View Crash

BillWhite:
Use the right topend when communicating with Java.

Sometimes sites will do work on behalf of other sites, particularly when updating materialized views with are joins of shared replicated tables and partitioned tables.  In this case we need to use the top end of the actual working site to communicate to Java, even if the working site is working on behalf of some other site.

https://issues.voltdb.com/browse/ENG-13694


src/ee/common/executorcontext.cpp
```diff
@@ -135,30 +151,44 @@ ExecutorContext::~ExecutorContext() {
...
-    pthread_setspecific(static_key, NULL);
+    pthread_setspecific(logical_executor_context_static_key, NULL);
+    pthread_setspecific(physical_topend_static_key, NULL);
 }
 
 void ExecutorContext::assignThreadLocals(const EngineLocals& mapping)
 {
-    pthread_setspecific(static_key, const_cast<ExecutorContext*>(mapping.context));
+    // At this point the logical and actual executor contexts
+    // must be equal.
```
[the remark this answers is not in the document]
wolffcm: But I'm not sure this comment is accurate: there's a few places where the lowest site ID thread needs to act on behalf of the executor context for each site, so it could be the case that the logical context is some context that is not the current physical context.

src/ee/common/executorcontext.hpp
```diff
@@ -426,6 +448,14 @@ class ExecutorContext {
...
 
   private:
+    /**
+     * This holds the top end for this executor context.  Don't
+     * use this, however.  Use the result of calling getPhysicalTopend().
+     * This is because sometimes this ExecutorContext is used by some
+     * other site when this site is a free rider.  In this case we will
+     * always, always, always want to use the top end of the site
+     * actually doing the work.
+     */
```
wolffcm: You can just get rid of m_topend, right?  Why keep it?

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
```diff
@@ -2351,6 +2350,43 @@ public void testCreateViewWithParams() throws Exception {
                 expectedMsg);
     }
 
+    public void testEng13694() throws Exception {
```
wolffcm: I would just omit this test.  If it takes 20 seconds on your local machine, it could take much longer under other conditions.

src/ee/common/executorcontext.cpp
```diff
@@ -135,30 +151,44 @@ ExecutorContext::~ExecutorContext() {
...
+    pthread_setspecific(logical_executor_context_static_key, NULL);
+    pthread_setspecific(physical_topend_static_key, NULL);
     ThreadLocalPool::resetStateForDebug();
 }
 
 void ExecutorContext::bindToThread()
 {
-    pthread_setspecific(static_key, this);
+    pthread_setspecific(logical_executor_context_static_key, this);
+    // At this point the logical and working sites must be the
```
wolffcm: "Working" sites... hmm.  I'd really prefer these comments to use consistent terms for clarity.  I like "logical" vs "physical" just fine.

src/ee/common/executorcontext.cpp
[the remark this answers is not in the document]
BillWhite: Sorry, this was an old comment.  I deleted it.

src/ee/common/executorcontext.hpp
BillWhite: No, we get a value for m_topend when the executor context is created.  We need to hold on to it until bind_to_thread is called.  After that it's never used.  Maybe bind_to_thread should zero it out, just to make sure that it's never used again?

tests/frontend/org/voltdb/regressionsuites/TestMaterializedViewSuite.java
BillWhite: Ok.  I'm only interested in seeing if this test fails without my fix anyway.

src/ee/common/executorcontext.cpp
BillWhite: Ok.  Fixed.

src/ee/common/serializeio.cpp
```diff
@@ -38,7 +38,7 @@ void FallbackSerializeOutput::expand(size_t minimum_desired) {
     ::memcpy(fallbackBuffer_, data(), position_);
     setPosition(position_);
     initialize(fallbackBuffer_, maxAllocationSize);
-    ExecutorContext::getExecutorContext()->getTopend()->fallbackToEEAllocatedBuffer(fallbackBuffer_, maxAllocationSize);
+    ExecutorContext::getExecutorContext()->getPhysicalTopend()->fallbackToEEAllocatedBuffer(fallbackBuffer_, maxAllocationSize);
```
wweiss-voltdb: So to be clear, you are calling a static function to get the singleton ExecutorContext to call it's static? Will the compiler compile ExecutorContext::getExecutorContext()->getPhysicalTopend() into ExecutorContext:getPhysicalTopend()?

PR #5211, opened 2018-03-13
Eng 13693 gram shared repl

johnTcrawford:
ENG-13693: SQL-grammar-gen: modified the run.sh script (and related test-tools.sh), adding build-pro[-if-needed], server-pro[-if-needed], and all-pro options/functions; and now allow all of the build... options to accept arguments (e.g. -Dbuild=debug and/or -DVOLT_POOL_CHECKING=true, which are needed for ENG-13693). Plus, it distinguishes between build args and other arguments, and passes a
'--build=debug' arg to the build_udf_jar.sh script, when appropriate.

### DIFF
--- a/tests/sqlgrammar/run.sh
+++ b/tests/sqlgrammar/run.sh
@@ -57,7 +57,6 @@ function find-directories-if-needed() {
 
 # Build VoltDB: 'community', open-source version
 function build() {
-    BUILD_ARGS=$ARGS
     echo -e "\n$0 performing: build$BUILD_ARGS"
     test-tools-build
     code[0]=$code_tt_build
@@ -65,7 +64,6 @@ function build() {
 
 # Build VoltDB: 'pro' version
 function build-pro() {
-    BUILD_ARGS=$ARGS
     echo -e "\n$0 performing: build-pro$BUILD_ARGS"
     test-tools-build-pro
     code[0]=$code_tt_build
@@ -73,14 +71,12 @@ function build-pro() {
 
 # Build VoltDB, only if not built already
 function build-if-needed() {
-    BUILD_ARGS=$ARGS
     test-tools-build-if-needed
     code[0]=$code_tt_build
 }
 
 # Build VoltDB 'pro' version, only if not built already
 function build-pro-if-needed() {
-    BUILD_ARGS=$ARGS
     test-tools-build-pro-if-needed
     code[0]=$code_tt_build
 }
@@ -122,6 +118,7 @@ function debug() {
     echo "UDF_TEST_DIR   :" $UDF_TEST_DIR
     echo "UDF_TEST_DDL   :" $UDF_TEST_DDL
     echo "BUILD_ARGS     :" $BUILD_ARGS
+    echo "BUILD_UDF_ARGS :" $BUILD_UDF_ARGS
     echo "DEFAULT_ARGS   :" $DEFAULT_ARGS
     echo "ARGS           :" $ARGS
     echo "SEED           :" $SEED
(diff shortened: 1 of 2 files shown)


tests/sqlgrammar/run.sh
```diff
@@ -55,19 +55,32 @@ function find-directories-if-needed() {
     fi
 }
 
-# Build VoltDB
+# Build VoltDB: 'community', open-source version
 function build() {
-    echo -e "\n$0 performing: build"
+    echo -e "\n$0 performing: build$BUILD_ARGS"
```
wolffcm: You probably want a space in there before `$BUILD_ARGS` so the output looks right?

johnTcrawford: $BUILD_ARGS (and also $ARGS) always starts with a space, so this is deliberate.

johnTcrawford:
ENG-13693: SQL-grammar-gen: modified the run.sh script (and related test-tools.sh), adding build-pro[-if-needed], server-pro[-if-needed], and all-pro options/functions; and now allow all of the build... options to accept arguments (e.g. -Dbuild=debug and/or -DVOLT_POOL_CHECKING=true, which are needed for ENG-13693). Plus, it distinguishes between build args and other arguments, and passes a
'--build=debug' arg to the build_udf_jar.sh script, when appropriate.

tests/sqlgrammar/run.sh
```diff
@@ -104,6 +117,8 @@ function debug() {
     echo "SQLGRAMMAR_DIR :" $SQLGRAMMAR_DIR
     echo "UDF_TEST_DIR   :" $UDF_TEST_DIR
     echo "UDF_TEST_DDL   :" $UDF_TEST_DDL
+    echo "BUILD_ARGS     :" $BUILD_ARGS
+    echo "BUILD_UDF_ARGS :" $BUILD_UDF_ARGS
```
BillWhite: Aren't all these environment variables confusing?  Why didn't you just have arguments
to the run.sh script which get passed on to the build_udf.sh script?  Figuring this out to
fix ENG-13725 cost me a day.

tests/sqlgrammar/run.sh
```diff
@@ -264,27 +320,33 @@ function tests-help() {
...
 # Print a simple help message, describing the options for this script
 function help() {
-    echo -e "\nUsage: ./run.sh {build|init|debug|jars|server|ddl|prepare|tests-only|tests|shutdown|all|tests-help|test-tools-help|help}\n"
+    echo -e "\nUsage: ./run.sh {build[-pro]|init|debug|jars|server[-pro]|ddl|prepare[-pro]|"
+    echo -e "  tests-only|tests|shutdown|all[-pro]|tests-help|test-tools-help|help}\n"
     echo -e "This script is used to run the SQL-grammar-generator tests, and the various things that"
     echo -e "  go with them, e.g., building and running a VoltDB server, or shutting it down afterward."
     echo -e "Options:"
     echo -e "    build           : builds VoltDB ('community', open-source version)"
+    echo -e "    build-pro       : builds VoltDB ('pro' version)"
```
BillWhite: Why are these build and build-pro arguments here at all?  Why not just call ant before calling run.sh?

One could ask the same thing of the server and server-pro arguments, but they seem more like run time things.

tests/sqlgrammar/run.sh
johnTcrawford: Yes, you're probably right. I'm not going to change it now, since these changes are already on master, and I thought they were already on ENG-13611-shared-replicated-tables last week; but I'll try to fix this at some point in the future.

tests/sqlgrammar/run.sh
johnTcrawford: Actually, that's because just before I created this run.sh file, you said to me, half-joking "friends don't let friends use ant", so I thought: you know, he's right, ant can be pretty annoying, I'll just write a run.sh (bash) script instead. The run.sh script does call ant for the build part, so you totally could call ant first, and then call the script with whatever (non-build) args you want. But back then I did not envision running SQL-grammar-gen on any config other than a voltdb (community), release build, so the 'build' option only had one version, with no args passed to it.

PR #5221, opened 2018-03-20
Make loading of replicated table CUSTOMER_NAME not use the deprecated…

wolffcm:
… voltLoadTable API

Also, this benchmark seemed to have a bug where in CUSTOMER_NAME has
more than a million rows (it should have the same as the CUSTOMER
table)

PR #5222, opened 2018-03-20
Make loading of replicated table CUSTOMER_NAME not use the deprecated…

wolffcm:
… voltLoadTable API

Also, this benchmark seemed to have a bug where in CUSTOMER_NAME has
more than a million rows (it should have the same as the CUSTOMER
table)


wolffcm:
Reviewed by Ethan in pull request #5221

PR #5223, opened 2018-03-20
ENG-13468 ZooKeeper client xid overflow handling

zeemanhe:
Client uses signed 32-int as xids which are assumed to be non-negative. Zookeeper uses some negative values as special XIDs (e.g. -2 for ping, -4 for auth). However, ZK client does not ensure the xids it generates are non-negative, and the server doesn't reject negative xids. If signed 32-int overflows, xids become negative, which will break the system.

The solution is to reset xid to 1 when it overflows.


luyangco:
The Xid overflows in just 4 days in customer environment, which means the `getXid()` was called every 0.08 millisecond, I'd like to know more about which code path calls it in such an extreme manner.

Anyway the fix looks straight forward to me.

PR #5230, opened 2018-03-29
ENG-13785 Add GCC7 support.

BillWhite:
It turns out that U17.10 uses GCC7.  This has been improved, so that it causes extra warnings.  Since a warning for us is an error, this is quite annoying.  This commit fixes all the warnings, and things seem to build ok.

Mostly I added a lot of comments with the word "FALLTHRU".  Gcc is happy with this.  There is a compiler neutral way of doing this, but it is not in the standard until C++17.  There is a less creepy way to do this as a gnu extension, but it will not work for clang, and will cause compilation errors on the mac, I fear.  This works for GCC and G++, and is not likely to break anything else.


BillWhite:
Closed to recreate this.

BillWhite:
Note that all the .iml files and all files inside .idea are JetBrains metadata.  This will be merged when ENG-13698 is merged.

BillWhite:
It turns out that the form of comment you wanted didn't work.  The compiler and the documentation don't match 100%.  But I found a form which is close and which seems to work ok.

PR #5231, opened 2018-03-29
ENG-13582 locking mechanism for uac, rejoining etc

zeemanhe:
add a locking mechanism to avoid stepping on each other while creating blockers on ZooKeeper for  exclusive validations of rejoining, UAC, etc

### DIFF
--- a/src/frontend/org/voltcore/zk/ZooKeeperLock.java
+++ b/src/frontend/org/voltcore/zk/ZooKeeperLock.java
@@ -49,16 +49,21 @@ public class ZooKeeperLock implements Watcher {
     }
 
     /**
-     *
-     * @param timeout  timeout in seconds
+    * @return true if a lock is successfully acquired
+    */
+    public boolean acquireLock() {
+        return acquireLockWithTimeout(0);
+    }
+
+    /**
+     * @param timeout  timeout in millisecond
      * @return true if a lock is successfully acquired
      */
-    public boolean acquireLockck(long timeout) {
+    public boolean acquireLockWithTimeout(long timeout) {
         try {
             //Create a sequential node, example: /db/action_lock/lock0000000000
             m_currentPath = m_zk.create(m_lockPath, null, Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL_SEQUENTIAL);
             synchronized(m_lock) {
-                long totalWaitingTime = TimeUnit.SECONDS.toMillis(timeout);
                 long startTime = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
                 while(true) {
                     List<String> nodes = getAllLockingNodes();
@@ -70,11 +75,15 @@ public class ZooKeeperLock implements Watcher {
                         return true;
                     } else {
                         //does not get the lock this time. Wait for next update upon node deletion or addition if not timeout
-                        long nextWaitingTime = totalWaitingTime - (TimeUnit.NANOSECONDS.toMillis(System.nanoTime()) - startTime);
-                        if (nextWaitingTime <= 0 && totalWaitingTime > 0) {
-                            return false;
+                        if (timeout > 0) {
+                            long nextWaitingTime = timeout - (TimeUnit.NANOSECONDS.toMillis(System.nanoTime()) - startTime);
+                            if (nextWaitingTime <= 0) {
+                                return false;
+                            }
+                            m_lock.wait(nextWaitingTime);
+                        } else {
+                            m_lock.wait();
                         }
-                        m_lock.wait(nextWaitingTime);
                     }
                 }
             }
--- a/src/frontend/org/voltcore/zk/ZooKeeperLock.java
+++ b/src/frontend/org/voltcore/zk/ZooKeeperLock.java
@@ -20,24 +20,27 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
-
+import java.util.concurrent.TimeUnit;
 import org.apache.zookeeper_voltpatches.AsyncCallback.VoidCallback;
 import org.apache.zookeeper_voltpatches.CreateMode;
 import org.apache.zookeeper_voltpatches.KeeperException;
 import org.apache.zookeeper_voltpatches.WatchedEvent;
 import org.apache.zookeeper_voltpatches.Watcher;
 import org.apache.zookeeper_voltpatches.ZooDefs.Ids;
+import org.voltcore.logging.VoltLogger;
+
 import org.apache.zookeeper_voltpatches.ZooKeeper;
 
 //Use the mechanism similar to leader election by creating sequential ephemeral nodes.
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltcore/zk/ZooKeeperLock.java
```diff
@@ -0,0 +1,109 @@
...
+    private String m_currentPath = null;
+    final Object lock = new Object();
+
+    public ZooKeeperLock(ZooKeeper zk, String nodePath, String lockName) {
+        m_zk = zk;
+        m_basePath = nodePath;
+        m_lockPath = ZKUtil.joinZKPath(m_basePath, lockName);
+    }
+
+    public void acquireLock() throws IOException {
```
luyangco: It would give caller more flexibility to add another API `acquireLockWithTimeout(int milliseconds)`.

src/frontend/org/voltcore/zk/ZooKeeperLock.java
```diff
@@ -0,0 +1,109 @@
...
+        m_zk.sync(m_basePath, new VoidCallback() {
+            @Override
+            public void processResult(int rc, String path, Object ctx) {
+                latch.countDown();
+            }
+        }, null);
+        latch.await();
+
+        //get a list of sequential nodes and watch on updates on this node
+        return m_zk.getChildren(m_basePath, this);
```
luyangco: Wrong path? If set watch on base path the watcher will be waked up after uac, rejoin, elastic join, ban elastics join and balanceSPI blockers creation. Maybe you just want to watch on ``m_lockPath``.

src/frontend/org/voltcore/zk/ZooKeeperLock.java
```diff
@@ -0,0 +1,109 @@
...
+            throw new IOException(e);
+        }
+    }
+
+    private List<String> getAllLockingNodes() throws KeeperException, InterruptedException {
+        //ensure that any updates with the lower sequence numbers are committed on both leader
+        //and all followers to avoid any missing reads on followers because ZooKeeper does not guarantee a write
+        //is immediately visible on all followers.
+        CountDownLatch latch = new CountDownLatch(1);
+        m_zk.sync(m_basePath, new VoidCallback() {
```
luyangco: `m_lockPath` instead? Because with the new lock, it's impossible to race with action blocker creations.

src/frontend/org/voltcore/zk/ZooKeeperLock.java
```diff
@@ -0,0 +1,109 @@
...
+    public ZooKeeperLock(ZooKeeper zk, String nodePath, String lockName) {
+        m_zk = zk;
+        m_basePath = nodePath;
+        m_lockPath = ZKUtil.joinZKPath(m_basePath, lockName);
+    }
+
+    public void acquireLock() throws IOException {
+        try {
+            //Create a sequential node, example: /db/action_blockers/lock0000000000
+            m_currentPath = m_zk.create(m_lockPath, null, Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL_SEQUENTIAL);
```
luyangco: I'm a little confused, is the lock created at `/db/action_blockers/lock/000000000` or `/db/action_blockers/lock000000000`? If the latter is true, then all other action blocker nodes are also create at `/db/action_blockers/` path, would sorting always put sequential nodes first?

src/frontend/org/voltcore/zk/ZooKeeperLock.java
```diff
@@ -0,0 +1,109 @@
...
+            synchronized(lock) {
```
luyangco: I feel like the locking is unnecessary because the protocol is thread-safe even if two threads on the same node attempt to acquire the lock, and also every acquire create a new `ZooKeeperLock` instance so everyone who failed to acquire is waiting on their own locks.

luyangco: Same applies to `public void process(WatchedEvent event)`.

src/frontend/org/voltdb/VoltZK.java
```diff
@@ -382,6 +384,23 @@ public static boolean zkNodeExists(ZooKeeper zk, String node)
      * @return null for success, non-null for error string
      */
     public static String createActionBlocker(ZooKeeper zk, String node, CreateMode mode, VoltLogger hostLog, String request) {
+        //Acquire a lock before creating a blocker and validate actions.
+        ZooKeeperLock lock = new ZooKeeperLock(zk, VoltZK.actionBlockers, "lock");
+        try {
+            lock.acquireLock();
+            return setActionBlocker(zk, node, mode, hostLog, request);
+        } catch (IOException e) {
+            VoltDB.crashLocalVoltDB("Unable to create action blocker " + node, true, e);
```
luyangco: Error message doesn't reflect the actual cause.

src/frontend/org/voltcore/zk/ZooKeeperLock.java
zeemanhe: will update

src/frontend/org/voltcore/zk/ZooKeeperLock.java
zeemanhe: basePath is the correct directory to be watched because the sequential nodes are under the basePath. Any update on the basePath will trigger the callback and notify all who are waiting for the lock-next one with lowest sequence gets the lock.

src/frontend/org/voltcore/zk/ZooKeeperLock.java
zeemanhe: The lock is  /db/action_blockers/lock000000000.   The children under  /db/action_blockers will be lock000000000, lock000000001, lock000000002, ...   Whoever has the lowest sequence will get the lock, others wait for their turns

luyangco: If the new lock is created under /db/action_blockers, would other action blocker nodes (uac, rejoin, elastic join etc.) affect the sorting order of children, for example after sort /db/action_blockers/elastic_join is in front of /db/action_blockers/lock0000000000.

zeemanhe: good catch

src/frontend/org/voltdb/VoltZK.java
```diff
@@ -382,6 +389,22 @@ public static boolean zkNodeExists(ZooKeeper zk, String node)
...
+        try {
+            if(!zklock.acquireLockWithTimeout(TimeUnit.SECONDS.toMillis(60))) {
+                return "Could not acquire a lock to create action blocker:" + request;
+            }
+        } finally {
+            try {
+                zklock.releaseLock();
+            } catch (IOException e) {}
+        }
+        return setActionBlocker(zk, node, mode, hostLog, request);
```
luyangco: So why not call `setActionBlocker()` within the try catch block? What if after zklock released, another node took the zk lock again and both nodes run setActionBlocker in parallel?

luyangco:
It looks fine to me, feel free to merge when Jenkins is happy.

PR #5232, opened 2018-03-29
Eng 13698 Add JetBrains Metadata Maybe For Real This Time.

BillWhite:
This commit adds metadata files for IntelliJ and CLion to the VoltDB Community branch, including the pro files.  I don't know what happens if you don't have a pro repository in "voltdb/../pro".

To use IntelliJ with this, open up the IntelliJ project in voltdb.  To use this with CLion open up the CLion project in voltdb/src/ee.  IntelliJ and CLion projects pretty much can't share the same folder, so they have to be separate.

The IntelliJ project has several modules:
1. The voltcore module, which is the core of the VoltDB server.
1. The voltcore.examples.* modules, which are all the examples.
1. The voltcore.test_apps.* modules, which are all the test applications.

So, we can edit examples and test_apps using IntelliJ.  There are no Eclipse projects for these, so editing them does not have IDE support easily.  Navigation seems to work for these, though I'm not 100% sure about all the examples and test applications.

The CLion project has the sources in src/ee and the tests in tests/ee.  Navigation seems to work for both of them, though I only did a cursory inspection.  The JetBrains advice is to put the file .idea/workspace.xml in .gitignore.  But that's where the CMake configurations are stored.  So, I added it.  That means that some user configuration settings will be pushed into github.  I don't think this is a security issue, but people have different ideas about user settings, and this may cause problems.  Just keep an eye out for it. 

To review these I would advise checking out the branch and then opening it up in IntelliJ and CLion.

All the .iml files are to describe modules.


BillWhite:
I don't think this works.  I'm withdrawing it.

PR #5235, opened 2018-04-02
Fix ENG-13747.

BillWhite:
We didn't return a boolean typed null value from is_valid_timestamp.
I also changed the return types of some other functions which return
null values.  Better the right type than the null type.


BillWhite:
There is one which I don't know how to resolve.  The function DECODE can return multiple types.  We  could do a better job, but it would be some work.  If the Planner gave us some direction here perhaps it would be better.

PR #5237, opened 2018-04-03
Suppress -Wuser-defined-warnings

yzhang1991:


tools/VoltDBCompilation.cmake
```diff
@@ -199,6 +199,9 @@ ELSEIF (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     # related to the LLVM versioning numbers.
     VOLTDB_ADD_COMPILE_OPTIONS(-Wno-unused-local-typedefs -Wno-absolute-value)
   ENDIF()
+  IF ( "9.1.0" VERSION_LESS ${CMAKE_CXX_COMPILER_VERSION} )
```
wolffcm: The compiler version on my machine is equal to 9.1.0, so does this work?  I would have expected you to either use a less-than-or-equal operator, or to test for 9.0.0.

PR #5238, opened 2018-04-04
Eng 13611 add thread info to debug logs

wweiss-voltdb:


src/ee/common/ThreadLocalPool.cpp
```diff
@@ -596,6 +600,18 @@ void ThreadLocalPool::setPartitionIds(int32_t partitionId) {
...
+    return partitionId;
+}
+
+int32_t ThreadLocalPool::getEnginePartitionId() {
+    int32_t partitionId =
+        *static_cast< int32_t* >(pthread_getspecific(m_enginePartitionIdKey));
+    return partitionId;
+}
+
+int32_t ThreadLocalPool::debugGetThreadPartitionId() {
```
wolffcm: So there are three variants of this method now?  `debugGetThreadPartitionId` as well as `getThreadPartitionIdForTest` and the vanilla `getThreadPartitionId`?

wolffcm: Ok, I understand better what you're doing here.  I think we should be consistent in how we name methods that are not going to be used in the product as it's deployed in production.  Prefixing a method with `debug` doesn't seem to have any precedent...  How about `getThreadPartitiondIdWithNullCheck`? or `getThreadPartitionIdForDebug`?

Also, do you think the null check is so costly that you really need two separate methods?  I feel like the answer is no, since the null check should be really cheap, and there's already so many additional changes introduced since 8.0 that are way more costly.

src/ee/common/ThreadLocalPool.cpp
```diff
@@ -191,15 +191,19 @@ void ThreadLocalPool::assignThreadLocals(const PoolLocals& mapping)
...
     pthread_setspecific(m_enginePartitionIdKey, NULL);
-
-    delete static_cast<int32_t*>(pthread_getspecific(m_threadPartitionIdKey));
     pthread_setspecific(m_threadPartitionIdKey, NULL);
 }
+int32_t* ThreadLocalPool::getThreadPartitionIdForTest() {
+    return static_cast< int32_t* >(pthread_getspecific(m_threadPartitionIdKey));
+}
+void ThreadLocalPool::setThreadPartitionIdForTest(int32_t* partitionId) {
+    pthread_setspecific(m_threadPartitionIdKey, static_cast<const void*>(partitionId));
```
wolffcm: So...I think we need to be explicit about who owns the memory referenced by a thread local key.  Back when we didn't need to update them really at all it wasn't a big deal.  What is the ownership policy?  It would be good to encapsulate this in a comment in ThreadLocalPool.h.

PR #5242, opened 2018-04-06
ENG-13778-srt-swap-drop

yzhang1991:
This is an old problem that manifested recently with the shared replicated table commit.
In EE, there are two types of `StatsSource`: one for tables and one for indexes. Both of them cache the table name in an `NValue` so they don't have to construct one every time when they produce a stats temp table.
Before we have the @SwapTables procedure, a table's name cannot be changed once it is created because we do not support `ALTER TABLE RENAME`. But when you swap two tables, the table names are also swapped. Previously, we forgot to update the cached `NValue` in the index and table stats to reflect the new table name. So the statistics that we have been providing, in this case, is always wrong.

In the shared replicated table work, we have a new change in Java where, before returning the stats, we retrieve the table name from the stats table and get the catalog table object to see if it is replicated. If you swap A and B, then drop A, B's stats still show A's name, but A does not exist in the catalog anymore. That's how the NPE is resulted from.

The fix for this problem is to update the table name cached in the index or table stats when swapping two tables.

In this fix, I moved the table name NValue from IndexStats and TableStats to the base class because both index and table stats need it.
I also removed the unused `m_name` string in the base class. In addition, I found the previous bug ENG-13798 I fixed can also be rewritten using the new `getTableName()` I added to the StatsSource base class. So I updated the fix.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -907,8 +907,6 @@ VoltDBEngine::processCatalogDeletes(int64_t timestamp, bool updateReplicated,
     // delete tables in the set
     bool isReplicatedTable;
     BOOST_FOREACH (auto path, deletions) {
-        VOLT_TRACE("delete path:");
-
         // If the delete path is under the catalog functions item, drop the user-defined function.
         if (startsWith(path, catalogFunctions.path())) {
             catalog::Function* catalogFunction =
@@ -1628,14 +1626,17 @@ VoltDBEngine::loadTable(int32_t tableId,
 }
 
 /*
- * Delete and rebuild id based table collections. Does not affect
- * any currently stored tuples.
+ * Delete and rebuild the relativeIndex based table collections.
+ * It does not affect any currently stored tuples.
  */
 void VoltDBEngine::rebuildTableCollections(bool updateReplicated, bool fromScratch) {
+    VOLT_DEBUG("UpdateReplicated = %s,%s from scratch",
+               updateReplicated ? "true" : "false",
+               fromScratch ? "" : " not");
     // 1. See header comments explaining m_snapshottingTables.
     // 2. Don't clear m_exportTables. They are still exporting, even if deleted.
     // 3. Clear everything else.
-    if (!updateReplicated && fromScratch) {
+    if (! updateReplicated && fromScratch) {
         m_tables.clear();
         m_tablesByName.clear();
         m_tablesBySignatureHash.clear();
@@ -1645,18 +1646,17 @@ void VoltDBEngine::rebuildTableCollections(bool updateReplicated, bool fromScrat
         getStatsManager().unregisterStatsSource(STATISTICS_SELECTOR_TYPE_INDEX);
     }
 
-    // walk the table delegates and update local table collections
+    // Walk through table delegates and update local table collections
     BOOST_FOREACH (LabeledTCD cd, m_catalogDelegates) {
         auto tcd = cd.second;
         assert(tcd);
-        if (!tcd) {
+        if (! tcd) {
             continue;
         }
         Table* localTable = tcd->getTable();
         assert(localTable);
-        if (!localTable) {
-            VOLT_ERROR("DEBUG-NULL");//:%s", cd.first.c_str());
-            std::cout << "DEBUG-NULL:" << cd.first << std::endl;
+        if (! localTable) {
+            VOLT_ERROR("DEBUG-NULL: %s", cd.first.c_str());
             continue;
         }
         assert(m_database);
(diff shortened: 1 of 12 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1676,30 +1676,29 @@ void VoltDBEngine::rebuildTableCollections(bool updateReplicated, bool fromScrat
...
+                        for (auto engineIt = execAllSites.begin(); engineIt != execAllSites.end(); ++engineIt) {
                             EngineLocals& curr = engineIt->second;
                             VoltDBEngine* currEngine = curr.context->getContextEngine();
                             SynchronizedThreadLock::assumeSpecificSiteContext(curr);
                             currEngine->m_tablesBySignatureHash[hash] = persistentTable;
                         }
                     }
                 }
-                else {
+                else if (! updateReplicated) {
```
wolffcm: Was this a bug?  I guess it was just a redundant assignment before your change?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1628,14 +1626,17 @@ VoltDBEngine::loadTable(int32_t tableId,
 }
 
 /*
- * Delete and rebuild id based table collections. Does not affect
- * any currently stored tuples.
+ * Delete and rebuild the relativeIndex based table collections.
+ * It does not affect any currently stored tuples.
  */
 void VoltDBEngine::rebuildTableCollections(bool updateReplicated, bool fromScratch) {
```
wolffcm: One could imagine refactoring our code so that we can simplify logic in methods like this.  We do a lot of stuff twice---once for the partitioned tables, on all threads, and then again on the lowest site thread for the replicated tables.  It seems like we actually want to keep the tables separated into partitioned and replicated collections.

The "updatedRelicated" flag is easy to lose track of.

src/ee/indexes/IndexStats.cpp
```diff
@@ -129,16 +129,10 @@ IndexStats::IndexStats(TableIndex* index)
  * Configure a StatsSource superclass for a set of statistics. Since this class is only used in
  * the EE it can be assumed that it is part of an Execution Site and that there is a site Id.
  * @parameter name Name of this set of statistics
```
wolffcm: What sort of name is passed in here?  Is it the type of statistics gathered, like cardinality, unique cardinality, etc?

src/ee/stats/StatsAgent.cpp
```diff
@@ -61,8 +61,11 @@ void StatsAgent::registerStatsSource(StatisticsSelectorType sst,
                                      CatalogId catalogId,
                                      StatsSource* statsSource) {
     assert(statsSource != NULL);
-    m_statsCategoryByStatsSelector[sst].insert(
-        pair<CatalogId, StatsSource*>(catalogId, statsSource));
+    m_statsCategoryByStatsSelector[sst].insert(pair<CatalogId, StatsSource*>(catalogId, statsSource));
```
wolffcm: This could be made a little more concise with `std::make_pair`.

src/ee/stats/StatsSource.cpp
```diff
@@ -83,16 +83,16 @@ void StatsSource::configure(string name) {
...
     m_hostname.free();
 }
 
-/**
- * Retrieve the name of this set of statistics
- * @return Name of statistics
- */
-string StatsSource::getName() {
-    return m_name;
+const char* StatsSource::getTableName() {
```
wolffcm: Will this return a null-terminated string?  If not you should add a comment saying so. 
 You could return an `std::string` here instead...?

src/ee/stats/StatsSource.cpp
```diff
@@ -83,16 +83,16 @@ void StatsSource::configure(string name) {
     m_statsTuple = m_statsTable->tempTuple();
 }
 
+void StatsSource::updateTableName(std::string tableName) {
```
wolffcm: This method could take `const std::string &` as an argument.

src/ee/execution/VoltDBEngine.cpp
yzhang1991: Yes, redundant assignment.

src/ee/stats/StatsSource.cpp
yzhang1991: Yes, I always lose attention to things like this.

src/ee/execution/VoltDBEngine.cpp
yzhang1991: I created https://issues.voltdb.com/browse/ENG-13804 to document this.

src/ee/stats/StatsAgent.cpp
yzhang1991: I made this change.

src/ee/stats/StatsSource.cpp
yzhang1991: Yes, this is actually coming from the NValue storage. I don't want to change it back to string because we only use this now in the DEBUG output, where I have to call `c_str()` anyway.

PR #5249, opened 2018-04-12
C++ header cleanup

yzhang1991:


wolffcm:
You did this with an automated tool, like CLion or something?

Did you make sure this still builds with different conditionally compiled settings (sadly we have a lot of these now)?
- release
- debug
- memcheck
- debug with pool checking
- with logging enabled

If so then this looks great.

yzhang1991:
I will double-check with pool checking, the rest are all good.

PR #5255, opened 2018-04-16
Adjust testAdHocLengthLimit

yzhang1991:
The length prefix was expanded from a 2-byte integer to a 4-byte integer (signed).
Obviously, it now can support way more than 10x longer query string.

Getting a `10 * Short.MAX_VALUE` long query string does not cover that much of this length range, and it brings the timeout issue in the test itself.

Change to the loop condition below, so that it just verifies you can go beyond 2^15 in query string, which is enough for the scope of [ENG-9560](https://issues.voltdb.com/browse/ENG-9560).


johnTcrawford:
Looks right to me.

PR #5257, opened 2018-04-17
ENG-13816: refactor exception handling for VoltDBEngine::loadTable

wolffcm:
TestLoadingSuite was failing in memcheck mode because of subtle differences in exception handling between the IPC and JNI back ends.  It turns out we were throwing a fatal exception in some cases from non-lowest site threads when encountering a unique constraint violation when loading a replicated table.

### DIFF
--- a/src/ee/common/ExecuteWithMpMemory.h
+++ b/src/ee/common/ExecuteWithMpMemory.h
@@ -54,8 +54,23 @@ private:
 
 class ConditionalSynchronizedExecuteWithMpMemory {
 public:
+    template<class ExceptionTracker>
     ConditionalSynchronizedExecuteWithMpMemory(bool needMpMemoryOnLowestThread,
-                                               bool isLowestSite);
+                                               bool isLowestSite,
+                                               ExceptionTracker* tracker,
+                                               const ExceptionTracker& initValue)
+    : m_usingMpMemoryOnLowestThread(needMpMemoryOnLowestThread && isLowestSite)
+    , m_okToExecute(!needMpMemoryOnLowestThread || m_usingMpMemoryOnLowestThread)
+    {
+        if (needMpMemoryOnLowestThread) {
+            if (SynchronizedThreadLock::countDownGlobalTxnStartCount(isLowestSite)) {
+                VOLT_DEBUG("Entering UseMPmemory");
+                SynchronizedThreadLock::assumeMpMemoryContext();
+                // This must be done in here to avoid a race with the non-MP path.
+                *tracker = initValue;
+            }
+        }
+    }
 
     ~ConditionalSynchronizedExecuteWithMpMemory();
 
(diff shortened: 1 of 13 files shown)


src/ee/executors/deleteexecutor.cpp
```diff
@@ -96,9 +96,10 @@ bool DeleteExecutor::p_execute(const NValueArray &params) {
         assert(targetTable->isCatalogTableReplicated() ==
                 (m_replicatedTableOperation || SynchronizedThreadLock::isInSingleThreadMode()));
         ConditionalSynchronizedExecuteWithMpMemory possiblySynchronizedUseMpMemory(
-                m_replicatedTableOperation, m_engine->isLowestSite(), s_modifiedTuples);
+                m_replicatedTableOperation, m_engine->isLowestSite());
         if (possiblySynchronizedUseMpMemory.okToExecute()) {
             if (m_truncate) {
+                s_modifiedTuples = -1;
```
JadenW: Also need set s_modifiedTuples to -1 in none truncate case, right?

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1584,43 +1584,76 @@ VoltDBEngine::loadTable(int32_t tableId,
...
+        else {
+            // Offending rows will be serialized on lowest site thread.
+            return false;
         }
-        throw;
     }
+    else if (s_loadTableException != VOLT_EE_EXCEPTION_TYPE_NONE) { // some other kind of exception occurred on lowest site thread
+        // This is fatal.
+        std::ostringstream oss;
+        oss << "An unknown exception occurred on another thread when loading table \"" << table->name() << "\".";
```
JadenW: Do you want to output oss to VOLT_DEBUG?

src/ee/common/ExecuteWithMpMemory.h
```diff
@@ -54,8 +54,23 @@ class ConditionalExecuteOutsideMpMemory {
...
+                                               const ExceptionTracker& initValue)
+    : m_usingMpMemoryOnLowestThread(needMpMemoryOnLowestThread && isLowestSite)
+    , m_okToExecute(!needMpMemoryOnLowestThread || m_usingMpMemoryOnLowestThread)
+    {
+        if (needMpMemoryOnLowestThread) {
+            if (SynchronizedThreadLock::countDownGlobalTxnStartCount(isLowestSite)) {
+                VOLT_DEBUG("Entering UseMPmemory");
+                SynchronizedThreadLock::assumeMpMemoryContext();
+                // This must be done in here to avoid a race with the non-MP path.
+                *tracker = initValue;
```
JadenW: Oh, that's because we don't want SP path to change this exception tracker, right?

wolffcm: Yes, when I took it out of here, and we assign this value unconditionally, there were lots of races.

PR #5259, opened 2018-04-20
Eng 13698 Jet Brains Metadata

BillWhite:
This commit wants to add Jet Brains Metadata to VoltDB projects.

We want to support CLion and IntelliJ projects, for C++ and Java development respectively.  The two problems are that these projects cannot both reside in the same directory, and the CLion project must reside in the directory which contains the main CMakeLists.txt file.  The latter means that the CLion project has to reside in the top level directory, and the IntelliJ project must reside some place else.  

We put the CLion files in the root of the voltdb community hierarchy.  We put the IntelliJ files in a directory named "tools/IntelliJ".  The metadata files are found in a directory called .idea for both CLion and IntelliJ.

So, to open an IntelliJ project one just needs to open voltdb/tools/IntelliJ.  The .idea file will be there, and if the pro directory exists at voltdb/tools/IntelliJ/../../../pro, it will be picked up.  To open a CLion project one needs to open the project at voltdb.

Note that the pro directory needs to be a sibling of voltdb and needs to be named "pro".  If this is a problem we can try to fix this.


wolffcm:
I'm going to merge this into the branch I'm working on now, and then "do some work" like normal and see how it goes.

wolffcm:
CLion seems to work fine.

If I want to use your new project files in IntelliJ how do I do that?

wolffcm:
Ok, I can just open voltdb/tools/IntelliJ like you clearly said above.  I should have looked there first.

When I try to run regression suite JUnits, I'm not running with the right JVM settings... another thing that maybe should be in source control?

What is %MODULE_WORKING_DIR% set to in this scheme?

wolffcm:
I have some changes in my working directory after trying to use IntelliJ... in tools/IntelliJ/.idea/codeStyles/Project.xml
Not sure why.

wolffcm:
I think it would be worth defining some default cmake profiles for new users, to make it easy for them to get set up.  Maybe we put it into a template file that can just be copied into the right place.  Or maybe we can have a set of cmake profiles that might be suitable for everyone.

My only real issue is with regressionsuite tests... I haven't been able to discover the magic to get them to work.

Overall this looks very nice!  Thank you for doing this!

BillWhite:
We figured out how the regression suites need to be changed.  Unfortunately these can't be easily changed in version controlled files.  Some files are not under version control, and these are where the CLion build configurations and the JUnit test configurations are kept.  I suspect we would have to make IntelliJ and CLion plugins which set the defaults.  I'm not sure this is possible, I'm just guessing.

BillWhite:
Closing to squash to the integration branch.

PR #5261, opened 2018-04-20
ENG-13638: Disallow transaction on the producer side if the combined

JadenW:
binary log size exceeds 50 MB.

### DIFF
--- a/src/frontend/org/voltdb/exceptions/SQLException.java
+++ b/src/frontend/org/voltdb/exceptions/SQLException.java
@@ -19,6 +19,7 @@ package org.voltdb.exceptions;
 
 import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Exceptions that are intended to be caught by the user in a stored procedure are SQLExceptions
@@ -41,16 +42,19 @@ public class SQLException extends SerializableException {
         assert(state.length() == 5);
     }
 
+    public SQLException(String sqlState, String message) {
+        super(message);
+        assert(sqlState.length() == 5);
+        byte[] sqlStateBytes = null;
+        sqlStateBytes = sqlState.getBytes(StandardCharsets.UTF_8);
+        assert(sqlStateBytes.length == 5);
+        m_sqlState = sqlStateBytes;
+    }
+
     public SQLException(String sqlState) {
         assert(sqlState.length() == 5);
         byte[] sqlStateBytes = null;
-        try {
-            sqlStateBytes = sqlState.getBytes("UTF-8");
-        } catch (UnsupportedEncodingException e1) {
-            sqlStateBytes = new byte[5];
-            e1.printStackTrace();
-            assert(false);
-        }
+        sqlStateBytes = sqlState.getBytes(StandardCharsets.UTF_8);
         assert(sqlStateBytes.length == 5);
         m_sqlState = sqlStateBytes;
     }
--- a/src/frontend/org/voltdb/iv2/FragmentTask.java
+++ b/src/frontend/org/voltdb/iv2/FragmentTask.java
@@ -345,13 +345,14 @@ public class FragmentTask extends FragmentTaskBase
                 // fragment response to the network
                 final int tableSize;
                 final byte fullBacking[];
+                final int drBufferChanged;
                 try {
                     // read the complete size of the buffer used
                     fragResult.readInt();
                     // read number of dependencies (1)
                     fragResult.readInt();
                     // read the dependencyId() -1;
-                    fragResult.readInt();
+                    drBufferChanged = fragResult.readInt();
                     tableSize = fragResult.readInt();
                     fullBacking = new byte[tableSize];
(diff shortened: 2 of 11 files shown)


src/ee/common/executorcontext.cpp
```diff
@@ -323,6 +324,11 @@ void ExecutorContext::resetExecutionMetadata(ExecutorVector* executorVector) {
     }
     assert (m_tuplesModifiedStack.size() == 0);
 
+    if (m_drBufferChangedStack.size() != 0) {
```
wolffcm: So will the tuples-modified stack and DR-buffer-changed stack always be kept in sync?  If so, then maybe it makes sense to have a single stack with a non-scalar type, maybe a stack of `std::pair` or something.

src/ee/executors/insertexecutor.h
```diff
@@ -79,6 +79,7 @@ class InsertExecutor : public AbstractExecutor
         m_nowFields(),
         m_targetTable(NULL),
         m_modifiedTuples(0),
+        m_drBufferChanged(0),
```
wolffcm: This name `m_drBufferChanged` makes me think that the variable woud have a Boolean type.  Would a better name be `m_drBinaryLogByteCount or something?

src/ee/storage/persistenttable.cpp
```diff
@@ -926,9 +926,9 @@ void PersistentTable::doInsertTupleCommon(TableTuple& source, TableTuple& target
 }
 
 void PersistentTable::insertTupleCommon(TableTuple& source, TableTuple& target,
-                                        bool fallible, bool shouldDRStream, bool delayTupleDelete) {
+                                        bool fallible, bool shouldDRStream, bool delayTupleDelete, size_t* drBufferChanged) {
```
wolffcm: I wonder if there's a better way to report this information that does not involve adding a new argument to every API that inserts, updates or deletes tuples.  Temp tables, for example, won't have anything to do with DR.

It seems more appropriate to use the current undo quantum to tally the number of bytes that would be put in the binary log.  Persistent tables that are adding undo actions could update the byte count at the same time.  When the send node is executed, you could get this byte count from the current undo quantum.

Would this work, or am I misunderstanding something?

src/frontend/org/voltdb/exceptions/SQLException.java
```diff
@@ -41,6 +41,21 @@ public SQLException(ByteBuffer buffer) {
         assert(state.length() == 5);
     }
 
+    public SQLException(String sqlState, String message) {
+        super(message);
+        assert(sqlState.length() == 5);
+        byte[] sqlStateBytes = null;
+        try {
+            sqlStateBytes = sqlState.getBytes("UTF-8");
```
manjujames: If we use getBytes(StandardCharsets.UTF_8), we can avoid this exception and clean up this part.

src/frontend/org/voltdb/iv2/FragmentTask.java
```diff
@@ -343,13 +343,15 @@ public FragmentResponseMessage processFragmentTask(SiteProcedureConnection siteC
...
+                final int drBufferChanged;
                 try {
                     // read the complete size of the buffer used
                     fragResult.readInt();
                     // read number of dependencies (1)
                     fragResult.readInt();
                     // read the dependencyId() -1;
-                    fragResult.readInt();
+                    drBufferChanged = fragResult.readInt();
+                    LOG.debug("DR Buffer" + drBufferChanged);
```
manjujames: Add if check here for log stmt.

src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -398,7 +409,17 @@ private void checkForException(FragmentResponseMessage msg)
         }
     }
 
-    private boolean trackDependency(long hsid, int depId, VoltTable table)
+    private void checkForDRBufferLimit() {
+        tmLog.debug("drBufferChanged: " + m_drBufferChangedAgg + " limit:" + MAX_DR_BUFFER_SIZE);
```
manjujames: If check here.

src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -431,6 +452,9 @@ private boolean trackDependency(long hsid, int depId, VoltTable table)
             // null dependency table is from a joining node, has no content, drop it
             if (table.getStatusCode() != VoltTableUtil.NULL_DEPENDENCY_STATUS) {
                 tables.add(table);
+                tmLog.debug("[trackDependency]:  drBufferSize:" + drBufferSize + " m_drBufferChangedAgg: " + m_drBufferChangedAgg);
```
manjujames: Add if check

manjujames:
Is there some way to add a unit test for this? If there is a way to change the max buffer size just for the test, then we can test positive and negative cases.

src/ee/storage/persistenttable.cpp
JadenW: I have simply the the tracking path, this should no longer a problem.

PR #5270, opened 2018-04-25
ENG-13815: Pass in XDCR flag when we create delta table so that times…

manjujames:
…tamp column will be included in schema.


manjujames:
Right, I was wondering why isXDCR flag is passed around because it is available in the engine. I will take out the new one that I added, so that we don't make it worse. 

Regarding junits - I think I just need to add a view for one of the DR'ed tables in one of the existing tests to cover this. I am looking into that.

wolffcm:
Thanks.  

Since this bug is related to "delta tables", the test would need a DR'ed table with a view that contains a join (single-table views do not use delta tables).

manjujames:
Yes, I added the test and verified that it failed before the fix and passes after. I will merge the change now. Thanks for the help.

PR #5271, opened 2018-04-26
ENG-13752:

wweiss-voltdb:
Previously it was assumed that a trucated PBD buffer would use less space than the original buffer. However, when compression is involved, this is not necessarily the case. To protect against this case, we get the necessary length from the truncation response rather than assuming that the original compressed buffer size will be enough.

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -165,7 +165,7 @@ public abstract class PBDSegment {
     // TODO: javadoc
     abstract int size();
 
-    abstract protected int writeTruncatedEntry(BinaryDeque.TruncatorResponse entry, int length) throws IOException;
+    abstract protected int writeTruncatedEntry(BinaryDeque.TruncatorResponse entry, int hdrLength) throws IOException;
 
     /**
      * Parse the segment and truncate the file if necessary.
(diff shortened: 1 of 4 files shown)


manjujames:
Looks like there is no compression when we write truncated object. So, it looks like there was always an bug here - we are allocating buffer based on compressed size, but writing uncompressed bytes. So your fix appears to be correct, but for different reasons?

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -226,9 +226,7 @@ int parseAndTruncate(BinaryDeque.BinaryDequeTruncator truncator) throws IOExcept
                         final long partialEntryBeginOffset = reader.readOffset();
                         m_fc.position(partialEntryBeginOffset);
 
-                        // It is conceivable that a truncated buffer uses up more compressed space than the original
-                        // compressed buffer, but we won't worry about that until it happens.
-                        final int written = writeTruncatedEntry(retval, compressedLength + OBJECT_HEADER_BYTES);
+                        final int written = writeTruncatedEntry(retval, OBJECT_HEADER_BYTES);
```
manjujames: Minor - We don't need to pass in the OBJECT_HEADER_BYTES because it is available as a constant.

PR #5274, opened 2018-04-27
ENG-11221: Added 'Steps to Reproduce' to SqlCoverage output (on separate

johnTcrawford:
HTML pages, with links). Main changes to sql_coverage_test.py &
SQLCoverageReport.py (also removed an obsolete try/except); plus extra
args in normalizer.py, not-a-normalizer.py, nulls-lowest-normalizer.py,
and also build.xml.

### DIFF
--- a/tests/sqlcoverage/SQLCoverageReport.py
+++ b/tests/sqlcoverage/SQLCoverageReport.py
@@ -32,6 +32,8 @@ import traceback
 from datetime import datetime
 from distutils.util import strtobool
 from optparse import OptionParser
+from os.path import basename, isfile
+from shutil import copyfile
 from time import mktime
 from voltdbclient import VoltColumn, VoltTable, FastSerializer
 
@@ -58,7 +60,7 @@ ReproduceStatementTypes[Reproduce.DML].extend(ReproduceStatementTypes[Reproduce.
 ReproduceStatementTypes[Reproduce.CTE]  = ['WITH']
 ReproduceStatementTypes[Reproduce.CTE].extend(ReproduceStatementTypes[Reproduce.DML])
 ReproduceStatementTypes[Reproduce.ALL]  = ['']    # any statement, including SELECT
-### print "DEBUG: ReproduceStatementTypes:', str(ReproduceStatementTypes)
+### print "DEBUG: ReproduceStatementTypes:", str(ReproduceStatementTypes)
 
 def highlight(s, flag):
     if not isinstance(s, basestring):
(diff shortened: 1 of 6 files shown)


tests/sqlcoverage/SQLCoverageReport.py
```diff
@@ -32,11 +32,36 @@
...
+from os.path import basename, isfile
+from shutil import copyfile
 from time import mktime
 from voltdbclient import VoltColumn, VoltTable, FastSerializer
 
 __quiet = True
 
+# One way of representing Enums in earlier versions of Python,
+# per Stack Overflow
+def enum(*sequential, **named):
```
BillWhite: This is clever.  I like it.

tests/sqlcoverage/SQLCoverageReport.py
```diff
@@ -74,14 +99,51 @@ def generate_table_str(res, key):
...
+    if not isfile(local_ddl_file):
+        ### print 'DEBUG: copying %s to %s' % (ddl_file, local_ddl_file)
+        copyfile(ddl_file, local_ddl_file)
+    return '<p>First, run: <a href="%s">%s</a></p>' % (local_ddl_file, 'the DDL file')
+
+def generate_reproducer(item, output_dir, reproducer, ddl_file):
+    result = ''
+    if item.get("reproducer"):
+        reproducer_html = """
+<html>
```
BillWhite: It might be easier to adopt a template engine of some sort, if you find you are doing this a lot.  But I wouldn't ask for changes for that, it's just a random thought.

tests/sqlcoverage/sql_coverage_test.py
```diff
@@ -267,8 +268,9 @@ def get_config_path(basedir, config_key, config_value):
 
 
 def run_config(suite_name, config, basedir, output_dir, random_seed,
-               report_invalid, report_all, generate_only, subversion_generation,
-               submit_verbosely, ascii_only, args, testConfigKit):
+               report_invalid, report_all, reproducer, generate_only,
+               subversion_generation, submit_verbosely, ascii_only,
```
BillWhite: Don't you find these long argument lists to be annoying?  I never remember what the parameters are.  Maybe putting them all in a dict would be easier.  It's more of a style comment, though.  This is fine.

tests/sqlcoverage/sql_coverage_test.py
```diff
@@ -422,7 +424,8 @@ def run_config(suite_name, config, basedir, output_dir, random_seed,
         compare_results = imp.load_source("normalizer", config["normalizer"]).compare_results
         success = compare_results(suite_name, random_seed, statements_path, cmpdb_path,
                                   jni_path, output_dir, report_invalid, report_all, extraStats,
-                                  comparison_database, modified_sql_path, max_mismatches, within_minutes)
+                                  comparison_database, modified_sql_path, max_mismatches,
+                                  within_minutes, reproducer, config.get("ddl"))
```
BillWhite: Is the ddl file always named 'ddl.sql'?  Should this be a constant somehow?

tests/sqlcoverage/sql_coverage_test.py
johnTcrawford: Yes. They were long before I ever started working on SqlCoverage, and I've only made them longer (slowly). However, the level of annoyance is mitigated here by the fact that each one is only called in one place, in the same file where they are defined, and not changed very often.

tests/sqlcoverage/sql_coverage_test.py
johnTcrawford: No, different test suites used different DDL files. The most commonly used one is in fact called DDL.sql; but there are others called int-DDL.sql, strings-DDL.sql, etc. So this cannot be a constant.

johnTcrawford:
Thanks for the review, Bill.

PR #5276, opened 2018-04-30
Eng 13856 better message for multi txn id msg in restore

wweiss-voltdb:


luyangco:
Since the SRT branch is merged into master, could you merge this to master too.

PR #5287, opened 2018-05-02
ENG-13866, fix restart sequence generator that resart flag is not set…

luyangco:
… to the right bit.

Add unit test for restart sequence generator.

Change-Id: I1be1256462c31dd60cf1d99129382e78c63c8fca


JadenW:
LGTM.

PR #5311, opened 2018-05-10
ENG-13834: Enable distributed sorting for ORDER BY in MP queries

wolffcm:
This is work by Mike Alexeev.  Previously we only use MergeReceive in MP plans when the collector fragment produced sorted results via an index scan.  Now, we will produce an OrderBY plan node in the collector fragment, and merge-receive in the coordinator.


wolffcm:
This was reviewed by Bill and I on a separate branch.

PR #5319, opened 2018-05-14
ENG-13894:

wweiss-voltdb:
Load table was not generating consistent results on all sites in the event of a failure. While this failure should not happen, it would generate a Hash Mismatch if it does, so this code fixes this path and addes unit tests for both this failure scenario and the exception failure scenario.

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -1551,7 +1551,7 @@ VoltDBEngine::loadTable(int32_t tableId,
                         ReferenceSerializeInputBE &serializeIn,
                         int64_t txnId, int64_t spHandle, int64_t lastCommittedSpHandle,
                         int64_t uniqueId,
-                        bool throwUniqueViolations,
+                        bool returnConflictRows,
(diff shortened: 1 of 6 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1627,13 +1629,24 @@ VoltDBEngine::loadTable(int32_t tableId,
...
+        }
+
         // Indicate to other threads that load happened successfully.
         s_loadTableException = VOLT_EE_EXCEPTION_TYPE_NONE;
     }
     else if (s_loadTableException == VOLT_EE_EXCEPTION_TYPE_CONSTRAINT_VIOLATION) {
         // An constraint failure exception was thrown on the lowest site thread and
         // handle it on the other threads too.
-        if (throwUniqueViolations) {
+        if (!returnConflictRows) {
```
JadenW: I think we don't need this if check here.
For returnConflictRows is true case, there is an uncaught ConstraintFailureException for loadTuplesForLoadTable (which should not happen). If this is happeing, we did want to throw ReplicationTableException.

PR #5320, opened 2018-05-14
Add guards for large queries

yzhang1991:


wolffcm:
Hey Ethan, the ticket is actually a little ambiguous.  In some places it talks about throwing errors for stuff that we can't handle, and in other places (like the ticket's title) it talks about falling back to normal planning/execution.  It looks like you've gone the error-throwing route here.

I think falling back to normal query planning and execution is actually what we want.  It will make John's job of testing easier.

Also, the ticket mentions collecting statistics for how often we need to fall back to normal planning, I don't see that here.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -3306,6 +3306,10 @@ public boolean shutdown(Thread mainSiteThread) throws InterruptedException {
                 did_it = true;
                 m_mode = OperationMode.SHUTTINGDOWN;
 
+                if (m_catalogContext.m_ptool.getAdHocLargeFallbackCount() > 0) {
```
wolffcm: Is there just one planner tool per host?  I think it's just one, but I could be wrong.

PR #5326, opened 2018-05-15
ENG-13856:

wweiss-voltdb:
A rejoined node can participate in a snapshot as it's first transaction but does not appear to have the latest txnId because that is normally assigned as part of fragment processing. Therefore the digest of a rejoin node can contain incorrect txnIds. By assigning the TxnId from the rejoin snapshot, the value will be correctly initialized.


luyangco:
Looks good to me, wait for unit tests pass then I'll merge.

PR #5329, opened 2018-05-16
Eng 13810

GuaGuaDean:
Corrected the error message when cluster name ("database") is not consistent with what's in the .initialized file


luyangco:
Can you ask Andrew to review your error message change?

GuaGuaDean:
I have. That is Andrew's suggestion. @luyangco

PR #5330, opened 2018-05-16
ENG-13801: Extra column when aggregate appears in ORDER BY clause

wolffcm:


wolffcm:
In this bug a statement that contains `ORDER BY COUNT(*)` added the COUNT(*) to a select statement's output in addition to the items on the statement's select list.

HSQL does not have this problem, so I traced HSQL execution to see how HSQL determines the output schema for a statement.  HSQL uses a member of QuerySpecification called `indexLimitVisible` (that is, the 0-based ordinal position of the first column that is not visible in the query output.

The class `SimpleColumnContext` is used to determine which "simple columns" (HSQL's notion of a column produced by a query specification) should appear in the query output.  I added `indexLimitVisible` to this class so it can determine which columns will be in the output in a manner similar to HSQL.

src/hsqldb19b3/org/hsqldb_voltpatches/Expression.java
```diff
@@ -1636,7 +1638,7 @@ public boolean disabledTheColumnForDisplay(int jj) {
      * @throws HSQLParseException
      */
     VoltXMLElement voltGetXML(Session session) throws HSQLParseException {
-        return voltGetXML(new SimpleColumnContext(session, null), null);
+        return voltGetXML(new SimpleColumnContext(session, null, 0), null);
```
BillWhite: Are all the ignored columns at the beginning of the row?

Or is this just for HSQL objects which for which the display list is irrelevant?

That's what it seems like is happening here.

tests/frontend/org/voltdb/regressionsuites/TestFixedSQLSuite.java
```diff
@@ -3130,9 +3130,57 @@ public void testEng13852() throws Exception {
...
+                        "  WHERE POINT != OUTER_TBL.POINT " +
+                        "  ORDER BY COUNT(*)) " +
+                        "ORDER BY NUM;").getResults()[0];
+        assertContentOfTable(new Object[][] {{"foo", 1}}, vt);
+    }
+
     //
     // JUnit / RegressionSuite boilerplate
-    //
+    //z
```
BillWhite: Is this a typo?

tests/test_apps/adhocbenchmark/voltcore.testapps.adhocbenchmark.iml
```diff
@@ -2,7 +2,7 @@
 <module type="JAVA_MODULE" version="4">
   <component name="FacetManager">
     <facet type="Python" name="Python">
-      <configuration sdkName="Python 2.7" />
+      <configuration sdkName="" />
```
BillWhite: I guess this is caused by my having the Python plugin installed while nobody else does.  We should agree on one way of working.  I would vote that we not have the module installed, and we use Pycharm if we need python.
 
Not relevant to this PR, though.

tools/IntelliJ/.idea/codeStyles/Project.xml
```diff
@@ -24,29 +24,5 @@
         <option name="ITERATION_ELEMENTS_WRAPPING" value="chop_down_if_not_single" />
       </formatting-settings>
     </DBN-SQL>
-    <DBN-PSQL>
```
BillWhite: I think we should probably choose one set of formatting rules and stick to them.

Not in this PR, though.

wolffcm:
Merged on different branch!

PR #5334, opened 2018-05-18
Eng 12067

Yusarin:
Changed error message and removed all "user lacks privilege"

### DIFF
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -48,302 +48,301 @@ import org.voltdb.utils.Encoder;
  * Used only for ad hoc queries.
  */
 public class PlannerTool {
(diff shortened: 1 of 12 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -280,9 +280,9 @@ public synchronized AdHocPlannedStatement planSql(String sqlIn, StatementPartiti
                             }
 
                             AdHocPlannedStatement ahps = new AdHocPlannedStatement(sql.getBytes(Constants.UTF8ENCODING),
-                                                                                   core,
-                                                                                   params,
-                                                                                   null);
+                                    core,
```
yzhang1991: I see there is inconsistent code style here in this file. Here you used two tab spaces to indent the parameters in the new line, but you aligned rest parameters with the first one in the RuntimeException construction below. Try to use one way or the other, not mixing both.

PR #5335, opened 2018-05-19
ENG-13907 A fix: flush out mp readonly txn out of backlog on node failure

zeemanhe:


src/frontend/org/voltdb/iv2/TransactionTaskQueue.java
```diff
@@ -522,11 +521,13 @@ public String toString()
         return pendingTasks;
     }
 
-    public synchronized TransactionTask peekFirstBacklogTask() {
-        if (m_backlog.isEmpty()) {
-            return null;
+    //flush mp readonly transactions out of backlog
+    public void removeMPReadTransactions() {
```
luyangco: This method needs to be `synchronized` since multiple thread tries to touch `m_backlog`

PR #5344, opened 2018-05-23
Eng 12641

Yusarin:
Initial attempt to run Jenkins


src/frontend/org/voltdb/exportclient/ExportToFileClient.java
```diff
@@ -842,6 +892,7 @@ else if (type.equalsIgnoreCase("tsv")) {
 
         BinaryEncoding encoding = BinaryEncoding.valueOf(
                 conf.getProperty("binaryencoding", "HEX").trim().toUpperCase());
+        boolean uniquename = Boolean.parseBoolean(conf.getProperty("uniquename"));
```
manjujames: The property name suggested by Andrew is uniquenames, not uniquename.

src/frontend/org/voltdb/exportclient/ExportToFileClient.java
```diff
@@ -179,16 +180,42 @@ public boolean equals(Object obj) {
             }
 
             String getPath(String prefix) {
+                String Host = ""+VoltDB.instance().getHostMessenger().getHostId();
```
manjujames: Variable names must start with lower case. Also make this hostId, instead of host because that is more accurate.

src/frontend/org/voltdb/exportclient/ExportToFileClient.java
Yusarin: Changed to unique names

manjujames: Also, pass in default value of false.

src/frontend/org/voltdb/exportclient/ExportToFileClient.java
```diff
@@ -179,16 +180,42 @@ public boolean equals(Object obj) {
...
                 if (m_batched) {
-                    return m_dirContainingFiles.getPath() +
+                    return m_uniquename ?
```
manjujames: With this uniquenames check as well, it might be better to separate out filename construction part into a separate utility method. This utility method can do the filename part without the extension, which means the same method can be used for getPathForSchema as well. This will make it more readable and make this logic live in once place.

manjujames:
I will wait for the junit tests as well.

tests/frontend/org/voltdb/exportclient/TestExportToFileClient.java
```diff
@@ -417,6 +420,34 @@ public void testFailGetWriter() throws Exception
...
+        props.put("with-schema", "true");
+        props.put("uniquenames", "true");
+        client.configure(props);
+
+        boolean validName = true;
+        final File dir = new File(m_dir);
+        final File[] subdirs = dir.listFiles();
+        if (subdirs != null) {
+            for (File file : subdirs) {
+                if (!file.getName().startsWith("active") && !(file.getName().contains("(") && file.getName().contains(")"))) {
```
manjujames: You want junit tests to verify the following 4 cases:
- with uniquenames in batch mode
- with uniquenames in non-batch mode
- without uniquenames in batch mode
- without uniquenames in non-batch mode

Also, when you verify the filename, please change it look for <tablename>_<hostid> in the filename. Otherwise we are not really verifying that host id is getting added to the filename.

Yusarin: What do you mean by "look for _" ? I'm looking for parentheses because the hostId I created is associated with parentheses like this "(1)", and other parts of the filename will not contain parentheses.

src/frontend/org/voltdb/exportclient/ExportToFileClient.java
```diff
@@ -280,7 +280,11 @@ public String toString() {
 
         String getPathOfBatchDir(String prefix) {
             assert(m_batched);
-            return m_outDir.getPath() + File.separator + prefix + m_nonce + "-" + m_dateformat.get().format(start);
+            String hostId = "";
+            if(m_uniquenames) {
+                hostId = "-("+VoltDB.instance().getHostMessenger().getHostId()+")";
+            }
+            return m_outDir.getPath() + File.separator + prefix + m_nonce + "-" + m_dateformat.get().format(start) + hostId;
```
manjujames: This is changing the batch directory name. This shouldn't be changed. Only the filenames need to be changed.

tests/frontend/org/voltdb/exportclient/TestExportToFileClient.java
```diff
@@ -417,6 +423,203 @@ public void testFailGetWriter() throws Exception
...
+        final File dir = new File(m_dir);
+        final File[] subdirs = dir.listFiles();
+
+        List<File> remainingFiles = new ArrayList<>();
+        if (subdirs != null) {
+            for (File file : subdirs) {
+                if (file.isDirectory()) {
+                    remainingFiles.addAll(Arrays.asList(file.listFiles()));
+                }
+                else if (!file.getName().matches("\\d\\-mytable\\-\\(\\d\\).*\\.[a-z]{3,4}")) {
```
manjujames: You can just add file to remainingFiles list, which will validate the name in the loop below. That way you don't have this regex pattern in 2 places.

tests/frontend/org/voltdb/exportclient/TestExportToFileClient.java
```diff
@@ -417,6 +423,203 @@ public void testFailGetWriter() throws Exception
...
+                if (file.isDirectory()) {
+                    remainingFiles.addAll(Arrays.asList(file.listFiles()));
+                }
+                else if (!file.getName().matches("\\d\\-mytable.*\\.[a-z]{3,4}")) {
+                    validName = false;
+                    break;
+                }
+            }
+            for (File file : remainingFiles) {
+                if (!file.getName().matches("\\d\\-mytable.*\\.[a-z]{3,4}")) {
```
manjujames: I assume the hostId is always 0 in this test. In which case, without uniquenames, you probably should also verify that filename does not contain "(0)"

PR #5346, opened 2018-05-23
ENG-13857, another corner case for non-restartable MP sysproc, when t…

luyangco:
…his kind of sysprocs are interrupted by SP leader promotion, we generates an error response to MpScheduler. In MpScheduler, we fake a CompleteTransactionMessage to repair log, but in this case, we mark it as a normal CTM, it should be a CTM with abortDuringRepair flag on.

Change-Id: Ic914ecaf38ab9ece24d777ae0a0434d54e384f4c

### DIFF
--- a/src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
+++ b/src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
@@ -58,6 +58,9 @@ public class InitiateResponseMessage extends VoltMessage {
     //when the site was leader partition
     boolean m_isForOldLeader = false;
 
+    // No need to serialize it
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/messaging/InitiateResponseMessage.java
```diff
@@ -58,6 +58,9 @@
     //when the site was leader partition
     boolean m_isForOldLeader = false;
 
+    // No need to serialize it
+    public boolean m_isFromNonRestartableSysproc = false;
```
zeemanhe: Should the be serialized?

PR #5359, opened 2018-05-29
ENG-13854:

wweiss-voltdb:
The SP RepairLog and the ScoreBoard share the same message. When a local repairlog is sent to a local MPI, the MPI will modify any found completion messages before resending them back to site leaders, these modifications alter the scoreboard message causing the scoreboard to treat the original, modified completion message as a repair message before the actual repair message arrives. To prevent this, we need to make new completion messages from the original repair messages rather than reusing the message we received.


luyangco:
Looks good to me.

PR #5364, opened 2018-05-31
ENG-13854:

wweiss-voltdb:
All completion handling (even for aborted txns) need to be done on the correct site thread to avoid a locking deadlock between the TransactionTaskQueue and the Scoreboard locks.

Also make sure that abort repair messages for non-restartable txns do not have the restart flag set (for non-restartables) because otherwise the txn will not be cleaned up by the SPs.

Lastly clean up duplicate counters for all abort completions that never made it past the scoreboard because a newer abort completion (with more valid SP leaders) arrived.

### DIFF
--- a/src/frontend/org/voltdb/iv2/CompleteTransactionTask.java
+++ b/src/frontend/org/voltdb/iv2/CompleteTransactionTask.java
@@ -35,6 +35,7 @@ public class CompleteTransactionTask extends TransactionTask
 {
     final private Mailbox m_initiator;
     final private CompleteTransactionMessage m_completeMsg;
+    private boolean m_fragmentNotExecuted = false;
 
     public CompleteTransactionTask(Mailbox initiator,
                                    TransactionState txnState,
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/iv2/CompleteTransactionTask.java
```diff
@@ -35,6 +35,7 @@
 {
     final private Mailbox m_initiator;
     final private CompleteTransactionMessage m_completeMsg;
+    private boolean m_fragmentNotExecuted = false;
```
zeemanhe: I believe that a transaction with fragments executed can also be missing if a site has received CompleteTransactionResposnemessage. If that is true,  the variable name can be changed to reflect the scenario

src/frontend/org/voltdb/iv2/MpProcedureTask.java
```diff
@@ -161,7 +161,7 @@ public void run(SiteProcedureConnection siteConnection)
                     0,
                     true,
                     false,  // really don't want to have ack the ack.
-                    !m_txnState.isReadOnly(),
+                    true,   // Don't clear/flush the txn because we are restarting it
```
luyangco: Good catch, I think this flag should keep the original meaning that the CTM indicates a transaction restart, no matter it's read-only or not.

src/frontend/org/voltdb/iv2/TransactionTaskQueue.java
```diff
@@ -31,7 +31,6 @@
 import org.voltcore.messaging.Mailbox;
 import org.voltcore.utils.Pair;
 import org.voltdb.dtxn.TransactionState;
-import org.voltdb.messaging.CompleteTransactionResponseMessage;
 
 public class TransactionTaskQueue
 {
```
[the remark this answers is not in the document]
wweiss-voltdb: I think you mean TransactionTaskQueue is not referenced any more. I have removed it. Thanks.

PR #5367, opened 2018-06-04
Eng 13198: modified: run.sh, UDFBenchmark.java, UDFBenchmarkConfig.java

johnTcrawford:
(in /udfbenchmark/).


tests/test_apps/udfbenchmark/run.sh
```diff
@@ -80,7 +80,8 @@ function server() {
 
 # run the client that drives the example
 function client() {
-    $APPNAME
+    $APPNAME udf-partitioned
```
prosegay: I really don't like the two benchmarks back to back. It's not what we usually do, I think its confusing, I prefer separate client callouts for each and make one (the partitioned one) the default if they just specify client.

PR #5369, opened 2018-06-06
ENG-13941: Updated the license checking for the community edition to …

mreading:
…throw an error if the user attempts to enable Database Replication in the deployment file settings.


luyangco:
Looks good to me.

PR #5373, opened 2018-06-06
ENG-13941: Throw an error if the user attempts to enable Database Rep…

mreading:
…lication in the deployment file settings while using the community edition.


GuaGuaDean:
Good job.

PR #5374, opened 2018-06-07
ENG-12062: AAUIW Adhoc planning problems no longer print java excepti…

mreading:
…on names.


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -158,7 +158,10 @@ public synchronized CompiledPlan planSqlCore(String sql, StatementPartitioning p
                 logException(e, "Error compiling query");
                 loggedMsg = " (Stack trace has been written to the log.)";
             }
-            throw new RuntimeException("Error compiling query: " + e.toString() + loggedMsg, e);
+            if (e instanceof NumberFormatException) {
```
yzhang1991: Why is `NumberFormatException` special?

tests/sqlcmd/baselines/simple/parsedates.errbaseline
```diff
@@ -1,8 +1,8 @@
...
-Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: Error compiling query: org.voltdb.VoltTypeException: tryToMakeCompatible: The provided value: (2015-11-111 10:10:10) of type: java.lang.String is not a match or is out of range for the target parameter type: org.voltdb.types.TimestampType (Stack trace has been written to the log.)
-Unexpected Ad Hoc Planning Error: java.lang.RuntimeException: Error compiling query: org.voltdb.VoltTypeException: tryToMakeCompatible: The provided value: (2015-11-11 10:10:10.) of type: java.lang.String is not a match or is out of range for the target parameter type: org.voltdb.types.TimestampType (Stack trace has been written to the log.)
+SQL error while compiling query: tryToMakeCompatible: The provided value: (20150-11-11 10:10:10) of type: java.lang.String is not a match or is out of range for the target parameter type: org.voltdb.types.TimestampType (Stack trace has been written to the log.)
```
yzhang1991: tryToMakeCompatible is also not needed. You can remove it from the code. Look at `ParameterConverter.java`

src/frontend/org/voltdb/compiler/PlannerTool.java
mreading: While running the TestFixedSQLSuite tests, I encountered this error without this addition:

    [junit] Testcase: testFixedTickets-localCluster-1-1-HSQLDB took 1.167 sec
    [junit] 	FAILED
    [junit] Expected the statement "SELECT NUM + 11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111.1111E1 FROM R1" to throw an exception matching the pattern "java.lang.NumberFormatException", but instead it threw an exception containing "SQL error while compiling query: null (Stack trace has been written to the log.)".
    [junit] junit.framework.AssertionFailedError: Expected the statement "SELECT NUM + 11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111.1111E1 FROM R1" to throw an exception matching the pattern "java.lang.NumberFormatException", but instead it threw an exception containing "SQL error while compiling query: null (Stack trace has been written to the log.)".
    [junit] 	at org.voltdb.regressionsuites.RegressionSuite.verifyProcFails(RegressionSuite.java:1213)
    [junit] 	at org.voltdb.regressionsuites.RegressionSuite.verifyStmtFails(RegressionSuite.java:1188)
    [junit] 	at org.voltdb.regressionsuites.TestFixedSQLSuite.subTestENG7480(TestFixedSQLSuite.java:2683)
    [junit] 	at org.voltdb.regressionsuites.TestFixedSQLSuite.testFixedTickets(TestFixedSQLSuite.java:899)

Without printing the java exception error, the overall error messages makes no sense - see the "...query: null" - so I chose to keep the original format for NumberFormatException cases

mreading: (you have to scroll far to the right)

tests/sqlcmd/baselines/simple/parsedates.errbaseline
mreading: Should I remove all VoltTypeException instances of tryToMakeCompatible in that file?

src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -158,10 +158,10 @@ public synchronized CompiledPlan planSqlCore(String sql, StatementPartitioning p
                 logException(e, "Error compiling query");
                 loggedMsg = " (Stack trace has been written to the log.)";
             }
-            if (e instanceof NumberFormatException) {
-                throw new RuntimeException("Error compiling query: " + e.toString() + loggedMsg, e);
+            if (e.getMessage() != null) {
+                throw new RuntimeException("SQL error while compiling query: " + e.getMessage() + loggedMsg, e);
             }
-            throw new RuntimeException("SQL error while compiling query: " + e.getMessage() + loggedMsg, e);
+            throw new RuntimeException("Error compiling query: " + e.toString() + loggedMsg, e);
```
yzhang1991: Why is this message different? Should they be both "SQL error while..." or just both "Error compiling query"?

mreading: The ticket suggested to say "SQL error while compiling query", but the original version was "Error compiling query". I only added the change to the code I added, but I can make it consistent across. Which one do you prefer?

PR #5375, opened 2018-06-07
ENG-13941-warnDR: Issue warning if community user attempts DR, no longer...

mreading:
...shuts down the database.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2457,6 +2457,10 @@ boolean readDeploymentAndCreateStarterCatalogContext(VoltDB.Configuration config
                             "in the community edition of VoltDB.");
                     shutdownDeployment = true;
                 }
+                if (deployment.getDr() != null) {
```
manjujames: I think this should also check that role is not NONE. So it should be:
if (deployment.getDr() != null && deployment.getDr().getRole() != DrRoleType.NONE)

PR #5378, opened 2018-06-08
ENG-6163 Add DESCRIBE tablename to SQLCMD

benjaminballard:


benjaminballard:
This should be ready to merge. I added a test in sqlcmdtest that creates a table with all the various data types. The desc table output automatically adjusts the width if the table has longer column names. If you try the command on a table name that does not exist, you get a simple "Table does not exist" response.

PR #5381, opened 2018-06-08
Eng 13678 clean

GuaGuaDean:
Changed the logic in ssl setup so server can start up without configuring truststore property.
Fixed default truststore path

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -2835,6 +2835,8 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
     //TODO: Is there a better place for this ssl setup work and constant defns
     private static final String DEFAULT_KEYSTORE_RESOURCE = "keystore";
     private static final String DEFAULT_KEYSTORE_PASSWD = "password";
+    private static final String DEFAULT_TRUSTSTORE_RESOURCE = System.getProperty("java.home") + File.separator + "lib" + File.separator + "security" + File.separator + "cacerts";
+    private static final String DEFAULT_TRUSTSTORE_PASSWD = "changeit";
     private void setupSSL(ReadDeploymentResults readDepl) {
         SslType sslType = readDepl.deployment.getSsl();
         m_config.m_sslEnable = m_config.m_sslEnable || (sslType != null && sslType.isEnabled());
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2835,6 +2835,8 @@ void setBare(boolean flag) {
     //TODO: Is there a better place for this ssl setup work and constant defns
     private static final String DEFAULT_KEYSTORE_RESOURCE = "keystore";
     private static final String DEFAULT_KEYSTORE_PASSWD = "password";
+    private static final String DEFAULT_TRUSTSTORE_RESOURCE = System.getProperty("java.home") + File.separator + "lib" + File.separator + "security" + File.separator + "cacerts";
+    private static final String DEFAULT_TRUSTSTORE_PASSWD = "changeit";
     private void setupSSL(ReadDeploymentResults readDepl) {
         SslType sslType = readDepl.deployment.getSsl();
         m_config.m_sslEnable = m_config.m_sslEnable || (sslType != null && sslType.isEnabled());
```
manjujames: If SSL is not enabled, we can just return from this method right here. I think the code will be more readable that way.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2892,25 +2894,27 @@ private SslContextFactory getSSLContextFactory(SslType sslType) {
...
-        }
-        if (trustStorePath == null || trustStorePath.trim().isEmpty()) {
-            throw new IllegalArgumentException("A path for the SSL truststore file was not specified.");
-        }
-        if (! new File(trustStorePath).exists()) {
-            throw new IllegalArgumentException("The specified SSL truststore file " + trustStorePath + " was not found.");
-        }
-        sslContextFactory.setTrustStorePath(trustStorePath);
+        if (m_config.m_startAction != StartAction.PROBE || trustStorePath != null) {
+            if (m_config.m_sslEnable) {
```
manjujames: Move this check to earlier and remove it from here.

src/frontend/org/voltdb/client/ClientConfig.java
```diff
@@ -506,4 +506,11 @@ public void enableSSL() {
             m_sslConfig = new SSLConfiguration.SslConfig();
         }
     }
+
+    public void setTrustStoreConfigFromDefault() {
+        String trustStore = System.getProperty("java.home") + File.separator + "lib" + File.separator + "security" + File.separator + "cacerts";
```
manjujames: Is there another place where we can define this, so that this does not need to be defined in RealVoltDB as well as here?

src/frontend/org/voltdb/RealVoltDB.java
GuaGuaDean: If SSL is not enabled, should we just skip the ssl setup process instead of read default truststore?

manjujames: Right

PR #5385, opened 2018-06-11
ENG-13654: in VoltDBManagementCenterPage.groovy's 'getTableContents',

johnTcrawford:
use '@innerHTML' rather than 'text()', because the latter removes
leading and trailing whitespace. In sqlQueries.txt, use '\t' to pass a
tab character to JSON, but not '\\t', since the VMC no longer interprets
'\t' as a tab character, but does work with a literal tab character; and
uncomment the 2 '...EmbeddedWhitespace' tests, since they now work, with
the above changes.


johnTcrawford:
Thanks, Chris. I'm unaware of anyone with groovy or web programming expertise, aside from me.

PR #5390, opened 2018-06-14
ENG 12194

Yusarin:


src/frontend/org/voltdb/utils/CSVLoader.java
```diff
@@ -155,18 +154,16 @@ public void run() {
...
 
-                } catch (FileNotFoundException e) {
-                    m_log.error("CSV report directory '" + config.reportdir
-                            + "' does not exist.");
                 } catch (Exception x) {
-                    m_log.error(x.getMessage());
+                    try {
+                        out_logfile.write(x.getMessage());
+                    } catch (IOException e) {
+                        e.printStackTrace();
```
manjujames: We want to print the original message to std error, not the exception that happened trying to write it to log file.

manjujames: Actually, thinking about this again, previously this error used to go to volt.log and console. So you can ignore this exception (do nothing if this IOexception happens) and add a System.stderr.println  for x.getMessage.

src/frontend/org/voltdb/utils/CSVLoader.java
```diff
@@ -543,28 +540,28 @@ public static void main(String[] args) throws IOException,
             try {
                listReader.close();
             } catch (Exception ex) {
-                m_log.error("Error closing reader: " + ex);
+               System.err.println("Error closing reader: " + ex);
```
manjujames: I think we should ignore these errors. It is completely OK if we cannot close the reader because we have already finished loading.  Do nothing in this catch block and just add a comment saying why we are doing nothing.

src/frontend/org/voltdb/utils/CSVLoader.java
```diff
@@ -543,28 +540,28 @@ public static void main(String[] args) throws IOException,
...
             } finally {
-                m_log.debug("Rows Queued by Reader: " + rowsQueued);
+               System.out.println("Rows Queued by Reader: " + rowsQueued);
```
manjujames: This used to be a debug, we can take this out completely.

src/frontend/org/voltdb/utils/CSVLoader.java
```diff
@@ -543,28 +540,28 @@ public static void main(String[] args) throws IOException,
...
-                           + config.maxerrors);
+               System.out.println("The number of failed rows exceeds the configured maximum failed rows: "
+                                  + config.maxerrors);
             }
 
-            m_log.debug("Parsing CSV file took " + readerTime + " milliseconds.");
-            m_log.debug("Inserting Data took " + ((insertTimeEnd - insertTimeStart) - readerTime) + " milliseconds.");
-            m_log.info("Read " + insertCount + " rows from file and successfully inserted "
-                       + ackCount + " rows (final)");
+            System.out.println("Parsing CSV file took " + readerTime + " milliseconds.");
```
manjujames: These used to debug output, which we should just remove completely.

src/frontend/org/voltdb/utils/JDBCLoader.java
```diff
@@ -417,15 +416,12 @@ public static void main(String[] args) throws IOException,
...
-            }
-
-            if (m_log.isDebugEnabled()) {
-                m_log.debug("Parsing CSV file took " + readerTime + " milliseconds.");
-                m_log.debug("Inserting Data took " + ((insertTimeEnd - insertTimeStart) - readerTime) + " milliseconds.");
+                System.out.println("The number of failed rows exceeds the configured maximum failed rows: "
+                        + m_config.maxerrors);
             }
-            m_log.info("Read " + insertCount + " rows from file and successfully inserted "
+            System.out.println("Parsing CSV file took " + readerTime + " milliseconds.");
```
manjujames: We should just get rid of the old debug log stmts.

manjujames:
That looks good to me. However, if we are no longer using log4j, we should remove configuring it from the scripts that start csvloader and jdbcloader.

bin/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
manjujames: What does this change do?

Yusarin: Oh, I think this file was automatically generated when I install PyDev on eclipse and create a sample project, should be removed.

manjujames: You should probably not use 'git add' with -A option, unless you are very careful to not add anything extra. I also see an invalidrows.csv file, which should be removed as well.

PR #5391, opened 2018-06-14
ENG-13965: Update adperformance to show off TTL

mreading:
Ready to merge (?)

### DIFF
--- a/examples/adperformance/ddl.sql
+++ b/examples/adperformance/ddl.sql
@@ -42,9 +42,11 @@
 -------------- REPLICATED TABLES ------------------------------------------------
 
 CREATE TABLE inventory (
-  inventory_id           INTEGER        NOT NULL,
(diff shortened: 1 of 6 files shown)


examples/adperformance/ddl.sql
```diff
@@ -68,14 +73,21 @@ CREATE TABLE event_data (
   -- derived from creative_id:
   campaign_id            INTEGER       NOT NULL,
   advertiser_id          INTEGER       NOT NULL,
+  creative_image_url     VARCHAR(2000 BYTES)     NOT NULL,
+  creative_name          VARCHAR(256)  NOT NULL,
+  creative_description   VARCHAR(256)  NOT NULL,
```
rmorgenstein: The event_data doesn't need to have the extra VARCHAR fields in it from creative and inventory.  I think it would be better just to have the extra fields in the replicated tables: creatives and inventory.

PR #5393, opened 2018-06-15
ENG-13880:

wweiss-voltdb:
In order to migrate an SPI on a DR Consumer, the BufferReceivers associated with that Consumer must be migrated as well. The first step in the process is to shutdown the PartitionBufferReceivers pinned to that consumer partition.

While the shutdown process is straightforward, the problem is that the DRConsumerMpCoordinator may not have any transaction in process from this PartitionBufferReceiver. To enable this a new mechanism has been added to reneg (take back) a pending message from the DRConsumerMpCoordinator. A response will be provided by the DRConsumerMpCoordinator when either, the message (if any) was successfully reneged, or the transaction associated with the message (and is currently in progress), has been completed.

### DIFF
--- a/src/frontend/org/voltdb/ConsumerDRGateway.java
+++ b/src/frontend/org/voltdb/ConsumerDRGateway.java
@@ -59,6 +59,8 @@ public interface ConsumerDRGateway extends Promotable {
 
     void restart(final boolean blocking) throws InterruptedException, ExecutionException;
 
+    void shutdownPartitions(final int consumerPartitionId) throws InterruptedException, ExecutionException;
+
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/ConsumerDRGateway.java
```diff
@@ -59,6 +59,8 @@
 
     void restart(final boolean blocking) throws InterruptedException, ExecutionException;
 
+    void shutdownPartitions(final int consumerPartitionId) throws InterruptedException, ExecutionException;
```
manjujames: I don't think we need to add this to the interface because we just need to shutdown the partitions when we get the callback that Ning added.

PR #5402, opened 2018-06-19
Fix no viable snapshot error resulted by ENG-11578

yzhang1991:


src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -854,20 +854,33 @@ private SnapshotInfo checkSnapshotIsComplete(Long key, Snapshot s)
                                 .append(" because catalog CRC did not match digest.");
                 return null;
             }
-            // Make sure this is not a partial snapshot.
-            // Compare digestTableNames with all normal table names in catalog file.
-            // A normal table is one that's NOT a materialized view, nor an export table.
-            Set<String> catalogNormalTableNames = CatalogUtil.getNormalTableNamesFromInMemoryJar(jarfile);
-            if (!catalogNormalTableNames.equals(digestTableNames)) {
+            // Make sure that the snapshot we are using is not a partial snapshot.
+            // All the "normal" tables in the current catalog must be present in the snapshot digest.
```
wolffcm: Would be helpful to define "normal" here, like it was in the previous revision.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2538,53 +2579,61 @@ public static void updateUsageAnnotations(Database db,
      * Get all normal table names from a in-memory catalog jar file.
      * A normal table is one that's NOT a materialized view, nor an export table.
      * @param InMemoryJarfile a in-memory catalog jar file
+     * @param optionalTableNames  List of table names that is not required to be in the snapshot.
      * @return A set of normal table names
      */
-    public static Set<String> getNormalTableNamesFromInMemoryJar(InMemoryJarfile jarfile) {
+    public static Set<String> getNormalTableNamesFromInMemoryJar(
```
wolffcm: A little strange to return two sets, one as a return value, and the other as an output parameter.  Consider returning a Pair?

The name of this function doesn't mention snaphots, the code might be clearer if it did.

src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -2538,53 +2579,61 @@ public static void updateUsageAnnotations(Database db,
...
             tableNames.add(table.getTypeName());
         }
         return tableNames;
     }
 
     /**
-     * Get all normal tables from the catalog. A normal table is one that's NOT a materialized
-     * view, nor an export table. For the lack of a better name, I call it normal.
+     * Get all normal tables from the catalog. A normal table is one that's NOT
+     * an export table and non-snapshotted views.
```
wolffcm: A "snapshottable table" maybe?

PR #5405, opened 2018-06-20
ENG-14154 wait leader promotions completed before rejoining moves forward

zeemanhe:
When a node is down, partitions will promote their new leaders. Before new leaders are fully promoted, the failed node may join back to the cluster, which will interference the partition replica calculations.  A partition replica from the rejoining node may be prematurely added to the partition replica list.  Transaction fragments may be forwarded to the partition replica on the rejoining node which may  fail to complete the scoreboard. And the repair process may hang on the new partition leader. The pull request  will make the rejoining wait until partition leader promotions are finished.

### DIFF
--- a/src/frontend/org/voltdb/iv2/LeaderAppointer.java
+++ b/src/frontend/org/voltdb/iv2/LeaderAppointer.java
@@ -278,6 +278,7 @@ public class LeaderAppointer implements Promotable
                         partitionId);
             try {
                 m_iv2appointees.put(partitionId, masterHSId);
+                VoltZK.registerPromotingPartition(m_zk, partitionId, tmLog);
             } catch (Exception e) {
                 VoltDB.crashLocalVoltDB("Unable to appoint new master for partition " + partitionId, true, e);
             }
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/iv2/LeaderAppointer.java
```diff
@@ -260,11 +260,14 @@ private long assignLeader(int partitionId, List<Long> children)
...
             } else {
-                //node down
+                // promote new partition leader when nodes are down
                 masterHostId = newLeaderHostId;
                 tmLog.debug("[assignLeader]moving leader of partition " + m_partitionId + " to host " + newLeaderHostId + " ["
                         + CoreUtils.hsIdCollectionToString(children) + "]");
                 newLeaderHostId = -1;
+
+                // register this partition so rejoining has to wait before the new leader is fully promoted.
+                VoltZK.registerPromotingPartition(m_zk, partitionId, tmLog);
```
luyangco: What if rejoin happens before assignLeader() is called, it's still possible that rejoin process is initialized during leader promotion time.

zeemanhe: Rejoin may happen before assignLeader() is called, but the chance should be very small.  An internal or external cluster monitoring tool needs detecting node failure and issue a rejoining.  The partition callbacks  are triggered by ZooKeeper.  The ZooKeeper's callbacks should be triggered before rejoining instances are coming up under most circumstances.

PR #5408, opened 2018-06-21
Eng 8921

MarcxJin:


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -3365,6 +3374,30 @@ private ExpressionLogical XreadLikePredicateRightPart(Expression a) {
                                   this.isCheckOrTriggerCondition);
     }
 
+    /**
+     *  Scan the right-side string value, return a Startswith Expression for generating XML
+     *
+     * @param a ExpressionColumn
+     */
+    private ExpressionLogical XStartsWithPredicateRightPart(Expression a) {
```
kaikaizi: Do we handle the case that parameter is on the left? That seems a legitimate use case when right side is a column.

I suspect that we do, as it would be harder to support. If we don't, make sure to issue a proper error, and test the case.

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -0,0 +1,209 @@
...
+package org.hsqldb_voltpatches;
+
+import org.hsqldb_voltpatches.lib.StringUtil;
+import org.hsqldb_voltpatches.types.Type;
+import org.hsqldb_voltpatches.types.CharacterType;
+import org.hsqldb_voltpatches.types.BinaryData;
+import org.hsqldb_voltpatches.lib.HsqlByteArrayOutputStream;
+
+/**
+ * Reusable object for processing STARTS WITH queries.
```
kaikaizi: Where is it reused?

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -0,0 +1,209 @@
+
```
kaikaizi: copyright

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        if (nodes[LEFT].dataType == null || nodes[RIGHT].dataType == null) {
+            throw Error.error(ErrorCode.X_42567);
+        }
+
+        if (nodes[LEFT].dataType.isCharacterType()
+                && nodes[RIGHT].dataType.isCharacterType()) {
+            boolean ignoreCase =
+                nodes[LEFT].dataType.typeCode == Types.VARCHAR_IGNORECASE
```
kaikaizi: When do we ignore case for startsWith? Is it covered in the test?

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -0,0 +1,209 @@
...
+    private final static BinaryData maxByteValue =
+        new BinaryData(new byte[]{ -128 }, false);
+    private char[]   cLike;
+    private int      iLen;
+    private boolean  isIgnoreCase;
+    private boolean  isNull;
+    boolean          hasCollation;
+    boolean          isVariable      = true;
+    boolean          isBinary        = false;
```
kaikaizi: Is `isBinary` always false? I see L107 of ExpressionStartsWith.java sets it true when both left and right argument are VARBINARY, but again do we support that?

If it's always false, then you can remove code that checks/ does whatever when it is true.

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -0,0 +1,209 @@
...
+        if (isBinary) {
+            l = (int) ((BinaryData) o).length(session);
+        } else {
+            l = ((String) o).length();
+        }
+
+        return l;
+    }
+
+    private boolean compareAt(Object o, int i, int j, int jLen) {
```
kaikaizi: This method looks horrible to me (and to whoever is a sane engineer, I should think).

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        if (j != jLen) {
+            return false;
+        }
+
+        return true;
+    }
+
+    void setPattern(Session session, Object pattern) {
```
kaikaizi: I challenge @MarcxJin to explain to me what this is doing (and many other functions; but I'll just take this one). If he cannot, then this shouldn't be here.

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -0,0 +1,209 @@
...
+    boolean isEquivalentToNotNullPredicate() {
+
+        if (isVariable || isNull) {
+            return false;
+        }
+
+        return true;
+    }
+
+    // An VoltDB extension for STARTS WITH operator
```
kaikaizi: ?? This whole file is all about `starts with`. How come this particular method is an extension?

src/hsqldb19b3/org/hsqldb_voltpatches/Tokens.java
```diff
@@ -1418,6 +1421,9 @@
     static final int        SCRIPTFORMAT          = 584;
     static final int        SEMICOLON             = 585;
     static final int        SHUTDOWN              = 586;
+    // A VoltDB extension to support fast match with some ranges starting with some patterns
+    static final int        STARTS                = 1011;
+    // Eng of VoltDB extension
```
kaikaizi: typo

src/hsqldb19b3/org/hsqldb_voltpatches/Tokens.java
```diff
@@ -1418,6 +1421,9 @@
     static final int        SCRIPTFORMAT          = 584;
     static final int        SEMICOLON             = 585;
     static final int        SHUTDOWN              = 586;
+    // A VoltDB extension to support fast match with some ranges starting with some patterns
+    static final int        STARTS                = 1011;
```
kaikaizi: Why 1011? I see `TTL = 1003`, `APPROX_COUNT_DISTINCT = 1304` and `STREAM = 1305`, etc. Anyway, it should follow the line before 1010, or be the next number of previous line.

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,270 @@
...
+    /** Procedures used by this suite */
+    static final Class<?>[] PROCEDURES = { Insert.class };
+
+    static final int ROWS = 10;
+
+    public void testStartsWith() throws IOException, ProcCallException
+    {
+        Client client = getClient();
+        StartsWithSuite tests = new StartsWithSuite(client);
+//        TestStartsWithQueries.StartsWithSuite tests = new TestStartsWithQueries.StartsWithSuite(client);
```
kaikaizi: clean up.

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,270 @@
...
+        final String schema =
+                "create table STRINGS (" +
+                "ID int default 0 not null, " +
+                "VAL varchar(32) default null," +
+                "PAT varchar(32) default null," +
+                "PRIMARY KEY(ID));";
+
+        try {
+            project.addLiteralSchema(schema);
+        } catch (IOException e) {
```
kaikaizi: you can just have it propagate up.

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,270 @@
...
+            e.printStackTrace();
+        }
+        project.addPartitionInfo("STRINGS", "ID");
+        project.addStmtProcedure("Insert", "insert into strings values (?, ?, ?);");
+        project.addStmtProcedure("SelectStartsWith", "select * from strings where val starts with ?;");
+        project.addStmtProcedure("NotStartsWith", "select * from strings where val not starts with ?;");
+
+        config = new LocalCluster("sqllike-onesite.jar", 1, 1, 0, BackendTarget.NATIVE_EE_JNI);
```
kaikaizi: What is this?

tests/sqlcmd/baselines/sysproc/explaincatalog.outbaseline
```diff
@@ -103,7 +103,7 @@ set $PREV elasticduration 50
 set $PREV elasticthroughput 2
 set $PREV querytimeout 10000
 add /clusters#cluster logconfig log
-set /clusters#cluster/logconfig#log enabled true
+set /clusters#cluster/logconfig#log enabled false
```
kaikaizi: You didn't mean to change this?

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,270 @@
+/* This file is part of VoltDB.
```
kaikaizi: The structure of this test looks very esoteric. Turns out that it's largely based on org.voltdb.TestLikeQueries.java, which was created in 2012.

Now there should be better/easier way to test this in regressionsuite. Copy-pasting-edit arcane test code makes it harder to understand/maintain.

I suggest rewrite the test

src/ee/common/NValue.hpp
```diff
@@ -4012,6 +4013,64 @@ inline NValue NValue::like(const NValue& rhs) const {
...
+            return getTrue();
+        } else {
+            return getFalse();
+        }
+    }
+
+    assert(valueChars);
+    assert(patternChars);
+
+    UTF8Iterator m_value(valueChars, valueChars + valueUTF8Length);
```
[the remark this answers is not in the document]
MarcxJin: It deals with the problem that we may read beyond the end of sequence, and the function 'uft8::unchecked::next' it uses can extract the right bytes of one utf-8 char. Anyways, this is a highly encapsulated class.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -3365,6 +3374,30 @@ private ExpressionLogical XreadLikePredicateRightPart(Expression a) {
...
+
+        readThis(Tokens.WITH);
+
+        if (token.tokenType == Tokens.QUESTION) {    // handle user parameter
+            Expression left = XreadRowValuePredicand();
+            if (a.isParam() && left.isParam()) {
```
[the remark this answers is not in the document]
MarcxJin: So I tried to create procedure as 'select * from t where ? like ?' or 'select * from t where ? between ? and ?' and they are both failed. If this is the case, we need to file some tickets for both these two expressions.

src/ee/common/NValue.hpp
```diff
@@ -4012,6 +4013,64 @@ inline NValue NValue::like(const NValue& rhs) const {
...
+ * The LHS (this) should always be the string being compared
+ * and the RHS should always be a plain string.
+ *
+ * Null check should have been handled in comparisonexpression.h already.
+ */
+inline NValue NValue::startsWith(const NValue& rhs) const {
+    /*
+     * Validate that all params are VARCHAR
+     */
+    const ValueType mType = getValueType();
```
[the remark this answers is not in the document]
MarcxJin: This is only for Java SP. If you define 'int' as the parameter type in Java class, 'create procedure' can be successful, but then if you do 'exec xxx 1' it will throw errors. While for adhoc procedure it automatically turns 'exec xxx 1' to " exec xxx'1' " since the type requirements. This is consistent with LIKE operation.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,237 @@
...
+            synchronized (startsWithObject) {
+               startsWithObject.setPattern(session, rightValue);
+               return startsWithObject.compare(session, leftValue);
+            }
+        }
+
+        return startsWithObject.compare(session, leftValue);
+    }
+
+    @Override
```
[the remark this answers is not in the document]
MarcxJin: It actually doesn't follow in that way. Both STARTS WITH and LIKE expression have there own resolveTypes function. And LIKE did not deal with Optypes.ROW as well. But by your example, 'col STARTS WITH col' actually won't arrive here. It is handled in another way.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,237 @@
...
+            startsWithObject.isBinary = true;
+        } else if (false == (nodes[LEFT].dataType.isBooleanType()
+                              && nodes[RIGHT].dataType.isBooleanType())
+                              && dataType.isBooleanType()) {
+            // If both argument nodes are boolean we have resolved
+            // this before.  So, this is ok.  Otherwise, this is not
+            // properly typed.          throw Error.error(ErrorCode.X_42565);
+        }
+
+        if (startsWithObject != null) {
```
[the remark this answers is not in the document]
MarcxJin: In most case it won't be null, but just want to make sure startsWithObject.dataType does not throw a NullPointer Exception.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,237 @@
...
+        return sb.toString();
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (!super.equals(other)) {
+            return false;
+        }
+
+        if (other instanceof ExpressionStartsWith) {
```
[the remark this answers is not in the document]
MarcxJin: Other requirements are checked by 'equals' function in its super class.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
MarcxJin: Yes, we do.  You can put string on the left and a column on the right. I will add a new junit test.

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
MarcxJin: It is to read what users have been inputed (BTW check if it is null) and store string length and itself as fields in the class.

src/hsqldb19b3/org/hsqldb_voltpatches/Tokens.java
MarcxJin: It should leave there to keep alphabetic order. 1011 is simply numbered after 1010.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,245 @@
...
+        super(OpTypes.STARTS_WITH);
+
+        nodes               = new Expression[BINARY];
+        nodes[LEFT]         = left;
+        nodes[RIGHT]        = right;
+        startsWithObject    = new StartsWith();
+        this.noOptimisation = noOptimisation;
+    }
+
+    private ExpressionStartsWith(ExpressionStartsWith other) {
```
yzhang1991: Nobody is using this constructor, remove it.

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
yzhang1991: This comment is FYI, @kaikaizi 
I am OK with not making changes here because I care less about the HSQL implementation.
I would also argue that in this class `isBinary` is not useful in our system. Because we do not allow STARTS WITH or LIKE to be applied on non-VARCHAR columns.
This implementation here is for HSQL. I think they are used when we run SQLCoverage against HSQL and will eventually be deprecated as we are moving the backend target to PostgreSQL.

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * System tests for basic STARTS WITH functionality
+ */
+public class TestSqlStartsWithRegressionSuite extends RegressionSuite {
+    /** Procedures used by this suite */
+    static final Class<?>[] PROCEDURES = { Insert.class };
+
+    static final int ROWS = 7;
+
+    static final String[] testString = {"aaa", "abc", "AbC", "zzz", "", "a", "âxxx", "aaaaaaa",
```
yzhang1991: I need a comment above this. It seems to me that the `testString` is the test pattern string.
Meanwhile, the name of this variable should be in plural form, i.e. `testStrings`

src/frontend/org/voltdb/types/ExpressionType.java
```diff
@@ -89,6 +89,7 @@
         // greater than equal operator between left and right
     COMPARE_LIKE                 (ComparisonExpression.class, 16, "LIKE", true),
         // LIKE operator (left LIKE right). both children must be string.
+    COMPARE_STARTSWITH           (ComparisonExpression.class, 150, "STARTS WITH", true),
```
yzhang1991: Add comment

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDQL.java
```diff
@@ -3365,6 +3374,30 @@ private ExpressionLogical XreadLikePredicateRightPart(Expression a) {
...
+
+        if (token.tokenType == Tokens.QUESTION) {    // handle user parameter case
+            Expression right = XreadRowValuePredicand();
+            if (left.isParam() && right.isParam()) {  // again make sure the left side is valid
+                throw Error.error(ErrorCode.X_42567);
+            }
+            Expression l = new ExpressionLogical(OpTypes.GREATER_EQUAL, left, right);
+            Expression r = new ExpressionLogical(OpTypes.SMALLER_EQUAL, left,
+                    new ExpressionArithmetic(OpTypes.CONCAT, right, new ExpressionValue("\uffff", Type.SQL_CHAR)));
+            return new ExpressionLogical(OpTypes.AND, l, r);
```
yzhang1991: This part needs a comment explaining how the rewriting is done.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,237 @@
...
+               startsWithObject.setPattern(session, rightValue);
+               return startsWithObject.compare(session, leftValue);
+            }
+        }
+
+        return startsWithObject.compare(session, leftValue);
+    }
+
+    @Override
+    public void resolveTypes(Session session, Expression parent) {
```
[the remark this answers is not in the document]
yzhang1991: **Cover it in your junit.**

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
yzhang1991: Rename the jar file name.
You should replace the BackendTarget of this case with `HSQLDB_BACKEND`
You add the implementation of `STARTS WITH` to HSQL, I don't want to rely on the SQLCoverage to test if it works.

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,208 @@
...
+        project.addPartitionInfo("STRINGS", "ID");
+        project.addStmtProcedure("Insert", "insert into strings values (?, ?, ?);");
+        project.addStmtProcedure("SelectStartsWith", "select * from strings where val starts with ?;");
+        project.addStmtProcedure("NotStartsWith", "select * from strings where val not starts with ?;");
+
+        config = new LocalCluster("sqllike-onesite.jar", 1, 1, 0, BackendTarget.NATIVE_EE_JNI);
+        if (!config.compile(project)) fail();
+        builder.addServerConfig(config);
+
+        config = new LocalCluster("sqllike-twosites.jar", 2, 1, 0, BackendTarget.NATIVE_EE_JNI);
```
yzhang1991: Rename

tests/frontend/org/voltdb/regressionsuites/TestSqlStartsWithRegressionSuite.java
```diff
@@ -0,0 +1,208 @@
...
+        if (!config.compile(project)) fail();
+        builder.addServerConfig(config);
+
+        config = new LocalCluster("sqllike-twosites.jar", 2, 3, 1, BackendTarget.NATIVE_EE_JNI);
```
yzhang1991: Rename

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -0,0 +1,237 @@
...
+                nodes[LEFT].dataType.typeCode == Types.VARCHAR_IGNORECASE
+                || nodes[RIGHT].dataType.typeCode == Types.VARCHAR_IGNORECASE;
+            startsWithObject.setIgnoreCase(ignoreCase);
+        } else if (nodes[LEFT].dataType.isBinaryType()
+                   && nodes[RIGHT].dataType.isBinaryType()) {
+            startsWithObject.isBinary = true;
+        } else if (false == (nodes[LEFT].dataType.isBooleanType()
+                              && nodes[RIGHT].dataType.isBooleanType())
+                              && dataType.isBooleanType()) {
+            // If both argument nodes are boolean we have resolved
```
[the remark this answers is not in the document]
yzhang1991: Just like `isBinary`, this is not needed in our system.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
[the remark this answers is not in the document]
MarcxJin: Look at testStartsWithWrongType method in the test case

PR #5409, opened 2018-06-21
Fix ENG-14164

yzhang1991:


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -2287,18 +2287,20 @@ private AbstractPlanNode handleMVBasedMultiPartQuery(
         AbstractPlanNode sendNodeChild = sendNode.getChild(0);
 
         HashAggregatePlanNode reAggNodeForReplace = null;
-        if (m_parsedSelect.getScanCount() > 1 && !edgeCaseOuterJoin) {
+        // For cases that joining an implicitly partitioned view
```
wolffcm: I don't really understand what this method does... I've just spent a fair amount of time staring this code but it's not any clearer to me.

Seems like when we call processScanNodeWithReAggNode below, that we are placing the reagg node into the plan just above the scan of the MV.  But somehow it's below the send node?  I thought we only re-aggregated above the receive node (on the coordinator fragment)?

I guess that we are using `instanceof AbstractJoinPlanNode` to see if we are doing a join, instead of counting the table scans like before.  Why is it better to do it that way?  In general, I don't like to see assumptions about the plan topology---are we making the assumption that if the query is doing a join, then the child of the send node must be a join node?

PR #5414, opened 2018-06-25
Eng 14165 sqlcov internal err

johnTcrawford:
Modified SqlCoverage to fail when "unexpected internal error occurred" (and certain other errors) is returned by VoltDB. Also, labeled old & deprecated config files as such, mostly by adding 'old_' prefix to file names or, in one case, via a comment.


kaikaizi:
Looks good to me.

PR #5419, opened 2018-06-26
ENG-14186 update DDL syntax for TTL

zeemanhe:
also do not execute nibble deletes when cluster is paused to allow buffers to be drained for export and dr

### DIFF
--- a/src/frontend/org/voltdb/TTLManager.java
+++ b/src/frontend/org/voltdb/TTLManager.java
@@ -89,9 +89,9 @@ public class TTLManager extends StatsSource{
         @Override
         public void run() {
 
-            //do not run TTL when cluster is paused to allow proper draini of stream and dr buffer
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/TTLManager.java
```diff
@@ -88,11 +88,17 @@ public TTLTask(String table, TimeToLive timeToLive, TTLStats ttlStats) {
         }
         @Override
         public void run() {
-            ClientInterface cl = VoltDB.instance().getClientInterface();
+
+            //do not run TTL when cluster is paused to allow proper draini of stream and dr buffer
+            final VoltDBInterface voltdb = VoltDB.instance();
+            if (voltdb.getMode() == OperationMode.PAUSED) {
```
JadenW: It may not happen, but we probably don't want issue TTL on case other than Operation.Running ? (i.e. Initilize and shuttingDown and Paused )

PR #5427, opened 2018-06-27
ENG-6163 print columns in the correct order

benjaminballard:


src/frontend/org/voltdb/utils/SQLCommand.java
```diff
@@ -424,84 +426,112 @@ else if (subcommand.equals("config") || subcommand.equals("configuration")) {
...
                     break;
                 }
             }
-
             if (tableName.equals("")) {
                 System.err.println("Table does not exist");
                 return true;
             }
 
+            // Class to temporarily store and sort column attributes
```
mreading: I find this comment rather misleading; I have more experience with C# than Java, so I see the purpose of this class as a [struct](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/struct) - you are using it to temporarily store data, but not to sort column attributes (see your comment, you are using the sorted set on line 466 instead). Since you do not use compareTo nor toString, I would remove them from this class, and does it still need to implement Comparable? Otherwise, this is lovely and I like how you only iterate through the Columns table once now, it is much more efficient.

mreading: My apologies, the implements Comparable and compareTo is being used by the SortedSet

benjaminballard: I am using it basically like a struct, except implementing Comparable makes it implicitly sortable, so when the items are added to the TreeSet, they are inserted in to the tree in order. No extra sorting command needs to be used, the iterator will give the items in sorted order. The alternative was to use a simpler struct class, but then the items would have to be put into sortable Map using the position attribute as a key, which seemed more complicated than this.

Thanks for the feedback.

PR #5431, opened 2018-07-02
Eng 13833 - Backport HTTP session capability to v7.7.2

GuaGuaDean:


manjujames:
Are the protobuf jar changes part of this fix? Do we need that change?

src/frontend/org/voltdb/utils/ApiRequestServlet.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.servlet.http.HttpServletResponse;
+import org.eclipse.jetty.server.Request;
+
+import org.voltdb.VoltDB;
+import org.voltdb.client.ClientResponse;
+
+/**
+ *
+ * @author akhanzode
+=======
```
manjujames: We should get rid of these merge artifacts.

manjujames:
What are the last 4 jar changes for (zk jars, commons and httpclient jar updates)?

src/frontend/org/voltdb/utils/ApiRequestServlet.java
GuaGuaDean: Sure. 
For the protobuf jar, Anish made this change in his pro commit's classpath file. https://github.com/VoltDB/pro/commit/dd259789bdce4585f64bf39a4b971b96c066304f
ZK, commons.io and httpclient used for test only and were introduced by earlier commits

PR #5438, opened 2018-07-03
ENG-13951: Adding callbacks for Export system on Host Failure

JadenW:


src/frontend/org/voltdb/iv2/SpTerm.java
```diff
@@ -86,11 +87,14 @@ public void run(List<String> children)
                 }
                 m_mailbox.updateReplicas(replicas, null);
                 m_replicasUpdatedRequired = false;
+                // TODO ? JOIN case for export
```
luyangco: Join likes rejoin, new node promotes partition leader which in turn elects export master. In my opinion I don't think export needs to do something here.

PR #5440, opened 2018-07-03
Eng 13881

manjujames:
On SPI migration, the DR partition buffer receiver(s) on old leader must be shutdown and DR partition buffer receiver must be started on the new leader. We must also make sure that the new leader buffer receiver is started only after the old has shutdown.

On SPI migration, DR receives 3 callbacks:

leader migration started - This callback happens only on the old leader. This will add a ZK blocker per producer-cluster and consumer partition id
leader migration failed - This callback happens only on the old leader. This will remove the ZK blocker that was added by 'started' callback
leader migration completed: This will be received on both old leader and new leader.
old leader will shutdown the receivers corresponding to this partition and once the receivers are shutdown, deletes the ZK node (description on how shutdown happens is below)
new leader will initiate repairCompleted, which will wait for the ZK blocker to be gone before it starts receivers on the new leader for the partitions
Shutting down partition receivers:

Pending and rate limited queues for the receivers are cleared (anything that hasn't been sent to the client interface will be removed).
send a message to consumer mp coordinator to reneg any MP txns for the producer partitions.
if there are any mp txns for the partitions waiting to get data from other partitions, they will be removed
if there are any mp txns for the partitions that were already sent down to the client interface, we will wait to receive a responses for those
once reneg work is complete, receivers are shutdown so that they won't accept anymore binary log data
once all the receivers are done, we delete the ZK blocker node

### DIFF
--- a/src/frontend/org/voltdb/ClientInterface.java
+++ b/src/frontend/org/voltdb/ClientInterface.java
@@ -2265,6 +2265,11 @@ public class ClientInterface implements SnapshotDaemon.DaemonInitiator {
 
             notifyPartitionMigrationStatus(partitionId, targetHSId, false);
 
+            Integer.parseInt(System.getProperty("MIGRATE_PARTITION_LEADER_DELAY", "30"));
+            if (Boolean.getBoolean("TEST_MIGRATION_FAILURE")) {
+                Thread.sleep(100);
+                throw new IOException("failure simulation");
+            }
             synchronized (m_executeTaskAdpater) {
                 createTransaction(m_executeTaskAdpater.connectionId(),
                         spi,
--- a/src/frontend/org/voltdb/ClientInterface.java
+++ b/src/frontend/org/voltdb/ClientInterface.java
@@ -2265,7 +2265,6 @@ public class ClientInterface implements SnapshotDaemon.DaemonInitiator {
 
             notifyPartitionMigrationStatus(partitionId, targetHSId, false);
 
-            Integer.parseInt(System.getProperty("MIGRATE_PARTITION_LEADER_DELAY", "30"));
             if (Boolean.getBoolean("TEST_MIGRATION_FAILURE")) {
                 Thread.sleep(100);
                 throw new IOException("failure simulation");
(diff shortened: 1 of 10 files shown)


src/frontend/org/voltdb/ClientInterface.java
```diff
@@ -2282,29 +2290,36 @@ void startMigratePartitionLeader() {
                 //not necessary a failure.
                 tmLog.warn(String.format("Fail to move the leader of partition %d to host %d. %s",
                         partitionId, targetHostId, resp.getStatusString()));
+                notifyPartitionMigrationStatus(partitionId, targetHSId, true);
             }
         } catch (IOException | InterruptedException e) {
             tmLog.warn(String.format("errors in leader change for partition %d: %s", partitionId, e.getMessage()));
+            notifyPartitionMigrationStatus(partitionId,
+                    m_cartographer.getHSIDForPartitionHost(targetHostId, partitionId),
```
JadenW: nitpick, use targetHsid instead.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2467,6 +2468,7 @@ boolean readDeploymentAndCreateStarterCatalogContext(VoltDB.Configuration config
              * Trick here is to print out all applicable problems and then stop, rather than stopping
              * after the first one is found.
              */
+            deployment.getDr();
```
JadenW: never used here?

src/frontend/org/voltdb/ClientInterface.java
manjujames: You are right, it should targetHSId.

src/frontend/org/voltdb/RealVoltDB.java
manjujames: Don't know how that got in there. I will take it out.

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2221,6 +2221,7 @@ private void startMigratePartitionLeaderTask() {
             int hostId = it.next();
             final int currentMasters = m_cartographer.getMasterCount(hostId);
             if (currentMasters > minimalNumberOfLeaders) {
+                hostLog.debug("Host " + hostId + " has more than " + minimalNumberOfLeaders + ". Sending migrate partition message");
```
luyangco: Wrap debug message with isDebugEnabled() check.

luyangco:
Looks good to me.

src/frontend/org/voltdb/RealVoltDB.java
manjujames: My mistake. Will take care of all of those.

PR #5444, opened 2018-07-05
ENG-14213 do not replay logged tasks before snapshotting in join path

zeemanhe:
When tasks in task log before stream snapshotting are replayed  in join path, EE could hang since these tasks are not synchronized via score board. Filter out these tasks during replay.


luyangco:
It looks good to me. After Scoreboard was introduced this mechanism was removed, but for join path it is still essential to filter out pre-snapshot messages.
Thanks Zeeman.

PR #5449, opened 2018-07-06
Eng 10675 - a SQL function to convert a timestamp to a String in a given timezone

Chaozz:
Implement **FORMAT_TIMESTAMP(timestamp, timezone_or_time_offset)**

+ We assume the timestamp stored in Volt is in UTC

+ *timezone_or_time_offset* must be a valid timezone from [Postgres timezones](https://www.postgresql.org/docs/7.2/static/timezones.html) or in a **[+-]HH:MM** format.

+ The time zone info in my [code](https://github.com/VoltDB/voltdb/blob/73535877cda1967becf6acd07358f132e7d3c0b4/src/ee/expressions/datefunctions.h#L33) is get from https://www.postgresql.org/docs/7.2/static/timezones.html. To save time, I have a script [here](https://gist.github.com/Chaozz/1da6cead158e1b9c8bb5fad48aa9ba9c) to auto convert the html data to c++ map.


src/ee/expressions/datefunctions.h
```diff
@@ -1059,6 +1085,56 @@ template<> inline NValue NValue::call<FUNC_VOLT_DATEADD_MICROSECOND>(const std::
     }
 }
 
+template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_TIMESTAMP>(const std::vector<NValue>& arguments){
+    assert (arguments.size() == 2);
+
+    const NValue& time_stamp = arguments[0];
+    if (time_stamp.isNull()) {
```
BillWhite: Hoorah!!  Testing for null before testing for the type is always the right thing to do.

Sometimes we have null parameter values whose type is VALUE_TYPE_NULL.  If you check that the type is VALUE_TYPE_TIMESTAMP before you test for null we get spurious errors.  A null is a null, after all, and it only has a stylized type.

src/ee/expressions/datefunctions.h
```diff
@@ -19,6 +19,9 @@
 #define SRC_EE_EXPRESSIONS_DATEFUNCTIONS_H
 
 #include <ctime>
+#include <unordered_map>
+#include <regex>
```
BillWhite: I think c++ <regex> is c++-11.  Has this been run on Centos6?  The Centos6 g++ compiler does not actually have c++11.  It has something almost but not completely c++11.  Sometimes Jenkins misses C6, so you will need to build this on C6 if it hasn't already to test this.  I would run the regression suite test on C6 as well, just to be sure the C6 regexp works well enough.

src/ee/expressions/datefunctions.h
```diff
@@ -1059,6 +1085,56 @@ template<> inline NValue NValue::call<FUNC_VOLT_DATEADD_MICROSECOND>(const std::
...
+    std::string time_offset_str(time_offset_buffer, length);
+    auto search = TIMEZONE_OFFSET_FROM_UTC.find(time_offset_str);
+    if(search!=TIMEZONE_OFFSET_FROM_UTC.end()){
+        time_offset_str = search->second;
+    }
+    if(!std::regex_match(time_offset_str, offset_re)){
+        throw SQLException(SQLException::data_exception_invalid_parameter, "time offset must use valid timezone name or meet the [+-]HH:MM format");
+    }
+
+    int64_t micro_seconds_offset = 1000000 * (stoll(time_offset_str.substr(1, 2)) * 3600 +
```
BillWhite: What if the string starts with some spaces or some other white space?  Does this affect this?

I don't know the answer but time_offset_str.at(0) might be a space, not '-'.

src/ee/expressions/datefunctions.h
```diff
@@ -1059,6 +1085,56 @@ template<> inline NValue NValue::call<FUNC_VOLT_DATEADD_MICROSECOND>(const std::
     }
 }
 
+template<> inline NValue NValue::call<FUNC_VOLT_FORMAT_TIMESTAMP>(const std::vector<NValue>& arguments){
```
BillWhite: It looks like you always use exactly two arguments.  Should this be a binary function and not a general function?  I would guess so, but maybe there's something I am missing.

src/ee/expressions/functionexpression.cpp
```diff
@@ -627,6 +627,9 @@ ExpressionUtil::functionFactory(int functionId, const std::vector<AbstractExpres
         case FUNC_VOLT_DWITHIN_POLYGON_POINT:
             ret = new GeneralFunctionExpression<FUNC_VOLT_DWITHIN_POLYGON_POINT>(*arguments);
             break;
+        case FUNC_VOLT_FORMAT_TIMESTAMP:
+            ret = new GeneralFunctionExpression<FUNC_VOLT_FORMAT_TIMESTAMP>(*arguments);
```
BillWhite: Again, it seems like BinaryFunctionExpression might be more natural here.

Chaozz: I agree with you. But currently we only have **UnaryFunctionExpression** and **GeneralFunctionExpression**. It seemthe way we use function with more than one arguments is to use GeneralFunctionExpression with *assert(argument_size)*.

I can implement a BinaryFunctionExpression if necessary, but it may out of the scope of this ticket.

src/ee/expressions/datefunctions.h
Chaozz: Nice catch! Our Jenkins server use GCC 4.4.7 which don't support std::regex. I have to work around.

src/ee/expressions/datefunctions.h
Chaozz: now I trim the offset_str before processing

src/ee/expressions/datefunctions.h
```diff
@@ -1059,6 +1083,63 @@ template<> inline NValue NValue::call<FUNC_VOLT_DATEADD_MICROSECOND>(const std::
...
+    if(search!=TIMEZONE_OFFSET_FROM_UTC.end()){
+        time_offset_str = search->second;
+    }
+    // equivalent to [\+\-][0-1][0-9]:[0-5][0-9], since <regex> is not support by our c++ compiler
+    if (!(time_offset_str.length() == 6 && (time_offset_str.at(0) == '-' || time_offset_str.at(0) == '+') &&
+          (time_offset_str.at(1) == '0' || time_offset_str.at(1) == '1') && isdigit(time_offset_str.at(2)) &&
+          time_offset_str.at(3) == ':' && isdigit(time_offset_str.at(4)) && stoi(time_offset_str.substr(4, 1)) < 6 &&
+          isdigit(time_offset_str.at(5)))) {
+        throw SQLException(SQLException::data_exception_invalid_parameter,
+                           "time offset must use valid timezone name or meet the [+-]HH:MM format");
```
yzhang1991: This error message needs to be checked with Andrew. As I pointed out in the last comment, there should be a description of valid ranges for offset string. For example, "-20:00" satisfies the  `[+-]HH:MM` format, but will fail this check. Getting this error message in such case is confusing to users.

src/ee/expressions/datefunctions.h
```diff
@@ -1059,6 +1083,63 @@ template<> inline NValue NValue::call<FUNC_VOLT_DATEADD_MICROSECOND>(const std::
...
+
+    int64_t total_micro_seconds = time_stamp.getTimestamp() + micro_seconds_offset;
+    int64_t total_seconds = total_micro_seconds / 1000000;
+    int64_t micro_seconds_tail = total_micro_seconds % 1000000;
+
+    std::time_t t = total_seconds;
+    struct std::tm * time_info = std::gmtime (&t);
+    char time_buffer [60];
+    std::strftime (time_buffer,60,"%F %T.",time_info);
+    snprintf(time_buffer+strlen(time_buffer), 10, "%06ld", long(micro_seconds_tail));
```
yzhang1991: We already have a way to get the string of a timestamp. Check `NValue::streamTimestamp` and do some refactors as needed.

src/ee/expressions/functionexpression.cpp
yzhang1991: That I don't think is necessary.

src/ee/expressions/functionexpression.h
```diff
@@ -293,6 +293,7 @@ static const int FUNC_VOLT_MIN_VALID_TIMESTAMP         = 21021;     // minimum v
 static const int FUNC_VOLT_MAX_VALID_TIMESTAMP         = 21022;     // maximum valid timestamp
 static const int FUNC_VOLT_IS_VALID_TIMESTAMP          = 21023;     // is a timestamp valid
 static const int FUNC_VOLT_MAKE_VALID_POLYGON          = 21024;     // Make a polygon valid if necessary.
+static const int FUNC_VOLT_FORMAT_TIMESTAMP            = 21025;
```
yzhang1991: Add comment please

src/hsqldb19b3/org/hsqldb_voltpatches/FunctionForVoltDB.java
```diff
@@ -195,6 +195,7 @@ private FunctionDescriptor(String name, Type type, int id, int typeParameter, Ty
         static final int FUNC_VOLT_IS_VALID_TIMESTAMP           = 21023;    // Is a timestamp value in range?
         static final int FUNC_VOLT_MAKE_VALID_POLYGON           = 21024;    // Make an invalid polygon valid by reversing rings.
                                                                             // Note: This will only correct orientation errors.
+        static final int FUNC_VOLT_FORMAT_TIMESTAMP             = 21025;
```
yzhang1991: Add comment please

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -3345,4 +3347,59 @@ public void testTimestampValidityFunctions() throws Exception {
...
+
+        // null offset means offset == 0
+        cr = client.callProcedure("FORMAT_TIMESTAMP", null, 1);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        result = cr.getResults()[0];
+        assertEquals(1, result.getRowCount());
+        assertTrue(result.advanceRow());
+        assertEquals(result.getString(0), "2013-07-18 02:00:00.123457");
+
+        String[] invalid_offsets = {"I_AM_INVALID", "10:00", "-1:00"};
```
yzhang1991: Need to test out-of-range ones like "+20:00"

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -3345,4 +3347,59 @@ public void testTimestampValidityFunctions() throws Exception {
...
+        ClientResponse cr;
+        VoltTable result;
+
+        cr = client.callProcedure("P2.insert", 1, "2013-07-18 02:00:00.123457");
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure("P2.insert", 2, null);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+
+        // null timestamp will return null
+        cr = client.callProcedure("FORMAT_TIMESTAMP", "UTC", 2);
```
yzhang1991: Use `assertContentOfTable()`

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -3345,4 +3347,59 @@ public void testTimestampValidityFunctions() throws Exception {
...
+        result = cr.getResults()[0];
+        assertEquals(1, result.getRowCount());
+        assertTrue(result.advanceRow());
+        assertNull(result.getString(0));
+
+        // null offset means offset == 0
+        cr = client.callProcedure("FORMAT_TIMESTAMP", null, 1);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        result = cr.getResults()[0];
+        assertEquals(1, result.getRowCount());
```
yzhang1991: Use `assertContentOfTable()`

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -3345,4 +3347,59 @@ public void testTimestampValidityFunctions() throws Exception {
...
+        assertTrue(result.advanceRow());
+        assertEquals(result.getString(0), "2013-07-18 02:00:00.123457");
+
+        String[] invalid_offsets = {"I_AM_INVALID", "10:00", "-1:00"};
+
+        for (String invalid_offset : invalid_offsets) {
+            try {
+                client.callProcedure("FORMAT_TIMESTAMP", invalid_offset, 1);
```
yzhang1991: Use `verifyProcFails()`

yzhang1991:
We switched to a Java implementation in https://github.com/VoltDB/voltdb/pull/5463

PR #5451, opened 2018-07-09
ENG-14198: SQL coverage auto filer

mreading:


tools/jenkins/sql_coverage_reporter.py
```diff
@@ -0,0 +1,93 @@
...
+            + ' : Total=' + str(stats['total_statements']) + '\n' \
+            + 'Test Failures:\n' \
+            + ' : Mismatched=' + str(stats['mismatched_statements']) + ' (' + stats['mismatched_percent'] + '%)\n' \
+            + 'Exceptions:\n' \
+            + ' : Volt Fatal=' + str(stats['total_volt_fatal_excep']) \
+            + ' : Volt Nonfatal=' + str(stats['total_volt_nonfatal_excep']) + '\n' \
+            + 'Crashes:\n' \
+            + ' : Volt=' + str(stats['total_volt_crashes']) \
+            + ' : Diff=' + str(stats['total_diff_crashes']) \
+            + ' : Comparison=' + str(stats['total_cmp_crashes'])
```
johnTcrawford: These 2 are in the wrong order: "Diff" is checking the difference between the VoltDB and Comparison database (HSQL or PostgreSQL/PostGIS) results; so the "Diff" is the last thing it does, and usually only has a problem if one or both of the previous two also had a problem. Also, again, it might be clearer to use the label "Comparison DB", rather than just "Comparison"; or to use the name of the DB, but that would require adding it to your stats.txt file (see above).

tests/sqlcoverage/sql_coverage_test.py
```diff
@@ -885,6 +885,27 @@ def usage():
...
+            'invalid_percent' : invalid_percent,
+            'total_statements' : total_statements,
+            'mismatched_statements' : mismatched_statements,
+            'mismatched_percent' : mismatched_percent,
+            'total_volt_fatal_excep' : total_volt_fatal_excep,
+            'total_volt_nonfatal_excep' : total_volt_nonfatal_excep,
+            'total_cmp_excep' : total_cmp_excep,
+            'total_volt_crashes' : total_volt_crashes,
+            'total_cmp_crashes' : total_cmp_crashes,
+            'total_diff_crashes' : total_diff_crashes
```
johnTcrawford: You might want to also include comparison_database here, i.e., the name of the database whose results are being compared to VoltDB, which will always be either 'HSQL' or 'PostgreSQL', or 'PostGIS' (which is PostgreSQL with its Geospatial extension). See below for why this might be useful.

tools/jenkins/sql_coverage_reporter.py
```diff
@@ -0,0 +1,93 @@
...
+import os
+import sys
+from jenkinsbot import JenkinsBot
+from jira import JIRA
+from sql_grammar_reporter import Issues # using symlink alias to reference file
+
+JIRA_USER = os.environ.get('jirauser', None)
+JIRA_PASS = os.environ.get('jirapass', None)
+JIRA_PROJECT = os.environ.get('jiraproject', None)
+JUNIT = os.environ.get('junit', None)
```
johnTcrawford: I presume these are the same as those used by the JUnit and sql-grammar-gen auto-filers? (I'm mostly curious, not a big deal either way, as long as it works.)

tools/jenkins/sql_coverage_reporter.py
```diff
@@ -0,0 +1,93 @@
...
+                continue
+
+            config_name = config_report.get('fullDisplayName')
+            config_data_url = config_url + 'artifact/obj/release/sqlcoverage'
+            config_stats = eval(self.read_url(config_data_url + '/stats.txt'))
+            description += self.build_description(config_name, config_data_url, config_stats)
+
+        jenkinsbot = JenkinsBot()
+        jenkinsbot.create_bug_issue(JUNIT, summary, description, 'Core', current_version, ['sqlcoverage-failure', 'automatic'],
+                                    JIRA_USER, JIRA_PASS, JIRA_PROJECT, DRY_RUN)
```
johnTcrawford: I'm not really familiar with the JenkinsBot, so I can't comment on this 'report' function specifically, but presuming that it works, it looks reasonable.

tools/jenkins/sql_coverage_reporter.py
prosegay: JC: do you have a question or comment?

tests/sqlcoverage/sql_coverage_test.py
prosegay: We will point to the sqlcoverage report that is in Jenkins.

tools/jenkins/sql_coverage_reporter.py
prosegay: what do you mean by 'wrong order'? what is the correct order?

prosegay:
JC: I don't understand most of your comments. If you want to tweak this to a particular result/format, feel free to do so.

tools/jenkins/sql_coverage_reporter.py
johnTcrawford: It was a question. It makes sense to me that this would use the same login info (username, password, etc.) as the other two auto-filers, so I guess I'm also expressing a slight preference for that, but I don't really care if different login info was used, for some reason.

tools/jenkins/sql_coverage_reporter.py
johnTcrawford: Volt, Comparison DB, Diff.

tests/sqlcoverage/sql_coverage_test.py
johnTcrawford: Yes, but I think the Jira ticket will be clearer if the description includes the name of the database, rather than 'Comparison', or even 'Comparison DB': it's clear to me what that means, but it won't be to people unfamiliar with SqlCoverage.

johnTcrawford:
Phil, I have clarified where you were confused. If Matt is unclear about my comments, he is free to ask me for help.

tools/jenkins/sql_coverage_reporter.py
mreading: For consistency across the auto-filers, it uses the same login info!

tools/jenkins/sql_coverage_reporter.py
mreading: It works for JUnit, so I believe it will work - I will check it out

tools/jenkins/sql_coverage_reporter.py
```diff
@@ -0,0 +1,93 @@
...
+        description += \
+            'SQL Statements:\n' \
+            + ' : Valid=' + str(stats['valid_statements']) + ' (' + stats['valid_percent'] + '%)' \
+            + ' : Invalid=' + str(stats['invalid_statements']) + ' (' + stats['invalid_percent'] + '%)' \
+            + ' : Total=' + str(stats['total_statements']) + '\n' \
+            + 'Test Failures:\n' \
+            + ' : Mismatched=' + str(stats['mismatched_statements']) + ' (' + stats['mismatched_percent'] + '%)\n' \
+            + 'Exceptions:\n' \
+            + ' : Volt Fatal=' + str(stats['total_volt_fatal_excep']) \
+            + ' : Volt Nonfatal=' + str(stats['total_volt_nonfatal_excep']) + '\n' \
```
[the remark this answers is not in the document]
mreading: Done!

PR #5453, opened 2018-07-09
Eng-13919 : Disable HTTP TRACE for embedded Jetty.

Yusarin:


tests/frontend/org/voltdb/TestHttpTraceDisabled.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public void testCannotAccessRootPathByTrace() throws IOException {
+        URL url = new URL(URL);
+        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+        conn.setRequestMethod("TRACE");
+        int code = conn.getResponseCode();
+        assertEquals(HttpServletResponse.SC_FORBIDDEN, code);
+    }
+
+    public void testCannotAccessCSSPathByTrace() throws IOException {
```
manjujames: Looks good. Only thing I would say is to have one method, may be called testCannotAccessPathByTrace, which takes in a URL path, so that you don't have to duplicate the same code in multiple methods. This method can have an array of URLs -- { URL, URL+"css", URL+"images"...}. Then you can loop through this array to do the test.

If we add any new handlers, all we need to do is to add to this array.

PR #5457, opened 2018-07-11
ENG-13765, export data source send out MIGRATE_MASTER event when part…

luyangco:
…ition leader migrates to different host.

Cartographer notifies export data sources that partition leader is migrated. Export generation holds the
event until all streams of the same partition are ready to migrate. On receiver side, receiver picks up from
where prior master left and continue exporting.

Change-Id: I5280236e395964a1e2f6be62dfa266d3dc100db2

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -284,6 +284,10 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
     }
 
     public synchronized void updateAckMailboxes(final Pair<Mailbox, ImmutableList<Long>> ackMailboxes) {
+        if (exportLog.isDebugEnabled()) {
+            exportLog.debug("Mailbox " + CoreUtils.hsIdToString(ackMailboxes.getFirst().getHSId()) + " is registered for " + this.toString() +
+                    " : replicas " + CoreUtils.hsIdCollectionToString(ackMailboxes.getSecond()) );
+        }
         m_ackMailboxRefs.set( ackMailboxes);
     }
 
@@ -833,6 +837,11 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
 
             for( Long siteId: p.getSecond()) {
                 mbx.send(siteId, bpm);
+            }
+            if (exportLog.isDebugEnabled()) {
+                exportLog.debug("Send RELEASE_BUFFER for partition " + m_partitionId + "source signature " + m_tableName + " with uso " + uso
+                        + " from " + CoreUtils.hsIdToString(mbx.getHSId())
+                        + " to " + CoreUtils.hsIdCollectionToString(p.getSecond()));
             }
         }
     }
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -79,7 +79,25 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
     private final byte [] m_signatureBytes;
     private final int m_partitionId;
     private final ExportFormat m_format;
+
     private long m_firstUnpolledUso = 0;
+    private long m_lastReleaseUso = 0;
+    // End uso of most recently pushed export buffer
+    private long m_lastPushedUso = 0L;
+    // Relinquishes export master after this uso
+    private long m_usoToDrain = -1L;
+    // EDS that will be demoted soon
+    private boolean m_retiredMaster = false;
+    // EDS that will be promoted as export master soon
+    private boolean m_wannabeMaster = false;
+    // This flag indicates there is no export master for the stream
+    private boolean m_anarchy = false;
+    //This is released when all mailboxes are set.
+    private final Semaphore m_allowAcceptingMastership = new Semaphore(0);
+    // This EDS is export master when the flag set to true
+    private volatile AtomicBoolean m_mastershipAccepted = new AtomicBoolean(false);
+
+    private volatile boolean m_closed = false;
     private final StreamBlockQueue m_committedBuffers;
     private Runnable m_onMastership;
     private SettableFuture<BBContainer> m_pollFuture;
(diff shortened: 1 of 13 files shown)


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -703,8 +796,64 @@ void handlePartitionFailure(int partitionId) {
...
+
+            ByteBuffer buf = ByteBuffer.allocate(msgLen);
+            buf.put(ExportManager.MIGRATE_MASTER);
+            buf.putInt(partitionId);
+            buf.putInt(partitionDataSourceMap.size());
+
+            for (ExportDataSource eds : partitionDataSourceMap.values()) {
+                buf.putInt(eds.getTableSignature().length);
+                buf.put(eds.getTableSignature());
+                buf.putLong(eds.getLastReleaseUso());
```
manjujames: This should be the last pushed uso, correct?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -79,24 +79,35 @@
     private final byte [] m_signatureBytes;
     private final int m_partitionId;
     private final ExportFormat m_format;
+
     private long m_firstUnpolledUso = 0;
+    private long m_lastReleaseUso = 0;
```
manjujames: May be m_lastReleasedUso

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -114,8 +116,23 @@ public ExportGeneration(File exportOverflowDirectory) throws IOException {
...
+        initializeGenerationFromCatalog(catalogContext, connectors, hostId, messenger, partitions);
+        Set<Integer> allPartitions = new HashSet<Integer>(partitionsFromDisk);
+        allPartitions.addAll(partitions);
+        // One export mailbox per node, since we only keep one generation
+        createAckMailboxes(messenger, allPartitions);
+        updateReplicaList(messenger, allPartitions);
+    }
+
+    List<Integer> initializeGenerationFromDisk(HostMessenger messenger) {
+        List<Integer> partitions = new ArrayList<Integer>();
```
manjujames: Originally this was a Set. Why did you have to change it to a List?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -891,20 +915,36 @@ public boolean isRunEveryWhere() {
...
+                if (exportLog.isDebugEnabled()) {
+                    exportLog.debug("Export table " + getTableName() + " mastership prepare to be demoted for partition " + getPartitionId());
+                }
+                m_retiredMaster = true;
+                // memorize end USO of the most recently pushed buffer from EE
+                m_usoToDrain = m_lastPushedUso;
+                // if no new buffer to be drained, send the migrate event right away
+                if (m_usoToDrain == m_lastReleaseUso) {
+                    unacceptMastership();
+                    m_generation.migrateDataSource(ExportDataSource.this);
```
manjujames: Can we name this method migrateMasterFromHere or something that like that, which will make it clear that this call is to migrate away from this node? I think it will make it a lot easier to read the code and understand.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -891,20 +915,36 @@ public boolean isRunEveryWhere() {
...
-        // TODO
-        // ENG-13952
-        // ENG-13765
+    void prepareUnacceptMastership() {
+        m_es.submit(new Runnable() {
+            public void run() {
+                if (exportLog.isDebugEnabled()) {
+                    exportLog.debug("Export table " + getTableName() + " mastership prepare to be demoted for partition " + getPartitionId());
+                }
+                m_retiredMaster = true;
```
manjujames: This should really be m_retiringMaster, right? Because this gets set to true before we retire this master and once it is fully retired in unacceptMastership, we set this to false.

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -703,8 +796,64 @@ void handlePartitionFailure(int partitionId) {
...
+            if (exportLog.isDebugEnabled()) {
+                exportLog.debug("Partition " + partitionId + " mailbox hsid (" +
+                        CoreUtils.hsIdToString(m_mbox.getHSId()) + ") send MIGRATE_MASTER message to " +
+                        CoreUtils.hsIdCollectionToString(m_replicasHSIds.get(partitionId)));
+            }
+        }
+    }
+
+    public void migrateDataSource(ExportDataSource source) {
+        Map<String, ExportDataSource> partitionDataSourceMap =
```
manjujames: One question - Does this mean that all other streams for this partition will be at a standstill until the last stream for this partition has drained it's lastPushedUso? This will be a problem if one export target is down for some reason. I didn't think about this earlier, but I think we should migrate export data sources as they become available.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -79,24 +79,35 @@
...
+    // End uso of most recently pushed export buffer
+    private long m_lastPushedUso = 0L;
+    // Relinquishes export master after this uso
+    private long m_usoToDrain = -1L;
+    // EDS that will be demoted soon
+    private boolean m_retiredMaster = false;
+    // EDS that will be promoted as export master soon
+    private boolean m_wannabeMaster = false;
+    // This flag indicates there is no export master for the stream
+    private boolean m_anarchy = false;
```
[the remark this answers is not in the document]
manjujames: Actually, reading through the code again, this is set to true when the EDS on this node is not the master. So, we should probably name it variable accordingly as well.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -878,9 +891,20 @@ private void ackImpl(long uso) {
                 VoltDB.crashLocalVoltDB("Error attempting to release export bytes", true, e);
                 return;
             }
+            // Current master sends migrate event to all export replicas
+            if (m_retiredMaster && uso >= m_usoToDrain) {
+                unacceptMastership();
+                m_generation.migrateDataSource(this);
+            }
+            // Receiving a ACK means new master is promoted, clear the anarchy state
+            if (m_anarchy && uso >= m_usoToDrain) {
```
manjujames: It is still not clear to me what this anarchy flag is for. And why this usoToDrain check here?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -915,24 +906,59 @@ public boolean isRunEveryWhere() {
...
+                // Just send to the ack mailbox on the new master
+                if (CoreUtils.getHostIdFromHSId(siteId) != m_newLeaderHostId) {
+                    continue;
+                }
+                mbx.send(siteId, bpm);
+                if (exportLog.isDebugEnabled()) {
+                    exportLog.debug("Partition " + m_partitionId + " mailbox hsid (" +
+                            CoreUtils.hsIdToString(mbx.getHSId()) + ") send TAKE_MASTERSHIP message to " +
+                            CoreUtils.hsIdToString(siteId));
+                }
```
manjujames: We can break out of the for loop here, correct?

PR #5458, opened 2018-07-12
export flush

stakutis:
Making the export flush timeout settable

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -146,7 +146,8 @@ class EnginePlanSet : public PlanSet { };
 VoltEEExceptionType VoltDBEngine::s_loadTableException = VOLT_EE_EXCEPTION_TYPE_NONE;
 
 VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
-    : m_currentIndexInBatch(-1),
+    :
+      m_currentIndexInBatch(-1),
       m_currentUndoQuantum(NULL),
       m_siteId(-1),
       m_isLowestSite(false),
@@ -162,7 +163,6 @@ VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
       m_executorContext(NULL),
       m_drPartitionedConflictStreamedTable(NULL),
       m_drReplicatedConflictStreamedTable(NULL),
-      m_drStream(NULL),
       m_drReplicatedStream(NULL),
       m_currExecutorVec(NULL)
 {
@@ -179,7 +179,8 @@ VoltDBEngine::initialize(int32_t clusterIndex,
                          int32_t defaultDrBufferSize,
                          int64_t tempTableMemoryLimit,
                          bool isLowestSiteId,
-                         int32_t compactionThreshold)
+                         int32_t compactionThreshold,
+                         int32_t maxBufferAge)
 {
     m_clusterIndex = clusterIndex;
     m_siteId = siteId;
@@ -187,6 +188,7 @@ VoltDBEngine::initialize(int32_t clusterIndex,
     m_partitionId = partitionId;
     m_tempTableMemoryLimit = tempTableMemoryLimit;
     m_compactionThreshold = compactionThreshold;
+    m_maxBufferAge = maxBufferAge;
 
     // Instantiate our catalog - it will be populated later on by load()
     m_catalog.reset(new catalog::Catalog());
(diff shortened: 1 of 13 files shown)


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1532,9 +1534,7 @@ void
 VoltDBEngine::markAllExportingStreamsNew() {
     //Mark all streams new so that schema is sent on next tuple.
     BOOST_FOREACH (LabeledStreamWrapper entry, m_exportingStreams) {
-        if (entry.second != NULL) {
```
wweiss-voltdb: You need to rebase master. Here you are reverting a previous commit. Talk with Ruth about how to rebase.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -162,7 +163,6 @@ VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
       m_executorContext(NULL),
       m_drPartitionedConflictStreamedTable(NULL),
       m_drReplicatedConflictStreamedTable(NULL),
-      m_drStream(NULL),
```
wweiss-voltdb: Not sure why this line is deleted.

src/ee/execution/VoltDBEngine.h
```diff
@@ -121,6 +121,9 @@ typedef std::unique_ptr<AbstractTempTable, TempTableTupleDeleter> UniqueTempTabl
 
 const int64_t DEFAULT_TEMP_TABLE_MEMORY = 1024 * 1024 * 100;
 
+static  int32_t m_maxBufferAge=4*1005; /** maxBufferAge in ms for export stream flushing STAKUTIS**/
```
wweiss-voltdb: Can you rename this variable so it is clear that this is for export. Something like exportFlushTimeout.
Also by convention, m_ is reserved for member variables. Statics are usually prefixed with s_.
Not sure why this is 1005 rather than 1000?

src/ee/execution/VoltDBEngine.h
```diff
@@ -142,7 +145,8 @@ class __attribute__((visibility("default"))) VoltDBEngine {
                         int32_t defaultDrBufferSize,
                         int64_t tempTableMemoryLimit,
                         bool createDrReplicatedStream,
-                        int32_t compactionThreshold = 95);
+                        int32_t compactionThreshold = 95,
+                        int32_t maxBufferAge = 4*1002);
```
wweiss-voltdb: Why is this 1002?

src/ee/voltdbjni.cpp
```diff
@@ -269,7 +269,8 @@ SHAREDLIB_JNIEXPORT jint JNICALL Java_org_voltdb_jni_ExecutionEngine_nativeIniti
     jint defaultDrBufferSize,
     jlong tempTableMemory,
     jboolean createDrReplicatedStream,
-    jint compactionThreshold)
+    jint compactionThreshold,
+    jint maxBufferAge) // STAKUTIS2
```
wweiss-voltdb: Remove comments: STAKUTIS2 and STAKUTIS...

src/ee/execution/VoltDBEngine.h
wweiss-voltdb: Where is this variable actually used?

src/ee/execution/VoltDBEngine.cpp
stakutis: Ok, hmmm.  I wrote her.  Not sure what exactly I did wrong; i took "latest" with JW but I'm guessing that is not how to do things correctly.  Will fix.

src/ee/execution/VoltDBEngine.cpp
stakutis: Hmmm.  I have no clue. I dont think I did it...? I had no reason to be near that code.  Maybe it is the source-tree/wrong issue again?  Maybe I somehow fat-fingered (but can't easily see how).  I can easily restore  and put it back, but it just seems curious to me.

src/ee/execution/VoltDBEngine.h
stakutis: Right . Good point.  It *WAS* an m_ thing for a few days (which didn't work).  I've changed it to "s_"; i didn't see that convention elsewhere but will follow.  Also I changed all the 100x* numbers back to 1000 (it was just convenient to test and find where things were coming from).   And I've taken my STAKUTIS name out (again, it was good to know exactly what I touched for a while).

src/ee/execution/VoltDBEngine.h
stakutis: done

PR #5461, opened 2018-07-13
added an xdcr tutorial

abradley201:


examples/HOWTOs/xdcr_tutorial.md
```diff
@@ -0,0 +1,98 @@
...
+
+		voltdb start --host="node1,node2,node3"
+		voltdb start --host="node4,node5,node6"
+
+As you did when first creating the xdcr cluster. The two node clusters will use the command logs and dr_overflow files to recreate the data and start following each other again.
+
+
+However, if you need to make a change to the configuration file or schema that cannot be done while the clusters are running, you may need to stop the cluster and restore from snapshot.
+
+First, create a directory to use as a place to store the snapshot files. It should be easily accessible by all the nodes. It can be saved like this:
```
benjaminballard: Instead of "It should be easily accessible by all the nodes." it should say "The directory needs to exist on all the nodes." so as not to imply that you should use a shared network drive.

examples/HOWTOs/xdcr_tutorial.md
```diff
@@ -0,0 +1,98 @@
...
+
+		sqlcmd < ddl.sql
+
+Note that if the schema changes too much, such as deleting tables, the snapshot might not restore correctly.
+Finally, use the following command to restore from snapshot:
+
+		voltadmin restore ~/restore snapshot_1
+
+Once this completes and is successful, restart the second cluster and reload its schema with sqlcmd < ddl.sql. Your two xdcr clusters should be ready and copying each other as normal with their new configuration file and/or schema.
+
```
benjaminballard: Here is where we should add some steps on how to monitor the status of the snapshot sync and verify when replication reaches the point where it is running normally. For testing, you may need a large snapshot so that it takes more than a few seconds and you can capture each stage in the process.

benjaminballard: I think this is great, and it's an excellent starting point.

It needs to have at least one link to the Using VoltDB chapter on XDCR, so it doesn't duplicate what is there. That is the reference, and this is an example.

Also, before we merge this into master and into the public repo and the product, I think it needs something more so that no one can say this duplicates what is in the documentation or doesn't add anything that wasn't already documented. The instructions on how to check the status and interpret the results would be helpful. The Using VoltDB chapter on XDCR doesn't mention the @Statistics, and the reference page for @Statistics has definitions for the different states, but doesn't tie it in with the process of establishing a cluster, or other scenarios.

PR #5463, opened 2018-07-13
ENG-10675-UDF - a SQL function to convert a timestamp to a String in a given timezone

Chaozz:
Implement **FORMAT_TIMESTAMP(timestamp, timezone_or_time_offset)**

+ I reimplemented this in Java, using the UDF mechanism. 

+ *timezone_or_time_offset* must be a valid string either a [offset](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#of-java.lang.String-) or a [timezone](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#of-java.lang.String-).

### DIFF
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -2806,7 +2806,7 @@ void VoltDBEngine::setViewsEnabled(const std::string& viewNames, bool value) {
     } while (updateReplicated);
 }
 
-void VoltDBEngine::loadBuiltInJavaFunctions(){
+void VoltDBEngine::loadBuiltInJavaFunctions() {
     // Hard code the info of format_timestamp function
     UserDefinedFunctionInfo *info = new UserDefinedFunctionInfo();
     info->returnType = VALUE_TYPE_VARCHAR;
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -69,6 +69,7 @@
 
 #include "common/SynchronizedThreadLock.h"
 #include "executors/abstractexecutor.h"
+#include "expressions/functionexpression.h"
 
 #include "indexes/tableindex.h"
 #include "indexes/tableindexfactory.h"
@@ -166,6 +167,7 @@ VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
       m_drReplicatedStream(NULL),
       m_currExecutorVec(NULL)
 {
+    initJavaBuildinSQLFunctionInfo();
 }
 
 void
@@ -2802,6 +2804,17 @@ void VoltDBEngine::setViewsEnabled(const std::string& viewNames, bool value) {
         }
         updateReplicated = ! updateReplicated;
     } while (updateReplicated);
+}
+
+void VoltDBEngine::initJavaBuildinSQLFunctionInfo(){
+    // Hard code the info of format_stamp function
+    UserDefinedFunctionInfo *info = new UserDefinedFunctionInfo();
+    info->returnType = VALUE_TYPE_VARCHAR;
+    info->paramTypes.resize(2);
+    info->paramTypes.at(0) = VALUE_TYPE_TIMESTAMP;
+    info->paramTypes.at(1) = VALUE_TYPE_VARCHAR;
+
+    m_functionInfo[FUNC_VOLT_FORMAT_TIMESTAMP] = info;
 }
 
 } // namespace voltdb
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -167,7 +167,7 @@ VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
       m_drReplicatedStream(NULL),
       m_currExecutorVec(NULL)
 {
-    initJavaBuildInSQLFunctionInfo();
+    initJavaBuiltInSQLFunctionInfo();
 }
 
 void
@@ -2806,7 +2806,7 @@ void VoltDBEngine::setViewsEnabled(const std::string& viewNames, bool value) {
     } while (updateReplicated);
 }
 
-void VoltDBEngine::initJavaBuildInSQLFunctionInfo(){
+void VoltDBEngine::initJavaBuiltInSQLFunctionInfo(){
     // Hard code the info of format_stamp function
     UserDefinedFunctionInfo *info = new UserDefinedFunctionInfo();
     info->returnType = VALUE_TYPE_VARCHAR;
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -167,7 +167,7 @@ VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
       m_drReplicatedStream(NULL),
       m_currExecutorVec(NULL)
 {
-    initJavaBuiltInSQLFunctionInfo();
+    loadBuiltInJavaFunctions();
 }
 
 void
@@ -2806,8 +2806,8 @@ void VoltDBEngine::setViewsEnabled(const std::string& viewNames, bool value) {
     } while (updateReplicated);
 }
 
-void VoltDBEngine::initJavaBuiltInSQLFunctionInfo(){
-    // Hard code the info of format_stamp function
+void VoltDBEngine::loadBuiltInJavaFunctions(){
+    // Hard code the info of format_timestamp function
     UserDefinedFunctionInfo *info = new UserDefinedFunctionInfo();
     info->returnType = VALUE_TYPE_VARCHAR;
     info->paramTypes.resize(2);
--- a/src/ee/execution/VoltDBEngine.cpp
+++ b/src/ee/execution/VoltDBEngine.cpp
@@ -167,7 +167,7 @@ VoltDBEngine::VoltDBEngine(Topend* topend, LogProxy* logProxy)
       m_drReplicatedStream(NULL),
       m_currExecutorVec(NULL)
 {
-    initJavaBuildinSQLFunctionInfo();
+    initJavaBuildInSQLFunctionInfo();
 }
 
 void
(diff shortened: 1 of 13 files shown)


src/frontend/org/voltdb/UserDefinedFunctionManager.java
```diff
@@ -96,9 +99,31 @@ public void loadFunctions(CatalogContext catalogContext) {
...
 
+    private void loadBuildinJavaFunctions(ImmutableMap.Builder<Integer,UserDefinedFunctionRunner> builder) {
+        // define the function object
+        Function func = new Function();
+        func.setFunctionname("FORMAT_TIMESTAMP");
+
+        func.setClassname("org.voltdb.utils.JavaBuildinFunctions");
+        func.setMethodname("format_timestamp");
+        func.setReturntype(Types.VARCHAR);
+        CatalogMap<FunctionParameter> params = new CatalogMap<>(null, null, "params", FunctionParameter.class, 1);
```
Chaozz: I am not sure it is the correct way to create CatalogMap<FunctionParameter> params. But currently it is the only way I found.

src/frontend/org/voltdb/utils/JavaBuildinFunctions.java
```diff
@@ -0,0 +1,60 @@
...
+        if (tz == null) {
+            return ts.toString();
+        }
+        tz = tz.trim();
+        long millis = ts.getTime() / 1000;
+        short usecs = ts.getUSec();
+        Instant fromEpochMilli = Instant.ofEpochMilli(millis);
+        ZoneOffset offset;
+        try {
+            offset = ZoneOffset.of(tz);
```
yzhang1991: From reading the documentation I got the understanding that ZoneId can handle both string abbrev. case and +-offset case. So you do not need to nest two try-catch blocks here? And, maybe `fromEpochMilli` is not needed?

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -3345,4 +3347,45 @@ public void testTimestampValidityFunctions() throws Exception {
...
+        ClientResponse cr;
+
+        cr = client.callProcedure("P2.insert", 1, "2013-07-18 02:00:00.123457");
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+        cr = client.callProcedure("P2.insert", 2, null);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
+
+        // null timestamp will return null
+        cr = client.callProcedure("FORMAT_TIMESTAMP", "UTC", 2);
+        assertEquals(ClientResponse.SUCCESS, cr.getStatus());
```
yzhang1991: You don't need to assert on the status code if you are checking the result table. Same below.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
```diff
@@ -3345,4 +3347,45 @@ public void testTimestampValidityFunctions() throws Exception {
...
+        assertContentOfTable(new Object[][]{{"2013-07-18 02:00:00.123457"}}, cr.getResults()[0]);
+
+        String[] invalid_offsets = {"I_AM_INVALID", "HOME/TRASH", "10:00", "-1:00"};
+        String expectedError = "VOLTDB ERROR: SQL ERROR\\s*" +
+                "invalid timezone string\\s*";
+        for (String invalid_offset : invalid_offsets) {
+            verifyProcFails(client, expectedError, "FORMAT_TIMESTAMP", invalid_offset, 1);
+        }
+
+        String[] offsets = {"UTC", "Australia/Adelaide", "America/Atikokan", "America/Belize", "Asia/Shanghai",
```
yzhang1991: Add PST EDT etc. We store in UTC time zone, testing UTC abbrev. only does not say much.

tests/frontend/org/voltdb/regressionsuites/TestAddDropUDF.java
```diff
@@ -113,7 +113,8 @@ String catalogMatchesCompilerFunctionSet(Client client) throws NoConnectionsExce
         StringBuffer success = new StringBuffer();
         Set<String> dfns = FunctionForVoltDB.getAllUserDefinedFunctionNamesForDebugging();
         VoltTable vt = client.callProcedure("@SystemCatalog", "functions").getResults()[0];
-        if (dfns.size() != vt.getRowCount()) {
+        // FORMAT_TIMESTAMP is a build-in function, not UDF
+        if (!(dfns.size() - 1 == vt.getRowCount() && dfns.contains("FORMAT_TIMESTAMP"))) {
```
yzhang1991: FORMAT_TIMESTAMP should not be in the catalog, neither is this change.

src/frontend/org/voltdb/utils/JavaBuildinFunctions.java
Chaozz: Yeah, I just need the `offset =  ZonedDateTime.ofInstant(fromEpochMilli, ZoneId.of(tz)).getOffset()`
But we still need `fromEpochMilli` cause offset rule is related to current time.

tests/frontend/org/voltdb/regressionsuites/TestFunctionsForVoltDBSuite.java
Chaozz: actually PST, EDT are invalid timezone string in java&tz databae standard, see [wiki](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).  They use country/city pairs mostly.

src/frontend/org/voltdb/UserDefinedFunctionManager.java
```diff
@@ -165,6 +158,55 @@ public UserDefinedFunctionRunner(Function catalogFunction, Object funcInstance)
...
+            for (int i = 0; i < m_paramCount; i++) {
+                m_paramTypes[i] = VoltType.typeFromClass(paramTypeClasses[i]);
+                m_boxUpByteArray[i] = paramTypeClasses[i] == Byte[].class;
+            }
+            m_returnType = VoltType.typeFromClass(m_functionMethod.getReturnType());
+
+            m_logger.debug(String.format("The user-defined function manager is defining function %s (ID = %s)",
+                    m_functionName, m_functionId));
+
+            // We register the token again when initializing the user-defined function manager because
```
yzhang1991: There is still duplicated code that may be able to fold.

Chaozz: These are code blocks that init the final variables.

Chaozz: Done, I rewrited them.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -2804,4 +2806,15 @@ void VoltDBEngine::setViewsEnabled(const std::string& viewNames, bool value) {
     } while (updateReplicated);
 }
 
+void VoltDBEngine::loadBuiltInJavaFunctions(){
```
yzhang1991: Space before `{`

src/frontend/org/voltdb/UserDefinedFunctionManager.java
```diff
@@ -96,9 +97,17 @@ public void loadFunctions(CatalogContext catalogContext) {
...
             builder.put(catalogFunction.getFunctionid(), new UserDefinedFunctionRunner(catalogFunction, funcInstance));
         }
+
+        loadBuiltInJavaFunctions(builder);
         m_udfs = builder.build();
     }
 
+    private void loadBuiltInJavaFunctions(ImmutableMap.Builder<Integer, UserDefinedFunctionRunner> builder) {
+        // define the function object
+        builder.put(FunctionForVoltDB.getFunctionID("format_timestamp"), new UserDefinedFunctionRunner("FORMAT_TIMESTAMP",
```
yzhang1991: Assign the function name `FORMAT_TIMESTAMP` to a variable and use it.
`fn_by_name` is case-insensitive, so you can use the uppercase version for all the occurrences here.
The function ID should also be cached in a variable and reused instead of querying the map repeatedly.

src/frontend/org/voltdb/UserDefinedFunctionManager.java
```diff
@@ -117,14 +126,40 @@ public void loadFunctions(CatalogContext catalogContext) {
...
+            // in a cluster setting the token may only be registered on the node where the CREATE FUNCTION DDL
+            // is executed. We uses a static map in FunctionDescriptor to maintain the token list.
+            FunctionForVoltDB.registerTokenForUDF(m_functionName, m_functionId, m_returnType, m_paramTypes);
+        }
+
+        private void initFunctionMethod(String methodName) {
             for (final Method m : m_functionInstance.getClass().getDeclaredMethods()) {
                 if (m.getName().equals(methodName)) {
-                    if (! Modifier.isPublic(m.getModifiers())) {
+                    if (!Modifier.isPublic(m.getModifiers())) {
```
yzhang1991: Space after `!`

src/frontend/org/voltdb/UserDefinedFunctionManager.java
```diff
@@ -103,9 +105,13 @@ public void loadFunctions(CatalogContext catalogContext) {
...
     private void loadBuiltInJavaFunctions(ImmutableMap.Builder<Integer, UserDefinedFunctionRunner> builder) {
-        // define the function object
-        builder.put(FunctionForVoltDB.getFunctionID("format_timestamp"), new UserDefinedFunctionRunner("FORMAT_TIMESTAMP",
-                FunctionForVoltDB.getFunctionID("format_timestamp"), "format_timestamp", new JavaBuiltInFunctions()));
+        // define the function objects
+        String[] fn_names = {"format_timestamp"};
+        for (String fn_name : fn_names) {
+            int fn_id = FunctionForVoltDB.getFunctionID(fn_name);
+            builder.put(fn_id, new UserDefinedFunctionRunner(fn_name,
+                    fn_id, fn_name, new JavaBuiltInFunctions()));
```
yzhang1991: RE: `fn_names`, `fn_id`
We normally use lower camel case for variable naming.

PR #5469, opened 2018-07-19
Eng 14251 Better Planner Testing Framework

BillWhite:
We have a small framework for testing plans.  This is not plannertester tests, but it's for describing and matching expected plans.

The existing framework just matched on plan node types.  I wanted to match other things, kind of like hamcrest testing, so I refactored this to have a more general plan matcher, which PlanNodeType implements.  This lets us have matchers which have a plan node type and require inline nodes of a given type.  In subsequent commits I have added some more functionality, but that is not easily separated into this commit.

### DIFF
--- a/src/frontend/org/voltdb/types/PlanMatcher.java
+++ b/src/frontend/org/voltdb/types/PlanMatcher.java
@@ -1,0 +1,35 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2018 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.voltdb.types;
+
+import org.voltdb.plannodes.AbstractPlanNode;
+
+/**
+ * A class which can match an abstract plan node.
+ * @author bwhite
+ *
+ */
+public interface PlanMatcher {
+    /**
+     * Match an AbstractPlanNode.  Return null for a match
+     * and an error message otherwise.
+     *
+     * @param node
+     * @return
+     */
+    String match(AbstractPlanNode node);
+}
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/types/PlanNodeType.java
```diff
@@ -49,9 +49,13 @@
 import org.voltdb.plannodes.WindowFunctionPlanNode;
 
 /**
+ * This is the type of plan nodes.
```
BillWhite: This is really the central change.  PlanNodeType can match plan nodes for the purpose of testing plans.

tests/frontend/org/voltdb/planner/PlannerTestCase.java
```diff
@@ -958,7 +992,21 @@ protected PlanMatcher noneOf(String failMessage, Object ... nodeSpecs) {
      */
     protected void validatePlan(String SQL,
                                 FragmentSpec ... spec) {
-        List<AbstractPlanNode> fragments = compileToFragments(SQL);
+        // All this System.out.printf nonsense should be changed
```
BillWhite: Perhaps we should convert PlannerTestCase.compileToTopDownTree to use validatePlan, or else just convert compileToTopDownTree calls to validatePlan calls?

src/frontend/org/voltdb/types/PlanMatcher.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * A class which can match an abstract plan node.
+ *
+ * PlanNodeType implements this, so plan nodes of
+ * a given type can be matched.
+ *
+ * @author bwhite
+ *
+ */
+public interface PlanMatcher {
```
kaikaizi: The name "Matcher" gives me the impression that it accepts two plans and check if they match with each other. But it only take a single plan node and returns null for match. From that POV it's unclear what contract this interface poses.

A different way to do this might be not having this interface (and therefore not requiring PlanNodeType to implements this interface), have the match() method that returns boolean, override toString() method of PlanNodeType enum, and have the caller (tester) call validate() method in a way like:
<code>
   Assert.assertTrue("Expecting " + node1.toString() + ", got " + node2.toString(), node1.match(node2));
</code>

This is my thought to make it a little bit easier to understand.

BillWhite: Well, I wanted to have a description of the plan, and then a way to test the actual plan to the description.  The description is created by a call to fragSpec.  This can be something like a sequence of plan node types.  But it can be more complicated, such as an object which requires that its plan node be an index scan node which scans a particular named index.  Or it can be an object which matches any scan plan.  We have index scan and sequential scans.  Making this an interface allows this kind of flexibility.

Making a description such as you can see in the function TestWindowFunctions.testWindowFunctionsWithIndexes is very compact.  
Look at the functions in TestPlansGroupBy, for example, or even the other functions in TestWindowFunctions to see how painful this has been in the past.

PR #5470, opened 2018-07-19
Eng 11039 - Change internal temporary table name

MarcxJin:


tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
```diff
@@ -993,10 +993,10 @@ private void checkGroupByAliasFeature(String sql1, String sql2, boolean exact) {
         pns = compileToFragments(sql2);
         explainStr2 = buildExplainPlan(pns);
         if (! exact) {
-            explainStr1 = explainStr1.replaceAll("TEMP_TABLE\\.column#[\\d]",
-                    "TEMP_TABLE.column#[Index]");
-            explainStr2 = explainStr2.replaceAll("TEMP_TABLE\\.column#[\\d]",
-                    "TEMP_TABLE.column#[Index]");
+            explainStr1 = explainStr1.replaceAll("TEMP_TABLE_\\$\\$\\.column#[\\d]",
```
BillWhite: Is there some way to use the constant TEMP_TABLE_NAME to construct this string?
It would be nice if the string "$$TEMP_TABLE_NAME$$" appeared only once in the
source, and could be changed if necessary.  You would think this is the final time we would
change this, but that's what we thought the last time we did this change.

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
```diff
@@ -993,10 +993,10 @@ private void checkGroupByAliasFeature(String sql1, String sql2, boolean exact) {
...
+                    "TEMP_TABLE_\\$\\$.column#[Index]");
+            explainStr2 = explainStr2.replaceAll("TEMP_TABLE_\\$\\$\\.column#[\\d]",
+                    "TEMP_TABLE_\\$\\$.column#[Index]");
```
BillWhite: Why is this not \\$\\$_TEMP_TABLE_\\$\\$?  I guess it doesn't really matter, but can TEMP_TABLE_\\$\\$ occur in any other table name?  I wouldn't think so.

tests/scripts/plannertester/ENG-4077-prefer-filtering-index/baseline/ENG-4077-prefer-filtering-index.plan526
```diff
@@ -1,2 +1,2 @@
...
...1","COLUMN_NAME":"ID","COLUMN_ALIAS":"ID","EXPRESSION":{"TYPE":"VALUE_TUPLE","VALUE_TYPE":"INTEGER","VALUE_SIZE":4,"COLUMN_IDX":0,"TABLE_NAME":"IR1","COLUMN_NAME":"ID","COLUMN_ALIAS":"ID"},"TYPE":"INTEGER","SIZE":4},{"TABLE_NAME":"IR1","COLUMN_NAME":"NUM","COLUMN_ALIAS":"NUM","EXPRESSION":{"TYPE":"VALUE_TUPLE","VALUE_TYPE":"INTEGER","VALUE_SIZE":4,"COLUMN_IDX":2,"TABLE_NAME":"IR1","COLUMN_NAME":"NUM","COLUMN_ALIAS":"NUM"},"TYPE":"INTEGER","SIZE":4}]}],"CHILDREN_IDS":[],"PARENT_IDS":[5],"OUTPUT_SCHEMA":[{"TABLE_NAME":"IR1","COLUMN_NAME":"ID","COLUMN_ALIAS":"ID","EXPRESSION":{"TYPE":"VALUE_TUPLE","VALUE_TYPE":"INTEGER","VALUE_SIZE":4,"COLUMN_IDX":0,"TABLE_NAME":"IR1","COLUMN_NAME":"ID","COLUMN_ALIAS":"ID"},"TYPE":"INTEGER","SIZE":4},{"TABLE_NAME":"IR1","COLUMN_NAME":"NUM","COLUMN_ALIAS":"NUM","EXPRESSION":{"TYPE":"VALUE_TUPLE","VALUE_TYPE":"INTEGER","VALUE_SIZE":4,"COLUMN_IDX":2,"TABLE_NAME":"IR1","COLUMN_NAME":"NUM","COLUMN_ALIAS":"NUM"},"TYPE":"INTEGER","SIZE":4}],"PREDICATE":null,"TARGET_TABLE_NAME":"IR1","KEY_ITERATE":true,"LOOKUP_TYPE":"GTE","SORT_DIRECTION":"ASC","TARGET_INDEX_NAME":"SYS_IDX_SYS_PK_10031_10032","END_EXPRESSION":null,"SEARCHKEY_EXPRESSIONS":[]}],"PARAMETERS":[]}
```
BillWhite: Plannertester tests are all very fragile.  You have my sympathies.  Their only saving grace is that they are very fast.

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
MarcxJin: We do actually have $$_VOLT_TEMP_TABLE_$$ as a static variable,  but we couldn't directly use that in the replaceAll function (error: illegal group reference) since the dollar sign needs escaping.  I prefer to fill the hard-type string here.

tests/frontend/org/voltdb/planner/TestPlansGroupBy.java
MarcxJin: Well.. this naming convention can prevent the majority of 'accidentally' name conflicts, but I agree with you for being more strict.

PR #5480, opened 2018-07-24
ENG-14266, LIKE/STARTS WITH Null Pointer Exception

MarcxJin:


src/hsqldb19b3/org/hsqldb_voltpatches/Like.java
```diff
@@ -336,8 +345,13 @@ boolean hasWildcards() {
         return iFirstWildCard != -1;
     }
 
+    // Specific check for 'col LIKE CAST(NULL AS VARCHAR)' case
+    boolean isEquivalentToCastPredicate() {
```
Chaozz: The function name doesn't make sense to me, how about `isEquivalentToCastNullPredicate`?

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -157,7 +166,12 @@ void setPattern(Session session, Object pattern) {
     }
 
     boolean isEquivalentToUnknownPredicate() {
-        return isNull;
+        return isNull && !isRightNull;
+    }
+
+    // Specific check for 'col STARTS WITH CAST(NULL AS VARCHAR)' case
+    boolean isEquivalentToCastPredicate() {
```
Chaozz: Same, try a better name.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
```diff
@@ -286,23 +286,24 @@ public void resolveTypes(Session session, Expression parent) {
...
-        if (likeObject.isEquivalentToUnknownPredicate()) {
+        if (likeObject.isEquivalentToCastPredicate()) {
+            // ENG-14266 solve 'col LIKE CAST(NULL AS VARCHAR)' problem
+            // If it is this case, we are already set.
+            // EE can handle this (left expression is a ExpressionColumn, right expression is a null VALUE).
+            return;
+        } else if (likeObject.isEquivalentToUnknownPredicate()) {
             this.setAsConstantValue(null);
-
-            likeObject = null;
```
Chaozz: Could you explain why you delete this line? 

I guess they assign the `likeObject` to null in order to let the object garbage collected as soon as we don't need it anymore.  It could save the memory usage if we have a very large parse tree and need to parse these expression nodes recursively.

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionStartsWith.java
```diff
@@ -168,14 +171,12 @@ public void resolveTypes(Session session, Expression parent) {
             nodes[RIGHT] = new ExpressionLogical(OpTypes.SMALLER_EQUAL,
                                                  leftOld, rightBound);
             opType     = OpTypes.AND;
-            startsWithObject = null;
```
Chaozz: same question above

src/hsqldb19b3/org/hsqldb_voltpatches/Like.java
```diff
@@ -261,6 +261,15 @@ void setPattern(Session session, Object pattern, Object escape,
...
 
         if (isNull) {
+            /* ENG-14266, solve 'col LIKE CAST(NULL AS VARCHAR)' Null Pointer problem
+             * In this particular case, the right expression has been turned into VALUE type whose valueData is null.
+             * EE can handle this case.
+             * isRightNull set to be true if it is this case.
+             */
+            isRightNull = (nodes[Expression.LEFT] instanceof ExpressionColumn) &&
+                          (nodes[Expression.RIGHT] instanceof ExpressionOp) &&
+                          (nodes[Expression.RIGHT].getType() == 1) &&
```
Chaozz: More readable if you could use the defined variable, is it `OpTypes.VALUE`?

src/hsqldb19b3/org/hsqldb_voltpatches/StartsWith.java
```diff
@@ -136,11 +136,20 @@ private boolean compareAt(Object o, int i, int j, int jLen) {
...
 
         if (isNull) {
+            /* ENG-14266, solve 'col STARTS WITH CAST(NULL AS VARCHAR)' Null Pointer problem
+             * In this particular case, the right expression has been turned into VALUE type whose valueData is null.
+             * EE can handle this case.
+             * isRightNull set to be true if it is this case.
+             */
+            isRightNull = (nodes[Expression.LEFT] instanceof ExpressionColumn) &&
+                          (nodes[Expression.RIGHT] instanceof ExpressionOp) &&
+                          (nodes[Expression.RIGHT].getType() == 1) &&
```
Chaozz: same above

src/hsqldb19b3/org/hsqldb_voltpatches/ExpressionLike.java
MarcxJin: I didn't figure out what this means (but now I know it may be due to GC mechanism), and just thought it's redundant, deleted that and no side-effects happened. I can fix this.

PR #5481, opened 2018-07-24
Query Optimization - Expression normalization and simplifcation for integers

kaikaizi:
A pretty big changeset, rewrites all AbstractExpressions.

As per team discussion, the current plan is to merge in the expression simplifier/normalizer without having the normalizer module in use. Because the normalization/optimization of expression can be very aggressive, user may get surprised especially when trying to match indexes of some table with predicates.

Also not that the MV optimizer has been updated to take advantage of what this normalization has to offer.

__To help you start this review, look at `tests/frontend/org/voltdb/expressions/TinySexpSerializer.java` and `tests/frontend/org/voltdb/planner/optimizer/TestExpressionNormalizer.java` first. These tests helps you understand how I do JUnit test below SQL statement level, what each component achieves and how end-to-end change look like. Then you can look at each individual piece.__

### DIFF
--- a/src/frontend/org/voltdb/planner/optimizer/ArithmeticExpressionFlattener.java
+++ b/src/frontend/org/voltdb/planner/optimizer/ArithmeticExpressionFlattener.java
@@ -1,0 +1,188 @@
+package org.voltdb.planner.optimizer;
+
+import org.voltcore.utils.Pair;
+import org.voltdb.expressions.AbstractExpression;
+import org.voltdb.expressions.OperatorExpression;
+import org.voltdb.types.ExpressionType;
+import static org.voltdb.planner.optimizer.NormalizerUtil.ArithOpType;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+/**
+ * Flattens arithmetic expression by grouping +/- together, * / together, into layered multi-way tree structure.
+ * e.g. (a + b - c) * (a + c - d) / c - (a + c) / (a + d) * b ==>
+ * bottom layer: [ a + b - c, a + c - d, c] and [a + c, a + d, b]
+ */
+final class ArithmeticExpressionFlattener {
+    private final ArithOpType m_op;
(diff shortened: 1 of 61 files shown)


src/frontend/org/voltdb/planner/optimizer/ArithmeticExpressionFlattener.java
```diff
@@ -0,0 +1,204 @@
+/* This file is part of VoltDB.
```
kaikaizi: New things in this new `optimizer/` folder were crafted and (quite extensively) tested, both in unit test and in real SQL regression suite, etc. They are the backbone of my multi-month effort for expression normalization.
It heavily uses FP with its advantages, which suites best for this kind of work.

tests/frontend/org/voltdb/planner/TestMVOptimization.java
```diff
@@ -164,10 +164,9 @@ public void testGroupByMultipleColumns() {
     public void testComplexGbyExpr() {  // test complex GBY expressions
         checkQueriesPlansAreTheSame("SELECT a * 2 + a1, b - a, SUM(a1), MIN(b1), COUNT(*) FROM t1 GROUP BY a * 2 + a1, b - a;",
                 "SELECT a2pa1, b_minus_a, sum_a1, min_b1, counts FROM v4");
-        // negative test: equivalent but different expression
-        checkQueriesPlansAreDifferent("SELECT a1 + a * 2, b - a, SUM(a1), MIN(b1), COUNT(*) FROM t1 GROUP BY a1 + a * 2, b - a;",
-                "SELECT a2pa1, b_minus_a, sum_a1, min_b1, counts FROM v4",
-                "GBY expression equivalent but different: should not match view");
+        // equivalent but different GBY expression
```
kaikaizi: The simple transformation of arithmetic expression (from `a * 2 + a1` to `a1 + a * 2`) happen to be detected by the equivalence relation of expression, hence this ad-hoc SELECT query gets to be optimized against existing view.

PR #5488, opened 2018-07-27
ENG-13878 Force shutdown even if not all exporter transactions are processed.

Yusarin:
If we do soft shutdown when the exporter transactions are not completed, it will shutdown immediately without hanging and printing out export pending information repeatedly. Not sure whether this is the correct behavior.


lib/python/voltcli/checkstats.py
```diff
@@ -45,6 +45,7 @@ def check_exporter(runner):
                 notifyInterval = 10
                 if last_table_stat_time > 1 and export_tables_with_data:
                     print_export_pending(runner, export_tables_with_data)
+                    return
```
dave-voltdb, line 48: This does not seem right to me.  Your change says that if there are any export tables with data then we will skip the check for progress.  

The intent of the code in lines 44 to 47 is notification. The decision of whether to give up waiting is made by the monitorStatsProgress routine.

PR #5489, opened 2018-07-27
ENG-14332-DR Tab not appearing in VMC since jetty sessions was changed

Chaozz:
It seems a small javascript issue, not a jetty one.

I also noticed the bug happened since **7.7.2**, should I also backport it?

### DIFF
--- a/src/frontend/org/voltdb/dbmonitor/js/voltdb.service.js
+++ b/src/frontend/org/voltdb/dbmonitor/js/voltdb.service.js
@@ -70,7 +70,7 @@
          };
 
         this.GetClusterInformation = function (onConnectionAdded) {
-            this.processTask(onConnectionAdded, "CLUSTER_INFORMATION", ['@SystemInformation'], ["OVERVIEW"], [undefined]);
+            this.processTask(onConnectionAdded, "CLUSTER_INFORMATION", ['@SystemInformation'], ["OVERVIEW"], []);
         };
 
         this.GetSystemInformationDeployment = function (onConnectionAdded) {
@@ -409,7 +409,7 @@
 
         //Check if cluster is replica or not
         this.GetClusterReplicaInformation = function (onConnectionAdded) {
-            this.processTask(onConnectionAdded, "CLUSTER_REPLICA_INFORMATION", ['@SystemInformation'], ["Overview"], [undefined]);
+            this.processTask(onConnectionAdded, "CLUSTER_REPLICA_INFORMATION", ['@SystemInformation'], ["Overview"], []);
         };
 
         this.GetDrRoleInformation = function (onConnectionAdded) {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/dbmonitor/js/voltdb.ui.js
```diff
@@ -1202,8 +1202,6 @@ var loadPage = function (serverName, portid) {
                         var currentServer = getCurrentServer();
                         if (replicaDetail.hasOwnProperty(currentServer))
```
Chaozz: The `replicaDetail` here is an empty object.
I diffed with the old version **7.3**, they have the same **if-else** clause here. Maybe this `replicaDetail` is not initialized correctly, I am not sure if I should fix it. Currently the webpage looks good to me.

Chaozz: Well, it seems not the root case, needs some more work to do.

src/frontend/org/voltdb/dbmonitor/js/voltdb.service.js
```diff
@@ -409,7 +409,7 @@
 
         //Check if cluster is replica or not
         this.GetClusterReplicaInformation = function (onConnectionAdded) {
-            this.processTask(onConnectionAdded, "CLUSTER_REPLICA_INFORMATION", ['@SystemInformation'], ["Overview"], ['0']);
+            this.processTask(onConnectionAdded, "CLUSTER_REPLICA_INFORMATION", ['@SystemInformation'], ["Overview"], [undefined]);
```
Chaozz: Here is the root cause, in **7.7.2** they rewrite this *voltdb.service.js* and made an error here, any version under **7.7.2** is OK.

yzhang1991: I feel that you can remove the `undefined`

src/frontend/org/voltdb/dbmonitor/js/voltdb.ui.js
```diff
@@ -1256,10 +1258,6 @@ var loadPage = function (serverName, portid) {
                                                             $("#divDrReplication").hide();
                                                         }
                                                     } else {
-                                                        voltDbRenderer.GetDrInformations(function (clusterInfo) {
-                                                            $('#clusterId').show();
```
Chaozz: May not relate to this ticket, but here is another bug. Master try to get the cluster ID using  `@Statistics DRCONSUMER`

According to the [doc](https://docs.voltdb.com/UsingVoltDB/sysprocstatistics.php),  **DRCONSUMER** only valid when the role is *replica* or *XDCR*. 
Now I get the cluster id in https://github.com/VoltDB/voltdb/blob/01d48e7cef3d44727525618b394a317969ee7a3d/src/frontend/org/voltdb/dbmonitor/js/voltdb.ui.js#L1167

PR #5495, opened 2018-07-31
ENG-14347, master stream forwards an ACK to newly rejoined streams.

luyangco:
When newly joined or rejoined streams miss any RELEASE_BUFFER event, and client is stopped, it causes
export buffers on rejoined nodes never get drained. Forwarding an ACK from master stream fixes the edge case.

Change-Id: If14d0a2a36e8835f7ad3e6b660af19d0ee5d9118

### DIFF
--- a/voltdb/log4j.xml
+++ b/voltdb/log4j.xml
@@ -44,7 +44,7 @@
         <level value="INFO"/>
     </logger -->
 
-    <!-- logger name="HOST">
+    <!--logger name="HOST">
         <level value="INFO"/>
     </logger -->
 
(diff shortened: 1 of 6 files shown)


voltdb/log4j.xml
```diff
@@ -64,7 +64,7 @@
         <level value="INFO"/>
     </logger -->
 
-    <!--logger name="EXPORT">
+    <!-- logger name="EXPORT">
         <level value="INFO"/>
     </logger -->
```
JadenW: drop this

luyangco:
Right, it's pre-existing issue before migration branch.

PR #5531, opened 2018-08-15
ENG-14464 VoltTable get by column name optimization

benjaminballard:
This adds a map to cache the column index for a given column name.

The map is not even created unless you use a column name to retrieve a value, and only caches columns as they are accessed, so this shouldn't add any undue overhead.

### DIFF
--- a/src/frontend/org/voltdb/VoltTable.java
+++ b/src/frontend/org/voltdb/VoltTable.java
@@ -23,6 +23,7 @@ import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.util.Arrays;
+import java.util.HashMap;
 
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/VoltTable.java
```diff
@@ -706,9 +710,20 @@ final String getColumnDefaultValue(int index) {
...
+                return cachedIndex;
+            }
+        } else {
+            m_columnNameIndexMap = new HashMap<String,Integer>(m_colCount);
+        }
+
         assert(verifyTableInvariants());
         for (int i = 0; i < m_colCount; i++) {
             if (getColumnName(i).equalsIgnoreCase(name))
+                m_columnNameIndexMap.put(name, Integer.valueOf(i));
```
yzhang1991: There should be a curly brace pair here for the if too.
Otherwise, it will always return 0.

PR #5532, opened 2018-08-15
 ENG-14449: handle Misrouted Fragment Exception on the MP Site thread

JadenW:


src/frontend/org/voltdb/iv2/MpScheduler.java
```diff
@@ -518,11 +518,9 @@ public void handleFragmentResponseMessage(FragmentResponseMessage message)
         // RTB: Didn't we decide early rollback can do this legitimately.
         if (txn != null) {
             SerializableException ex = message.getException();
-            if (ex != null && ex instanceof TransactionRestartException) {
-                if (((TransactionRestartException)ex).isMisrouted()) {
-                    ((MpTransactionState)txn).restartFragment(message, m_iv2Masters, m_partitionMasters);
-                    return;
-                }
+            if (ex instanceof TransactionRestartException && ((TransactionRestartException)ex).isMisrouted()) {
```
wweiss-voltdb: was the null check not needed?

n3world:
Instanceof returns false if the value is null.

src/frontend/org/voltdb/iv2/MpTransactionState.java
```diff
@@ -472,22 +465,48 @@ private FragmentResponseMessage pollForResponses()
...
+
+                SerializableException se = msg.getException();
+                if (se instanceof TransactionRestartException) {
+                    if (tmLog.isDebugEnabled()) {
+                        tmLog.debug("Transaction exception, txnid: " + TxnEgo.txnIdToString(msg.getTxnId()) + " status:" + msg.getStatusCode()  + " isMisrouted:"+ ((TransactionRestartException) se).isMisrouted()
+                                + " msg: " + msg);
+                    }
+
+                    // If this is a restart exception from the inject poison pill, we don't need to match up the DependencyId
+                    // Don't rely on the restartTimeStamp check since it's not reliable for poison
```
manjujames: I don't understand this comment about poison pill here. Could you please explain?

JadenW: The initial change I had was filtering out the all stale fragmentResponse based on the restartTimeStamp. 
Then I found this could cause we lose track of some dummy FramgmentResponseMessage (i.e. the inject *poison pill* for abort/restart current MP txn) when repairs happens on MPI.(https://github.com/VoltDB/voltdb/blob/master/src/frontend/org/voltdb/iv2/MpTransactionTaskQueue.java#L137) 
The restartTimeStamp in poison is unreliable because it was assigned on the MPI deliver thread during repair. If there are multiple round of repair (each sp leader promotion could trigger a new round of repair), there are no guarantee that the current MP txn has been restart or not before it received the next round of poison. (Mp restart happens on the MP Site thread, and will increment the its restartTimeStamp).
Since we always want to handle the poison and there are no "stale" poison per se, so I intentionally ignore the restartTimeStamp check for the "poison". 
I will update the comment.

PR #5533, opened 2018-08-16
 ENG-14408: Use a thread to wait for timeout 

n3world:
The implementation of using a ScheduledThreadPoolExecutor with a core
pool size of 0 and keepAliveTime is 0 any thread in the pool will do
spin loop until the queue is empty. Instead of using a thread pool which
has no real defined start and end to its lifetime use a thread to track
handle tracking the timeout.

Create DelayedExecutionThread which wait for a delay to pass and then
invokes a runnable. The delayed thread can be canceled while waiting for
the delay to pass but not before.

### DIFF
--- a/tests/frontend/org/voltdb/client/TestConnectionUtilDelayedExecutionThread.java
+++ b/tests/frontend/org/voltdb/client/TestConnectionUtilDelayedExecutionThread.java
@@ -1,0 +1,159 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2018 VoltDB Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/client/TestConnectionUtilDelayedExecutionThread.java
```diff
@@ -0,0 +1,157 @@
...
+
+        cdl1.await();
+        assertTrue("Executed too early", System.nanoTime() - before >= TimeUnit.MILLISECONDS.toNanos(delayMs));
+
+        state = det.state();
+        assertTrue("State not running: " + state, state == State.RUNNING);
+
+        assertEquals(0, cdl1.getCount());
+
+        state = det.state();
```
manjujames: Shouldn't you call cdl2. countDown() before you check the state again?

PR #5534, opened 2018-08-16
ENG-14468 verify buffer length before allocation to handle late arrived responses

zeemanhe:
Client Interface may prematurely send response back to client due to leader changes. Responses could then be routed to client interface which could not locate the transaction and does no know how to allocate buffer.

### DIFF
--- a/src/frontend/org/voltdb/SimpleClientResponseAdapter.java
+++ b/src/frontend/org/voltdb/SimpleClientResponseAdapter.java
@@ -176,7 +176,11 @@ public class SimpleClientResponseAdapter implements Connection, WriteStream {
             // in NIOWriteStream.enqueue().
             ByteBuffer buf = null;
             synchronized(this) {
-                buf = ByteBuffer.allocate(ds.getSerializedSize());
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/SimpleClientResponseAdapter.java
```diff
@@ -176,7 +176,11 @@ public void enqueue(DeferredSerialization ds) {
             // in NIOWriteStream.enqueue().
             ByteBuffer buf = null;
             synchronized(this) {
-                buf = ByteBuffer.allocate(ds.getSerializedSize());
+                final int serializedSize = ds.getSerializedSize();
+                if (serializedSize <= 0) {
```
luyangco: Correct me if I am wrong. Regarding readability I feel comparing with `DeferredSerialization.EMPTY_MESSAGE_LENGTH` is better. Since the only way to return negtive serializedSize is through the early return.

zeemanhe: agree

PR #5538, opened 2018-08-17
ENG-14347, keep mastership designation runnable when give up mastership.

luyangco:
Initialization and catalog update setup new runnable so won't be affected.

The cause of the issue is that when export master give up mastership, it somehow also clears the mastership promotion callback, which is installed only during initialization or catalog update. When the old master was asked to be master again, the promotion will fail because it couldn't find the callback, in turn blocks export buffer to drain.

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -1102,12 +1102,10 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
                         if (exportLog.isDebugEnabled()) {
                             exportLog.debug("Export table " + getTableName() + " accepting mastership for partition " + getPartitionId());
                         }
-                        if (m_onMastership != null) {
-                            if (m_mastershipAccepted.compareAndSet(false, true)) {
-                                // Either get enough responses or have received TRANSFER_MASTER event, clear the response sender HSids.
-                                m_responseHSIds.clear();
-                                m_onMastership.run();
-                            }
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1065,7 +1065,6 @@ public synchronized void unacceptMastership() {
             exportLog.debug("Export table " + getTableName() + " for partition " + getPartitionId() + " mailbox hsid (" +
                     CoreUtils.hsIdToString(m_ackMailboxRefs.get().getFirst().getHSId()) + ") gave up export mastership");
         }
-        m_onMastership = null;
```
n3world: You should also be able to remove the null check in the Runnable in acceptMastership since this was the only location where the runnable could go from set to unset and it was checked at the start of acceptMastership.

luyangco: Good catch. Will do.

luyangco: After a second thought, I think it still worth to keep the null check to guard any attempt to call acceptMastership() before export initialization. But maybe I am just overprotective.

n3world: You can keep the first null check in acceptMastership() but remove the second one in the runnable that is created by acceptMastership().

luyangco: Done, thanks.

PR #5540, opened 2018-08-17
ENG-12549  MPI promotion could fail if more nodes are down before promotion is completed.

zeemanhe:
When a MPI node is down, global service elector will promote a new MPI leader. The promotion may be blocked if the cluster endures more node failures.  MPI promotion could be interrupted while partition leaders are being promoted and will not be completed before all partition leaders are promoted.  MPI promotion and the partition leader promotion callbacks share the same executor service.  Upon more node failures, more rounds of partition leader promotions will be triggered by ZooKeeper.  Unfortunately these followup partition leader promotions are not executed by the executor service since the service is blocked by the MPI promotion.  Thus MPI promotion will never be completed under this circumstance and the cluster hangs. The issue is addressed in this pull request:  the partition leader promotions use their own single thread executor services to avoid being blocked.


zeemanhe:
[Test case](https://github.com/VoltDB/pro/pull/2319)

PR #5545, opened 2018-08-20
ENG-14471 printout thread dumps while a node is programmatically terminated. 

zeemanhe:
A node may be programmatically shutdown.  Under such certain circumstances,   the thread dump of this node won't be collected. The dump may also be of help for the analysis of issues from productions and  system tests.

### DIFF
--- a/src/frontend/org/voltdb/VoltDB.java
+++ b/src/frontend/org/voltdb/VoltDB.java
@@ -22,6 +22,9 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/VoltDB.java
```diff
@@ -1507,6 +1516,16 @@ public static void exit(int status) {
         System.exit(status);
     }
 
+    private static String generateThreadDump() {
+        StringBuilder threadDumps = new StringBuilder();
+        ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+        ThreadInfo[] threadInfos = threadMXBean.getThreadInfo(threadMXBean.getAllThreadIds(), 100);
```
luyangco: Integer.MAX_VALUE could be used to request the entire stack to be dumped.

PR #5548, opened 2018-08-21
Eng 14447 remove action blocker

luyangco:


src/frontend/org/voltdb/iv2/SpInitiator.java
```diff
@@ -196,10 +202,13 @@ private void acceptPromotionImpl(long lastLeaderHSId, boolean migratePartitionLe
         try {
             long startTime = System.currentTimeMillis();
             Boolean success = false;
-            m_term = createTerm(m_messenger.getZK(),
-                    m_partitionId, getInitiatorHSId(), m_initiatorMailbox,
-                    m_whoami);
-            m_term.start();
+            if (!m_hasTerm) {
```
zeemanhe: Can this check be replaced with m_term != null? This new flag may not be needed.

PR #5549, opened 2018-08-21
Bug fix for ENG-14415 core dump when using materialized view to rewrite SELECT query.

kaikaizi:
Added check against LIMIT/OFFSET of SELECT stmt. When SELECT contains LIMIT and/or OFFSET, it skips materialized view matching.


tests/frontend/org/voltdb/planner/TestMVOptimization.java
```diff
@@ -209,8 +218,9 @@ public void testStoredProcedures() {
         // But could match when parameter is somewhere else.
         assertMatch("SELECT distinct a1 distinct_a1, COUNT(b1) count_b1, SUM(a) sum_a, COUNT(*) counts " +
```
yzhang1991: Add one distinct without limit offset.
This can be optimized to use views.

PR #5550, opened 2018-08-21
Fix ENG-14478

yzhang1991:
When we plan an UPDATE statement, the in-byte flags of the target table's VARCHAR columns are copied to the corresponding input VARCHAR columns. This may cause some mistakes when we try to copy the input value out.


src/frontend/org/voltdb/planner/PlanAssembler.java
```diff
@@ -1540,7 +1546,6 @@ private CompiledPlan getNextUpdatePlan() {
             Column col = colEntry.getKey();
             String colName = col.getTypeName();
             AbstractExpression expr = colEntry.getValue();
-            expr.setInBytes(colEntry.getKey().getInbytes());
```
Chaozz: I understand we set the flag with a wrong target. My concern is if we don't set the `in-byte` flag, would it cause some other errors? 
What would happen if we try to update a string which is larger than the byte limit of the target column?

yzhang1991: The in-byte flag of the source and the target table shouldn't mess with each other. Overriding the in-byte flag of the source column with the in-byte flag of the target column is wrong.

Chaozz: I see.

PR #5551, opened 2018-08-22
ENG-14481 - Allow skipping during mp non replicated binary log apply

n3world:
Disabling skip for transactions with TXN_PAR_HASH_SPECIAL caused row
moditications to be applied to partitions which they should not be. Allowing
skip for this transacations allows truncates to be applied to all
partitions but only applies modifications to the correct partition.


manjujames:
These changes are fine as long as we don't support DR protocol version 9 (NO_REPLICATED_STREAM).

Walter says SRT does not work with NO_REPLICATED_STREAM, so we should just not switch to that.

PR #5552, opened 2018-08-22
ENG-14480 stop rejoining node when it sees another node failure

zeemanhe:
If the current node hasn't finished rejoin when another node fails, this rejoining node should be shutdown to prevent locking up of the cluster.  If  the host failure notification could come before mesh determination: that is, the current node does not know it is rejoining, and the shutdown is skipped.  The host failure  processing is handed off to a shared executor which could be blocked by other running tasks. To ensure the accurate and timely rejoining node shutdown, use a dedicated executor instead.


luyangco:
Looks good to me.

PR #5553, opened 2018-08-22
Eng 14166 sqlcov starts w

johnTcrawford:
ENG-14166: added tests of STARTS WITH to SqlCoverage (in the existing advanced-strings and new advanced-starts-with test suites, the latter running only against PostgreSQL); and a new QueryTransformer that converts VoltDB STARTS WITH clauses into equivalent PostgreSQL LIKE clauses; also improved LIKE tests; improved debugging of QueryTransformers, added a new groupReplaceAll (boolean) parameter, tweaked the handling of special characters like '\' and '$', and the handling of group name substitutions; and tweaked start_postgresql.sh to ensure that PostgreSQL uses UTF-8 encoding (which it already was on Jenkins machines, but not on my Mac).

### DIFF
--- a/src/frontend/org/voltdb/NonVoltDBBackend.java
+++ b/src/frontend/org/voltdb/NonVoltDBBackend.java
@@ -184,6 +184,7 @@ public abstract class NonVoltDBBackend {
         private Integer m_minimum = null;
         private List<String> m_groups = new ArrayList<String>();
         private List<String> m_groupReplacementTexts = new ArrayList<String>();
+        private boolean m_groupReplaceAll = false;
         private List<String> m_exclude = new ArrayList<String>();
         private boolean m_debugPrint = false;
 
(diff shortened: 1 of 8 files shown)


src/frontend/org/voltdb/NonVoltDBBackend.java
```diff
@@ -283,6 +284,22 @@ protected QueryTransformer groupReplacementText(String ... text) {
             return this;
         }
 
+        /** Specifies whether or not, when replacing groups within the whole
+         *  match (see "groupReplacementText" above), to replace all instances
+         *  of each group, rather than just the first instance of each group;
+         *  default is <b>false</b>. */
+        protected QueryTransformer groupReplaceAll(boolean groupReplaceAll) {
+            this.m_groupReplaceAll = groupReplaceAll;
```
kaikaizi: `this.` is redundant.

johnTcrawford: Thanks. 'this.' is indeed redundant, but seems clearer to me. As for NonVoltDBBackend.java (& its subclasses), Chris used to say that I probably ought to replace all this Regex stuff with ANTLR (or something similar), which is probably true, but a big job. Perhaps some Innovation Week I'll take a crack at it.

PR #5556, opened 2018-08-23
ENG-14488, fix export poll data source twice issue.

luyangco:
THere are two problems under the hood, first export doesn't set m_pendingContainer properly during export processor shutdown, it may cause data lose in downstream export client. Second, when export master gives up mastership, it no longer needs the m_pollFuture, set to null unconditionally.

Change-Id: Iaf17be7011aba0698da11d756b9c00f5babe8199


zeemanhe:
The changes look good

PR #5557, opened 2018-08-23
ENC-14490 skip truncate on replicated tables

n3world:
Only apply truncate for replicated tables on lowest site. Make sure it is skipped on all other sites. Update truncate case to skip if it is a replicatedTableOperation and if it is a replicatedTable just skip all operations on the non lowest sites.


manjujames:
Looks fine to me. You are still working on getting all the unit tests working for this one and 14481 along with the original truncate ticket, right?

n3world:
Yes. I still would need to fix ENG-14444 to get all the unit tests I wrote to pass. Currently most pass with the two fixes I have already put out of pull request.

PR #5559, opened 2018-08-24
ENG-9600: voltadmin jstack feature

JadenW:
use the same code path for thread dump when crash
Handled locally on InvocationDispatcher for local
and on ForeigHost when remote.


lib/python/voltcli/voltadmin.d/jstack.py
```diff
@@ -0,0 +1,66 @@
...
+    bundles = VOLT.AdminBundle(),
+    description = 'Dumping stacktrace for one or all hosts of a running VoltDB Cluster.',
+    arguments =(
+            VOLT.StringArgument('target_host', 'the target hostname[:port] or address[:port]. (default all hosts)', optional=True)
+    ),
+)
+
+def jstack(runner):
+    # take the jstack using exec @JStack HOST_ID
+    if runner.opts.target_host is None:
```
luyangco: Could you also provide an option to take jstack by host ip address?

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -623,6 +627,50 @@ private final static ClientResponseImpl dispatchGetPartitionKeys(StoredProcedure
...
+                    task.clientHandle);
+        }
+
+        boolean success = true;
+        if (ihid < 0) { // all hosts
+            //collect thread dumps
+            String dumpDir = new File(VoltDB.instance().getVoltDBRootPath(), "thread_dumps").getAbsolutePath();
+            String fileName =  hostMessenger.getHostname() + "_host-" + hostMessenger.getHostId() + "_" + System.currentTimeMillis()+".jstack";
+            success = VoltDB.dumpThreadTraceToFile(dumpDir, fileName );
+            hostMessenger.sendPoisonPill(liveHids, "@Jstack called", ForeignHost.PRINT_STACKTRACE);
```
luyangco: `sendPoisonPill` usually means do something then kill the node/cluster, since here we're not going to kill cluster, please rename the function. Suggestion: `sendNonIv2Message`.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -623,6 +627,50 @@ private final static ClientResponseImpl dispatchGetPartitionKeys(StoredProcedure
...
+            String fileName =  hostMessenger.getHostname() + "_host-" + hostMessenger.getHostId() + "_" + System.currentTimeMillis()+".jstack";
+            success = VoltDB.dumpThreadTraceToFile(dumpDir, fileName );
+        } else { // only remote
+            hostMessenger.sendPoisonPill(ihid, "@Jstack called", ForeignHost.PRINT_STACKTRACE);
+        }
+        if (success) {
+            return new ClientResponseImpl(ClientResponse.SUCCESS, new VoltTable[0], "SUCCESS", task.clientHandle);
+        }
+        return gracefulFailureResponse(
+                "Failed to create the thread dump of localhost.",
```
luyangco: Maybe a remote node, print host id instead of `localhost`.

src/frontend/org/voltdb/InvocationDispatcher.java
zeemanhe: Locally dumped already. The local host should be excluded from **liveHids**

src/frontend/org/voltdb/SystemProcedureCatalog.java
```diff
@@ -148,14 +148,15 @@ Procedure asCatalogProcedure() {
...
         builder.put("@AdHoc_RW_SP",             new Config("org.voltdb.sysprocs.AdHoc_RW_SP",              true,  false, false, 0,    VoltType.VARBINARY, false, false, false, false,     true,   false,            true,         true  ));
         builder.put("@AdHoc_RO_MP",             new Config("org.voltdb.sysprocs.AdHoc_RO_MP",              false, true,  false, 0,    VoltType.INVALID,   false, false, false, true,      false,  false,            true,         true  ));
         builder.put("@MigratePartitionLeader",  new Config("org.voltdb.sysprocs.MigratePartitionLeader",   true,  true,  false, 0,    VoltType.BIGINT,    false, true,  true,  false,     false,  false,            true,         false  ));
         builder.put("@AdHoc_RO_SP",             new Config("org.voltdb.sysprocs.AdHoc_RO_SP",              true,  true,  false, 0,    VoltType.VARBINARY, false, false, false, true,      false,  false,            true,         true  ));
+        builder.put("@JStack",                  new Config(null,                                           false, false, false, 0,    VoltType.INVALID,   false, false, true,  true,      false,  true,             true,         false ));
```
zeemanhe: @JStack should be made to non-transactional.

src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -623,6 +627,50 @@ private final static ClientResponseImpl dispatchGetPartitionKeys(StoredProcedure
...
+            String fileName =  hostMessenger.getHostname() + "_host-" + hostMessenger.getHostId() + "_" + System.currentTimeMillis()+".jstack";
+            success = VoltDB.dumpThreadTraceToFile(dumpDir, fileName );
+            hostMessenger.sendPoisonPill(liveHids, "@Jstack called", ForeignHost.PRINT_STACKTRACE);
+        } else if (ihid == hostMessenger.getHostId()) { // only local
+            //collect thread dumps
+            String dumpDir = new File(VoltDB.instance().getVoltDBRootPath(), "thread_dumps").getAbsolutePath();
+            String fileName =  hostMessenger.getHostname() + "_host-" + hostMessenger.getHostId() + "_" + System.currentTimeMillis()+".jstack";
+            success = VoltDB.dumpThreadTraceToFile(dumpDir, fileName );
+        } else { // only remote
+            hostMessenger.sendPoisonPill(ihid, "@Jstack called", ForeignHost.PRINT_STACKTRACE);
```
zeemanhe: An invalid positive **ihid**  can be passed in  via scripted procedure calls.

lib/python/voltcli/voltadmin.d/jstack.py
rdykiel: It works both with IP or hostnames:
~/test/multinode-eclipse                                                                                                              
▶ voltadmin jstack 10.10.183.245:3023                     
INFO: Taking jstack of host: 10.10.183.245:3023
Status: 1
Information: SUCCESS


~/test/multinode-eclipse                                                                                                              
▶ voltadmin jstack rdykiel-DH270:3022
INFO: Taking jstack of host: rdykiel-DH270:3022
Status: 1
Information: SUCCESS

src/frontend/org/voltdb/InvocationDispatcher.java
rdykiel: There are only 2 callers of this variant of sendPoisonPill. I am loath duplicating code just for this use case. I'd rather keep this unchanged until we have other use cases justifying this change.

src/frontend/org/voltdb/InvocationDispatcher.java
rdykiel: Changed message:

~/test/multinode-eclipse                                                                                                             ⍉
▶ sqlcmd --port=21213                    
SQL Command :: localhost:21213
1> execute @JStack 0;
Failed to create the thread dump of Host Id 0.
2> execute @JStack -1;
Failed to create the thread dump of all hosts.

src/frontend/org/voltdb/InvocationDispatcher.java
zeemanhe: I wonder if this can be made simpler by asking ClientInterface to dump instead of going through host messenger poison pills.

rdykiel: No, this does not generate 2 stack traces for the localhost. This is because somehow sendPoisonPill only sends to foreign hosts. However for the sake of clarity I'll remove the localHost id.

src/frontend/org/voltdb/SystemProcedureCatalog.java
rdykiel: done

src/frontend/org/voltdb/InvocationDispatcher.java
rdykiel: I believe this should be caught before:
        if (ihid >=0 && !liveHids.contains(ihid)) {
            return gracefulFailureResponse(
                    "Invalid Host Id or Host Id not member of cluster: " + ihid,
                    task.clientHandle);
        }

src/frontend/org/voltdb/SystemProcedureCatalog.java
rdykiel: Reverting change, as it introduces NPEs. Leaving as it was originally.

src/frontend/org/voltdb/InvocationDispatcher.java
rdykiel: This would involve more refactoring. Not sure we want to spend this kind of time right now.

luyangco: In my original comment I mean to rename `sendPoisonPill ` to `sendNonIv2Message ` for two reasons,
1) the third parameter can tell the purpose of the message,
2) the function does more than send poison pill.

I am also dislike duplicating entire function to a slightly different variant.

PR #5565, opened 2018-08-27
Eng 13952 test

GuaGuaDean:
When a node rejoining, it will ask the surviving node to save snapshot, which will call @Quiesce SysProc. So we don't need to do this again in MigratePartitionLeader process. Test case attached.

### DIFF
--- a/tests/frontend/org/voltdb/TestExportSPIMigration.java
+++ b/tests/frontend/org/voltdb/TestExportSPIMigration.java
@@ -48,7 +48,6 @@ import org.junit.Test;
 import org.voltdb.client.Client;
 import org.voltdb.client.ClientConfig;
 import org.voltdb.client.ClientFactory;
-import org.voltdb.client.NoConnectionsException;
 import org.voltdb.compiler.VoltProjectBuilder;
 import org.voltdb.export.ExportDataProcessor;
 import org.voltdb.regressionsuites.JUnit4LocalClusterTest;
@@ -57,11 +56,11 @@ import org.voltdb.utils.VoltFile;
 
 import au.com.bytecode.opencsv_voltpatches.CSVParser;
 
-public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
+public class TestExportSPIMigration extends JUnit4LocalClusterTest
 {
     private ServerListener m_serverSocket;
-    private final List<ClientConnectionHandler> m_clients = Collections.synchronizedList(new ArrayList<ClientConnectionHandler>());
-    private final ConcurrentMap<Long, AtomicLong> m_seenIds = new ConcurrentHashMap<Long, AtomicLong>();
+    private final List<ClientConnectionHandler> clients = Collections.synchronizedList(new ArrayList<ClientConnectionHandler>());
+    private final ConcurrentMap<Long, AtomicLong> seenIds = new ConcurrentHashMap<Long, AtomicLong>();
     private static final String SCHEMA = "CREATE STREAM t partition on column a (a integer not null, b integer not null);";
     private static int PORT = 5001;
     private volatile Set<String> exportMessageSet = null;
@@ -73,7 +72,7 @@ public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
     }
 
     @Test
-    public void testFlushEEBufferWithoutQuiesce() throws Exception {
+    public void testFlushEEBufferWhenRejoin() throws Exception {
         resetDir();
         m_serverSocket = new ServerListener(5001);
         m_serverSocket.start();
@@ -126,18 +125,15 @@ public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
             VoltTable vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
             System.out.println("TOPO at the start:");
             System.out.println(vt.toFormattedString());
-            assertEquals("0:0", vt.fetchRow(0).get(2, VoltType.STRING));
-            assertEquals("1:1", vt.fetchRow(1).get(2, VoltType.STRING));
+            assertTrue(!vt.fetchRow(0).getString(2).split(":")[0].equals(vt.fetchRow(1).getString(2).split(":")[0]));
 
             cluster.killSingleHost(1);
             vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
             System.out.println("TOPO after kill:");
             System.out.println(vt.toFormattedString());
-            assertEquals("0:0", vt.fetchRow(0).get(2, VoltType.STRING));
-            assertEquals("0:1", vt.fetchRow(1).get(2, VoltType.STRING));
+            assertTrue(vt.fetchRow(0).getString(2).split(":")[0].equals(vt.fetchRow(1).getString(2).split(":")[0]));
 
             for (int i = 0; i < 5; i++) {
-                Thread.sleep(50);
                 //add data to stream table
                 client.callProcedure("@AdHoc", "insert into t values(" + i + ", 1)");
             }
@@ -146,33 +142,40 @@ public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
             long tss = System.currentTimeMillis();
             while (true) {
                 Thread.sleep(100);
-                // rejoin time limit (20s) reached, test marked as failed
-                if (System.currentTimeMillis() - tss > 20000) {
-                    assertTrue(false);
-                }
-                // rejoin has finished and partition leader migrated
-                else {
-                    vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
-                    if ("2:0".equals(vt.fetchRow(0).get(2, VoltType.STRING)) && "0:1".equals(vt.fetchRow(1).get(2, VoltType.STRING))) {
-                        break;
-                    }
+                assertTrue("Rejoin time has reached 20s limit, test failed", System.currentTimeMillis() - tss < 20000);
+                // rejoin has finished and partition leader balanced after migration
+                vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
+                if (!vt.fetchRow(0).getString(2).split(":")[0].equals(vt.fetchRow(1).getString(2).split(":")[0])) {
+                    break;
                 }
             }
 
             // all 5 export messages should be captured
             // because during rejoin a @Quiesce call will be triggered by @Snapshotsave
             assertEquals(exportMessageSet.size(), 5);
+
             vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
             System.out.println("TOPO in the end:");
             System.out.println(vt.toFormattedString());
-            assertEquals("2:0", vt.fetchRow(0).get(2, VoltType.STRING));
-            assertEquals("0:1", vt.fetchRow(1).get(2, VoltType.STRING));
+
+            for (int i = 5; i < 10; i++) {
+                client.callProcedure("@AdHoc", "insert into t values(" + i + ", 1)");
+            }
+            client.callProcedure("@Quiesce");
+            while (true) {
+                Thread.sleep(100);
+                assertTrue("Insertion time has reached 10s limit, test failed", System.currentTimeMillis() - tss < 10000);
+                // making sure all export message
+                if (exportMessageSet.size() == 10) {
+                    break;
+                }
+            }
         } catch (Exception e) {
             e.printStackTrace();
         } finally {
             m_serverSocket.close();
             cleanup(client, cluster);
-            m_clients.clear();
+            clients.clear();
             System.out.println("Everything shut down.");
         }
     }
@@ -194,18 +197,16 @@ public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
     }
 
     private void cleanup(Client client, LocalCluster cluster) {
-        if ( client != null) {
-            try {
-                client.drain();
-                client.close();
-            } catch (InterruptedException | NoConnectionsException e) {
-            }
-        }
         if (cluster != null) {
             try {
                 cluster.shutDown();
             } catch (InterruptedException e) {
-                //e.printStackTrace();
+            }
+        }
+        if ( client != null) {
+            try {
+                client.close();
+            } catch (InterruptedException e) {
             }
         }
     }
@@ -234,7 +235,7 @@ public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
                 try {
                     Socket clientSocket = ssocket.accept();
                     ClientConnectionHandler ch = new ClientConnectionHandler(clientSocket);
-                    m_clients.add(ch);
+                    clients.add(ch);
                     ch.start();
                 } catch (IOException ex) {
                     //ex.printStackTrace();
@@ -272,8 +273,8 @@ public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
                             continue;
                         }
                         Long i = Long.parseLong(parts[0]);
-                        if (m_seenIds.putIfAbsent(i, new AtomicLong(1)) != null) {
-                            m_seenIds.get(i).incrementAndGet();
+                        if (seenIds.putIfAbsent(i, new AtomicLong(1)) != null) {
+                            seenIds.get(i).incrementAndGet();
                         }
                     }
                     m_clientSocket.close();
--- a/tests/frontend/org/voltdb/TestExportSPIMigration.java
+++ b/tests/frontend/org/voltdb/TestExportSPIMigration.java
@@ -288,12 +288,6 @@ public class TestExportSPIMigration extends JUnit4LocalClusterTest
         }
     }
 
-    private class TopoMapEntry {
-        int m_partitionKey;
-        ArrayList<HostSiteId> m_hsIds = new ArrayList<>();
-        HostSiteId m_hsLeader;
-    }
-
     private static class HostSiteId implements Comparable<HostSiteId> {
         int m_hostId;
         int m_siteId;
--- a/tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
+++ b/tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
@@ -25,7 +25,6 @@ package org.voltdb;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
-import org.voltcore.logging.VoltLogger;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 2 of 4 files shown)


tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+        File f = new File("/tmp/" + System.getProperty("user.name"));
+        f.mkdirs();
+        super.setUp();
+        m_serverSocket = new ServerListener(5001);
+        m_serverSocket.start();
+        //PeriodicTopoCheck check = new PeriodicTopoCheck();
+        //check.start();
+    }
+
+    @Test
```
n3world: The @Test notation is junit4 the TestExportBase extends TestCase which is junit3. You really shouldn't mix which different junit semantics because you could confuse the runner and it might use the wrong one.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+        cluster.setJavaProperty("DISABLE_MIGRATE_PARTITION_LEADER", "true");
+        boolean compile = cluster.compile(project);
+        assertTrue(compile);
+        builder.addServerConfig(cluster);
+        return builder;
+    }
+
+    class ServerListener extends Thread {
+        private ServerSocket ssocket;
+        private boolean shuttingDown = false;
```
n3world: If this variable is going to be accessed without synchronization between multiple threads it should be volatile.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+    public TestFlushEEBufferWhenExport(String name) {
+        super(name);
+    }
+
+    @Override
+    public void setUp() throws Exception {
+        m_username = "default";
+        m_password = "password";
+        VoltFile.recursivelyDelete(new File("/tmp/" + System.getProperty("user.name")));
+        File f = new File("/tmp/" + System.getProperty("user.name"));
```
n3world: Why not define this before the call to recursivelyDelete and pass f into recursivelyDelete?

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+                            break;
+                        }
+                        if (line != null) {
+                            set.add(line);
+                        }
+                        String parts[] = m_parser.parseLine(line);
+                        if (parts == null) {
+                            continue;
+                        }
+                        Long i = Long.parseLong(parts[0]);
```
n3world: It looks like `parseLine` can return an empty array when `line == null` which means this could cause an ArrayIndexOutOfBoundsException.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+
+public class TestFlushEEBufferWhenExport extends TestExportBase
+{
+    private ServerListener m_serverSocket;
+    private final List<ClientConnectionHandler> m_clients = Collections.synchronizedList(new ArrayList<ClientConnectionHandler>());
+    private final ConcurrentMap<Long, AtomicLong> m_seenIds = new ConcurrentHashMap<Long, AtomicLong>();
+    private Map<Integer, TopoMapEntry> m_topoMap = new HashMap<Integer, TopoMapEntry>();
+    private static final String SCHEMA =
+            "CREATE STREAM t partition on column a (a integer not null, b integer not null);";
+    private static int m_portCount = 5001;
```
n3world: Usually the name of static variables is all upper case like the above SCHEMA.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+                        if (line != null) {
+                            set.add(line);
+                        }
+                        String parts[] = m_parser.parseLine(line);
+                        if (parts == null) {
+                            continue;
+                        }
+                        Long i = Long.parseLong(parts[0]);
+                        if (m_seenIds.putIfAbsent(i, new AtomicLong(1)) != null) {
+                            synchronized(m_seenIds) {
```
n3world: It is unusual to synchronize on a concurrent map.
I think this would be better as:
m_seenIds.computeIfAbsent(i, () -> new AtomicLong()).getAndIncrement();

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+    private ServerListener m_serverSocket;
+    private final List<ClientConnectionHandler> m_clients = Collections.synchronizedList(new ArrayList<ClientConnectionHandler>());
+    private final ConcurrentMap<Long, AtomicLong> m_seenIds = new ConcurrentHashMap<Long, AtomicLong>();
+    private Map<Integer, TopoMapEntry> m_topoMap = new HashMap<Integer, TopoMapEntry>();
+    private static final String SCHEMA =
+            "CREATE STREAM t partition on column a (a integer not null, b integer not null);";
+    private static int m_portCount = 5001;
+    private static LocalCluster cluster = null;
+    private static VoltProjectBuilder project = null;
+    private static Set<String> set = null;
```
n3world: `set` is a very ambiguous name for a set.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+        if (cluster != null) {
+            try {
+                cluster.shutDown();
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+    static public junit.framework.Test suite() throws Exception {
```
n3world: It seems strange that the `suite` method is doing so much configuration/setup.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated. I was referring to [TestExportBaseSocketExport](https://github.com/VoltDB/voltdb/blob/master/tests/frontend/org/voltdb/export/TestExportBaseSocketExport.java) when writing this test. So I am wondering if we should do the same update on that one, too.

manjujames: I think all caps is for final variables and for statics that are not final is s_<name>? Not sure if we have this defined in our standards doc.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+    public void setUp() throws Exception {
+        m_username = "default";
+        m_password = "password";
+        VoltFile.recursivelyDelete(new File("/tmp/" + System.getProperty("user.name")));
+        File f = new File("/tmp/" + System.getProperty("user.name"));
+        f.mkdirs();
+        super.setUp();
+        m_serverSocket = new ServerListener(5001);
+        m_serverSocket.start();
+        //PeriodicTopoCheck check = new PeriodicTopoCheck();
```
manjujames: Remove the commented out code, if we don't need it.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+                        cluster.rejoinOne(1);
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    }
+                }
+            });
+            insertThread.start();
+            rejoinThread.start();
+
+            Thread.sleep(6000);
```
manjujames: Is this sleep of 6 seconds for SPI migration to complete?

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+            m_serverSocket.close();
+            m_clients.clear();
+            cleanup(client, cluster);
+        }
+    }
+
+    public static void wireupExportTableToSocketExport(String tableName) {
+        String streamName = tableName;
+        Map<String, Integer> m_portForTable = new HashMap<String, Integer>();
+        if (!m_portForTable.containsKey(streamName)) {
```
manjujames: Isn't this Map created in the line right above? So it will not have any key, correct? Why this 'if' check?

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+            if (m_clients.size() > 0) m_clients.get(0).stopClient();
+            m_serverSocket.close();
+            m_clients.clear();
+            cleanup(client, cluster);
+        }
+    }
+
+    public static void wireupExportTableToSocketExport(String tableName) {
+        String streamName = tableName;
+        Map<String, Integer> m_portForTable = new HashMap<String, Integer>();
```
manjujames: Local variables should not have m_ in its name.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,430 @@
...
+        if (!m_portForTable.containsKey(streamName)) {
+            m_portForTable.put(streamName, getNextPort());
+        }
+        Properties props = new Properties();
+        boolean m_isExportReplicated = false;
```
manjujames: This is also a local variable. So, it should just be isExportReplicated, no m_

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated, now wrapped with JUnit4LocalCluster as test base

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Now sleep() removed from test

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated, thx

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: In this case since multi flag is set to false so pending will be null, thus parseLine will only return null when line is null. Still I added an addition condition for array.length == 0.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: I tried to use computeIfAbsent, but with syntax complaints. Can I just remove synchronized?

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+            client.createConnection(cluster.getListenerAddress(0));
+
+            // create a table to enable rejoin
+            client.callProcedure("@AdHoc", "create table test (a int);");
+            client.callProcedure("@AdHoc", "insert into test values(1)");
+
+            VoltTable vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
+            System.out.println("TOPO at the start:");
+            System.out.println(vt.toFormattedString());
+            assertEquals("0:0", vt.fetchRow(0).get(2, VoltType.STRING));
```
manjujames: This assertion is assuming that node 0 will always have partition 0 and node 1 will always have portion 1. This might be OK for now, but if we make changes to the placement algorithm, this test would break. These assertions are not required in this test since this is not a test for placement algorithm, right?

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+
+            cluster.killSingleHost(1);
+            vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
+            System.out.println("TOPO after kill:");
+            System.out.println(vt.toFormattedString());
+            assertEquals("0:0", vt.fetchRow(0).get(2, VoltType.STRING));
+            assertEquals("0:1", vt.fetchRow(1).get(2, VoltType.STRING));
+
+            for (int i = 0; i < 5; i++) {
+                Thread.sleep(50);
```
manjujames: Why do we need this sleep here?

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+            while (true) {
+                Thread.sleep(100);
+                // rejoin time limit (20s) reached, test marked as failed
+                if (System.currentTimeMillis() - tss > 20000) {
+                    assertTrue(false);
+                }
+                // rejoin has finished and partition leader migrated
+                else {
+                    vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
+                    if ("2:0".equals(vt.fetchRow(0).get(2, VoltType.STRING)) && "0:1".equals(vt.fetchRow(1).get(2, VoltType.STRING))) {
```
manjujames: This again assumes that partition 0 will be migrated to the rejoined node, which again might be OK for now. Better might be to verify that the partitions are balanced, which is done is TestSPIMigration written by Zeeman.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+                }
+            }
+
+            // all 5 export messages should be captured
+            // because during rejoin a @Quiesce call will be triggered by @Snapshotsave
+            assertEquals(exportMessageSet.size(), 5);
+            vt = client.callProcedure("@Statistics", "TOPO").getResults()[0];
+            System.out.println("TOPO in the end:");
+            System.out.println(vt.toFormattedString());
+            assertEquals("2:0", vt.fetchRow(0).get(2, VoltType.STRING));
```
manjujames: Same comment about these asserts here.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+            assertEquals("0:1", vt.fetchRow(1).get(2, VoltType.STRING));
+        } catch (Exception e) {
```
manjujames: At this point you could insert more rows into the stream and do a quiesce and verify that all the rows are seen. This will verify that the new export master is sending export data correctly.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+import org.voltdb.client.NoConnectionsException;
+import org.voltdb.compiler.VoltProjectBuilder;
+import org.voltdb.export.ExportDataProcessor;
+import org.voltdb.regressionsuites.JUnit4LocalClusterTest;
+import org.voltdb.regressionsuites.LocalCluster;
+import org.voltdb.utils.VoltFile;
+
+import au.com.bytecode.opencsv_voltpatches.CSVParser;
+
+public class TestFlushExportBufferWhenRejoin extends JUnit4LocalClusterTest
```
manjujames: May be make the test name more generic (for example, TestExportSPIMigration), so that we can add more export related SPI migration tests here as needed.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
```diff
@@ -0,0 +1,335 @@
...
+{
+    private ServerListener m_serverSocket;
+    private final List<ClientConnectionHandler> m_clients = Collections.synchronizedList(new ArrayList<ClientConnectionHandler>());
+    private final ConcurrentMap<Long, AtomicLong> m_seenIds = new ConcurrentHashMap<Long, AtomicLong>();
+    private static final String SCHEMA = "CREATE STREAM t partition on column a (a integer not null, b integer not null);";
+    private static int PORT = 5001;
+    private volatile Set<String> exportMessageSet = null;
```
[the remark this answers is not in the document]
GuaGuaDean: Updated. Thx

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: My intention was to check if partition leader is balanced in the cluster, so that we can later make sure that rejoin is finished. Figured out a new way that won't involve hard coded placement.

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Updated

tests/frontend/org/voltdb/TestFlushExportBufferWhenRejoin.java
GuaGuaDean: Sure, updated

tests/frontend/org/voltdb/TestExportSPIMigration.java
```diff
@@ -0,0 +1,336 @@
...
+            } catch (IOException ioe) {
+            }
+        }
+
+        public void stopClient() {
+            m_closed = true;
+        }
+    }
+
+    private class TopoMapEntry {
```
manjujames: Is this used anywhere?

GuaGuaDean: Yes, you are right. This class is not used any more. Updated.

PR #5568, opened 2018-08-28
ENG-14504:

wweiss-voltdb:
We don't need to wait for the pico network to go away before it is safe to start the repair because there is a failedhost check in the message deliver path.

### DIFF
--- a/src/frontend/org/voltcore/messaging/HostMessenger.java
+++ b/src/frontend/org/voltcore/messaging/HostMessenger.java
@@ -879,7 +879,8 @@ public class HostMessenger implements SocketJoiner.JoinHandler, InterfaceToMesse
     }
 
     public synchronized boolean canCompleteRepair(int hostId) {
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -879,7 +879,8 @@ public synchronized void piconetworkThreadShutdown(int hostId, ForeignHost fh) {
     }
 
     public synchronized boolean canCompleteRepair(int hostId) {
-        return !m_foreignHosts.containsKey(hostId) && !m_picoZombieForeignHosts.containsKey(hostId) && !m_zkZombieForeignHosts.containsKey(hostId);
+        return !m_foreignHosts.containsKey(hostId) && (m_knownFailedHosts.containsKey(hostId) ||
+                (!m_picoZombieForeignHosts.containsKey(hostId) && !m_zkZombieForeignHosts.containsKey(hostId)));
```
zeemanhe: is check m_knownFailedHosts.containsKey(hostId) needed? Look like failed hosts are added to m_knownFailedHosts and then removed from m_foreignHosts  so   m_knownFailedHosts.containsKey(hostId) == !m_foreignHosts.containsKey(hostId)

PR #5573, opened 2018-08-28
ENG-13873-system property for sql execution to use large mode

Chaozz:
- `LARGE_TABLE_RATIO` is a system property that will force execution of SQL queries to use the "large" path (for read-only queries) a certain percentage of the time. It can be used for any tests that are invoked using ant, or for system tests. 
- The default value is 0, means no query will use large mode. If set it to 1, all queries will use large mode.

usage
------
It can be set either from system property:
```bash
export VOLTDB_OPTS="-DLARGE_TABLE_RATIO=0.5"
```
or from ant
```bash
# run in large mode 75% of the time:
ant -Dlarge_mode_ratio=0.75 ...
```

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -1620,6 +1620,13 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
             m_configLogger.start();
 
             scheduleDailyLoggingWorkInNextCheckTime();
+
+            // Write a message to the log file if LARGE_MODE_RATIO system property is set to anything other than 0.
+            // This way it will be possible to verify that various frameworks are exercising large mode.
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1620,6 +1620,13 @@ public Boolean get() {
             m_configLogger.start();
 
             scheduleDailyLoggingWorkInNextCheckTime();
+
+            // Write a message to the log file if LARGE_MODE_RATIO system property is set to anything other than 0.
+            // This way it will be possible to verify that various frameworks are exercising large mode.
+            final double largeModeRatio = Double.valueOf(System.getenv("LARGE_MODE_RATIO") == null ? System.getProperty("LARGE_MODE_RATIO", "0") : System.getenv("LARGE_MODE_RATIO"));
```
yzhang1991: Add a comment here.
If -Dlarge_mode_ratio=xx is specified via ant, the value will show up in the environment variables and take higher priority. Otherwise, the value specified via VOLTDB_OPTS will take effect.

PR #5576, opened 2018-08-30
Eng 14519 sync rejoin finish across sites

wweiss-voltdb:


src/frontend/org/voltdb/iv2/RejoinProducer.java
```diff
@@ -315,11 +324,20 @@ public void runForRejoin(SiteProcedureConnection siteConnection,
                 }
                 m_rejoinSiteProcessor.close();
 
+                boolean allSitesFinishStreaming;
                 if (m_streamSnapshotMb != null) {
                     VoltDB.instance().getHostMessenger().removeMailbox(m_streamSnapshotMb.getHSId());
+                    m_streamSnapshotMb = null;
+                    allSitesFinishStreaming = s_streamingSiteCount.decrementAndGet() == 0;
```
luyangco: Since this line is the only place to decrement the counter, I don't think following else block is needed because `pendingSites` would never go to 0.

src/frontend/org/voltdb/iv2/RejoinProducer.java
```diff
@@ -315,11 +324,20 @@ public void runForRejoin(SiteProcedureConnection siteConnection,
...
+                }
+                else {
+                    int pendingSites = s_streamingSiteCount.get();
+                    assert(pendingSites >= 0);
+                    allSitesFinishStreaming = pendingSites == 0;
+                }
+                if (allSitesFinishStreaming) {
+                    doFinishingTask(siteConnection);
```
luyangco: I wonder this patch won't work because only the last site that finishes receiving snapshot would call `doFinishingTask()`, other sites will skip `doFinishingTask()`.

PR #5579, opened 2018-08-31
ENG-14511 add a blocker to ZooKeeper before partition callbacks are registered.

zeemanhe:
also block  uac during promotion and repair

### DIFF
--- a/src/frontend/org/voltdb/RealVoltDB.java
+++ b/src/frontend/org/voltdb/RealVoltDB.java
@@ -1667,6 +1667,11 @@ public class RealVoltDB implements VoltDBInterface, RestoreAgent.Callback, HostM
             @Override
             public void run()
             {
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1667,6 +1667,11 @@ public void hostsFailed(Set<Integer> failedHosts) {
             @Override
             public void run()
             {
+                //create a blocker for repair if this is a MP leader and partition leaders change
+                if (m_leaderAppointer.isLeader() && m_cartographer.hasPartitionMastersOnHosts(failedHosts)) {
+                    VoltZK.createMpRepairBlocker(m_messenger.getZK(), hostLog);
```
luyangco: Unless there are performance considerations, otherwise I feel like it's better to call with `createActionBlocker`.

PR #5590, opened 2018-09-06
ENG-14541:

wweiss-voltdb:
After MpPromoteAlgo collects repair logs, it identifies the incomplete, non-restartable sysprocs and generates a rollback message to all partitions to clear out the transaction. However, it does not assign the truncation handle to the completion which can cause a random value to be in the truncation handle (in some cases). If the random value happens to be large enough, the partitions can truncation in progress transactions from the repair log ultimately causing mp deadlocks. To ensure that these rollback completions never alter the repair log, the truncation handle must be explicitly assigned to Long.MIN_VALUE.

### DIFF
--- a/src/frontend/org/voltdb/iv2/MpPromoteAlgo.java
+++ b/src/frontend/org/voltdb/iv2/MpPromoteAlgo.java
@@ -391,6 +391,7 @@ public class MpPromoteAlgo implements RepairAlgo
                             true,
                             false);
                     rollback.setTimestamp(m_restartSeqGenerator.getNextSeqNum());
+                    rollback.setTruncationHandle(Long.MIN_VALUE);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/iv2/MpPromoteAlgo.java
```diff
@@ -391,6 +391,7 @@ VoltMessage createRepairMessage(Iv2RepairLogResponseMessage msg)
                             true,
                             false);
                     rollback.setTimestamp(m_restartSeqGenerator.getNextSeqNum());
+                    rollback.setTruncationHandle(Long.MIN_VALUE);
```
luyangco: Truncation handle is defined in TransactionInfoBaseMessage.java as 
```Java
private long m_truncationHandle;
 ```
In my understanding Uninitialized int/long variable in Java is initialized to 0, do you know why the truncation handle in rollback message is a very large value?

PR #5594, opened 2018-09-06
Eng 14539 json api v2.0 http only

prosegay:
give it a check, but I'm waiting for tests, and I need to merge this for a kit that is being cut tonight, so don't merge, thanks for the help, I think this worked out nicely.


prosegay:
ignore this pr

PR #5595, opened 2018-09-06
ENG-14539 JSON API V2.0 for HTTP requests only

prosegay:
give it a check, but I'm waiting for tests, and I need to merge this for a kit that is being cut tonight, so don't merge, thanks for the help, I think this worked out nicely.

### DIFF
--- a/src/frontend/org/voltdb/VoltTable.java
+++ b/src/frontend/org/voltdb/VoltTable.java
@@ -1597,6 +1597,34 @@ public final class VoltTable extends VoltTableRow implements JSONString {
     }
 
     /**
+     * Get a JSON /api/2.0/ representation of this table.
+     * @return A string containing a JSON representation of this table.
+     */
+    public JSONStringer toJSONStringerV2(JSONStringer js) {
+        try {
+            // array of row data
+            VoltTableRow row = cloneRow();
+            row.resetRowPosition();
+            js.array();
+            while (row.advanceRow()) {
+                js.object();
+                for (int i = 0; i < getColumnCount(); i++) {
+
+                    js.key(getColumnName(i));
+                    row.putJSONRep(i, js);
+                }
+                js.endObject();
+            }
+            js.endArray();
+        }
+        catch (JSONException e) {
+            e.printStackTrace();
+            throw new RuntimeException("Failed to serialized a table to JSON V2.", e);
+        }
+        return js;
+    }
+
+    /**
      * Construct a table from a JSON string. Only parses VoltDB VoltTable JSON format.
      *
      * @param json String containing JSON-formatted table data.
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/utils/HTTPAdminListener.java
```diff
@@ -279,6 +279,7 @@ public HTTPAdminListener(
             //Following are the servelets jetty is configured with see URL pattern for what they handle.
             servlets.addServletWithMapping(DBMonitorServlet.class, "/").setAsyncSupported(true);
             servlets.addServletWithMapping(ApiRequestServlet.class, "/api/1.0/*").setAsyncSupported(true);
+            servlets.addServletWithMapping(ApiRequestServlet.class, "/api/2.0/*").setAsyncSupported(true);
```
n3world: You know that the version is 2.0 here you could have a ApiRequestServletV2 which does the v2 response and then you can remove the request from JSONProcCallback

src/frontend/org/voltdb/VoltTable.java
```diff
@@ -1596,6 +1596,34 @@ public String toJSONString() {
...
+                for (int i = 0; i < getColumnCount(); i++) {
+
+                    js.key(getColumnName(i));
+                    row.putJSONRep(i, js);
+                }
+                js.endObject();
+            }
+            js.endArray();
+        }
+        catch (JSONException e) {
```
n3world: I would just throw this exception from this method and let the caller handle it so there is one catch for this.

src/frontend/org/voltdb/utils/ClientResponseToJsonApiV2.java
```diff
@@ -0,0 +1,46 @@
...
+            for (int i=0; i<results.length; i++) {
+                js.key(String.valueOf(i));
+                VoltTable o = results[i];
+                o.toJSONStringerV2(js);
+            }
+            js.endObject();
+            js.endObject();
+        }
+        catch (JSONException e) {
+            e.printStackTrace();
```
n3world: Printing the exception to stdout is not ideal I would remove this since you are throwing an exception with the original exception as the cause on the next line.

src/frontend/org/voltdb/utils/ClientResponseToJsonApiV2.java
```diff
@@ -0,0 +1,46 @@
...
+import org.voltdb.client.ClientResponse;
+
+public class ClientResponseToJsonApiV2 {
+    /* Takes a ClientResonse and builds a api/v2.0 string
+    Currently, this is only used in HTTPClientInterface. If we want to version the response generally, it will
+    require extensive refactoring. This one-off bypasses the implementations of toJSONString()
+     */
+
+    // JSON KEYS FOR SERIALIZATION
+    static final String JSON_STATUS_KEY = "status";
```
n3world: These variables are already defined in ClientResponseImpl and it would be better not redefine them.

PR #5607, opened 2018-09-11
ENG-14547 remove mp repair blocker in the last round of repair

zeemanhe:
Provide MpRepairTask with partition leader promotion status. If all the partitions have been promoted, then the MpRepairTask is the last repair task in this round of repair process. Then after this MpRepairTask is completed, remove mp repair blocker.

### DIFF
--- a/src/frontend/org/voltdb/VoltZK.java
+++ b/src/frontend/org/voltdb/VoltZK.java
@@ -42,7 +42,6 @@ import org.voltcore.zk.ZooKeeperLock;
 import org.voltdb.iv2.LeaderCache;
 import org.voltdb.iv2.LeaderCache.LeaderCallBackInfo;
 import org.voltdb.iv2.MigratePartitionLeaderInfo;
-import org.voltdb.iv2.MpInitiator;
 
 /**
  * VoltZK provides constants for all voltdb-registered
@@ -458,7 +457,7 @@ public class VoltZK {
                 } else if (blockers.contains(migrate_partition_leader)){
                     errorMsg = "while leader migration is active. Please retry node rejoin later.";
                 } else if (isRepairProgress(zk)){
-                    errorMsg = "while leader promotion or transaction repair are in progress. Please retry node rejoin later.";
+                    errorMsg = "while leader promotions or transaction repairs are in progress. Please retry node rejoin later.";
                 }
                 break;
             case elasticJoinInProgress:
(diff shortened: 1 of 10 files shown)


src/frontend/org/voltdb/VoltZK.java
```diff
@@ -445,7 +444,10 @@ private static String setActionBlocker(ZooKeeper zk, String node, CreateMode mod
             case catalogUpdateInProgress:
                 if (blockers.contains(leafNodeRejoinInProgress)) {
                     errorMsg = "while a node rejoin is active. Please retry catalog update later.";
-                } else if (isRepairProgress(zk)) {
+                } else if (blockers.contains(mpRepairBlocker)){
+                    // Upon node failures, a MP repair blocker may be registered right before they
+                    // unregistered after repair is done. Let rejoining nodes wait to avoid any
```
luyangco: Comment update: Let rejoining nodes wait -> Let schema change wait.
Another question, I understand rejoin would interfere transaction repair as new sites are added to replica list, but would catalog update affect repair also? Beyond the nature of changing live schema, catalog update is just a common MP transaction.

zeemanhe: UAC will trigger global service registration/startup, which may step each other with MPI and other promotions. So UAC is more than a normal MP transaction.  Specifically UAC and MPI/or other promotions may compete for object monitor on GlobalServiceElector, resulting in a deadlock.

src/frontend/org/voltdb/rejoin/Iv2RejoinCoordinator.java
```diff
@@ -175,11 +174,7 @@ public static void acquireLock(HostMessenger messenger )
 
             if (elapsed % 10 == 5) {
                 // log the info message every 10 seconds, log the initial message under 5 seconds
-                REJOINLOG.info(String.format("Rejoin node is waiting for catalog update or elastic join to finish, "
-                        + "time elapsed " + elapsed + " seconds"));
-                if (REJOINLOG.isDebugEnabled()) {
-                    REJOINLOG.debug("Rejoin fails to acquire lock:" + blockerError);
-                }
+                REJOINLOG.info("Rejoin node is waiting " + blockerError + "time elapsed " + elapsed + " seconds");
```
luyangco: Need a space between blockerError and "time elapsed"

zeemanhe: updated

PR #5608, opened 2018-09-12
ENG-14547 block rejoin or uac during tx repair

zeemanhe:
Provide MpRepairTask with partition leader promotion status. If all the partitions have been promoted, then the MpRepairTask is the last repair task in this round of repair process. Then after this MpRepairTask is completed, remove mp repair blocker.


zeemanhe:
[forward ported from 8.3](https://github.com/VoltDB/voltdb/pull/5607)

PR #5611, opened 2018-09-12
ENG-14563:

wweiss-voltdb:
Don't disable the truncation handle in all messages generated by the mp repair / mp promote algorithms. Inadvertently advancing the handles can remove repair messages that are needed by partitions that have not completed repair yet.

Also don't all MP transactions that have no fragments for MP transactions from advancing the MP repair truncation point. Otherwise we may remove completions from some parititons that have not been seen at all by other partitions.

### DIFF
--- a/src/frontend/org/voltdb/iv2/MpScheduler.java
+++ b/src/frontend/org/voltdb/iv2/MpScheduler.java
@@ -161,7 +161,7 @@ public class MpScheduler extends Scheduler
             VoltMessage resp = counter.getLastResponse();
             if (resp != null && resp instanceof InitiateResponseMessage) {
                 InitiateResponseMessage msg = (InitiateResponseMessage)resp;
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/iv2/MpScheduler.java
```diff
@@ -475,8 +476,8 @@ else if (result == DuplicateCounter.MISMATCH) {
             // doing duplicate suppresion: all done.
         }
         else {
-            // Only advance the truncation point on committed transactions.
-            if (message.shouldCommit()) {
+            // Only advance the truncation point on committed transactions that sent fragments to SPIs.
+            if (message.shouldCommit() && message.haveSentMpFragment()) {
```
luyangco: What about line 164?  When a node fails, we may also advance the truncation handle, it should also be protected by `haveSentMpFragment()`

PR #5612, opened 2018-09-12
ENG-14563:

wweiss-voltdb:
Don't disable the truncation handle in all messages generated by the mp repair / mp promote algorithms. Inadvertently advancing the handles can remove repair messages that are needed by partitions that have not completed repair yet.

Also don't all MP transactions that have no fragments for MP transactions from advancing the MP repair truncation point. Otherwise we may remove completions from some parititons that have not been seen at all by other partitions.

### DIFF
--- a/src/frontend/org/voltdb/iv2/MpPromoteAlgo.java
+++ b/src/frontend/org/voltdb/iv2/MpPromoteAlgo.java
@@ -25,6 +25,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeSet;
 import java.util.concurrent.Future;
+
+import org.voltcore.messaging.TransactionInfoBaseMessage;
 import org.voltcore.messaging.VoltMessage;
 import org.voltcore.utils.CoreUtils;
 import org.voltcore.utils.Pair;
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/iv2/MpPromoteAlgo.java
```diff
@@ -352,7 +355,9 @@ VoltMessage createRepairMessage(Iv2RepairLogResponseMessage msg)
             // anyway, so not restarting the read is currently harmless.
             boolean restart = !ftm.isReadOnly();
             if (restart) {
-                assert(ftm.getInitiateTask() != null);
+                Iv2InitiateTaskMessage initTaskMsg = ftm.getInitiateTask();
+                assert(initTaskMsg != null);
+                initTaskMsg.setTruncationHandle(TransactionInfoBaseMessage.UNUSED_TRUNC_HANDLE);
```
zeemanhe: Is the truncation handle reset needed here?  MpInitiatorMailbox.repairReplicasWithInternal() does not seem to carry the handle over using copy constructor **Iv2InitiateTaskMessage(long initiatorHSId,long coordinatorHSId, Iv2InitiateTaskMessage rhs)**

PR #5621, opened 2018-09-14
Eng 14335 vmc new tabs fix

johnTcrawford:
ENG-14335: Fix tests broken by original change:
Modified GEB VMC tests' run.sh file, and related test-tools.sh, and SQL
grammar gen's run.sh, to explicitly pass build parameters, not rely on a
BUILD_ARGS environment variable, to avoid confusion when Jenkins also
has a BUILD_ARGS parameter, as happened here. (Though this was a
secondary issue, the main one being in the Jenkins script, not starting
with "#!/bin/bash".) Plus other trivial improvements.


tests/sqlgrammar/run.sh
```diff
@@ -58,26 +58,26 @@ function find-directories-if-needed() {
 # Build VoltDB: 'community', open-source version
 function build() {
     echo -e "\n$0 performing: build$BUILD_ARGS"
-    test-tools-build
+    test-tools-build$BUILD_ARGS
```
kaikaizi: Should we add a whites pace in between?

tests/test-tools.sh
```diff
@@ -66,13 +66,13 @@ function test-tools-find-directories-if-needed() {
 }
 
 # Build VoltDB: 'community', open-source version
-# Optionally, you may specify BUILD_ARGS
+# Optionally, you may specify one or more build arguments ($@)
 function test-tools-build() {
     test-tools-find-directories-if-needed
-    echo -e "\n$0 performing: [test-tools-]build $BUILD_ARGS"
+    echo -e "\n$0 performing: [test-tools-]build" "$@"
```
kaikaizi: fyi  `$@` could also be in the same `""` mark.

tests/test-tools.sh
```diff
@@ -66,13 +66,13 @@ function test-tools-find-directories-if-needed() {
...
 
     cd $VOLTDB_COM_DIR
-    ant clean dist $BUILD_ARGS
+    ant clean dist "$@"
```
kaikaizi: I think it's safe to leave `$@` unquoted - quoting should generate a single argument, and won't work for multiple args? Not sure about this, but that was my impression.

tests/sqlgrammar/run.sh
johnTcrawford: Not needed currently, since BUILD_ARGS is always defined starting with a space; but probably a good idea anyway, in case that ever changes; so I will add a space.

tests/test-tools.sh
johnTcrawford: Yeah, good idea.

tests/test-tools.sh
johnTcrawford: I think it's fine either way, here, but I was guided by this:
http://wiki.bash-hackers.org/scripting/posparams#mass_usage
and especially: 'Well, let's just say: You almost always want a quoted "$@"!'
So, either would work here, but using double-quotes is good practice.

johnTcrawford:
Thanks, Lukai. I checked the double-quoting, and added some (optional) debug print, to make sure it works as expected, including with multiple build args. Will run it through Jenkins a couple more times, to make sure it works there, but, assuming all is well, I will then merge to master.

PR #5623, opened 2018-09-17
ENG-14566 misrouted every partition proc

zeemanhe:
For EveryPartition stored procedures, Initial task message could be misrouted back to MpScheduler upon leader migration, resulting in no hash exception.  Upon leader migration, Now the misrouted message will be routed to the new partition leader for transactional processing.

### DIFF
--- a/src/frontend/org/voltdb/iv2/MpScheduler.java
+++ b/src/frontend/org/voltdb/iv2/MpScheduler.java
@@ -72,6 +72,10 @@ public class MpScheduler extends Scheduler
     private final List<Long> m_iv2Masters;
     private final Map<Integer, Long> m_partitionMasters;
     private final List<Long> m_buddyHSIds;
+
+    // Leader migrated from one site to another
+    private final Map<Long, Long> m_leaderMigrationMap;
+
     private int m_nextBuddy = 0;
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/iv2/MpScheduler.java
```diff
@@ -454,17 +483,30 @@ private void handleIv2InitiateTaskMessageRepair(Iv2InitiateTaskMessage message)
...
-        if (message.getClientResponseData().getHashes() == null) {
-            tmLog.warn("The transaction response has no hash:" + message);
+        // A transaction may be routed back here for EveryPartitionTask via leader migration
+        if (counter != null && message.isMisrouted()) {
+            tmLog.info("The message on the partition is misrouted. TxnID: " + TxnEgo.txnIdToString(message.getTxnId()));
+            Long newLeader = m_leaderMigrationMap.get(message.m_sourceHSId);
+            if (newLeader != null) {
+                // Update the DuplicateCounter with new replica
+                counter.updateReplica(message.m_sourceHSId, newLeader);
+                m_leaderMigrationMap.clear();
```
luyangco: What if misrouted response comes back late after second migration finishes. Although I think it's unlikely, but it's theoretically possible.

PR #5653, opened 2018-09-20
ENG-14539 JSON API V2.0 for HTTP requests only

prosegay:


src/frontend/org/voltdb/utils/ApiRequestServletV2.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ *
+ * The servlet that handles all API request for HTTP-JSON interface.
+ * This servlet calls procedures system or otherwise.
+ */
+public class ApiRequestServletV2 extends VoltBaseServlet {
+
+    private static final long serialVersionUID = -6240161897983329796L;
+
```
JadenW: Nitpick: ServletV2 has exactly same logic as V1 for now, right? 
Change serialVersionUID to make it different in case we later change V2 API?

PR #5654, opened 2018-09-21
Eng 14316 revert rm rep stream

GuaGuaDean:
Revert the changes in [ENG-12827](https://github.com/VoltDB/voltdb/pull/4957/files)


src/frontend/org/voltdb/PartitionDRGateway.java
```diff
@@ -51,11 +51,6 @@
         SPECIAL
     }
 
-    // Keep sync with EE REPLICATED_TABLE_MASK at types.h
```
n3world: Shouldn't REPLICATED_TABLE_MASK also be removed from types.h?

GuaGuaDean: Thank you, updated.

PR #5655, opened 2018-09-21
ENG-14619-voltsql-improvement-2

Chaozz:
This PR is a mixed fix for ENG-14619 and ENG-14620.
Also improve the layout of the bottom toolbar.


lib/python/voltsql/voltsql/voltliterals/volt_literals.json
```diff
@@ -89,14 +90,15 @@
         ],
         "PRIMARY": [],
         "RECURSIVE": [],
+        "REMOVE CLASSES": [],
```
yzhang1991: Is there a LOAD CLASSES?

lib/python/voltsql/voltsql/voltliterals/volt_literals.json
```diff
@@ -113,7 +115,10 @@
         "VIEW": [],
         "WHEN": [],
         "WHERE": [],
-        "WITH": []
+        "WITH": [],
+        "QUIT": [],
+        "REFRESH": [],
```
yzhang1991: Where do we use refresh?

lib/python/voltsql/voltsql/voltliterals/volt_literals.json
```diff
@@ -251,17 +256,16 @@
     ],
     "datatypes": [
         "BIGINT",
-        "BOOLEAN",
-        "CHAR",
-        "DATE",
-        "DOUBLE PRECISION",
-        "INT",
```
yzhang1991: INT is valid

lib/python/voltsql/voltsql/voltliterals/volt_literals.json
Chaozz: Yes, there is.

lib/python/voltsql/voltsql/voltliterals/volt_literals.json
Chaozz: I change previous voltsql special command UPDATE to REFRESH(cause UPDATE is already a keyword). It is can be used to force the voltsql refreshing the schemas from catalog.

PR #5663, opened 2018-09-27
ENG-14487-disallow truncate from single partitioned procedures

Chaozz:
If we found a truncate statement in a single partitioned procedure, we will disallow it and throw an exception.


tests/frontend/org/voltdb/fullddlfeatures/TestDDLFeatures.java
```diff
@@ -187,6 +197,12 @@ public void testCreateProcedureFromClass() throws Exception {
         // ENG-14210 more than 1025 parameters
         RegressionSuite.verifyProcFails(m_client, "The statement's parameter count 1200 must not exceed the maximum 1025",
                 "@AdHoc", "CREATE PROCEDURE FROM CLASS org.voltdb_testprocs.fullddlfeatures.testJavaProcTooManyParams;");
+
+        // ENG-14487 truncate statement is not allowed for single partitioned procedures.
+        RegressionSuite.verifyProcFails(m_client,
+                "Single partitioned procedure: org.voltdb_testprocs.fullddlfeatures.testSinglePartitionedTruncateProc has TRUNCATE statement: \"truncate table t2;\".",
+                "@AdHoc",
+                "CREATE PROCEDURE PARTITION ON TABLE T2 COLUMN area FROM CLASS org.voltdb_testprocs.fullddlfeatures.testSinglePartitionedTruncateProc;");
```
yzhang1991: All of the test cases you added here are using the "CREATE PROCEDURE PARTITION ON ..." syntax. It would be nice to add one more case to test the "PARTITION PROCEDURE" statement.

Chaozz: done.

PR #5667, opened 2018-09-28
ENG-14459: Add row_number() to SqlCoverage

johnTcrawford:
Mostly refactored analytic-template.sql; minor changes to 3 other files.

### DIFF
--- a/tests/sqlcoverage/config/postgis-analytic-config.py
+++ b/tests/sqlcoverage/config/postgis-analytic-config.py
@@ -33,11 +33,11 @@
 {
     # Tests of SQL Analytic Windowed Functions (e.g. RANK, DENSE_RANK)
     "analytic": {"schema": "schema.py",
-                 "ddl": "DDL.sql",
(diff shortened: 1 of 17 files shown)


tests/sqlcoverage/config/postgis-analytic-config.py
```diff
@@ -33,11 +33,11 @@
 {
     # Tests of SQL Analytic Windowed Functions (e.g. RANK, DENSE_RANK)
     "analytic": {"schema": "schema.py",
-                 "ddl": "DDL.sql",
+                 "ddl": "DDL-null.sql",
```
yzhang1991: What does the new `-null` postfix mean?

johnTcrawford: It means to use a different DDL file, which has that name because it allows NULL values for all but the Primary Key column, unlike the standard DDL.sql file. (This is not strictly needed for the new ROW_NUMBER tests, but seems like a good idea for all these tests.)

PR #5668, opened 2018-09-28
ENG-14485: LIKE operator handles two or more consecutive '%' 

Chaozz:
Fix a bug that LIKE operator does not handle two or more consecutive '%' characters correctly.

The patterns `abc%%`, `%%abc`, `a%%bc`, `%%` should all match the string `abc`.

### DIFF
--- a/src/ee/common/NValue.hpp
+++ b/src/ee/common/NValue.hpp
@@ -3944,7 +3944,19 @@ inline NValue NValue::like(const NValue& rhs) const {
                     }
 
                     const char *postPercentPatternIterator = m_pattern.getCursor();
-                    const uint32_t nextPatternCodePointAfterPercent = m_pattern.extractCodePoint();
+                    uint32_t nextPatternCodePointAfterPercent = m_pattern.extractCodePoint();
+
+                    // ENG-14485 handle two or more consecutive '%' characters at the end of the pattern
+                    if (m_value.atEnd()) {
+                        while (nextPatternCodePointAfterPercent == '%') {
+                            if (m_pattern.atEnd()) {
+                                return true;
+                            }
+                            nextPatternCodePointAfterPercent = m_pattern.extractCodePoint();
+                        }
+                        return false;
+                    }
+
                     const bool nextPatternCodePointAfterPercentIsSpecial =
                             (nextPatternCodePointAfterPercent == '_') ||
                             (nextPatternCodePointAfterPercent == '%');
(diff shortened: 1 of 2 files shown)


src/ee/common/NValue.hpp
```diff
@@ -3944,7 +3944,19 @@ inline NValue NValue::like(const NValue& rhs) const {
                     }
 
                     const char *postPercentPatternIterator = m_pattern.getCursor();
-                    const uint32_t nextPatternCodePointAfterPercent = m_pattern.extractCodePoint();
+                    uint32_t nextPatternCodePointAfterPercent = m_pattern.extractCodePoint();
```
kaikaizi: Can this be replaced with `nextPatternCodePoint` on L3933:
`uint32_t nextPatternCodePointAfterPercent = nextPatternCodePoint;`?
Or even better, make L3939 non-const, and use that variable instead of declaring a new local variable?

src/ee/common/NValue.hpp
```diff
@@ -3944,7 +3944,19 @@ inline NValue NValue::like(const NValue& rhs) const {
...
+
+                    // ENG-14485 handle two or more consecutive '%' characters at the end of the pattern
+                    if (m_value.atEnd()) {
+                        while (nextPatternCodePointAfterPercent == '%') {
+                            if (m_pattern.atEnd()) {
+                                return true;
```
kaikaizi: L3953 and L3957 both should correspond to two or more '%' characters, right? Why do they return different values?

Chaozz: if the rest of pattern string are all '%', return true.
Otherwise, return false.

src/ee/common/NValue.hpp
Chaozz: I prefer 2 variables cause they are for different uses.
`nextPatternCodePoint` is the current character at the pattern string.
`nextPatternCodePointAfterPercent` is to read & check the characters after `nextPatternCodePoint`

src/ee/common/NValue.hpp
kaikaizi: I would personally grunt how `Liker` does things, by doing part of what a full-power regex with do: the way `Liker` is a class defined inside a function; and the comment on L3952-3962 (which does some of regex matching process, and was crafted that it is quite hard to grok now).
But unless we were to re-do string matching with regex `.*<string>.*`, it should be fine now that it passes those tests.

PR #5672, opened 2018-10-01
Added necessary source code backing up CatalogAdapter.java

kaikaizi:
Short-lived branch serving the purpose of adding SchemaPlus interface to non-DDL work.


src/frontend/org/voltdb/calciteadapter/rel/AbstractVoltDBTableScan.java
```diff
@@ -0,0 +1,78 @@
...
+        String dg = super.computeDigest();
+        return dg + "_" + m_voltDBTable.getCatTable().getTypeName();
+    }
+
+    public VoltTable getVoltDBTable() {
+        return m_voltDBTable;
+    }
+
+    @Override
+    public RelDataType deriveRowType() {
```
Chaozz: seems identical to base class's method, no need to override

src/frontend/org/voltdb/calciteadapter/rel/AbstractVoltDBTableScan.java
```diff
@@ -0,0 +1,78 @@
...
+        return table.getRowType();
+    }
+
+    @Override
+    public RelOptCost computeSelfCost(RelOptPlanner planner,
+                                      RelMetadataQuery mq) {
```
Chaozz: seems identical to base class's method, no need to override

src/frontend/org/voltdb/calciteadapter/rel/AbstractVoltDBTableScan.java
kaikaizi: It only "seems" identical, because the `table` (`VoltTable` type) is different from base class's member `table` (`RelOptTable` type).

src/frontend/org/voltdb/calciteadapter/rel/AbstractVoltDBTableScan.java
kaikaizi: Indeed. Will remove this one.

src/frontend/org/voltdb/calciteadapter/rel/AbstractVoltDBTableScan.java
Chaozz: I think `m_voltDBTable` is `VoltTable` type, but `table` is actually a `RelOptTable` defined in the base class `TableScan`.

kaikaizi: You're right. I missed that. Thanks for the catch!

PR #5682, opened 2018-10-09
ENG-14441-row-number-function (#5554)

Chaozz:
# back-port row number function to 8.3

The `ROW_NUMBER()` function is a window function that assigns a unique integer to each row in a result set. 

The syntax:
```sql
SELECT ROW_NUMBER() OVER(
    [PARTITION BY column_1, column_2,…]
    [ORDER BY column_3,column_4,…]
) FROM table_1;
```

notes
------
We don't support use of both a windowed function call and GROUP BY in a single query, like
```sql
select distinct A, row_number() over(order by A) as R from T;
```
although it is valid in Postgresql.
instead, you can do
```sql
SELECT A, ROW_NUMBER () OVER (ORDER BY A) FROM ( SELECT DISTINCT A FROM T) T2;
```


yzhang1991:
We already made the kit, so closing the pull request.

PR #5696, opened 2018-10-17
ENG-14545: Sending DR BinaryLog Invocation after transaction commit on

JadenW:
master and replicas


src/frontend/org/voltdb/RestoreAgent.java
```diff
@@ -1291,6 +1291,7 @@ private void changeState() {
             fetchSnapshotTxnId();
 
             exitRestore();
+            //?  directly transfer to replay state and did the replay, not using state transition
```
manjujames: Is this a note to do something here? Or did you just add a clarifying comment?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -41,16 +41,8 @@
 import org.voltcore.messaging.TransactionInfoBaseMessage;
 import org.voltcore.messaging.VoltMessage;
 import org.voltcore.utils.CoreUtils;
-import org.voltdb.ClientResponseImpl;
-import org.voltdb.CommandLog;
+import org.voltdb.*;
```
manjujames: I think importing everything in a package this way is highly discouraged.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1671,17 +1663,18 @@ private void setRepairLogTruncationHandle(long newHandle, boolean isForLeader)
      */
     private void scheduleRepairLogTruncateMsg()
     {
-        if (m_sendToHSIds.length == 0) {
```
manjujames: May be you can check if there are any txncommit interests registered also here. Then we don't need to schedule the task unnecessarily, if DR is not enabled.

src/frontend/org/voltdb/RestoreAgent.java
JadenW: For my investigation only, will remove

src/frontend/org/voltdb/iv2/SpScheduler.java
JadenW: Ide, will revoke the change.

src/frontend/org/voltdb/iv2/SpScheduler.java
JadenW: Right.

src/ee/common/Topend.cpp
```diff
@@ -86,15 +86,15 @@ namespace voltdb {
         receivedExportBuffer = true;
     }
 
-    int64_t DummyTopend::pushDRBuffer(int32_t partitionId, voltdb::StreamBlock *block) {
+    int64_t DummyTopend::pushDRBuffer(int32_t partitionId, int64_t committedSpHandle, voltdb::StreamBlock *block) {
         receivedDRBuffer = true;
         partitionIds.push(partitionId);
+        // TODO
```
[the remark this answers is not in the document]
JadenW: I haven't fully check the IPC path will involve with dummyTopend here . Will address or remove this after that.

src/frontend/org/voltdb/iv2/RepairLog.java
```diff
@@ -148,17 +148,26 @@ void setLeaderState(boolean isLeader)
         }
     }
 
+    public boolean hasNoTxnCommitInterests() {
+        return m_txnCommitInterests.isEmpty();
+    }
+
+    public void notifyTxnCommitInterests(long handle) {
```
n3world: This probably doesn't have to be public it could be either protected or default protection level.

PR #5704, opened 2018-10-18
ENG-14792: Calcite SqlValidator

Chaozz:
- checkout from `calcite-SchemaPlus` branch and remove the redundant wrappers for views and indexes.

- Make the `SchemaPlus` a member of` CatalogContext`.
- Make a new `SchemaPlus` on catalog update.
- Use the SchemaPlus in the CatalogContext, and other operatorTable etc. to construct the SqlValidator.
- This validator is used in `PlannerTool.planSqlCalcite()` to validate the query tree.


src/frontend/org/voltdb/CatalogContext.java
```diff
@@ -117,6 +119,9 @@ public CatalogInfo(byte[] catalogBytes, byte[] catalogBytesHash, byte[] deployme
     // database settings. contains both cluster and path settings
     private final DbSettings m_dbSettings;
 
+    // the SchemaPlus object
+    private final SchemaPlus m_schemaPlus;
```
kaikaizi: Each time the catalog is updated, a new SchemaPlus instance needs to be created/replaced, because of how all catalog objects work. So it cannot be final here.

kaikaizi: Edit: since the SchemaPlus instance does not outlive Catalog instance in the constructor, setting it final should work.

PR #5708, opened 2018-10-19
[ENG-14621] CREATE TABLE through Calcite!

kaikaizi:
Runs CREATE TABLE DDL through calcite. Should have full support and no
behavior change, unless we see otherwise from Jenkins...

It has all the CalciteAdapter changes, since that is not yet merged to calcite master branch.
The CREATE TABLE stmt got added to white list (renamed class AlwaysFail to 
RealCalciteCheck), so that Jenkins ideally should not fail.

The change appears to be quite large, but is largely in class CreateTableUtils.
I intentionally keep the "Table X has already been created!" println statement in sqlcmd,
to remind me that it is going through the new code path for catalog construction.


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1987,6 +1997,7 @@ private void addConstraintToCatalog(Table table,
             if (deleteStmt != null) {
                 Statement catStmt = table.getTuplelimitdeletestmt().add("limit_delete");
                 catStmt.setSqltext(deleteStmt);
+                System.err.println(deleteStmt);
```
n3world: Debug println which should be removed?

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -701,7 +709,7 @@ private Catalog replayFailedCatalogRebuildUnderDebug(
         // Or step OVER to debug just the catalog diff process, retried with verbose output --
         // maybe it's just being too sensitive to immaterial changes?
         Catalog autoGenCatalog = autoGenCompiler.compileCatalogInternal(null, null,
-                autogenReaderList, autoGenJarOutput);
+                autogenReaderList, new ArrayList<>(), autoGenJarOutput);
```
n3world: Wouldn't Collections.emptyList() be better than creating a new list?

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -1023,7 +1035,8 @@ public Catalog loadSchema(HSQLInterface hsql,
         List<VoltCompilerReader> ddlReaderList = DDLPathsToReaderList(ddlFilePaths);
         final VoltDDLElementTracker voltDdlTracker = new VoltDDLElementTracker(this);
         InMemoryJarfile jarOutput = new InMemoryJarfile();
-        compileDatabase(db, hsql, voltDdlTracker, null, null, ddlReaderList, null, whichProcs, jarOutput);
+        // NOTE/TODO: this is not from AdHoc code branch. We use the old code path here, and don't update CalciteSchema from VoltDB catalog.
+        compileDatabase(db, hsql, voltDdlTracker, null, null, ddlReaderList, new ArrayList<>(), null, whichProcs, jarOutput);
```
n3world: Wouldn't Collections.emptyList() be better?

src/frontend/org/voltdb/expressions/OperatorExpression.java
```diff
@@ -171,30 +171,28 @@ public void finalizeValueTypes()
...
-            return "(" + m_left.explain(impliedTableName) + " IS NULL)";
-        }
-        if (type == ExpressionType.OPERATOR_NOT) {
-            return "(NOT " + m_left.explain(impliedTableName) + ")";
-        }
-        if (type == ExpressionType.OPERATOR_CAST) {
-            return "(CAST (" + m_left.explain(impliedTableName) + " AS " + m_valueType.toSQLString() + "))";
+        switch(getExpressionType()) {
+            case OPERATOR_IS_NULL:
+                return "(" + m_left.explain(impliedTableName) + " IS NULL)";
```
n3world: It might be better to use a StringBuilder to construct the result string and pass that into m_left.explain() instead of constructing and concatenating multiple strings.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -199,7 +210,7 @@ else if (ddlToken.equals("create")) {
         return updateApplication("@AdHoc",
                                 null,
                                 null,
-                                sqlStatements.toArray(new String[0]),
+                                sqlStatements.toArray(new String[0]), sqlNodes,
```
n3world: sqlNodes should probably be on its own line to match the formatting of the other arguments or put them all on the same line with wrapping at line length.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -132,10 +136,17 @@ else if (mix == AdHocSQLMix.ALL_DML_OR_DQL) {
      * @author Yiqun Zhang
      */
     private CompletableFuture<ClientResponse> runDDLBatchThroughCalcite(SqlBatch batch) {
-        return null;
+        // We batch SqlNode with original SQL DDL stmts together, because we need both.
+        final List<Pair<String, SqlNode>> args =
```
n3world: This creates a list of Pairs just to create two lists one of the sql and one of the parsed query. This is completely unnecessary. It is much more efficient to iterate over the batch once and generate the two lists during a single iteration.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -120,8 +124,8 @@ else if (mix == AdHocSQLMix.ALL_DML_OR_DQL) {
 
         // at this point assume all DDL
         assert(mix == AdHocSQLMix.ALL_DDL);
-
-        return runDDLBatch(sqlStatements);
+        // Since we are not going through Calcite, there is no need to update CalciteSchema.
+        return runDDLBatch(sqlStatements, new ArrayList<>());
```
n3world: Should probably use Collections.emptyList().

src/frontend/org/voltdb/sysprocs/Promote.java
```diff
@@ -46,7 +47,7 @@
         return updateApplication("@UpdateApplicationCatalog",
                                 null,
                                 null,
-                                new String[0],
+                                new String[0], new ArrayList<>(),
```
n3world: Should probably use Collections.emptyList() and the placement of the argument does not match the formatting.

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -494,7 +503,7 @@ public static long getNextGenerationId() {
             ccr = prepareApplicationCatalogDiff(invocationName,
                                                 operationBytes,
                                                 operationString,
-                                                adhocDDLStmts,
+                                                adhocDDLStmts, sqlNodes,
```
n3world: Argument placement does not match other arguments.

src/frontend/org/voltdb/sysprocs/UpdateApplicationCatalog.java
```diff
@@ -54,11 +55,11 @@
         }
 
         logCatalogUpdateInvocation("@UpdateApplicationCatalog");
-
+        // NOTE/TODO: this is not from AdHoc code branch. We use the old code path here, and don't update CalciteSchema from VoltDB catalog.
         return updateApplication("@UpdateApplicationCatalog",
                                 catalogJarBytes,
                                 deploymentString,
-                                new String[0],
+                                new String[0], new ArrayList<>(),
```
n3world: Should use Collections.emptyList() and argument formatting is inconsistent.

src/frontend/org/voltdb/sysprocs/UpdateClasses.java
```diff
@@ -45,11 +46,11 @@
         }
 
         logCatalogUpdateInvocation("@UpdateClasses");
-
+        // NOTE/TODO: this is not from AdHoc code branch. We use the old code path here, and don't update CalciteSchema from VoltDB catalog.
         return updateApplication("@UpdateClasses",
                                 jarfileBytes,
                                 classesToDeleteSelector,
-                                new String[0],
+                                new String[0], new ArrayList<>(),
```
n3world: Should use Collections.emptyList() and argument formatting is inconsistent.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1249,7 +1248,18 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
         HashMap<String, Index> indexMap = new HashMap<>();
 
         final String name = node.attributes.get("name");
-
+        final AtomicBoolean hasCreated = new AtomicBoolean(false);
+        // Skip "CREATE TABLE" in here, since we have already added it using Calcite parser.
+        db.getTables().forEach(tbl -> {
+            if (tbl.getTypeName().equals(name)) {
+                System.err.println("Table " + name + " has already been created!");
```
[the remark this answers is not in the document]
kaikaizi: It is more of a temporay remind to myself that the code part has be rerouted to the new CreateTableUtils class. Prefer having it hanging around in here for now. Will remove this before merging to master.

src/frontend/org/voltdb/sysprocs/AdHoc.java
kaikaizi: Will do.

src/frontend/org/voltdb/compiler/DDLCompiler.java
kaikaizi: will do.

src/frontend/org/voltdb/sysprocs/AdHoc.java
kaikaizi: Sure, in the end that will be the goal. This is just a shortcut without changing too many parts of what is not directly relevant to the key of CREATE TABLE (and probably also update lots of tests).

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1249,7 +1248,18 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
...
+                hasCreated.set(true);
+            }
+        });
+        //assert(hasCreated.get()) : "All tables should have been created for CalciteSchema.";
+        if (hasCreated.get()) {
+            return;
+        }       // Code below this point is not executed any more.
```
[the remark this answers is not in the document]
kaikaizi: I would personally prefer delaying removing all deprecated code in DDLCompiler till I'm absolutely sure that the new CREATE TABLE works as the old one. At that time I'll clean up all unused code here, and everywhere else.

src/frontend/org/voltdb/sysprocs/AdHoc.java
kaikaizi: I split & passed two lists to `runDDLBatch()` method here because, the current behavior is it only expects a list of SQL strings, so the split has to be done somewhere down there.

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -17,15 +17,13 @@
...
 package org.voltdb.sysprocs;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
+import java.util.*;
```
n3world: I thought imports of entire packages were discouraged?

kaikaizi: Not sure if it is or isn't. My IntelliJ grouped them together. Maybe there is an Eclipse setting? Will change them back.

src/frontend/org/voltdb/sysprocs/AdHoc.java
n3world: But you are creating three different lists one of pairs of String and SqlNode and then you take that list and split it into two lists one of just the strings and one of just the SqlNodes. The list of Pairs is just an unnecessary intermediate step. This could be done more efficiently like
`
List<String> sqlTasks = new ArrayList<>(batch.getTaskCount());
List<String> parsedQuaries = new ArrayList<>(batch.getTaskCount());
List<String> validatedHomogeonousSQL = new ArrayList<>();
for (SqlTask task : batch) {
    String sql = task.getSQL();
    sqlTasks.add(sql);
    parsedQuaries.add(task.getParsedQuery());
    processAdHocSQLStmtTypes(sql, validatedHomogeonousSQL);
    validatedHomogeonousSQL.clear();
}
runDDLBatch(sqlTasks, parsedQuaries);
`

src/frontend/org/voltdb/parser/SqlParserFactory.java
```diff
@@ -26,16 +28,15 @@
...
      * create a {@link SqlParser} for it.
      * @param sql the SQL statement to parse.
      * @return a SQL parser created from the SQL statement.
      */
-    public static SqlParser create(String sql) {
-        return SqlParser.create(sql, SqlParser.configBuilder()
-                                              .setParserFactory(SqlDdlParserImpl.FACTORY)
-                                              .build());
+    public static SqlNode parse(String sql) throws SqlParseException {
+        return SqlParser.create(sql,
```
Chaozz: This change may affect my other branches, but it's not a big deal. 
The parser has some other functions may be useful:
`parseExpression()` can be used in testing
`parseQuery(String sql)` can let you reuse a parser.

So I still vote for the original one.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+            throw new PlanningErrorException(String.format(
+                    "Function %s not allowed for DEFAULT value of column \"%s\" of %s type",
+                    funName, colName, vt.getName()));
+        } else {
+            return "CURRENT_TIMESTAMP:43";
+        }
+    }
+
+    private static int addColumn(SqlColumnDeclarationWithExpression col, Table t, String tableName,
+                                 AtomicInteger index, Map<Integer, VoltType> columnTypes, Map<String, Index> indexes) {
```
Chaozz: the `Map<String, Index> indexes` is not used.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+        final List<SqlNode> nameAndType = col.getOperandList();
+        final String colName = nameAndType.get(0).toString();
+        final Column column = t.getColumns().add(colName);
+        column.setName(colName);
+
+        final SqlDataTypeSpec type = (SqlDataTypeSpec) nameAndType.get(1);
+        //final int scale = type.getScale();
```
Chaozz: commented code here

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+     */
+    private static Pair<String, String> genIndexAndConstraintName(
+            SqlNode constraintName, SqlKind type, String tableName, List<Column> cols, boolean hasExpression) {
+        if (constraintName != null) {
+            return Pair.of("VOLTDB_AUTOGEN_CONSTRAINT_IDX_" + constraintName.toString(), constraintName.toString());
+        } else {
+            final String suffix,
+                    type_str = type == SqlKind.PRIMARY_KEY ? "PK" : "CT",
+                    index = cols.stream().map(Column::getName)
+                            .reduce(String.format("VOLTDB_AUTOGEN_IDX_%s_%s", type_str, tableName),
```
Chaozz: I guess we want a unique name for a unique index here.
What if we have a table name with underline? say "FOO_BAR"

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+                                    (acc, colName) -> acc + "_" + colName),
+                    constr = cols.stream().map(Column::getName)
+                            .reduce(String.format("VOLTDB_AUTOGEN_CT__%s_%s", type_str, tableName),
```
Chaozz: `VOLTDB_AUTOGEN_CT__%s_%s` -> `VOLTDB_AUTOGEN_CT_%s_%s`

src/frontend/org/voltdb/sysprocs/AdHoc.java
kaikaizi: I think that's just the difference of procedural vs. functional style to express things. I liked the way I did: it's easier to reason that the list of SQL query and the list of SqlNode is pair-wise relation. 

True there is additional cost of creating a list of pairs, run a `void` function on the list of queries, then splitting into two lists as separate args to `runDDLBatch` method. But the cost is rather negligible compared with the clarity it brings.

(P.S. the `processAdHocSqlStmtTypes()` method is only needed for its side effects: there is no need to store `validHomogenousSQL` list because most of its deeds had been taken care of by Calcite).

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+                                    (acc, colName) -> acc + "_" + colName);
+            if (hasExpression) {        // For index involving expression(s), generate a UID string as suffix, and pretending that
+                // expression-groups for different constraints are never the same
```
Chaozz: can we hash the expression string instead of generating a UUID?

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+
+    private static Index genIndex(SqlKind type, Table t, String indexName, String constraintName, boolean hasGeog,
+                                  List<Column> indexCols, List<AbstractExpression> exprs) {
+        final Index index = t.getIndexes().add(indexName);
+        if (hasGeog) {
+            index.setCountable(false);
+            index.setType(IndexType.COVERING_CELL_INDEX.getValue());
+        } else {
+            index.setCountable(true);
+            index.setType(IndexType.BALANCED_TREE.getValue());
```
Chaozz: I found we use `IndexType.HASH_TABLE` in some places but marked it as deprecated.
https://github.com/VoltDB/voltdb/blob/be5e47a191300710ed535bc6c3f51dabe45b5309/src/frontend/org/voltdb/compiler/DDLCompiler.java#L1798
So I am not sure if we should have a hash_index option here.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -0,0 +1,601 @@
...
+    private static void addTableAnnotation(Table t) {
+        // TODO: assuming that no mat view is associated with t; and t is not a streaming table
+        final TableAnnotation annotation = new TableAnnotation();
+        annotation.ddl = CatalogSchemaTools.toSchema(new StringBuilder(), t,
+                null, false, null, null);
+        t.setAnnotation(annotation);
+    }
+
+    // NOTE: sadly, HSQL is pretty deeply ingrained and it's hard to make this method independent of HSQL.
+    public static Pair<SchemaPlus, Pair<Statement, VoltXMLElement>> addTable(
```
Chaozz: java doc for this method

src/frontend/org/voltdb/sysprocs/AdHoc.java
```diff
@@ -17,15 +17,13 @@
...
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
```
Chaozz: import not used

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
kaikaizi: No. The naming was specifically used to match HSQL auto generated names (a few Jenkins tests would fail if I miss an underscore).
We might come back to index/constraint naming later, but why fix it if not broken?

build.xml
```diff
@@ -1212,6 +1212,7 @@ JAVA COMPILATION
    <exec executable="patch" dir="${build.calcite.dir}" failonerror="true" input="${vendor.tar.dir}/commits.patch">
       <arg value="-p1"/>
    </exec>
+   <delete file="${vendor.tar.dir}/commits.patch"/>
```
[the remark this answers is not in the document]
kaikaizi: Having git track the patch file is a hassle when I updated local Calcite repo, and changing Volt branches.
Since the shell script basically gets all the diffs on Calcite repo since 1.17 official release commit, there is no need to keep a separate copy. In fact, I would love to not generating the patch file at all (but that's impossible because the script works across GIT repo boundaries).
Manually generating a diff file and store in the directory, as a separate process of updating Calcite.jar files is bad, since all you need to do to track Calcite changes should be done in Calcite repo, not VoltDB repo.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
kaikaizi: we might be able to, but I'm not so ready to do that, now that using UUID does not seem to do anything bad, as a step to replace how HSQL generates index/constraint names.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
kaikaizi: Good point! Some how I missed that branch. We should stick whatsoever/however Volt had been working.

src/frontend/org/voltdb/sysprocs/AdHoc.java
n3world: The list passed into processAdHocSqlStmtTypes is cleared immediately so the result is not saved. I was just reusing the same list to prevent unnecessary object creation.

I would argue that streams are hardly if ever more readable then a standard for loop, they are more compact and they are always more expensive to use. I think if I remember correctly doing a simple forEach on a stream is takes 10-15% more time than a traditional for loop. With your implementation you are doing the forEach loop 4x and with the sample I proposed there is 1 traditional for loop. That makes the original code more than 350% slower and a lot of extra object creation.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1249,7 +1248,17 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
         HashMap<String, Index> indexMap = new HashMap<>();
 
         final String name = node.attributes.get("name");
-
+        final AtomicBoolean hasCreated = new AtomicBoolean(false);
+        // Skip "CREATE TABLE" in here, since we have already added it using Calcite parser.
+        db.getTables().forEach(tbl -> {
```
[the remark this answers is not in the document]
kaikaizi: `db.getTables()` returns a iterable of tables. It has no method called `get(String)`.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1947,8 +1956,8 @@ private void validateTupleLimitDeleteStmt(Statement catStmt) throws VoltCompiler
     }
 
     /** Accessor */
-    Collection<Map.Entry<Statement, VoltXMLElement>> getLimitDeleteStmtToXmlEntries() {
```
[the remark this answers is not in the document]
kaikaizi: Thanks. Done.

src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -780,6 +777,7 @@ private InMemoryJarfile compileInternal(
             final VoltCompilerReader cannonicalDDLIfAny,
             final Catalog previousCatalogIfAny,
             final List<VoltCompilerReader> ddlReaderList,
+            final List<SqlNode> sqlNodes,
```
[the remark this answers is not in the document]
kaikaizi: The doc was pretty darn outdated. I'm not fully sure what each param was meant to do, will give it a try.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
kaikaizi: [ENG-12475](https://issues.voltdb.com/browse/ENG-12475) has deprecated support for hash index, and it is now the time to officially ditch it. Will add a comment in my code about that.

src/frontend/org/voltdb/compiler/VoltCompiler.java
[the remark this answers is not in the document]
kaikaizi: On a closer look, the call path of `compileDatabase` is a horribly mess (or, in @BillWhite's words, "Ugly, ugly hack") underneath. It's not easy to track, much less than to change anything about it.
I may not be able to update what was not already updated.

src/frontend/org/voltdb/parser/SqlParserFactory.java
kaikaizi: Tried `parseQuery(String sql)`, it breaks a few JUnit tests. It's best to create a new parser on each query encountered.
Separating parser creation and concrete query -> SqlNode construction is not really interesting outside the database query parser use. Putting them together makes it easier to use.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
kaikaizi: Then the index name will be "VOLTDB_AUTOGEN_IDX_CT_FOO_BAR_COL".
The name was auto generated by HSQL to differentiate from other index names. Since an index is associated with the associated table, **NOT** under global catalog space, there is no risk that two indices of two different tables end up with the same index name.

src/frontend/org/voltdb/compiler/DDLCompiler.java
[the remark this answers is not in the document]
n3world: org.voltdb.catalog.Database.getTables() returns a org.voltdb.catalog.CatalogMap which does have a get(String) method.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -87,6 +87,17 @@
...
  */
 public class CreateTableUtils {
+
+    /**
+     * Mimick user-observable error message from VoltCompiler.VoltCompilerException, by
+     * clearing stack traces.
+     */
+    private static final class DDLException extends PlanningErrorException {
+        public DDLException(String msg) {
+            super(msg);
```
n3world: You could invoke super(String, Throwable, boolean, boolean) which allows you to disable the stack trace and suppressing exceptions. `super(msg, null, true, false);`

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -108,30 +114,29 @@ private static void except(String msg) {
             VoltType vt, String tableName, String colName, int size, boolean inBytes) {
         if (size < 0) {        // user did not provide, e.g. CREATE TABLE t(i VARCHAR);
             size = vt.defaultLengthForVariableLengthType();
-        } else if (size > VoltType.MAX_VALUE_LENGTH) {
-            except(String.format("%s column %s in table %s has unsupported length %s",
-                    vt.toSQLString(), colName, tableName, VoltType.humanReadableSize(size)));
-        } else if (vt == VoltType.STRING && size > VoltType.MAX_VALUE_LENGTH_IN_CHARACTERS) {
+        }
+        exceptWhen(size > VoltType.MAX_VALUE_LENGTH,
+                String.format("%s column %s in table %s has unsupported length %s",
```
n3world: One problem with using a utility method like this is that you might be constructing the string every time rather than only when the exception is to be thrown. I know java will inline private methods if they are small enough so you might be OK here but usually you want to avoid rendering a string you don't actually need. It would normally be slightly cheaper to change the method to take in the format string and args and do the format only if the exception is to be thrown.

src/frontend/org/voltdb/planner/PlanningErrorException.java
```diff
@@ -23,4 +23,15 @@
...
     }
+
+    /**
+     * Create an exception with stacktrace erased. Used for reporting
+     * parsing/validation errors without the long backtrace list, to mimick
+     * user-observable error message from VoltCompiler.VoltCompilerException.
+     * @param msg exception message
+     * @param ignored not used
+     */
+    public PlanningErrorException(String msg, int ignored) {
```
n3world: Why not just take in the boolean for stackTrace and just pass that down rather than an ignored argument?

kaikaizi: `int` is only for differentiate from the other, more popular constructor. `0` is easier to write than `true` or `false`; and I ignore it so that caller don't need to reference what the value means. Passing `true` down is largely the same as calling the other constructor.
If you would make the other constructor `private`, and expose only this ctor, it would involve changing additional hundreds of lines of code scattered elsewhere. Exposing two ctors makes the 2nd argument mostly useless (ignored), only serving to differentiate the behavior.

PR #5710, opened 2018-10-24
ENG-14774: Improve partition placement for host removal

n3world:
Simplify the partition placement algorithm by forcing all partitions to have the same k factor and all hosts to have the site count. Modify the protection group creation algorithm to create all protection groups and then begin placing the partitions within each group.

Simplify public API of AbstractTopology so only the methods needed are exposed.

Pairs with https://github.com/VoltDB/pro/pull/2380


src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -295,566 +233,756 @@ public int compareTo(Host o) {
...
+         * @return {@code this}
+         */
+        TopologyBuilder addPartitionsToHosts() {
+            int hostsNotInAGroup = m_hosts.size();
+            BalancedPartitionChecker bpc = m_selector.getBalancedPartitionChecker();
+
+            List<HostBuilder[]> protectionGroups = new ArrayList<>(hostsNotInAGroup / m_replicaCount);
+
+            do {
+                int protectionGoupCount = (hostsNotInAGroup / m_replicaCount);
```
manjujames: Nit: typo - protectionGoupCount is missing r in group

tests/frontend/org/voltdb/TestAbstractTopology.java
```diff
@@ -209,6 +217,39 @@ Metrics validate(AbstractTopology topo) throws JSONException {
...
+            }
+            Set<Integer> hosts = new HashSet<>();
+            @SuppressWarnings("unchecked")
+            Range<Integer>[] partitionRange = new Range[1];
+            buildProtectionGroup(topo, visited, host, partitionRange, hosts);
+            assertEquals(hosts.size() * topo.getSitesPerHost() / (topo.getReplicationFactor() + 1),
+                    partitionRange[0].upperEndpoint() - partitionRange[0].lowerEndpoint() + 1);
+            protectionGroups.put(partitionRange[0], hosts);
+        }
+
```
JadenW: With, seems don't the "collect distinct peer groups" calculation from L163 - L172

n3world: Yes the protection groups are being calculated 2x. I'll remove the original way and keep the new once since that generates more information, which is checked.

PR #5711, opened 2018-10-25
ENG-14766: add VoltDB option to disable compression

rdykiel:
We discussed offering this option for customers that want to try improving export throughput at the expense of their disk space.

### DIFF
--- a/src/frontend/org/voltdb/export/StreamBlockQueue.java
+++ b/src/frontend/org/voltdb/export/StreamBlockQueue.java
@@ -72,9 +72,8 @@ public class StreamBlockQueue {
         if (exportLog.isDebugEnabled()) {
             exportLog.debug(m_nonce + " At SBQ creation, PBD size is " + (m_reader.sizeInBytes() - (8 * m_reader.getNumObjects())));
         }
-        if (DISABLE_COMPRESSION) {
-            exportLog.info(m_nonce + " has compression disabled in " + path);
-        }
+        exportLog.info(m_nonce + " has compression "
+                + (DISABLE_COMPRESSION ? "disabled" : "enabled") + " in " + path);
     }
 
     public boolean isEmpty() throws IOException {
--- a/src/frontend/org/voltdb/export/StreamBlockQueue.java
+++ b/src/frontend/org/voltdb/export/StreamBlockQueue.java
@@ -46,6 +46,8 @@ public class StreamBlockQueue {
     private static final VoltLogger exportLog = new VoltLogger("EXPORT");
     private static final int EXPORT_BUFFER_VERSION = 1;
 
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/export/StreamBlockQueue.java
```diff
@@ -60,14 +62,20 @@
     private final String m_nonce;
     private final String m_path;
     private BinaryDequeReader m_reader;
+    private final boolean m_compress = !DISABLE_COMPRESSION;
```
zeemanhe: This instance variable may not be needed.  use DISABLE_COMPRESSION directly on line 215

src/frontend/org/voltdb/export/StreamBlockQueue.java
```diff
@@ -66,8 +68,12 @@ public StreamBlockQueue(String path, String nonce) throws java.io.IOException {
         m_path = path;
         m_nonce = nonce;
         m_reader = m_persistentDeque.openForRead(m_nonce);
-        // temporary debug stmt
-        exportLog.info(m_nonce + " At SBQ creation, PBD size is " + (m_reader.sizeInBytes() - (8 * m_reader.getNumObjects())));
+
+        if (exportLog.isDebugEnabled()) {
+            exportLog.debug(m_nonce + " At SBQ creation, PBD size is " + (m_reader.sizeInBytes() - (8 * m_reader.getNumObjects())));
+        }
+        exportLog.info(m_nonce + " has compression "
```
wweiss-voltdb: We already have way too many log messages for export. If you want a log message do it in the Export Manager initialization path.

src/frontend/org/voltdb/export/StreamBlockQueue.java
```diff
@@ -204,7 +210,7 @@ public StreamBlock pop() {
      * Only allow two blocks in memory, put the rest in the persistent deque
      */
     public void offer(StreamBlock streamBlock) throws IOException {
-        m_persistentDeque.offer(streamBlock.asBBContainer());
+        m_persistentDeque.offer(streamBlock.asBBContainer(), !DISABLE_COMPRESSION);
```
wweiss-voltdb: You need an additional check in the recovery/rejoin path when opening an existing PBD to detect a change in the compression state (host started when compression disabled and rejoined with compression enabled). It is probably good enough to stop the host with an error when the difference is detected.

rdykiel: Why is this an issue? The flag is checked only when inserting buffers. Once inserted the compressed/uncompressed state is stored as a flag for each buffer so the readers can uncompress if needed.

src/frontend/org/voltdb/export/StreamBlockQueue.java
rdykiel: I'll move the message in the ExportManager compression path.

PR #5727, opened 2018-11-05
ENG-14805 New Catalog API

yzhang1991:
Catalog APIs are pulled out to separate classes "CatalogOperator" and "CatalogSerializer".
The next step will be making them into a JNI library so it will become a black box.


src/catgen/catalog.py
```diff
@@ -71,6 +71,10 @@ def genjava( classes, javaOnlyClasses, prepath, postpath, package ):
     os.system( interp( "cp $prepath/CatalogException.java $postpath", locals() ) )
     os.system( interp( "cp $prepath/CatalogChangeGroup.java $postpath", locals() ) )
     os.system( interp( "cp $prepath/CatalogDiffEngine.java $postpath", locals() ) )
+    os.system( interp( "cp $prepath/CatalogOperator.java $postpath", locals() ) )
```
n3world: I see that all of the other copies are done the same way but it would be much more efficient to use shutil.copy to do a native copy in pythion and not start a new shell which creates a cp process.

src/catgen/in/javasrc/CatalogCommand.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Create a catalog command from a serialized string.
+     * @param cmdStr serialized command string.
+     */
+    public CatalogCommand(String cmdStr) {
+        int pos = 0;
+        cmdStr = cmdStr.trim();
+        cmd = cmdStr.charAt(pos++);
+        while (cmdStr.charAt(pos++) != ' ');
```
n3world: Wouldn't pos = cmdStr.indexOf(' ', pos) + 1; be a better way to do this?

src/catgen/in/javasrc/CatalogCommand.java
```diff
@@ -0,0 +1,54 @@
...
+
+        int pathStart = pos;
+        while (cmdStr.charAt(pos++) != ' ');
+        path = cmdStr.substring(pathStart, pos - 1);
+
+        int argStart = pos;
+        while (cmdStr.charAt(pos++) != ' ');
```
n3world: Again indexOf(int, int) would be a better way to find the next space.

src/catgen/in/javasrc/CatalogCommand.java
```diff
@@ -0,0 +1,54 @@
...
+        int pos = 0;
+        cmdStr = cmdStr.trim();
+        cmd = cmdStr.charAt(pos++);
+        while (cmdStr.charAt(pos++) != ' ');
+
+        int pathStart = pos;
+        while (cmdStr.charAt(pos++) != ' ');
+        path = cmdStr.substring(pathStart, pos - 1);
+
+        int argStart = pos;
```
n3world: Is it worth handling extra white space between the different elements? You could update this to also just ignore leading and trailing whitespace and get rid of the initial trim.

src/catgen/in/javasrc/CatalogOperator.java
```diff
@@ -0,0 +1,176 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* WARNING: THIS FILE IS AUTO-GENERATED
```
n3world: Why is auto generated code being checked in and not just generated during the build?

src/catgen/in/javasrc/CatalogSerializer.java
```diff
@@ -0,0 +1,150 @@
...
+        Object value = ct.getField(field);
+        if (value == null) {
+            m_builder.append("null");
+        }
+        else if (value.getClass() == Integer.class)
+            m_builder.append(value);
+        else if (value.getClass() == Boolean.class)
+            m_builder.append(Boolean.toString((Boolean)value));
+        else if (value.getClass() == String.class)
+            m_builder.append("\"").append(value).append("\"");
```
n3world: There is a single character append which would be better for the double quotes. `append('"')`

src/catgen/in/javasrc/CatalogSerializer.java
```diff
@@ -0,0 +1,150 @@
...
+                mapsToVisit.add(map);
+            }
+        }
+        for (CatalogMap<? extends CatalogType> map : mapsToVisit) {
+            map.accept(this);
+        }
+    }
+
+    void writeDeleteDiffStatement(CatalogType ct, String parentName) {
+        m_builder.append("delete ").append(ct.getParent().getCatalogPath()).append(" ")
```
n3world: minor: Single char appends would be better for the spaces and newline

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -43,25 +44,36 @@
     /* White list of fields that we care about for DR for table children classes.
        This is used only in serialize commands for DR method.
        There are duplicates added to the set because it lists all the fields per type */
-    private static Set<String> s_whiteListFields = Sets.newHashSet(
+    private static final Set<String> s_whiteListFields = Sets.newHashSet(
```
n3world: For paranoia you might want to use `ImmutableSet.of(...)` so nobody can accidentally mutate this set.

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -43,25 +44,36 @@
...
             /* Statement */
-            "sqltext", "querytype", "readonly", "singlepartition", "replicatedtabledml", "iscontentdeterministic", "isorderdeterministic", "nondeterminismdetail",
-            "cost", "seqscancount", "explainplan", "tablesread", "tablesupdated", "indexesused", "cachekeyprefix"
+            "sqltext", "querytype", "readonly", "singlepartition", "replicatedtabledml",
+            "iscontentdeterministic", "isorderdeterministic", "nondeterminismdetail", "cost",
+            "seqscancount", "explainplan", "tablesread", "tablesupdated", "indexesused", "cachekeyprefix"
             );
 
+    @SuppressWarnings("unchecked")
+    private static final Set<Class<? extends CatalogType>> s_whiteListChildren =
```
n3world: ImmutableSet would be good for safety.

src/catgen/in/javasrc/Catalog.java
```diff
@@ -79,180 +82,6 @@ public CatalogType getParent() {
         return null;
     }
 
-    /**
-     * Run one or more single-line catalog commands separated by newlines.
-     * See the docs for more info on catalog statements.
-     * @param commands A string containing one or more catalog commands separated by
-     * newlines
-     */
-    public void execute(final String commands) {
```
n3world: Why not keep this method on catalog and just have the implementation be `getOperator().execute(commands)`. It would reduce some of the other changes that were made to use the new Catalog class

Similar for `serialize()`

src/catgen/in/javasrc/CatalogOperator.java
```diff
@@ -0,0 +1,176 @@
...
+            resolved.getCollection(catCmd.arg1).add(catCmd.arg2);
+            break;
+        case 'd': // delete
+            resolved.getCollection(catCmd.arg1).delete(catCmd.arg2);
+            String toDelete = catCmd.path + "/" + catCmd.arg1 + MAP_SEPARATOR + catCmd.arg2;
+            m_catalog.m_pathCache.invalidate(toDelete);
+            break;
+        case 's': // set
+            resolved.set(catCmd.arg1, catCmd.arg2);
+            break;
```
n3world: Should probably either add a default or a comment saying default not needed.

src/catgen/in/javasrc/Catalog.java
```diff
@@ -64,178 +56,24 @@ public Catalog getCatalog() {
...
     /**
      * Run one or more single-line catalog commands separated by newlines.
      * See the docs for more info on catalog statements.
-     * @param commands A string containing one or more catalog commands separated by
-     * newlines
+     * @param commands a string containing one or more catalog commands
+     * separated by newlines.
      */
     public void execute(final String commands) {
-
```
kaikaizi: Admittedly, this is another place littered with parsing the (already encoded) input.

src/catgen/in/javasrc/CatalogMap.java
```diff
@@ -87,6 +78,10 @@ public T get(String name) {
         return m_items.get(name.toUpperCase());
     }
 
+    private Object get1(String name) {
```
kaikaizi: Is there a better name for this method?

src/catgen/in/javasrc/CatalogMap.java
```diff
@@ -166,6 +161,10 @@ public T add(String name) {
         }
     }
 
+    private void add1(String name) {
```
kaikaizi: Is there a better name for this method?

src/catgen/in/javasrc/Catalog.java
```diff
@@ -21,30 +21,22 @@
...
-    //private final PatriciaTrie<CatalogType> m_pathCache = new PatriciaTrie<>();
-    Cache<String, CatalogType> m_pathCache = CacheBuilder.newBuilder().maximumSize(8).build();
+public final class Catalog extends CatalogType {
 
-    private CatalogType m_prevUsedPath = null;
+    private Cache<String, CatalogType> m_pathCache = CacheBuilder.newBuilder().maximumSize(8).build();
 
     CatalogMap<Cluster> m_clusters;
 
+    final CatalogOperator m_operator = new CatalogOperator(this);
```
kaikaizi: Add a getter for `m_operator`?

src/catgen/in/javasrc/CatalogType.java
```diff
@@ -60,7 +60,7 @@ public void set(T value) {
         T resolve() {
             synchronized (m_lock) {
                 if (m_unresolvedPath != null) {
-                    m_value = (T) getCatalog().getItemForPath(m_unresolvedPath);
+                    m_value = (T) getCatalog().m_operator.getItemForPath(m_unresolvedPath);
```
kaikaizi: Use `.getOperator()` in place of `m_operator`?

src/frontend/org/voltdb/NativeLibraryLoader.java
```diff
@@ -0,0 +1,226 @@
...
+                msgBuilder.append("VoltDB provides builds on our website for 64-bit OS X systems >= 10.6, ");
+                msgBuilder.append("and 64-bit Linux systems with kernels >= 2.6.18.");
+                if (e instanceof UnsatisfiedLinkError) {
+                    msgBuilder.append("\nOr the library may have failed to load because java.io.tmpdir should be set ");
+                    msgBuilder.append("to a different directory. Use VOLTDB_OPTS='-Djava.io.tmpdir=<dirpath>' to set it.");
+                }
+                VoltDB.crashLocalVoltDB(msgBuilder.toString(), false, null);
+            } else {
+                s_hostLog.info("Failed to load shared library " + fullLibName + "\nLibrary path: "
+                        + System.getProperty("java.library.path"));
```
kaikaizi: This is the way before; but should it also crash on missing libraries like UnsatsifiedLinkError? Both case fails to load library anyway.

src/frontend/org/voltdb/NativeLibraryLoader.java
```diff
@@ -0,0 +1,226 @@
...
+     * @return the native library file copied into a readable location.
+     */
+    private static File getNativeLibraryFile(String libname) {
+
+        // for now, arch is always x86_64
+        String pathFormat = "/org/voltdb/native/%s/x86_64";
+        String libPath = null;
+        if (System.getProperty("os.name").toLowerCase().contains("mac")) {
+            libPath = String.format(pathFormat, "Mac");
+        } else {
```
kaikaizi: In a more ideal world, I'd like to have
```suggestion
        } else if (System.getProperty("os.name").toLowerCase().contains("linux")) {
            libPath = String.format(pathFormat, "Linux");
        } else {
            throw new RuntimeError("Unsupported system: " + System.getProperty("os.name"));
         }
```

src/frontend/org/voltdb/NativeLibraryLoader.java
```diff
@@ -0,0 +1,226 @@
...
+        // Using UUID as in Snappy, but we probably don't need it?
+        String uuid = UUID.randomUUID().toString();
+        String extractedLibFileName = String.format("voltdb-%s-%s", uuid, libFileName);
+        File extractedLibFile = new File(tmpFolder, extractedLibFileName);
+
+        String libPath = libFolder + "/" + libFileName;
+        // Extract a native library file into the target directory
+        InputStream reader = null;
+        FileOutputStream writer = null;
+        try {
```
kaikaizi: A try-with-resources would be the thing to do here.

src/frontend/org/voltdb/utils/SystemStatsCollector.java
```diff
@@ -478,7 +478,8 @@ public static void main(String[] args) {
         String pidString = processName.substring(0, processName.indexOf('@'));
         pid = Integer.valueOf(pidString);
 
-        org.voltdb.EELibraryLoader.loadExecutionEngineLibrary(false);
+        // ETHAN (11/7/2018): If loading the native library does not have to succeed, why load?
```
kaikaizi: Exactly, and we can make `loadVoltDB` function nullary.
Or maybe that's for testing purpose only?
If that's the case, then we can still have it be nullary, and mock it up in test.

build.xml
```diff
@@ -125,6 +125,12 @@
     <os family="mac"/>
 </condition>
 
+<condition property="jar.nativelib.dir"
```
yzhang1991: Moved here from below.

src/catgen/catalog.py
```diff
@@ -63,18 +64,21 @@ def genjava( classes, javaOnlyClasses, prepath, postpath, package ):
...
-    os.system( interp( "cp $prepath/CatalogException.java $postpath", locals() ) )
-    os.system( interp( "cp $prepath/CatalogChangeGroup.java $postpath", locals() ) )
-    os.system( interp( "cp $prepath/CatalogDiffEngine.java $postpath", locals() ) )
-    os.system( interp( "cp $prepath/FilteredCatalogDiffEngine.java $postpath", locals() ) )
-    os.system( interp( "cp $prepath/DRCatalogDiffEngine.java $postpath", locals() ) )
-    os.system( interp( "cp $prepath/DRCatalogCommands.java $postpath", locals() ) )
-    os.system( interp( "cp $prepath/DatabaseConfiguration.java $postpath", locals() ) )
+    shutil.rmtree(postpath, ignore_errors=True)
+    os.makedirs(postpath)
+    shutil.copy(prepath + "/Catalog.java", postpath)
```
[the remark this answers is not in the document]
yzhang1991: Thanks for the suggestion, revised.

src/catgen/in/javasrc/Catalog.java
yzhang1991: I don't think so, I do not want to expose it.

src/catgen/in/javasrc/CatalogMap.java
yzhang1991: I will change it to `_get`

src/catgen/in/javasrc/CatalogMap.java
yzhang1991: `_add`

PR #5728, opened 2018-11-05
Send an error response to avoid sqlcmd hanging, and support concurrent default procedures updates

rdykiel:
Send an error response when hitting exception in runnable executed in OpsAgent. This should prevent sqlcmd from hanging forever and system tests to report failures faster.

Use ConcurrentHashMap to support updates from LoadedProcedure.getNibbleDeleteProc

### DIFF
--- a/src/frontend/org/voltdb/DefaultProcedureManager.java
+++ b/src/frontend/org/voltdb/DefaultProcedureManager.java
@@ -20,9 +20,9 @@ package org.voltdb;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.voltdb.CatalogContext.ProcedurePartitionInfo;
 import org.voltdb.catalog.Catalog;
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/DefaultProcedureManager.java
```diff
@@ -49,7 +49,7 @@
  */
 public class DefaultProcedureManager {
 
-    Map<String, Procedure> m_defaultProcMap = new HashMap<>();
+    Map<String, Procedure> m_defaultProcMap = new ConcurrentHashMap<>();
```
zeemanhe: ConcurrentHashMap : Iterators and Enumerations return elements reflecting the state of the hash table at some point at or since the creation of the iterator/enumeration. They do not throw ConcurrentModificationException.  So the stats returned may not reflect the correct info.

src/frontend/org/voltdb/OpsAgent.java
```diff
@@ -270,6 +271,16 @@ public void run() {
                     collectStatsImpl(c, clientHandle, selector, params);
                 } catch (Exception e) {
                     hostLog.warn("Exception while attempting to collect stats", e);
+                    if (e instanceof ConcurrentModificationException) {
```
zeemanhe: ConcurrentModificationException for this stats won't be thrown after the change in DefaultProcedureManager.

PR #5739, opened 2018-11-08
ENG-14846 Make Calcite support DROP TABLE

kaikaizi:
Use no hint from HSQL for searching target table and dependencies.
Still need to effect HSQL session, as VoltDB's logic is to batch
effect HSQL session state, then update catalog all at once for either
ad-hoc query or batch.


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -374,16 +374,29 @@ private void doExecuteStatement(String statement, int lineNum) throws VoltCompil
...
+        final StringBuilder ddls = new StringBuilder();
+        boolean isBatch = false;        // When reader contains multiple stmts, set it to indicate we are in batch mode.
         while (stmt != null) {
-            // Some statements are processed by VoltDB and the rest are handled by HSQL.
-            processVoltDBStatements(db, whichProcs, stmt);
+            final String encodedDropTable;
+            if (isBatch) {      // We cannot query previous database for existing tables in batch mode, as
+                encodedDropTable = null;    // current DDL batch may contain CREATE statements, that is later dropped.
+            } else {
+                encodedDropTable = DropTableUtils.run(prevDb, stmt.statement, m_schema, m_compiler);
```
Chaozz: why we would like to use the previous db instead of the current one?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -374,16 +374,29 @@ private void doExecuteStatement(String statement, int lineNum) throws VoltCompil
...
+            }
+            if (encodedDropTable == null) {
+                processVoltDBStatements(db, whichProcs, stmt);
+            } else {        // matched & executed DROP TABLE stmt
+                ddls.append(encodedDropTable);
```
Chaozz: we can process only one ddl in calcite, I think we don't need the string builder, just `m_fullDDL+=encodedDropTable`

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
```diff
@@ -201,7 +201,10 @@ private static int addColumn(SqlColumnDeclarationWithExpression col, Table t, St
                 column.setDefaultvalue(defaultFunctionValue(vt, ((SqlIdentifier) expr).getSimple(), colName));
             } else if (expr.getKind() == SqlKind.LITERAL) {
                 final String defaultValue;
-                if (expr instanceof SqlNumericLiteral) {
+                if (((SqlLiteral) expr).getValue() == null) {
```
Chaozz: is it a previous bug in **create table**?

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/DropTableUtils.java
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Convert a SQL statement to a SQLNode, ignoring any Calcite parse errors.
+     * Eventually, when Calcite is patched to be fully compatible with VoltDB syntax,
+     * we shall not silence any exceptions.
+     * @param sql SQL statement
+     * @return parsed SqlNode
+     */
+    private static SqlNode toSqlNode(String sql) {
+        try {
+            return new SqlBatchImpl(sql).iterator().next().getParsedQuery();
```
Chaozz: I think will only pass a single sql statement here, can use `SqlParserFactory.parse(sql)`

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/DropTableUtils.java
```diff
@@ -0,0 +1,222 @@
...
+            // In future (ENG-7542, ENG-6353, doc), we will drop those stored procedures in cascade mode.
+            CreateTableUtils.exceptWhen(! procs.isEmpty(),
+                    String.format("Cannot drop table %s: stored procedure(s) %s depend on it.",
+                            tableName, String.join(", ", procs)));
+            final String viewNames = String.join(", ", materializedViews);
+            CreateTableUtils.exceptWhen(!cascaded && !materializedViews.isEmpty(),
+                    String.format("Dependent object exists: PUBLIC.%s in statement [%s]", viewNames, sql));
+            materializedViews.add(tableName);
+            final VoltXMLElement.VoltXMLDiff od = new VoltXMLElement.VoltXMLDiff("databaseschemadatabaseschema");
+            IntStream.range(0, schema.children.size())
```
Chaozz: if we can visit the index in reverse order then we don't need to sort and store the intermediate result.
I am not sure if `IntStream` can do this, but for loop definitely can.

src/frontend/org/voltdb/compiler/DDLCompiler.java
kaikaizi: That's because of how catalog is updated in AdHoc or batch mode:
- In batch mode (Yes, DDL can also run in batch mode), all statements are executed together, and if you drop a table that was created in the same batch, then the table catalog is in current database catalog. I don't support the batch mode, because it may also be created before current batch; and other reasons.
- In ad-hoc mode, the previous db catalog contains all catalog updated before current query. So if we drop a table, it has to be found from previous database. Current database starts as empty, and we use HSQL session to replay/track all DDL/DMLs.
Until we get to the point that we no longer need HSQL session to update catalog, we can only support ad-hoc mode, and query previous catalog status from previous database.

Things get a lot messier when we create an index (in the other PR), because it has a possibly long chain of dependencies of various catalog objects; and it needs tweaking around the CatalogDiffEngine, that provides the greater aurora of protection for snapshot/restore.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/CreateTableUtils.java
kaikaizi: That's correct, I found this when working on this feature.

src/frontend/org/voltdb/compiler/DDLCompiler.java
kaikaizi: I'll just leave it in case of other DDLs/drops might need it: DROP INDEX/VIEW/STREAM/etc.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/DropTableUtils.java
kaikaizi: Good one, will do.

src/frontend/org/voltdb/sysprocs/org/voltdb/calciteutils/DropTableUtils.java
kaikaizi: Will change to loop.

PR #5732, opened 2018-11-06
ENG-14847: Pick up the Calcite logical rules into our feature branch

Chaozz:
- Converts a validated query's parse tree into a relational expression(RelRoot).
- Create a heuristic HepPlanner, apply the calcite logical rules, and transform the converted RelNode to a new RelNode.

I also add [Janino](https://janino-compiler.github.io/janino/) to the lib. Calcite uses it to generate Java code.

The plan results for join statements have some redundant steps. I suspect it is related to the `HepPlanner`. I have verfied that if I used the `ValcanoPlanner` the result will be better. But the `ValcanoPlanner` need some addtional VoltDB Logical rules to apply which is out of the scope of this PR. I will create another PR for that(see https://github.com/VoltDB/voltdb/pull/5740).

**update:**
I have fixed the bad plans for join operators in `HepPlanner`. The key point is, in HepPlanner, the order of rules matters. You need to add the rules to a `group`, otherwise each rules will only apply once in order.

PR #5740, opened 2018-11-09
ENG-14873: Pick up the VoltDB logical rules into our feature branch

Chaozz:
- Create a VolcanoPlanner, apply the calcite logical rules and the VoltDB logical rules, and transform the converted RelNode to a new RelNode.
- What VoltDB Logical Rules do?
Use to convert the convention from `org.apache.calcite.plan.Convention.NONE` to `org.voltdb.calciteadapter.rel.logical.VoltDBLRel.VOLTDB_LOGICAL`.
- Why we need VoltDB Logical Rules?
`org.apache.calcite.plan.Convention.NONE` is not implementable, and has to be transformed to something else in order to be implemented. Otherwise, the Volcano Planner will throw a `CannotPlanException`.

note:
I checkout this branch from `ENG-14847`, so merge https://github.com/VoltDB/voltdb/pull/5732 first.


src/frontend/org/voltdb/calciteadapter/rel/logical/VoltDBLAggregate.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    private VoltDBLAggregate(
+            RelOptCluster cluster,
+            RelTraitSet traitSet,
+            RelNode child,
+            boolean indicator,
+            ImmutableBitSet groupSet,
+            List<ImmutableBitSet> groupSets,
+            List<AggregateCall> aggCalls) {
+        super(cluster, traitSet, child, indicator, groupSet, groupSets, aggCalls);
```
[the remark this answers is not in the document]
Chaozz: This is just a confirmation: I can see your comments now :smiley:

src/frontend/org/voltdb/newplanner/rules/logical/VoltDBLSortRule.java
```diff
@@ -0,0 +1,74 @@
...
+        RelNode logicalRel = null;
+        RelCollation sortCollation = sort.getCollation();
+        if (!sortCollation.getFieldCollations().isEmpty()) {
+            logicalRel = new VoltDBLSort(
+                    sort.getCluster(),
+                    convertedTraits,
+                    convertedInput,
+                    sort.getCollation());
+        }
+        // why split it to two phases? why instead of single VoltDBLSort we need VoltDBLSort and VoltDBLLimit
```
alexeevm: We split it into two res mainly for a couple of reasons:
1. During physical phase the VoltDBLSort is transformed into a collation trait that Volcano Planner would propagate  as appropriate -see the VoltDBPSortRule implementation. It won't be possible if the sort would have limit / offset as well
2. The rules for pushing down Limit and Sort RelNodes differ. For example, Limit / Exchange is transformed into Coordinator Limit / Exchange / Fragment Limit while  Sort / Exchange potentially can be transformed to Coordinator Sort / MergeExchange / fragment Sort.
3. The Limit node can be inlined with Scan or Serial Aggregate nodes

tests/frontend/org/voltdb/newplanner/TestLogicalRules.java
```diff
@@ -0,0 +1,427 @@
...
+
+    public void testJoin() {
+        assertPlanMatch("select R1.i, R2.v from R1, R2 " +
+                        "where R2.si = R1.i and R2.v = 'foo'",
+                "VoltDBLCalc(expr#0..11=[{inputs}], expr#12=[CAST($t7):INTEGER], expr#13=[=($t12, $t0)], expr#14=['foo'], expr#15=[=($t11, $t14)], expr#16=[AND($t13, $t15)], I=[$t0], V=[$t11], $condition=[$t16])\n" +
+                        "  VoltDBLJoin(condition=[true], joinType=[inner])\n" +
+                        "    VoltDBLTableScan(table=[[catalog, R1]])\n" +
+                        "    VoltDBLTableScan(table=[[catalog, R2]])\n");
+
+        assertPlanMatch("select R1.i, R2.v from R1 inner join R2 " +
```
alexeevm: Calcite provides several Logical rules related to joins - pushing filters, projections through joins, commuting join order.
What are your thoughts about the phase(s) where the join order permutations should be tried? The problem here is that with a multi-phase approach only a winner of a previous phase proceeds to the next phase. Consider a join of two tables
select * from R inner join RI on R.I = RI.I where RI.C > 0;
where RI has an index on column I.   VoltDB would produce a nested loop index plan where R would be an outer table. With Calcite, if our first logical phase contains the JoinCommuteRule, most likely the (RI , R) join would be a winner over the (R, RI) simply because of an extra filter  RI.C > 0. 
All the next phases would only consider the (RI , R) order even if at the physical stage the (R, RI) order could produce a better plan.
Potentially, we can place the JoinCommuteRule (or its equivalent) to a same phase which introduces index scan(s). The caveat is that the Calcite's JoinCommuteRule matches LogicalJoin only and we would have to rewrite it to target VoltDB joins.

tests/frontend/org/voltdb/newplanner/CalcitePlannerTestCase.java
```diff
@@ -60,10 +61,9 @@ protected CompiledPlan compileAdHocCalcitePlan(String sql,
         CompiledPlan cp = null;
         try {
             cp = m_aide.compileAdHocPlan(PlannerType.CALCITE, sql, inferPartitioning, forcedSP, detMode);
-            assertTrue(cp != null);
+            assertNotNull(cp);
         } catch (Exception ex) {
-            ex.printStackTrace();
-            fail();
+            fail(ex.getMessage());
```
n3world: Why not just let the exception bubble up and fail the test by having the test throw an exception. That way the entire stack trace is captured by junit.

tests/frontend/org/voltdb/newplanner/TestVoltSqlValidator.java
```diff
@@ -102,4 +102,5 @@ public void testType() {
         assertExceptionIsThrown("select ^True or i^ from R2",
                 "Cannot apply 'OR' to arguments of type '<BOOLEAN> OR <INTEGER>'.*");
     }
+
```
n3world: Random change line insert?

tests/frontend/org/voltdb/newplanner/VoltSqlValidatorTestCase.java
```diff
@@ -67,22 +66,19 @@ protected SqlNode parseAndValidate(String sql) {
         return m_validator.validate(sqlNode);
     }
 
-    private void checkParseEx(Throwable e, String expectedMsgPattern, String sql) {
-        try {
-            throw e;
-        } catch (SqlParseException spe) {
-            String errMessage = spe.getMessage();
+    private RuntimeException mapException(Throwable e, String expectedMsgPattern, String sql) {
```
n3world: Why not only accept SqlParseException? The only change would be the catch blocks would be SqlParseException instead of Throwable.

tests/frontend/org/voltdb/planner/PlannerTestAideDeCamp.java
```diff
@@ -159,13 +159,23 @@ public CompiledPlan compileAdHocPlan(PlannerType plannerType, String sql, boolea
                 partitioning, hsql, estimates, false,
                 costModel, null, joinOrder, detMode, false)) {
 
-            if (PlannerType.VOLTDB == plannerType) {
-                planner.parse();
-                plan = planner.plan();
-            } else if (PlannerType.CALCITE == plannerType) {
-                // TODO: plan using the calcite planner
+            switch (plannerType) {
+                case VOLTDB: {
```
n3world: The parenthesis are unnecessary

src/frontend/org/voltdb/calciteadapter/rel/logical/VoltDBLLimit.java
```diff
@@ -0,0 +1,83 @@
...
+    private RexNode m_limit;
+
+    public VoltDBLLimit(
+            RelOptCluster cluster,
+            RelTraitSet traitSet,
+            RelNode input,
+            RexNode offset,
+            RexNode limit) {
+            super(cluster, traitSet, input);
+            assert VoltDBLRel.VOLTDB_LOGICAL.equals(getConvention());
```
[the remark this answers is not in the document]
Chaozz: I prefer `assert traitSet.contains(VoltDBLRel.VOLTDB_LOGICAL);`
I have  updated the asserts to all VoltDBLRels

src/frontend/org/voltdb/calciteadapter/rel/logical/VoltDBLAggregate.java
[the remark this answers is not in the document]
Chaozz: done

src/frontend/org/voltdb/newplanner/rules/logical/VoltDBLSortRule.java
Chaozz: Thanks for your detailed explanation!

src/frontend/org/voltdb/calciteadapter/rel/logical/VoltDBLJoin.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Set;
+
+/**
+ * Logical Join with <code>VOLTDB_LOGICAL</code> convention trait.
+ *
+ * @author Chao Zhou
+ * @since 8.4
+ */
+public class VoltDBLJoin extends Join implements VoltDBLRel {
+    private final boolean semiJoinDone;
```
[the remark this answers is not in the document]
Chaozz: Refer to `computeDigest()` in `AbstractRelNode`, it will call its `explainTerms()`.
I think since I have overrided the `explainTerms()`, the `computeDigest()` should be fine.

tests/frontend/org/voltdb/newplanner/TestLogicalRules.java
Chaozz: Learned!
For `JoinCommuteRule`, its default `JoinCommuteRule.INSTANCE` matches `LogicalJoin`. We can change it to `Join` using the public constructor: `JoinCommuteRule(Join.class, RelFactories.LOGICAL_BUILDER, swapOuter)`

Chaozz: And for the rules that pushing filters, projections through joins, it seems not related to the index right?
I am thinking of introducing them to this logical plan stage.

tests/frontend/org/voltdb/newplanner/CalcitePlannerTestCase.java
Chaozz: Right, removed the try block.

tests/frontend/org/voltdb/newplanner/VoltSqlValidatorTestCase.java
Chaozz: I checked the Throwable type, if it is not `SqlParseException`, a `runtimeException` will return.

tests/frontend/org/voltdb/planner/PlannerTestAideDeCamp.java
Chaozz: It is unnecessary but it won't hurt. Just for helping me to read the code.

tests/frontend/org/voltdb/newplanner/TestLogicalRules.java
alexeevm: Actually, pushing filters down is related to indexes - for the current VoltDBPCalcScanToIndexRule rule to transform a sequential scan into an index scan(s) the VoltDBPCalc must be positioned directly above the VoltDBPTableSeqScan. But this happens later at the physical phase.
I agree with you that pushing projections, filters down should happen at the logical phase.
Permuting a join order could be actually done at both, logical and physical, phases. Initially the best order would be identified at the logical phase using Calcite's logical rules and then reevaluated if there is a potential for a nest loop index join at the physical phase.
The benefit of the two phases approach is a smaller planner's state at each phase

PR #5742, opened 2018-11-12
ProcedureRunnerNT: Consistent synchronization

n3world:
The allHost* member variables were sometimes protected by a synchronize on
m_allHostCallbackLock and sometimes on this. Update the locking to be
consistently on m_allHostCallbackLock.

Use compareAndSet to set m_outstandingAllHostProc so that way there is
race between checking the state and setting it.

### DIFF
--- a/src/frontend/org/voltdb/ProcedureRunnerNT.java
+++ b/src/frontend/org/voltdb/ProcedureRunnerNT.java
@@ -86,6 +86,12 @@ public class ProcedureRunnerNT {
     protected String m_appStatusString = null;
     protected final boolean m_isAdmin;
 
+    // Track the outstanding all host procedures
+    private final Object m_allHostCallbackLock = new Object();
+    private Set<Integer> m_outstandingAllHostProcedureHostIds;
+    private Map<Integer, ClientResponse> m_allHostResponses;
+    private CompletableFuture<Map<Integer, ClientResponse>> m_allHostFut;
+
     ProcedureRunnerNT(long id,
                       AuthUser user,
                       Connection ccxn,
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/ProcedureRunnerNT.java
```diff
@@ -283,21 +285,18 @@ public synchronized void allHostNTProcedureCallback(ClientResponse clientRespons
...
-        Set<Integer> liveHostIds = null;
-
         // hold this lock while getting the count of live nodes
         // also held when
+        long[] hsids;
         synchronized(m_allHostCallbackLock) {
             // collect the set of live client interface mailbox ids
-            liveHostIds = VoltDB.instance().getHostMessenger().getLiveHostIds();
-            m_outstandingAllHostProcedureHostIds = liveHostIds;
+            m_outstandingAllHostProcedureHostIds = VoltDB.instance().getHostMessenger().getLiveHostIds();
```
manjujames: Was the original code trying to reduce the amount time for which we hold the lock? Was there anything wrong in the way it was done?

n3world: I am not sure why the stream was outside of the lock. In theory yes in practice no. The call to stream is an iteration over a non concurrent Set which means that it could throw an Concurrent modification exception if any modifications occur on `m_outstandingAllHostProcedureHostIds`. This couldn't really happen because nothing should try and modify m_outstandingAllHostProcedureHostIds until after the send call is done. In theory with the way this class is written you could just remove the synchronization in this method because it is protected by the compareAndSet at the top. I just decide if we are going to bother with synchronization lets make it consistent.

I did trip over a concurrent modification when I moved the iteration of `m_outstandingAllHostProcedureHostIds`, which is how I even noticed that this was being done outside the protection of any lock and the inconsistent object used for synchronization.

manjujames: I see. Thanks.

PR #5743, opened 2018-11-12
Log a stacktrace as a single printf

n3world:
Since stack traces are written as multiple printf statements their
printing could be interrupted by other log statements. To prevent this
from happening build a single string stack trace and log that in one
printf.

Reduce the implementations of rendering a stack trace down to one method
which operates on an ostream.

### DIFF
--- a/src/ee/common/StackTrace.h
+++ b/src/ee/common/StackTrace.h
@@ -20,29 +20,42 @@
 #include <vector>
 #include <cstdio>
 #include <string>
+#include <iostream>
+#include <sstream>
 
 namespace voltdb {
 
 class StackTrace {
 public:
-    StackTrace();
+    // By default do not include the frame with the constructor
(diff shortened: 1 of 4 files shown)


src/ee/common/StackTrace.h
```diff
@@ -20,29 +20,42 @@
...
+#include <iostream>
+#include <sstream>
 
 namespace voltdb {
 
 class StackTrace {
 public:
-    StackTrace();
+    // By default do not include the frame with the constructor
+    StackTrace(uint32_t skipFrames = 1);
```
JadenW: Is there a case constructor frame won't be skipped? (I assume skipFrames always > 0)

n3world: If you set skipFrames = 0 then the constructor frame will be in the stack trace, otherwise any number greater than 0 will skip this constructor plus other frames.

PR #5744, opened 2018-11-12
ENG-14553: fix avoiding deadlock on MP RO transaction

rdykiel:
In this fix we release the pending buffered reads regardless of our partition leadership status. Therefore a leader losing leadership can still reply to pending MP RO transactions and avoid deadlock.


zeemanhe:
The fix should be back ported to 8.3

PR #5746, opened 2018-11-14
ENG-14925 Scoreboard refactoring to accommodate more than 2 transactions

zeemanhe:
The scoreboard for SRT can hold complete transaction messages from 2 transactions. However the  complete transaction messages could come from more than 2 transactions.  For example, node failure occurs when t1 is processing CompleteTransactionMessage, t2 is processing FragmentTaskMessage. t2 is a non-restartable system store procedure. When repair process kicks in,  CompleteTransactionMessages for both t1 and t2 will be broadcasted.  Since t2 is the ongoing transaction, t2 will be restarted, and then immediately aborted since it is non-restartable.  Then  t3 will kick in and will send out CompleteTransactionMessage if restarted.  Thus 3 CompleteTransactionMessages will reach the scoreboard but one of them will be ignored since the scoreboard can only take 2.  Under this circumstance, t3 will hit deadlock.
This issue is addressed in this pull request.

### DIFF
--- a/src/frontend/org/voltdb/iv2/Scoreboard.java
+++ b/src/frontend/org/voltdb/iv2/Scoreboard.java
@@ -58,7 +58,6 @@ public class Scoreboard {
             m_compTasks.put(task.getMsgTxnId(), Pair.of(task, missingTxn));
         } else {
             if ( task.getTimestamp() > pair.getFirst().getTimestamp() && isComparable(pair.getFirst(), task)) {
-                m_compTasks.remove(task.getMsgTxnId());
                 m_compTasks.put(task.getMsgTxnId(), Pair.of(task, missingTxn));
             }
         }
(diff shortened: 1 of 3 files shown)


rdykiel:
I'm ok with code but I see no new unit test cases to test > 2 transactions?

src/frontend/org/voltdb/iv2/Scoreboard.java
```diff
@@ -53,57 +53,59 @@ public void addCompletedTransactionTask(CompleteTransactionTask task, Boolean mi
...
-        return m_compTasks;
+    /**
+     * Remove the CompleteTransactionTask from the head and count the next CompleteTransactionTask
+     * @param nextTaskCounter CompletionCounter
+     * @return the removed CompleteTransactionTask
+     */
+    public Pair<CompleteTransactionTask, Boolean> pollFirstCompletionTask(CompletionCounter nextTaskCounter) {
+
+        // remove from the head
+        Pair<CompleteTransactionTask, Boolean> pair = m_compTasks.pollFirstEntry().getValue();
```
luyangco: Better to guard against empty map or at least add an assert.

PR #5751, opened 2018-11-16
ENG-11349: Make PostgreSQL, not HSQL, the default sqlcov comparison DB

johnTcrawford:
Added, to SqlCoverage, a new 'hsql' option (in both sql_coverage_test.py
and build.xml), so you can still use HSQL as your comparison database if
you wish; but made 'postgresql' the default option (though not
'postgis', FYI).

### DIFF
--- a/tests/sqlcoverage/sql_coverage_test.py
+++ b/tests/sqlcoverage/sql_coverage_test.py
@@ -254,8 +254,7 @@ def ignore_known_mismatches(comparison_database, suite_name, reproducer):
     to avoid producing lots of meaningless reproducer*.html files.
     """
     if (comparison_database.startswith('Post') and reproducer == Reproduce.DML and
-            suite_name in ['basic-joins', 'basic-index-joins', 'basic-compoundex-joins',
-            'basic-int-joins', 'joined-matview-default-full', 'joined-matview-int']):
+            suite_name in ['joined-matview-default-full', 'joined-matview-int']):
         reproducer = Reproduce.NONE
     return reproducer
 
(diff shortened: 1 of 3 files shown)


tests/sqlcoverage/sql_coverage_test.py
```diff
@@ -254,8 +254,7 @@ def ignore_known_mismatches(comparison_database, suite_name, reproducer):
     to avoid producing lots of meaningless reproducer*.html files.
     """
     if (comparison_database.startswith('Post') and reproducer == Reproduce.DML and
-            suite_name in ['basic-joins', 'basic-index-joins', 'basic-compoundex-joins',
-            'basic-int-joins', 'joined-matview-default-full', 'joined-matview-int']):
+            suite_name in ['joined-matview-default-full', 'joined-matview-int']):
```
kaikaizi: PostgreSQL dropped 4 test suites. Is that on purpose?

johnTcrawford: Yes, this is actually related to ENG-10775, but I missed it then (not a serious problem, BTW), and stuck it in here: this is a list of test suites whose detailed results should be deleted (they were taking up too much space in Jenkins, see ENG-14543), but ENG-10775 fixed these test suites, so any failed test results should not be deleted.

tests/test-tools.sh
```diff
@@ -358,11 +523,13 @@ function test-tools-all-pro() {
 }
 
 # Print a simple help message, describing the options for this script
-function test-tools-help() {
-    echo -e "\nUsage: ./test-tools.sh test-tools-{build|build-pro|init|debug|server|all|shutdown|all|help}\n"
-    echo -e "This script is mainly intended to provide useful functions that may be called"
-    echo -e "  by a variety of other test scripts, e.g., <voltdb>/tests/sqlgrammar/run.sh,"
-    echo -e "  but it may also be called directly on the command line."
+function tt-help() {
+    echo -e "\nUsage: ./test-tools.sh test-tools-{build[-pro]|init|debug|server[-pro]|all[-pro]|shutdown}"
```
kaikaizi: For such chunk of message to print, I would prefer doing it like:
```suggestion
    echo -e <<EOF
    
Usage: ./test-tools.sh test-tools-{build[-pro]|init|debug...
...
but options usually call other options that are prerequisites.

EOF
```

PR #5755, opened 2018-11-19
ENG-14934-index-unary-minus

yzhang1991:
Now you can create an index on an expression with a unary minus operator.


src/frontend/org/voltdb/expressions/OperatorExpression.java
```diff
@@ -192,6 +192,9 @@ public String explain(String impliedTableName) {
                     m_right.m_left.explain(impliedTableName) + " ELSE " +
                     m_right.m_right.explain(impliedTableName) + " END)";
         }
+        if (type == ExpressionType.OPERATOR_UNARY_MINUS) {
```
Chaozz: @kaikaizi did some refactoring on this class in [ENG-14621] CREATE TABLE through Calcite!. (if else -> switch)
It is on `calcite` branch but I believe we can bring it to master at this moment.
https://github.com/VoltDB/voltdb/blob/b682ecd11d2f5b01bf8ed55e4494c9dd358ea314/src/frontend/org/voltdb/expressions/OperatorExpression.java#L190

PR #5758, opened 2018-11-19
ENG-14923, better naming for @Statistics EXPORT, add LastQueuedTimest…

luyangco:
…amp and LastAckedTimestamp column.

Change-Id: I6f3f3953b35de1669d696684a320ac90e1b29fbe

### DIFF
--- a/src/ee/storage/ExportTupleStream.cpp
+++ b/src/ee/storage/ExportTupleStream.cpp
@@ -39,20 +39,20 @@ const size_t ExportTupleStream::s_mdSchemaSize = (19 + 21 + 27 + 17 + 12 + 21 //
                                                                 + ExportTupleStream::METADATA_COL_CNT // Volt Type byte
                                                                 + (ExportTupleStream::METADATA_COL_CNT * sizeof(int32_t)) // Int for column names string size
                                                                 + (ExportTupleStream::METADATA_COL_CNT * sizeof(int32_t))); // column length colInfo->length
-const size_t EXPORT_BUFFER_HEADER_SIZE = 12; // row count(4) + uniqueId(8)
+const size_t ExportTupleStream::s_EXPORT_BUFFER_HEADER_SIZE = 12; // row count(4) + uniqueId(8)
 const size_t ExportTupleStream::s_FIXED_BUFFER_HEADER_SIZE = 13; // Size of header before schema: Version(1) + GenerationId(8) + SchemaLen(4)
 const uint8_t ExportTupleStream::s_EXPORT_BUFFER_VERSION = 1;
 
 ExportTupleStream::ExportTupleStream(CatalogId partitionId, int64_t siteId, int64_t generation, std::string signature,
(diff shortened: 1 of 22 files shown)


src/ee/storage/ExportTupleStream.cpp
```diff
@@ -172,6 +172,8 @@ size_t ExportTupleStream::appendTuple(int64_t lastCommittedSpHandle,
     const size_t startingUso = m_uso;
     m_uso += (streamHeaderSz + io.position());
     m_exportSequenceNumber++;
+    // TODO: better to only add this before commit
```
rdykiel: Why is this still a TODO? Is there something else we should do here?

src/frontend/org/voltdb/ExportStatsBase.java
```diff
@@ -25,34 +25,36 @@
...
+            m_sourceName = sourceName;
             m_exportTarget = exportTarget;
+            m_isExporting = isExporting ? "TRUE" : "FALSE";
             m_tupleCount = tupleCount;
             m_tuplesPending = tuplesPending > 0 ? tuplesPending : 0;
+            m_lastQueuedTimestamp = lastQueuedTimestamp;
+            m_lastAckedTimestamp = lastAckedTimestamp;
             m_averageLatency = averageLatency;
             m_maxLatency = maxLatency;
+            m_missingTuple = missingTuple;
```
rdykiel: should it be something like "gapCount" ?

tests/frontend/org/voltdb/export/TestExportDataSource.java
```diff
@@ -214,23 +214,23 @@ public void run() {
             int buffSize = 20 + StreamBlock.HEADER_SIZE;
             ByteBuffer foo = ByteBuffer.allocateDirect(buffSize);
             foo.duplicate().put(new byte[buffSize]);
-            s.pushExportBuffer(1, 1, foo, false);
+            s.pushExportBuffer(1, 1, 0, foo, false);
```
rdykiel: You should be consistent: some place you're using "L" and some not.

src/ee/storage/ExportTupleStream.cpp
luyangco: Removed TODO. When I marked this as TODO I was thinking it's nice to have export stream that doesn't reply on next insert to trigger the commit. But this is beyond the scope of current ticket.

src/frontend/org/voltdb/ExportStatsBase.java
luyangco: Changed.

tests/frontend/org/voltdb/export/TestExportDataSource.java
luyangco: Changed.

luyangco:
It turns out the order of first a few columns that the scripts care about is not changed, so I didn't modify those files.

PR #5763, opened 2018-11-20
ENG-14744: ensure repair messages are not checked for misrouted by re…

rdykiel:
…plicas

### DIFF
--- a/src/frontend/org/voltdb/iv2/SpPromoteAlgo.java
+++ b/src/frontend/org/voltdb/iv2/SpPromoteAlgo.java
@@ -26,6 +26,7 @@ import java.util.Map.Entry;
 import java.util.TreeSet;
 import java.util.concurrent.Future;
 
+import org.voltcore.messaging.TransactionInfoBaseMessage;
 import org.voltcore.messaging.VoltMessage;
 import org.voltcore.utils.CoreUtils;
 import org.voltdb.messaging.Iv2RepairLogRequestMessage;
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/iv2/SpPromoteAlgo.java
```diff
@@ -251,6 +252,11 @@ public void repairSurvivors()
                 }
             }
             if (!needsRepair.isEmpty()) {
+                // ENG-14744: ensure replicas receiving those messages won't check for misrouted messages
+                if (li.getPayload() instanceof TransactionInfoBaseMessage) {
+                    TransactionInfoBaseMessage tibm = (TransactionInfoBaseMessage) li.getPayload();
+                    tibm.setForReplica(true);
```
luyangco: I don't understand the logic here. I understand marking message as `ForReplica` would bypass misrouted message check, but does it mean this message will skip partition leader too? How could it not break repair?
If I missed anything, please correct me. But it's worthy to update the comment with the answer to my question.

rdykiel: Isn't repair always originating from partition leader? Zeeman can you chime in?

zeemanhe: SP repair can only occur during leader promotion. repair is done before a leader for the partition is promoted.

PR #5764, opened 2018-11-20
ENG-12409: Shutdown client interface after dr

n3world:
When the ClientInterface is shutdown before dr the responses to the requests which the consumer has sent out will never be delivered which can block a graceful shutdown which waits for outstanding requests.


manjujames:
I don't see changes in here that changes anything to wait for PBRs to drain? The only kind of substantial change I see is that you moved clientInterface shutdown to after shutting down dr consumer.

manjujames:
Oh never mind. That is a separate PR. This one looks OK to me.

PR #5765, opened 2018-11-20
ENG-14430: Do not log accept ioexceptions at error

n3world:


src/frontend/org/voltcore/messaging/SocketJoiner.java
```diff
@@ -562,16 +563,18 @@ private void processSSC(ServerSocketChannel ssc) throws Exception {
                 } else {
                     throw new RuntimeException("Unexpected message type " + type + " from " + remoteAddress);
                 }
-            } catch (Exception ex) {
+                success = true;
+            } catch (IOException e) {
+                LOG.info("IOException occurred while processing new socket " + sc, e);
```
manjujames: We should probably also NOT log the stack trace. Only log the exception message. Also, log message is somewhat unclear as to whether we think this is a problem or not. May be say something to the effect that an exception occurred when another node was trying to connect with this one, but will probably retry.

We could also log the exception stack trace in debug level.

n3world: Updated the message and only log stack trace if debug is enabled.

PR #5766, opened 2018-11-20
E3 datasource

rdykiel:
Made the changes to support E3 versions for ExportGeneration and ExportDataSource; I backed out from trying to have a common implementation of ExportGeneration given the tight coupling with ExportDataSource.


luyangco:
Without looking inside your pull request (will do it later), I wonder if it's possible to get rid of export generation in E3 implementation, since E2 already has most parts of its implementation to be generation-less.

rdykiel:
@luyangco I thought of it but it's not my short term priority: short term is to finalize and merge the tickets I have been assigned several weeks ago. I suggest we add a ticket for this at the next grooming session.

PR #5774, opened 2018-11-26
Calcite DQL

kaikaizi:
Excercise Calcite parsing/planning for all ad-hoc DQL queries. Since
planning is not supported, the PlannerTool.planSqlCalcite() method will
just exercise to the point that Calcite planning is built, then return
null, and its caller chain will use fall back behavior to plan the
query.

Based off my latest, un-merged dev branch. Changes/fixes of indenting 
in file `template.classpath`.


kaikaizi:
Holding back this as Jenkins is failing hundreds of tests. Will reopen once resolved.

PR #5775, opened 2018-11-27
Eng 14941 export blocked

luyangco:
So far I identified two problems: 1) auto release feature checks cluster completeness meanwhile gap detection/resolution checks export mailboxes list, so it’s possible that before gap resolution gets a chance to query other nodes, cluster completeness check is passed. This causes premature gap release which lead to missing rows.
2) There is a race window between updating export mailboxes list and receiving query message, if query message is received before export mailboxes update, it causes blocked stream failed to find missing data on other nodes.

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -1436,7 +1436,7 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
                     return;
                 }
                 if (exportLog.isDebugEnabled()) {
-                    exportLog.debug(ExportDataSource.this.toString() + " decides to claw mastership back from other node.");
+                    exportLog.debug(ExportDataSource.this.toString() + " is going to export data because partition leader is on current node.");
                 }
                 // Query export membership if current stream is not the master
                 sendTakeMastershipMessage();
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1499,14 +1497,17 @@ public void run() {
                             if (voltdb.isClusterComplete()) {
                                 if (DISABLE_AUTO_GAP_RELEASE) {
                                     // Show warning only in full cluster.
-                                    exportLog.warn(ExportDataSource.this.toString() + " is blocked because stream hits a gap from sequence number " +
-                                            gap.getFirst() + " to " + gap.getSecond());
+                                    exportLog.warn("Export is blocked, missing [" + gap.getFirst() + ", " + gap.getSecond() + "] from " +
+                                            ExportDataSource.this.toString() + ". Please rejoin a node with the missing export queue data or " +
+                                            "use 'voltadmin release' command to skip the missing data.");
+                                } else {
+                                    releaseBlock();
```
zeemanhe: can call  processStreamControl(OperationMode.RELEASE) directly. eliminate releaseBlock().

luyangco: Replaced releaseBlock() with processStreamControl(). The reason to move the logic here is because every cluster member change/partition leader promotion will trigger sendGapQuery(), so it's safe to release blocked stream at handleQueryResponse.

PR #5776, opened 2018-11-28
ENG-14816: Use netty ssl context

n3world:
Use netty for ssl since it can use openssl if it is installed on the local system.


src/frontend/org/voltcore/utils/ssl/SSLBufferDecrypter.java
```diff
@@ -98,4 +101,65 @@ public int tlsunwrap(ByteBuffer srcBuffer, ByteBuffer dstBuffer) {
...
+     * Encrypt data in {@code srcBuffer} into {@code dstBuf} if it is large enough. If an error occurs {@code dstBuf}
+     * will be released. If {@code dstBuf} is not large enough a new buffer will be allocated from {@code allocator} and
+     * {@code dstBuf} will be released.
+     *
+     * @param srcBuffer holding encrypted data
+     * @param dstBuf    to attempt to write plain text data into
+     * @param allocator to be used to allocate new buffers if {@code dstBuf} is too small
+     * @return {@link ByteBuf} containing plain text data
+     */
+    public ByteBuf tlsunwrap(ByteBuffer srcBuffer, ByteBuf dstBuf, PooledByteBufAllocator allocator) {
```
manjujames: We don't use this method yet, correct? Other than the one in this class itself that calls this with app buffer size.

n3world: It is used in one place other than by the same class

PR #5777, opened 2018-11-28
Eng 15004 investigation notest (NOT FOR MERGING)

rdykiel:
NOT FOR MERGING

Can you review the instrumentation in ExportDataSource.java only and verify if I made a gross mistake. If not, then ENG-15004 may be caused by what is reported in method checkContinuity(), or it's a red herring and the root cause is something else.

What is disturbing is that the 'rowCount' variable consistently reports the exact number of rows that the CLIENT test progam claims to have inserted in the export stream.

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -168,8 +168,8 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
     static public AtomicLong rowCount = new AtomicLong(0);
     static public AtomicLong missCount = new AtomicLong(0);
     static public AtomicLong overCount = new AtomicLong(0);
-    private volatile long m_prevSeqNo = 0L;
-    private volatile int m_prevTuplesCount = 0;
+    private AtomicLong m_prevSeqNo = new AtomicLong(0);
+    private AtomicLong m_prevTuplesCount = new AtomicLong(0);
 
     static enum StreamStatus {
         ACTIVE,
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1006,6 +1017,36 @@ private synchronized void pollImpl(SettableFuture<AckingContainer> fut) {
         }
     }
 
+    // make this synchronized to avoid questions
+    private synchronized void checkContinuity(AckingContainer cont) {
```
zeemanhe: synchronized not needed

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1006,6 +1017,36 @@ private synchronized void pollImpl(SettableFuture<AckingContainer> fut) {
...
+                        + ", cur.tuplesCount: " + cont.m_tuplesCount
+                        + ", missing by: " + miss);
+            }
+        }
+        m_prevSeqNo = cont.m_seqNo;
+        m_prevTuplesCount = cont.m_tuplesCount;
+    }
+
+
     private void releaseBlock() {
```
zeemanhe: rebase with master to get the changes Yang committed today

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1004,9 +1015,39 @@ private synchronized void pollImpl(SettableFuture<AckingContainer> fut) {
         }
     }
 
+    // make this synchronized to avoid questions
+    private synchronized void checkContinuity(AckingContainer cont) {
+        if (m_prevSeqNo.get() != 0L) {
+            long nextSeqNo = m_prevSeqNo.get() + m_prevTuplesCount.get();
```
luyangco: Here is the problem, AckingContainer records the last sequence number and the row count of the buffer, so nextSeqNo should be m_prevSeqNo.get() + 1. The reason to remember the last sequence number instead of the start is because when EDS acks buffer it only cares about the last sequence number.

rdykiel: No this is not an issue of the instrumentation because seqNo + tuplesCount is lastsequencenumber + 1. The discrepancies detected are variable and always > 1

rdykiel:
UPDATE: I pushed a new version of the instrumentation which seems to prove that the Kafka client is losing messages. I'll put a more detailed analysis in https://issues.voltdb.com/browse/ENG-15004

PR #5781, opened 2018-11-28
ENG-14933: voltsql blank line reacts slowly

Chaozz:
immediately return when the input is an empty line instead of sending it to voltdb server

### DIFF
--- a/lib/python/voltsql/voltsql/voltsql.py
+++ b/lib/python/voltsql/voltsql/voltsql.py
@@ -202,17 +202,18 @@ class VoltCli(object):
             except EOFError:
                 break
             else:
-                if sql_cmd.lower() == "refresh":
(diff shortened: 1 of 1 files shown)


lib/python/voltsql/voltsql/voltsql.py
```diff
@@ -212,6 +212,9 @@ def run_cli(self):
                 if sql_cmd.strip().lower() in ("help", "help;"):
                     print(README)
                     continue
+                if not sql_cmd.strip():
```
kaikaizi: Not related: 
L205, do you need `.strip()`?
You can have an auxiliary method to check w/o semicolon.

Chaozz: done

PR #5782, opened 2018-11-29
ENG-15021, add logging when gaps are skipped or covered.

luyangco:
Change-Id: Ia30e27b2d1d24f2098b165a246ed2bb8c7e9d5fc

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -423,6 +423,7 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
         if (m_status == StreamStatus.BLOCKED &&
                 m_gapTracker.getFirstGap() != null &&
                 releaseSeqNo > m_gapTracker.getFirstGap().getSecond()) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -423,6 +423,7 @@ private synchronized void releaseExportBytes(long releaseSeqNo, int tuplesSent)
         if (m_status == StreamStatus.BLOCKED &&
                 m_gapTracker.getFirstGap() != null &&
                 releaseSeqNo > m_gapTracker.getFirstGap().getSecond()) {
+            exportLog.info("Export queue gap resolved. Resuming export for " + ExportDataSource.this.toString());
```
zeemanhe: use this instead of ExportDataSource.this.toString()

luyangco: It runs in a runnable, I'm afraid `this.toString()` would point to `Object.toString()`

zeemanhe: ok

PR #5788, opened 2018-12-03
Eng 15019 incorrect stat

luyangco:
This patch includes five changes:
1) completely reply on export sequence tracker to compute pending data,
2) fix a race condition that sometimes replica stream fails to retrieve local mailbox to send back gap query response,
3) fix the accounting problem in export sequence tracker itself,
4) fix the wrong ‘blocked’ status in some cases for active non-blocking stream.
5) fix a pre-8.0 bug that in `m_dataSourcesByPartition` it uses table signature as key but in snapshot digest the key of exportSequence map is table name, causing truncateToSeqNo never find the right stream.
Pro PR: https://github.com/VoltDB/pro/pull/2416

### DIFF
--- a/src/frontend/org/voltdb/export/ExportDataSource.java
+++ b/src/frontend/org/voltdb/export/ExportDataSource.java
@@ -438,8 +438,7 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
             lastSeqNo = releaseSeqNo;
         }
         m_firstUnpolledSeqNo = Math.max(m_firstUnpolledSeqNo, lastSeqNo + 1);
-        int pendingCount = m_tuplesPending.get();
-        m_tuplesPending.set(pendingCount - tuplesSent);
+        m_tuplesPending.addAndGet(-tuplesSent);
 
         return;
     }
@@ -1007,14 +1006,14 @@ public class ExportDataSource implements Comparable<ExportDataSource> {
     }
 
     public class AckingContainer extends BBContainer {
-        final long m_seqNo;
+        final long m_lastSeqNo;
         final int m_tuplesCount;
         final BBContainer m_backingCont;
         long m_startTime = 0;
 
         public AckingContainer(BBContainer cont, long seq, int tuplesSent) {
             super(cont.b());
-            m_seqNo = seq;
+            m_lastSeqNo = seq;
             m_tuplesCount = tuplesSent;
             m_backingCont = cont;
         }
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1007,14 +1013,14 @@ private synchronized void pollImpl(SettableFuture<AckingContainer> fut) {
     }
 
     public class AckingContainer extends BBContainer {
-        final long m_seqNo;
+        final long m_lastSeqNo;
```
rdykiel: Thanks :-)

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -419,27 +429,23 @@ private synchronized void releaseExportBytes(long releaseSeqNo, int tuplesSent)
...
+                    m_firstUnpolledSeqNo + ", tracker map is " + m_gapTracker.toString());
         }
-        // If releaseSeqNo falls in between stream buffers,
-        if (m_committedBuffers.isEmpty()) {
-            lastSeqNo = releaseSeqNo;
-        }
-        m_firstUnpolledSeqNo = Math.max(m_firstUnpolledSeqNo, lastSeqNo + 1);
-        int pendingCount = m_tuplesPending.get();
-        m_tuplesPending.set(pendingCount - tuplesSent);
+        m_tuplesPending.addAndGet(-tuplesSent);
```
zeemanhe: use decrementAndGet() instead

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -401,13 +401,23 @@ private synchronized void releaseExportBytes(long releaseSeqNo, int tuplesSent)
             tuplesSent = 0;
         }
 
-        long lastSeqNo = 0;
+        long lastSeqNo = releaseSeqNo;
```
rdykiel: We're using several different intermediate variables; I think this method would benefit from more comments explaining the intended behavior. This could miminize the risk of unintended maintenance errors.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: decrementAndGet() only decreases the current value by one.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Will do.

src/frontend/org/voltdb/export/ExportDataSource.java
zeemanhe: Oops, my bad. Regarding  the decrement of m_tuplesPending by tuplesSent , should tuplesSent be adjusted by StreamBlock released or discarded? That is, count the rows from the StreamBlocks which are discarded/released.

luyangco: This is addressed in later commits.

src/frontend/org/voltdb/export/ExportSequenceNumberTracker.java
```diff
@@ -110,27 +112,43 @@ public void append(long start, long end) {
...
+     * @return number of sequence be truncated
      */
-    public void truncate(long newTruncationPoint) {
-        if (size() == 0) {
+    public int truncate(long newTruncationPoint) {
+        int truncated = 0;
+        if (m_map.isEmpty()) {
             m_map.add(range(newTruncationPoint, newTruncationPoint));
-            return;
+            m_hasSentinel = true;
```
rdykiel: I find it very strange that m_sentinel is never reset to false and am concerned that we will count it more than once.

PR #5790, opened 2018-12-06
ENG-14559:

manjujames:
Deserialize params immediately in initFromBuffer because
sometimes the buffers are returned back to the pool before the bytes are used for deserialization.

### DIFF
--- a/src/frontend/org/voltdb/StoredProcedureInvocation.java
+++ b/src/frontend/org/voltdb/StoredProcedureInvocation.java
@@ -277,6 +277,8 @@ public class StoredProcedureInvocation implements JSONString {
                 return ParameterSet.fromByteBuffer(duplicate);
             }
         });
+        params.run(); // do this because in some cases the buffer may be a pooled buffer,
+                      // which may be returned back to the pool before we deserialize.
     }
 
     @Override


src/frontend/org/voltdb/StoredProcedureInvocation.java
```diff
@@ -277,6 +277,8 @@ public ParameterSet call() throws Exception {
                 return ParameterSet.fromByteBuffer(duplicate);
             }
         });
+        params.run(); // do this because in some cases the buffer may be a pooled buffer,
```
n3world: Wouldn't it be a bit simpler as:
```
ParameterSet parameterSet = ParameterSet.fromByteBuffer(serializedParams);
params = new FutureTask<>(() -> parameterSet);
```

wweiss-voltdb: It would be better to remove the future altogether. Since we are incurring the cost of the deserialize, we might as well remove the added cost of the future and the synchronization overhead associated with the call/get.

manjujames: Nate, it is Java 1.7 with Volt 4.9.

manjujames: Walter, the future is set in other paths too. So it is much more work to remove it altogether and definitely not worth it for Volt 4.9. May be something we can look into for the current version.

wweiss-voltdb: Then if you want to hack something to solve the problem for 4.9, I think you should isolate the change to the tasklog path. You can force the param expansion in replayFromTaskLog when you process a fragmenttaskmessage that includes an initiateTaskMessage in it.

manjujames: That was my first attempt, but that still gave the error. So there might have been paths I missed. So, I went this fix.

When I said params are set in other paths too, I meant setParams path. That is not something I want to touch at this poing. It is really not worth it to spend a lot of time on this for 4.9, especially because the code looks very different now.

PR #5800, opened 2018-12-13
Basically enable the DQL code path through Calcite

kaikaizi:
With the last switch left in CalciteCheckImpl.doCheck() method.
Also spent some amount of effort on refactoring how PlannerTool plans with caches.
(And fixed a bug on TestAdhocDropTable for the error message mismatch.)

### DIFF
--- a/src/frontend/org/voltdb/compiler/PlannerTool.java
+++ b/src/frontend/org/voltdb/compiler/PlannerTool.java
@@ -208,7 +208,7 @@ public class PlannerTool {
      * Plan a query with the Calcite planner.
      * @param task the query to plan.
      * @param batch the query batch which this query belongs to.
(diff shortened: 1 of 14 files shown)


src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -212,7 +208,8 @@ public synchronized CompiledPlan planSqlCore(String sql, StatementPartitioning p
      * Plan a query with the Calcite planner.
      * @param task the query to plan.
      * @param batch the query batch which this query belongs to.
-     * @return a planned statement.
+     * @return a planned statement. Untill DQL is fully supported, it returns null, and the caller
```
Chaozz: typo: untill -> until

PR #5729, opened 2018-11-06
Eng 14768: System Proc for starting elastic removal

JadenW:

PR #5801, opened 2018-12-14
ENG-14768: System Proc for starting elastic removal

JadenW:


JadenW:
Old reviews has been addressed. https://github.com/VoltDB/voltdb/pull/5729

PR #5802, opened 2018-12-14
 Add a utility class to handle pro classes

n3world:
Currently there is a single method which assits in loading a pro class but that
still requires the caller to retrieve the constructor and do the construction.
The ProClass utility class wraps that and simplifies the error handling for
the common use cases.

### DIFF
--- a/src/frontend/org/voltdb/utils/ProClass.java
+++ b/src/frontend/org/voltdb/utils/ProClass.java
@@ -1,0 +1,261 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2018 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.voltdb.utils;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Executable;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.Objects;
+
+import org.voltcore.logging.VoltLogger;
+import org.voltdb.VoltDB;
+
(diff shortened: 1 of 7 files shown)


src/frontend/org/voltdb/utils/ProClass.java
```diff
@@ -0,0 +1,262 @@
...
+     */
+    public static final ErrorHandler HANDLER_CRASH = (m, t) -> VoltDB.crashLocalVoltDB(m, true, t);
+
+    private final Class<T> m_proClass;
+    private ErrorHandler m_errorHandler = HANDLER_CRASH;
+    private boolean m_loadConstructor = true;
+    private Constructor<T> m_constructor;
+
+    /**
+     * Try to load a PRO class. If it's running the community edition, an error message will be logged if not
```
manjujames: 'if not suppressed'?

src/frontend/org/voltdb/utils/ProClass.java
```diff
@@ -0,0 +1,262 @@
...
+                    + feature + '.', e);
+        }
+
+        return (ProClass<T>) NO_CLASS;
+    }
+
+    /**
+     * Construct a new instance of {@code className} using {@code parameters}. If the class cannot be loaded
+     * {@code null} is returned. If the class cannot be constructed
+     * {@link VoltDB#crashLocalVoltDB(String, boolean, Throwable)} will be invoked with the instantiation exception.
```
manjujames: Need to update the java doc?

src/frontend/org/voltdb/utils/ProClass.java
```diff
@@ -0,0 +1,262 @@
...
+     * @param parameters  to pass to the constructor of {@code className}
+     * @return an instance of {@code T}
+     */
+    public static <T> T newInstanceOf(String className, String feature, ErrorHandler loadFailure,
+            Object... parameters) {
+        return ProClass.<T>load(className, feature, loadFailure).newInstance(parameters);
+    }
+
+    private ProClass(Class<T> proClass) {
+        super();
```
manjujames: Nit: Do we need the super() call here?

manjujames:
Nice work. Thanks for doing this.

src/frontend/org/voltdb/utils/ProClass.java
n3world: Probably not it is just the way eclipse by default auto generates constructors it always puts a call into super();

src/frontend/org/voltdb/utils/ProClass.java
n3world: A little but the part about crash being invoked on failure to instantiate is still true.

src/frontend/org/voltdb/utils/ProClass.java
n3world: Updated to say the errorHandler will be invoked.

PR #5804, opened 2018-12-14
 ENG-14769: voltadmin resize command for test|start|status elastic resizing

JadenW:


lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,82 @@
...
+                              [2, runner.opts.hostOrPartition]).table(0)
+    status = result.tuple(0).column_integer(0)
+    message = result.tuple(0).column_string(1)
+    if status == 0:
+        runner.info(message)
+    else:
+        runner.error(message)
+        sys.exit(1)
+
+def clearup(runner):
```
n3world: Should this be `cleanup`?

lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+import sys
+
+RELEASE_MAJOR_VERSION = 8
+RELEASE_MINOR_VERSION = 5
+
+def test(runner):
+    result = runner.call_proc('@ElasticRemoveNT', [VOLT.FastSerializer.VOLTTYPE_TINYINT, VOLT.FastSerializer.VOLTTYPE_STRING],
```
n3world: The body of this function and most of the others are all identical except for the first parameter to the procedure. It should probably become a utility method that takes in that one parameter

lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,82 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+import sys
+
+RELEASE_MAJOR_VERSION = 8
+RELEASE_MINOR_VERSION = 5
```
n3world: Does the version really have to be defined in this module, it isn't being used?

lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test(runner):
+    result = runner.call_proc('@ElasticRemoveNT', [VOLT.FastSerializer.VOLTTYPE_TINYINT, VOLT.FastSerializer.VOLTTYPE_STRING],
+                              [0, runner.opts.hostOrPartition]).table(0)
```
manjujames: Is there something we can do in python than the hardcoded 0, 1, 2 values in these calls? It is not easily readable what is going on here. May be it will become a bit better with Nate's suggested change.

On the Java side, we should use enums or something instead of the hardcoded numbers, whose meaning are just defined in comments.

lib/python/voltcli/voltadmin.d/resize.py
manjujames: In the sysproc it is clear, so I think @JadenW does mean clearup. May be we should just call it clear or reset?

lib/python/voltcli/voltadmin.d/resize.py
JadenW: Forgot to add the version verification using this. If using new voltadmin targeting to an old votdb, it will reject.

lib/python/voltcli/voltadmin.d/resize.py
JadenW: Agree, will simplify it.

lib/python/voltcli/voltadmin.d/resize.py
JadenW: I thought in ticket it's clear, will rename

lib/python/voltcli/voltadmin.d/resize.py
JadenW: For java side, volt procedure seems cannot take enums as input value type, so maybe still use hardcode numbers with comments?

For client side, I simply the python code with enums and shared function.

lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,92 @@
...
+        runner.error(message)
+        sys.exit(1)
+
+@VOLT.Multi_Command(
+    bundles = VOLT.AdminBundle(),
+    description = 'Elastic resizing cluster command.',
+    options = (
+            VOLT.BooleanOption('-f', '--force', 'forcing', 'bypass precheck', default = False),
+            VOLT.IntegerOption('-t', '--timeout', 'timeout', 'The timeout value in seconds if @Statistics is not progressing.', default = 0),
+            VOLT.StringOption(None, '--hostOrPartition', 'hostOrPartition', 'Specified hosts or partitions to be removed.', default = ''), # TODO: use StringListOptition if hostOrPartition format changes
```
n3world: It might not make sense to put this in until it is supported otherwise it will be displayed in the help and could confuse a user.

lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,92 @@
...
+        runner.info(message)
+    else:
+        runner.error(message)
+        sys.exit(1)
+
+@VOLT.Multi_Command(
+    bundles = VOLT.AdminBundle(),
+    description = 'Elastic resizing cluster command.',
+    options = (
+            VOLT.BooleanOption('-f', '--force', 'forcing', 'bypass precheck', default = False),
```
n3world: It doesn't look like this actually does anything

lib/python/voltcli/voltadmin.d/resize.py
```diff
@@ -0,0 +1,92 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+import sys
+from voltcli.hostinfo import Hosts
+
+# TODO: change to specific version when feature branch merge into master
+RELEASE_MAJOR_VERSION = 8
```
n3world: This should be at least 9.0

PR #5805, opened 2018-12-14
ENG-15030 Query with many joins drains during planning

kaikaizi:
Changed exhaustive planning To periodically estimate heap memory usage
against maximum allowed memory in planning. Stop further plan searching
when the limit is reached.

### DIFF
--- a/src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
+++ b/src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
@@ -19,13 +19,7 @@ package org.voltdb.planner;
 
 import java.util.*;
 
-import org.hsqldb_voltpatches.HSQLInterface;
-import org.voltdb.catalog.MaterializedViewInfo;
 import org.voltdb.catalog.Database;
-import org.voltdb.catalog.Table;
-import org.voltdb.compiler.DatabaseEstimates;
-import org.voltdb.compiler.DeterminismMode;
-import org.voltdb.compiler.VoltCompiler;
 import org.voltdb.expressions.AbstractExpression;
 import org.voltdb.expressions.ExpressionUtil;
 import org.voltdb.expressions.TupleValueExpression;
@@ -63,6 +57,16 @@ public class SelectSubPlanAssembler extends SubPlanAssembler {
     /** The list of all possible join orders, assembled by queueAllJoinOrders */
     private ArrayDeque<JoinNode> m_joinOrders = new ArrayDeque<>();
 
+    private static final long S_TOTAL_JVM_BYTES = Runtime.getRuntime().maxMemory();
+    // Approximate size in bytes of each plan: about 50 KB with some experiments.
+    private static final long S_PLAN_SIZE = 50_000;
+    // Number of times generateSubPlanForJoinNode() gets called recursively that we collect an estimate of heap size,
+    // and early exit if too large heap size had been used.
+    private static final int S_PLAN_ESTIMATE_PERIOD = 100;
+    // Stop generating any further possible plans, if we have reached xx% of available JVM heap memory
+    private static final short S_MAX_HEAP_MEMORY_USAGE_PCT = 98;
+    private static final long S_MAX_ALLOWED_PLAN_MEMORY = (long) (S_TOTAL_JVM_BYTES / 100. * S_MAX_HEAP_MEMORY_USAGE_PCT);
+
     /**
      *
      * @param db The catalog's Database object.
@@ -98,7 +102,7 @@ public class SelectSubPlanAssembler extends SubPlanAssembler {
         ArrayList<List<JoinNode>> joinOrderList = generateJoinOrders(subTrees);
         // Reassemble the all possible combinations of the sub-tree and queue them
         ArrayDeque<JoinNode> joinOrders = new ArrayDeque<>();
-        queueSubJoinOrders(joinOrderList, 0, new ArrayList<JoinNode>(), joinOrders, findAll);
+        queueSubJoinOrders(joinOrderList, 0, new ArrayList<>(), joinOrders, findAll);
         return joinOrders;
     }
 
@@ -156,8 +160,7 @@ public class SelectSubPlanAssembler extends SubPlanAssembler {
                 return generateFullJoinOrdersForTree(subTree);
             } else {
                 // Shouldn't get there
-                assert(false);
-                return null;
+                throw new PlanningErrorException("Internal error: unsupported join type " + joinType.toString());
             }
         } else {
             // Single tables and subqueries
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -538,18 +542,45 @@ private void generateSubPlanForJoinNodeRecursively(JoinNode rootNode,
...
             generateSubPlanForJoinNodeRecursively(rootNode, nextNode+1, nodes);
         }
     }
 
+    /**
+     * Evaluate size of currently generated plans.
+     * @return the total number of plan nodes contained in generated plan candidates.
+     */
+    private long evaluatePlanSize() {
+        return m_plans.stream().mapToLong(SelectSubPlanAssembler::evaluetePlanSize).sum();
```
n3world: Wouldn't it be better to just keep a running total of number of nodes in the tree since they are all being created in this class instead of periodically doing a full tree walk and counting the number of nodes?

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -63,6 +57,16 @@
...
 
+    private static final long S_TOTAL_JVM_BYTES = Runtime.getRuntime().maxMemory();
+    // Approximate size in bytes of each plan: about 50 KB with some experiments.
+    private static final long S_PLAN_SIZE = 50_000;
+    // Number of times generateSubPlanForJoinNode() gets called recursively that we collect an estimate of heap size,
+    // and early exit if too large heap size had been used.
+    private static final int S_PLAN_ESTIMATE_PERIOD = 100;
+    // Stop generating any further possible plans, if we have reached xx% of available JVM heap memory
+    private static final short S_MAX_HEAP_MEMORY_USAGE_PCT = 98;
+    private static final long S_MAX_ALLOWED_PLAN_MEMORY = (long) (S_TOTAL_JVM_BYTES / 100. * S_MAX_HEAP_MEMORY_USAGE_PCT);
```
n3world: If you swap the multiplication and division you wouldn't need to make the 100 a float or do the cast.
`S_MAX_ALLOWED_PLAN_MEMORY = (S_TOTAL_JVM_BYTES * S_MAX_HEAP_MEMORY_USAGE_PCT) / 100;`

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -63,6 +57,16 @@
...
     private ArrayDeque<JoinNode> m_joinOrders = new ArrayDeque<>();
 
+    private static final long S_TOTAL_JVM_BYTES = Runtime.getRuntime().maxMemory();
+    // Approximate size in bytes of each plan: about 50 KB with some experiments.
+    private static final long S_PLAN_SIZE = 50_000;
+    // Number of times generateSubPlanForJoinNode() gets called recursively that we collect an estimate of heap size,
+    // and early exit if too large heap size had been used.
+    private static final int S_PLAN_ESTIMATE_PERIOD = 100;
+    // Stop generating any further possible plans, if we have reached xx% of available JVM heap memory
+    private static final short S_MAX_HEAP_MEMORY_USAGE_PCT = 98;
```
n3world: You want to use up to 98% of the heap for the planning of one join operation ?

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
```diff
@@ -538,18 +542,45 @@ private void generateSubPlanForJoinNodeRecursively(JoinNode rootNode,
             return;
         }
 
+        if (! m_plans.isEmpty() && m_plans.size() % S_PLAN_ESTIMATE_PERIOD == 0) {
+            final long usedHeapMemory = evaluatePlanSize() * S_PLAN_SIZE;
```
n3world: I am not sure this calculation is right. evaluatePlanSize returns the number of nodes in the plan tree. Is S_PLAN_SIZE supposed to be the max size of a plan or node?

kaikaizi: S_PLAN_SIZE is an estimate of the size of each plan node (~50KB) including all the overhead in the recursive structure.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
kaikaizi: Yes, at least this solves the planning GC/crash problem we have: it avoid having to GC until the planner crashes. I may GC once or twice, but would give up after that.
This results in short plan time at the cost of potentially longer execution time, because we may not have found an optimal plan that considers all indexes, etc.
And 98% is a rough estimate. We do not want to give up too soon.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
kaikaizi: Good idea, will do.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
kaikaizi: Good one.

src/frontend/org/voltdb/types/JoinType.java
```diff
@@ -30,17 +30,34 @@
...
             "Enum element " + this.name() +
             " in position " + this.ordinal() +
             " instead of position " + val;
+        m_val = val;
     }
 
     public int getValue() {
         return this.ordinal();
     }
+    public String toString() {
```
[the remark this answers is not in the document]
kaikaizi: Thanks, learnt.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
kaikaizi: Update: to keep track of plan nodes inside plan, I need to make some extensive changes in the class hierarchy of PlanNode (`AbstractPlanNode`, `SendPlanNode`, `WindowFunctionPlanNode`, `OrderByPlanNode`, `AbstractJointPlanNode`, etc. etc.).

In spirit of Calcite planner that is going to replace (most of) our VoltDB planning, I'll limit the change to periodic counting strategy.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
Chaozz: Why not using Java API to get the actual memory usage instead of estimating.

src/frontend/org/voltdb/planner/SelectSubPlanAssembler.java
n3world: But you don't prevent the OOM if you have multiple operations all using some heap it is easy for other parts of the system to be using up more than 2% of heap cumulative

PR #5810, opened 2018-12-18
ENG-15075: Throw Exception when a query is determined to be multi-partition

Chaozz:
Check if a query is Multi-partition after the logical planning, throw an Exception to force a fallback in Calcite planner.
I also remove all the Exchange Nodes since they are not used in SP query plan.

Note: I throw a `UnsupportedOperationException` rather than `PlannerFallbackExecption`, cause I can only throw runtime exceptions in the overridden `onMatch` functions.

### how to decide a query without join is single-partition?
1. push the filters down, make the related filter sit above its target tablescan
2. make a rule match the
```
VoltDBLCalc
      VoltDBLTableScan
```
check if the filtered result is locate in a single partition. If not, throw.
### how to decide a query with join is single-partition?
1. do the previous check
2. join on two partitioned table is MP, throw
3. join on two replicated table is SP
4. join on one replicated table and one nicely filtered(what the previous section does) partitioned table is SP.
5. join on two (replicated table or `case 3 join`) is SP, and it is equivalent to a `case 3 join`.
6. join on two `case 4 join` is MP, throw.
7. join on one  (replicated table or `case 3 join`) and one `case 4 join` is SP, and it is equivalent to a `case 4 join`.
8. apply the above rules recursively.


src/frontend/org/voltdb/calciteadapter/rel/logical/VoltDBLCalc.java
```diff
@@ -30,24 +30,32 @@
  * @author Michael Alexeev
  * @since 8.4
  */
-public class VoltDBLCalc extends Calc implements VoltDBLRel{
+public class VoltDBLCalc extends Calc implements VoltDBLRel {
+    private boolean m_isReplicated = true;
```
alexeevm: I am not sure that a logical relation should be aware of a distribution property. I think it belongs to the physical realm and the SP / MP determination should be part of one(the first one?) of the physical stages.
Have you considered to use RelDistribution trait instead of the m_isReplicated attribute?
For example, all Logical relations are assigned the RelDistribution.ANY trait to begin with.
During the VoltDBLTableScan to VoltDBPTableSeqScan conversion, the trait is set to either SINGELTON or HASH
All rules that currently relay on the  m_isReplicated attribute can access the trait instead
m_isReplicated = distributionTrait != RelDistribution.SINGLE && distributionTrait != RelDistribution.ANY
The only problem I am not sure what is the best way to propagate scan's distribution trait up. Maybe a series of rules matching any node (SingleRel) with a trait RelDistribution.ANY and any() with RelDistribution.SINGLE or RelDistribution.HASH? Plus a rule for joins to match a join with distribution ANY and its children both having their distribution trait set to something other than ANY

src/frontend/org/voltdb/newplanner/rules/logical/MPJoinQueryFallBackRule.java
```diff
@@ -0,0 +1,105 @@
...
+                    operand(VoltDBLJoin.class,
+                            operand(VoltDBLJoin.class, any()),
+                            operand(VoltDBLJoin.class, any())), "MPJoinQueryFallBackRule3");
+
+    public static final MPJoinQueryFallBackRule INSTANCE_4 =
+            new MPJoinQueryFallBackRule(
+                    operand(VoltDBLJoin.class,
+                            unordered(operand(VoltDBLJoin.class, any()),
+                                    operand(VoltDBLTableScan.class, any()))), "MPJoinQueryFallBackRule4");
+
```
alexeevm: Consider the
select RI1.bi from RI1, (select I from RI2 order by I) RI22 where RI1.i = RI22.I

The logical tree is
```
LogicalProject
  LogicalFilter
    LogicalJoin
      VoltDBLTableScan
      LogicalSort
        LogicalProject
          VoltDBLTableScan
```

which won't be caught by the MPJoinQueryFallBackRule. The example is contrive but still...
If by using RelDestribution trait instead we can eliminate the need to enumerate all possible variants it would be a big gain

src/frontend/org/voltdb/calciteadapter/rel/logical/VoltDBLCalc.java
Chaozz: Yeah, I agree. We can take advantage of the RelDistribution Trait and it will eliminate some edge cases. I have replaced m_isReplicated with a new RelDistribution approach.

src/frontend/org/voltdb/newplanner/rules/logical/MPJoinQueryFallBackRule.java
Chaozz: done using RelDestribution trait. I also add this one to my test. good catch!

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public void onMatch(RelOptRuleCall call) {
+        if (call.rel(0) instanceof VoltLogicalCalc && call.rel(1) instanceof VoltLogicalTableScan) {
+            VoltLogicalCalc calc = call.rel(0);
+            VoltLogicalTableScan tableScan = call.rel(1);
+            RelDistribution tableDist = tableScan.getTable().getDistribution();
+            if (tableDist.getType() != RelDistribution.Type.SINGLETON) {
+                if (calc.getProgram().getCondition() == null ||
```
kaikaizi: Add some comment on what this condition matches: `SELECT ... FROM t`

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -0,0 +1,114 @@
...
+                        !isSinglePartitioned(calc.getProgram(), calc.getProgram().getCondition(), tableDist.getKeys())) {
+                    throw new UnsupportedOperationException("MP query not supported in Calcite planner.");
+                }
+            }
+            call.transformTo(calc.copy(calc.getTraitSet().replace(tableDist), calc.getInputs()));
+        } else {
+            RelNode child = call.rel(1);
```
kaikaizi: What are some examples that this branch correspond to?

src/frontend/org/voltdb/plannerv2/rules/physical/VoltPSeqScanRule.java
```diff
@@ -55,35 +51,16 @@ public void onMatch(RelOptRuleCall call) {
         RelDistribution tableDist = tableScan.getTable().getDistribution();
         int scanSplitCount = (Type.SINGLETON == tableDist.getType()) ?
                 1 : Constants.DISTRIBUTED_SPLIT_COUNT;
+        // TODO: we make the distribution trait here ALWAYS SINGLE because we only support SP in our initial release.
```
kaikaizi: Would exceptions you through for MP be thrown in L48? In that way, can you add an assertion/Precondition check (with comment) in L52 like:
`assert tableDist.getType() == RelDistribution.Type.SINGLETON`?

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,294 @@
...
+import org.apache.calcite.rel.RelDistributions;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelRoot;
+import org.voltdb.calciteadapter.CatalogAdapter;
+import org.voltdb.calciteadapter.rel.logical.VoltDBLRel;
+import org.voltdb.newplanner.rules.PlannerPhase;
+import org.voltdb.newplanner.util.VoltDBRelUtil;
+import org.voltdb.types.CalcitePlannerType;
+
+public class TestMPQueryFallbackRules extends VoltConverterTestCase {
```
kaikaizi: Add some document: what it tests, why it's temporal, when we expect this test to go away, etc.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,294 @@
...
+    private void assertNotFallback(String sql) {
+        RelRoot root = parseValidateAndConvert(sql);
+
+        // apply logical rules
+        RelTraitSet logicalTraits = root.rel.getTraitSet().replace(VoltDBLRel.VOLTDB_LOGICAL);
+        RelNode nodeAfterLogicalRules = CalcitePlanner.transform(CalcitePlannerType.VOLCANO, PlannerPhase.LOGICAL,
+                root.rel, logicalTraits);
+
+        // Add RelDistributions.ANY trait to the rel tree.
+        nodeAfterLogicalRules = VoltDBRelUtil.addTraitRecurcively(nodeAfterLogicalRules, RelDistributions.ANY);
```
kaikaizi: s/addTraitRecurcively/addTraitRecursively/

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,294 @@
...
+
+        // do the MP fallback check
+        CalcitePlanner.transform(CalcitePlannerType.HEP_BOTTOM_UP, PlannerPhase.MP_FALLBACK,
+                nodeAfterLogicalRules);
+    }
+
+    private void assertFallback(String sql) {
+        try {
+            assertNotFallback(sql);
+        } catch (RuntimeException e) {
```
kaikaizi:
```suggestion
        } catch (UnsupportedOperationException e) {
            assertEquals("MP query not supported in Calcite planner.", e.getMessage());
        } catch (RuntimeException e) {
            assertTrue(e.getMessage().startsWith("...");
         }
```

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,294 @@
...
+            assertTrue(e.getMessage().startsWith("Error while applying rule") ||
+                    e.getMessage().equals("MP query not supported in Calcite planner."));
+            // we got the exception, we are good.
+            return;
+        }
+        fail("Expected fallback.");
```
kaikaizi: minor: you may move this inside try block.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,294 @@
...
+        assertNotFallback("select * from P1 where i = 1");
+
+        assertNotFallback("select * from P1 where 1 = i");
+
+        // other conditions on partition key
+        assertFallback("select * from P1 where i > 10");
+        assertFallback("select * from P1 where i <> 10");
+
+        // equal condition on partition key with ANDs
+        assertNotFallback("select si, v from P1 where 7=si and i=2");
```
kaikaizi: Does `SELECT si + 1 FROM P1 WHERE 7 = si` or `SELECT max(v) FROM P1 WHERE 7 = si` work too?

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,294 @@
...
+        assertNotFallback("select si, v from P1 where (7>si or ti=2) and i=2");
+        assertFallback("select si, v from P1 where (7>si or ti=2) or i=2");
+
+        // equal condition with some expression that always TURE
+        assertNotFallback("select si, v from P1 where (7=si and i=2) and 1=1");
+        assertFallback("select si, v from P1 where (7=si and i=2) or 1=1");
+
+        // equal condition with some expression that always FALSE
+        assertNotFallback("select si, v from P1 where (7=si and i=2) and 1=2");
+        // TODO: we should pass the commented test below if the planner is clever enough
```
kaikaizi: Sure! I'm wondering how Calcite can get to be with RexNode evaluation.

tests/frontend/org/voltdb/plannerv2/TestPhysicalConversion.java
```diff
@@ -45,14 +45,19 @@ void assertPlanMatch(String sql, String expectedPlan) {
         RelNode nodeAfterLogicalRules = CalcitePlanner.transform(CalcitePlannerType.VOLCANO, PlannerPhase.LOGICAL,
                 root.rel, logicalTraits);
 
+        // Add RelDistributions.ANY trait to the rel tree.
```
kaikaizi: Why is this test commented?

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -0,0 +1,114 @@
...
+                case EQUALS:
+                    return isSinglePartitioned(program, rexCall.getOperands().get(0), partitionKeys) ||
+                            isSinglePartitioned(program, rexCall.getOperands().get(1), partitionKeys);
+                case AND:
+                    return isSinglePartitioned(program, rexCall.getOperands().get(0), partitionKeys) ||
+                            isSinglePartitioned(program, rexCall.getOperands().get(1), partitionKeys);
+                case OR:
+                    return isSinglePartitioned(program, rexCall.getOperands().get(0), partitionKeys) &&
+                            isSinglePartitioned(program, rexCall.getOperands().get(1), partitionKeys);
+                default:
```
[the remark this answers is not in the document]
Chaozz: Oh, My initial thinking is that `NOT` is always guarantee to be not single Partitioned condition.
Now I came up with some corner cases:
`Not PK<> 15`,
`Not ( Not PK = 15)` 
`Not ( PK1 <> 15 or Col = 16 or ...)` 
`Not ( Not (PK = 15 and Col = 16 and ...))`
a liitle wired usage though

tests/frontend/org/voltdb/plannerv2/TestPhysicalConversion.java
Chaozz: It is not compiled since the result the refactor. I will rewrite it

src/frontend/org/voltdb/plannerv2/rules/physical/VoltPSeqScanRule.java
alexeevm: I think that the comment here is a bit misleading. If a table is partitioned, then its distribution here will be HASH and not a SINGLETON.  The MPQueryFallBackRule guarantees that the final result is effectively single partitioned because of the partitioning filter (T.PATR_KEY = ABC)

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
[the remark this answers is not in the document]
kaikaizi: No need to go into the recursive direction. Does calcite simplify those conditions?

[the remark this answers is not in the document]
Chaozz: It seems calcite does not simplify NOT(NOT ...).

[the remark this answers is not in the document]
Chaozz: OK, there is a `ReduceExpressionsRule` can simplify this.

src/frontend/org/voltdb/compiler/PlannerTool.java
```diff
@@ -228,14 +228,19 @@ public synchronized AdHocPlannedStatement planSqlCalcite(SqlTask task, NonDdlBat
         RelNode nodeAfterLogical = CalcitePlanner.transform(CalcitePlannerType.VOLCANO, PlannerPhase.LOGICAL,
                 root.rel, logicalTraits);
 
+        // Add RelDistributions.ANY trait to the rel tree.
+        nodeAfterLogical = VoltDBRelUtil.addTraitRecurcively(nodeAfterLogical, RelDistributions.ANY);
```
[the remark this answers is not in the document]
Chaozz: done

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param rexNode       the condition of the program.
+     * @param partitionKeys the list of partition keys for the target table.
+     * @return true if the filtered result is located at a single partition.
+     */
+    private boolean isSinglePartitioned(RexProgram program, RexNode rexNode, List<Integer> partitionKeys) {
+        if (rexNode instanceof RexCall) {
+            RexCall rexCall = (RexCall) rexNode;
+            SqlKind sqlKind = rexCall.getOperator().getKind();
+            switch (sqlKind) {
+                case EQUALS:
```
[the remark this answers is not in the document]
Chaozz: done

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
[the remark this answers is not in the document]
Chaozz: Ok, I add the addtional check for NOT and add the corresponding tests in  `testPartitionedWithNotFilter`

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -0,0 +1,114 @@
...
+        return false;
+    }
+
+    @Override
+    public void onMatch(RelOptRuleCall call) {
+        if (call.rel(0) instanceof VoltLogicalCalc && call.rel(1) instanceof VoltLogicalTableScan) {
+            VoltLogicalCalc calc = call.rel(0);
+            VoltLogicalTableScan tableScan = call.rel(1);
+            RelDistribution tableDist = tableScan.getTable().getDistribution();
+            if (tableDist.getType() != RelDistribution.Type.SINGLETON) {
```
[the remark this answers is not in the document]
Chaozz: done

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
Chaozz: added

src/frontend/org/voltdb/plannerv2/rules/physical/VoltPSeqScanRule.java
Chaozz: you are right, it can be HASH or SINGLETON. 
Updated.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
Chaozz: added.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
Chaozz: Actually, after applying the `ReduceExpressionsRule`, calcite can do it now.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
Chaozz: Yes, it works.
I added them to the test.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
Chaozz: I want the test fail if there is not exception thrown, so I move it out from the try block

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
Chaozz: if it is not a VoltLogicalCalc / VoltLogicalTableScan pattern,
propagate the DistributionTrait bottom up.
I have updated the comment

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Helper function to decide whether the filtered result is located at a single partition.
+     *
+     * @param program       the program of a {@link LogicalCalc}.
+     * @param rexNode       the condition of the program.
+     * @param partitionKeys the list of partition keys for the target table.
+     * @return true if the filtered result is located at a single partition.
+     */
+    private boolean isSinglePartitioned(RexProgram program, RexNode rexNode, List<Integer> partitionKeys) {
+        if (rexNode instanceof RexCall) {
```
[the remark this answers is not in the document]
Chaozz: Cause those if returns, I prefer using separate if's.

src/frontend/org/voltdb/plannerv2/rel/physical/VoltPhysicalRel.java
```diff
@@ -26,8 +26,8 @@
 
 import java.util.Objects;
 
-public interface VoltDBPRel extends RelNode {
-    final static Convention VOLTDB_PHYSICAL = new Convention.Impl("VOLTDB_PHYSICAL", VoltDBPRel.class) {
+public interface VoltPhysicalRel extends RelNode {
+    Convention CONVENTION = new Convention.Impl("CONVENTION", VoltPhysicalRel.class) {
         public boolean canConvertConvention(Convention toConvention) {
```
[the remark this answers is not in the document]
Chaozz: done

tests/frontend/org/voltdb/plannerv2/Plannerv2TestCase.java
```diff
@@ -166,23 +170,59 @@ void checkEx(Exception ex) throws AssertionError {
...
-                m_transformedNode = m_planner.transform(m_ruleSetIndex, m_requiredOutputTraits, m_root.rel);
-                if (m_expectedTransform != null) {
-                    String actualTransform = RelOptUtil.toString(m_transformedNode);
-                    assertEquals(m_expectedTransform, actualTransform);
-                }
-            } catch (Exception ex) {
-                checkEx(ex);
+                test();
+            }
+            catch (PlannerFallbackException e){
```
kaikaizi: Are there any cases that a `PlannerFallbackException` is unexpected? I'm thinking that the exception class does not need a string for cause, and when we catch it in test here, we ignore it.

tests/frontend/org/voltdb/plannerv2/TestLogicalRules.java
```diff
@@ -401,50 +397,53 @@ public void testDistinct() {
 
     public void testJoin() {
         m_tester.sql("select R1.i, R2.v from R1, R2 where R2.si = R1.i and R2.v = 'foo'")
-                .transform("VoltLogicalCalc(expr#0..11=[{inputs}], expr#12=[CAST($t7):INTEGER], expr#13=[=($t12, $t0)], expr#14=['foo'], expr#15=[=($t11, $t14)], expr#16=[AND($t13, $t15)], I=[$t0], V=[$t11], $condition=[$t16])\n" +
-                           "  VoltLogicalJoin(condition=[true], joinType=[inner])\n" +
-                           "    VoltLogicalTableScan(table=[[public, R1]])\n" +
-                           "    VoltLogicalTableScan(table=[[public, R2]])\n")
+                .transform("VoltLogicalCalc(expr#0..11=[{inputs}], I=[$t0], V=[$t11])\n" +
+                        "  VoltLogicalJoin(condition=[=(CAST($7):INTEGER, $0)], joinType=[inner])\n" +
```
kaikaizi: Add a TODO here about Calcite casting to INTEGER instead of BigInt? We may fail some JUnit tests when we execute such Calcite plan.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,298 @@
...
+        // equal condition on partition key with ANDs
+        m_tester.sql("select si, v from P1 where 7=si and i=2").test();
+        m_tester.sql("select si, v from P1 where 7>si and i=2 and ti<3").test();
+
+        m_tester.sql("SELECT si + 1 FROM P1 WHERE 7 = i").test();
+        m_tester.sql("SELECT max(v) FROM P1 WHERE 7 = i").test();
+
+        // equal condition on partition key with ORs
+        m_tester.sql("select si, v from P1 where 7=si or i=2").testFail();
+        m_tester.sql("select si, v from P1 where 7=si or i=2 or ti=3").testFail();
```
kaikaizi: well, L84 implies L85.

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,298 @@
...
+        m_tester.sql("select si, v from P1 where 7>si and (i=2 and ti<3)").test();
+        m_tester.sql("select si, v from P1 where (7>si or ti=2) and i=2").test();
+        m_tester.sql("select si, v from P1 where (7>si or ti=2) or i=2").testFail();
+
+        // equal condition with some expression that always TURE
+        m_tester.sql("select si, v from P1 where (7=si and i=2) and 1=1").test();
+        m_tester.sql("select si, v from P1 where (7=si and i=2) or 1=1").testFail();
+
+        // equal condition with some expression that always FALSE
+        m_tester.sql("select si, v from P1 where (7=si and i=2) and 1=2").test();
```
kaikaizi: or `and true`

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,298 @@
...
+                + "where t1.i = 3").test();
+
+        m_tester.sql("select RI1.bi from RI1, (select I from P2 order by I) P22 where RI1.i = P22.I").testFail();
+
+        m_tester.sql("select RI1.bi from RI1, (select I from P2 where I = 5 order by I) P22 where RI1.i = P22.I").test();
+    }
+
+    public void testIn() {
+        // calcite will use equal to rewrite IN
+        m_tester.sql("select * from P1 where i in (16)").test();
```
kaikaizi: Add `m_tester.sql("select * from P1 where i in (16, 16)").test();`?

tests/frontend/org/voltdb/plannerv2/TestMPQueryFallbackRules.java
```diff
@@ -0,0 +1,298 @@
...
+
+        m_tester.sql("select * from P1 where i in (1,2,3,4,5,6,7,8,9,10)").testFail();
+
+        m_tester.sql("select * from P1 where i Not in (1, 2)").testFail();
+
+        m_tester.sql("select si from P1 where i in (1,2) and i in (1,3)").testFail();
+
+        m_tester.sql("select si from P1 where i in (1,2) or i not in (1,3)").testFail();
+        // calcite will use Join to rewrite IN (sub query)
+        m_tester.sql("select si from P1 where si in (select i from R1)").testFail();
```
kaikaizi: Add `m_tester.sql("select i from R1 where i in (select si from P1)").testFail();`?

Chaozz:
@yzhang1991 junit testing fails on your commit https://github.com/VoltDB/voltdb/pull/5810/commits/4958ec3ff62cc55196474ef5099655a83257eb46,
can you check it?

PR #5813, opened 2018-12-19
Eng 14912 - Handle Mis-parition for Elastic Shrink

GuaGuaDean:
handled no partition leader case in createTransaction() and will return mis-partitioned client response accordingly

### DIFF
--- a/src/frontend/org/voltdb/InvocationDispatcher.java
+++ b/src/frontend/org/voltdb/InvocationDispatcher.java
@@ -518,12 +518,22 @@ public final class InvocationDispatcher {
             // when VoltDB.crash... is called, we close off the client interface
             // and it might not be possible to create new transactions.
             // Return an error.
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/InvocationDispatcher.java
```diff
@@ -1332,6 +1342,13 @@ public  boolean createTransaction(
             }
         }
 
+        if (initiatorHSId == null) {
+            hostLog.rateLimitedLog(60, Level.WARN, null,
```
n3world: I am not sure warn is necessary here. You are probably fine with an INFO since this can be something that happens during normal operation now.

GuaGuaDean: Updated

manjujames:
createTransaction is also used by DRInvocationAdapter (DR consumer code). We will need to handle this case there as well. If there is a way to return mispartitioned response, that will be better.

PR #5816, opened 2018-12-20
ENG-15083:

wweiss-voltdb:
There is a race between UAC notifcations to export and EOS notifications from Sites. This means that if either is used to trigger the removal of the ExportDataSource, the other fails an assertion. This is exposed with the TestExportStatsSuite which is currently failing.

However, the m_eos flag is no longer used so the notification of the EndOfStream is no longer necessary. There is a small chance that the Export streams may start injecting events into the PBD. If this ends up being the case, we will still need the CPP api so the Topend callback will be kept until the need for this is decided.

### DIFF
--- a/src/frontend/org/voltdb/export/ExportGeneration.java
+++ b/src/frontend/org/voltdb/export/ExportGeneration.java
@@ -684,28 +684,6 @@ public class ExportGeneration implements Generation {
         source.pushExportBuffer(startSequenceNumber, tupleCount, uniqueId, buffer, sync);
     }
 
-    @Override
-    public void pushEndOfStream(int partitionId, String signature) {
-        assert(m_dataSourcesByPartition.containsKey(partitionId));
-        assert(m_dataSourcesByPartition.get(partitionId).containsKey(signature));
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -684,28 +684,6 @@ public void pushExportBuffer(int partitionId, String signature,
         source.pushExportBuffer(startSequenceNumber, tupleCount, uniqueId, buffer, sync);
     }
 
-    @Override
```
rdykiel: Deleting useless code is good!

src/frontend/org/voltdb/export/ExportManager.java
```diff
@@ -673,16 +673,6 @@ public void shutdown() {
     public static void pushEndOfStream(
             int partitionId,
             String signature) {
-        ExportManager instance = instance();
```
luyangco: Why leave pushEndOfStream() empty instead of completely removing the call from EE?

luyangco:
I approve the fix but have question about whether the scope of the fix should touch EE.

PR #5824, opened 2019-01-02
ENG-15127, if an export stream is drained and discarded between two @…

luyangco:
…Statistics EXPORT queries issued by either "voltadmin pause --wait" or "voltadmin shutdown --save" command, a bug buried in the python script causes it always show export is not drained. This commit fixes that issue.

Change-Id: I08ec64c09048d8cdfd02a755e60edffd773d2c4a


rdykiel:
Looks good to me

PR #5827, opened 2019-01-02
ENG-15112 clean up SubPlanAssembler

kaikaizi:
and things that have to do with it. This task is to clean up logic of very long methods with long nested-if's, strip out nested classes, move PlanningErrorException to where VoltDB defines exception types, and other relevant changes.


src/frontend/org/voltdb/exceptions/PlanningErrorException.java
```diff
@@ -15,14 +15,20 @@
...
+package org.voltdb.exceptions;
 
 public class PlanningErrorException extends RuntimeException {
     private static final long serialVersionUID = 1L;
 
     public PlanningErrorException(String msg) {
         super(msg);
     }
+    public PlanningErrorException(Throwable e) {
+        super(e.getCause());
```
n3world: Why are you setting the cause of this exception as the cause of e? Shouldn't it just be e same as the next constructor?

src/frontend/org/voltdb/planner/InsertSubPlanAssembler.java
```diff
@@ -24,8 +24,8 @@
 import java.util.Set;
 
 import org.voltdb.catalog.Column;
-import org.voltdb.catalog.Database;
 import org.voltdb.catalog.Table;
+import org.voltdb.exceptions.PlanningErrorException;
```
n3world: Why was this import added to this class and others which don't use this exception?

src/frontend/org/voltdb/planner/ExpressionOrColumn.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    ExpressionOrColumn(int aIndexEntryNumber,
+                       StmtTableScan tableScan,
+                       AbstractExpression expr,
+                       SortDirectionType sortDir,
+                       ColumnRef colRef) {
+        // Exactly one of expr or colRef can be null.
+        assert (expr == null) == (colRef != null);
+        assert colRef == null || tableScan != null;
```
n3world: Wouldn't IllegalArgumentException be better than assert since you are doing argument checking and asserts are off during normal run time?

src/frontend/org/voltdb/planner/ExpressionOrColumn.java
```diff
@@ -0,0 +1,307 @@
...
+        m_tableScan = tableScan;
+        m_sortDirection = sortDir;
+    }
+
+    SortDirectionType sortDirection() {
+        return m_sortDirection;
+    }
+
+    @Override
+    public boolean equals(Object otherObj) {
```
n3world: It is usually highly encouraged to implement hashCode if you implement equals

src/frontend/org/voltdb/planner/WindowFunctionScore.java
```diff
@@ -0,0 +1,272 @@
...
+                    // If there are expressions later on in the
+                    // m_partitionByExprs or m_unmatchedOrderByExprs
+                    // which match this expression we need to
+                    // delete them.  We can safely ignore them.
+                    //
+                    // If there are more than one instances of
+                    // an expression, say with "PARTITION BY A, A",
+                    // we need to remove them all.  But guard against
+                    // an infinite loop here;
+                    while (m_partitionByExprs.remove(eorc)) {
```
n3world: Not sure if this is better but it is a bit more clear if this were `m_partitionByExprs.removeIf(eorc::equals)` and it guarantees at most one iteration.

Another option would be `m_partitionByExprs.removeAll(Collections.singleton(eorc))`

src/frontend/org/voltdb/planner/WindowFunctionScore.java
```diff
@@ -0,0 +1,272 @@
...
+            if (moreBindings != null) {
+                // We matched, because moreBindings != null, and
+                // the sort direction matched as well.  So
+                // add nextEOC to the order matching expressions
+                // list and add the bindings to the bindings list.
+                m_orderedMatchingExpressions.add(nextStatementEOC.m_expr);
+                m_bindings.addAll(moreBindings);
+                // Remove the next statement EOC from the unmatched OrderByExpressions
+                // list since we matched it.  We need to remove all of them,
+                while (m_unmatchedOrderByExprs.remove(nextStatementEOC)) {
```
n3world: Also could use removeIf or removeAll with a singleton set of `nextStatementEOC`

src/frontend/org/voltdb/planner/WindowFunctionScoreboard.java
```diff
@@ -0,0 +1,187 @@
...
+     * the AccessPath, the order spoilers list and the bindings.
+     *
+     * We prefer window functions to statement level
+     * order by expressions.  If there is more than one
+     * window functions, we prefer the one with the least
+     * number of order spoilers.  In the case of ties we
+     * return the first one in the list, which is essentially
+     * random.  Perhaps we can do better.
+     *
+     * @return The number of order spoilers in the candidate we choose.
```
n3world: It would be nice to add doc on the parameters too

src/frontend/org/voltdb/planner/AccessPath.java
```diff
@@ -120,7 +123,7 @@ private String indexPurposeString() {
         switch (m_windowFunctionUsesIndex) {
         case SubPlanAssembler.STATEMENT_LEVEL_ORDER_BY_INDEX:
             return "Statement Level Order By";
-        case SubPlanAssembler.NO_INDEX_USE:
+        case WindowFunctionScoreboard.NO_INDEX_USE:
```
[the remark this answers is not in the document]
kaikaizi: It is indeed weird, and I failed to see that. Thanks, will combine the two enum defs.

src/frontend/org/voltdb/planner/AccessPath.java
```diff
@@ -133,4 +136,32 @@ private String indexPurposeString() {
...
+    }
+
+    public void setSortDirection(SortDirectionType newSortDirection) {
+        sortDirection = newSortDirection;
+    }
+    public IndexLookupType getIndexLookupType() {
+        return lookupType;
+    }
+
+    public List<AbstractExpression> getIndexExpressions() {
```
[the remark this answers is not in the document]
kaikaizi: These are left-overs when picking Mike's changes. I'll clean them for the current task.

src/frontend/org/voltdb/planner/WindowFunctionScoreboard.java
kaikaizi: Will do.

src/frontend/org/voltdb/exceptions/AdHocPlanningException.java
```diff
@@ -0,0 +1,29 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.voltdb.exceptions;
+
+public class AdHocPlanningException extends PlanningErrorException {
```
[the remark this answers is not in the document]
kaikaizi: It came as a subclass of AdHocNTBase, that `extends Exception`. I guess at that time we may not even have a `PlanningErrorException`. 
I agree that we don't need a specialized PlanningErrorException. Will combine the two.

PR #5831, opened 2019-01-03
ENG-14909 elastic remove operator

n3world:
Minor utility additions and plumbing for elastic remove feature that is needed in the community edition. As well as rebranding join to elastic for everything except logging.

Required by https://github.com/VoltDB/pro/pull/2437

### DIFF
--- a/src/frontend/org/voltcore/zk/BabySitter.java
+++ b/src/frontend/org/voltcore/zk/BabySitter.java
@@ -20,19 +20,25 @@ package org.voltcore.zk;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.concurrent.*;
+import java.util.concurrent.Callable;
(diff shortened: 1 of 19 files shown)


src/frontend/org/voltcore/zk/BabySitter.java
```diff
@@ -177,10 +183,13 @@ public static BabySitter nonblockingFactory(ZooKeeper zk, String dir,
         public void process(WatchedEvent event)
         {
             try {
-                m_es.submit(m_eventHandler);
+                if (event.getType() != EventType.NodeDeleted) {
```
manjujames: Why this check?

n3world: Because if you do a getChildren on a deleted node ZooKeeper throws NoNodeException and I saw that exception take down the node.

PR #5836, opened 2019-01-08
Eng 15153 st jdbc vertica startup failure

jwiebema1:
delay prepared statement creation until we need them incase the DB isn't initialized.


tests/test_apps/genqa/src/genqa/JDBCGetData.java
```diff
@@ -38,15 +38,15 @@
...
     /**
      * Connect and prepare the select to make the per row handling
      * as efficient as possible, though it's still pretty pokey
      * @param config
      * @return
      */
-    public static Connection jdbcConnect(Config config) {
-
+    public static Connection jdbcConnect(Config lconfig) {
+        config = lconfig;
```
prosegay: this doesn't look right

jwiebema1: I made it a private static value so I could use 'config'  when we read the rows

PR #5837, opened 2019-01-08
Eng 14998 gram views

johnTcrawford:
Modified sql-grammar.txt to include CREATE (and DROP) VIEW statements,
and put them in the "DDL" section, along with the existing CREATE (and
DROP) INDEX and PROCEDURE AS statements; and greatly improved the use of
GROUP BY clauses (which are needed by CREATE VIEW). With a few other
minor tweaks, such as: more use of reusable table and column names; and
reducing the frequency of joins, views, sub-queries, nonexistent column
names. (And deleted a few things from sql-ddl-grammar.txt, which were
moved into sql-grammar.txt, with modifications). Modified
sql_grammar_generator.py to include multiple new error messages that we
see in response to CREATE VIEW statements; and to allow some new syntax
involving reusable symbol names.

### DIFF
--- a/tests/sqlgrammar/sql-grammar.txt
+++ b/tests/sqlgrammar/sql-grammar.txt
@@ -80,6 +80,29 @@
 #      table R1 has been specified elsewhere in the SQL statement, so using a
 #      definition such as {table-name;t1,t2,t3,t4,t5}.{any-column-name} makes
 #      this much more likely.
+#  11. An addition to the colon syntax for reusable names, described in #9
+#      above: if you wish to set multiple (possibly unique) values to multiple
+#      reusable names, you may use, for example, this syntax:
+#      {column-name:c1,c2,c3,c4,c5:10}. The first time that this is used (in a
+#      given SQL statement), it will find one possible value of {column-name},
+#      and set 'c1' equal to that; the second time, it will find another
+#      possible value, and set 'c2' to that; the third time, to 'c3', etc. The
+#      final number after the second colon, here '10', means that there is a 10%
+#      chance that repeat values of {column-name} will be allowed; if that value
+#      is instead '0', then distinct values of {column-name} will be used, with
+#      no duplicates, though if it runs out of distinct possible values of
+#      {column-name}, or out of reusable names, then duplicates will be used.
+#      The second colon and the repeat-percentage following it are optional; if
+#      not specified, there will be no attempt to avoid repeats.
+#  12. An addition to the semi-colon syntax for reusable names, described in
+#      #10 above: if you wish to get a list of reusable values, such as those
+#      defined in #11 above, you may use, for example, this syntax:
+#      {column-name;c1,c2,c3,c4,c5;', '}. This will produce a list of all the
+#      values of all of the reusable names listed, separated by the separator
+#      string (here, ', '); any reusable names that have not yet been given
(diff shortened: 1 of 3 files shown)


tests/sqlgrammar/sql-grammar.txt
```diff
@@ -1738,13 +1826,19 @@ window-order-by-item    ::= {column-expression} 2| {int-expression} | {timestamp
 # Aggregate functions - used with various data types defined below
 # (and in the window functions and scalar sub-queries above)
 #
-# TODO: More use of these in GROUP BY queries
 num-only-aggregate-func ::= SUM | AVG
 num-result-aggr-func    ::= COUNT 3| APPROX_COUNT_DISTINCT
-non-num-aggregate-func  ::= MIN | MAX
+legit-non-num-aggr-func ::= MIN | MAX
```
kaikaizi: `COUNT(?)` and `COUNT(DISTINCT ?)` and `COUNT(*)`?
`COUNT` is also an aggregation:
`SELECT i, COUNT(j) FROM foo GROUP BY i;`

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -1477,6 +1521,50 @@ comparison-operator     ::= {simple-comparison-op} 19| {str-comparison-operator}
 simple-comparison-op    ::= = 3| <> | != | < | > | <= | >= | IS NOT DISTINCT FROM
 str-comparison-operator ::= {simple-comparison-op} | LIKE | STARTS WITH
 
+################################################################################
+# GROUP BY clauses, and the SELECT statements that use them; including SELECT
+# statements without GROUP BY that use aggregate functions in a similar way
+#
+groupby-select-statement::= {simple-groupby-select} 3| {random-groupby-select}
+
+simple-groupby-select   ::= SELECT [{top-all-or-distinct}] {groupby-select-list} \
```
kaikaizi: `... HAVING...`? `SELECT i, COUNT(*) FROM foo HAVING COUNT(*) > 0;`

tests/sqlgrammar/sql-grammar.txt
```diff
@@ -281,8 +318,8 @@ table-name-or-alias-set ::= {table-name-set} 3| {table-alias-set}
 table-or-view-name-set  ::= {table-name-set} 3| {view-name-set}
 view-name-or-alias-set  ::= {view-name-set}  3| {view-alias-set}
 
-table-ref-set           ::= {table-name-or-alias-set} 5| \
-                            {view-name-or-alias-set}  2| \
+table-ref-set           ::= {table-name-or-alias-set} 8| \
+                            {view-name-or-alias-set}   | \
```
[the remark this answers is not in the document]
johnTcrawford: I reduced the frequency of this because it provides, among other things, the name of the thing on which a View will be created, and it is not legal to CREATE VIEW on another View; this lower frequency increases the percentage of valid CREATE VIEW statements, and some other statements. However, I may have gone a bit too far here; perhaps I'll do something in between.

tests/sqlgrammar/sql-grammar.txt
johnTcrawford: Most of those are covered, in the lines above and below; however, I believe I'm missing COUNT(DISTINCT ?), both for CREATE VIEW / GROUP BY, and also elsewhere. I'll check, and add it if it's not there.

PR #5843, opened 2019-01-09
ENG-15073  prevent replicas from initializing repair log truncation unless

zeemanhe:
a transaction is interleaved by partition leader migration.


src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1208,7 +1208,7 @@ private void handleFragmentResponseMessage(FragmentResponseMessage message)
             int result = counter.offer(message);
             if (result == DuplicateCounter.DONE) {
                 if (txn != null && txn.isDone()) {
-                    setRepairLogTruncationHandle(txn.m_spHandle, message.isForOldLeader());
+                    setRepairLogTruncationHandle(txn.m_spHandle, txn.getOldFragmentExecutedSiteId() == m_mailbox.getHSId());
```
luyangco: `OldLeader` flag and `OldFragmentExecutedSiteId` are set at the same place and have same meaning, so why replacing `OldLeader` flag with the new variable?

zeemanhe: OldLeader will be carried around in the message to other sites.  OldFragmentExecutedSiteId is local to the site. With the new flag, the change in the transaction path is minimized (only on truncation).  It may be possible to merge these two flags but more changes are required in the transaction path. The transaction path seems pretty stable for system test perspective from last couple of releases.  The additional changes in the transaction path to fix the hash mismatch may not worth the risk.

rdykiel: Since the task is always local you don't need the hsid. Could it be just a boolean?

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1630,25 +1630,25 @@ private long getRepairLogTruncationHandleForReplicas()
         return m_repairLogTruncationHandle;
     }
 
-    private void setRepairLogTruncationHandle(long newHandle, boolean isForLeader)
+    private void setRepairLogTruncationHandle(long newHandle, boolean isExecutedOnOldLeader)
     {
         if (newHandle > m_repairLogTruncationHandle) {
             m_repairLogTruncationHandle = newHandle;
```
rdykiel: I think Walter wanted this assignment of m_repairLogTruncationHandle to only occur on the master, and NEVER on a replica.

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1630,25 +1630,25 @@ private long getRepairLogTruncationHandleForReplicas()
...
-            m_bufferedReadLog.releaseBufferedReads(m_mailbox, newHandle);
+            m_bufferedReadLog.releaseBufferedReads(m_mailbox, m_repairLogTruncationHandle);
             // We have to advance the local truncation point on the replica. It's important for
             // node promotion when there are no missing repair log transactions on the replica.
             // Because we still want to release the reads if no following writes will come to this replica.
             // Also advance the truncation point if this is not a leader but the response message is for leader.
-            if (m_isLeader || isExecutedOnOldLeader) {
-                scheduleRepairLogTruncateMsg(newHandle);
+            if (m_isLeader) {
+                scheduleRepairLogTruncateMsg();
```
rdykiel: Should the assignment of m_repairLogTruncationHandle be made here , instead of at line 1636?

rdykiel:
I like the latest version: much simpler code IMO, which is goodness.

PR #5845, opened 2019-01-10
ENG-15166: Always consume partition data and replicas

n3world:
The root cause of the issue was that there were two queues one for the partition
data and one for the children data. If an ignored exception was thrown
retrieving the partition data then the item in the children queue was not
removed which caused the two queues to become out of sync.

This change simplifies the async retrieval of partition information by making a
single utility method which returns a list AsynPartitions. This ties the async
request for the partition data to the child request so that the processing of
both will not get out of sync. Also, this condenses two different implementation
of this algorithm down to one common implementation.

### DIFF
--- a/src/frontend/org/voltdb/iv2/Cartographer.java
+++ b/src/frontend/org/voltdb/iv2/Cartographer.java
@@ -18,7 +18,6 @@
 package org.voltdb.iv2;
 
 import java.util.AbstractMap;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/iv2/Cartographer.java
```diff
@@ -756,51 +795,31 @@ private boolean isPartitionLeadersBalanced() {
...
-                        return "StopNode is disallowed in initialization phase";
-                    }
+                if (partition.isInitializing()) {
+                    return "StopNode is disallowed in initialization phase";
                 }
 
-                List<String> replicas = childrenCallbacks.poll().getChildren();
-                //This is here just so callback is polled.
-                if (pid == MpInitiator.MP_INIT_PID) {
+                if (partition.getPid() == MpInitiator.MP_INIT_PID) {
```
JadenW: Since Mp partition would always be skipped (has no replica), we can set skipMp to true at getPartitionsAsync call?

n3world:
This method still returns an error if the mp is initializing. So we still
have to check that first.

PR #5856, opened 2019-01-14
Eng 14854

manjujames:


src/frontend/org/voltdb/iv2/Site.java
```diff
@@ -1167,12 +1167,12 @@ private static void handleUndoLog(List<UndoAction> undoLog, boolean undo) {
             return;
         }
 
-        for (final ListIterator<UndoAction> iterator = undoLog.listIterator(undoLog.size()); iterator.hasPrevious();) {
-            final UndoAction action = iterator.previous();
+        ListIterator<UndoAction> iterator = (undo) ? undoLog.listIterator(undoLog.size()) : undoLog.listIterator();
+        for (; (undo ? iterator.hasPrevious() : iterator.hasNext());) {
```
n3world: This should probably be just a while since there is no initialization or end of loop block

PR #5861, opened 2019-01-15
Eng-15164 delete VDM

rmorgenstein:
VDM/voltdeploy last year. This removes the VDM source code for V9.0 as well as the various python packages that it required.


prosegay:
its a big pull and i'm not sure how good a review I can provide, lets just go with it and see if there's any adverse effects.

PR #5867, opened 2019-01-16
OP-362:

wweiss-voltdb:
Sometimes PBD files that would normally be deleted need to truncated. However, the truncation does not occur if the file is already closed so this branch opens the file to truncate the file before closing it again. Unit test was also updated to ignore truncated PBD files in counting the number of files with data.

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -213,9 +213,10 @@ public class PBDRegularSegment extends PBDSegment {
     public void closeAndTruncate() throws IOException {
         try
         {
-            if (m_ras != null) {
-                m_ras.setLength(0);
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -213,9 +213,10 @@ public void closeAndDelete() throws IOException {
     public void closeAndTruncate() throws IOException {
         try
         {
-            if (m_ras != null) {
-                m_ras.setLength(0);
+            if (m_ras == null) {
+                m_ras = new RandomAccessFile( m_file, "rw");
             }
+            m_ras.setLength(0);
```
n3world: Do you want to do a metadata flush here so that you guarantee the metadata is stable on disk `FileChannel.force(true)`?

PR #5868, opened 2019-01-16
Op 362 truncate never deletes m

wweiss-voltdb:


src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -209,6 +209,20 @@ public void closeAndDelete() throws IOException {
...
     }
 
+    @Override
+    public void closeAndTruncate() throws IOException {
+        try
+        {
+            if (m_ras == null) {
+                m_ras = new RandomAccessFile( m_file, "rw");
+            }
+            m_ras.setLength(0);
```
n3world: Might want to perform flush here either `m_ras.getFD().sync()` or `m_ras.getChannel().force(true)`

PR #5870, opened 2019-01-17
ENG-14778 restore partition assignments for cluster recovery and node rejoin

zeemanhe:
Partitions assignments among the cluster nodes are related to host Ids.  Upon cluster  recover, a node may get a set of partitions, totally different from the set of partitions before it is shutdown, depending how fast a node races to get its host id.  The behavior would introduce the inconsistency of preserved persist data with the partition layout upon recovery, such as export overflow PBD files. The match of  persistent data with partition layout after recovery would help speed up  the recovery process and eliminate export gaps.


src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2161,9 +2190,26 @@ private AbstractTopology getTopology(StartAction startAction, Map<Integer, HostI
...
+                        List<Integer> hosts = restoredPartitionsByHosts.get(v.m_recoveredPartitions);
+                        if (hosts == null) {
+                            hosts = Lists.newArrayList();
+                            restoredPartitionsByHosts.put(v.m_recoveredPartitions, hosts);
+                        }
+                        hosts.add(k);
+                    }
+                });
+                long version = topology.version;
+                topology = AbstractTopology.mutateRecoverTopology(topology, m_messenger.getLiveHostIds(), -1, "", restoredPartitionsByHosts, true);
```
n3world: AbstractTopology.getTopology() already takes the HostInfos as an argument and could probably do the recover as part of that method

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -279,14 +294,25 @@ public HostInfo(String hostIp, String group, int localSitesCount) {
...
 
         public static HostInfo fromBytes(byte[] bytes) throws JSONException
         {
             final JSONObject obj = new JSONObject(new String(bytes, StandardCharsets.UTF_8));
-            return new HostInfo(obj.getString(HOST_IP), obj.getString(GROUP), obj.getInt(LOCAL_SITES_COUNT));
+            return new HostInfo(obj.getString(HOST_IP), obj.getString(GROUP), obj.getInt(LOCAL_SITES_COUNT),
+                    obj.getString(RECOVERED_PARTITION_IDS));
+        }
+
+        public List<Integer> getRecoveredPartitions() {
```
n3world: Is this method ever used?

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -985,6 +1022,83 @@ public static AbstractTopology mutateRecoverTopology(AbstractTopology topology,
...
+        if (restoredPartitionsByHosts == null || restoredPartitionsByHosts.isEmpty()) {
+            return topology;
+        }
+
+        Map<Integer, Host> hostsById = new TreeMap<>(topology.hostsById);
+        List<Host> hosts = Lists.newArrayList();
+        for (Map.Entry<String, List<Integer>> entry : restoredPartitionsByHosts.entrySet()) {
+            List<Integer> restoredHosts = entry.getValue();
+            List<Integer> matchedHosts = new ArrayList<>();
+            for (Iterator<Map.Entry<Integer, Host>> it = hostsById.entrySet().iterator(); it.hasNext();) {
```
n3world: It would make a lot more sense to invert these loops. You can iterate over the hostsById and then lookup the recovered partitions as the key in restoredPartitionsByHosts. This assumes that the lists of partitions are always in the same order since you are doing a string compare and I am not sure if that is guaranteed anywhere in these APIs.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -985,6 +1022,83 @@ public static AbstractTopology mutateRecoverTopology(AbstractTopology topology,
...
+                Host host = it.next().getValue();
+                String partitions = host.getPartitionIdList().stream()
+                        .map(n -> n.toString()).collect(Collectors.joining( "," ));
+
+                // host the same list of partitions but different host id, a candidate for swap
+                if (partitions.equals(entry.getKey())) {
+                    if (!restoredHosts.remove(new Integer(host.id))) {
```
n3world: Use of `new Integer()` is discouraged. It is usually better to use `Integer.valueOf()`

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -262,14 +265,26 @@ public String toString() {
         private final static String HOST_IP = "hostIp";
         private final static String GROUP = "group";
         private final static String LOCAL_SITES_COUNT = "localSitesCount";
+        private final static String RECOVERED_PARTITION_IDS = "recoveredPartitions";
         public final String m_hostIp;
         public final String m_group;
         public final int m_localSitesCount;
 
+        // comma separated partition ids
+        public final String m_recoveredPartitions;
```
n3world: Instead of a String would it make more sense for this to be a Set of Inetgers so that type safety and other guarantees are maintained explicitly rather than implicitly

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -2161,9 +2190,26 @@ private AbstractTopology getTopology(StartAction startAction, Map<Integer, HostI
...
             }
-            TopologyZKUtils.registerTopologyToZK(m_messenger.getZK(), topology);
+            if (m_config.m_restorePlacement && m_config.m_startAction.doesRecover()){
+                Map<String, List<Integer>> restoredPartitionsByHosts = Maps.newHashMap();
+                hostInfos.forEach((k, v) -> {
+                    if (!StringUtils.isEmpty(v.m_recoveredPartitions)) {
+                        List<Integer> hosts = restoredPartitionsByHosts.get(v.m_recoveredPartitions);
+                        if (hosts == null) {
+                            hosts = Lists.newArrayList();
+                            restoredPartitionsByHosts.put(v.m_recoveredPartitions, hosts);
```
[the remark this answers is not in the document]
zeemanhe: will update this section

src/frontend/org/voltdb/RealVoltDB.java
zeemanhe: will do

src/frontend/org/voltcore/messaging/HostMessenger.java
zeemanhe: will be used

src/frontend/org/voltdb/AbstractTopology.java
zeemanhe: the partitions are always sorted. but I will update the codes with set diff

src/frontend/org/voltcore/messaging/HostMessenger.java
zeemanhe: This is a raw string retrieved from properties and used during mesh determination. use as is to avoid parsing.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -92,14 +94,18 @@
     public static class Partition implements Comparable<Partition>{
         public final int id;
         public int leaderHostId;
```
n3world: Would you mind making leaderHostId final like all of the other variables

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -139,7 +145,7 @@ public int compareTo(Partition o) {
     public static class Host implements Comparable<Host> {
         public final int id;
         public final String haGroup;
-        public final ImmutableSortedSet<Partition> partitions;
+        public ImmutableSortedSet<Partition> partitions;
```
n3world: Having this not be final breaks the immutability of this class

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -936,18 +966,38 @@ public static AbstractTopology getTopology(Map<Integer, HostInfo> hostInfos, Set
...
+            }
+            if (replaceHost == null) {
+                replaceHost = h;
+            }
+
+            // use the matched host. if no match found, use any one available.
+            if (recoverPartitions != null && !recoverPartitions.isEmpty()) {
+                Set<Integer> partitions = Sets.newHashSet();
+                partitions.addAll(h.getPartitionIdList());
+                if (Sets.difference(partitions, recoverPartitions).isEmpty()) {
```
n3world: You could use `partitions.equals(recoverPartitions)`

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -180,6 +186,7 @@ private Host(int id, String haGroup, ImmutableSortedSet<Partition> partitions, b
         public List<Integer> getPartitionIdList() {
             return partitions.stream()
                     .map(p -> p.id)
+                    .sorted()
```
n3world: I believe the order is guaranteed in partitions by the sort order of Partition which is on the partition ID.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -924,8 +935,27 @@ private static TopologyBuilder addPartitionsToHosts(Map<Integer, HostInfo> hostI
...
+     * @param  restorePartition a flag for partition assignment recovery
+     * @throws RuntimeException if hosts are not valid for topology
+     */
+    public static AbstractTopology getTopology(Map<Integer, HostInfo> hostInfos, Set<Integer> missingHosts,
+            int kfactor, boolean restorePartition) {
         TopologyBuilder builder = addPartitionsToHosts(hostInfos, missingHosts, kfactor, 0);
-        return new AbstractTopology(EMPTY_TOPOLOGY, builder);
+        AbstractTopology topo = new AbstractTopology(EMPTY_TOPOLOGY, builder);
+        if (restorePartition) {
+            topo = restorePartitionsForRecovery(topo,hostInfos);
```
n3world: It seems unnecessary to have to create a topology in order to create the topology you actually want. Why not try to incorporate the recoveredPartitions into `addPartitionsToHosts()`?

src/frontend/org/voltcore/messaging/HostMessenger.java
```diff
@@ -279,14 +290,30 @@ public HostInfo(String hostIp, String group, int localSitesCount) {
...
+        public Set<Integer> getRecoveredPartitions() {
+            Set<Integer> partitionSet = Sets.newHashSet();
+            if (StringUtils.isEmpty(m_recoveredPartitions)) {
+                return partitionSet;
+            }
+            String partitions[] = m_recoveredPartitions.split(",");
+            for (String partition : partitions) {
+                try {
+                    partitionSet.add(Integer.valueOf(partition));
+                } catch (NumberFormatException e) {}
```
n3world: Isn't this a bad programming bug if the String does not contain a valid Integer? Should probably log this error.

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -936,18 +966,38 @@ public static AbstractTopology getTopology(Map<Integer, HostInfo> hostInfos, Set
...
+                continue;
+            }
+            if (replaceHost == null) {
+                replaceHost = h;
+            }
+
+            // use the matched host. if no match found, use any one available.
+            if (recoverPartitions != null && !recoverPartitions.isEmpty()) {
+                Set<Integer> partitions = Sets.newHashSet();
+                partitions.addAll(h.getPartitionIdList());
```
n3world: Instead of converting the List to a set you could replace `getPartitionIdList` with a method which returns a Set it looks like all except one caller in a test wouldn't really care

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1244,6 +1253,15 @@ public Boolean get() {
                     partitionGroupPeers = topo.getPartitionGroupPeers(m_messenger.getHostId());
                 }
                 m_messenger.setPartitionGroupPeers(partitionGroupPeers, m_clusterSettings.get().hostcount());
+
+                // persist the merged settings
+                Collections.sort(partitions);
+                m_config.m_recoveredPartitions = partitions.stream().map( n -> n.toString()).collect(Collectors.joining( "," ));
```
n3world: Better to use method lambdas when possible `Integer::toString`.

Not sure if it is better but it looks simpler to use Joiner: `m_config.m_recoveredPartitions = Joiner.on(",").join(partitions)`

src/frontend/org/voltdb/RealVoltDB.java
```diff
@@ -1244,6 +1253,15 @@ public Boolean get() {
...
+                clusterSettings = ClusterSettings.create(
+                        m_config.asClusterSettingsMap(), fromPropertyFile.asMap(), fromDeploymentFile);
+                clusterSettings.store();
+                m_clusterSettings.set(clusterSettings, 2);
```
n3world: Is it safe to do a set here and with a static stamp value?

src/frontend/org/voltdb/settings/ClusterSettings.java
```diff
@@ -40,6 +41,10 @@
     @DefaultValue("7")
     public int cangamanga();
 
+    @Key(PARTIITON_IDS)
+    @DefaultValue("")
+    public String partitionids();
```
n3world: Shouldn't the `i` in `ids` be capitalized

src/frontend/org/voltcore/messaging/HostMessenger.java
zeemanhe: The logging is intentionally ignored. The partition list is retrieved from previous run, and users are not aware of it. The partition restoration is best-effort try anyway. Warning or errors may confuse users.

src/frontend/org/voltdb/RealVoltDB.java
zeemanhe: The update with stamp 2 should be safe since previous stamp  is 1 and no other will update between. could use a variable though.

src/frontend/org/voltdb/settings/ClusterSettings.java
zeemanhe: Will do

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -106,6 +106,17 @@ public void updateHosts(Collection<Integer> hostIds) {
...
 
+        public ImmutableSortedSet<Integer> getHostIds(){
+            return hostIds;
+        }
+
+        public int getLeaderHostId() {
+            return leaderHostId;
+        }
+
+        public void setLeaderHostId(int hostId) {
```
n3world: This should probably be package protected method to at least keep the illusion of immutability

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -183,11 +194,14 @@ private Host(int id, String haGroup, ImmutableSortedSet<Partition> partitions, b
             this.isMissing = missing;
         }
 
-        public List<Integer> getPartitionIdList() {
+        public Set<Integer> getPartitionIdList() {
```
n3world: Method name should probably be updated

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -193,6 +214,10 @@ public int getleaderCount() {
             return leaders;
         }
 
+        public void updatePartitions(Collection<Partition> partitions) {
```
n3world: This should probably be package protected method to at least keep the illusion of immutability

PR #5880, opened 2019-01-20
Eng 15212 path.properties not atomic

prosegay:
saving path.properties (after normalizing paths, etc.) was not atomic. This showed up on k8s with path.properties being left empty. k8s has a tendency to kill the process at unexpected times. The fix is to create the new file as a temp file and then use atomic rename operation to update the file. Atomicity of the rename operation is platform dependent, however.

### DIFF
--- a/src/frontend/org/voltdb/settings/NodeSettings.java
+++ b/src/frontend/org/voltdb/settings/NodeSettings.java
@@ -233,8 +233,23 @@ public interface NodeSettings extends Settings {
     }
 
     default void store() {
+        for (File f : Settings.getConfigDir().listFiles()) {
+            if (f.getName().endsWith(".tmp")) {
+                f.delete();
+            }
+        }
+        File tempFH = null;
+        try {
+            tempFH = File.createTempFile("path", null, Settings.getConfigDir());
+        } catch (IOException e) {
+            throw new SettingsException("failed to create temp file in config dir");
+        }
         File configFH = new File(Settings.getConfigDir(), "path.properties");
-        store(configFH, "VoltDB path settings. DO NOT MODIFY THIS FILE!");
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -233,8 +233,23 @@ default Properties asProperties() {
     }
 
     default void store() {
+        for (File f : Settings.getConfigDir().listFiles()) {
+            if (f.getName().endsWith(".tmp")) {
```
n3world: I don't think the temporary files created by `File.createTempFile` have the suffix .tmp unless you set that as the second argument to the method.

src/frontend/org/voltdb/settings/NodeSettings.java
```diff
@@ -233,8 +233,23 @@ default Properties asProperties() {
...
+        }
+        File tempFH = null;
+        try {
+            tempFH = File.createTempFile("path", null, Settings.getConfigDir());
+        } catch (IOException e) {
+            throw new SettingsException("failed to create temp file in config dir");
+        }
         File configFH = new File(Settings.getConfigDir(), "path.properties");
-        store(configFH, "VoltDB path settings. DO NOT MODIFY THIS FILE!");
+        store(tempFH, "VoltDB path settings. DO NOT MODIFY THIS FILE!");
```
n3world: This could still technically lead to a corrupt file on a system crash. You would have to add the equivalent of at least a `fdatasync` here and a `fsync` on the final file after the rename.

src/frontend/org/voltdb/settings/NodeSettings.java
n3world: Never mind it does use the `.tmp` suffix if the supplied suffix is `null`.

PR #5883, opened 2019-01-22
ENG-15070: Rename JOIN log facility to ELASTIC

n3world:
Use the log component ELASTIC instead of JOIN for all elastic operatins.
Also change references to elastic join to elastic operation when it is
not specifically about joins.


manjujames:
When the upgrade, JOIN log will stop working. This might be OK because it is a major version, but we will have to doc it in release notes or somewhere.

Other option would be to make JOIN also work and write to ELASTICLOG. Still needs a release note, I think.

prosegay:
this will impact system tests we match messages in the log and no doubt some of them may have the logger source included.

n3world:
> won't this will impact system tests we match messages in the log and no doubt some of them may have the logger source included.

I tried to update all the tests I found in the pro version of this change. https://github.com/VoltDB/pro/pull/2461

PR #5884, opened 2019-01-22
ENG-15220

kaikaizi:
Warn and convert AssumeUnique index on replicated table into Unique index on `CREATE ASSUMEUNIQUE INDEX ...` statement,
since EE treats the two indexes identically for replicated table.

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -1833,7 +1833,7 @@ public class DDLCompiler {
         }
 
         index.setUnique(unique);
-        if (! index.getTypeName().startsWith("VOLTDB_AUTOGEN_IDX_") && table.getIsreplicated() && assumeUnique) {
+        if (! index.getTypeName().startsWith(HSQLInterface.AUTO_GEN_PREFIX) && table.getIsreplicated() && assumeUnique) {
             // Warn and convert AssumeUnique -> Unique index only on
             // non-auto-generated index (i.e. from "CREATE ASSUMEUNIQUE INDEX ..." statement rather than "CREATE TABLE(...)" statement),
             // because "PARTITION TABLE ..." statement cannot be specified together with "CREATE TABLE" statement!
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1827,7 +1833,16 @@ else if (isHashIndex) {
         }
 
         index.setUnique(unique);
-        if (assumeUnique) {
+        if (! index.getTypeName().startsWith("VOLTDB_AUTOGEN_IDX_") && table.getIsreplicated() && assumeUnique) {
```
yzhang1991: There are a bunch of string constants in HSQLInterface.java, use them instead of having magic strings here. And yes, they should be moved over to some new places after HSQL is deprecated.

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1827,7 +1833,16 @@ else if (isHashIndex) {
...
+            // Warn and convert AssumeUnique -> Unique index only on
+            // non-auto-generated index (i.e. from "CREATE ASSUMEUNIQUE INDEX ..." statement rather than "CREATE TABLE(...)" statement),
+            // because "PARTITION TABLE ..." statement cannot be specified together with "CREATE TABLE" statement!
+            final String warn = String.format("On replicated table %s, ASSUMEUNIQUE index %s is converted to UNIQUE index.",
```
yzhang1991: Ask Andrew, "is this message OK?"

tests/frontend/org/voltdb/TestAdhocCreateDropIndex.java
```diff
@@ -474,6 +474,50 @@ public void testENG15213() throws Exception {
 
     }
 
+    @Test
+    public void testENG15220() throws Exception {
+        final VoltDB.Configuration config = new VoltDB.Configuration();
+        final String ddl = "CREATE TABLE R1 (ID INTEGER NOT NULL PRIMARY KEY, TINY TINYINT);";
+        try {
+            createSchema(config, ddl, 2, 1, 0);
+            startSystem(config);
```
yzhang1991: A minor thing, you can determine by yourself if it is worth doing.
If you can use `LocalCluster` to write this test instead of `ServerThread`, then there is an API called `verifyLogMessage` that you can use to verify the warning was given by the system.

src/frontend/org/voltdb/compiler/DDLCompiler.java
kaikaizi: Updated as per Andrew's suggestion.

PR #5885, opened 2019-01-22
ENG-15141: createTransaction use return code

n3world:
With the ability to remove partitions from the system a boolean is not
descriptive enough of a return code for InvocationDispatcher.createTransaction().
Instead return an enum which allows the caller to determine the cause of the
error better.


manjujames:
Looks good.

CreateTransactionResult is so small and specific that I wonder if it should be part of ClientInterface or InvocationDispatcher or so.

n3world:
> Looks good.
> 
> CreateTransactionResult is so small and specific that I wonder if it should be part of ClientInterface or InvocationDispatcher or so.

I originally had it as a class inside of InvocationDispatcher but that made the path strange since most handlers of the result were not directly invoking InvocationDispatcher they were invoking something that wrapped it. Because of that I made it its own standalone class.

PR #5888, opened 2019-01-29
ENG-15078, fix issue that during rollback EE generates invalid sequen…

luyangco:
…ce number range for export buffers.

In general row count of the block needs to be updated when
1) new transaction starts,
2) new block is created after the old one is full,
3) rollback one or more blocks,

### DIFF
--- a/src/ee/common/StreamBlock.h
+++ b/src/ee/common/StreamBlock.h
@@ -230,7 +230,7 @@ namespace voltdb
         }
 
         void truncateTo(size_t mark, int64_t exportSeqNo) {
-            // just move offset and row count (only export block cares). pretty easy.
+            // just move offset and update export row count. pretty easy.
             if (((m_uso + offset()) >= mark ) && (m_uso <= mark)) {
                 m_offset = mark - m_uso;
                 if (lastExportSequenceNumber() >= exportSeqNo && startExportSequenceNumber() <= exportSeqNo) {
--- a/src/ee/common/StreamBlock.h
+++ b/src/ee/common/StreamBlock.h
@@ -229,10 +229,13 @@ namespace voltdb
             m_offset += consumed;
         }
 
-        void truncateTo(size_t mark) {
-            // just move offset. pretty easy.
+        void truncateTo(size_t mark, int64_t exportSeqNo) {
+            // just move offset and row count (only export block cares). pretty easy.
             if (((m_uso + offset()) >= mark ) && (m_uso <= mark)) {
                 m_offset = mark - m_uso;
+                if (lastExportSequenceNumber() >= exportSeqNo && startExportSequenceNumber() <= exportSeqNo) {
+                    m_rowCountForExport = exportSeqNo - startExportSequenceNumber() + 1;
+                }
             }
             else {
                 throwFatalException("Attempted Export block truncation past start of block."
--- a/src/ee/storage/ExportTupleStream.h
+++ b/src/ee/storage/ExportTupleStream.h
@@ -61,7 +61,7 @@ public:
         assert(m_uso == 0);
         m_uso = count;
         // this is for start sequence number of stream block
-        m_exportSequenceNumber = seqNo + 1;
+        m_exportSequenceNumber = seqNo;
         //Extend the buffer chain to replace any existing stream blocks with a new one
         //with the correct sequence number
         extendBufferChain(0);
--- a/src/ee/storage/TupleStreamBase.cpp
+++ b/src/ee/storage/TupleStreamBase.cpp
@@ -133,9 +133,6 @@ void TupleStreamBase::commit(int64_t lastCommittedSpHandle, int64_t currentSpHan
         return;
     }
 
-    m_currBlock->updateRowCountForExport(m_uncommittedTupleCount);
-    m_uncommittedTupleCount = 0;
-
     // If the current TXN ID has advanced, then we know that:
     // - The old open transaction has been committed
     // - The current transaction is now our open transaction
@@ -156,6 +153,8 @@ void TupleStreamBase::commit(int64_t lastCommittedSpHandle, int64_t currentSpHan
         m_committedSpHandle = m_openSpHandle;
         m_openSpHandle = currentSpHandle;
         m_openUniqueId = uniqueId;
+        m_currBlock->updateRowCountForExport(m_uncommittedTupleCount);
+        m_uncommittedTupleCount = 0;
 
         if (flush) {
             extendBufferChain(0);
@@ -171,6 +170,8 @@ void TupleStreamBase::commit(int64_t lastCommittedSpHandle, int64_t currentSpHan
         m_committedUso = m_uso;
         m_committedSpHandle = m_openSpHandle;
         m_committedUniqueId = m_openUniqueId;
+        m_currBlock->updateRowCountForExport(m_uncommittedTupleCount);
+        m_uncommittedTupleCount = 0;
 
         if (flush) {
             extendBufferChain(0);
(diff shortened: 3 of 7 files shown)


src/ee/storage/TupleStreamBase.cpp
```diff
@@ -223,9 +223,16 @@ void TupleStreamBase::rollbackTo(size_t mark, size_t, int64_t exportSeqNo)
...
-    m_uncommittedTupleCount -= m_exportSequenceNumber - exportSeqNo;
-    assert (m_uncommittedTupleCount >= 0 || exportSeqNo == SIZE_MAX);
-    m_exportSequenceNumber = exportSeqNo;
+    // TODO:
+    // Because right now every appendTuple in export stream incorporates a commit,
+    // most of time uncommitted tuple count will be 1, if there are multiple stream table inserts
+    // in a transaction and it is rolled back, uncommitted tuple count may be negative.
+    // The real fix would be stream table commits at end of transaction or flush.
+    m_uncommittedTupleCount -= m_exportSequenceNumber - rollbackTo;
+    if (m_uncommittedTupleCount < 0 && rollbackTo != SIZE_MAX) {
```
rdykiel: Why keep the test on SIZE_MAX? Does this mean you accept a negative uncommitted tuple count? Why no floor it to zero systematically?

luyangco: `rollbackTo` equals to SIZE_MAX when DR rollback its transaction, that's the reason for testing the condition but I agree with your comment, `m_uncommittedTupleCount` is only used in export, DR doesn't care negative `m_uncommittedTupleCount `. Walter's ongoing patch might change the encapsulation of `TupleStreamBase`, but before that `m_uncommittedTupleCount` can be floor to zero systematically.

src/ee/common/StreamBlock.h
```diff
@@ -229,10 +229,13 @@ namespace voltdb
             m_offset += consumed;
         }
 
-        void truncateTo(size_t mark) {
-            // just move offset. pretty easy.
+        void truncateTo(size_t mark, int64_t exportSeqNo) {
+            // just move offset and row count (only export block cares). pretty easy.
```
rdykiel: Clarify this comment I don't understand "only export block cares"

src/ee/storage/TupleStreamBase.h
```diff
@@ -118,10 +118,10 @@ class TupleStreamBase {
     size_t m_headerSpace;
 
     /**
-     * The number of Export Tuples applied to the Export Stream Block in the current txn;
-     * Note that before the Export Tuples are only committed by the *next* Txn that updates the StreamBLock
-     */
-    int64_t m_uncommittedTupleCount;
+    * The number of Export Tuples applied to the Export Stream Block in the current txn;
```
rdykiel: Should it be commented as this (help me understand the comment):

"Number of export tuples that were inserted in the Export Stream Block in the current transaction"

Then add something about how you are using this.

src/ee/storage/TupleStreamBase.h
```diff
@@ -118,10 +118,10 @@ class TupleStreamBase {
...
+    * Note that before the Export Tuples are only committed by the *next* Txn that updates the StreamBLock
+    */
+    int64_t m_stashedTupleCount;
```
rdykiel: Should this be named "m_currTxnTupleCount" ? Or something like this? "stash" doesn't mean anything to me. Maybe that's just me?

src/ee/storage/TupleStreamBase.cpp
```diff
@@ -286,6 +288,8 @@ void TupleStreamBase::extendBufferChain(size_t minLength)
 
     if (m_currBlock) {
         if (m_currBlock->offset() > 0) {
+            m_currBlock->updateRowCountForExport(m_stashedTupleCount);
```
rdykiel: Should it be commented like this? "Update tuple count in current block with the number of tuples inserted so far by the current transaction"

src/ee/storage/ExportTupleStream.h
```diff
@@ -61,7 +61,7 @@ class ExportTupleStream : public voltdb::TupleStreamBase {
         assert(m_uso == 0);
         m_uso = count;
         // this is for start sequence number of stream block
-        m_exportSequenceNumber = seqNo + 1;
+        m_exportSequenceNumber = seqNo;
```
rdykiel: Why are we changing the renumbering? This looks suspicious to me.

src/ee/storage/TupleStreamBase.cpp
wweiss-voltdb: m_uncommittedTupleCount is only relevant to the first buffer of the transaction so if a transaction fills a block partially filled by another txn and fills another block, the only thing we care about and the only thing that m_uncommittedTupleCount should be used for is to revert the tuple count of the original block back to it's previous value.

src/ee/common/StreamBlock.h
luyangco: Update the comment to "just move offset and update export row count"

src/ee/storage/TupleStreamBase.h
luyangco: Comment is updated.

src/ee/storage/TupleStreamBase.h
luyangco: Maybe another name can describe the intention better, but to use "stash" I mean the export tuples that hasn't be reflected in the header (row count field) of stream block. 
For example, a single transaction inserts 10 export tuples, at the 3rd insert the "stash" tuple count is 3, now current stream block is full, so the row count field increments by 3 then reset the "stash" tuple count, after all this is done the new block is created. Similarly at end of transaction the "stash" tuple count is 7, it waits for either commit or rollback or extend new block to flush the value to stream block header.

src/ee/storage/TupleStreamBase.cpp
luyangco: In my new commit, `m_uncommittedTupleCount` is replaced by `m_stashedTupleCount` because stream block doesn't need additional book keeping to know how to restore the previous value of row count during rollback. `rollbackToSeqNo` - `startSeqNo` + 1 can restore the value.

The only purpose to have `m_stashedTupleCount` is that it simplifies extendBufferChain() between transaction spans multiple blocks and tick (or periodicFlush), because we don't know whether the insert is successful until very late and tick() needs to know whether to update row count header of current block for previous insert.

src/ee/storage/TupleStreamBase.cpp
luyangco: Updated.

src/ee/storage/ExportTupleStream.h
luyangco: Actually the previous way of tracking `m_exportSequenceNumber` in tuple stream is a little bit hacky, it assumes the starting sequence number of new block is currentSeqNo + 1, so the `m_exportSequenceNumber` advances even before the new insert happens. 

I changed it so `m_exportSequenceNumber` advances only in appendTuple(), new block updates its starting sequence number at first tuple insert to the block.

src/ee/storage/TupleStreamBase.h
rdykiel: I think the grammar of this comment needs fixing; should it be the following:

Count of exported tuples not yet counted in the row count header of the Export Stream Block for the current transaction; this count will be added to the row count header (and this count reset to 0) on the following conditions:
1) current transaction appears "committed" when we start handling the *next* transaction
2) when the current transaction needs a new Stream Block
3) current transaction is rolled back (with the rollback affecting one or more Stream blocks)

luyangco: Thanks, I updated the comment.

luyangco:
Merge externally, close this one.

PR #5889, opened 2019-01-29
ENG-15254: Once all parititon data is migrated away from partitions b…

manjujames:
…eing removed,

generate elastic change event for DR consumers and set the sites for the partitions
being removed to a mode where they won't write anymore DR or export data.

### DIFF
--- a/src/ee/storage/persistenttable.cpp
+++ b/src/ee/storage/persistenttable.cpp
@@ -332,7 +332,11 @@ void PersistentTable::deleteAllTuples(bool, bool fallible) {
 }
 
 bool PersistentTable::doDRActions(AbstractDRTupleStream* drStream) {
-    return m_drEnabled && drStream && drStream->drStreamStarted();
+    //TODO: See if we can set drStream to NULL when streams are disabled
+    // to avoid adding another if check here.
+    ExecutorContext* ec = ExecutorContext::getExecutorContext();
+    return m_drEnabled && drStream && drStream->drStreamStarted() &&
+          (isReplicatedTable() || ec->externalStreamsEnabled());
 }
 
 void PersistentTable::truncateTableUndo(TableCatalogDelegate* tcd,
--- a/src/ee/storage/streamedtable.cpp
+++ b/src/ee/storage/streamedtable.cpp
@@ -131,6 +131,13 @@ void StreamedTable::nextFreeTuple(TableTuple *) {
                                   "May not use nextFreeTuple with streamed tables.");
 }
 
+bool StreamedTable::shouldStreamToExport() {
+    //TODO: See if we can set wrapper to null or similar trick when streams are disabled
+    // to avoid the extra if check
+    ExecutorContext* ec = ExecutorContext::getExecutorContext();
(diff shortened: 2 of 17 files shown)


src/ee/storage/persistenttable.cpp
```diff
@@ -332,7 +332,11 @@ void PersistentTable::deleteAllTuples(bool, bool fallible) {
 }
 
 bool PersistentTable::doDRActions(AbstractDRTupleStream* drStream) {
-    return m_drEnabled && drStream && drStream->drStreamStarted();
+    //TODO: See if we can set drStream to NULL when streams are disabled
+    // to avoid adding another if check here.
+    ExecutorContext* ec = ExecutorContext::getExecutorContext();
+    return m_drEnabled && drStream && drStream->drStreamStarted() &&
+          (isReplicatedTable() || ec->allStreamsEnabled());
```
n3world: Why do you still want to do DR actions for replicated tables when all streams are disabled?

src/ee/storage/streamedtable.h
```diff
@@ -92,6 +92,7 @@ class StreamedTable : public Table {
 
     //Override and say how many bytes are in Java and C++
     int64_t allocatedTupleMemory() const;
+    bool streamToExport();
```
n3world: The name of this method sounds like it is doing the stream to the export not if it should do the stream to export

src/frontend/org/voltdb/SiteProcedureConnection.java
```diff
@@ -262,4 +262,16 @@ public void exportAction(boolean syncAction,
     public void generateElasticChangeEvents(int oldPartitionCnt, int newPartitionCnt, long txnId, long spHandle, long uniqueId);
 
     public void generateElasticRebalanceEvents(int srcPartition, int destPartition, long txnId, long spHandle, long uniqueId);
+
+    /**
+     * Use this to disable all external streams (DR, export) from this site.
```
n3world: This doc implies that we only support disabling. What would happen if somebody called this method with a value of true when it was previously false. That would break things right? Maybe it would be better to just have a disable method that doesn't take any arguments so that we only allow a one way transition

src/ee/storage/streamedtable.cpp
```diff
@@ -139,7 +146,8 @@ bool StreamedTable::insertTuple(TableTuple &source)
     }
 
     size_t mark = 0;
-    if (m_wrapper) {
+    //if (m_wrapper) {
```
n3world: Remove commented out code same for line 203

src/frontend/org/voltdb/jni/ExecutionEngineIPC.java
```diff
@@ -132,7 +132,8 @@
         , ExecuteTask(28)
         , ApplyBinaryLog(29)
         , ShutDown(30)
-        , SetViewsEnabled(31);
+        , SetViewsEnabled(31)
+        , SetExternalStreamsEnabled(32);
```
n3world: This should be updated to DisableExternalStreams

PR #5891, opened 2019-01-30
Removed unused DtxnConstants.

manjujames:
Removed unnecessary DEP_ constants from sysprocs.


src/frontend/org/voltcore/agreement/DtxnConstants.java
```diff
@@ -18,13 +18,5 @@
...
 
 public abstract class DtxnConstants {
-
-    /** If the txn id is this, do a dump from the ExecSite */
-    public static final long DUMP_REQUEST_TXNID = -101;
-
-    /** Dtxn requires 1 dependency response per partition */
-    public static final int MULTIPARTITION_DEPENDENCY = 0x40000000;
-
     public static final long DUMMY_LAST_SEEN_TXN_ID = -1;
```
n3world: I wonder if there is a better home for this rather than a class with a single static value

src/frontend/org/voltdb/VoltSystemProcedure.java
```diff
@@ -134,6 +133,34 @@ void initSysProc(SiteProcedureConnection site,
...
+
+        /**
+         * Most MP sysprocs use this pattern of MP fragment and a non-MP aggregator fragment.
+         * Utility method to create that.
+         * @param fragmentId
+         * @param aggFragmentId
+         * @param params
+         * @return Array of fragments in the correct order
+         */
+        public static SynthesizedPlanFragment[] createFragmentAndAggregator(long fragmentId, long aggFragmentId, Object... params) {
```
n3world: Should this be protected or does it actually need to be public?

src/frontend/org/voltdb/VoltSystemProcedure.java
```diff
@@ -134,6 +133,34 @@ void initSysProc(SiteProcedureConnection site,
          * like user sysprocs.
          */
         public boolean suppressDuplicates = false;
+
+        /**
+         * Most MP sysprocs use this pattern of MP fragment and a non-MP aggregator fragment.
+         * Utility method to create that.
+         * @param fragmentId
```
n3world: If it isn't worth adding the doc for the parameters might as well just remove them from the javadoc

src/frontend/org/voltdb/VoltSystemProcedure.java
```diff
@@ -134,6 +133,34 @@ void initSysProc(SiteProcedureConnection site,
...
+         * @param aggFragmentId
+         * @param params
+         * @return Array of fragments in the correct order
+         */
+        public static SynthesizedPlanFragment[] createFragmentAndAggregator(long fragmentId, long aggFragmentId, Object... params) {
+            SynthesizedPlanFragment pfs[] = new SynthesizedPlanFragment[2];
+
+            pfs[0] = new SynthesizedPlanFragment();
+            pfs[0].fragmentId = fragmentId;
+            pfs[0].inputDepIds = new int[] {};
```
n3world: Could use `ArrayUtils.EMPTY_INT_ARRAY` to avoid creating a new object

src/frontend/org/voltdb/sysprocs/ExportControl.java
```diff
@@ -127,7 +120,7 @@ public DependencyPair executePlanFragment(Map<Integer, List<VoltTable>> dependen
         SynthesizedPlanFragment[] pfs = new SynthesizedPlanFragment[2];
```
n3world: Couldn't this use your new helper method or are you planning on doing that over time?

src/frontend/org/voltdb/VoltSystemProcedure.java
```diff
@@ -134,6 +133,34 @@ void initSysProc(SiteProcedureConnection site,
...
+            pfs[0].parameters = ParameterSet.fromArrayNoCopy(params);
+
+            pfs[1] = new SynthesizedPlanFragment();
+            pfs[1].fragmentId = aggFragmentId;
+            pfs[1].inputDepIds = new int[]{ (int) fragmentId };
+            pfs[1].outputDepId = (int) aggFragmentId;
+            pfs[1].multipartition = false;
+            pfs[1].parameters = ParameterSet.emptyParameterSet();
+
+            return pfs;
```
n3world: You could probably even do the execute call in this helper method since that is what most sys procs do next too.

`return executeSysProcPlanFragments(pfs, (int) aggFragmentId);`

src/frontend/org/voltdb/sysprocs/SysProcFragmentId.java
```diff
@@ -21,14 +21,14 @@
 
 public class SysProcFragmentId
 {
-    // @LastCommittedTransaction
+    // @LastCommittedTransaction -- UNUSED
```
n3world: Can the unused fragment IDs be removed?

It would also clean up a lot of casting if all fragment IDs here were declared as int and then just expanded to long upon assignment to the fragment ID in SynthesizedPlanFragment

src/frontend/org/voltdb/VoltSystemProcedure.java
manjujames: This is meant to be used by any system procedure that needs to create this pattern of plan fragment, so public makes logical sense.

src/frontend/org/voltdb/VoltSystemProcedure.java
manjujames: Yes, I will add docs. Thanks.

src/frontend/org/voltdb/VoltSystemProcedure.java
manjujames: OK, thanks.

src/frontend/org/voltdb/sysprocs/ExportControl.java
manjujames: Yes. However, I was trying to avoid the number of manual and boring changes that is time consuming and error prone. We will have to clean up the rest as we work on these classes.

src/frontend/org/voltdb/sysprocs/SysProcFragmentId.java
manjujames: I tried making them int. But, then there are generated fragment ids that also use long (they just use numbers after 5000). Making all fragment ids ints is changing a lot of things and my first attempt broke a lot of basic junits. Might be good to it at some point (or make dependency ids longs), but it is not something I want to spend time on now.

manjujames: The unused fragment ids can probably be removed. But, may be another clean up task. I want to be done with this clean up for now :)

src/frontend/org/voltdb/VoltSystemProcedure.java
n3world: But don't all system procedures have to extend this class?

src/frontend/org/voltdb/sysprocs/SysProcFragmentId.java
n3world: I wouldn't make all fragment IDs int just the system procedure ones since they have to be int because of the fragment ID being used as a dependency ID

src/frontend/org/voltdb/VoltSystemProcedure.java
manjujames: This is on SynthesizedPlanFragment.

src/frontend/org/voltdb/sysprocs/SysProcFragmentId.java
manjujames: It is not a quick task to do it cleanly because there are some methods that are for all fragment IDs. I spent too much time on it already, which my goal was just to get rid of that MULTIPARTITION_DEPENDENCY, which had everyone confused.

src/frontend/org/voltdb/VoltSystemProcedure.java
n3world: This method is in VoltSystemProcedure and returns SynthesizedPlanFragment[]

PR #5892, opened 2019-01-30
ENG-14773: Add final snapshot to elastic remove

n3world:
Update NativeSnapshotWritePlan to be able to update the partition count in the system
Make Cartogropher the authoritative place to retrieve the partition count from
Do not pass partition count into CommandLog

### DIFF
--- a/src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
+++ b/src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
@@ -71,9 +71,10 @@ public class NativeSnapshotWritePlan extends SnapshotWritePlan
                                             HashinatorSnapshotData hashinatorData,
                                             long timestamp)
     {
+        Integer newPartitionCount = jsData == null ? null : (Integer) jsData.opt("partitionCount");
         return createSetupInternal(file_path, pathType, file_nonce, txnId, partitionTransactionIds,
                 jsData, context, result, extraSnapshotData, tracker, hashinatorData,
-                timestamp, context.getNumberOfPartitions());
+                timestamp, newPartitionCount);
     }
 
     Callable<Boolean> createSetupInternal(String file_path, String pathType,
@@ -87,7 +88,7 @@ public class NativeSnapshotWritePlan extends SnapshotWritePlan
                                                     SiteTracker tracker,
                                                     HashinatorSnapshotData hashinatorData,
                                                     long timestamp,
-                                                    int newPartitionCount)
+                                                    Integer newPartitionCountInteger)
     {
         assert(SnapshotSiteProcessor.ExecutionSitesCurrentlySnapshotting.isEmpty());
         if (hashinatorData == null) {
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
```diff
@@ -161,14 +159,14 @@
         placeReplicatedTasks(replicatedSnapshotTasks, tracker.getSitesForHost(context.getHostId()));
 
         boolean isTruncationSnapshot = true;
-        if (jsData != null) {
+        if (jsData != null && config.partitionCount == null) {
```
manjujames: What is this doing? If the request config doesn't have partition count and it has truncReqId, then it is truncation snapshot? May be add a comment as to when all this means and how it is used?

n3world: It is a continuation of the semi truncation snapshot hack that stream snapshot does. This makes it so we get the same behavior on the final snapshot from remove. I can add a comment but either way it is an ugly hack. It would be good to eventually cleanup the snapshot request/configuration api

src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
```diff
@@ -96,23 +95,22 @@
...
-        }
+        // TRAIL [SnapSave:5] - 3.2 [1 site/host] Get list of tables to save and create tasks for them.
+        final Table[] tableArray = config.tables;
 
         final int newPartitionCount;
-        if (newPartitionCountInteger == null) {
-            newPartitionCount = context.getNumberOfPartitions();
+        final int partitionCount;
+        if (config.partitionCount != null) {
+            partitionCount = newPartitionCount = config.partitionCount;
```
JadenW: This is for the final snapshot path?

src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
```diff
@@ -96,23 +95,22 @@
...
+        } else if (newPartitionCountInteger == null) {
+            partitionCount = newPartitionCount = context.getNumberOfPartitions();
         } else {
             newPartitionCount = newPartitionCountInteger.intValue();
+            partitionCount = context.getNumberOfPartitions();
```
JadenW: This is currently only for join path?

src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
n3world: Yes. I just pushed a change which cleans up the newPartitionCount handling

src/frontend/org/voltdb/sysprocs/saverestore/NativeSnapshotWritePlan.java
n3world: Yes. But this has changed in the latest diff where I collapsed remove and join handling of newPartitionCount into one path.

PR #5893, opened 2019-01-30
Calcite SetOps support ENG-15160

alexeevm:
Calcite Logical to Volt Logical Set Ops conversion rules and RelNodes. The MPFallBackRule for SetOps and Volt Logical to Volt Physical conversion rules are still in progress.

TestLogicalSetOpsRules.testSetOpsWithExpressionSubqueiries also requires additional validation


src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -81,9 +81,12 @@ private MPQueryFallBackRule() {
                             || ! isSinglePartitioned(calc.getProgram(), calc.getProgram().getCondition(), tableDist.getKeys()))) {
                     throw new PlannerFallbackException("MP query not supported in Calcite planner.");
             }
-            call.transformTo(calc.copy(calc.getTraitSet().replace(tableDist), calc.getInputs()));
+            RelNode newTableScan = tableScan.copy(tableScan.getTraitSet().replace(tableDist), tableScan.getInputs());
+            // @TODO Need to adjust indexes from the table distribution for a possible Calc projection
+            call.transformTo(calc.copy(calc.getTraitSet().replace(tableDist), newTableScan, calc.getProgram()));
```
Chaozz: I understand the issue, do you have a plan?

src/frontend/org/voltdb/plannerv2/rules/physical/VoltPSetOpsRule.java
```diff
@@ -0,0 +1,91 @@
...
+                return new VoltPhysicalMinus(setOp.getCluster(), convertedTraits, convertedInputs, setOp.all, splitCount);
+            }
+        };
+
+        public abstract RelNode convertToVoltPhysical(SetOp setOp, RelTraitSet convertedTraits, List<RelNode> convertedInputs, int splitCount);
+    }
+
+    private MatchType m_matchType;
+
+    VoltPSetOpsRule(RelOptRuleOperand operand, MatchType matchType) {
```
Chaozz: The constructor can be private.

alexeevm: Addressed

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
alexeevm: My VoltDBRexUtil provides a method to adjust collation indexes 
https://github.com/alexeevm/voltdb/blob/43d545598061dd11af58779f936659f925f446be/src/frontend/org/voltdb/calciteadapter/util/VoltDBRexUtil.java#L109

We can reuse the same implementation to adjust distribution's indexes as well. The IF part that matches Calc / Scan would have to combine both programs prior to adjusting the distribution. The ELSE part needs to do it only if the top node (call.rel(0)) is of VoltLogicaCalc type because of its project. All other Volt Logical types do not have a project.
Would this work?

Chaozz: Make sense. Thanks for the explanation.

alexeevm: @Chaozz Just to confirm, if C is a partitioning column for P and its index let say is 2, the scan's distribution trait will be [2].
If we have SQLs 
 - select C from P -- calc's distribution will be [0] // C's index becomes 0
- select C1 from P -- calc's distribution will be [-1] // C is projected out
- select C*2 from P -- calc's distribution will be [-1] // C is projected out and replaced with C*2 expression
Does it seem right to your?

Chaozz: what is the difference between
`select C1 from P` and `select C2 from P`?

Chaozz: It looks to me for `select C * 2 from P`, the calc's distribution will be [0]. Since the expression is only based on C

alexeevm: Ok, yes, in this case we can drill down to a single column but consider a bit more complicated scenario with a table partitioned on two columns (C1, C2) if VoltDB supports such cases

`select C1 + C2 from P`

what would be Calc's distribution here?

Chaozz: Oh, I got it.

alexeevm: According to the VoltDB documentation, there can be only one partitioning column per table. So, my example with two partitioning columns is invalid.
On the other hand,
`select SI from P where SI = 5;`
where the partitioning column's (SI) type is smallint is currently failing the MPFallback test simply because Calcite adds a  CAST expression `( INT CAST) SI = 5`
I wonder should we modify isSinglePartitioned to explicitly handle CAST?
Also, what about 
`select SI from P where abs(SI) = 5;`

alexeevm: Actually, it's a bit more complicated than I initially thought. Consider a Join
`select P1.PC*2 from P1 join P2 where P1.PC*3 = P2.PC;`
The outer child would be Calc / Scan. The scan project will have two expressions (`P1.PC*2` and `P1.PC*3`). Both of them refer to the partitioning column P1.PC which itself won't be part of the projected columns. Initially I thought, if a partitioning column itself is not part of the projection, I can take an index of the first RexCall expression that is solely based on the PC and create a new RelDistribution with it. But this won't work for the above example, because the join expression will be the second one and RelDistribution column index won't match.
Any thoughts?

kaikaizi: For `select P1.PC*2 from P1 join P2 where P1.PC*3 = P2.PC;`, PC is a partition column of both table P1 and P2, and are you seeking to make it a merge join?
Not sure that I could understand your statement: could you please be more explict for the part (first/second)?

>  Initially I thought, if a partitioning column itself is not part of the projection, I can take an index of the first RexCall expression that is solely based on the PC and create a new RelDistribution with it. But this won't work for the above example, because the join expression will be the second one and RelDistribution column index won't match.

alexeevm: Sorry for being unclear.  No, it's about whether this statement
`select P1.PC*2 from P1 join P2 where P1.PC*3 = P2.PC;`
should pass the MPFallback validation. As it stands today, it fails because  `P1.PC*3` is an expression itself similar to a CAST. If we do want it to pass we would have to propagate scan's partitioning info (distribution)  all the way up to the join. We do it today as part of the MPQueryFallback.rule but we never adjust initial partitioning column index - we simply copy the initial value.
A simplified example that should pass the MP validation but fails is
`SELECT p4.si FROM P4 join p2 on p4.i = p2.i WHERE 7 = p4.i`

Both tables P2 and P4 are partitioned on int column I. The P4.I column index is 3 for example. 

```
Join
   Calc projection (SI, I). Distribution [3]
      Scan P4 - Distribution [3]
  Calc projection (I) Distribution [0]
      Scan P2 - Distribution [0]

```
As you can see, the outer Calc has the distribution as [3] - the initial P4 distribution even the partitioning column I  index now should be 1 after the Calc's projection. As a result isSinglePartitioned gets invalid partitionsKey collection and returns false as a result.

Going back to the initial SQL, `select `P1.PC*2` from P1 join P2 where P1.PC*3 = P2.PC;` 

```
Join
   Calc projection (expr1, expr2). Distribution [????]
      Scan P1 - Distribution [0]
  Calc projection (I) Distribution [0]
      Scan P2 - Distribution [0]

```

What should be the outer Calc's distribution above? The outer Calc's projection would have two expression columns expr1 (`P1.PC*2`) and expr2(`P1.PC*3`) but not the P1.PC itself. For this query to pass, the distribution should be [1] indicating that the second expression is a partitioning column there. The problem is that the Calc's distribution is set when the MP rule matches a Calc / Scan expression and at this point we don't have any information about a potential join and its condition to set the expected distribution indexes. From the Calc point of view both expressions could be a partitioning one. Which one to pick?

PR #5896, opened 2019-02-01
ENG-15000 remove Kafka 8 reference and add Kerberos to Kafka 10 importer

zeemanhe:
Kafka 8, loader, importer and referenced libs are removed.


build-importers.xml
```diff
@@ -61,7 +61,7 @@
             <param name="include.classpattern" value="log4j/*.class"/>
         </antcall>
     </target>
-
+<!--
```
rdykiel: Why do you want to keep this? If you want to keep it, at least put a comment explaining

PR #5902, opened 2019-02-04
ENG-15291:

manjujames:
Store external streams states of sites in snapshot and use it at recover time.
Check for stream disabled was in the wrong place. Moved it out of persistent table
because it will never get there for partitioned table because there is nothing in the
hash ring for that partition. Added the check where we insert empty txn for MP txns for
partitions where it produced no actual data change.

### DIFF
--- a/src/ee/common/executorcontext.cpp
+++ b/src/ee/common/executorcontext.cpp
@@ -403,7 +403,7 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
 bool ExecutorContext::checkTransactionForDR() {
     bool result = false;
     if (UniqueId::isMpUniqueId(m_uniqueId) && m_undoQuantum != NULL) {
-        if (m_drStream && m_drStream->drStreamStarted()) {
+        if (m_externalStreamsEnabled && m_drStream && m_drStream->drStreamStarted()) {
             if (m_drStream->transactionChecks(m_lastCommittedSpHandle,
                     m_spHandle, m_uniqueId)) {
                 m_undoQuantum->registerUndoAction(
--- a/src/frontend/org/voltdb/ExtensibleSnapshotDigestData.java
+++ b/src/frontend/org/voltdb/ExtensibleSnapshotDigestData.java
@@ -19,10 +19,13 @@ package org.voltdb;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
+import org.json_voltpatches.JSONArray;
 import org.json_voltpatches.JSONException;
 import org.json_voltpatches.JSONObject;
 import org.json_voltpatches.JSONStringer;
@@ -41,6 +44,8 @@ public class ExtensibleSnapshotDigestData {
     public static final String EXPORT_SEQUENCE_NUMBER = "exportSequenceNumber";
     public static final String EXPORT_USO = "exportUso";
 
+    public static final String DISABLED_EXTERNAL_STREAMS = "disabledExternalStreams";
+
     /**
      * This field is the same values as m_exportSequenceNumbers once they have been extracted
      * in SnapshotSaveAPI.createSetup and then passed back in to SSS.initiateSnapshots. The only
@@ -54,6 +59,11 @@ public class ExtensibleSnapshotDigestData {
      * Same as m_exportSequenceNumbersToLogOnCompletion, but for m_drTupleStreamInfo
      */
     private final Map<Integer, TupleStreamStateInfo> m_drTupleStreamInfo;
+
+    /**
+     * Set of partitions where external streams are disabled.
+     */
+    private Set<Integer> m_disabledExternalStreams;
 
     /**
      * Used to pass the last seen unique ids from remote datacenters into the snapshot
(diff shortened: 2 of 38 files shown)


src/ee/common/executorcontext.cpp
```diff
@@ -403,7 +403,7 @@ void ExecutorContext::setDrReplicatedStream(AbstractDRTupleStream *drReplicatedS
 bool ExecutorContext::checkTransactionForDR() {
     bool result = false;
     if (UniqueId::isMpUniqueId(m_uniqueId) && m_undoQuantum != NULL) {
-        if (m_drStream && m_drStream->drStreamStarted()) {
+        if (m_drStream && m_drStream->drStreamStarted() && m_externalStreamsEnabled) {
```
n3world: Minor: I would put m_externalStreamsEnabled before m_drStream->drStreamStarted() just because it is cheaper to check a local variable than call a method

src/frontend/org/voltdb/ExtensibleSnapshotDigestData.java
```diff
@@ -159,6 +171,31 @@ private void mergeExportSequenceNumbersToZK(JSONObject jsonObj, VoltLogger log)
...
+            if (!jsonStates.has(partitionStr)) {
+                jsonStates.put(partitionStr, tableEntry.getValue());
+            } else {
+                boolean existing = jsonStates.getBoolean(partitionStr);
+                if (log.isDebugEnabled() && existing != tableEntry.getValue()) {
+                    // This should not happen. If it does, use false if it is false anywhere.
+                    log.debug("Found different external stream states on different hosts for partition " + partitionStr +
+                              ". This host has " + tableEntry.getValue());
+                }
+                jsonStates.put(partitionStr, tableEntry.getValue() & existing);
```
n3world: Minor but did you want logical and `&&` or did you mean bitwise and `&`? In this case they are roughly the same so it probably doesn't matter too much.

src/frontend/org/voltdb/ExtensibleSnapshotDigestData.java
```diff
@@ -54,6 +56,12 @@
      * Same as m_exportSequenceNumbersToLogOnCompletion, but for m_drTupleStreamInfo
      */
     private final Map<Integer, TupleStreamStateInfo> m_drTupleStreamInfo;
+    
+    /**
+     * Map from partition id to boolean indicating whether external streams (DR, export)
+     * are enabled for that partition.
+     */
+    private Map<Integer, Boolean> m_partitionExternalStreamsEnabled;
```
n3world: This might be simpler as a Set of partitions where streams are disabled

tests/frontend/org/voltdb/export/TestSocketExportServer.java
```diff
@@ -0,0 +1,197 @@
...
+                        continue;
+                    }
+                    String parts[] = m_parser.parseLine(line);
+                    if (parts == null) {
+                        continue;
+                    }
+                    Long i = Long.parseLong(parts[0]);
+                    if (m_seenIds.putIfAbsent(i, new AtomicLong(1)) != null) {
+                        synchronized(m_seenIds) {
+                            m_seenIds.get(i).incrementAndGet();
```
n3world: Since this is only accessed under a lock you don't really need an AtomicLong but it is an easy way to get a mutable numeric object.

src/frontend/org/voltdb/compiler/VoltProjectBuilder.java
```diff
@@ -739,14 +739,30 @@ public void addImport(boolean enabled, String importType, String importFormat, S
...
 
-    public void addExport(boolean enabled, String exportTarget, Properties config, String target) {
-        HashMap<String, Object> exportConnector = new HashMap<>();
-        exportConnector.put("elLoader", "org.voltdb.export.processors.GuestProcessor");
-        exportConnector.put("elEnabled", enabled);
+    public void addExport(boolean enabled, ServerExportEnum exportType, Properties config, String target) {
+        addExport(enabled, exportType, null, config, target);
+    }
+    public void addExport(boolean enabled, ServerExportEnum exportType, String connectorClass, Properties config, String target) {
+        org.voltdb.compiler.deploymentfile.ObjectFactory factory = new org.voltdb.compiler.deploymentfile.ObjectFactory();
```
n3world: Why does this use the fully qualified class and package?

src/frontend/org/voltdb/sysprocs/SnapshotRestore.java
```diff
@@ -266,21 +266,22 @@ private static synchronized BBContainer getNextChunk() throws IOException {
...
         {
-            assert(params.toArray()[0] != null);
-            assert(params.toArray().length == 6);
-            assert(params.toArray()[0] instanceof byte[]);
-            assert(params.toArray()[2] instanceof long[]);
-            assert(params.toArray()[3] instanceof Long);
-            assert(params.toArray()[4] instanceof Long);
-            assert(params.toArray()[5] instanceof Integer);
+            Object[] paramsArr = params.toArray();
+            assert(paramsArr[0] != null);
```
n3world: This is actually a bit redundant because of the instance of check a couple of lines down. null can not be an instance of anything.

src/frontend/org/voltdb/sysprocs/SnapshotRestore.java
```diff
@@ -266,21 +266,22 @@ private static synchronized BBContainer getNextChunk() throws IOException {
...
+            assert(paramsArr.length == 6);
+            assert(paramsArr[0] instanceof byte[]);
```
n3world: For completeness check that the object at index 1 is a Long

PR #5908, opened 2019-02-06
ENG-15339 consume complete transaction responses from replicas and do not forward

zeemanhe:


rdykiel:
I need a more detailed explanation of the scenario we're fixing.

rdykiel:
Approve

PR #5911, opened 2019-02-06
Eng 15337 pbd file management

rdykiel:
* Add a segment finalization mechanism: any file segment marked as "finalized" can be polled  without checking its internal consistency.
* Change the PBD file naming scheme to create an explicit linkage between PBD segments.

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
+import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -47,15 +48,17 @@ public class PBDRegularSegment extends PBDSegment {
 
     //ID of this segment
     private final Long m_index;
+    private final Instant m_tsCreated;
 
     private int m_numOfEntries = -1;
     private int m_size = -1;
 
     private DBBPool.BBContainer m_tmpHeaderBuf = null;
 
-    public PBDRegularSegment(Long index, File file) {
+    public PBDRegularSegment(Long index, Instant ts, File file) {
         super(file);
         m_index = index;
+        m_tsCreated = ts;
         reset();
     }
 
@@ -63,6 +66,11 @@ public class PBDRegularSegment extends PBDSegment {
     public long segmentId()
     {
         return m_index;
+    }
+
+    @Override
+    public Instant created() {
+        return m_tsCreated;
     }
 
     @Override
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -22,7 +22,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
-import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -46,31 +45,33 @@ public class PBDRegularSegment extends PBDSegment {
     private final Map<String, SegmentReader> m_readCursors = new HashMap<>();
     private final Map<String, SegmentReader> m_closedCursors = new HashMap<>();
 
-    //ID of this segment
+    // Index of this segment in the in-memory segment map
     private final Long m_index;
-    private final Instant m_tsCreated;
+
+    // Persistent ID of this segment, based on managing a monotonic counter
+    private final Long m_id;
 
     private int m_numOfEntries = -1;
     private int m_size = -1;
 
     private DBBPool.BBContainer m_tmpHeaderBuf = null;
 
-    public PBDRegularSegment(Long index, Instant ts, File file) {
+    public PBDRegularSegment(Long index, Long  id, File file) {
         super(file);
         m_index = index;
-        m_tsCreated = ts;
+        m_id = id;
         reset();
     }
 
     @Override
-    public long segmentId()
+    public long segmentIndex()
     {
         return m_index;
     }
 
     @Override
-    public Instant created() {
-        return m_tsCreated;
+    public long segmentId() {
+        return m_id;
     }
 
     @Override
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -46,10 +46,10 @@ public class PBDRegularSegment extends PBDSegment {
     private final Map<String, SegmentReader> m_closedCursors = new HashMap<>();
 
     // Index of this segment in the in-memory segment map
-    private final Long m_index;
+    private final long m_index;
 
     // Persistent ID of this segment, based on managing a monotonic counter
-    private final Long m_id;
+    private final long m_id;
 
     private int m_numOfEntries = -1;
     private int m_size = -1;
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -25,7 +25,6 @@ import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.attribute.UserDefinedFileAttributeView;
-import java.time.Instant;
 import java.util.List;
 
 import org.voltcore.utils.DBBPool;
@@ -130,8 +129,8 @@ public abstract class PBDSegment {
         m_file = file;
     }
 
+    abstract long segmentIndex();
     abstract long segmentId();
-    abstract Instant created();
     abstract File file();
 
     abstract void reset();
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -121,7 +121,7 @@ public abstract class PBDSegment {
     protected boolean m_closed = true;
     protected RandomAccessFile m_ras;
     protected FileChannel m_fc;
-    //Avoid unecessary sync with this flag
+    //Avoid unnecessary sync with this flag
     protected boolean m_syncedSinceLastEdit = true;
 
     public PBDSegment(File file)
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -20,7 +20,12 @@ package org.voltdb.utils;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
+import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.attribute.UserDefinedFileAttributeView;
+import java.util.List;
 
 import org.voltcore.utils.DBBPool;
 import org.voltcore.utils.DeferredSerialization;
@@ -99,6 +104,8 @@ public abstract class PBDSegment {
 
     private static final String TRUNCATOR_CURSOR = "__truncator__";
     private static final String SCANNER_CURSOR = "__scanner__";
+    protected static final String IS_FINAL_ATTRIBUTE = "VoltDB.PBDSegment.isFinal";
+
     static final int NO_FLAGS = 0;
     static final int FLAG_COMPRESSED = 1;
 
@@ -213,7 +220,7 @@ public abstract class PBDSegment {
                     if (retval.status == BinaryDeque.TruncatorResponse.Status.FULL_TRUNCATE) {
                         if (reader.readIndex() == 1) {
                             /*
-                             * If truncation is occuring at the first object
+                             * If truncation is occurring at the first object
                              * Whammo! Delete the file.
                              */
                             entriesTruncated = -1;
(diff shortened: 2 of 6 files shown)


src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -284,4 +291,55 @@ ExportSequenceNumberTracker scan(BinaryDeque.BinaryDequeScanner scanner) throws
...
+     *
+     * FIXME: true?
+     * Note that all calls to 'setFinal' are done by the class owning the segment because the segment
+     * itself generally lacks context to decide whether it's final or not.
+     *
+     * @param isFinal   true if segment is set to final, false otherwise
+     * @throws IOException
+     */
+
+    public void setFinal(boolean isFinal) throws IOException {
```
luyangco: I wonder whether it's simpler to call `m_file.setReadOnly()` directly to make it immutable.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -320,6 +320,13 @@ public boolean accept(File pathname) {
                     if (nonce.equals(parsedNonce) && "pbd".equals(extension)) {
                         if (pathname.length() == 4) {
                             //Doesn't have any objects, just the object count
+                            // Ensure it's marked as non-final before deletion
+                            try {
+                                PBDSegment.setFinal(pathname, false);
+                            }
+                            catch (IOException ioe) {
```
luyangco: Code style nitpick: to unify the style of whole class, closing bracket and catch statement should in the same line.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -333,12 +340,21 @@ public boolean accept(File pathname) {
...
                             }
-                            if (m_usageSpecificLog.isDebugEnabled()) {
-                                m_usageSpecificLog.debug("Segment " + qs.file() + " has been recovered");
+
+                            // Any recovered segment that is not final should be checked
+                            // for internal consistency.
+                            if (!qs.isFinal()) {
+                                LOG.warn("Segment " + qs.file()
+                                + " (final: " + qs.isFinal() + "), has been recovered but is not in a final state");
+                            }
```
luyangco: Another code style nitpick: closing bracket and else if statement should in the same line.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -493,8 +535,19 @@ private PBDSegment newSegment(long segmentId, File file) {
      */
     private void closeTailAndOffer(PBDSegment newSegment) throws IOException {
         PBDSegment last = peekLastSegment();
-        if (last != null && !last.isBeingPolled()) {
-            last.close();
+        if (last != null) {
+            if (!last.isBeingPolled()) {
+                last.close();
+            }
```
luyangco: Minor code style issue here too.

src/frontend/org/voltdb/utils/PBDSegment.java
rdykiel: Hell, no. Look at all the places we're opening the segment "for write" in the PBD implementation. I really don't like the idea of playing with permissions. Furthermore I think this is a brittle mechanism that can be too easily broken via "chmod". Also susceptible to be broken by programs (e.g. backup) that would play with those permissions.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
rdykiel: Will do

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
rdykiel: Yes

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
rdykiel: yes

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -377,24 +392,231 @@ public boolean accept(File pathname) {
...
-        } catch (NoSuchElementException e) {}
+        // Find the first and last segment for polling and writing (after); ensure the
+        // writing segment is not final
+        Instant now = Instant.now();
 
+        Map.Entry<Long, PBDSegment> lastEntry = m_segments.lastEntry();
+        Instant tsLast = lastEntry == null ? now.minusMillis(1) : lastEntry.getValue().created();
+        Long writeSegmentIndex = lastEntry == null ? 0L : lastEntry.getKey();
+
+        String fname = getSegmentFileName(now, tsLast);
```
n3world: This could be a problem if you could ever create two segments on the same millis you wouldn't be able to tell them apart since they would have the same now instant. `Instant.now()` is basically the same as `System.currentTimeMillis()`. This is highly unlikely to except that with the concept of forcing a new segment when a catalog change occurs for export. In that case you could start a new segment and immediately have a new catalog causing two segments with the same now. This could also happen with a clock change. Why not just use a monotonic counter which starts at last known `sequenceId + 1`?

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -284,4 +293,55 @@ ExportSequenceNumberTracker scan(BinaryDeque.BinaryDequeScanner scanner) throws
...
+
+    public static boolean isFinal(File file) throws IOException {
+
+        boolean ret = false;
+        UserDefinedFileAttributeView view = getFileAttributeView(file);
+
+        List<String> attrList = view.list();
+        if (attrList.contains(IS_FINAL_ATTRIBUTE)) {
+            ByteBuffer buf = ByteBuffer.allocate(view.size(IS_FINAL_ATTRIBUTE));
+            view.read(IS_FINAL_ATTRIBUTE, buf);
```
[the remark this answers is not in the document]
rdykiel: After discussing this, decided no change as it has little impact.

src/frontend/org/voltdb/utils/PairSequencer.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public void add(Pair<T, T> pair) {
+        m_pairs.addLast(pair);
+    }
+
+    public void addAll(Collection<Pair<T, T>> collection) {
+        m_pairs.addAll(collection);
+    }
+
+    public LinkedList<LinkedList<T>> getSequences() {
```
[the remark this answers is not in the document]
rdykiel: Changed API to Deque<Deque<T>>

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -377,24 +392,231 @@ public boolean accept(File pathname) {
...
+     *
+     * @param file
+     * @return
+     */
+    private Instant getPreviousTimestamp(File file) {
+        Instant ret;
+        String fname = file.getName();
+
+        String rootname = fname.substring(0, fname.lastIndexOf("."));
+        String[] parts = rootname.split("_");
```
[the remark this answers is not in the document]
rdykiel: How would you extract the 2 timestamp strings if not by substring() or by split()?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -377,24 +392,231 @@ public boolean accept(File pathname) {
...
+        PairSequencer<Instant> sequencer = new PairSequencer<>();
+        try {
+            for (File file : m_path.listFiles()) {
+
+                if (file.isDirectory() || !file.isFile() || file.isHidden()) {
+                    continue;
+                }
+
+                String fname = file.getName();
+                if (logMe) LOG.info("XXX parsing file " + fname);
```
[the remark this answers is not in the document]
rdykiel: Removed

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -635,34 +900,46 @@ public synchronized void push(BBContainer objects[]) throws IOException {
...
                 writeSegment.offer(currentSegmentContents.pollFirst(), false);
                 m_numObjects++;
             }
 
-            // Don't close the last one, it'll be used for writes
+            // If this segment is to become the writing segment, don't close and
+            // finalize if
             if (!m_segments.isEmpty()) {
                 writeSegment.close();
+                writeSegment.setFinal(true);
```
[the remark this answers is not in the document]
rdykiel: I want this to be centralized in this file, and not have the PBD implementation touch it.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
rdykiel: Our last conversation convinced me to go to an implementation using a monotonic counter.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -635,34 +900,46 @@ public synchronized void push(BBContainer objects[]) throws IOException {
...
+            String fname = getSegmentFileName(ts, prevTs);
+
             PBDSegment writeSegment =
                 newSegment(
                         nextIndex,
-                        new VoltFile(m_path, m_nonce + "." + nextIndex + ".pbd"));
+                        ts,
+                        new VoltFile(m_path, fname));
             writeSegment.openForWrite(true);
+            writeSegment.setFinal(false);
```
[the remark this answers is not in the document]
rdykiel: No because 2 reasons: 
1- I want the control of the final info to be centralized in the PBD: the segment doesn't have enough context
2- openForWrite is used in cases where we don't want to touch this flag, such as truncate or scan.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -290,109 +290,257 @@ public PersistentBinaryDeque(final String nonce, final File path, VoltLogger log
...
+     * Recover a PBD segment and add it to m_segments
+     *
+     * @param qs
+     * @param deleteEmpty
+     * @throws IOException
+     */
+    private void recoverSegment(PBDSegment qs, boolean deleteEmpty) throws IOException {
+
+        if (deleteEmpty) {
+            if (qs.getNumEntries() == 0) {
```
n3world: Why not just combine the two ifs `if (deleteEmpty && qs.getNumEntries() == 0) {`

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -463,28 +619,42 @@ public synchronized void parseAndTruncate(BinaryDequeTruncator truncator) throws
             PBDSegment segment = m_segments.get(segmentId);
             m_numObjects -= segment.getNumEntries();
             iterator.remove();
-            m_usageSpecificLog.debug("Segment " + segment.file() + " has been closed and deleted by truncator");
-            segment.closeAndTruncate();
+
+            // Ensure the file is not final before closing and truncating
+            segment.setFinal(false);
+            segment.closeAndDelete();
```
n3world: Why not have closeAndDelete call `setFinal(false);` since it has to be done

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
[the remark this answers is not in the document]
n3world: If you write to the segment at all then it isn't final. A scan should just be a read though and not a write.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -45,26 +45,35 @@
...
     @Override
-    public long segmentId()
+    public long segmentIndex()
     {
         return m_index;
     }
 
+    @Override
+    public long segmentId() {
+        return m_id;
```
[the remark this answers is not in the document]
rdykiel: good catch

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -290,109 +290,257 @@ public PersistentBinaryDeque(final String nonce, final File path, VoltLogger log
...
+     * Return a segment file name from m_nonce and current + previous segment ids.
+     *
+     * @see parseFiles for file name structure
+     * @param currentId   current segment id
+     * @param previousId  previous segment id
+     * @return  segment file name
+     */
+    private String getSegmentFileName(long currentId, long previousId) {
+
+        StringBuilder sb = new StringBuilder(m_nonce)
```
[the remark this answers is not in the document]
rdykiel: no change

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
rdykiel: Yes - I wasn't sure there would be just one branch

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
rdykiel: As explained before I want to keep the management of final() in the PBD for now.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -290,109 +292,263 @@ public PersistentBinaryDeque(final String nonce, final File path, VoltLogger log
...
+    }
+
+    /**
+     * @return the next segment id
+     */
+    private synchronized long getNextSegmentId() {
+        long newId = ++m_segmentCounter;
+        if (newId == Long.MAX_VALUE) {
+            // Extremely unlikely
+            VoltDB.crashLocalVoltDB("Unable to allocate segment id for " + m_nonce);
```
luyangco: Crash is bad. I'm fine with resetting m_segmentCounter back to 0 because it's less likely we keep 1e11 number of files than the server run for 10 years then the counter overflows.

rdykiel: The Java Long.MAX_VALUE is 9,223,372,036,854,775,807. Assuming we increment the counter every millisecond, we will maxt he counter in 291,672,107 YEARS. 

Check my math:  9,223,372,036,854,775,807 / (366 x 24 x 3600 x 1000) =  291,672,107?

rdykiel: But I don't object to Yang's change; actually Yang will do this change after I merge.

PR #5913, opened 2019-02-07
ENG-15309: On rejoin send the lowest pending uniqueId

n3world:
During a rejoin SpInitiator.updateReplicasForJoin() calls
SpScheduler.updateReplicas() and then calls
SpScheduler.forwardPendingTaskToRejoinNode(). The first method sends the most
recently generated uniqueId to the replica which the replica might record in
the SpDurabilityListener. Then the next method will send any pending tasks to
the replica whose uniqueId will be sent to SpDurabilityListener. This can cause
the spMaxUniqueId stored in SpDurabilityListener to decrease in value.

Instead of always sending the last generated uniqueId on a rejoin send the
lower value between the last generated uniqueId and the uniqueId of the oldest
pending transaction.


luyangco:
This change looks reasonable to me, I just have one question. Since the pending transactions that being sent to rejoin node are transactions generated between stream snapshot request is queued in site task queue and the request is fired on site thread, their spHandle and uniqueId should be in order, and snapshot save fragment's uniqueId should be the smallest among them. Why rejoin node would report last durable uniqueId going backward in this case?

n3world:
> This change looks reasonable to me, I just have one question. Since the pending transactions that being sent to rejoin node are transactions generated between stream snapshot request is queued in site task queue and the request is fired on site thread, their spHandle and uniqueId should be in order, and snapshot save fragment's uniqueId should be the smallest among them. Why rejoin node would report last durable uniqueId going backward in this case?

The pending transactions are in order but the order of operations is send the Iv2LogFaultMessage with the uniqueId in the most recent pending transactions. The rejoining node records the uniqueId as the latest and then receives the pending transactions. If there are more than 2 pending transactions the first one will have a lower uniqueId than the one sent in the Iv2LogFaultMessage.

n3world:
> I am uncertain about this change. It seems to me that the faultlog message operates outside the transaction path and is only using the last seen uniqueId on the host that detected the fault. This information is only useful when faultlogs are collected for recovery purposes, so I don't understand how this path which operates with different execution ordering rules from the transaction path is able to influence the transaction path?

I don't understand what you mean by it operates outside of the transaction path it is being created and sent as part of a plan fragment of SnapshotSave.

The original commit which added it was 4e49d1665fbb6d22f4cddc7a2020a55e306eba23 ENG-9383: Fix durability notification after rejoin

zeemanhe:
One thing I am not sure about, properly not related to the pull request, is  the potential side effect of transaction id advancement on Iv2FaultLog  (writeIv2ViableReplayEntry). The transaction is not logged and could be lost during repair process.

PR #5915, opened 2019-02-11
ENG-12876: Remove rejoin node from zk on failure

n3world:
There is a time when a the zk rejoin node exists for a rejoining host but the
host is not in zookeeper. If the host failed in this window the rejoin node
would be left in zookeeper not allowing any other hosts to rejoin.
    
Update MeshProber.detract(int) to also take in a zookepeer instance and call
CoreZK.removeRejoinNodeIndicatorForHost to remove the rejoin node from
zookeeper.

### DIFF
--- a/src/frontend/org/voltdb/probe/MeshProber.java
+++ b/src/frontend/org/voltdb/probe/MeshProber.java
@@ -349,7 +349,7 @@ public class MeshProber implements JoinAcceptor {
     }
 
     @Override
-    public void detract(int hostId) {
+    public void detract(ZooKeeper zooKeeper, int hostId) {
         checkArgument(hostId >= 0, "host id %s is not greater or equal to 0", hostId);
         Map<Integer,HostCriteria> expect;
         Map<Integer,HostCriteria> update;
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/probe/MeshProber.java
```diff
@@ -360,6 +360,7 @@ public void detract(int hostId) {
                 .putAll(Maps.filterKeys(expect, not(equalTo(hostId))))
                 .build();
         } while (!m_hostCriteria.compareAndSet(expect, update));
+        CoreZK.removeRejoinNodeIndicatorForHost(zooKeeper, hostId);
```
JadenW: Just curious, the culprit is the race of rejoin_node_blocker creation and the existence of rejoin node in the zookeeper, right?
So the clock skew (described in the ticket) is just "one" cause of the rejoin node failure, but nothing special about it?

n3world: It isn't a race. The rejoin node is created before the host joins the cluster so anything that causes an error between the rejoin node being created and the new host adding itself to zookeeper would have caused this issue. It is just that clock skew is the easiest to reproduce.

PR #5917, opened 2019-02-11
Eng 15035 junit flaky2

johnTcrawford:
ENG-15035-junit-flaky2, v2 after review commments:
Added FlakyTestRule (including @Flaky annotation) and
FlakyTestStandardRunner, to allow marking of Flaky tests as such;
included modifications of TestFuzzMeshArbiter to show how to use @flaky;
and minor changes to MultiConfigSuiteBuilder (including debug print) and
build.xml. Also modified test-tools.sh: added a tt-set-test-args function,
to set a TEST_ARGS environment variable, based on a value passed to it
(which will probably be called TEST_OPTS), for passing @Flaky-related
command-line args to the JUnit tests (very similar to the way the
tt-set-build-args function works for VoltDB builds args).

### DIFF
--- a/tests/frontend/org/voltdb/FlakyTestRule.java
+++ b/tests/frontend/org/voltdb/FlakyTestRule.java
@@ -1,0 +1,238 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2019 VoltDB Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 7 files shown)


tests/frontend/org/voltdb/FlakyTestRule.java
```diff
@@ -0,0 +1,238 @@
...
+            System.out.println("DEBUG:   base: "+base);
+            System.out.println("DEBUG:   desc: "+desc);
+        }
+
+        Statement result = base;
+        Flaky flakyAnnotation = desc.getAnnotation(Flaky.class);
+        if (flakyAnnotation != null) {
+            FlakyTestRunner runCondition = (new FlakyTestRunnerCreator(desc, flakyAnnotation)).create();
+            if (!runCondition.runFlakyTest(flakyAnnotation.isFlaky(), flakyAnnotation.description())) {
+                result = new IgnoreThisTest();
```
kaikaizi: So when a test is marked flaky, and we don't run flaky test, the marked test will be skipped, rather than running the test but ignore the test result. 
We might need some comment here to state this again in here.

tests/frontend/org/voltdb/FlakyTestRule.java
```diff
@@ -0,0 +1,238 @@
...
+public class FlakyTestRule implements TestRule {
+    // Print debug statements only if -Drun.flaky.tests.debug=TRUE
+    private final boolean DEBUG = FlakyTestStandardRunner.debug();
+
+    /**
+     * The @Flaky annotation is intended to be used for JUnit tests that do not
+     * pass reliably, failing either intermittently or consistently; such tests
+     * may or may not be skipped, depending on several factors:<br>
+     *     o The value of the system property run.flaky.tests (set on the command
+     *       line via '-Drun.flaky.tests=...'), when running the tests.<br>
```
kaikaizi: s/.../[true|false|none]/
and explain the concrete usage via Jenkins.
We need a wiki page for Flaky annotation, and it's best to also link to that page here.

PR #5920, opened 2019-02-12
ENG-15084:

luyangco:
1. Added a new mechanism for tracking streams that have a partial buffer to flush.
2. Refactored the StreamBlock to specialize it for DR and Export specific behavior.
3. Added commit semantics to the Export stream path so that pending buffers can be flushed immediately after the commit. The commit uses the release interest to commit once at the end of the release of the undo log.
4. Dedupped the release interest so that the same table is not repeatedly inserted into the release interest and compaction is not called once for each deleted row. This would imply that multiple rows injected into a stream by the same transaction will only generate a single commit call for each StreamTable.


src/ee/common/StreamBlock.h
```diff
@@ -110,186 +83,247 @@ namespace voltdb
...
+        const size_t m_uso;            // universal stream offset of m_offset 0.
+
+        StreamBlockType m_type;
+
+        int64_t m_lastCommittedSpHandle; // for record last CommittedSpHandle for Sp Txn in this block
+        int64_t m_lastSpUniqueId;
+    };
+
+
+    class ExportStreamBlock : public StreamBlock {
```
rdykiel: Why isnt this class and the DR one in their own implementation file?

src/ee/common/StreamBlock.h
```diff
@@ -110,186 +83,247 @@ namespace voltdb
...
+    public:
+        ExportStreamBlock(char* data, size_t headerSize, size_t capacity, size_t uso) :
+            StreamBlock(data, headerSize, capacity, uso),
+            m_rowCount(0),
+            m_needsSchema(true),
+            m_startSequenceNumber(0)
+        {}
+
+        ExportStreamBlock(ExportStreamBlock* other) :
```
rdykiel: Why don't we have copy constructors based on the other's reference? It seems pervasive so therre must be a good reason that I need to learn. Not having the copy constructor means bitwise copy, right?

src/ee/execution/VoltDBEngine.h
```diff
@@ -540,6 +543,14 @@ class __attribute__((visibility("default"))) VoltDBEngine {
 
         void setViewsEnabled(const std::string& viewNames, bool value);
 
+        virtual ExportTupleStream** getNewestExportStreamWithPendingRowsForAssignment() {
```
luyangco: Or simply make `m_newestExportStreamWithPendingRows ` and `m_oldestExportStreamWithPendingRows ` public.

luyangco: Another advantage of adding a setter is that `appendToList()` in ExportTupleStream can be replaced by the setter.

src/ee/storage/ExportTupleStream.cpp
```diff
@@ -181,6 +190,109 @@ size_t ExportTupleStream::appendTuple(int64_t lastCommittedSpHandle,
...
+        // Advance the tip to the new transaction.
+        m_committedSpHandle = m_openSpHandle;
+        if (m_currBlock->startSequenceNumber() > m_committedSequenceNumber) {
+            // Started a new block so reset the flush timeout
+            m_lastFlush = UniqueId::ts(m_committedUniqueId);
+            removeFromFlushList(engine, true);
+        }
+        m_committedSequenceNumber = m_nextSequenceNumber-1;
+
+        pushPendingBlocks();
```
luyangco: I think `commit` shouldn't push the pending blocks, because we already have 2 places to push the pending buffer:
1) when new buffer is created, push the old one,
2) when `tick` or `quiesce` is called.

src/ee/execution/VoltDBEngine.h
wweiss-voltdb: This implies that we have to look up the engine from the thread local variable. Passing the pointers avoids the lookup and simplifies testing.

src/ee/storage/ExportTupleStream.cpp
wweiss-voltdb: With commit semantics in place, the only place to process pending buffers is the commit. In all other paths, the m_pendingBlocks is either empty or there is an MP transaction in progress that has not been committed yet. Tick should ever only need to flush the current buffer and only if tick is not in the middle of an MP txn. Quiesce can also only have one buffer in progress (not in the pending queue), but since Quiesce is an MP txn, there are never MP txns in progress so the current buffer can always be flushed (unless it is empty).

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -2172,13 +2180,37 @@ void VoltDBEngine::tick(int64_t timeInMillis, int64_t lastCommittedSpHandle) {
...
+                m_oldestExportStreamWithPendingRows->appendToList(&oldestUnflushed, &newestUnflushed);
+            }
+            m_oldestExportStreamWithPendingRows = nextStreamToFlush;
+        }
+        else {
+            // We tried to flush a stream that has not yet hit it's flush timer
+            break;
+        }
+    }
+    if (nextStreamToFlush && newestUnflushed) {
```
n3world: If newestUnflushed is not NULL but nextStreamToFlush is NULL don't you need to update m_oldestExportStreamWithPendingRows and m_newestExportStreamWithPendingRows appropriately.

It might be a bit cleaner to make a utility class which is a ExportTupleStreamList that handles all of the list maintenance and pointers and iteration.

PR #5921, opened 2019-02-12
ENG-12733: Support multiple schema and procedure files

n3world:
Add support for providing multiple schema and jar files during the voltdb initialization

### DIFF
--- a/src/frontend/org/voltdb/VoltDB.java
+++ b/src/frontend/org/voltdb/VoltDB.java
@@ -331,7 +331,7 @@ public class VoltDB {
         public List<File> m_userSchemas = null;
 
         /** location of user supplied classes and resources jar file */
-        public File m_stagedClassesPath = null;
+        public List<File> m_stagedClassesPaths = null;
 
         /** Best effort to recover previous partition layout*/
         public boolean m_restorePlacement = !Boolean.valueOf(System.getenv("DISABLE_PLACEMENT_RESTORE") == null ?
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -551,16 +551,19 @@ public boolean compileFromSchemaAndClasses(
             return false;
         }
 
-        InMemoryJarfile inMemoryUserJar = null;
+        InMemoryJarfile inMemoryUserJar = new InMemoryJarfile();
         ClassLoader originalClassLoader = m_classLoader;
         try {
-            if (classesJarPath != null && classesJarPath.exists()) {
+            m_classLoader = inMemoryUserJar.getLoader();
```
JadenW: Why have to switch classLoader?

src/frontend/org/voltdb/VoltDB.java
```diff
@@ -718,18 +718,25 @@ else if (arg.equalsIgnoreCase("buildstringoverride")) {
...
-                    }
-                    if (!m_stagedClassesPath.canRead()) {
-                        System.err.println("FATAL: Supplied classes jar file " + m_stagedClassesPath + " can't be read.");
-                        referToDocAndExit();
-                    }
-                    if (!m_stagedClassesPath.isFile()) {
-                        System.err.println("FATAL: Supplied classes jar file " + m_stagedClassesPath + " is not an ordinary file.");
-                        referToDocAndExit();
+                    for (String jarPath : Splitter.on(",").trimResults().omitEmptyStrings().split(args[++i])) {
+                        File staggedJar = new File(jarPath);
```
manjujames: Nit: Should be stagedJar instead of staggedJar.

manjujames:
Looks fine to me. I assume we have to test voltadmin tool manually?

Only question I have is about the user interface - Should it be possible for the user to specify a directory containing the schema/jar files? Might be a question for Ruth and Andrew. Could even be a separate ticket, if this change is to satisfy a specific customer request.

src/frontend/org/voltdb/compiler/VoltCompiler.java
n3world: The change was always done before. It appears that the class loader is used in methods called by compileInternal. Not sure of the reason why it is needed.

n3world:
> Looks fine to me. I assume we have to test voltadmin tool manually?
> 
> Only question I have is about the user interface - Should it be possible for the user to specify a directory containing the schema/jar files? Might be a question for Ruth and Andrew. Could even be a separate ticket, if this change is to satisfy a specific customer request.

voltdb is partially tested by the voltdb-verbs-test which did find an issue with one of my earlier revisions. voltdb just blindly passes the value along to the java process so I would be surprised if there was an issue there.

Adding directory support wouldn't be hard but the question would be what do you take from the directory and is it a recursive search. For classes it could be all `.jar` files you find but for schema what extensions do you look for `.ddl` or `.sql`?

manjujames:
> > Looks fine to me. I assume we have to test voltadmin tool manually?
> > Only question I have is about the user interface - Should it be possible for the user to specify a directory containing the schema/jar files? Might be a question for Ruth and Andrew. Could even be a separate ticket, if this change is to satisfy a specific customer request.
> 
> voltdb is partially tested by the voltdb-verbs-test which did find an issue with one of my earlier revisions. voltdb just blindly passes the value along to the java process so I would be surprised if there was an issue there.
> 
> Adding directory support wouldn't be hard but the question would be what do you take from the directory and is it a recursive search. For classes it could be all `.jar` files you find but for schema what extensions do you look for `.ddl` or `.sql`?

Yes, exactly. Which is why depending on how quickly Ruth needs it, it might be something to discuss with Andrew.

PR #5923, opened 2019-02-13
ENG-15370 modernize kafka export verifiers

jwiebema1:
ENG-15370 and ENG-15175 replace genqa.ExportKafkaOnServerVerifier.java with testutils.KafkaClientVerifier.java a generic  KafkaConsumer API compliant verifier for export tests.


tests/test_apps/genqa/src/genqa/KafkaClientVerifier.java
```diff
@@ -0,0 +1,416 @@
...
+                                + row[7] + "," + row[8] + "," + row[9] + " foundsize:" + foundRowIds.size());
+                    }
+
+                    if (m_partitionFieldNum != null) {
+                        Integer partition = Integer.parseInt(row[m_partitionFieldNum].trim());
+
+                        if (TxnEgo.getPartitionId(rowTxnId) != partition) {
+                            System.err.println("ERROR mismatched exported partition for txid " + rowTxnId
+                                    + ", tx says it belongs to " + TxnEgo.getPartitionId(rowTxnId)
+                                    + ", while export record says " + partition);
```
prosegay: on errors, like this, how are we sure that the app exists with non-zero return code?

tests/test_apps/genqa/src/genqa/KafkaClientVerifier.java
```diff
@@ -0,0 +1,416 @@
...
+            Thread.sleep(5000);
+            if (cnt != consumedRows.get()) {
+                long delta = consumedRows.get() - cnt;
+                wtime = System.currentTimeMillis();
+                System.out.println("Train is still running, got " + delta + " more records");
+                continue;
+            }
+            if ((System.currentTimeMillis() - wtime) > 60000) {
+                System.out.println("Waited long enough looks like train has stopped.");
+                break;
```
prosegay: error exit here? or flagged from here?

tests/test_apps/genqa/src/genqa/KafkaClientVerifier.java
```diff
@@ -0,0 +1,416 @@
...
+        Arrays.sort(sortedIds);
+        List<Long> ids = Arrays.asList(sortedIds);
+        long missingCnt = 0;
+        long duplicateCnt = 0;
+        long lastId = 0;
+        int print_cnt = 0;
+        for (Long id : ids) {
+            if (lastId > id) {
+                // double check the sequence.
+                System.err.println("out of sequence id: " + id + " > " + lastId);
```
prosegay: error exit?

tests/test_apps/genqa/src/genqa/KafkaClientVerifier.java
```diff
@@ -0,0 +1,416 @@
...
+        CountDownLatch consumersLatch = new CountDownLatch(consumerGroup.size());
+
+        for (KafkaConsumer<String, String> consumer : consumerGroup) {
+
+            consumer.subscribe(Arrays.asList(topic));
+            System.out.println("Creating consumer for " + topic);
+            TopicReader reader = new TopicReader(consumer, consumersLatch, uniqueIndexFieldNum, sequenceFieldNum,
+                    partitionFieldNum, consumerTimeoutSecs );
+            executor.execute(reader);
+        }
```
prosegay: the topic reader runs on a thread so have you tested: 1) unhandled exceptions thrown from the thread causes the app to exit with a clear message and non-zero return code? 2) application errors (verification failure, for example) detected in a thread are reflected in a non-zero return code from the app when it exits?

jwiebema1: That is why I use executor.execute instead of submit so the errors are bubbled up.

tests/test_apps/genqa/src/genqa/KafkaClientVerifier.java
jwiebema1: It's possible that the stream is actually complete and valid, so we check all of the stats later and will fail if the stats don't match up.

tests/test_apps/genqa/src/genqa/KafkaClientVerifier.java
jwiebema1: yup, I need to add a flag to propopage the failure.

PR #5926, opened 2019-02-14
ENG-13631: Improve error handling while aggregating results 

n3world:
When an error occurs log a detailed message and return an error response to the client. Also, track the timer future and cancel the timer when the operation completes.

### DIFF
--- a/src/frontend/org/voltdb/sysprocs/SnapshotRestore.java
+++ b/src/frontend/org/voltdb/sysprocs/SnapshotRestore.java
@@ -781,7 +781,7 @@ public class SnapshotRestore extends VoltSystemProcedure {
                             resultSet.parseRestoreResultRow(vt);
                         }
                         if (result == null) {
-                            result = new VoltTable(VoltTableUtil.extractTableSchema(vt));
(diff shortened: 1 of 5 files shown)


src/frontend/org/voltdb/sysprocs/SnapshotRestore.java
```diff
@@ -781,7 +781,7 @@ else if (fragmentId == SysProcFragmentId.PF_restoreReceiveResultTables) {
                             resultSet.parseRestoreResultRow(vt);
                         }
                         if (result == null) {
-                            result = new VoltTable(VoltTableUtil.extractTableSchema(vt));
+                            result = new VoltTable(vt.getTableSchema());
```
manjujames: Was this just a cleanup or did this have issues?

n3world: Clean up there were two implementations for how to get the schema of a volt table one in VoltTable and one in VoltTableUtil. They were identical so I removed the one from VoltTableUtil

PR #5930, opened 2019-02-15
ride through errors on finalization operations

rdykiel:
See also https://issues.voltdb.com/browse/ENG-15429 which defines an improvement that will be implemented later.

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -323,7 +323,9 @@ public abstract class PBDSegment {
 
         try {
             UserDefinedFileAttributeView view = getFileAttributeView(file);
-            view.write(IS_FINAL_ATTRIBUTE, Charset.defaultCharset().encode(new Boolean(isFinal).toString()));
+            if (view != null) {
+                view.write(IS_FINAL_ATTRIBUTE, Charset.defaultCharset().encode(new Boolean(isFinal).toString()));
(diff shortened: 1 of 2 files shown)


luyangco:
Looks good to me.

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -305,36 +312,49 @@ ExportSequenceNumberTracker scan(BinaryDeque.BinaryDequeScanner scanner) throws
...
 
-    public static void setFinal(File file, boolean isFinal) throws IOException {
+    public static void setFinal(File file, boolean isFinal) {
 
-        UserDefinedFileAttributeView view = getFileAttributeView(file);
-        view.write(IS_FINAL_ATTRIBUTE, Charset.defaultCharset().encode(new Boolean(isFinal).toString()));
+        try {
+            UserDefinedFileAttributeView view = getFileAttributeView(file);
+            view.write(IS_FINAL_ATTRIBUTE, Charset.defaultCharset().encode(new Boolean(isFinal).toString()));
+        } catch (Exception e) {
```
n3world: I would prefer if not all exceptions were swallowed but instead maybe just IOExceptions and it still might be nice to log those at debug in case somebody ever wonders why this isn't working.

rdykiel: Right - restricting handling to IOException

PR #5931, opened 2019-02-15
ENG-15396:

wweiss-voltdb:
A JUnit test failure for Export identified an issue with the Quiesce command, which did not ensure that flushed export buffers had been delivered to the Export persistence layer.

This did not fix the JUnit test, which found additional issues with the synchronization of socket connection initialization as well as synchronization of the writes but this issue was only in the socket exporter used for testing.


rdykiel:
Approved

PR #5935, opened 2019-02-19
ENG-15408 DDL support for alter stream and TTL export

zeemanhe:
Support add, drop and modify stream column. Add new DDL syntax to mark a table as TTL export.


src/catgen/in/javasrc/CatalogDiffEngine.java
```diff
@@ -927,8 +922,18 @@ protected String checkModifyWhitelist(final CatalogType suspect,
...
+            suspect instanceof PlanFragment) {
+            return null;
+        }
+
+        if (suspect instanceof TimeToLive) {
+            TimeToLive current = (TimeToLive)suspect;
+            if (prevType != null) {
+                TimeToLive previous = (TimeToLive)prevType;
+                if (previous.getMigrationtarget() == null && current.getMigrationtarget() != null) {
+                    m_requiresNewExportGeneration= true;
```
rdykiel: Why only detecting one-way change? Why not  handling
previous.getMigrationtarget() != null && current.getMigrationtarget() == null

.. to detect we're removing the nibble export clause?

src/catgen/spec.txt
```diff
@@ -166,6 +166,9 @@ begin Table                                  "A table (relation) in the database
   bool isDRed                                "Is this table DRed?"
   Statement* tuplelimitDeleteStmt            "Delete statement to execute if tuple limit will be exceeded"
   TimeToLive* timeToLive                     "Time to live"
+  bool isStream                              "Is the table a steam?"
```
rdykiel: BEWARE: Walter is changing this as well! Coordinate with him....

src/catgen/spec.txt
```diff
@@ -174,6 +177,7 @@ begin TimeToLive javaonly             "Time to Live (TTL), limits the life span
   Column? ttlColumn                   "A column used to determine the duration of data"
   int batchSize                       "Number of tuples to be deleted in a batch. default: 1000"
   int maxFrequency                    "Maximal number of times of deleting batchSize rows per second.default: 1"
+  string migrationTarget              "The target where the data are migrated to"
```
rdykiel: Is this field only used in nibble export? Could it be used for stream exports? 
Why do we need to know the target name?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1352,6 +1355,11 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
             ttl.setBatchsize(ttlValue);
             ttlValue = Integer.parseInt(ttlNode.attributes.get("maxFrequency"));
             ttl.setMaxfrequency(ttlValue);
+            final String migrationTarget = ttlNode.attributes.get("migrartionTarget");
```
rdykiel: TYPO on atrribute name: did you mean "migrationTarget"?

src/catgen/in/javasrc/CatalogDiffEngine.java
zeemanhe: The intention is to allow nibble export to be added but not changed.  Nibble export can be dropped.

src/catgen/spec.txt
rdykiel: When we process "CREATE STREAM foo EXPORT TO TARGET bar" we don't need the 'migrationTarget' property. Why do we need this for "CREATE TABLE foo USING TTL MIGRATE..."? I'd like the handling of all the exporting tables/streams to be common.

src/catgen/spec.txt
```diff
@@ -166,6 +166,7 @@ begin Table                                  "A table (relation) in the database
   bool isDRed                                "Is this table DRed?"
   Statement* tuplelimitDeleteStmt            "Delete statement to execute if tuple limit will be exceeded"
   TimeToLive* timeToLive                     "Time to live"
+  int streamType                             "Types: no stream(persistent), view-only stream, stream with target, persistent table for export and migration"
```
rdykiel: It should not be called 'streamType' because it also contains the standard persistent table. Why not call this 'tableType'?

src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1352,6 +1385,11 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
             ttl.setBatchsize(ttlValue);
             ttlValue = Integer.parseInt(ttlNode.attributes.get("maxFrequency"));
             ttl.setMaxfrequency(ttlValue);
+            final String migrationTarget = ttlNode.attributes.get("migrationTarget");
+            if (migrationTarget != null) {
+                ttl.setMigrationtarget(migrationTarget);
+                table.setStreamtype(VoltTypeUtil.TABLE_STREAM_EXTENSION.MIGRATE_TABLE.get());
```
rdykiel: How do we set the streamType to the 'standard persistent table' value? Does this mean that 'streamType' would be null to indicate a standard persistent table? I would prefer streamType to be changed to tableType and have a value for the standard persistent table.

src/frontend/org/voltdb/compiler/VoltDDLElementTracker.java
```diff
@@ -47,6 +47,8 @@
             new HashMap<>();
     // map from export group name to a sorted set of table names in that group
     final NavigableMap<String, NavigableSet<String>> m_exportsByTargetName = new TreeMap<>();
+    final Map<String, String> m_persistentTableTargetMap = new HashMap<>();
```
rdykiel: I am not convinced we need to have a different way to handle persistent tables that export, and streams that export. We should build the connectors in the same way as if they were all streams and whoever wants to figure out what type the table is should just look up the catalog entry for the table.

zeemanhe: The export tables have special meanings for materialized view. The separate tracking is intended not to mess around with it.

src/frontend/org/voltdb/compiler/DDLCompiler.java
zeemanhe: fixed

src/catgen/spec.txt
zeemanhe: will change it

src/catgen/spec.txt
zeemanhe: The TTL with migrationTarget allows DDL compiler and catalog diff engine to identify modifications directly from the statement rather than going through the connectors in catalog. Compiler or diff engine does not need to know any stream modifications since streams are not alterable in previous releases.

src/frontend/org/voltdb/utils/VoltTypeUtil.java
```diff
@@ -358,4 +358,24 @@ public static VoltType getNumericLiteralType(VoltType vt, String value) {
         }
         return vt;
     }
+
+    public static enum TABLE_TYPE {
+        PERSISTENT(0),              // Regular PersistentTable
+        STREAM_VIEW_ONLY(1),        // StreamTable without ExportTupleStream (Views only)
+        STREAM(2),                  // StreamTable with ExportTupleStream
+        PERSISTENT_MIGRATE(3),      // PersistentTable with associated Stream for migrating DELETES
+        PERSISTENT_EXPORT(4);       // PersistentTable with associated Stream for leaking INSERTS
```
rdykiel: "leaking"? Did you mean "exporting"? And my IW prototype would export on INSERT and UPDATE.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -1060,40 +1057,70 @@ private Statement readTimeToLive(Table table, boolean alter) {
...
     private Statement createTimeToLive(Table table, boolean alter,int value, String unit, String column,
-            int batchSize, int maxFrequency) {
+            int batchSize, int maxFrequency, String migrationTargetName) {
         if (!alter) {
-            table.addTTL(value, unit, column, batchSize, maxFrequency);
+            table.addTTL(value, unit, column, batchSize, maxFrequency, migrationTargetName);
+        } else {
+            // The migration target can not be altered
+            if (table.getTTL() != null && !StringUtil.isEmpty(table.getTTL().migrationTarget) && !table.getTTL().migrationTarget.equalsIgnoreCase(migrationTargetName)){
+                throw unexpectedToken("The migration target can not be altered.");
```
rdykiel: "cannot"

PR #5937, opened 2019-02-19
Eng 14565 integration

luyangco:
This pull request contains two main changes.
1) move the export schema from per-entry to per-segment, the first entry of segment contains schema. Catalog update involves export streams will cause export to close current segment and start a new segment with new schema. Export client will update the decoder when new schema is detected.
2) add CRC field to segment header and entry header, check the CRC of PBD files when cluster is recovered or restored, truncate the file to last safe point if file corruption is detected.

New PBD segment layout:

> Segment Header (Each PBD file is a segment)
> --
> segment header CRC (4 bytes)
> total number of entries (4 bytes)
> total size in bytes (4 bytes)
>
> Export Segment Header
> --
>  export version (1 byte)
>  generation ID (8 bytes)
>  length of schema ( 4 bytes)
>  schema (variable length)
>
> Entry Header (Each segment contains one or more entries)
> --
> entry CRC (4 bytes)
> length of entry (4 bytes)
> flags ( COMPRESSED or NONE, 4 bytes)
>
> Export Entry Header
> --
> export sequence number (8 bytes)
> row count (4 bytes)
> unique ID (8 bytes)
>
> Per Row Data (for EXPORT)
> --
> // each row repeats
> { 
>     row length (4 bytes)
>     partition column index (4 bytes)
>     column count (4 bytes)
>     null array length (4 bytes)
>     null array (variable length)
>     export meta column (txnId + timestamp + seqNo + partitionId + siteId + exportOperation)
>     row data (variable length)
> }

### DIFF
--- a/src/frontend/org/voltdb/export/StreamBlockQueue.java
+++ b/src/frontend/org/voltdb/export/StreamBlockQueue.java
@@ -24,11 +24,14 @@ import java.util.Iterator;
 
 import org.voltcore.logging.VoltLogger;
 import org.voltcore.utils.DBBPool.BBContainer;
+import org.voltcore.utils.DeferredSerialization;
+import org.voltdb.VoltDB;
+import org.voltdb.export.ExportDataSource.StreamTableSchemaSerializer;
 import org.voltdb.utils.BinaryDeque;
-import org.voltdb.utils.BinaryDeque.BinaryDequeReader;
 import org.voltdb.utils.BinaryDeque.BinaryDequeScanner;
 import org.voltdb.utils.BinaryDeque.BinaryDequeTruncator;
 import org.voltdb.utils.BinaryDeque.TruncatorResponse;
+import org.voltdb.utils.BinaryDequeReader;
 import org.voltdb.utils.PersistentBinaryDeque;
 import org.voltdb.utils.PersistentBinaryDeque.ByteBufferTruncatorResponse;
 import org.voltdb.utils.VoltFile;
@@ -41,12 +44,15 @@ import org.voltdb.utils.VoltFile;
  * portion of the queue.
  *
  * Export PBD buffer layout:
- *    --- Buffer Header   ---
- *    seqNo(8) + tupleCount(4) + uniqueId(8) + exportVersion(1) + generationId(8) + schemaLen(4) + tupleSchema(var length)
- *    {
- *          ---Inside schema---
- *          tableNameLength(4) + tableName(var length) + colNameLength(4) + colName(var length) + colType(1) + colLength(4) + ...
- *    }
+ *    -- Segment Header ---
+ *    crc(8) + numberOfEntries(4) + totalBytes(4)
+ *    -- Export Segment Header ---
+ *    exportVersion(1) + generationId(8) + schemaLen(4) + tupleSchema(var length) +
+ *    tableNameLength(4) + tableName(var length) + colNameLength(4) + colName(var length) + colType(1) + colLength(4) + ...
+ *    --- Common Entry Header   ---
+ *    crc(8) + length(4) + flags(4)
+ *    --- Export Entry Header   ---
+ *    seqNo(8) + tupleCount(4) + uniqueId(8)
  *    --- Row Header      ---
  *    rowLength(4) + partitionColumnIndex(4) + columnCount(4, includes metadata columns) +
  *    nullArrayLength(4) + nullArray(var length)
@@ -78,10 +84,15 @@ public class StreamBlockQueue {
 
     private final String m_nonce;
     private final String m_path;
+    private final String m_streamName;
     private BinaryDequeReader m_reader;
 
-    public StreamBlockQueue(String path, String nonce) throws java.io.IOException {
-        m_persistentDeque = new PersistentBinaryDeque( nonce, new VoltFile(path), exportLog);
+    public StreamBlockQueue(String path, String nonce, String streamName)
+            throws java.io.IOException {
+        m_streamName = streamName;
+        StreamTableSchemaSerializer ds = new StreamTableSchemaSerializer(
+                VoltDB.instance().getCatalogContext(), m_streamName);
+        m_persistentDeque = new PersistentBinaryDeque( nonce, ds, new VoltFile(path), exportLog);
         m_path = path;
         m_nonce = nonce;
         m_reader = m_persistentDeque.openForRead(m_nonce);
@@ -108,8 +119,14 @@ public class StreamBlockQueue {
      */
     private StreamBlock pollPersistentDeque(boolean actuallyPoll) {
         BBContainer cont = null;
+        BBContainer schemaCont = null;
         try {
-            cont = m_reader.poll(PersistentBinaryDeque.UNSAFE_CONTAINER_FACTORY);
+            // Start to read a new segment
+            if (m_reader.isStartOfSegment()) {
+                schemaCont = m_reader.getSchema(PersistentBinaryDeque.UNSAFE_CONTAINER_FACTORY, false);
+            }
+            cont = m_reader.poll(PersistentBinaryDeque.UNSAFE_CONTAINER_FACTORY, false);
+
         } catch (IOException e) {
             exportLog.error(e);
         }
@@ -120,12 +137,13 @@ public class StreamBlockQueue {
             cont.b().order(ByteOrder.LITTLE_ENDIAN);
             //If the container is not null, unpack it.
             final BBContainer fcont = cont;
-            long seqNo = cont.b().getLong(0);
-            int tupleCount = cont.b().getInt(8);
-            long uniqueId = cont.b().getLong(12);
+            long seqNo = cont.b().getLong(StreamBlock.SEQUENCE_NUMBER_OFFSET);
+            int tupleCount = cont.b().getInt(StreamBlock.ROW_NUMBER_OFFSET);
+            long uniqueId = cont.b().getLong(StreamBlock.UNIQUE_ID_OFFSET);
             //Pass the stream block a subset of the bytes, provide
             //a container that discards the original returned by the persistent deque
             StreamBlock block = new StreamBlock( fcont,
+                schemaCont,
                 seqNo,
                 tupleCount,
                 uniqueId,
@@ -224,8 +242,8 @@ public class StreamBlockQueue {
     /*
      * Only allow two blocks in memory, put the rest in the persistent deque
      */
-    public void offer(StreamBlock streamBlock) throws IOException {
-        m_persistentDeque.offer(streamBlock.asBBContainer(), !DISABLE_COMPRESSION);
+    public void offer(StreamBlock streamBlock, DeferredSerialization ds, boolean createNewFile) throws IOException {
+        m_persistentDeque.offer(streamBlock.asBBContainer(), ds, !DISABLE_COMPRESSION, createNewFile);
         long unreleasedSeqNo = streamBlock.unreleasedSequenceNumber();
         if (m_memoryDeque.size() < 2) {
             StreamBlock fromPBD = pollPersistentDeque(false);
@@ -296,11 +314,6 @@ public class StreamBlockQueue {
                     return null;
                 }
                 b.getLong(); // uniqueId
-                byte version = b.get(); // export version
-                assert(version == EXPORT_BUFFER_VERSION);
-                b.getLong(); // generation id
-                int firstRowStart = b.getInt() + b.position();
-                b.position(firstRowStart);
 
                 // Partial truncation
                 int offset = 0;
@@ -331,7 +344,9 @@ public class StreamBlockQueue {
 
         // close reopen reader
         m_persistentDeque.close();
-        m_persistentDeque = new PersistentBinaryDeque(m_nonce, new VoltFile(m_path), exportLog);
+        StreamTableSchemaSerializer ds = new StreamTableSchemaSerializer(
+                VoltDB.instance().getCatalogContext(), m_streamName);
+        m_persistentDeque = new PersistentBinaryDeque(m_nonce, ds, new VoltFile(m_path), exportLog);
         m_reader = m_persistentDeque.openForRead(m_nonce);
         // temporary debug stmt
         exportLog.info("After truncate, PBD size is " + (m_reader.sizeInBytes() - (8 * m_reader.getNumObjects())));
@@ -354,7 +369,6 @@ public class StreamBlockQueue {
         });
     }
 
-
     @Override
     public void finalize() {
         try {
@@ -371,5 +385,4 @@ public class StreamBlockQueue {
             }
         }
     }
-
 }
--- a/src/frontend/org/voltdb/export/StreamBlockQueue.java
+++ b/src/frontend/org/voltdb/export/StreamBlockQueue.java
@@ -127,7 +127,7 @@ public class StreamBlockQueue {
             }
             cont = m_reader.poll(PersistentBinaryDeque.UNSAFE_CONTAINER_FACTORY, false);
         } catch (IOException e) {
-            exportLog.error(e);
+            exportLog.error("Failed to poll from persistent binary deque:" + e);
         }
 
         if (cont == null) {
--- a/src/frontend/org/voltdb/export/StreamBlockQueue.java
+++ b/src/frontend/org/voltdb/export/StreamBlockQueue.java
@@ -176,7 +176,7 @@ public class StreamBlockQueue {
         try {
             schemaCont = m_reader.getSchema(segmentIndex, true, false);
         } catch (IOException e) {
-            exportLog.error(e);
+            exportLog.error("Failed to poll schema: " + e);
         }
         return schemaCont;
 
--- a/src/frontend/org/voltdb/export/StreamBlockQueue.java
+++ b/src/frontend/org/voltdb/export/StreamBlockQueue.java
@@ -123,10 +123,9 @@ public class StreamBlockQueue {
         try {
             // Start to read a new segment
             if (m_reader.isStartOfSegment()) {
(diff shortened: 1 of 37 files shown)


src/frontend/org/voltdb/export/StreamBlockQueue.java
```diff
@@ -280,9 +285,17 @@ public void truncateToSequenceNumber(final long truncationSeqNo) throws IOExcept
         m_persistentDeque.parseAndTruncate(new BinaryDequeTruncator() {
 
             @Override
-            public TruncatorResponse parse(BBContainer bbc) {
+            public TruncatorResponse parse(BBContainer bbc, boolean firstObject) {
```
rdykiel: firstObject --> firstBuffer?

src/frontend/org/voltdb/export/processors/GuestProcessor.java
```diff
@@ -346,21 +347,28 @@ public void run() {
...
-                                    buf.position(buf.position() + schemaSize);
+                                if (cont.hasSchema()) {
+                                    byte version = buf.get();
+                                    assert(version == StreamBlockQueue.EXPORT_BUFFER_VERSION);
+                                    // update the global generation id of guest processor
+                                    m_genId = buf.getLong();
+                                    int schemaSize = buf.getInt();
+                                    ExportRow previousRow = edb.getPreviousRow();
+                                    // update the decoder if current generation is different than previous row
+                                    if (previousRow == null || previousRow.generation != m_genId) {
```
rdykiel: Can we get rid of the previousRow management and replace it by some logic allowing accessing schema whenever needed?

src/frontend/org/voltdb/utils/BinaryDeque.java
```diff
@@ -143,6 +146,13 @@
          * @throws IOException
          */
         public boolean isEmpty() throws IOException;
+
+        /**
+         * Is the object this reader going to read the first object of segment?
+         * @return true if the object this reader going to read is the first object of segment
+         * throws IOException
+         */
+        public boolean isReadFirstObjectOfSegment() throws IOException;
```
rdykiel: isStartOfSegment?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -282,44 +294,48 @@ public boolean hasAllFinishedReading() throws IOException {
...
             }
 
+            // Write entry
+            destBuf.b().flip();
             while (destBuf.b().hasRemaining()) {
                 m_fc.write(destBuf.b());
             }
-
+            // Update segment header
             incrementNumEntries(remaining);
```
rdykiel: So we're still updating the segment header on each buffer? I thoought we'd only update it when closing the segment?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -233,6 +235,20 @@ public void discard() {
                 }
             };
         }
+
+        public boolean isReadFirstObjectOfSegment() throws IOException {
```
rdykiel: 'isStartOfSegment' ?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -93,7 +93,7 @@ public ReadCursor(String cursorId, int numObjectsDeleted) throws IOException {
         }
 
         @Override
-        public BBContainer poll(OutputContainerFactory ocf) throws IOException {
+        public BBContainer poll(OutputContainerFactory ocf, boolean checkCRC) throws IOException {
```
rdykiel: Where is the logic that bypasses checks if the segment file is finalized?

src/frontend/org/voltdb/utils/PBDUtils.java
```diff
@@ -62,4 +62,26 @@ public static void readBufferFully(FileChannel fc, ByteBuffer buf, int startPos)
...
+        crc.update(size);
+        crc.update(flag);
+        crc.update(destBuf);
+        headerBuf.clear();
+        headerBuf.putLong(crc.getValue());
+        headerBuf.putInt(size);
+        headerBuf.putInt(flag);
+    }
+
+    public static byte[] longToBytes(long x) {
```
rdykiel: Those 2 methods are not thread-safe. Is this an issue? Why not use ThreadLocal?

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -194,20 +216,31 @@ int parseAndTruncate(BinaryDeque.BinaryDequeTruncator truncator) throws IOExcept
...
+        }
         int sizeInBytes = 0;
 
         DBBPool.BBContainer cont;
         while (true) {
             final long beforePos = reader.readOffset();
 
-            cont = reader.poll(PersistentBinaryDeque.UNSAFE_CONTAINER_FACTORY);
+            if (reader.readIndex() == 0) {
+                reader.getSchema(PersistentBinaryDeque.UNSAFE_CONTAINER_FACTORY, !isFinal());
```
rdykiel: I think we should optimize isFinal() , by caching an in-memory member variable that reflects the state of the property in the file.

src/frontend/org/voltdb/export/StreamBlockQueue.java
```diff
@@ -41,12 +42,16 @@
...
+ *    --- Entry Header   ---
+ *    crc(8) + length(4) + flags(4)
+ *    --- Export-specific Header ---
  *    {
+ *          // Optional, only exists in the first entry.
+ *          exportVersion(1) + generationId(8) + schemaLen(4) + tupleSchema(var length) +
  *          ---Inside schema---
  *          tableNameLength(4) + tableName(var length) + colNameLength(4) + colName(var length) + colType(1) + colLength(4) + ...
  *    }
+ *    seqNo(8) + tupleCount(4) + uniqueId(8)
```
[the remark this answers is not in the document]
luyangco: The layout here is a out-dated, I update the comment with the latest layout.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -451,6 +497,16 @@ public BBContainer poll(OutputContainerFactory factory) throws IOException {
                             }
                         }
                         compressedBuf.b().flip();
+                        if (checkCRC) {
+                            m_crc.update(compressedBuf.b());
+                            if (entryCRC != m_crc.getValue() || INJECT_PBD_CHECKSUM_ERROR) {
```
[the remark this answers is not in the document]
luyangco: I will do this in a follow-up improvement ticket.

luyangco:
To [rdykiel](/rdykiel)'s question:
1. schema data now is separate from row buffer, it stays in its own buffer and both schema and row buffer are passed to GuestProcessor.
2. Now there is only one refcount in StreamBlock.
3. Decoder no longer relies on schema from previous row, it's just the name that hasn't change. Because decoder only set the schema once per segment file now.

src/frontend/org/voltdb/utils/PBDSegment.java
luyangco: I can call isFinal() outside the while loop, so calling it once per segment file and this is only called in recover time.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
luyangco: Will change this in next improvement ticket.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -781,7 +867,7 @@ private void closeAndDeleteSegment(PBDSegment segment) throws IOException {
     }
 
     @Override
-    public synchronized void push(BBContainer objects[]) throws IOException {
+    public synchronized void push(BBContainer objects[], DeferredSerialization ds) throws IOException {
```
rdykiel: Did you re-run the JUnits? There is a test case exercising the push

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -240,6 +272,47 @@ public void discard() {
...
+                        throw new IOException("Tail segment file " + m_segment.file().getName() + " doesn't exist! ");
+                    }
+                    m_segment = m_segments.higherEntry(m_segment.segmentIndex()).getValue();
+                }
+                PBDSegmentReader segmentReader = m_segment.getReader(m_cursorId);
+                // push to PBD will rewind cursors. So, this cursor may have already opened this segment
+                if (segmentReader == null) {
+                    segmentReader = m_segment.openForRead(m_cursorId);
+                }
+                while (!segmentReader.hasMoreEntries()) {
```
rdykiel: As I said in our email exchanges, I think this logic that searches for a specific condition should be factorized because it's more or less similar in 3 methods

luyangco: I rewrite `isStartOfSegment()` and `pollSchema()` to simplify the logic of finding valid segment.

src/frontend/org/voltdb/utils/PBDUtils.java
```diff
@@ -62,13 +61,13 @@ public static void readBufferFully(FileChannel fc, ByteBuffer buf, int startPos)
...
             ByteBuffer destBuf, int size, int flag) {
         crc.update(size);
         crc.update(flag);
         crc.update(destBuf);
-        headerBuf.order(ByteOrder.LITTLE_ENDIAN);
         headerBuf.clear();
+        headerBuf.order(ByteOrder.LITTLE_ENDIAN);
         headerBuf.putLong(crc.getValue());
         headerBuf.putInt(size);
         headerBuf.putInt(flag);
```
n3world: The next line sets the order to BIG_ENDIAN but this assumes that the buffer was BIG_ENDIAN to start. It would be better to save the original endianess and restore it. Also, why change the endianess of the buffer at all?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1168,15 +1168,16 @@ public int getTruncatedBuffSize() throws IOException {
 
         @Override
         public int writeTruncatedObject(ByteBuffer output) throws IOException {
+            output.position(PBDSegment.ENTRY_HEADER_BYTES);
```
n3world: Is it guaranteed that the buffer passed in should be written to starting at position 0 both this line and the header writes below require it?

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1168,15 +1168,16 @@ public int getTruncatedBuffSize() throws IOException {
...
+            m_crc32.update(bytesWritten);
+            m_crc32.update(PBDSegment.NO_FLAGS);
+            output.position(PBDSegment.ENTRY_HEADER_BYTES);
+            m_crc32.update(output); // compute checksum of the entry data
             // write entry header
-            output.flip();
-            output.putLong(m_crc.getValue());
-            output.putInt(bytesWritten);
-            output.putInt(PBDSegment.NO_FLAGS);
+            output.putLong(PBDSegment.ENTRY_HEADER_CRC_OFFSET, m_crc32.getValue());
```
n3world: Not sure if it matters but the value returned by getValue is really an unsigned int so you could store an int here and save 4bytes if that matters at all

src/frontend/org/voltdb/utils/PBDUtils.java
```diff
@@ -62,13 +61,13 @@ public static void readBufferFully(FileChannel fc, ByteBuffer buf, int startPos)
...
-    public static void writeEntryHeader(HybridCrc32 crc, ByteBuffer headerBuf,
+    public static void writeEntryHeader(CRC32 crc, ByteBuffer headerBuf,
             ByteBuffer destBuf, int size, int flag) {
         crc.update(size);
         crc.update(flag);
         crc.update(destBuf);
-        headerBuf.order(ByteOrder.LITTLE_ENDIAN);
         headerBuf.clear();
+        headerBuf.order(ByteOrder.LITTLE_ENDIAN);
         headerBuf.putLong(crc.getValue());
```
n3world: Not sure if it matters but the value returned by getValue is really an unsigned int so you could store an int here and save 4bytes if that matters at all

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1168,15 +1168,16 @@ public int getTruncatedBuffSize() throws IOException {
 
         @Override
         public int writeTruncatedObject(ByteBuffer output) throws IOException {
+            output.position(PBDSegment.ENTRY_HEADER_BYTES);
             int bytesWritten = PBDUtils.writeDeferredSerialization(output, m_ds);
-            m_crc.update(bytesWritten);
-            m_crc.update(PBDSegment.NO_FLAGS);
-            m_crc.updateFromPosition(PBDSegment.ENTRY_HEADER_BYTES, output);
+            m_crc32.update(bytesWritten);
```
n3world: Why not use writeEntryHeader() so that there is only 1 implementation to write the header. You could pass a duplicate or slice of output in for the header buffer.

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -57,24 +57,26 @@
     //  - crc of segment entry (8 bytes),
     //  - total bytes of the entry (4 bytes),
     //  - entry flag (4 bytes)
-    // TODO: Does DR Segment Entry needs a header? PartitionDRGatewayImpl defines
-    //       DR_BLOCK_HEADER_SIZE but also says this header is unused.
     public static final int ENTRY_HEADER_BYTES = 16;
+    public static final int ENTRY_HEADER_CRC_OFFSET = 0;
```
rdykiel: Please put these in the correct order

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1168,15 +1168,16 @@ public int getTruncatedBuffSize() throws IOException {
...
+            m_crc32.update(PBDSegment.NO_FLAGS);
+            output.position(PBDSegment.ENTRY_HEADER_BYTES);
+            m_crc32.update(output); // compute checksum of the entry data
```
n3world: The position of output will still be at the end of the data which was written. You want to set the limit to the position and set the position to the start of the data.

src/frontend/org/voltdb/utils/PBDUtils.java
```diff
@@ -22,16 +22,15 @@
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.FileChannel;
+import java.util.zip.CRC32;
 
 import org.voltcore.utils.DeferredSerialization;
-import org.voltdb.HybridCrc32;
 
 public class PBDUtils {
     public static int writeDeferredSerialization(ByteBuffer mbuf, DeferredSerialization ds) throws IOException
```
n3world: There is nothing about this method which is explicitly for PBD now it seems like it might have a better home in a more generic class

src/frontend/org/voltdb/utils/PBDUtils.java
luyangco: Good advice, I'll use Int to store the checksum.

src/frontend/org/voltdb/utils/PBDUtils.java
luyangco: I'll restore the original endianness of the buffer. The reason to change it is because in the export decoder it always assume the buffer is in little endian and parse the buffer in that way.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
luyangco: Will follow your advice.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
luyangco: Good catch. This line should write the buffer starting from header offset，the line below (the header write) should write starting from position 0.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
luyangco: I don't think so, the line above set the position of output at after entry header so the CRC computes entire entry data.

src/frontend/org/voltdb/utils/PBDUtils.java
luyangco: Move it to `MiscUtils.java`.

luyangco:
@n3world the comment on PBDRegularSegment.offer(DeferredSerialization) is a good catch, the destBuf should be flipped before getting used by CRC32.

luyangco:
@rdykiel to Richard's comment about the approach of polling schema and exposing segmentIndex to StreamBlock, the main concern here is that trying to avoid extra file IO to poll the schema every time when we fetch the stream block.  Guest Processor needs the schema when reading a new segment or EDS is promoted due to leader migration or node failure, other than that bringing schema to Guest Processor is unnecessary. 
Let's set up a discussion session after standup tomorrow to go through your concerns.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
luyangco: The test case exercising push() is passed.

voltdb/log4j.xml
```diff
@@ -64,9 +64,9 @@
         <level value="INFO"/>
     </logger -->
 
-    <!-- logger name="EXPORT">
-        <level value="INFO"/>
-    </logger -->
+    <logger name="EXPORT">
+        <level value="DEBUG"/>
```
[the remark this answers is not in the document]
luyangco: Removed.

PR #5940, opened 2019-02-20
ENG-15223 update exportmeasure bench

jwiebema1:
This change allows the ExportBenchmark to run as it either a standalone client or a socket receiver so we can re-use the client for all of the exports benchmarks.  When the Benchmarks needs to be run as a standalone socket reciever for socket benchmarks the --socketmode=receiver option can be used.  This will also allow the socket receiver to be on a different host then benchmark client.


tests/test_apps/exportbenchmark/client/exportbenchmark/ExportBenchmark.java
```diff
@@ -651,17 +661,26 @@ public void run() {
...
+        if (!config.socketmode.equals("receiver")) {
+            client.close();
+        }
 
-        // Make sure we got serverside stats
-        if (config.target.equals("socket") && serverStats.size() == 0) {
-            System.err.println("ERROR: Never received stats from export clients");
-            success = false;
+        // Make sure we got serverside stats if we are acting as a receiver
+        if (config.target.equals("socket") && (config.socketmode.equals("both") || config.socketmode.equals("receiver"))) {
```
rdykiel: Second time I see this same condition in this code: factorize in "isSocketTest()"?

tests/test_apps/exportbenchmark/client/exportbenchmark/ExportBenchmark.java
```diff
@@ -566,36 +574,49 @@ private void setupSocketListener() {
...
-        } else {
-            benchmarkEndTS = benchmarkWarmupEndTS + (config.duration * 1000);
+        Thread writes = null;
+        if ( !config.socketmode.equals("receiver")) {
+            // Figure out how long to run for
+            benchmarkStartTS = System.currentTimeMillis();
+            if (config.warmup == 0) {
+                benchmarkWarmupEndTS = 0;
+            } else {
+                benchmarkWarmupEndTS = benchmarkStartTS + (config.warmup * 1000);
```
rdykiel: Did you change anything in the stats gathering and analysis? Should we ignore the stats that occur during the warmup period?

jwiebema1: I didn't change anything in the way the stats are collected in this app. If the stats should be ignored during the warmup, then the warmup should be set to 0.

PR #5941, opened 2019-02-20
ENG-15447: New type of SerializableEEException for DR table hash

manjujames:
not found error in EE, which needs to be handled in the Java layer.


src/ee/storage/DRTableNotFoundException.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef DRTABLENOTFOUNDEXCEPTION_H_
+#define DRTABLENOTFOUNDEXCEPTION_H_
+
+#include "common/SerializableEEException.h"
+
+/*
+ * Generated when a DR table cannot be found for table hash found in binary log.
+ */
+class DRTableNotFoundException : public SerializableEEException {
```
n3world: I would almost vote to put the entire exception implementation in the header

src/frontend/org/voltdb/exceptions/DRTableNotFoundException.java
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.ByteBuffer;
+
+/**
+ * Exception thrown by native Execution Engine
+ * when a table cannot be found for DR table hash that come from the remote cluster
+ */
+public class DRTableNotFoundException extends SerializableException {
+    public static final long serialVersionUID = 0L;
+    
+    private int m_hash;
```
JadenW: table handle hash is 64 bit in EE, why is it int in java?

manjujames: Thanks, I will fix it.

PR #5943, opened 2019-02-20
ENG-15376: ALTER TABLE allows you to add TTL to a STREAM, but causes problems

Chaozz:


zeemanhe:
Do we support TTL on stream, delete on stream?

PR #5944, opened 2019-02-20
ENG-15187 Only pause consumer on incompatible schema

yzhang1991:


manjujames:
There is a TestDRCatalogDiffs. Does anything need to be changed there?

We should probably add a test for this case in there.

PR #5946, opened 2019-02-21
ENG-15445: EE now throws DRTableNotFoundException when a table cannot…

manjujames:
… be found for hash

while applying binary log. This gets translated to new ClientResponse code in the Java layer.

### DIFF
--- a/src/ee/storage/BinaryLogSink.cpp
+++ b/src/ee/storage/BinaryLogSink.cpp
@@ -18,6 +18,7 @@
 #include "BinaryLogSink.h"
 
 #include "ConstraintFailureException.h"
+#include "DRTableNotFoundException.h"
 #include "persistenttable.h"
 #include "streamedtable.h"
 #include "tablefactory.h"
@@ -40,6 +41,9 @@
 #include <crc/crc32c.h>
 
 #include <string>
+
+#define throwDRTableNotFoundException(tableHash, ...) \
+do { char _msg_[1024]; snprintf(_msg_, 1024, __VA_ARGS__); throw DRTableNotFoundException(tableHash, _msg_); } while (false)
 
 namespace voltdb {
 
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/client/ClientResponse.java
```diff
@@ -94,6 +94,12 @@
      */
     public static final byte TXN_MISROUTED = -11;
 
+    /**
+     * Status code indicating that a record in DR binary log could not be applied because a table
+     * coult not be found for the table hash from the remote cluster. Used internally only.
```
n3world: could is misspelled. 

It is a little strange that internal only return codes are documented in the public client interface. It would make more sense to have a class for the internal errors that do not go back to a user's client.

src/ee/storage/BinaryLogSink.cpp
```diff
@@ -493,8 +497,8 @@ inline void truncateTable(boost::unordered_map<int64_t, PersistentTable*> &table
         VoltDBEngine *engine, bool replicatedTableOperation, int64_t tableHandle, std::string *tableName) {
     boost::unordered_map<int64_t, PersistentTable*>::iterator tableIter = tables.find(tableHandle);
     if (tableIter == tables.end()) {
-        throwSerializableEEException("Unable to find table %s hash %jd while applying binary log for truncate record",
-                                     tableName->c_str(), (intmax_t)tableHandle);
+        throwDRTableNotFoundException(tableHandle, "Unable to find table %s hash %jd while applying binary log for truncate record",
```
n3world: Very unlikely but if the table name is very very long, > ~924, you could go beyond the 1024 characters which means you end up with a non null terminated string. Not sure if you want to handle that in your macro or even the original macro

manjujames: I had looked at snprintf documentation and it does null terminate the string.

PR #5948, opened 2019-02-22
ENG-15328 fix rejoining timeout

zeemanhe:
A rejoining process could  be timed out if the rejoining node does not receive data for a specified period of time (60 s by default).  It seems that the mailboxes for initiators for partitions must be created in sorted order by partition id.  Otherwise,  the rejoining node could end up receiving no data.  Now the partition id order will be ensured.

### DIFF
--- a/src/frontend/org/voltdb/AbstractTopology.java
+++ b/src/frontend/org/voltdb/AbstractTopology.java
@@ -158,14 +158,7 @@ public class AbstractTopology {
         }
 
         private Host(Host oldHost, int newId, Map<Integer, Partition> allPartitions) {
-            ImmutableSortedSet.Builder<Partition> builder = ImmutableSortedSet.naturalOrder();
-            for (Partition p : oldHost.partitions) {
-                builder.add(allPartitions.get(p.id));
-            }
-            id = newId;
-            haGroup = oldHost.haGroup;
-            partitions = builder.build();
-            isMissing = false;
+           this(oldHost, newId, allPartitions, false);
         }
 
         private Host(HostBuilder hostBuilder, Map<Integer, Partition> allPartitions) {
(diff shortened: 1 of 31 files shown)


src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -93,19 +94,15 @@
 
     public static class Partition implements Comparable<Partition>{
         public final int id;
-        private int leaderHostId;
-        private ImmutableSortedSet<Integer> hostIds;
+        public int leaderHostId;
```
rdykiel: Question: why are we exposing state variables instead of providing accessor functions?

src/frontend/org/voltdb/AbstractTopology.java
```diff
@@ -225,14 +214,10 @@ public int getleaderCount() {
...
 
-        public void updatePartitions(Collection<Partition> thePartitions) {
-            partitions = ImmutableSortedSet.copyOf(thePartitions);
-        }
-
         @Override
         public String toString() {
             String[] partitionIdStrings = partitions.stream().map(p -> String.valueOf(p.id)).toArray(String[]::new);
-            return String.format("Host %d, ha:%s (Partitions %s)",
+            return String.format("Host %d sph:%d ha:%s (Partitions %s)",
```
rdykiel: Bug? format requires 4 params but only 3 are provided?

zeemanhe: good catch

src/frontend/org/voltdb/AbstractTopology.java
zeemanhe: good question. guess this is static class i s used only within the class. will make it final.

PR #5952, opened 2019-02-22
Eng 15347.enrich.calcite.rel distribution.class

kaikaizi:
Accompanying Calcite changes: https://github.com/VoltDB/calcite/pull/6

This is just a showcase of how we can take advantages of added methods to Calcite to facilitate decision on MP/SP on table joins.

Because Calcite fails to push filters either into or out of join condition in various cases, the distinction of MP/SP is still very unreliable. See updates, commented-out codes and `TODO`s in `TestMPQueryFallBackRules.java`.

@alexeevm has changes in https://github.com/alexeevm/voltdb/blob/calcite-prototype-2-mergejoin/src/frontend/org/voltdb/calciteadapter/rules/logical/VoltDBLCalcJoinTransposeRule.java that could probably alleviate this dilema.

The burden to marshal filters around `RelNode`s and condition checks are now shifted onto the new util class, `RelDistributionUtils.java`, which is, sadly becoming more debug-able than readable.

Also, I think that the MP/SP on set operations might be cumbersome as well (could be easier, as the filters are always in their individual calc/table scan node pairs, rather than being pushed down; except for the case that each set of UNION/INTERCEPT operations are table joins, which then relies on the ability of MP/SP distinction in table joining operation, which I am trying to address).


src/frontend/org/voltdb/plannerv2/rules/PlannerRules.java
```diff
@@ -90,6 +107,72 @@ public RuleSet getRules() {
         public abstract RuleSet getRules();
     }
 
+    private static final FilterJoinRule.Predicate MOVEABLE_TO_JOIN_COND = (join, joinType, expr) -> {     // moveableToOn
```
alexeevm: I might be wrong, but isn't it the case that a WHERE expression can't be moved to a ON clause if this is an OUTER join? In all other cases, WHERE and ON expressions can be combined.

src/frontend/org/voltdb/plannerv2/rules/PlannerRules.java
```diff
@@ -90,6 +107,72 @@ public RuleSet getRules() {
...
+        final RexNode rightExpression = ((RexCall) expr).getOperands().get(1);
+        if (leftExpression.isA(SqlKind.INPUT_REF) && rightExpression.isA(SqlKind.LITERAL)
+                || rightExpression.isA(SqlKind.INPUT_REF) && leftExpression.isA(SqlKind.LITERAL)) {
+            final RelOptTable outerRel = ((RelSubset) join.getLeft()).getOriginal().getTable(),
+                    innerRel = ((RelSubset) join.getRight()).getOriginal().getTable();
+            final Table outer = ((RelOptTableImpl) outerRel).getTable(),
+                    inner = ((RelOptTableImpl) innerRel).getTable();
+            final Integer outerPartitionColumn = outer.getPartitionColumn(),
+                    innerPartitionColumn = inner.getPartitionColumn();
+            if (outerPartitionColumn == null || innerPartitionColumn == null) {
```
alexeevm: should it be != instead of ==?

src/frontend/org/voltdb/plannerv2/rules/PlannerRules.java
```diff
@@ -90,6 +107,72 @@ public RuleSet getRules() {
...
+    private static final FilterJoinRule.Predicate MOVEABLE_TO_JOIN_COND = (join, joinType, expr) -> {     // moveableToOn
+        if (joinType != JoinRelType.INNER || ! (expr instanceof RexCall)
+                || ! expr.isA(SqlKind.EQUALS)) {
+            return false;
+        }
+        final RexNode leftExpression = ((RexCall) expr).getOperands().get(0);
+        final RexNode rightExpression = ((RexCall) expr).getOperands().get(1);
+        if (leftExpression.isA(SqlKind.INPUT_REF) && rightExpression.isA(SqlKind.LITERAL)
+                || rightExpression.isA(SqlKind.INPUT_REF) && leftExpression.isA(SqlKind.LITERAL)) {
+            final RelOptTable outerRel = ((RelSubset) join.getLeft()).getOriginal().getTable(),
```
alexeevm: getTable() will return NULL if an expression does not represent an access to a table. For example, in a multi table join left or right would a join itself

src/frontend/org/voltdb/plannerv2/rules/PlannerRules.java
```diff
@@ -90,6 +107,72 @@ public RuleSet getRules() {
...
     }
 
+    private static final FilterJoinRule.Predicate MOVEABLE_TO_JOIN_COND = (join, joinType, expr) -> {     // moveableToOn
+        if (joinType != JoinRelType.INNER || ! (expr instanceof RexCall)
+                || ! expr.isA(SqlKind.EQUALS)) {
+            return false;
+        }
+        final RexNode leftExpression = ((RexCall) expr).getOperands().get(0);
+        final RexNode rightExpression = ((RexCall) expr).getOperands().get(1);
+        if (leftExpression.isA(SqlKind.INPUT_REF) && rightExpression.isA(SqlKind.LITERAL)
```
alexeevm: could be a CAST. I have a helper method that extract input or field reference from an expression bypassing a possible CAST

src/frontend/org/voltdb/plannerv2/rules/PlannerRules.java
alexeevm: I am not quite sure what is the purpose of this predicate, but I think that a decision whether to push / or not to push filters down through a join should be based on join's semantic only and not whether join's tables are partitioned or not

A decision whether a join itself is a SP could be made by inspecting its ON expression whether it contains an equality expression involving outer and inner partitioning columns and children's distribution traits have compatible partitioning values. 
I think that the FilterJoinRule works as expected (pushing down filters to join's children whenever possible) but the new expression tree could have a higher cost than the original one once it's converted to the VoltLogical convention and gets rejected. To ensure that the new plan is always superior to the original one we can tweak the FilterJoinRule to match VoltLogicalFilter/ Join and manipulate Join self cost based on a number of pushed down filters.

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+     * "WHERE partitionColumn = LITERAL_VALUE"), then also set the literal value in the distribution trait.
+     * @param join join node
+     * @param outer outer relation
+     * @param inner inner relation
+     * @return true if the result of the join is SP, and sets the distribution trait for all partitioned relations.
+     */
+    static boolean checkSPAndPropogateDistribution(VoltLogicalJoin join, RelNode outer, RelNode inner) {
+        // NOTE: we MUST cleanse the table scans to get rid of partitionEqualValue that got propagated inside,
+        // because for multi-join, the result of outer join tree will set its partition equal value, which gets
+        // propagated to inner table scan.
```
alexeevm: This is what I can't figure out. If a distribution trait is always propagated bottom up, the partitioning value can only be set by the MPQueryFallback rule when it matches a Calc / Scan expression.

src/frontend/org/voltdb/plannerv2/rules/PlannerRules.java
kaikaizi: This was my earlier attempt to prevent certain conditions from being pushed down from join-condition. I should have remove this.

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
kaikaizi: This was actually because what I found when running 2 queries in succession, that in the 2nd query, the distribution trait left over from the first query was preserved:

> CREATE TABLE t1(i VARCHAR(8) NOT NULL, j FLOAT);
CREATE TABLE t2 (j FLOAT, i VARCHAR(8) NOT NULL, k VARCHAR(8));
PARTITION TABLE t1 ON COLUMN i;
PARTITION TABLE t2 ON COLUMN i;
SELECT t1.j FROM t1, t2 WHERE t1.i = t2.i AND t1.i = 'foobar' AND t2.i = 'foo'; -- After executing this, the distribution value 'foobar' is set in table scan of t1, and the distribution value 'foo' is set in the table scan of t2, and they are preserved!
SELECT t1.j FROM t1 INNER JOIN t2 ON t1.i = t2.i AND t2.i = 'foobar' AND t1.j = 0; -- Here, because the distribution value of t2 is carried over (if we don't cleanse it out), we will mistake the query as MP, effectively:
-- SELECT t1.j FROM t1 INNER JOIN t2 ON t1.i = t2.i AND t1.i = 'foobar' AND t2.i = 'foo' AND t1.j = 0;

I am guessing that it might have something to do with the way Calcite itself caches/reuses plans in the same session but across queries.

alexeevm: The reason might be in the isCalcScanSP implementation - the partitioning value is set(line 615) on RelDistribution trait that is associated with a table. It should be actually set on a trait instead (both scan and calc).

This is why it's cached

kaikaizi: Thanks! I am working on it.

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+            // WHERE col1 = col2 AND col1 = col3 AND col4 = col2 AND col5 = col6 ==>
+            // {col1, col2, col3, col4}, {col5, col6}
+            final Set<Set<Integer>> r = new HashSet<>();
+            joinCondition.getOperands()
+                    .forEach(node -> {
+                        if (node instanceof RexCall) {  // Calcite flattens cascaded AND into a list (x1 AND x2 AND x3 ...) => AND(x1, x2, x3, ...)
+                            final Pair<Integer, Integer> columnPair = getJoiningColumns((RexCall) node);
+                            if (columnPair != null) {
+                                final Integer fst = columnPair.getFirst(), snd = columnPair.getSecond();
+                                final AtomicBoolean updated = new AtomicBoolean(false);
```
alexeevm: Out of curiosity, why it has to be an atomic boolean? The access to "updated" is always from within a single thread of control as far as I can see

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+     * For other forms of join condition, return null.
+     * @param joinCondition The join condition
+     * @return ordered indexes of column references.
+     */
+    private static Pair<Integer, Integer> getJoiningColumns(RexCall joinCondition) {
+        if (! joinCondition.isA(SqlKind.EQUALS)) {
+            return null;
+        } else {
+            final RexNode leftConj = joinCondition.getOperands().get(0), rightConj = joinCondition.getOperands().get(1);
+            if (!(leftConj instanceof RexInputRef) || !(rightConj instanceof RexInputRef)) {
```
alexeevm: consider a case T1.c1 = T2.c2 where T1.c1 is an integer and T2.c2 is a BIG integer. The actual expression would be
CAST_TO_BIGINT(T1.C1) = T2.C2 and the IF condition will be satisfied. I think you need to pass join condition operands through a function similar to this
https://github.com/alexeevm/voltdb/blob/7c0de15d7e98f942aa4253371a75cce75e8b7ccb/src/frontend/org/voltdb/calciteadapter/util/VoltDBRexUtil.java#L374

to bypass a possible CAST expression

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+                }
+                return Pair.of(col, literal);
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Collects all the conjunction of ColumnRef = Literal forms.
+     * e.g. foo.i = 1 AND foo.j = 'foo' AND bar.k = 0, with foo(i int, j varchar), bar(i int, i int), gives:
```
alexeevm: bar(i int, k int)

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+     * {(0, 1), (1, 'foo'), (1, 0)}.
+     * @param call condition to be checked
+     * @return A collection of column literal values.
+     */
+    private static Map<Integer, RexLiteral> getAllColumnValuePairs(RexCall call) {
+        if (call.isA(SqlKind.AND)) {
+            return call.getOperands().stream().flatMap(entry -> {
+                if (entry instanceof RexCall) {
+                    final Pair<Integer, RexLiteral> p = getColumnValuePairs((RexCall) entry);
```
alexeevm: if getColumnValuePairs would return an Optional (possible an empty) then getOrElse could be used to complete a stream instead of if (p != null) check.

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+                return null;
+            } else {
+                return dist.getKeys().get(0);
+            }
+        } else if (tbl instanceof VoltLogicalJoin) {    // join tree with a child being a join node
+            final List<Integer> keys =
+                    tbl.getTraitSet().getTrait(RelDistributionTraitDef.INSTANCE).getKeys();
+            return keys.isEmpty() ? null : keys.get(0);
+        } else if (tbl instanceof VoltLogicalCalc) {
+            final VoltLogicalTableScan scan = getTableScan(tbl);
```
alexeevm: I think this is a dangerous assumption - getTableScan expect tbl's child to be either a Scan or Join. It can be anything actually, a SetOp for example.
Why can't it always be accessed via a trait set similar to 
tbl.getTraitSet().getTrait(RelDistributionTraitDef.INSTANCE).getKeys();

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+            index = null;
+            literal = null;
+        }
+        if (index != null) {
+            map.put(index, literal);
+        }
+    }
+
+    /**
+     * Helper to retrieve all "ColumnRef = LiteralValue" pairs from a Calc node
```
alexeevm: Do we need to support ColumnRef = FUNC(LiteralValue) type of expressions?
foo = TO_UPPER('something')

src/frontend/org/voltdb/plannerv2/rules/logical/RelDistributionUtils.java
```diff
@@ -0,0 +1,623 @@
...
+        } else if (partCol1 != null) {
+            return Pair.of(partCol1, eqCols.stream().filter(col -> col >= outerTableColumns).findAny().get());
+        } else {
+            return Pair.of(eqCols.stream().filter(col -> col < outerTableColumns).findAny().get(),
+                    partCol2 + outerTableColumns);
+        }
+    }
+
+    static RelDistribution getDistribution(RelNode node) {
+        if (node instanceof VoltLogicalTableScan) {
```
alexeevm: I would always return from the trait set to avoid modifying the trait associated with a table itself

src/frontend/org/voltdb/plannerv2/rules/logical/MPQueryFallBackRule.java
```diff
@@ -68,20 +56,9 @@ private MPQueryFallBackRule() {
...
-            // For partitioned tables, the distribution type will be HASH_DISTRIBUTED.
-            if (tableDist.getType() != RelDistribution.Type.SINGLETON
-                    && (calc.getProgram().getCondition() == null // SELECT ... FROM t (no filter)
-                            || ! isSinglePartitioned(calc.getProgram(), calc.getProgram().getCondition(), tableDist.getKeys()))) {
-                    throw new PlannerFallbackException("MP query not supported in Calcite planner.");
-            }
-            call.transformTo(calc.copy(calc.getTraitSet().replace(tableDist), calc.getInputs()));
+            // TODO: the exception thrown here could be too early, e.g. the result is used as a child of a join,
+            // which will change MP decision to SP at that point.
+            RelDistributionUtils.checkedFallBack(! RelDistributionUtils.isCalcScanSP(tableScan, calc));
```
alexeevm: RelDistributionUtils.isCalcScanSP seems to be doing two things - the actual SP/MP check and tries to update existing nodes's traits which won't work. It should only do a check. An additional method should create a new calc/scan pair with updated traits to be registered by the rule

alexeevm:
> > This is line does not have any effect at all. The replace(dist) creates a copy of a current trait set, modifies it and returns leaving the existing one intact.
> > The logic should be
> > ```
> > 1. Create a copy of a distilled trait with a partitioning value set
> > 
> > 2. Create a new trait set with updated distribution trait
> > 
> > 3. Create a copy of the scan / calc node with a new trait set
> > 
> > 4. Call call.transformTo(newRel) to register the new rel nodes
> > ```
> 
> Now I found that currently there is no true way to _clone_ a trait:
> 
> * The `AbstractRelNode.copy()` method either returns itself when you want to make a copy; or forbids being called in any other manner;
> * The `RelTraitSet.replace(trait)` method seems to be that it created a copy, and we can change on the copy while keep `this` unchanged; but the caveat is that the argument `trait` is referenced; so when we change the replace-d `RelTraitSet` object, i.e. by changing its partition equal value, the argument itself, which came from `tableScan.getTable().getTraitSet()` that reflects the partition-ness of the table, also gets changed.
> 
> This wasn't a problem for Calcite because, the only implementation of interface `RelDistribution`, the `RelDistributions.RelDistributionImpl`, is immutable: all its attributes were immutable; but now we added a mutable field, `partitionEqualValue`, breaking the previous assumption.
> 
> I could add a true _clone_ method for a trait, and change on the clone; but the better method would be to keep `RelDistributions.RelDistributionImpl` immutable, and change mutator `void RelDistribution.setPartitionEqualValue(RexNode)` to a constructor: `RelDistribution withPartitionEqualValue(RexNode)`.

Yeah, I would add a private constructor that accepts a RexLiteral and also overload the RelDistributions.hash method to accept collection of indexes and a partitioning value. Having an original trait you could always create a new one then

kaikaizi:
@alexeevm I have addressed your comments and (finally) passed all JUnit tests on join MP/SP that I could come up with.

kaikaizi:
These changes are now in calcite branch of voltdb repo; and voltdb branch of calcite repo.

PR #5958, opened 2019-02-26
ENG-15193:

manjujames:
- Removed configuration for flexible vs. strict schema mode and always use flexible mode.
- Pause DR streams when we receive a binary log record for which schema does not match and
  unpause on local catalog update.


johnTcrawford:
I see that sqlcmdtest is failing on this branch, see:
http://ci.voltdb.lan/view/Branch-jobs/view/branch-ENG-15193/job/branch-2-sqlcmdtest-ENG-15193/.
So this is setting us up for another ENG-15468, in reverse, unless the appropriate adjustment is made to tests/sqlcmd/baselines/sysproc/explaincatalog.outbaseline. See:
https://github.com/VoltDB/voltdb/pull/5951/files

PR #5963, opened 2019-02-27
ENG-15408 restrict nibble export for replicated tables

zeemanhe:
Compiler throws errors if nibble export is configured on replicated table.

### DIFF
--- a/src/frontend/org/voltdb/compiler/DDLCompiler.java
+++ b/src/frontend/org/voltdb/compiler/DDLCompiler.java
@@ -1390,7 +1390,7 @@ public class DDLCompiler {
             final String migrationTarget = ttlNode.attributes.get("migrationTarget");
             if (!StringUtil.isEmpty(migrationTarget)) {
                 if (table.getIsreplicated()) {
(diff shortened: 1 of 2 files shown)


src/frontend/org/voltdb/compiler/DDLCompiler.java
```diff
@@ -1389,7 +1388,10 @@ private void addTableToCatalog(Database db, VoltXMLElement node, boolean isXDCR)
             ttlValue = Integer.parseInt(ttlNode.attributes.get("maxFrequency"));
             ttl.setMaxfrequency(ttlValue);
             final String migrationTarget = ttlNode.attributes.get("migrationTarget");
-            if (migrationTarget != null) {
+            if (!StringUtil.isEmpty(migrationTarget)) {
+                if (table.getIsreplicated()) {
+                    throw m_compiler.new VoltCompilerException("Nibble export is not supported for replicated table:" + table.getTypeName());
```
rdykiel: I'm not sure that users will understand "Nibble Export" as this is an internal name for the feature. You should instead refer to the actual DDL command that's not allowed, e.g. "MIGRATE TO TARGET"?

PR #5964, opened 2019-02-27
ENG-13605: Use MessagingChannel for kerberos auth

n3world:
When a connection is encrypted the only way to send messages through that
connection is by using the MessageChannel. Kerberos authentication methods were
all using the raw SocketChannel and were sending unencrypted messages on an
encrypted connection. Always use the MessagingChannel to send messages over a
connection.

### DIFF
--- a/src/frontend/org/voltdb/client/ConnectionUtil.java
+++ b/src/frontend/org/voltdb/client/ConnectionUtil.java
@@ -415,7 +415,7 @@ public class ConnectionUtil {
                     throw new IOException("Wire protocol format violation error");
                 }
                 String servicePrincipal = SerializationHelper.getString(loginResponse);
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/client/ConnectionUtil.java
```diff
@@ -614,7 +596,6 @@ public GSSContext run() {
             });
         } catch (SecurityException ex) {
             // if we get here the authentication handshake failed.
-            try { channel.close(); } catch (Exception ignoreIt) {}
```
JadenW: Is MessagingChannel need to close the wrapped socketChannel in the error scenery?
The cleanup() method of it seems just a stub.

n3world: The calling method handles closing the socket on exception. The closes that were in this method were actually not needed but didn't hurt.

PR #5973, opened 2019-02-28
ENG-14942: Disable JVM stats if tmp is not a tmpfs

n3world:
Determine the temp directory to be used by java and if it is not on a tmpfs
filesystem set the option -XX:+PerfDisableSharedMem to disable JVM stats.

### DIFF
--- a/lib/python/voltcli/environment.py
+++ b/lib/python/voltcli/environment.py
@@ -91,6 +91,11 @@ if not [opt for opt in java_opts if opt.startswith('-Xmx')]:
         java_opts.append('-Xms2048m')
         java_opts.append('-XX:+AlwaysPreTouch')
 
+# Find the temp directory which is going to be used by java
+tmpDir = '/tmp'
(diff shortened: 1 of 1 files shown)


lib/python/voltcli/environment.py
```diff
@@ -118,6 +123,27 @@
...
+    while path and not os.path.ismount(path):
+        path = os.path.dirname(path)
+    if not path:
+        return False
+    is_tmpfs = False
+    with open('/proc/mounts') as f:
+        for line in f:
+            line = line.split()
+            if line[1] == path:
+                is_tmpfs = line[2] == 'tmpfs'
```
JadenW: nitpick: Early return if path has been found in /proc/mounts ?

n3world: Actually you don't want to do that. You can mount over mounts so technically there could be more than one entry for the same mount point and you want to take the last one

PR #5974, opened 2019-02-28
ENG-15230 k8s scripts add support multiple jars and ddl files

prosegay:


tools/kubernetes/bin/voltdbk8s.py
```diff
@@ -227,6 +227,22 @@ def hang():
     sp.wait()
 
 
+def get_files_list(dir):
+    files = [f for f in os.listdir(dir) if not f.startswith('..')]
```
n3world: listdir does not return `..` and `.` so you don't need to skip those

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
 }
 
 
+find_files() {
+    # takes a comma-separated list of file paths, which may include Unix-style wildcards (*?[ch-ranges])
+    # returns a comma separated list of all matching files, following links
+    files=`find -H ${@//,/\  } -maxdepth 1 -type f -print`
+    echo $files
```
n3world: You can skip the variable assignment and echo and just have the body of the method be `find -H ${@//,/\  } -maxdepth 1 -type f -print`

I think you could simplify this and just use `ls ${@//,/  }`

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
+}
+
+build_configmap_cmd() {
+    I=0
+    for f in $@
+    do
+        echo -n "--from-file=${f} "
+        let I+=1
```
n3world: Not sure what the minimum supported version of bash is but this can also be done as `((++I))`

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
 
+find_files() {
+    # takes a comma-separated list of file paths, which may include Unix-style wildcards (*?[ch-ranges])
+    # returns a comma separated list of all matching files, following links
+    files=`find -H ${@//,/\  } -maxdepth 1 -type f -print`
+    echo $files
+}
+
+build_configmap_cmd() {
+    I=0
```
n3world: I recommend using `local I=0` so that I doesn't go into global scope

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
+    for f in $@
+    do
+        echo -n "--from-file=${f} "
+        let I+=1
+    done
+    if [[ $I -gt 1 ]]; then
+        FNS=""
```
n3world: You could move the building of the FNS variable up to the first loop and then this if block just persists it if necessary

Also, should be local

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
+        for f in $@
+        do
+            FNS+="`basename ${f}`,"
+        done
+        ORDERFILE=`mktemp`
+        echo "$FNS" | sed -e 's/,$//' > $ORDERFILE
+        echo -n "--from-file=.loadorder=$ORDERFILE"
+    fi
+    echo -n ""
```
n3world: This doesn't print anything

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
+    for f in $@
+    do
+        echo -n "--from-file=${f} "
+        let I+=1
+    done
+    if [[ $I -gt 1 ]]; then
+        FNS=""
+        for f in $@
+        do
+            FNS+="`basename ${f}`,"
```
n3world: Also, dependent on bash version but using `$(...)` can be more flexible than the ticks. ie `FNS+="$(basename ${f},)"
You can nest commands together `A=$(cmd1 $(getArg1)  $(getArg2))`

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
+}
+
+
 make_configmap() {
+
+    MAPNAME=${CLUSTER_NAME}-init-classes
+    delete_configmap_ifexists $MAPNAME
+    CONFIG_MAP_ARGS=""
+    TYPE=classes
+    if [ ! -z "${CLASSES_FILES}" ]; then
```
n3world: You use `[[` in another conditional. It is probably good to be consistent. You can also express `! -z` as either `-n` or no flag since by default it tests for content `[[ $CLASSES_FILES ]]`

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -48,11 +103,11 @@ make_configmap() {
     [ ! -z "${LOG4J_CONFIG_PATH}"  ] && echo "LOG4J_CONFIG_PATH=\"/etc/voltdb/log4jcfg\"" >> ${PROP_FILE}
     [ ! -z "${VOLTDB_OPTS}" ] && echo "VOLTDB_OPTS=${VOLTDB_OPTS}"                    >> ${PROP_FILE}
     grep VOLTDB_HEAPMAX         ${CFG_FILE} >> ${PROP_FILE}
-    cat ${PROP_FILE}
+    #cat ${PROP_FILE}
```
n3world: Can remove or if you want for debugging you could do a debug mode for the script and only print if it is enabled

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -24,14 +24,69 @@ make_statefulset() {
...
+        echo -n "--from-file=.loadorder=$ORDERFILE"
+    fi
+    echo -n ""
+}
+
+
+delete_configmap_ifexists() {
+    set +e  # don't check errors, may not exist
+    kubectl delete configmap $1
+    set -e
```
n3world: Minor issue with this is that if somebody is running this script with `-e` this will disable it. If you want to allow that to work you can test if `-e` is set with something like `[[ $- =~ e ]]` and only set and unset if this returns false

tools/kubernetes/bin/voltdbk8s.py
prosegay: actually there are files named ..data, etc, that k8s uses in conjunction with configmaps, those are the files that we're skipping, they exist even if the configmap is empty.

tools/kubernetes/voltdb-k8s-utils.sh
prosegay: one thing I wanted to ask you if you agree that -H is appropriate, i've had second thoughts that I should remove it. the echo was for debugging, i'll remove it.

tools/kubernetes/voltdb-k8s-utils.sh
prosegay: I intentionally put it into a separate loop for "readability" and clarity, originally it was that way

tools/kubernetes/voltdb-k8s-utils.sh
prosegay: but it puts a \n at the end of the output stream, i forget why, but it was needed.

tools/kubernetes/voltdb-k8s-utils.sh
prosegay: cool

tools/kubernetes/voltdb-k8s-utils.sh
n3world: The `-H` would make it so if a directory was supplied as the path and there was a symlink to a file in that directory that symlink would not be followed so that file not used. However if all of the paths specified are to files or symlinks to files then it doesn't do anything. I didn't think about one of the paths being a directory when I made my `ls` comment so using `find` makes sense the `-H` depends on what behavior is desired. I think either with or without is fine just as long as the user knows how it should behave

tools/kubernetes/voltdb-k8s-utils.sh
n3world: It shuldn't the `-n` flag should say do not put a `\n`

tools/kubernetes/voltdb-k8s-utils.sh
```diff
@@ -1,111 +1,118 @@
...
-    set -e
+kubectl_ifexists() {
+    [[ $OPTS =~ e ]] && set +e  # don't check errors, may not exist
+    kubectl get "$2"/"$3" &>/dev/null
+    RC=$?
+    if [[ $RC = 0 ]]; then
+        kubectl $@
+        RC=$?
+    fi
+    [[ $OPTS =~ e ]] && set -e
```
n3world: This will always be false since you always make sure `e` is unset. To get this to work you will have to track if you unset `e` and only if it was unset reset it

tools/kubernetes/bin/voltdbk8s.py
n3world: Oh ok I was thinking you were trying to skip parent directory

PR #5978, opened 2019-03-04
Eng 14565 to nibble export ib

luyangco:


src/frontend/org/voltdb/utils/BinaryDequeReader.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+package org.voltdb.utils;
+
+import java.io.IOException;
+
+import org.voltcore.utils.DBBPool.BBContainer;
+import org.voltdb.utils.BinaryDeque.OutputContainerFactory;
+
+/**
+ * Reader class used to read entries from the deque. Multiple readers may be active at the same time,
```
rdykiel: "interface"

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -258,8 +267,14 @@ public boolean isClosed()
 
     @Override
     public void close() throws IOException {
+        if (exportLog.isDebugEnabled()) {
+            exportLog.debug("Close PBD Segment " + m_file.getName());
+        }
         if (m_fc != null) {
             m_writeOffset = m_fc.position();
```
rdykiel: Why are we setting the write offset on close? Shouldn't we recalculate it on open?

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -312,8 +238,8 @@ int parseAndTruncate(BinaryDeque.BinaryDequeTruncator truncator) throws IOExcept
     ExportSequenceNumberTracker scan(BinaryDeque.BinaryDequeScanner scanner) throws IOException {
         if (!m_closed) throw new IOException(("Segment should not be open before truncation"));
 
-        // Open for write because corrupted file may be truncated
-        openForWrite(false);
+//        // Open for write because corrupted file may be truncated
```
rdykiel: Is this supposed to be removed permanently or is this an oversight?

src/frontend/org/voltdb/utils/BinaryDequeReader.java
luyangco: Updated.

src/frontend/org/voltdb/utils/PBDSegment.java
luyangco: Good catch, this should be removed.

luyangco:
In TestStreamBlockQueue when StreamBlockQueue is created, the schema buffer needs to be written in  segment file. MockVoltdb is installed to give StreamBlockQueue the ability to construct serializer for current schema, because MockVoltDB provides catalog table and its column definition. Once schema buffer is written to PBD segment, the following poll() doesn't trigger IOException which fixes the test.

PR #5981, opened 2019-03-05
ENG-15512 ignore newly introduced tableType on DR catalog diff

zeemanhe:
ignore tableType to allow catalog to be synchronized between clusters

### DIFF
--- a/src/catgen/in/javasrc/DRCatalogDiffEngine.java
+++ b/src/catgen/in/javasrc/DRCatalogDiffEngine.java
@@ -157,6 +157,9 @@ public class DRCatalogDiffEngine extends CatalogDiffEngine {
             if ("tuplelimit".equals(field)) {
                 return null;
             }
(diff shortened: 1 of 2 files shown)


zeemanhe:
@manjujames agree.    The list could be added in ENG-15516

src/catgen/in/javasrc/DRCatalogDiffEngine.java
```diff
@@ -54,7 +54,9 @@
             "sqltext", "querytype", "readonly", "singlepartition", "replicatedtabledml", "iscontentdeterministic", "isorderdeterministic", "nondeterminismdetail",
             "cost", "seqscancount", "explainplan", "tablesread", "tablesupdated", "indexesused", "cachekeyprefix"
             );
-
+    private static Set<String> s_filterListFields = Sets.newHashSet(
```
n3world: It would be safer to set this variable as final and use an immutable map

PR #5983, opened 2019-03-05
ENG-15117: subselect gives differing results when rerun multiple times

Chaozz:


src/frontend/org/voltdb/planner/StatementPartitioning.java
```diff
@@ -499,6 +499,15 @@ public void analyzeForMultiPartitionAccess(Collection<StmtTableScan> scans,
                     break;
                 }
             }
+        } else if (scans.size() == 1) {
+            // For query like "select * from (select DL_REGISTER_IP_CITY  from SJYH_DENGLU2 where DL_USER_ID = '113001' ) as result;"
```
kaikaizi: Also put JIRA number in here.

src/frontend/org/voltdb/planner/StatementPartitioning.java
```diff
@@ -499,6 +499,15 @@ public void analyzeForMultiPartitionAccess(Collection<StmtTableScan> scans,
...
+            // There is no filter on the outer query,
+            // the inferredParameterIndex should be inferred from the subquery instead of uninitialized (-1).
+            StmtTableScan tableScan = scans.iterator().next();
+            if (tableScan instanceof StmtSubqueryScan) {
```
kaikaizi: Should it be an assert? If it fails no other tests, let's change it to assert.

src/frontend/org/voltdb/planner/StatementPartitioning.java
Chaozz: done.

src/frontend/org/voltdb/planner/StatementPartitioning.java
Chaozz: No.
`Select * from T` will also hit this branch. In this case the tableScan is a instance of `StmtTargetTableScan`. Only subselect query have the type `StmtSubqueryScan`.

kaikaizi: Right. Thanks for the clarification.

Chaozz:
The SQL Coverage tests look good, merge it to master.

PR #6001, opened 2019-03-11
ENG-15557 retain TTL while table is altered. 

zeemanhe:
The pull request addressed the issue that TTL definition is accidentally dropped  while a table is altered. And enforce a restriction that a column used by TTL is not allowed to be dropped.

### DIFF
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
@@ -4212,7 +4212,7 @@ public class ParserDDL extends ParserRoutine {
         if (table.timeToLive != null) {
             final String ttlColumn = table.timeToLive.ttlColumn.getNameString();
             if (colName.equalsIgnoreCase(ttlColumn)) {
-                throw Error.error("This column used by TTL cannot be dropped.");
+                throw Error.error("Columns used by TTL cannot be dropped.");
             }
         }
         Object[] args = new Object[] {
--- a/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
+++ b/src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
@@ -4209,6 +4209,12 @@ public class ParserDDL extends ParserRoutine {
             throw Error.error(ErrorCode.X_42591);
         }
 
+        if (table.timeToLive != null) {
+            final String ttlColumn = table.timeToLive.ttlColumn.getNameString();
+            if (colName.equalsIgnoreCase(ttlColumn)) {
+                throw Error.error("This column used by TTL cannot be dropped.");
+            }
+        }
         Object[] args = new Object[] {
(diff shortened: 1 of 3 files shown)


src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -4209,6 +4209,12 @@ Statement compileAlterTableDropColumn(Table table, String colName,
             throw Error.error(ErrorCode.X_42591);
         }
 
+        if (table.timeToLive != null) {
+            final String ttlColumn = table.timeToLive.ttlColumn.getNameString();
+            if (colName.equalsIgnoreCase(ttlColumn)) {
+                throw Error.error("This column used by TTL cannot be dropped.");
```
n3world: I would remove the worth `This` from the error message and just have it be `Columns used by TTL cannot be dropped`

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -4209,6 +4209,12 @@ Statement compileAlterTableDropColumn(Table table, String colName,
             throw Error.error(ErrorCode.X_42591);
         }
 
+        if (table.timeToLive != null) {
```
n3world: If cascade is `true` would this also drop the TTL since the TTL column is being removed?

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
zeemanhe: updated.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
zeemanhe: Per doc, CASCADE is only supported for alter table drop  column or change definition.  This should cover any cases that a column is being removed.

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
```diff
@@ -4209,6 +4209,12 @@ Statement compileAlterTableDropColumn(Table table, String colName,
...
+            final String ttlColumn = table.timeToLive.ttlColumn.getNameString();
+            if (colName.equalsIgnoreCase(ttlColumn)) {
+                throw Error.error("Columns used by TTL cannot be dropped.");
```
rdykiel: Andrew might need to know this for doc purposes

src/hsqldb19b3/org/hsqldb_voltpatches/ParserDDL.java
n3world: That is basically what I was trying to get at. If cascade is `true` shouldn't this check not be done and the TTL just removed from the table ie the drop column cascades into dropping the TTL?

PR #6004, opened 2019-03-12
ENG-15503: fix grammar-gen failures due to sqlcmd hanging

johnTcrawford:
... by adding 3 missing error messages that currently go unrecognized,
and modifying a fourth, plus some special code for CREATE VIEW
statements that just return 'null' (see ENG-15587); also, include the
current time in the final SUMMARY message.

### DIFF
--- a/tests/sqlgrammar/sql_grammar_generator.py
+++ b/tests/sqlgrammar/sql_grammar_generator.py
@@ -635,7 +635,8 @@ def print_summary(error_message=''):
                          + get_last_n_sql_statements(last_n_sql_statements, False, False)
         seconds = time() - start_time
         summary_message  = '\n\nSUMMARY: in ' + re.sub('^0:', '', str(timedelta(0, round(seconds))), 1) \
-                         + ' ({0:.3f} seconds)'.format(seconds) + ', SQL statements by type:'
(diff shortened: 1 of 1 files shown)


tests/sqlgrammar/sql_grammar_generator.py
```diff
@@ -1344,8 +1354,11 @@ def generate_sql_statements(sql_statement_type, num_sql_statements=0, max_save_s
                             ['Materialized view only supports INNER JOIN'],
                             ['windowed function call and GROUP BY in a single query is not supported'],
                             ['materialized view does not support self-join'],
-                            ['Windowed RANK function call expressions require an ORDER BY specification'],
```
kaikaizi: Is this removal intentional?

johnTcrawford: Yes, it's replaced with:
'Windowed', 'function call expressions require an ORDER BY specification'
which catches both:
'Windowed RANK function call expressions require an ORDER BY specification'
as before, and:
'Windowed DENSE_RANK function call expressions require an ORDER BY specification'
which I saw recently, as well as any other function name that might appear in that message.

PR #6009, opened 2019-03-13
ENG-15347 reset migrate hidden column value for elastic join

zeemanhe:
The migrate hidden column holds transaction ids. While tuples are streamed over to a new elastically joined node, the transaction ids are not relevant on the new node since the new node hosts new partitions and partition ids are part of transaction ids. The values of migrate hidden column will be reset to null while tuples are streamed over to newly joined hosts.

### DIFF
--- a/src/ee/common/tabletuple.h
+++ b/src/ee/common/tabletuple.h
@@ -499,6 +499,7 @@ public:
     int compare(const TableTuple &other) const;
 
     void deserializeFrom(voltdb::SerializeInputBE &tupleIn, Pool *stringPool);
+    void deserializeFrom(voltdb::SerializeInputBE &tupleIn, Pool *stringPool, bool elasticJoin);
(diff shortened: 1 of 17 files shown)


zeemanhe:
[change on pro](https://github.com/VoltDB/pro/pull/2520)

src/ee/common/tabletuple.h
```diff
@@ -1103,9 +1108,15 @@ inline void TableTuple::deserializeFrom(voltdb::SerializeInputBE &tupleIn, Pool
             throw SerializableEEException(VOLT_EE_EXCEPTION_TYPE_EEEXCEPTION, message.str().c_str());
         }
 
-        char *dataPtr = getWritableDataPtr(columnInfo);
-        NValue::deserializeFrom(tupleIn, dataPool, dataPtr, columnInfo->getVoltType(),
+        if (hiddenColumnMigrateElastic && j == hiddenColumnCount -1) {
```
rdykiel: I would always recommend to add a comment about the migrate column being the last column

PR #6016, opened 2019-03-14
Eng 15590 support migrate delete rollback

wweiss-voltdb:


src/frontend/org/voltdb/compiler/VoltCompiler.java
```diff
@@ -1391,6 +1391,11 @@ void addExportTableToConnector(final String targetName, final String tableName,
                     }
                 }
             }
+            // This is used to enforce default connectors assigned to streams through the Java Property
+            // for streams that don't have the Export To Target clause
+            if (TableType.isStreamViewOnly(tableref.getTabletype())) {
+                tableref.setTabletype(TableType.STREAM.get());
```
zeemanhe: Should we  just not distinguish  streams by connectors, remove STEAM_VIEW_ONLY  type ? Then the conversion may not be needed.

PR #6017, opened 2019-03-14
ENG-15591: MIGRATING SQL function

Chaozz:
Create a SQL function: 
```
MIGRATING()
```
that returns whether the migrating hidden column of the table is NOT NULL. It returns TRUE if the hidden column is NOT NULL, and FALSE otherwise.
The function then can be used like:
```sql
SELECT COUNT(*) FROM tbl WHERE MIGRATING() AND TTL < 1000;
SELECT TTL FROM tbl WHERE NOT MIGRATING();
```
## Limitations
Note that the `MIGRATING()` is **NOT SUPPORTED** in the SELECT clasue and Joined tables, such as
```sql
SELECT MIGRATING() FROM tbl;
SELECT * FROM t1, t2 WHERE NOT MIGRATING();
```
Why? 
Because supporting them will significantly increase the complextiy of the implementation, and there is no obvious use cases.

### DIFF
--- a/src/ee/expressions/functionexpression.cpp
+++ b/src/ee/expressions/functionexpression.cpp
@@ -98,7 +98,20 @@ public:
         : AbstractExpression(EXPRESSION_TYPE_FUNCTION) {
     };
 
-    NValue eval(const TableTuple *, const TableTuple *) const {
+    NValue eval(const TableTuple *tuple1, const TableTuple *tuple2) const {
+        // special handler for MIGRATING() function
+        if (F == FUNC_VOLT_MIGRATING) {
+            // For MIGRATING(), check if we are evaluating on a migrating table(the table with a migrate target).
+            if (tuple1 != NULL && tuple1->getSchema()->isTableWithStream()) {
+                // we have at most 3 hidden columns, DR Timestamp, count for view and transaction id for migrating
+                // and transaction id for migrating is always the last one.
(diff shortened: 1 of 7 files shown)


src/ee/expressions/functionexpression.cpp
```diff
@@ -98,7 +98,20 @@ class ConstantFunctionExpression : public AbstractExpression {
         : AbstractExpression(EXPRESSION_TYPE_FUNCTION) {
     };
 
-    NValue eval(const TableTuple *, const TableTuple *) const {
+    NValue eval(const TableTuple *tuple1, const TableTuple *tuple2) const {
+        // special handler for MIGRATING() function
+        if (F == FUNC_VOLT_MIGRATING) {
+            // For MIGRATING(), check if we are evaluating on a migrating table(the table with a migrate target).
+            if (tuple1 != NULL && tuple1->getSchema()->isTableWithStream()) {
```
yzhang1991: Need to enforce `tuple2 == NULL` before this.
Otherwise, this function will give a wrong answer if a user tries to run something like:
```sql
SELECT * FROM t1 JOIN ttl ON t1.a = ttl.a WHERE migrating;
```

We should give an error message like "Cannot use the migrating function in join queries."

yzhang1991: I see this can be prevented by another guard in Java, it should be OK.

PR #6019, opened 2019-03-14
GCC 8 adds some new warnings

n3world:
Disable warning about parentheses.
Do not error when memcpy or memset could be violating invariants of the class
Fix passing a very large string buffer as a format argument to the exception
Correct catching of polymorphic exception


src/ee/storage/ElasticIndexReadContext.cpp
```diff
@@ -209,7 +209,7 @@ bool ElasticIndexReadContext::parseHashRange(
                                                  boost::lexical_cast<int32_t>(rangeStrings[1]));
                 success = true;
             }
-            catch(boost::bad_lexical_cast) {
+            catch(boost::bad_lexical_cast&) {
```
kaikaizi: A good practice is to use `const&` when catching exceptions.

PR #6027, opened 2019-03-15
ENG-15630: Rejoin snapshot timeout

n3world:
Snapshot streams only send data if there are persistent tables. Rejoin handled
when there were no tables in the cluster but not when there were only streams
in the system. Instead of checking if there are tables, check if there are any
persistent tables.

### DIFF
--- a/src/frontend/org/voltdb/iv2/RejoinProducer.java
+++ b/src/frontend/org/voltdb/iv2/RejoinProducer.java
@@ -59,8 +59,8 @@ public class RejoinProducer extends JoinProducerBase {
     // Stores the name of the views to pause/resume during a rejoin stream snapshot restore process.
     private String m_commaSeparatedNameOfViewsToPause = null;
 
-    // True if we're handling a table-less rejoin.
-    boolean m_schemaHasNoTables = false;
+    // True if there are any persistent tables in the schema.
+    boolean m_hasPersistentTables = true;
 
     // Barrier that prevents the finish task for firing until all sites have finished the stream snapshot
     private static AtomicInteger s_streamingSiteCount;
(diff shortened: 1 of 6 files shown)


src/frontend/org/voltdb/iv2/RejoinProducer.java
```diff
@@ -59,8 +59,8 @@
     // Stores the name of the views to pause/resume during a rejoin stream snapshot restore process.
     private String m_commaSeparatedNameOfViewsToPause = null;
 
-    // True if we're handling a table-less rejoin.
-    boolean m_schemaHasNoTables = false;
+    // True if there are any persistent tables in the schema.
+    boolean m_hasPersistentTables = true;
```
rdykiel: It's funny that this condition should be initialized to 'true' instead of to 'false' and flipped to 'true' on the first table.

tests/frontend/org/voltdb/TestRejoinEndToEnd.java
```diff
@@ -1085,4 +1086,24 @@ public void testRejoinWithExportWithActuallyExportedTables() throws Exception {
...
+        LocalCluster lc = new LocalCluster("rejoin.jar", 2, 3, 1, BackendTarget.NATIVE_EE_JNI);
+        lc.overrideAnyRequestForValgrind();
+        VoltProjectBuilder vpb = new VoltProjectBuilder();
+        vpb.setUseDDLSchema(true);
+        assertTrue(lc.compile(vpb));
+        lc.setHasLocalServer(false);
+        try {
+            lc.startUp();
+            Client client = lc.createClient(new ClientConfig());
+            client.callProcedure("@AdHoc", "CREATE STREAM stream_towns PARTITION ON COLUMN state "
```
rdykiel: I see that my test scenario will now gain persistence in our code base :-)

n3world: It was the easy way to write the test. You gave me all of the code I needed :^)

src/frontend/org/voltdb/iv2/RejoinProducer.java
n3world: This variable is actually assigned in doInitialize `m_hasPersistentTables = message.schemaHasPersistentTables();`

I think you are thinking about `Iv2RejoinCoordinator.startJoin()` but in that case it is a local variable for the method.

PR #6030, opened 2019-03-15
ENG-15608. Fix libcatalog placement.

rmorgenstein:
The libcatalog is checked out of source control for both the Mac( .jnilib) and the Linux (.so) versions.  But during the ant build, only the version for the current OS was being put in the right place.

There is a step for official final kit builds that calls the tools/lib-copy.sh script and it was attempting to get the libcatalog, but unfortunately, it wasn't in place for the copy at this step. Rather than fix that problem, I reverted lib-copy.sh to its old task of just making libvoltdb get from the Mac into the final Mac/Linux kit.

The real fix is that now build.xml put the libcatalog into a Mac and Linux dir regardless of which build is being done.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -840,7 +840,7 @@ DISTRIBUTION
     <antcall target="call_strip_lib" />
 
     <!-- copy to dir where jar picks it up as well -->
-    <!-- libcatalog for OS is copied  in the catalog" target (earlier than this) -->
+    <!-- libcatalog for both OS copied in the "catalog" target (earlier than this) -->
     <copy todir="${jar.nativelib.dir}">
         <fileset dir="${base.dir}/voltdb">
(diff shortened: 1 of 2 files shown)


build.xml
```diff
@@ -842,7 +840,7 @@ DISTRIBUTION
     <antcall target="call_strip_lib" />
 
     <!-- copy to dir where jar picks it up as well -->
-    <!-- libcatalog is copied to ${jar.nativelib.dir} in the ee target (earlier than this) -->
+    <!-- libcatalog for OS is copied  in the catalog" target (earlier than this) -->
```
yzhang1991:
```suggestion
    <!-- libcatalog for OS is copied  in the "catalog" target (earlier than this) -->
```

PR #6033, opened 2019-03-18
ENG-15587 Some CREATE VIEW statements just reply 'null' (in sqlcmd)

GuaGuaDean:
This issue is cause by queries that contains columns not existing in temp table. Fixed by adding an error-handling clause.


src/frontend/org/voltdb/planner/PlanningErrorException.java
```diff
@@ -23,4 +23,14 @@
...
     }
+    /**
+     * Create an exception with stacktrace erased. Used for reporting
+     * parsing/validation errors without the long backtrace list, to mimick
+     * user-observable error message from VoltCompiler.VoltCompilerException.
+     * @param msg exception message
+     * @param ignored not used
+     */
+    public PlanningErrorException(String msg, int ignored) {
+        super(msg, null, true, false);
```
Chaozz: `ignored` is unused?

kaikaizi: It is used to differentiate from the other constructor, to suppress stacktrace.

Chaozz: got it

PR #6041, opened 2019-03-20
ENG-15349 nibble export 

zeemanhe:
This pull request covers end-to-end of nibble export.  Use TTL mechanism to periodically move data from persistent table to its companion stream which will  transfer data to export target.  As soon as export acknowledges, the migrated rows in the persistent table will be deleted.


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -650,6 +655,10 @@ private void addDataSource(File adFile,
             final long genId) throws IOException {
         ExportDataSource source = new ExportDataSource(this, adFile, localPartitionsToSites, processor, genId);
         adFilePartitions.add(source.getPartitionId());
+        int migrateBatchSize = CatalogUtil.getPersistentMigrateBatchSize(source.getTableName());
+        if (migrateBatchSize > 0) {
+            source.setupMigrateRowsDeleter(migrateBatchSize);
```
n3world: setupMigrateRowsDeleter checks if the batchSize is > 0 so having the check here is redundant

src/frontend/org/voltdb/sysprocs/MigrateRowsBase.java
```diff
@@ -0,0 +1,234 @@
...
+            throw new VoltAbortException(
+                    "Chunk size must be positive, current value is" + chunksize);
+        }
+        Table catTable = ctx.getDatabase().getTables().getIgnoreCase(tableName);
+        if (catTable == null) {
+            throw new VoltAbortException(
+                    String.format("Table %s doesn't present in catalog", tableName));
+        }
+
+        if (replicated ^ catTable.getIsreplicated()) {
```
n3world: I think using `replicated != catTable.getIsreplicated()` is a clearer way to express this

src/frontend/org/voltdb/sysprocs/MigrateRowsNT.java
```diff
@@ -0,0 +1,260 @@
...
+        status.rowsToBeMoved = 0;
+        status.rowsJustMoved = 0;
+        int attempts = 1;
+        while (attempts <= attemptsLeft) {
+            ExportTask task = new ExportTask(attempts);
+            es.schedule(task, delay * attempts, TimeUnit.MILLISECONDS);
+            attempts++;
+        }
+        try {
+            latch.await(timeoutms, TimeUnit.MILLISECONDS);
```
n3world: If this times out what result is returned? It seems like it would be a success with rowMoved set but rowsToBeMoved and rowsJustMove as 0

src/frontend/org/voltdb/sysprocs/MigrateRowsNT.java
```diff
@@ -0,0 +1,260 @@
...
+            public ExportTask(int attempt) {
+                this.attempt = attempt;
+            }
+            @Override
+            public void run() {
+                NibbleStatus thisStatus = migrateRows(tableName, columnName, comparisonOp,
+                        value, chunksize, catTable.getIsreplicated());
+                if (!thisStatus.errorMessages.isEmpty()) {
+                    errors[attempt-1] = thisStatus.errorMessages;
+                    success.set(false);
```
n3world: When this occurs should any other attempts be cancelled and the result returned immediately? You could make this a Callable and have this throw an exception and then use the futures returned by the schedule method to check for errors and cancel any future attempts

src/ee/executors/migrateexecutor.cpp
```diff
@@ -43,41 +43,180 @@
...
+    setDMLCountOutputTable(executorVector.limits());
+
+    AbstractPlanNode *child = m_node->getChildren()[0];
+    ProjectionPlanNode *proj_node = NULL;
+    if (NULL == child) {
+        VOLT_ERROR("Attempted to initialize migrate executor with NULL child");
+        return false;
+    }
+
+    PlanNodeType pnt = child->getPlanNodeType();
```
wweiss-voltdb: I don't think you need to do any projection at all because the migrate executor should be the last node in the chain. Further, the migrate should not alter anything in the row other than the hidden column so we should not need to iterate over m_inputtargetmap in the execute path. Just update the hidden column.

src/ee/executors/migrateexecutor.cpp
```diff
@@ -43,41 +43,180 @@
...
+            BOOST_FOREACH(TableIndex *index, allIndexes) {
+                if (index->isMigratingIndex()) {
+                    indexesToUpdate.push_back(index);
+                }
+            }
+
+            assert(m_inputTuple.columnCount() == m_inputTable->columnCount());
+            assert(targetTuple.columnCount() == targetTable->columnCount());
+            TableIterator input_iterator = m_inputTable->iterator();
+            uint16_t migrateColumnIndex = targetTuple.getSchema()->hiddenColumnCount() -1;
```
wweiss-voltdb: I would put this in the init and make this an m_ because it will be the same column index for all rows that the migrate is called on.

zeemanhe:
@n3world will address your comments on a separate pull request

PR #6046, opened 2019-03-20
ENG-15614:

manjujames:
Use old way of looking at connectors to see if a table is a stream
because old snapshots didn't have tableType field.


src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -645,7 +645,19 @@ public static Index getPrimaryKeyIndex(Table catalogTable) throws Exception {
      */
     public static boolean isTableExportOnly(org.voltdb.catalog.Database database,
                                             org.voltdb.catalog.Table table) {
-        return TableType.isStream(table.getTabletype());
```
n3world: Why not use the tableType field if it exists and if not fall back to the old way?

PR #6058, opened 2019-03-22
ENG-15706: CRC was not getting reset in uncompressed mode, which resu…

manjujames:
…lted

in a CRC error. In addition, the buffer wasn't getting discarded on a CRC error,
which resulted in the error seen in the system test.

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -585,6 +585,7 @@ public class PBDRegularSegment extends PBDSegment {
                     }
                     retcont.b().flip();
                     if (checkCRC) {
+                        m_crc32.reset();
                         m_crc32.update(length);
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -596,6 +597,7 @@ public boolean allReadAndDiscarded() throws IOException {
                             openForWrite(false);
                             initNumEntries(m_objectReadIndex, m_bytesRead);
                             m_fc.truncate(m_readOffset);
+                            retcont.discard();
```
JadenW: Does the compressed path also need this discard on the ERROR case?

n3world: No, it hasn't allocated retcont before the crc check and it has a finally block to discard the buffer which it read the compressed data into.

PR #6061, opened 2019-03-25
PBD harden a bit more and handle check errors

n3world:
Add a length and CRC field in the header for the extra header
Remove checkCRC from the poll api since the only valid places to check crc are either parseAndTruncate and scanForGap
When a CRC error is found in a segment entry report back to the PBD the number of entries truncated
Only mark a segment final if all of the data has been syned to disk and there were no errors writing the data

relates to https://github.com/VoltDB/pro/pull/2538

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -29,6 +29,7 @@ import java.util.zip.CRC32;
 
 import org.voltcore.logging.VoltLogger;
 import org.voltcore.utils.DBBPool;
+import org.voltcore.utils.DBBPool.BBContainer;
 import org.voltcore.utils.DeferredSerialization;
 import org.voltdb.utils.BinaryDeque.OutputContainerFactory;
 
@@ -40,7 +41,8 @@ import com.google_voltpatches.common.base.Preconditions;
  * to insert an object that exceeds the remaining space is made. A segment can be used
  * for reading and writing, but not both at the same time.
  */
-public class PBDRegularSegment extends PBDSegment {
+class PBDRegularSegment extends PBDSegment {
+    private static final int VERSION = 2;
     private static final VoltLogger LOG = new VoltLogger("HOST");
     private final Map<String, SegmentReader> m_readCursors = new HashMap<>();
     private final Map<String, SegmentReader> m_closedCursors = new HashMap<>();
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -54,39 +62,41 @@
...
     private int m_extraHeaderSize = 0;
+    // Not guaranteed to be valid unless m_extraHeaderSize > 0
+    private int m_extraHeaderCrc = 0;
 
     private DBBPool.BBContainer m_segmentHeaderBuf = null;
     private DBBPool.BBContainer m_entryHeaderBuf = null;
     Boolean INJECT_PBD_CHECKSUM_ERROR = Boolean.getBoolean("INJECT_PBD_CHECKSUM_ERROR");
 
-    public PBDRegularSegment(Long index, long id, File file) {
+    PBDRegularSegment(Long index, long id, File file) {
```
rdykiel: I think I made these public for E3 but I'm not sure to recall. I may have to do this again.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -97,61 +107,71 @@ public void reset()
             m_entryHeaderBuf.discard();
             m_entryHeaderBuf = null;
         }
-        m_segmentHeaderCRC.reset();
-        m_entryCRC.reset();
+        m_crc.reset();
+        m_crc.reset();
```
rdykiel: Duplicated line (probably global replace)

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -631,46 +670,93 @@ public void discard() {
...
+                            + "Truncate the file to last safe point.");
+                    truncateToCurrentReadIndex();
+                    return false;
+                }
+            }
+
+            entry.position(origPosition);
+
+            if (checkCrc) {
+                m_crc32.reset();
```
rdykiel: m_crc32 could be renamed m_crcReader so we know what it is.

src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1237,32 +1216,27 @@ public synchronized void setAwaitingTruncation(boolean m_awaitingTruncation)
     }
 
     @Override
-    public ExportSequenceNumberTracker scanForGap(BinaryDequeScanner scaner) throws IOException
+    public void scanEntries(BinaryDequeScanner scaner) throws IOException
```
rdykiel: "scanner" :-)

tests/frontend/org/voltdb/utils/TestPBDMultipleReaders.java
```diff
@@ -39,7 +39,7 @@
 
     private final static VoltLogger logger = new VoltLogger("EXPORT");
 
-    private static final int s_segmentFillCount = 47;
+    private static final int s_segmentFillCount = 31;
```
rdykiel: It might be useful to have a comment explaining the value for the poor maintainers following.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
n3world: Yeah just a find and replace. This reset is not actually even really needed since the object is reset each time before use anyways and it isn't a running crc.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
n3world: PBDRegularSegment is a package protected class so only package member can access the constructor. Having a constructor with a higher level of access than the class doesn't make much sense.

PR #6063, opened 2019-03-25
ENG-15606, remove assertion generating backtrace in tests

rdykiel:
Why was this assert put in? It seems we now can get gap queries when our gap tracker is empty.


luyangco:
The assert is now out-of-date, which should be removed.

PR #6064, opened 2019-03-26
Eng 15711 quiesec sigsegv

luyangco:


src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1482,13 +1487,12 @@ void VoltDBEngine::attachTupleStream(StreamedTable* streamedTable,
...
-    ExportTupleStream *wrapper = m_exportingStreams[streamName];
+    ExportTupleStream* wrapper = streamedTable->getWrapper();
     if (wrapper == NULL) {
         wrapper = new ExportTupleStream(m_executorContext->m_partitionId,
                                         m_executorContext->m_siteId,
                                         timestamp,
                                         streamName);
-        m_exportingStreams[streamName] = wrapper;
         streamedTable->setWrapper(wrapper);
         VOLT_TRACE("created stream export wrapper stream %s", name.c_str());
```
rdykiel: Please fix this, "name" is undefined AFAIK (saw this yesterday while debugging)

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -2248,13 +2251,15 @@ void VoltDBEngine::tick(int64_t timeInMillis, int64_t lastCommittedSpHandle) {
 /** Bring the Export and DR system to a steady state with no pending committed data */
 void VoltDBEngine::quiesce(int64_t lastCommittedSpHandle) {
     m_executorContext->setupForQuiesce(lastCommittedSpHandle);
-    BOOST_FOREACH (auto exportStream, m_exportingStreams) {
+    for (auto const &streamTable : m_exportingTables) {
```
rdykiel: Why don't we use BOOST_FOREACH here? (I'm not familiar with the tradeoffs)

src/ee/storage/TableCatalogDelegate.cpp
```diff
@@ -743,7 +743,10 @@ TableCatalogDelegate::processSchemaChanges(catalog::Database const& catalogDatab
             int64_t seqNo;
             size_t streamBytesUsed;
             existingStreamedTable->getExportStreamPositions(seqNo, streamBytesUsed);
-            ExportTupleStream* wrapper = new ExportTupleStream(*existingStreamedTable->getWrapper());
+            ExportTupleStream* wrapper = existingStreamedTable->getWrapper();
```
rdykiel: Why don't we check against wrapper == NULL? It's done everywhere else in the code...

src/ee/execution/VoltDBEngine.cpp
luyangco: c++11 has built-in range based for-loop support to iterate the map, which is equivalent to BOOST_FOREACH. Though the performance difference is minor, there is a general movement to replace BOOST_FOREACH with the built-in for-loop because 1) clarity 2) compilation speed, as BOOST_FOREACH does a lot of magic under the hood.

src/ee/storage/TableCatalogDelegate.cpp
luyangco: Good catch. I'll include the null checker and variable name fix into a followup commit.

src/ee/execution/VoltDBEngine.cpp
luyangco: Renamed to streamName.

PR #6066, opened 2019-03-26
ENG-15721:

wweiss-voltdb:
Reset back to NULL any rows that were marked as migrating in a Snapshot when that row is being Restored (as opposed to recovered) from a Snapshot.

Updated the unit test (in Pro) that verifies the row counts of non-migrating rows in the face of Joins, Rejoins, Recover, and Restore. Note that this test will break when and if the migrate index is implicitly created by the compiler as it does today for the primary key column(s) because it explicitly removes the index to hold the migrating row count constant.

### DIFF
--- a/tests/frontend/org/voltdb/sysprocs/saverestore/TestSavedTableConverter.java
+++ b/tests/frontend/org/voltdb/sysprocs/saverestore/TestSavedTableConverter.java
@@ -437,6 +437,169 @@ public class TestSavedTableConverter extends TestCase {
         assertFalse(SavedTableConverter.needsConversion(inputTableWithViewCol, m_catalogTable, false, true, false, true));
     }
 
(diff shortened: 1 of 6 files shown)


tests/frontend/org/voltdb/sysprocs/saverestore/TestSavedTableConverter.java
```diff
@@ -164,7 +165,7 @@ private void runNumberConversion(String in, String out, VoltType tin, VoltType t
         boolean failed = false;
         try {
             result = SavedTableConverter.convertTable(inputTable,
-                    catalogTable, false, false, false);
+                    catalogTable, false, false, false, true);
```
rdykiel: I notice that we only test the new flag value to "true" but we don't exercise all the combinations between flags that the code allegedly supports. Shouldn't we make it a priority to add these test cases?

PR #6067, opened 2019-03-26
PBD harden a bit more and handle check errors

n3world:
Add a length and CRC field in the header for the extra header
Remove checkCRC from the poll api since the only valid places to check crc are either parseAndTruncate and scanForGap
When a CRC error is found in a segment entry report back to the PBD the number of entries truncated
Only mark a segment final if all of the data has been syned to disk and there were no errors writing the data

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -79,6 +79,8 @@ public abstract class PBDSegment {
     protected CRC32 m_crc;
     // Mirror of the isFinal metadata on the filesystem
     private boolean m_isFinal;
+    // Whether or not this is the current active segment being written to
+    boolean m_isActive = false;
 
     public PBDSegment(File file)
     {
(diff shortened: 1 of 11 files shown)


src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -212,105 +253,146 @@ int parseAndTruncate(BinaryDeque.BinaryDequeTruncator truncator) throws IOExcept
...
-    public static void setFinal(File file, boolean isFinal) {
+    public static boolean setFinal(File file, boolean isFinal) {
 
         try {
             UserDefinedFileAttributeView view = getFileAttributeView(file);
             if (view != null) {
-                view.write(IS_FINAL_ATTRIBUTE, Charset.defaultCharset().encode(new Boolean(isFinal).toString()));
+                view.write(IS_FINAL_ATTRIBUTE, Charset.defaultCharset().encode(Boolean.toString(isFinal)));
             }
+            return true;
```
luyangco: UserDefinedFileAttributeView is always null on Mac, even though HFS+/APFS fully supports xattr. It leaves inconsistency between m_isFinal and isFinal(file) method, which also means on mac all pbd readers are forced to do checksum check in constructor.

n3world: I meant to put that return in the if block so it only returns true when the write is done, but thinking about it more as far as this segment is concerned it is final just the system couldn't make it official by writing out the attribute.

The limitation on Mac is a problem is an issue with this implementation design for final. I am not sure it is a big deal since since Mac is not really supported for production.

PR #6070, opened 2019-03-26
ENG-15349 delete or update on migrated rows for TTL export

zeemanhe:
Migrated rows in persistent tables can be updated or deleted via AdHoc queries.  In these cases, the migrating indexes must be tracked correctly. The pull request uses unit tests with connector disabling and enabling, via catalog update,  to simulate various scenario and verify all the migrated rows which should be accounted for.  The pull request also fixed catalog update for migrated streams.


src/frontend/org/voltdb/export/ExportManager.java
```diff
@@ -792,10 +792,8 @@ public void clientInterfaceStarted(ClientInterface clientInterface) {
         m_ci.bindAdapter(m_adapter, null);
     }
 
-    public void invokeMigrateRowsDelete(StoredProcedureInvocation spi, int partition, SimpleClientResponseAdapter.Callback cb) {
-        Long handle = m_adapter.registerCallback(cb);
-        spi.setClientHandle(handle);
-        m_ci.createTransaction(m_adapter.connectionId(), spi, false, true, false, partition, spi.getSerializedSize(),
-                System.nanoTime());
+    public void invokeMigrateRowsDelete(int partition, String tableName, int batchSize, long deletableTxnId,  ProcedureCallback cb) {
+        m_ci.getDispatcher().getInternelAdapterNT().callProcedure(m_ci.getInternalUser(), true, 1000 * 120, cb,
```
rdykiel: What is tthis "1000 * 120" hard coded value and should it stay hard coded? Define a constant and comment.

src/frontend/org/voltdb/export/MigrateRowsDeleter.java
```diff
@@ -17,103 +17,46 @@
...
+                                m_tableName, resp.getStatusString()));
+                    }
+                    latch.countDown();
+                }
+            };
+            ExportManager.instance().invokeMigrateRowsDelete(m_partitionId, m_tableName, m_batchSize, deletableTxnId, cb);
             try {
-                StoredProcedureInvocation serializedSPI = MiscUtils.roundTripForCL(m_deleterSPI);
-                ExportManager.instance().invokeMigrateRowsDelete(serializedSPI, m_partitionId, new DeleterCB(deletableTxnId));
+                latch.await(1, TimeUnit.MINUTES);
```
rdykiel: Another hard coded constant... Should we have them configurable?

src/frontend/org/voltdb/jni/ExecutionEngineJNI.java
```diff
@@ -698,16 +697,16 @@ public void exportAction(boolean syncAction,
...
         m_nextDeserializer.clear();
-        boolean txnFullyDeleted = nativeDeleteMigratedRows(pointer, txnid, spHandle, uniqueId,
+        int txnFullyDeleted = nativeDeleteMigratedRows(pointer, txnid, spHandle, uniqueId,
                 getStringBytes(tableName), deletableTxnId, maxRowCount, undoToken);
-        if (EXPORT_LOG.isDebugEnabled()) {
-            EXPORT_LOG.debug("deleteMigrated rows from table " + tableName +
-                    " up to txnId: " + TxnEgo.txnIdToString(deletableTxnId) +
-                    (txnFullyDeleted ? " FULLY processed" : " PARTIALLY processed"));
-        }
+//        if (EXPORT_LOG.isDebugEnabled()) {
```
rdykiel: Why not remove this code

src/frontend/org/voltdb/sysprocs/SwapTables.java
```diff
@@ -41,6 +42,13 @@
         List<String> sqlStatements = new ArrayList<>(1);
         sqlStatements.add(sql);
 
+        for (Table t: context.database.getTables()) {
+            if (t.getTypeName().equalsIgnoreCase(theTable) && t.getTimetolive() != null) {
+                return makeQuickResponse(ClientResponse.GRACEFUL_FAILURE,
+                        String.format("Table %s cannot be swaped since it uses TTL.",theTable));
```
rdykiel: Typo , should be "swapped"

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -962,7 +962,7 @@ VoltDBEngine::processCatalogDeletes(int64_t timestamp, bool updateReplicated,
                 // FIXME: factorize deletion logic in common method
                 auto streamedtable = persistenttable->getStreamedTable();
                 if (streamedtable) {
-                    VOLT_TRACE("delete a streamed companion wrapper for %s", tcd->getTable()->name().c_str());
+                    VOLT_DEBUG("delete a streamed companion wrapper for %s", tcd->getTable()->name().c_str());
```
wweiss-voltdb: I think this is fine for now but in the future we should get used to using the HOST logger for all catalog update functions that need to be instrumented, so that the logs can be read in conjunction with java logs. VOLT_????? logs are only useful when you run a single host and you have direct access to the console or are running IPC.

src/ee/storage/persistenttable.cpp
```diff
@@ -1004,14 +1020,15 @@ void PersistentTable::updateTupleWithSpecificIndexes(TableTuple& targetTupleToUp
...
-    ExecutorContext* ec = ExecutorContext::getExecutorContext();
     if (hasDRTimestampColumn() && updateDRTimestamp && !updateMigrate) {
         setDRTimestampForTuple(ec, sourceTupleWithNewValues, true);
     }
 
-    if (updateMigrate && m_shadowStream != nullptr) {
-        setMigrateTxnIdForTuple(ec, sourceTupleWithNewValues);
+    // Set or reset txnId on migrate hidden column
+    if (m_shadowStream != nullptr) {
+        setMigrateTxnIdForTuple(ec, targetTupleToUpdate, sourceTupleWithNewValues);
```
wweiss-voltdb: This does not seem right to me. If a row has not been migrated, it's hidden column will be null. But the code in setMigrateTxnIdForTuple says that if the column is null set the TxnId in the hidden column. Doesn't that mean that all updates will invert the hidden column even for rows that have not yet been migrated?

src/ee/storage/persistenttable.cpp
```diff
@@ -2435,7 +2453,27 @@ bool PersistentTable::migratingRemove(int64_t txnId, TableTuple& tuple) {
     return found == 1;
 }
 
-bool PersistentTable::deleteMigratedRows(int64_t deletableTxnId, int32_t maxRowCount) {
+void PersistentTable::migratingUndo(TableTuple& tuple) {
+    if (m_shadowStream != nullptr) {
+        NValue txnId = tuple.getHiddenNValue(getMigrateColumnIndex());
+        if (!txnId.isNull()) {
```
wweiss-voltdb: This does not seem right to me either. We should only need to add a tuple back to the hidden index if the tuple's hidden column was reset AND it was previously set. But if it is currently set to null, it could more likely be that the row was never migrated rather then that the row was migrated and then updated.

src/ee/execution/VoltDBEngine.cpp
```diff
@@ -1285,6 +1285,12 @@ VoltDBEngine::processCatalogAdditions(int64_t timestamp, bool updateReplicated,
...
                 // update exporting tables with new stream
                 StreamedTable* stream = tcd->getStreamedTable();
+                if (!stream) {
+                   PersistentTable *persistenttable = dynamic_cast<PersistentTable*>(tcd->getTable());
+                   if (persistenttable) {
+                       stream = persistenttable->getStreamedTable();
+                   }
+                }
                 if (stream) {
                     m_exportingTables[stream->name()] = stream;
```
wweiss-voltdb: add assert that there is no purgestream with the same name

src/ee/storage/persistenttable.cpp
```diff
@@ -1004,17 +1006,26 @@ void PersistentTable::updateTupleWithSpecificIndexes(TableTuple& targetTupleToUp
...
+    if (m_shadowStream != nullptr) {
+       uint16_t migrateColumnIndex = getMigrateColumnIndex();
+       NValue txnId = sourceTupleWithNewValues.getHiddenNValue(migrateColumnIndex);
+       if (txnId.isNull()) {
+           if (fromMigrate) {
+               int64_t spHandle = ec->currentSpHandle();
+               sourceTupleWithNewValues.setHiddenNValue(migrateColumnIndex, ValueFactory::getBigIntValue(spHandle));
+           }
+       } else {
+           sourceTupleWithNewValues.setHiddenNValue(migrateColumnIndex, NValue::getNullValue(VALUE_TYPE_BIGINT));
```
wweiss-voltdb: add assert not fromMigrate

PR #6071, opened 2019-03-26
ENG-15711:

wweiss-voltdb:
Fix Cpp Unit tests that were using a test function but that is now used in other paths and therefore renamed.


luyangco:
I shouldn't trust Eclipse, I thought refactor-> rename should do the work for me ;-)

PR #6076, opened 2019-03-27
PBD quarantine unreadable

n3world:
Add an optional flag encoded into a segment name to mark the segment as quarantined
If there is an IOError reading a segment header or crc checksum fails mark the segment as quarantined. If all readers poll beyond the quarantined segment delete it.
Add a little more paranoia around the isFinal flag

### DIFF
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -89,13 +89,12 @@ class PBDRegularSegment extends PBDSegment {
     }
 
     @Override
-    long segmentId() {
+    public long segmentId() {
         return m_id;
     }
 
     @Override
-    File file()
-    {
+    public File file() {
         return m_file;
     }
 
--- a/src/frontend/org/voltdb/utils/PBDRegularSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDRegularSegment.java
@@ -614,11 +614,8 @@ class PBDRegularSegment extends PBDSegment {
                 //Get the length and size prefix and then read the object
                 ByteBuffer b = m_entryHeaderBuf.b();
                 b.clear();
-                while (b.hasRemaining()) {
-                    int read = m_fc.read(b);
-                    if (read == -1) {
-                        throw new EOFException();
-                    }
+                if (!read(b, checkCRC)) {
+                    return null;
                 }
                 b.flip();
                 final int entryCRC = b.getInt();
@@ -713,17 +710,8 @@ class PBDRegularSegment extends PBDSegment {
         private boolean fillBuffer(ByteBuffer entry, int entryId, char flags, int crc, boolean checkCrc)
                 throws IOException {
             int origPosition = entry.position();
-            while (entry.hasRemaining()) {
-                int read = m_fc.read(entry);
-                if (read == -1) {
-                    if (!checkCrc) {
-                        throw new EOFException();
-                    }
-                    LOG.warn("File corruption detected in " + m_file.getName() + ": invalid entry length. "
-                            + "Truncate the file to last safe point.");
-                    truncateToCurrentReadIndex();
-                    return false;
-                }
+            if (!read(entry, checkCrc)) {
+                return false;
             }
 
             entry.position(origPosition);
@@ -739,6 +727,32 @@ class PBDRegularSegment extends PBDSegment {
                 entry.position(origPosition);
             }
 
+            return true;
+        }
+
+        private boolean read(ByteBuffer buffer, boolean canTruncate) throws IOException {
+            do {
+                try {
+                    int read = m_fc.read(buffer);
+                    if (read == -1) {
+                        String message = "EOF encountered reading " + m_file + " at position " + m_fc.position()
+                                + " expected to be able to read " + buffer.remaining() + " more bytes";
+                        if (canTruncate) {
+                            LOG.warn(message);
+                            truncateToCurrentReadIndex();
+                            return false;
+                        }
+                        throw new EOFException(message);
+                    }
+                } catch (IOException e) {
+                    if (canTruncate) {
+                        LOG.warn("Error encountered reading: " + m_file, e);
+                        truncateToCurrentReadIndex();
+                        return false;
+                    }
+                    throw new IOException("Error encountered reading: " + m_file, e);
+                }
+            } while (buffer.hasRemaining());
             return true;
         }
 
--- a/src/frontend/org/voltdb/utils/PBDSegment.java
+++ b/src/frontend/org/voltdb/utils/PBDSegment.java
@@ -172,9 +172,11 @@ public abstract class PBDSegment {
 
     /**
      * Parse the segment and truncate the file if necessary.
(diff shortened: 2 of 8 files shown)


src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1230,11 +1346,22 @@ public void scanEntries(BinaryDequeScanner scanner) throws IOException
...
+
         /*
          * Iterator all the objects in all the segments and pass them to the scanner
          */
         for (PBDSegment segment : m_segments.values()) {
-            m_numObjects -= segment.scan(scanner);
+            try {
+                m_numObjects -= segment.scan(scanner);
+            } catch (IOException e) {
+                LOG.warn("Error scanning segment: " + segment.file() + ". Qurantining segment.");
```
rdykiel: Typo

tests/frontend/org/voltdb/utils/TestPersistentBinaryDeque.java
```diff
@@ -1238,122 +1233,7 @@ public void cancel() {}
...
-    }
-
-    @SuppressWarnings("unchecked")
-    private NavigableMap<Long, PBDSegment> getSegmentMap() throws IllegalArgumentException, IllegalAccessException {
-        return ((NavigableMap<Long, PBDSegment>) FieldUtils
-                .getDeclaredField(PersistentBinaryDeque.class, "m_segments", true).get(m_pbd));
-    }
-
-    private BBContainer pollOnceWithoutDiscard(BinaryDequeReader reader) throws IOException {
+    static BBContainer pollOnceWithoutDiscard(BinaryDequeReader reader) throws IOException {
```
rdykiel: I can't check but am wondering why all these test cases have become irrelevant

tests/frontend/org/voltdb/utils/TestPersistentBinaryDequeCorruption.java
```diff
@@ -0,0 +1,270 @@
+/* This file is part of VoltDB.
```
rdykiel: I must say I hate to see a file totally devoid of comments explaining the developer's (or test case) intent

tests/frontend/org/voltdb/utils/TestPersistentBinaryDeque.java
n3world: They didn't they were all moved to TestPersistentBinaryDequeCorruption

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -615,11 +615,8 @@ public boolean allReadAndDiscarded() {
                 //Get the length and size prefix and then read the object
                 ByteBuffer b = m_entryHeaderBuf.b();
                 b.clear();
-                while (b.hasRemaining()) {
-                    int read = m_fc.read(b);
-                    if (read == -1) {
-                        throw new EOFException();
-                    }
+                if (!read(b, checkCRC)) {
```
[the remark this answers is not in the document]
n3world: I should probably rename that argument. Currently the only time the caller can handle a segment auto truncating is when checkCrc is true. They could be split so there is a checkCrc and a canTruncate argument but if checkCrc is true canTruncate is also true. Currently checkCrc == canTruncate so there is only one variable

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -743,6 +731,31 @@ private boolean fillBuffer(ByteBuffer entry, int entryId, char flags, int crc, b
...
+                            LOG.warn(message);
+                            truncateToCurrentReadIndex();
+                            return false;
+                        }
+                        throw new EOFException(message);
+                    }
+                } catch (IOException e) {
+                    if (canTruncate) {
+                        LOG.warn("Error encountered reading: " + m_file, e);
+                        return false;
```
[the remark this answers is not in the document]
n3world: A bug

src/frontend/org/voltdb/utils/PBDSegment.java
```diff
@@ -338,8 +338,11 @@ void setFinal(boolean isFinal) throws IOException {
                     // It is dangerous to leave final on a segment so make sure the metadata is flushed
                     m_fc.force(true);
                 }
-                m_isFinal = isFinal;
+            } else if (isFinal(m_file) && !isFinal) {
+                throw new IOException("Could not remove the final attribute from " + m_file.getName());
             }
+            // It is OK for m_isFinal to be true when isFinal(File) returns false but not the other way
```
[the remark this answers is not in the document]
n3world: If m_isFinal false and the program wants to mark it final it is OK for the in memory variable for isFinal to be true since the voltdb will behave like it is final and not modify regardless of what the flag on the file. When the db restarts it will see that it isn't final and go through the checks to make sure it is OK. There is no harm in the running program in treating a known good segment as final even though it cannot set the flag on the filesystem

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -743,6 +731,31 @@ private boolean fillBuffer(ByteBuffer entry, int entryId, char flags, int crc, b
...
+                    int read = m_fc.read(buffer);
+                    if (read == -1) {
+                        String message = "EOF encountered reading " + m_file + " at position " + m_fc.position()
+                                + " expected to be able to read " + buffer.remaining() + " more bytes";
+                        if (canTruncate) {
+                            LOG.warn(message);
+                            truncateToCurrentReadIndex();
+                            return false;
+                        }
+                        throw new EOFException(message);
```
luyangco: This doesn't look right to me, `canTruncate == false` means file attribute marks as final but at this line it means file is actually corrupted(something horribly wrong), should we try to truncate the file to last safe point regardless the `canTruncate` flag?

src/frontend/org/voltdb/utils/PBDRegularSegment.java
```diff
@@ -714,17 +711,8 @@ private void handleCorruptHeader(String message, boolean checkCrc) throws IOExce
         private boolean fillBuffer(ByteBuffer entry, int entryId, char flags, int crc, boolean checkCrc)
                 throws IOException {
             int origPosition = entry.position();
-            while (entry.hasRemaining()) {
```
luyangco: Github disallows me from commenting on lines outside changeset. 
I want to discuss line https://github.com/VoltDB/voltdb/pull/6076/files#diff-5cfd6961fecf46d96a41f7ef1103726cR630, if header is corrupted, the segment can be marked as quarantined which will be deleted later regardless the `checkCrc` flag.

src/frontend/org/voltdb/utils/PBDRegularSegment.java
n3world: No because that breaks PersistentBinaryDeque. Currently the PersistentBinaryDeque can only handle a segment auto truncating itself during scanEntries or parseAndTruncate. Outside of those calls an auto truncate will cause  PersistentBinaryDeque to have incorrect state.

n3world:
Moved to https://github.com/VoltDB/voltdb/pull/6091

PR #6077, opened 2019-03-27
ENG-15337:

wweiss-voltdb:
There are some tests that wait for tuple counts to go to 0, in the Export Statistics after the Export Stream is dropped. However, when the stream is dropped the EDS and PBD is immediately removed and therefore the stats will have no rows for the Stream. The tests that look for rows have been updated to look for no rows instead.

### DIFF
--- a/tests/frontend/org/voltdb/export/TestExportBaseSocketExport.java
+++ b/tests/frontend/org/voltdb/export/TestExportBaseSocketExport.java
@@ -306,7 +306,6 @@ public class TestExportBaseSocketExport extends RegressionSuite {
         // Wait 10 mins only
         long end = System.currentTimeMillis() + (10 * 60 * 1000);
         while (true) {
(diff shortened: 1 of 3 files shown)


tests/frontend/org/voltdb/export/TestExportBaseSocketExport.java
```diff
@@ -361,6 +361,63 @@ public static void waitForStreamedTargetAllocatedMemoryZero(Client client) throw
...
+        quiesce(client);
+        System.out.println("Quiesce done....");
+
+        VoltTable stats = null;
+        long ftime = 0;
+        long st = System.currentTimeMillis();
+        // Wait 10 mins only
+        long end = System.currentTimeMillis() + (10 * 60 * 1000);
+        while (true) {
+            stats = client.callProcedure("@Statistics", "export", 0).getResults()[0];
```
zeemanhe: can move the stats pulling after timeout check

PR #6082, opened 2019-03-28
ENG-15730: scripts to turn on/off Calcite debug output

Chaozz:
```console
# turn off calcite debug output
tools/calcite/debug-off 

# turn on calcite debug output
tools/calcite/debug-on
```
1. you don't need to recompile after running the script, the change is taken on the fly
2. the debug is **off** by default, so don't forget to turn it off before commiting to git.


Chaozz:
> Does this turn on/off command-console debugging or IDE debugging?
> It could make more sense to combine 2 scripts into one, used like `calcite-debug on/off`

PR #6090, opened 2019-03-29
ENG-15616 Fixed Crash using CTE with Geo Index

GuaGuaDean:
The root cause of this crash is that the query tries to access the index on a polygon based column, which should be only used for geo related query like "geometry contains point". This solution adjusted the discount and penalty when calculating plan cost and thus prevented geo index being used by non-geo queries.


src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
```diff
@@ -668,9 +668,16 @@ public void computeCostEstimates(long unusedChildOutputTupleCountEstimate,
...
+            // otherwise geo index can't be used and thus will be given a penalty, worse than seq scan
             if (m_catalogIndex.getType() == IndexType.COVERING_CELL_INDEX.getValue()) {
-                final double GEO_INDEX_ARTIFICIAL_TUPLE_DISCOUNT_FACTOR = 0.08;
-                tuplesToRead *= GEO_INDEX_ARTIFICIAL_TUPLE_DISCOUNT_FACTOR;
+                if (m_lookupType == IndexLookupType.GEO_CONTAINS) {
+                    final double GEO_INDEX_ARTIFICIAL_TUPLE_DISCOUNT_FACTOR = 0.08;
+                    tuplesToRead *= GEO_INDEX_ARTIFICIAL_TUPLE_DISCOUNT_FACTOR;
+                } else {
+                    final double GEO_INDEX_ARTIFICIAL_TUPLE_PENALTY_FACTOR = 10.0;
+                    tuplesToRead *= GEO_INDEX_ARTIFICIAL_TUPLE_PENALTY_FACTOR;
```
Chaozz: I would like to set the cost to **MAX** since we don't want it to be picked.

kaikaizi:
Can you add tests?

src/frontend/org/voltdb/plannodes/IndexScanPlanNode.java
GuaGuaDean: The cost is calculated in a recursive way, so a maximal value might cause integer overflow

PR #6093, opened 2019-04-01
ENG-15740, bug fix, when stream is dropped it's PBDs are also removed…

luyangco:
…, but for rejoined node with old PBDs of dropped stream it causes trouble because dropped stream doesn't exist in catalog, PBD doesn't know how to generate schema header for it.

During initialization ExportGeneration shouldn't create streams that doesn't exist in catalog, delete underlying PBD files of those dropped streams.

Change-Id: I75ba5da97e8310b564deb2139b8178ea326d2a3d

Pro side: https://github.com/VoltDB/pro/pull/2554

### DIFF
--- a/src/frontend/org/voltdb/export/ExportGeneration.java
+++ b/src/frontend/org/voltdb/export/ExportGeneration.java
@@ -134,7 +134,7 @@ public class ExportGeneration implements Generation {
     {
         File files[] = exportOverflowDirectory.listFiles();
         if (files != null) {
-            initializeGenerationFromDisk(messenger, processor, files, localPartitionsToSites, catalogContext.m_genId);
+            initializeGenerationFromDisk(connectors, messenger, processor, files, localPartitionsToSites, catalogContext.m_genId);
         }
         initializeGenerationFromCatalog(catalogContext, connectors, processor, hostId, messenger, localPartitionsToSites);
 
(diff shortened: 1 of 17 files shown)


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -145,18 +145,30 @@ void initialize(HostMessenger messenger,
...
-    private void initializeGenerationFromDisk(HostMessenger messenger,
+    private void initializeGenerationFromDisk(final CatalogMap<Connector> connectors,
+            HostMessenger messenger,
             final ExportDataProcessor processor,
             File[] files, List<Pair<Integer, Integer>> localPartitionsToSites,
             long genId) {
 
         List<Integer> onDiskPartitions = new ArrayList<Integer>();
-
+        Set<String> exportedTables = new HashSet<>();
```
rdykiel: Please factorize this logic into a method as it is used in 2 places.

tests/frontend/org/voltdb/export/ExportLocalClusterBase.java
```diff
@@ -96,4 +98,27 @@ public Client getClient(LocalCluster cluster) throws IOException {
         }
         return client;
     }
+
+    static void resetDir() throws IOException {
+        File f = new File("/tmp/" + System.getProperty("user.name"));
```
n3world: This directory should probably be defined in a common place so if we want to move it to another location it is a single update

luyangco:
@rdykiel and @n3world  I addressed all your comments.

PR #6095, opened 2019-04-01
ENG-15714:

manjujames:
Make the default table type of 0 an invalid value.
This way we can distinguish between unset table types
from tables from older version snapshots and tables that
are actually persistent tables.


src/frontend/org/voltdb/utils/CatalogUtil.java
```diff
@@ -645,19 +645,24 @@ public static Index getPrimaryKeyIndex(Table catalogTable) throws Exception {
      */
     public static boolean isTableExportOnly(org.voltdb.catalog.Database database,
```
n3world: Should this method be renamed to isStream so it matches the table type names?

manjujames: Yes, probably makes sense, but I will leave it for now.

PR #6102, opened 2019-04-01
ENG-15747: Fail license check if node count exceeds number specified …

manjujames:
…in license, for enterprise license as well.

### DIFF
--- a/src/frontend/org/voltdb/utils/MiscUtils.java
+++ b/src/frontend/org/voltdb/utils/MiscUtils.java
@@ -383,18 +383,10 @@ public class MiscUtils {
 
         // check node count
         if (licenseApi.maxHostcount() < numberOfNodes) {
-            // Enterprise gets a pass on this one for now
-            if (licenseApi.isEnterprise()) {
-                hostLog.error("Warning, VoltDB commercial license for " + licenseApi.maxHostcount() +
-                        " nodes, starting cluster with " + numberOfNodes + " nodes.");
-                valid = false;
-            }
-            // Trial, Pro & AWS licenses have a hard enforced limit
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/utils/MiscUtils.java
```diff
@@ -383,18 +383,10 @@ public static boolean validateLicense(LicenseApi licenseApi,
...
-                        " nodes, starting cluster with " + numberOfNodes + " nodes.");
-                valid = false;
-            }
-            // Trial, Pro & AWS licenses have a hard enforced limit
-            else {
-                hostLog.fatal("Warning, VoltDB license for a " + licenseApi.maxHostcount() + " node " +
-                        "attempted for use with a " + numberOfNodes + " node cluster.");
-                return false;
-            }
+            hostLog.fatal("Attempting to start with too many nodes (" + numberOfNodes + "). " +
```
JadenW: This seems enforce on the start and recovery cases, but not in the elastic add case.
So there might be a case that user will be able to grow the cluster to number of hosts exceeding licensed, but will not be able to recovery afterwards. Is this an expected behavior?
Also, those enforcements are in clear code?

manjujames: That's a good point. I will check with Ruth.

manjujames: Yes, the enforcements are in clear code as of now.

manjujames: Ruth opened a separate ticket to handle elastic add case. So I will merge this one.

PR #6104, opened 2019-04-02
ENG-15746, add debug message to capture future NPE in quiesce(). Fix …

luyangco:
…a bug that stream name containing underscore may not be initialized from disk.

Change-Id: Ic911dead39f22b896ce604bed297255be3968cfb

### DIFF
--- a/src/frontend/org/voltdb/export/ExportGeneration.java
+++ b/src/frontend/org/voltdb/export/ExportGeneration.java
@@ -1101,7 +1101,8 @@ public class ExportGeneration implements Generation {
 
     // Naming convention for export pdb file: [table name]_[partition]_[segmentId]_[prevId].pdb,
     private static String getStreamNameFromNonce(String nonce) {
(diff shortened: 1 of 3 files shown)


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -1101,7 +1101,8 @@ public void processStreamControl(String exportSource, List<String> exportTargets
 
     // Naming convention for export pdb file: [table name]_[partition]_[segmentId]_[prevId].pdb,
     private static String getStreamNameFromNonce(String nonce) {
-        return nonce.substring(0, nonce.indexOf('_'));
+        // it's possible the stream name contains underscore
```
rdykiel: I would be more reassured it works well if I saw JUnit test cases...

luyangco: I'll add it in follow-up pull request.

PR #6105, opened 2019-04-02
ENG-15288: Print more details when krb login fails

n3world:
If the servers login to kerberos fails there is not much information printed.
Always add the configuration entry name and if possible some of the
configuration entry information.


src/frontend/org/voltdb/AuthSystem.java
```diff
@@ -516,6 +519,28 @@ private void finish() {
         }
     }
 
+    private static String getLoginConfigurationInfo(String prefix) {
```
manjujames: This could even call VoltDB.crashGlobalVoltDB since all the callers use this to crash global in the same exact way.

PR #6107, opened 2019-04-02
Added VoltRelOptCost

alexeevm:
Introduced VoltRelOptCost which mimics Calcite's VolcanoCost but throws an OverflowException if cost is about to overflow.

Attempt to propagate an error message thrown by the Calcite all the way to a ClientResponse


Chaozz:
I have run this branch in Jenkins and the test results look good except some trivial failure on `TestInlineRules` due to the changes in `VoltPhysicalRel`. 
I have updated the `VoltPhysicalRel` and I think it is ready to merge.

PR #6117, opened 2019-04-05
Aggregate distribution

kaikaizi:
Update distribution trait of aggregation:
Set distribution trait to SINGLETON (corresponding to replicated table) only if all the aggregation function appeared in the aggregation node refer to the partition column(s) of the child distribution.

### DIFF
--- a/src/frontend/org/voltdb/plannerv2/rel/logical/VoltLogicalAggregate.java
+++ b/src/frontend/org/voltdb/plannerv2/rel/logical/VoltLogicalAggregate.java
@@ -17,10 +17,16 @@
 
 package org.voltdb.plannerv2.rel.logical;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


src/frontend/org/voltdb/plannerv2/rel/logical/VoltLogicalAggregate.java
```diff
@@ -49,23 +55,45 @@
...
+                final Set<Integer> partCols = new HashSet<>(childDist.getKeys());
+                if (aggs.stream().flatMap(call -> {
+                    assert call.getArgList().size() <= 1;
+                    if (call.getArgList().isEmpty()) {  // COUNT removes argument
+                        return Stream.empty();
+                    } else {
+                        return Stream.of(call.getArgList().get(0));
+                    }
+                }).allMatch(partCols::contains)) {
+                    return traitSet.replace(RelDistributions.SINGLETON.with(null, true));
```
Chaozz: It seems we don't update the trait to SINGLETON if the `childDist` is replicated.
What would happen when we have an aggregation on the replicated table?

PR #6118, opened 2019-04-05
Added advance() method

kaikaizi:
of TableIterator to help advance iterator arbitrary times

### DIFF
--- a/src/ee/executors/limitexecutor.cpp
+++ b/src/ee/executors/limitexecutor.cpp
@@ -100,18 +100,16 @@ LimitExecutor::p_execute(const NValueArray &params)
     int offset = -1;
     node->getLimitAndOffsetByReference(params, limit, offset);
 
-    while ((limit == -1 || tuple_ctr < limit) && iterator.next(tuple))
-    {
-        // TODO: need a way to skip / iterate N items.
-        if (tuples_skipped < offset)
-        {
-            tuples_skipped++;
-            continue;
+    while ((limit == -1 || tuple_ctr < limit) && iterator.next(tuple)) {
+        if (tuples_skipped < offset) {
+           tuples_skipped += iterator.advance(tuple, offset - tuples_skipped);
+           if (tuples_skipped < offset) { // offset > table count
+              break;
+           }
         }
         tuple_ctr++;
 
-        if (!output_table->insertTuple(tuple))
-        {
+        if (!output_table->insertTuple(tuple)) {
             VOLT_ERROR("Failed to insert tuple from input table '%s' into"
                        " output table '%s'",
                        input_table->name().c_str(),
--- a/src/ee/storage/tableiterator.h
+++ b/src/ee/storage/tableiterator.h
@@ -88,6 +88,15 @@ public:
     */
     bool next(TableTuple &out);
 
+    /**
+     * Updates the given tuple so that it points to the next
+     * off-th tuple in the table.
+     * @param out the tuple will advance at most off positions
+     * @param off offset to advance
+     * @return actual advanced position
+     */
+    size_t advance(TableTuple& out, size_t const off);
+
     bool hasNext() const;
 
     uint32_t getFoundTuples() const {
@@ -442,6 +451,12 @@ inline bool TableIterator::next(TableTuple &out) {
         assert(m_iteratorType == LARGE_TEMP);
         return largeTempNext(out);
     }
+}
+
+inline size_t TableIterator::advance(TableTuple& out, size_t const off) {
+   size_t advanced = 0;
+   while(next(out) && ++advanced < off);
+   return advanced;
 }
 
 inline bool TableIterator::persistentNext(TableTuple &out) {
--- a/src/frontend/org/voltdb/plannerv2/rel/physical/VoltPValuesRule.java
+++ b/src/frontend/org/voltdb/plannerv2/rel/physical/VoltPValuesRule.java
@@ -1,0 +1,55 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2019 VoltDB Inc.
+ *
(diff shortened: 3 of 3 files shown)


src/frontend/org/voltdb/plannerv2/rel/physical/VoltPValuesRule.java
```diff
@@ -0,0 +1,55 @@
...
+import org.voltdb.plannerv2.rel.physical.VoltPhysicalRel;
+import org.voltdb.plannerv2.rel.physical.VoltPhysicalValues;
+
+/**
+ * VoltDB physical rule that transform {@link VoltLogicalValues} to {@link VoltPhysicalValues}.
+ *
+ * @author Chao Zhou
+ * @since 9.0
+ */
+public class VoltPValuesRule extends RelOptRule {
```
Chaozz: I put this class in the wrong place. Here it should be dropped.

src/ee/storage/tableiterator.h
```diff
@@ -88,6 +88,15 @@ class TableIterator : public TupleIterator {
     */
     bool next(TableTuple &out);
 
+    /**
+     * Updates the given tuple so that it points to the next
+     * off-th tuple in the table.
```
n3world: This is a bit confusing. It might be better to phrase it that off is a relative offset from the current position. Also this will return a tuple before current position + off if the end of the table is reached and in that case it will return the last tuple in the table. Also, it can return and not populate out if you are already at the end of the table. If out is only suppose to be valid when the result == offset then I think a boolean would be a more obvious return

src/ee/storage/tableiterator.h
```diff
@@ -444,6 +453,12 @@ inline bool TableIterator::next(TableTuple &out) {
     }
 }
 
+inline size_t TableIterator::advance(TableTuple& out, size_t const off) {
+   size_t advanced = 0;
+   while(next(out) && ++advanced < off);
```
n3world: Wouldn't it better to not populate the tuple for the skipped rows and just move the pointer to the correct total offset if it is valid and then only copy that one row.

kaikaizi: Certainly that's where we should be heading to.
If you take a look at the `next()` method, you'll see that it dispatches different method based on a enumerator type, which is not the *right* way to do things (in this case, OO). Without changing that implementation and all the implications involved, it is hard to just skip by `offset` in `advance` method.
One thing at a time...

src/ee/storage/tableiterator.h
kaikaizi: Updated.

src/ee/storage/tableiterator.h
```diff
@@ -88,6 +88,16 @@ class TableIterator : public TupleIterator {
     */
     bool next(TableTuple &out);
 
+    /**
+     * Updates the given tuple so that it points to the next
```
GuaGuaDean: Update(**_s_**) the given **

src/ee/storage/tableiterator.h
```diff
@@ -455,7 +456,7 @@ inline bool TableIterator::next(TableTuple &out) {
 
 inline size_t TableIterator::advance(TableTuple& out, size_t const off) {
    size_t advanced = 0;
-   while(next(out) && ++advanced < off);
+   while(++advanced < off && next(out));
```
n3world: I think your first ordering was correct. Now if you pass in 1 as off then next will never be called. Also, if there is no next you will return 1 as the result even though it did not advance at all.

src/ee/executors/limitexecutor.cpp
```diff
@@ -77,48 +77,37 @@ LimitExecutor::p_init(AbstractPlanNode* abstract_node,
...
-        }
-        tuple_ctr++;
-
-        if (!output_table->insertTuple(tuple))
-        {
-            VOLT_ERROR("Failed to insert tuple from input table '%s' into"
-                       " output table '%s'",
-                       input_table->name().c_str(),
-                       output_table->name().c_str());
+   if (limit < 0 || iterator.advance(tuple, offset) < offset) {
```
yzhang1991:
```
2> create table t (a int not null);
Table T has already been created!
Command succeeded.
3> partition table t on column a;
Command succeeded.
4> insert into t values (1);
(Returned 1 rows in 0.00s)
5> insert into t values (2);
(Returned 1 rows in 0.00s)
6> insert into t values (3);
(Returned 1 rows in 0.00s)
7> explain select * from t offset 1;
EXECUTION_PLAN
 LIMIT with parameter
  RECEIVE FROM ALL PARTITIONS
    SEQUENTIAL SCAN of "T"


(Returned 1 rows in 0.00s)
8> select * from t offset 1;
A
--
 1
```
Please correct this and add it to junit.

src/ee/executors/limitexecutor.cpp
```diff
@@ -77,48 +77,37 @@ LimitExecutor::p_init(AbstractPlanNode* abstract_node,
...
+      return true;     // offset beyond table count: empty table
+   } else if (limit >= 0) {
```
yzhang1991: Do we need to include limit == 0 here? or better in the block above

kaikaizi: Below loop should do: the `limit ==0` is included in the `tuple_ctr++ < limit` part of check.
```suggestion
    if (iterator.advance(tuple, offset) < offset) {
      return true;     // offset beyond table count: empty table
   } else {
      while((limit < 0 || tuple_ctr++ < limit) && iterator.next(tuple)) {
         if (!output_table->insertTuple(tuple)) {
            VOLT_ERROR("Failed to insert tuple from input table '%s' into output table '%s'",
                  input_table->name().c_str(), output_table->name().c_str());
            return false;
         }
      }
      return true;
   }

```

PR #6125, opened 2019-04-08
ENG-15811, satisfy outstanding poll request on mastership transfer

rdykiel:


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1669,7 +1669,13 @@ public void run() {
                         if (m_mastershipAccepted.compareAndSet(false, true)) {
                             // Either get enough responses or have received TRANSFER_MASTER event, clear the response sender HSids.
                             m_queryResponses.clear();
-                            m_onMastership.run();
+                            // If we have an outstanding poll request, try satisfying it,
+                            // otherwise start a polling runnable.
+                            if (m_pollTask != null) {
+                                pollImpl(m_pollTask);
```
luyangco: `m_onMastership.run()` calls `GuestProcessor.runDataSource()`, which in turn calls `poll()` in `buildListener()`, I think we'll get the root cause if we can answer why poll() doesn't call pollImpl() to start polling data.

rdykiel: The problem is that the runnable is run regardless of whether it is already running.

PR #6128, opened 2019-04-09
ENG-15741, unregister the mailbox id from export generation mailboxes…

luyangco:
…… (#6112)

Unregister the mailbox id from export generation mailboxes when the last data source for given partition is drained.
Keep a reference of HostMessenger in ExportGeneration to avoid excessive parameter passing.
Discard ACKs and suppress the error message if dangling data source is unregistering export mailbox.
Fix a test issue that causing NPE in testExportGeneration.

Change-Id: I319e2aae21cadd6b8d508a39bef88b48c3eab210

### DIFF
--- a/src/frontend/org/voltdb/export/ExportGeneration.java
+++ b/src/frontend/org/voltdb/export/ExportGeneration.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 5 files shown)


benjaminballard:
Lu Yang, can you resolve and merge this? It looks like it was left undone.

luyangco:
> Lu Yang, can you resolve and merge this? It looks like it was left undone.

Thanks for reminding it to me, I should close it. This change has already in v8.4.x as part of the coordinator-based export backport work, under the ticket of ENG-17592.

PR #6131, opened 2019-04-09
ENG-15827:

manjujames:
Rejoining node was getting a newer catalog than the rest of the clusters. It looks like
the cluster started UpdateClasses NTProc, which called UpdateCore sysproc as the final step.
UpdateCore wrote the new catalog to ZK node, but the MPI node failed before any fragments were run.
Node rejoined a few seconds later and picked up the new catalog from ZK, while the
rest of the cluster stayed on the old catalog. To avoid this, write new catalog bytes
to ZK node, only after the first fragment of UpdateCore completes, thus ensuring that UpdateCore
will be restarted on the new MPI even on an MPI node failure.

### DIFF
--- a/src/frontend/org/voltdb/VoltZK.java
+++ b/src/frontend/org/voltdb/VoltZK.java
@@ -52,7 +52,6 @@ public class VoltZK {
 
     public static final String buildstring = "/db/buildstring";
     public static final String catalogbytes = "/db/catalogbytes";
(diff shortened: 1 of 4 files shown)


manjujames:
Ignore the commented out stuff for now. I will remove that and push a cleaned up version if the tests run fine on the branch.

PR #6134, opened 2019-04-09
ENG-15233:

wweiss-voltdb:
Log messages from MP dump messages (after a possible MP Deadlock) don't include the current repair log. This makes it harder to detect inconsistencies between the repair log and the transaction task queue. This branch adds additional log output as well as cleaning the formatting of some of the more common task messages that are logged.

### DIFF
--- a/src/frontend/org/voltdb/iv2/CompleteTransactionTask.java
+++ b/src/frontend/org/voltdb/iv2/CompleteTransactionTask.java
@@ -267,7 +267,8 @@ public class CompleteTransactionTask extends TransactionTask
             sb.append("  SP HANDLE: ").append(TxnEgo.txnIdToString(getSpHandle()));
             sb.append("  UNDO TOKEN: ").append(m_txnState.getBeginUndoToken());
         }
-        sb.append("  MSG: ").append(m_completeMsg.toString());
+        sb.append("  MSG:\n    ");
+        m_completeMsg.indentedString(sb, 8);
         return sb.toString();
     }
 
--- a/src/frontend/org/voltdb/iv2/RepairLog.java
+++ b/src/frontend/org/voltdb/iv2/RepairLog.java
@@ -18,6 +18,7 @@
 package org.voltdb.iv2;
 
 import java.util.ArrayDeque;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Deque;
@@ -336,5 +337,28 @@ public class RepairLog
     void registerTransactionCommitInterest(TransactionCommitInterest interest)
     {
         m_txnCommitInterests.add(interest);
+    }
+
+    private void logSummary(StringBuilder sb, Deque<Item> log, String indentStr) {
+        final int txnIdsPerLine = 15;
+        Iterator<Item> itemator = log.iterator();
+        for (int lineCnt = 0; lineCnt <= (log.size() / txnIdsPerLine); lineCnt++) {
+            sb.append(indentStr);
+            for(int i = 0; i < txnIdsPerLine; i++) {
+                if (itemator.hasNext()) {
+                    sb.append(" ").append(TxnEgo.txnIdSeqToString(itemator.next().getTxnId()));
+                }
+            }
+        }
+    }
+
+    public void indentedString(StringBuilder sb, int indentCnt) {
+        char[] array = new char[indentCnt-1];
+        Arrays.fill(array, ' ');
+        String indentStr = new String("\n" + new String(array));
+        sb.append(indentStr).append("MP RepairLog:");
+        logSummary(sb, m_logMP, indentStr);
+        sb.append(indentStr).append("SP RepairLog:");
+        logSummary(sb, m_logSP, indentStr);
     }
 }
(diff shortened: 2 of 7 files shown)


src/frontend/org/voltdb/iv2/CompleteTransactionTask.java
```diff
@@ -267,7 +267,8 @@ public String toString()
             sb.append("  SP HANDLE: ").append(TxnEgo.txnIdToString(getSpHandle()));
             sb.append("  UNDO TOKEN: ").append(m_txnState.getBeginUndoToken());
         }
-        sb.append("  MSG: ").append(m_completeMsg.toString());
+        sb.append("  MSG:\n    ");
+        m_completeMsg.indentedString(sb, 8);
```
rdykiel: nit: why not indent to 4 only

src/frontend/org/voltdb/messaging/CompleteTransactionMessage.java
```diff
@@ -172,43 +173,50 @@ public void initFromBuffer(ByteBuffer buf) throws IOException
         assert(buf.capacity() == buf.position());
     }
 
-    @Override
-    public String toString() {
-        StringBuilder sb = new StringBuilder();
-
+    public void indentedString(StringBuilder sb, int indentCnt) {
+        char[] array = new char[indentCnt];
```
rdykiel: Same comment for C++: you could have a constant static array of 4 spaces and append the static multiple times, indentCnt becoming the depth e.g 2 would append the constant array 2 times

src/frontend/org/voltdb/iv2/ReplaySequencer.java
```diff
@@ -364,17 +362,17 @@ else if (in instanceof CompleteTransactionMessage) {
...
-        hostLog.warn(String.format("%s: REPLAY SEQUENCER DUMP, LAST POLLED FRAGMENT %d (%s), LAST SEEN TXNID %d (%s), %s%s",
+        sb.append(String.format("%s: REPLAY SEQUENCER DUMP, LAST POLLED FRAGMENT %d (%s), LAST SEEN TXNID %d (%s), %s%s",
                                  who,
                                  m_lastPolledFragmentUniqueId, TxnEgo.txnIdToString(m_lastPolledFragmentUniqueId),
                                  m_lastSeenUniqueId, TxnEgo.txnIdToString(m_lastSeenUniqueId),
                                  m_mpiEOLReached ? "MPI EOL, " : "",
                                  m_mustDrain ? "MUST DRAIN" : ""));
         for (Entry<Long, ReplayEntry> e : m_replayEntries.entrySet()) {
-            hostLog.warn(String.format("%s: REPLAY ENTRY %s: %s", who, e.getKey(), e.getValue()));
+            sb.append(String.format("\n    %s: REPLAY ENTRY %s: %s", who, e.getKey(), e.getValue()));
```
n3world: It seems counter productive to format a string for a StringBuilder. This could be written as
`sb.append("\n    ").append(who).append(": REPLAY ENTRY ").append(e.getKey()).append(": ").append(e.getValue());`

Same for the line above

src/frontend/org/voltdb/iv2/SpScheduler.java
```diff
@@ -1646,9 +1646,12 @@ void safeAddToDuplicateCounterMap(DuplicateCounterKey dpKey, DuplicateCounter co
     @Override
     public void dump()
     {
-        m_replaySequencer.dump(m_mailbox.getHSId());
-        hostLog.warn("[dump] current truncation handle: " + TxnEgo.txnIdToString(m_repairLogTruncationHandle) + " "
+        StringBuilder sb = new StringBuilder();
+        m_replaySequencer.dump(m_mailbox.getHSId(), sb);
+        sb.append("\n    current truncation handle: " + TxnEgo.txnIdToString(m_repairLogTruncationHandle) + " "
```
n3world: Why is string concatenation being done as an argument to append? That is the purpose of append. Also you usually want to avoid calling toString() directly since that is handled for you by java and java handles nulls correctly

tests/frontend/org/voltdb/iv2/TestReplaySequencer.java
```diff
@@ -112,7 +115,7 @@ public void testOfferSentinelThenFragments()
         TransactionInfoBaseMessage frag2 = makeFragment(1L);
 
         result = dut.offer(1L, sntl);
-        try { dut.dump(1); } catch (Exception e) { fail(e.getMessage()); } // toString should not throw
+        try { dut.dump(1, sb); hostLog.warn(sb.toString()); } catch (Exception e) { fail(e.getMessage()); } // toString should not throw
```
n3world: It would be easier to debug if the exception was just allowed to be caught and handled by junit instead of being lost except for the message, which doesn't even include the exception type. Same comment for other occurrence of this.

src/frontend/org/voltdb/iv2/RepairLog.java
```diff
@@ -337,4 +338,27 @@ void registerTransactionCommitInterest(TransactionCommitInterest interest)
...
+            for(int i = 0; i < txnIdsPerLine; i++) {
+                if (itemator.hasNext()) {
+                    sb.append(" ").append(TxnEgo.txnIdSeqToString(itemator.next().getTxnId()));
+                }
+            }
+        }
+    }
+
+    public void indentedString(StringBuilder sb, int indentCnt) {
+        char[] array = new char[indentCnt-1];
```
n3world: Why subtract 1 from the array length?

PR #6135, opened 2019-04-09
ENG-15811, close potential race conditions on m_pollTask

rdykiel:
Refactored the interface between ExportDataSource and GuestProcessor. Implemented a new ExportCoordinator class to coodinate export mastership, gap detection and gap correction.

### DIFF
--- a/src/frontend/org/voltdb/export/ExportCoordinator.java
+++ b/src/frontend/org/voltdb/export/ExportCoordinator.java
@@ -834,7 +834,7 @@ public class ExportCoordinator {
                     + "], look for candidate replicas");
         }
 
-        Integer replicaId = Integer.MIN_VALUE;
+        Integer replicaId = NO_HOST_ID;
         long leaderNextSafePoint = gap.getSecond() + 1;
         long  replicaSafePoint = 0L;
 
--- a/src/frontend/org/voltdb/export/ExportCoordinator.java
+++ b/src/frontend/org/voltdb/export/ExportCoordinator.java
@@ -491,7 +491,7 @@ public class ExportCoordinator {
             task.setInitialState(initialState);
             task.registerStateMachineWithManager(initialState);
 
-            exportLog.info("Created ssm for topic " + topicName + ", and hostId " + m_hostId
+            exportLog.info("Created export coordinator for topic " + topicName + ", and hostId " + m_hostId
                     + ", leaderHostId: " + m_leaderHostId);
 
         } catch (Exception e) {
@@ -509,7 +509,7 @@ public class ExportCoordinator {
      */
     public void shutdown() throws InterruptedException {
         if (exportLog.isDebugEnabled()) {
-            exportLog.debug("Shutting down...");
+            exportLog.debug("Export coordinator shutting down...");
         }
         m_ssm.ShutdownSynchronizedStatesManager();
     }
@@ -520,6 +520,7 @@ public class ExportCoordinator {
     public void becomeLeader() {
         if (m_hostId.equals(m_leaderHostId)) {
             exportLog.warn(m_eds + " is already the partition leader");
+            return;
         }
 
         m_task.invoke(new Runnable() {
--- a/src/frontend/org/voltdb/export/ExportCoordinator.java
+++ b/src/frontend/org/voltdb/export/ExportCoordinator.java
@@ -130,7 +130,7 @@ public class ExportCoordinator {
                             m_leaderHostId = newLeaderHostId;
                             StringBuilder sb = new StringBuilder("Initialized export coordinator: host ")
                                     .append(m_leaderHostId)
-                                    .append(isLeader() ? " (localHost) " : " ")
+                                    .append(isPartitionLeader() ? " (localHost) " : " ")
                                     .append("is the leader at initial state");
                             exportLog.info(sb.toString());
                             setCoordinatorInitialized();
@@ -342,14 +342,14 @@ public class ExportCoordinator {
                             m_leaderHostId = newLeaderHostId;
                             StringBuilder sb = new StringBuilder("Host ")
                                     .append(m_leaderHostId)
-                                    .append(isLeader() ? " (localHost) " : " ")
+                                    .append(isPartitionLeader() ? " (localHost) " : " ")
                                     .append("is the new leader");
                             exportLog.info(sb.toString());
 
                             // If leader and maps empty request ExportSequenceNumberTracker from all nodes.
                             // Note: cannot initiate a coordinator task directly from here, must go
                             // through another runnable and the invocation path.
-                            if (isLeader() && m_trackers.isEmpty()) {
+                            if (isPartitionLeader() && m_trackers.isEmpty()) {
                                 requestTrackers();
                             }
 
@@ -405,7 +405,7 @@ public class ExportCoordinator {
                             }
 
                             // If not the leader, check if the leader is behind or ahead on acks.
-                            if (!isLeader()) {
+                            if (!isPartitionLeader()) {
                                 if (lastReleasedSeqNo < m_eds.getLastReleaseSeqNo()) {
                                     // Leader is behind on acks, ask the ExportDataSource to resend
                                     // an ack to all the other hosts.
@@ -589,7 +589,7 @@ public class ExportCoordinator {
                             // may have lost leadership. But this is not important as the goal is to start the task
                             // from one host. Note also that the request goes through another EDS runnable.
                             if (!addedMembers.isEmpty()) {
-                                if (isLeader()) {
+                                if (isPartitionLeader()) {
                                     exportLog.info("Leader requests trackers for added members: " + addedMembers);
                                     requestTrackers();
 
@@ -897,7 +897,7 @@ public class ExportCoordinator {
     /**
      * @return true if this host is the partition leader.
      */
-    public boolean isLeader() {
+    public boolean isPartitionLeader() {
         return m_hostId.equals(m_leaderHostId);
     }
 
@@ -991,12 +991,10 @@ public class ExportCoordinator {
 
         // Note: the trackers are truncated so the seqNo should not be past the first gap
         Pair<Long, Long> gap = leaderTracker.getFirstGap();
-        if (gap != null) {
-            assert (exportSeqNo <= gap.getSecond());
-        }
+        assert (gap == null || exportSeqNo <= gap.getSecond());
         if (gap == null || exportSeqNo < (gap.getFirst() - 1)) {
 
-            m_isMaster = isLeader();
+            m_isMaster = isPartitionLeader();
             if (gap == null) {
                 m_safePoint = INFINITE_SEQNO;
             } else {
@@ -1051,7 +1049,7 @@ public class ExportCoordinator {
 
         // If no replicas were found, the leader is Export Master and will become BLOCKED
         m_safePoint = gap.getFirst();
-        m_isMaster = isLeader();
+        m_isMaster = isPartitionLeader();
         if (exportLog.isDebugEnabled()) {
             exportLog.debug("Leader host " + m_leaderHostId + " is Export Master and will be blocked");
         }
--- a/src/frontend/org/voltdb/export/ExportCoordinator.java
+++ b/src/frontend/org/voltdb/export/ExportCoordinator.java
@@ -530,8 +530,8 @@ public class ExportCoordinator {
     private final ExportDataSource m_eds;
 
     // State machine will only be instantiated on initialize()
-    private SynchronizedStatesManager m_ssm;
-    private ExportCoordinationTask m_task;
+    private final SynchronizedStatesManager m_ssm;
+    private final ExportCoordinationTask m_task;
 
     private Integer m_leaderHostId = NO_HOST_ID;
     private static final int NO_HOST_ID =  -1;
@@ -597,8 +597,38 @@ public class ExportCoordinator {
         m_rootPath = rootPath;
         m_hostId = hostId;
         m_eds = eds;
+
+        SynchronizedStatesManager ssm = null;
+        ExportCoordinationTask task = null;
+        try {
+            ZKUtil.addIfMissing(m_zk, m_rootPath, CreateMode.PERSISTENT, null);
+
+            // Set up a SynchronizedStateManager for the topic/partition
+            String topicName = getTopicName(m_eds.getTableName(), m_eds.getPartitionId());
+            ssm = new SynchronizedStatesManager(
+                    m_zk,
+                    m_rootPath,
+                    topicName,
+                    m_hostId.toString(),
+                    1);
+
+            task = new ExportCoordinationTask(ssm);
+
+            exportLog.info("Created export coordinator for topic " + topicName + ", and hostId " + m_hostId
+                    + ", leaderHostId: " + m_leaderHostId);
+
+        } catch (Exception e) {
+            VoltDB.crashLocalVoltDB("Failed to initialize ExportCoordinator state machine", true, e);
+        } finally {
+            m_ssm = ssm;
+            m_task = task;
+        }
     }
 
+    private String getTopicName(String tableName, int partitionId) {
+        return tableName + "_" + partitionId;
+
+    }
     /**
      * @return true if coordinator is initialized
      *
@@ -633,23 +663,12 @@ public class ExportCoordinator {
         }
         try {
             m_state = State.INITIALIZING;
-            ZKUtil.addIfMissing(m_zk, m_rootPath, CreateMode.PERSISTENT, null);
-
-            // Set up a SynchronizedStateManager for the topic/partition
-            String topicName = m_eds.getTableName() + "_" + m_eds.getPartitionId();
-            m_ssm = new SynchronizedStatesManager(
-                    m_zk,
-                    m_rootPath,
-                    topicName,
-                    m_hostId.toString(),
-                    1);
-
-            m_task = new ExportCoordinationTask(m_ssm);
             ByteBuffer initialState = ByteBuffer.allocate(4);
             initialState.putInt(m_leaderHostId);
             initialState.flip();
             m_task.registerStateMachineWithManager(initialState);
 
+            String topicName = getTopicName(m_eds.getTableName(), m_eds.getPartitionId());
             exportLog.info("Initialized export coordinator for topic " + topicName + ", and hostId " + m_hostId
                     + ", leaderHostId: " + m_leaderHostId);
 
@@ -682,6 +701,7 @@ public class ExportCoordinator {
      * NOTE: we accept this invocation even if not initialized yet
      */
     public void becomeLeader() {
+
 
         if (m_hostId.equals(m_leaderHostId)) {
             exportLog.warn(m_eds + " is already the partition leader");
--- a/src/frontend/org/voltdb/export/ExportCoordinator.java
+++ b/src/frontend/org/voltdb/export/ExportCoordinator.java
@@ -36,8 +36,6 @@ import org.voltdb.VoltDB;
 
 
 /**
- * @author rdykiel
- *
  * An {@code ExportCoordinator} instance determines whether the associated
  * {@code ExportDataSource} instance should be an export master. It coordinates
  * the transfer of export mastership between the replicas of an export partition.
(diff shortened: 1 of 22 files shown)


src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -1019,13 +1019,21 @@ public void close() {
...
         synchronized(m_dataSourcesByPartition) {
             for (Map<String, ExportDataSource> partitionDataSourceMap : m_dataSourcesByPartition.values()) {
                 for (ExportDataSource source : partitionDataSourceMap.values()) {
-                    source.unacceptMastership();
+                    tasks.add(source.unacceptMastership());
                 }
             }
         }
+        try {
+            Futures.allAsList(tasks).get();
```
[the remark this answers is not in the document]
rdykiel: OK I like this, as it simplifies the implementation.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1093,61 +1056,186 @@ public void run() {
...
+            // Ensure this first block has a schema
+            BBContainer schemaContainer = m_committedBuffers.pollSchema();
+            if (schemaContainer == null) {
+                try {
+                    pollTask.setException(new IOException("No schema for committedSeqNo " + cont.m_commitSeqNo
+                            + ", discarding buffer (rows may be lost)."));
+                } catch (RejectedExecutionException reex) {
+                    /* Ignore */
+                }
+                if (exportLog.isDebugEnabled()) {
```
luyangco: Highly unlikely to hit this, so change the logging to error to make it obvious.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1093,61 +1056,186 @@ public void run() {
...
+
+    // Check the continuity of a pending buffer, which stands BEFORE the unpolled ones
+    private ContinuityCheckResult checkPendingContinuity(long startSeqNo, long lastSeqNo) {
+
+        if (lastSeqNo <= m_lastReleasedSeqNo) {
+            return ContinuityCheckResult.ACKED;
+        }
+
+        if ((startSeqNo <= m_firstUnpolledSeqNo && m_firstUnpolledSeqNo <= lastSeqNo)
+                || (lastSeqNo + 1) == m_firstUnpolledSeqNo) {
```
luyangco: `(lastSeqNo + 1) == m_firstUnpolledSeqNo` is wrong, for example, buffer [10,20] is in pendingContainer, and `m_firstUnpolledSeqNo` is 21, polling buffer[10,20] doesn't give GP the needed data.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1020,68 +1036,15 @@ public void run() {
                         } catch (RejectedExecutionException rej) {
                             // Ignore: the {@code GuestProcessor} was shut down
                         }
-                        m_pollTask = null;
```
luyangco: I think we should call `pollTask.setFuture(null)` here to terminate GP, because when the code reaches here it means data source is no longer the master but GP still trying to poll. This is possible because when data source gives up mastership, m_pollTask may not be set.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1630,46 +1630,70 @@ private void sendQueryResponse(long senderHSId, long requestId, long lastSeq) {
...
-                        }
+                        exportLog.info("Accepting mastership for " + this);
                         if (m_mastershipAccepted.compareAndSet(false, true)) {
                             // Either get enough responses or have received TRANSFER_MASTER event, clear the response sender HSids.
                             m_queryResponses.clear();
-                            m_onMastership.run();
+                            // ENG-15811 - not sure that the {@code GuestProcessor} is not already
+                            // running when we get there.
+                            // FIXME: must remove info messages
+                            if (m_pollTask != null) {
```
luyangco: I don't think we need this, because if we always terminate GP after master is migrated away, m_pollTask will never be null here.

luyangco: Oh there is still another race, if EDS gave up mastership but left GP running,  and accepts mastership before GP does another poll, in that case we leaks old GP while a new GP is created.

src/frontend/org/voltdb/export/ExportDataSource.java
rdykiel: I'm surprised: if pending buffer is [10, 20] and m_firstUnpolledSeqNo  == 21 isn't this exactly the case why we are managing the pending buffer? Recall that if we put a buffer in pending, it is because we polled PAST it. Actually our first versions of the fix were inconsistent:

- if we poll the buffer and pass it to GP and if GP sets it as the pending buffer because it is shutting down, then m_firstUnpolledSeqNo == pendingBuffer.lastSeqNo + 1

- if we poll the buffer and detect a gap, we pend the buffer and don't set m_firstUnpolledSeqNo; I think we should.

src/frontend/org/voltdb/export/ExportDataSource.java
rdykiel: I will change this by implementing the change that I outlined in slack this morning.

src/frontend/org/voltdb/export/ExportDataSource.java
rdykiel: Right, will do

src/frontend/org/voltdb/export/ExportDataSource.java
rdykiel: No I don't think so; if you hit RejectedExecutionException here it's because this GP instance that is waiting on the fut has already been shut down.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: I see, the problem is that we don't set m_firstUnpolledSeqNo consistently in the two cases above. I agree we should advance m_firstUnpolledSeqNo in GAP detection case too.

src/frontend/org/voltdb/export/processors/GuestProcessor.java
```diff
@@ -246,19 +244,22 @@ private void runDataSource() {
                                 m_source.m_columnLengths,
                                 m_source.getExportFormat());
 
-                // in this case we cannot poll until the initial truncation is complete
+                // Ee cannot poll until the initial truncation is complete
```
luyangco: Ea => We

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1630,46 +1655,54 @@ private void sendQueryResponse(long senderHSId, long requestId, long lastSeq) {
...
                     if (!m_es.isShutdown() || !m_closed) {
-                        if (exportLog.isDebugEnabled()) {
-                            exportLog.debug("Export table " + getTableName() + " accepting mastership for partition " + getPartitionId());
-                        }
+                        exportLog.info("Accepting mastership");
                         if (m_mastershipAccepted.compareAndSet(false, true)) {
                             // Either get enough responses or have received TRANSFER_MASTER event, clear the response sender HSids.
                             m_queryResponses.clear();
-                            m_onMastership.run();
+                            // FIXME: must remove info messages
```
luyangco: Downgrade info log to warning? Remove FIXME?

luyangco:
I imagine two scenarios that are involved around `m_pollFuture`:

1. the given data source gains or loses mastership (possibly re-gains mastership shortly after) while GP thread is running, e.g. export mastership migration/gap detection. In this case, `m_pollFuture` defined in ExportDataSource keeps the poll pending while GP on the other side waits for its completion, which is good.
2. all data sources give up mastership, GP thread shuts down and re-created, all data sources re-gain mastership, e.g. catalog update. In this case, leaving `m_pollFuture` non-empty in data source is dangerous, because the pending poll may come from the terminated GP thread (imagine that, GP queues a poll in data source, after that it exits and do catalog update, then a new GP polls same data source again).

To fix the issue, I think we should add a callback in ExportDataSource, when GP is shutting down, calls the callback to null `m_pollFuture` because it will become irrelevant soon.

src/frontend/org/voltdb/export/ExportCoordinator.java
```diff
@@ -0,0 +1,1130 @@
...
+        if (leaderTracker == null) {
+            // This means that the leadership has been resolved but the
+            // trackers haven't been gathered
+            return false;
+        }
+
+        // Note: the trackers are truncated so the seqNo should not be past the first gap
+        Pair<Long, Long> gap = leaderTracker.getFirstGap();
+        if (gap != null) {
+            assert (exportSeqNo <= gap.getSecond());
```
luyangco: Could rewrite to `assert (gap == null || exportSeqNo <= gap.getSecond());`.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -784,11 +808,9 @@ public void discard() {
...
         }
         if (poll) {
-            try {
-                pollImpl(m_pollTask);
-            } catch (RejectedExecutionException ex) {
-                //Its ok.
-            }
+            // Note: this should only be executed in a runnable
+            assert(!m_es.isShutdown());
+            pollImpl(m_pollTask);
```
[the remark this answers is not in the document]
rdykiel: Are we sure the thread is always the same? I could imagine the system recycling an idle thread and re-creating as needed. I haven't checked the executor creation parameters. Anyway that's an optimization I'd like to defer.

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -693,6 +605,7 @@ private void addDataSource(File adFile,
             final ExportDataProcessor processor,
             final long genId) throws IOException {
         ExportDataSource source = new ExportDataSource(this, adFile, localPartitionsToSites, processor, genId);
+        source.setCoordination(m_messenger.getZK(), m_messenger.getHostId());
```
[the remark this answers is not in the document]
rdykiel: My reasons were:
1- I thought it would facilitate JUnit testing in that it would make it easier to mock the eds (when testing the coordinator), or mock the coordinator (when testing the eds)

2- I was reluctant to add a couple more parameters to the (already crowded) constructor API. So I decided to keep with this pattern for now.

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -758,6 +671,7 @@ private void addDataSources(Table table, int hostId,
                                 table.getColumns(),
                                 table.getPartitioncolumn(),
                                 m_directory.getPath());
+                        exportDataSource.setCoordination(m_messenger.getZK(), m_messenger.getHostId());
```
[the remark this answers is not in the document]
rdykiel: See previous comment.

src/frontend/org/voltdb/export/ExportCoordinator.java
```diff
@@ -0,0 +1,1130 @@
...
+            public String toString() {
+                return "becomeLeader request for host:" + m_hostId;
+            }
+        });
+    }
+
+    /**
+     * @return true if this host is the partition leader.
+     */
+    public boolean isLeader() {
```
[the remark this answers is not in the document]
rdykiel: Changed, thanks

src/frontend/org/voltdb/export/ExportCoordinator.java
rdykiel: Done.

src/frontend/org/voltdb/export/processors/GuestProcessor.java
```diff
@@ -52,12 +52,12 @@
 
 public class GuestProcessor implements ExportDataProcessor {
 
+    private static final VoltLogger exportLog = new VoltLogger("EXPORT");
```
wweiss-voltdb: Either prepend with s_ or UpperCase this. Otherwise it is assumed to be a method local variable.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -933,10 +953,10 @@ public void run() {
                     }
                     m_committedBuffers.closeAndDelete();
                     m_adFile.delete();
-                } catch(IOException e) {
+                    m_coordinator.shutdown();
+
```
wweiss-voltdb: remove whitespace

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -950,15 +970,24 @@ public void run() {
                 try {
                     m_committedBuffers.close();
                     m_ackMailboxRefs.set(null);
-                } catch (IOException e) {
+                    m_coordinator.shutdown();
+
```
wweiss-voltdb: ditto

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1174,138 +1253,79 @@ private synchronized void pollImpl(PollTask pollTask) {
...
-                        }
-                        assert(m_startTime != 0);
-                        long elapsedMS = System.currentTimeMillis() - m_startTime;
-                        m_blocksSentSinceClear += 1;
-                        m_totalLatencySinceClearInMS += elapsedMS;
-                        m_averageLatency = m_totalLatencySinceClearInMS / m_blocksSentSinceClear;
-                        if (m_averageLatency > m_maxLatency) {
-                            m_maxLatency = m_averageLatency;
-                        }
+    private BinaryPayloadMessage getReleaseBufferMessage() {
```
wweiss-voltdb: createReleaseBufferMessage is more descriptive.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1455,229 +1464,97 @@ private void ackImpl(long seq) {
      /**
       * Notify the generation when source is drained on an unused partition.
       *
+      * @param pollTask the current poll request or null
+      * @return true if handled a drained source
+      *
       * @throws IOException
       */
-     private void handleDrainedSource() throws IOException {
+     /**
```
wweiss-voltdb: Remove this comment

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -979,150 +1033,184 @@ public void setPendingContainer(AckingContainer container) {
...
+                    if (!m_coordinator.isExportMaster(m_firstUnpolledSeqNo)) {
+                        if (exportLog.isDebugEnabled()) {
+                            exportLog.debug("Not export master for seqNo " + m_firstUnpolledSeqNo);
                         }
+                        // Put the poll aside until we become Export Master
+                        m_pollTask = pollTask;
+                        return;
+                    }
+
+                    // If the next block is not in sequence, and we were told we're Export Master,
```
wweiss-voltdb: Who told us we're Export Master? Also, does this imply that the status for the non-masters is ACTIVE?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -979,150 +1033,184 @@ public void setPendingContainer(AckingContainer container) {
...
-                                        "] from " + this.toString() + ". Searching other sites for missing data.");
-                                m_seqNoToDrain = firstUnpolledSeq - 1;
-                                mastershipCheckpoint(firstUnpolledSeq - 1);
-                            }
-                            break;
+                        continue;
+                    }
+
+                    // Are we the Export Master for the unpolled sequence number?
+                    if (!m_coordinator.isExportMaster(m_firstUnpolledSeqNo)) {
```
wweiss-voltdb: Does isExportMaster() return true if this is a RunEverywhere stream or will RunEverywhere streams break out here? I am not sure if RunEverywhere semantics is more appropriate in the coordinator or in the EDS but it should ideally not be in both.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1989,16 +1618,57 @@ private void resetStateInRejoinOrRecover(long initialSequenceNumber, boolean isR
...
+                    + ", first unpolled " + m_firstUnpolledSeqNo);
+        }
     }
 
     public String getTarget() {
         return m_exportTargetName;
     }
 
+    private void blockOnGap(long start, long end) {
+
```
wweiss-voltdb: Whitespace?

src/frontend/org/voltdb/export/processors/GuestProcessor.java
```diff
@@ -274,28 +275,42 @@ public void run() {
...
                         }
                     }
                 };
-                if (m_shutdown) return;
-                if (!m_source.getExecutorService().isShutdown()) try {
+                if (m_shutdown) {
+                    exportLog.warn("Got shutdown while starting.");
+                    return;
+                }
+                if (m_source.getExecutorService().isShutdown()) {
```
wweiss-voltdb: This block and the next could all be implemented by the EDS. This would be safer since it could then be treated the same as other queued tasks (which may or may not be protected by a sync block).

src/frontend/org/voltdb/export/ExportCoordinator.java
```diff
@@ -0,0 +1,1128 @@
...
+        private AtomicBoolean m_pending = new AtomicBoolean(false);
+        private AtomicBoolean m_shutdown = new AtomicBoolean(false);
+
+        public ExportCoordinationTask(SynchronizedStatesManager ssm) {
+            ssm.super(s_coordinatorTaskName, exportLog);
+        }
+
+        @Override
+        protected ByteBuffer notifyOfStateMachineReset(boolean isDirectVictim) {
+            // FIXME: should we crash voltdb
```
wweiss-voltdb: Resets are ordinarily caused by an issue in the notifications that was unresolvable. This is mostly defensive but should not crash the node or the cluster. It is intended for survivability.

src/frontend/org/voltdb/export/ExportDataSource.java
rdykiel: the call to m_coordinator.isExportMaster several lines above. Yes the state of the replicas is ACTIVE, AFAIK

src/frontend/org/voltdb/export/ExportCoordinator.java
```diff
@@ -0,0 +1,1128 @@
...
+        protected String stateToString(ByteBuffer state) {
+            state.rewind();
+            Integer leaderHostId = state.getInt();
+            StringBuilder sb = new StringBuilder("Leader hostId: ").append(leaderHostId);
+            return sb.toString();
+        }
+
+        @Override
+        protected String taskToString(ByteBuffer task) {
+            return "ExportCoordinationTask";
```
wweiss-voltdb: Since you have added info to the task request, you can include that here.

src/frontend/org/voltdb/export/ExportDataSource.java
rdykiel: Unfortunately it's not that easy because the 'replicated' is set by the GuestProcessor so this is transmitted to the coordinator as a mode.

src/frontend/org/voltdb/export/ExportCoordinator.java
```diff
@@ -0,0 +1,1128 @@
...
+     *  - we want to catch any exceptions occurring in the SSM daemon callbacks in order
+     *    to preserve the integrity of the SSM.
+     */
+    private class ExportCoordinationTask extends SynchronizedStatesManager.StateMachineInstance {
+
+        // A queue of invocation runnables: each invocation needs the distributed lock
+        private ConcurrentLinkedQueue<Runnable> m_invocations = new ConcurrentLinkedQueue<>();
+
+        private AtomicBoolean m_pending = new AtomicBoolean(false);
+        private AtomicBoolean m_shutdown = new AtomicBoolean(false);
```
wweiss-voltdb: It might be better to use the EDS version of shutdown. As it stands, you have 2 processes in possibly 2 different (atomic) states.

src/frontend/org/voltdb/export/processors/GuestProcessor.java
rdykiel: The GP and EDS are shut down independently and this method tries to detect both cases. Anyway this is pre-existing code that I'm loath touching at this point.

src/frontend/org/voltdb/export/ExportCoordinator.java
rdykiel: Is this method always called with the task REQUEST?

src/frontend/org/voltdb/export/ExportCoordinator.java
rdykiel: NO: this is ORDERED shutdown: the EDS shuts down the coordinator; when the coordinator shuts down it calls back into the EDS to let it shut down. I had to do this because of concurrent callbacks from SSM.

src/frontend/org/voltdb/export/ExportDataSource.java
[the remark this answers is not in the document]
luyangco: Unless you specified the non-zero `keepalive` in the constructor of executor service, the thread pool won't terminate idle thread to avoid performance cost of creating thread.

src/frontend/org/voltdb/export/ExportGeneration.java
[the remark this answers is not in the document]
luyangco: Understand.

PR #6136, opened 2019-04-10
 ENG-15826: Access to m_hasMoreChunks and m_buffers

n3world:
There is a race between TableSaveFile.close() discarding all buffers and
TableSaveFile.Container.discard(). discard() can m_hasMoreChunks prior to
close() setting it to false but insert the buffer into m_buffers after close()
has drained m_buffers(). By synchronizing the body of discard() that race goes
away and the buffers is guaranteed to be discarded and not leaked.

### DIFF
--- a/tests/ee/catalog/ExportTupleStreamTest.cpp
+++ b/tests/ee/catalog/ExportTupleStreamTest.cpp
@@ -88,7 +88,7 @@ TEST_F(ExportTupleStreamTest, TestExportTableChange) {
     initParamsBuffer();
     voltdb::ReferenceSerializeInputBE params(m_parameter_buffer.get(), m_smallBufferSize);
     // The insert statement does not have any query parameters.
-    boost::scoped_ptr<fragmentId_t> scopedPlanfragmentIds(new fragmentId_t[4]);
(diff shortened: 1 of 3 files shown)


tests/ee/catalog/ExportTupleStreamTest.cpp
```diff
@@ -88,7 +88,7 @@ TEST_F(ExportTupleStreamTest, TestExportTableChange) {
     initParamsBuffer();
     voltdb::ReferenceSerializeInputBE params(m_parameter_buffer.get(), m_smallBufferSize);
     // The insert statement does not have any query parameters.
-    boost::scoped_ptr<fragmentId_t> scopedPlanfragmentIds(new fragmentId_t[4]);
+    boost::scoped_array<fragmentId_t> scopedPlanfragmentIds(new fragmentId_t[4]);
```
manjujames: I don't understand this change, but looks good otherwise.

n3world: Oh it is to fix the valgrind failure http://ci.voltdb.lan/view/Master/job/branch-2-memcheck-lite-master/4380/

PR #6137, opened 2019-04-10
ENG-15811, fix an edge case that causing ReentrantPollException.

luyangco:
When stream hits a gap, it may poll a stream block but doesn't assign it to first_unpolled_block or m_pendingContainer, because of that, first_unpolled_block is null in pollImpl() and m_pollTask is set. The next time the stream becomes master again, guest processor tries to poll data source but because m_pollTask is not null EDS throws ReentrantPollException.

Also moving AckingContainer class out of ExportDataSource.

Change-Id: I985d0928eea664c962f07efbce12c74f4c0a6751

### DIFF
--- a/src/frontend/org/voltdb/export/AckingContainer.java
+++ b/src/frontend/org/voltdb/export/AckingContainer.java
@@ -44,7 +44,7 @@ public class AckingContainer extends BBContainer {
         m_schemaCont = schemaCont;
     }
 
-    public static AckingContainer of(ExportDataSource source,
+    public static AckingContainer create(ExportDataSource source,
                                      StreamBlock sb,
                                      StreamBlockQueue sbq,
                                      boolean forcePollSchema) {
--- a/src/frontend/org/voltdb/export/AckingContainer.java
+++ b/src/frontend/org/voltdb/export/AckingContainer.java
@@ -1,0 +1,137 @@
+/* This file is part of VoltDB.
+ * Copyright (C) 2008-2019 VoltDB Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with VoltDB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.voltdb.export;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.RejectedExecutionException;
+
+import org.voltcore.logging.VoltLogger;
+import org.voltcore.utils.DBBPool.BBContainer;
+
+public class AckingContainer extends BBContainer {
+    ExportDataSource m_source;
+    final long m_startSeqNo;
+    final long m_lastSeqNo;
+    final long m_commitSeqNo;
+    final BBContainer m_backingCont;
+    BBContainer m_schemaCont;
+    long m_startTime = 0;
+    long m_commitSpHandle = 0;
+    private static VoltLogger EXPORT_LOG = new VoltLogger("EXPORT");
+
+    private AckingContainer(ExportDataSource source, BBContainer cont,
+                            BBContainer schemaCont, long startSeqNo, long lastSeqNo, long commitSeq) {
+        super(cont.b());
+        m_source = source;
+        m_startSeqNo = startSeqNo;
+        m_lastSeqNo = lastSeqNo;
+        m_commitSeqNo = commitSeq;
+        m_backingCont = cont;
+        m_schemaCont = schemaCont;
+    }
+
+    public static AckingContainer of(ExportDataSource source,
+                                     StreamBlock sb,
+                                     StreamBlockQueue sbq,
+                                     boolean forcePollSchema) {
+        BBContainer schemaContainer = null;
+        if (forcePollSchema) {
+            schemaContainer = sbq.pollSchema();
+        } else {
+            schemaContainer = sb.getSchemaContainer();
+        }
+        return new AckingContainer(source,
+                        sb.unreleasedContainer(),
+                        schemaContainer,
+                        sb.startSequenceNumber(),
+                        sb.startSequenceNumber() + sb.rowCount() - 1,
+                        sb.committedSequenceNumber());
+    }
+
+    public void updateStartTime(long startTime) {
+        m_startTime = startTime;
+    }
+
+    // Synchronized because schema is settable
+    public synchronized ByteBuffer schema() {
+        if (m_schemaCont == null) {
+            return null;
+        }
+        return m_schemaCont.b();
+    }
+
+    public synchronized void setSchema(BBContainer schemaCont) {
+        if (m_schemaCont != null) {
+            throw new IllegalStateException("Overwriting schema");
+        }
+        m_schemaCont = schemaCont;
+    }
+
+    public long getCommittedSeqNo() {
+        return m_commitSeqNo;
+    }
+
+    public void setCommittedSpHandle(long spHandle) {
+        m_commitSpHandle = spHandle;
+    }
+
+    long getStartSeqNo() {
+        return m_startSeqNo;
+    }
+
+    // Package private
+    long getLastSeqNo() {
+        return m_lastSeqNo;
+    }
+
+    private void internalDiscard(boolean checkDoubleFree) {
+        if (checkDoubleFree) {
+            checkDoubleFree();
+        }
+        m_backingCont.discard();
+        synchronized(this) {
+            if (m_schemaCont != null) {
+                m_schemaCont.discard();
+            }
+        }
+    }
+    // Package private
+    void internalDiscard() {
+        internalDiscard(true);
+    }
+
+    @Override
+    // Invoked from GuestProcessor after the container is delivered.
+    public void discard() {
+        checkDoubleFree();
+        try {
+            m_source.advance(m_lastSeqNo, m_commitSeqNo, m_commitSpHandle, m_startTime);
+        } catch (RejectedExecutionException rej) {
+            //Don't expect this to happen outside of test, but in test it's harmless
+            if (EXPORT_LOG.isDebugEnabled()) {
+                EXPORT_LOG.debug("Acking export data task rejected, this should be harmless");
+            }
+        } finally {
+            internalDiscard(false);
+        }
+    }
+
+    @Override
+    public String toString() {
+        return new String("Container: ending at " + m_lastSeqNo + " (Committed " + m_commitSeqNo + ")");
+    }
+}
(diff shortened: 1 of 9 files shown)


src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -524,7 +524,7 @@ private synchronized void releaseExportBytes(long releaseSeqNo) throws IOExcepti
         if (m_status == StreamStatus.BLOCKED &&
                 m_gapTracker.getFirstGap() != null &&
                 releaseSeqNo >= m_gapTracker.getFirstGap().getSecond()) {
-            exportLog.info("Export queue gap resolved. Resuming export for " + ExportDataSource.this.toString());
+            exportLog.info("Export queue gap resolved. releaseExportBytes. Resuming export for " + ExportDataSource.this.toString());
```
rdykiel: This is an info message that is visible to the customer so make it "Released bytes resolve export gap. Resuming export..."

rdykiel: When we go through that branch, how are we resuming exporting? Is it only at the next buffer pushed?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1024,64 +1024,22 @@ public void run() {
...
-                            }
-                            m_pollTask = null;
-                            return;
-                        }
                         /*
                          * The poll is blocking through the cached future, shouldn't
                          * call poll a second time until a response has been given
                          * which satisfies the cached future.
                          */
                         if (m_pollTask != null) {
```
rdykiel: Move this check at the beginning of the runnable

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1093,6 +1051,94 @@ public void run() {
...
+            // Ensure this first block has a schema
+            BBContainer schemaContainer = m_committedBuffers.pollSchema();
+            if (schemaContainer == null) {
+                try {
+                    pollTask.setException(new IOException("No schema for committedSeqNo " + cont.m_commitSeqNo
+                            + ", discarding buffer (rows may be lost)."));
+                } catch (RejectedExecutionException reex) {
+                    /* Ignore */
+                }
+                if (exportLog.isDebugEnabled()) {
```
rdykiel: Make this an error, at least. Perhaps even a crash voltdb because we should always be able to get a schema

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1093,6 +1051,94 @@ public void run() {
...
+
+    private void initiateMastershipMigration(long nextSeqNo, AckingContainer ackingContainer) {
+        Preconditions.checkNotNull(ackingContainer, "Acking container must be not null.");
+        Pair<Long, Long> gap = m_gapTracker.getFirstGap();
+        assert (gap != null && nextSeqNo >= gap.getFirst() && nextSeqNo <= gap.getSecond());
+        // Mark the recently polled buffer as pending then query other nodes.
+        setPendingContainer(ackingContainer);
+        // If an existing mastership migration is in progress and is before hitting gap,
+        // don't start new migration.
+        if (m_seqNoToDrain > nextSeqNo - 1) {
```
rdykiel: why not check simply m_seqNoToDrain >= nextSeqNo

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1101,53 +1147,50 @@ private synchronized void pollImpl(PollTask pollTask) {
...
-                                m_seqNoToDrain = firstUnpolledSeq - 1;
-                                mastershipCheckpoint(firstUnpolledSeq - 1);
-                            }
-                            break;
-                        }
+                        assert (result == ContinuityCheckResult.HAS_GAP);
+                        final AckingContainer ackingContainer = AckingContainer.of(
+                                this, block, m_committedBuffers, pollTask.forcePollSchema());
+                        initiateMastershipMigration(nextSeqNo, ackingContainer);
+                        return;
```
rdykiel: BUG: what happens to the passed-in pollTask? Shouldn't we set it to null to make the GuestProcessor thread exit?

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1093,6 +1051,94 @@ public void run() {
...
+                setPendingContainer(cont);
+                result = ContinuityCheckResult.NONE;
+            }
+        } else if (result == ContinuityCheckResult.BUFFER_IS_ACKED) {
+            int tuplesDeleted = m_gapTracker.truncate(cont.getLastSeqNo());
+            m_tuplesPending.addAndGet(-tuplesDeleted);
+            cont.internalDiscard();
+        } else {
+            assert (result == ContinuityCheckResult.HAS_GAP);
+            initiateMastershipMigration(nextSeqNo, cont);
```
rdykiel: BUG: the polltask is not satisfied! It's actually leaked here.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1172,21 +1215,11 @@ private synchronized void pollImpl(PollTask pollTask) {
...
-                } else {
-                    schemaContainer = first_unpolled_block.getSchemaContainer();
-                }
-                final AckingContainer ackingContainer =
-                        new AckingContainer(first_unpolled_block.unreleasedContainer(),
-                                schemaContainer,
-                                first_unpolled_block.startSequenceNumber() + first_unpolled_block.rowCount() - 1,
-                                first_unpolled_block.committedSequenceNumber(),
-                                m_migrateRowsDeleter);
+                final AckingContainer ackingContainer = AckingContainer.of(
```
rdykiel: AckingContainer.create()

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1210,128 +1243,48 @@ private synchronized void pollImpl(PollTask pollTask) {
...
+     * @param commitSpHandle the committed SpHandle
+     * @param startTime the time of when the buffer is delivered to export client
+     * @throws RejectedExecutionException - if the stream's task executor cannot accept the task
+     */
+    public void advance(long lastSeqNo, long commitSeqNo, long commitSpHandle, long startTime) {
+        m_es.execute(new Runnable() {
+            @Override
+            public void run() {
+                if (exportLog.isTraceEnabled()) {
+                    exportLog.trace("AckingContainer.discard with sequence number: " + lastSeqNo);
```
rdykiel: Change message to "Advance past sequence number "

rdykiel:
Also, you might want to pull in the changes from my PR, which try to improve the synchnronization.

src/frontend/org/voltdb/export/AckingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+        super(cont.b());
+        m_source = source;
+        m_startSeqNo = startSeqNo;
+        m_lastSeqNo = lastSeqNo;
+        m_commitSeqNo = commitSeq;
+        m_backingCont = cont;
+        m_schemaCont = schemaCont;
+    }
+
+    public static AckingContainer of(ExportDataSource source,
```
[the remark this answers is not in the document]
luyangco: Updated.

src/frontend/org/voltdb/export/AckingContainer.java
```diff
@@ -0,0 +1,137 @@
...
+    final long m_startSeqNo;
+    final long m_lastSeqNo;
+    final long m_commitSeqNo;
+    final BBContainer m_backingCont;
+    BBContainer m_schemaCont;
+    long m_startTime = 0;
+    long m_commitSpHandle = 0;
+    private static VoltLogger EXPORT_LOG = new VoltLogger("EXPORT");
+
+    private AckingContainer(ExportDataSource source, BBContainer cont,
```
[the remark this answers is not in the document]
luyangco: I added ExportDataSource in constructor because when container is discarded it needs a way to tell the owner (ExportDataSource) sends acknowledgment.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: My mistake, accidentally checked in debug message.

luyangco: I rewrite the block to 
```java
if (m_status == StreamStatus.BLOCKED &&
     m_gapTracker.getFirstGap() != null &&
     releaseSeqNo >= m_gapTracker.getFirstGap().getSecond()) {
    assert (!m_mastershipAccepted.get()); //master stream cannot resolve a gap by receiving an ACK from itself, only replica stream can do.
    exportLog.info("Received an ACK for the sequence number after data gap. Set stream status to ACTIVE: " + ExportDataSource.this.toString());
}
```

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Updated.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Escalate to error.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Updated.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Updated.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: The entire check can be removed because stream master can be migrated right away if it is blocked by gap.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: If we call `pollTask.setFuture(null)` to exit GuestProcessor thread here, what if user issues `voltadmin release` or ENABLE_AUTO_GAP_RELEASE is set therefore gap is automatically resolved? In both cases EDS does a poll and assumes GuestProcessor will pick the buffer up.

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Same reason as above.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -126,15 +126,15 @@
     // The committed sequence number is the sequence number of the last row
     // of the last committed transaction pushed up from the EE, and acknowledged by
     // the export client.
-    private long m_committedSeqNo = 0L;
+    private long m_lastCommittedSeqNo = 0L;
```
rdykiel: Why rename this? Name is longer and doesn't bring more info

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1429,7 +1426,9 @@ private void ackImpl(long seq) {
             try {
                 releaseExportBytes(seq);
                 handleDrainedSource();
-                mastershipCheckpoint(seq);
+                if (seq >= m_seqNoToDrain) {
+                    migrateMastershipTo(m_newLeaderHostId);
```
rdykiel: Are we sure m_newLeaderHostId is set at this point? This seems dangerous to me.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -1936,23 +1940,21 @@ public String toString() {
            + " (" + m_status + ", " + (m_mastershipAccepted.get() ? "Master":"Replica") + ")";
     }
 
-    private void mastershipCheckpoint(long seq) {
+    private void migrateMastershipTo(Integer newHostId) {
```
rdykiel: I doubt this new logic simplifies much. I'm not sure it helps fixing the problems we have now so I would recommend deferring to after we have fixed the problems

PR #6148, opened 2019-04-15
Fix hang when importing >1MB VARCHAR/STRING columns

rkennedy-mode:
Fixes an issue in PerPartitionTable.buildTable() where VARCHAR/STRING column values >1MB result in the bulk loader (and by relation the CSV loader) hanging indefinitely during drain while waiting for m_outstandingRowCount to be decremented to zero.

### DIFF
--- a/tests/frontend/org/voltdb/utils/TestCSVLoader.java
+++ b/tests/frontend/org/voltdb/utils/TestCSVLoader.java
@@ -44,7 +44,7 @@ import java.util.List;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
-import com.google.common.base.Strings;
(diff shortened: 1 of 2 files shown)


tests/frontend/org/voltdb/utils/TestCSVLoader.java
```diff
@@ -44,6 +44,7 @@
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.base.Strings;
```
JadenW: Can you change this to import "com.google_voltpatches.common.base.Strings;" instead?

rkennedy-mode: Done.

PR #6152, opened 2019-04-17
ENG-15819 check snapshot for uac

zeemanhe:
If uac kicks in after a rejoining node fails but before its stream snapshot is completed, uac could be blocked by the stream snapshot, which may trigger mp deadlock warning if uac is blocked by more than 5 min.  If the uac requires snapshot isolation and snapshot is in progress, abort the uac.

### DIFF
--- a/src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
+++ b/src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
@@ -26,7 +26,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.zookeeper_voltpatches.CreateMode;
-import org.apache.zookeeper_voltpatches.KeeperException;
 import org.apache.zookeeper_voltpatches.ZooKeeper;
 import org.hsqldb_voltpatches.HSQLInterface;
 import org.voltcore.logging.VoltLogger;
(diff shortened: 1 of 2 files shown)


rdykiel:
Go for it: were you able to exercise the code paths in tests?

manjujames:
Where does it acquire the ZK lock and release it?

zeemanhe:
> Where does it acquire the ZK lock and release it?

do you mean when uac acquirs lock/blocker?

manjujames:
No, the snapshot nodes. Are they pre-existing ZK nodes?

zeemanhe:
yeah, created during snapshotting.

manjujames:
Never mind, I found it.

src/frontend/org/voltdb/sysprocs/UpdateApplicationBase.java
```diff
@@ -544,6 +543,13 @@ public static long getNextGenerationId() {
             compilerLog.info("No more @UpdateApplicationCatalog calls when using DDL mode");
         }
 
+        // catalog update will wait for snap shot to be completed, which will block site thread.
+        if (ccr.requiresSnapshotIsolation && VoltZK.hasHostsSnapshotting(zk)) {
+            VoltZK.removeActionBlocker(zk, VoltZK.catalogUpdateInProgress, hostLog);
+            errMsg = "Snapshot in progress. Please retry catalog update later.";
+            return makeQuickResponse(ClientResponseImpl.GRACEFUL_FAILURE, errMsg);
```
manjujames: I know that nodes_currently_snapshotting is not node under blockers, but I wonder if we can do this check also right at the beginning when we try to acquire uac_blocker similar to other mp_repair, rejoin etc.

zeemanhe: Not all uac are blocked by snapshotting.  Catalog diff engine determines if snapshotting will block uac. So check snapshotting here to avoid unneeded blocking.

manjujames: I see. Makes sense to me then.

PR #6154, opened 2019-04-17
Eng 15850 async quiesce

luyangco:


src/frontend/org/voltdb/export/ExportManager.java
```diff
@@ -577,6 +577,7 @@ public synchronized void updateCatalog(CatalogContext catalogContext, boolean re
             swapWithNewProcessor(catalogContext, generation,
                     connectors, localPartitionsToSites, m_processorConfig);
         }
+        generation.updateCatalog(catalogContext);
```
wweiss-voltdb: Why are you calling this here? Can't you do this work in initializeGenerationFromCatalog?

src/frontend/org/voltdb/export/ExportGeneration.java
```diff
@@ -1128,6 +1132,26 @@ public void processStreamControl(String exportSource, List<String> exportTargets
...
+        // Flip the data source mapping from partition-to-sources to name-to-sources
+        Multimap<String, ExportDataSource> dataSourcesByName = ArrayListMultimap.create();
+        synchronized (m_dataSourcesByPartition) {
+            for (Map<String, ExportDataSource> sourceByName : m_dataSourcesByPartition.values()) {
+                for (Entry<String, ExportDataSource> entry : sourceByName.entrySet()) {
+                    dataSourcesByName.put(entry.getKey(), entry.getValue());
+                }
+            }
+        }
+        for (Entry<String, Collection<ExportDataSource>> entry : dataSourcesByName.asMap().entrySet()) {
```
wweiss-voltdb: This is troubling. It implies that the list of data sources that receive the runnable are potentially not the same list as the one protected by the synchronized block. If you move this code into initializeGenerationFromCatalog, rather than synchronizing in the add and the remove and here, do all the work inside a single synchronized block. That way it is not possible to process an ackMailbox message in between any of these steps.

src/frontend/org/voltdb/export/ExportDataSource.java
```diff
@@ -2080,12 +2054,33 @@ public void setupMigrateRowsDeleter(int batchSize) {
         }
     }
 
+    public void updateCatalog(Table table, long genId) {
+        m_es.execute(new Runnable() {
+            public void run() {
+                if (genId != m_previousGenId) {
```
zeemanhe: I would check the equality first and add a runable if needed to avoid unneeded runs.

src/frontend/org/voltdb/export/ExportGeneration.java
luyangco: Fixed.

src/frontend/org/voltdb/export/ExportManager.java
luyangco: Moved the method into `initializeGenerationFromCatalog`

src/frontend/org/voltdb/export/ExportDataSource.java
luyangco: Done.

PR #6158, opened 2019-04-20
ENG 15253 Still WIP

alexeevm:
Resolve ambiguous columns from USING expression in SELECT for FULL joins.
select I from T1 full join T2 using(I);
is transformed to
select CASE WHEN T1.I is not null THEN T1.I ELSE T2.I as I from T1 full join T2 using(I);


tests/frontend/org/voltdb/plannerv2/TestLogicalRules.java
```diff
@@ -567,4 +567,90 @@ public void testLimit0() {
                 .transform("VoltLogicalValues(tuples=[[]])\n")
                 .pass();
     }
+
+    public void testFullJoinWithUsing() {
```
kaikaizi: Can we have more tests covering `ON`; having qualified/unqualified identifier in the select list; in the `CASE WHEN...` list; in sub-queries; and in multi-way joins?

alexeevm: Sure.
select I as ii from R1 full join R2 using(i);
is currently failing because "I" itself is expanded as alias for the CASE WHEN...

alexeevm: @kaikaizi 
Another interesting case not directly related to the ambiguous join column
select r1.I as ii from R1 full join R2 on R1.I = R2.I group by ii;

Calcite complains about the II column not been found in any table. But, according to the VoltSqlConformance,
aliases of the SELECT columns can be used in the HAVING clause.
On the other hand,
select r1.I as ii from R1 group by ii;
does compile

kaikaizi: Will take a look. Can you add that test case and comment it out?

alexeevm: Done

PR #6160, opened 2019-04-22
ENG-15832: Track close of each ReadCursor

n3world:
Even though an individual read cursor could be closed the read cursor was not
aware of it being closed. Add an explicit close method to ReadCursor and use
the reader closed state to determine when to throw an error on reader methods.
On close always close all readers too.

### DIFF
--- a/src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
+++ b/src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
@@ -90,6 +90,7 @@ public class PersistentBinaryDeque implements BinaryDeque {
         private int m_numRead;
         // If a rewind occurred this is set to the segment id where this cursor was before the rewind
         private long m_rewoundFromId = -1;
+        private boolean m_cursorClosed = false;
 
         public ReadCursor(String cursorId, int numObjectsDeleted) {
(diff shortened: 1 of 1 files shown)


src/frontend/org/voltdb/utils/PersistentBinaryDeque.java
```diff
@@ -1118,17 +1114,29 @@ public synchronized void close() throws IOException {
...
+                // recover a new segment will be opened for writing
+                segment.finalize(true);
+                if (m_usageSpecificLog.isDebugEnabled()) {
+                    m_usageSpecificLog.debug(
+                            "Closed segment " + segment.file() + " (final: " + segment.isFinal() + "), on PBD close");
+                }
             }
-            closeAndDeleteSegment(qs);
         }
         m_segments.clear();
```
luyangco: If `delete == false`, is `m_segments.clear();` still needed?

n3world: I would say it doesn't hurt. It just removes all entries from the segments map and since they are all deleted no need to keep them there. In theory you don't really have to clear it during close ever since the pbd instance is not usable anymore and all references should be dropped

PR #6167, opened 2019-04-24
ENG-15584 shutdown rejoining node without going through agreement process

zeemanhe:
A rejoining node could still wait for stream snapshot from a failed node and end up 60 second no-data timeout.  Shutdown it early to avoid the error warning.


src/frontend/org/voltcore/agreement/AgreementSite.java
```diff
@@ -620,6 +621,13 @@ private void discoverGlobalFaultData(FaultMessage faultMessage) {
                     false,
                     null);
         }
+
+        // Don't try to go through agreement process for a rejoining node.
+        if (VoltDB.instance().rejoining()) {
```
wweiss-voltdb: This violates the assumption of voltcore of no dependencies on cluster state (i.e. VoltCore should be able to stand on it's own).

